PR #1, opened 2016-11-03
Add cecil as a submodule.

erozenfeld:
Fix Mono.Linker.csproj to point to the new location of cecil.

Change a call to XmlWriter.Dispose() (it's declared private on desktop .NET).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "cecil"]
+	path = cecil
+	url = git://github.com/mono/cecil.git
--- a/linker/Mono.Linker.csproj
+++ b/linker/Mono.Linker.csproj
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="14.0">
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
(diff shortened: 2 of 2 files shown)


erozenfeld:
@migueldeicaza PTAL

marek-safar:
XmlWriter::Dispose is public check https://msdn.microsoft.com/en-us/library/ff928369(v=vs.110).aspx

marek-safar:
Also we are using mono fork of cecil instead of official cecil

erozenfeld:
It appears there is an inconsistency in XmlWriter.Dispose accessibility between different versions of .NET Framework. Here is what I see in the reference assembly for 4.0:
```
.method family hidebysig newslot virtual 
        instance void  Dispose(bool disposing) cil managed
{
  // Code size       0 (0x0)
} // end of method XmlWriter::Dispose
```
while in later reference assemblies (4.5, 4.5.1, 4.5.2, 4.6, 4.6.1) I see
```
.method public hidebysig newslot virtual final 
        instance void  Dispose() cil managed
{
  // Code size       0 (0x0)
} // end of method XmlWriter::Dispose
```

The change I made makes the code work against all of these versions.

erozenfeld:
I used
```
url = git://github.com/mono/cecil.git
```
Isn't that a mono fork of cecil? If not, please let me know the correct URL.

marek-safar:
Do you really need to support .net 4.0 runtime? We target at least .net 4.5 in most cases so more issues like that could show up later. Looking at the doc pre-4.5 versions seems to prefer XmlWriter::Close.

Your github url is correct I misread it, sorry for that

erozenfeld:
https://github.com/mono/cecil/blob/master/Mono.Cecil.props sets TargetFrameworkVersion to v4.0 for net_4_0 configurations and there is no net_4_5 configuration. That's why a reference assembly for 4.0 gets picked up during the build. Should we update Mono.Cecil.props to include net_4_5 configurations?

erozenfeld:
I added
```
<TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
to Mono.Linker.csproj
and reverted the change to the call to Dispose().

PR #2, opened 2016-11-09
Fixes for type forwarders.

erozenfeld:
The changes in MarkStep.cs and SweepStep.cs are mostly from https://github.com/xamarin/xamarin-macios/blob/master/tools/linker/MobileMarkStep.cs#L33 and https://github.com/xamarin/xamarin-macios/blob/master/tools/linker/MobileSweepStep.cs#L45 . One additional line
Annotations.Mark (assembly.MainModule);
was added to MarkStep.Process().

The changes in ResolveFromXmlStep.cs are new and allow type forwarders to be specified in xml descriptors.

The change in Mono.Linker.csproj allows opt-out from embedding the default xml descriptors.

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -115,6 +115,22 @@ namespace Mono.Linker.Steps {
 				ProcessQueue ();
 				ProcessVirtualMethods ();
 			}
+
+			// deal with [TypeForwardedTo] pseudo-attributes
+			foreach (AssemblyDefinition assembly in _context.GetAssemblies ()) {
+				if (!assembly.MainModule.HasExportedTypes)
+					continue;
+
+				foreach (var exported in assembly.MainModule.ExportedTypes) {
+					if (!exported.IsForwarder)
+						continue;
+					var type = exported.Resolve ();
+					if (!Annotations.IsMarked (type))
+						continue;
+					Annotations.Mark (exported);
+					Annotations.Mark (assembly.MainModule);
+				}
+			}
 		}
 
 		void ProcessQueue ()
--- a/linker/Mono.Linker.Steps/ResolveFromXmlStep.cs
+++ b/linker/Mono.Linker.Steps/ResolveFromXmlStep.cs
@@ -125,6 +125,23 @@ namespace Mono.Linker.Steps {
 				}
 
 				TypeDefinition type = assembly.MainModule.GetType (fullname);
+
+				if (type == null) {
+					if (assembly.MainModule.HasExportedTypes) {
+						foreach (var exported in assembly.MainModule.ExportedTypes) {
+							if (fullname == exported.FullName) {
+								Annotations.Mark (exported);
+								Annotations.Mark (assembly.MainModule);
+								var resolvedExternal = exported.Resolve ();
+								if (resolvedExternal != null) {
+									type = resolvedExternal;
+									break;
+								}
+							}
+						}
+					}
+				}
+
 				if (type == null)
 					continue;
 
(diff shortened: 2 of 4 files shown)


erozenfeld:
@marek-safar PTAL

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -115,6 +115,22 @@ void Process ()
...
+					continue;
+
+				foreach (var exported in assembly.MainModule.ExportedTypes) {
+					if (!exported.IsForwarder)
+						continue;
+					var type = exported.Resolve ();
+					if (!Annotations.IsMarked (type))
+						continue;
+					Annotations.Mark (exported);
+					Annotations.Mark (assembly.MainModule);
```
radekdoulik: What is the reasoning for adding this line?

erozenfeld: If a type forwarder needs to be kept, then the module/assembly containing the type forwarder also needs to be kept. In .NET Core we have some assemblies that only have type forwarders and we ran into a case where the assembly was incorrectly deleted.

radekdoulik: Thank you for the explanation. I run into similar issue and wasn't sure about the reasoning/history there.

I noticed it was even there in the original MonoTouchMarkStep.cs (https://github.com/xamarin/monotouch/commit/49fc552e1d43bc1dd39b3e12552d16cb9ac7dc53), but I am not sure why it was removed later (there were many moves of this code so it is not easy to follow)

akoeplinger: Found the commit where it was removed: https://github.com/xamarin/monotouch/commit/7c0e0a1df9276b52be92146ee2c7094167e4e2b4

erozenfeld: I'm getting 404 error trying to open xamarin/monotouch@7c0e0a1 . Is there an explanation in that commit of why the line was removed?

akoeplinger: The commit message is: `Only mark the destnation (the type being forwarded to) not the origin. Allows competely removing System.Core (if no types are needed from it)`.

@spouliot do you remember why this change was made?

erozenfeld: I looked some more into code that deals with type forwarders. I believe the intention is that if nothing in an assembly is marked except type forwarders, then all references to the type forwarders in other assemblies should be resolved in SweepStep.SweepReferences and the assembly can be deleted. There is another piece of code in SweepStep.Process that resolves all references in Copy assemblies to make sure they don't have references to type forwarders in deleted assemblies.

I have some scenarios where I don't want this optimization to kick in and need to keep assemblies with type forwarders and not resolve references to them. So, I'll create another pull request where I will add a command line switch to disable this optimization. assembly.MainModule above will only be marked when the switch is passed. Likewise, the code resolving references in  SweepStep.Process and SweepStep.SweepReferences will only run if the new switch is not passed.

erozenfeld: New PR: #12

erozenfeld: New PR: #12

erozenfeld: New PR: #12

erozenfeld: New PR: #12

erozenfeld: New PR: https://github.com/mono/linker/pull/12

erozenfeld: New PR: https://github.com/mono/linker/pull/12

spouliot: Right, things can be a bit easier for XI since we can assume there will be no code generation (or download) so the linker sees everything. That is not a correct assumption outside XI fullAOT compilation. That piece of code might be better left for XI (not mobile) alone - and not used in .net core.

There's also an issue in C9 (Cecil 0.10) where some Scope are not updated (on exported types). That did not happen with C8 (Cecil 0.9) but it might not be related to Cecil (I'm investigating this).

PR #3, opened 2016-11-27
Fix marking of exported nested types.

erozenfeld:
The linker should mark exported nested types if they resolve to
marked types. IsForwarder property is only set on the outermost
enclosing type so nested types weren't marked.


erozenfeld:
@marek-safar PTAL

PR #5, opened 2016-12-08
Fix generated jobs by removing \b from job steps, and update submodules in netci.groovy

sbomer:
The \ wasn't escaped, causing strange behavior with jobs not showing up.
This also fixes build failures due to the submodule for Mono.Cecil not being initialized.


sbomer:
test ci please

PR #9, opened 2017-01-04
Preserve the default constructor and instance fields of return types …

erozenfeld:
…and byref parameter types of PInvokeImpl methods. They are necessary for il stub generation.

@swaroop-sridhar @sbomer @marek-safar PTAL


swaroop-sridhar:
👍

PR #10, opened 2017-01-13
Fix the names of output assembly files

erozenfeld:
The file name of an output assembly of the linker should match the file name of the corresponding input assembly.

Before this fix, OutputStep.GetAssemblyFileName inferred the name of the output file from assembly name and the main module kind.
The name of the input assembly doesn't have to be the same as the name of the corresponding file;
moreover, a Windows or Console main module can live in a dll file instead of an exe file.


erozenfeld:
@swaroop-sridhar @russellhadley @sbomer @marek-safar @jbevain PTAL

swaroop-sridhar:
👍

PR #14, opened 2017-02-03
Bump cecil and use the new setter for ExportedType.Scope. Fixes #51805

spouliot:
Older Cecil (0.9) were saving the scope of the actual type. This changed
with Cecil 0.10 where the ExportedType.Scope is saved, which could lead
to invalid assemblies being saved with the current linker code (which
could not update the scope without the cecil bump)

reference:
https://bugzilla.xamarin.com/show_bug.cgi?id=51805

### DIFF
--- a/linker/Mono.Linker.Steps/SweepStep.cs
+++ b/linker/Mono.Linker.Steps/SweepStep.cs
@@ -175,6 +175,7 @@ namespace Mono.Linker.Steps {
 					if ((td != null) && Annotations.IsMarked (td)) {
 						scope = assembly.MainModule.ImportReference (td).Scope;
 						hash.Add (td, scope);
+						et.Scope = scope;
 					}
 				}
 			}


linker/Mono.Linker.Steps/SweepStep.cs
```diff
@@ -175,6 +175,7 @@ bool ResolveAllTypeReferences (AssemblyDefinition assembly)
 					if ((td != null) && Annotations.IsMarked (td)) {
 						scope = assembly.MainModule.ImportReference (td).Scope;
 						hash.Add (td, scope);
```
erozenfeld: I don't think the line above your change is needed. The scope of the type definition doesn't need to be updated.

erozenfeld: The line above the if is also not needed.

marek-safar: @erozenfeld send separate PR with the clean up

PR #16, opened 2017-02-14
Ignore unresolved type forwarders and assembly references.

erozenfeld:
This change modifies the linker to ignore unresolved type forwarders and assembly references.
This is needed to support .NET Core 2.0 scenarios where mscorlib.dll (and possibly other assemblies) will
contain type forwarders that point to assemblies that may not always be resolved.

### DIFF
--- a/linker/Mono.Linker.Steps/LoadReferencesStep.cs
+++ b/linker/Mono.Linker.Steps/LoadReferencesStep.cs
@@ -48,8 +48,16 @@ namespace Mono.Linker.Steps {
 
 			_references.Add (assembly.Name, assembly);
 
-			foreach (AssemblyNameReference reference in assembly.MainModule.AssemblyReferences)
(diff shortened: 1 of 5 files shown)


erozenfeld:
@marek-safar @jbevain PTAL

marek-safar:
I am not sure the approach of removing any unresolved reference without trace is the best one.

@spouliot how do you report missing references during linking?

erozenfeld:
@spouliot If you are busy, is there anyone else who can answer @marek-safar question? I really want to get  this issue resolved.
/cc:@russellhadley

erozenfeld:
@spouliot Thank you for the review.
I believe you will still get an error if the linker encounters a forwarder that is **used** but can't be resolved. That's why I didn't add a catch inside the resolver but instead added them only when processing forwarder and assembly reference declarations.

PR #17, opened 2017-02-23
Support .NET Core

sbomer:
We would like to run the linker on .NET Core. Some changes have already been made to Cecil to support a netstandard build. This change adds a NetCore.props file, which supports building against netcoreapp1.1, using the netcore_Debug and netcore_Release configurations, which will depend on the netstandard_Debug and netstandard_Release configurations of Cecil, respectively.

The CI build scripts have been updated to use the netcore configuration, but the project file still supports building the old Debug and Release configurations on desktop (in addition to the netcore configurations).

This change includes a DirectoryAssemblyResolver, which is a simple resolver following the example of BaseAssemblyResolver in cecil.

For now, this change will not fall back on the default parameterless ReadSymbols() if there is no SymbolReaderProvider. Once cecil supports ReadSymbols() on .NET Core ([335](https://github.com/jbevain/cecil/pull/335)), we can remove the #if !NET_CORE.

@erozenfeld, would you mind taking a look?


sbomer:
test ci please

sbomer:
Just confirming that the ci job passes with my changes to netci.groovy: [Job output](https://ci.dot.net/job/mono_linker/job/master/job/GenPRTest/job/windows_nt_prtest/1/console). (I'm not sure how long this temporary job will exist, so the link might break.)

linker/Mono.Linker/Driver.cs
```diff
@@ -225,10 +225,12 @@ static IStep ResolveStep (string type)
...
-				string line;
-				while ((line = reader.ReadLine ()) != null)
-					lines.Add (line);
+			using (FileStream fileStream = new FileStream (file, FileMode.Open)) {
+				using (StreamReader reader = new StreamReader (fileStream)) {
+					string line;
+					while ((line = reader.ReadLine ()) != null)
+						lines.Add (line);
+				}
 			}
```
erozenfeld: StreamReader(string) seems to be available in coreclr: https://github.com/dotnet/coreclr/blob/c35e8dbc37e5380f46553510d0368aad04a677d2/src/mscorlib/src/System/IO/StreamReader.cs#L160

marek-safar:
What are all these infrastructure files for and can we move them to separate folder?

sbomer:
Some of the infrastructure files are for ci, and some are scripts that bootstrap a specific cli version used to build the linker with the build scripts (also used in ci). It's still possible to build using the plain project files if desired.

More specifically:
- restore.{cmd/sh}, build.{cmd/sh}: wrapper scripts to restore and build the netcore version of the linker. These pass arguments through run.{cmd/sh}
- run.{cmd/ps1/sh}: bootstraps the cli and build tools using bootstrap.{ps1/sh}, and uses the cli to invoke the run.exe tool with the passed-in arguments. Run.exe is provided by a package in .toolversions.
- config.json: Run.exe reads this to determine how to translate its arguments to arguments to underlying tools.
- .cliversion: version of the cli we use to build
- .toolversions: extra tools used in the build, including run.exe
- bootstrap.{ps1/sh}: This downloads and runs the dotnet-install.ps1 script, which is used to obtain the version set in .cliversion. Extra tools in .toolversions are then restored. This replaces init-tools.cmd.
- netci.groovy: script read by the jenkins jobs to determine what to run in ci


None of these are necessary if all you want to do is restore and build the linker and you already have an installed cli.

It's possible to move these to a separate infrastructure folder, but I'm not sure if it's worth the effort, as the examples I followed all use a similar layout with these files in the root of the repo (see [coreclr](https://github.com/dotnet/coreclr), [corefx](https://github.com/dotnet/corefx), [cli](https://github.com/dotnet/cli), [buildtools](https://github.com/dotnet/buildtools)), and I'd prefer not to diverge too much from them.

linker/Mono.Linker.Steps/BlacklistStep.cs
```diff
@@ -46,7 +46,7 @@ protected override bool ConditionToProcess()
 
 		protected override void Process ()
 		{
-			foreach (string name in Assembly.GetExecutingAssembly ().GetManifestResourceNames ()) {
+			foreach (string name in typeof (BlacklistStep).GetTypeInfo ().Assembly.GetManifestResourceNames ()) {
```
sbomer: See [1784](https://github.com/dotnet/corefx/issues/1784). It looks like some of the removed APIs are not available to build against, despite existing in the coreclr sources.


> It's quite strange that Assembly.Load(byte[]) exists in CoreCLR project: https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Reflection/Assembly.cs#L392
> Can someone explain me why it's there and we can't use it?

> @exyi It actually gets modified to be internal in the build process. It is complicated by various factors around how this code gets shared in many different configurations.

linker/Mono.Linker/LinkContext.cs
```diff
@@ -189,8 +189,12 @@ public void SafeReadSymbols (AssemblyDefinition assembly)
 
 					_annotations.AddSymbolReader (assembly, symbolReader);
 					assembly.MainModule.ReadSymbols (symbolReader);
-				} else
+				}
+#if !NET_CORE
```
sbomer: We won't once it's in mono/cecil.

linker/Mono.Linker/DirectoryAssemblyResolver.cs
```diff
@@ -0,0 +1,118 @@
+using System;
+using System.Reflection;
+using System.Collections.Generic;
+using System.IO;
+using Mono.Collections.Generic;
+using Mono.Cecil;
+
```
sbomer: I agree about copying code. For now this gives us the flexibility to modify our assembly resolution for the scenarios we care about, as discussed with @jbevain. If cecil ends up supporting the resolution methods we need (see [306](https://github.com/jbevain/cecil/issues/306)) it should be easy enough to remove this.

I mostly took this directly from BaseAssemblyResolver.cs, but I must have changed a few of the conventions. I'll fix those - thanks!

linker/Mono.Linker/Driver.cs
sbomer: This is another one where it's not available in the [apis](https://docs.microsoft.com/en-us/dotnet/core/api/system.io.streamreader) despite existing in the source. I believe this is being reintroduced in 2.0 (see [http://stackoverflow.com/questions/40697200/no-streamreader-constructor-accepting-a-string](http://stackoverflow.com/questions/40697200/no-streamreader-constructor-accepting-a-string)), but I tried building against netcoreapp2.0 and that didn't fix the problems I was having with these APIs not existing yet - seems to be a work in progress.

linker/Mono.Linker/LinkContext.cs
```diff
@@ -147,7 +147,7 @@ public AssemblyDefinition Resolve (string name)
 				return assembly;
 			}
 
-			return Resolve (new AssemblyNameReference (name, new Version ()));
+			return Resolve (new AssemblyNameReference (name, new Version (0, 0)));
```
sbomer: Another API that was removed. [https://github.com/dotnet/coreclr/pull/6810](https://github.com/dotnet/coreclr/pull/6810) gives a clue as to how the APIs available to build against differ from the coreclr sources. I don't fully understand when and how these are used, but it looks like everything is compiled against the reference assemblies such as ([mscorlib.cs](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/ref/mscorlib.cs)), which was missing the parameterless constructor. I think these are taken from nuget packages during the build.

linker/Mono.Linker/DirectoryAssemblyResolver.cs
```diff
@@ -0,0 +1,118 @@
...
+			assembly = SearchDirectory(name, new[] { framework_dir }, parameters);
+			if (assembly != null)
+				return assembly;
+
+			throw new AssemblyResolutionException(name);
+		}
+
+		AssemblyDefinition SearchDirectory(AssemblyNameReference name, IEnumerable<string> directories, ReaderParameters parameters)
+		{
+			var extensions = new [] { ".dll" };
```
sbomer: On .NET Core, the main assembly is a .dll, and the main executable (.exe, or no extension on linux) is just a renamed host. I tested this on musicstore.

linker/Mono.Linker/Annotations.cs
```diff
@@ -316,9 +316,10 @@ public void SaveDependencies ()
 			writer.WriteEndElement ();
 			writer.WriteEndDocument ();
 			writer.Flush ();
+#if !NET_CORE
```
sbomer: XmlWriter and GZipStream don't have Close methods on core: [4797](https://github.com/dotnet/corefx/issues/4797). Looking at this again, I think it would be safe to get rid of the calls to Close() altogether, as Dispose() should take care of it for us on desktop: [GZipStream](https://referencesource.microsoft.com/#System/sys/System/IO/compression/GZipStream.cs,18658ab48f2f7361), [XmlWriter](https://referencesource.microsoft.com/#System.Xml/System/Xml/Core/XmlWriter.cs,e5032b905808fabc). What do you think?

linker/Mono.Linker/DirectoryAssemblyResolver.cs
sbomer: I adjusted the formatting in a new commit.

linker/Mono.Linker/DirectoryAssemblyResolver.cs
erozenfeld: That's not always true. You can certainly have a .exe main assembly and run it via corerun.

linker/Mono.Linker/LinkContext.cs
erozenfeld: Can you create a PR for merging it to mono/cecil?

linker/Mono.Linker.Steps/BlacklistStep.cs
erozenfeld: Can you please check with @terrajobst regarding all of the appi's you had to change? I believe they should all be available in 2.0.

linker/Mono.Linker/LinkContext.cs
sbomer: Merged: [bd98f212](https://github.com/mono/cecil/commit/bd98f212fa37a5c868a55f2ed295fe8032f7da7f)

sbomer:
@erozenfeld, I figured out how to properly target netcoreapp2.0 with some pointers from @terrajobst (thanks!). This lets me revert most of the code changes I made. Could you take another look?

erozenfeld:
Sven, thank you for following up on netcoreapp2.0 targeting and reverting code changes.

Regarding location of the scripts: I agree with @marek-safar that they should go into a separate folder. The linker can run in a variety of environments (desktop, mono, .net core, maybe others) and is used by several teams so having files named run.*, restore.*, bootstrap.*, etc. at the root is confusing when they are .NET Core - specific.

sbomer:
@joncham, could you take a look? I'd like to ensure that my change will keep it easy to share this repo across the different scenarios, which you seem to know about. Note that I'll be moving the build infrastructure scripts to a separate directory as per @marek-safar and @erozenfeld 's suggestions.

linker/Mono.Linker/DirectoryAssemblyResolver.cs
mrvoorhe: As an fyi, I made some changes related to the AssemblyResolver in our fork.  We have a custom assembly resolver that we need to use.  Sounds like you have a similar problem.

Here is the PR to our fork that will allow us to rebase on top of the latest linker code.  (I will be upstreaming most of this soon)
https://github.com/Unity-Technologies/linker/pull/1

If you want to see what we did, Search for ILinkerAssemblyResolver.  That's the abstract I came up with to let us hook in our own resolver.

mrvoorhe:
@sbomer This PR looks ok.  There's going to be a minor conflict with the new AseemblyResolver and an ILinkerAssemblyResolver interface I'll be attempting to upstream soon.  But that's not a big deal.

When it comes to the .NET Core build stuff, I don't see us jumping on that in the short term, so no concerns from us regarding the build scripts at this point.

Currently the build paths we use are JAM, xbuild, and various IDE's (VS, Project Rider, VS for Mac).  As long as the .csproj's continue to work we'll be ok.

sbomer:
test ci please

sbomer:
test ci please

sbomer:
test ci please

sbomer:
test ci please

sbomer:
test ci please

sbomer:
test ci please

linker/Mono.Linker/DirectoryAssemblyResolver.cs
```diff
@@ -0,0 +1,113 @@
...
+			get { return reference; }
+		}
+
+		public AssemblyResolutionException (AssemblyNameReference reference)
+			: base (string.Format ("Failed to resolve assembly: '{0}'", reference))
+		{
+			this.reference = reference;
+		}
+	}
+
```
erozenfeld: This may conflict with my PR that was recently merged: https://github.com/mono/linker/pull/16/ . In that PR I'm catching AssemblyResolutionException, which is now ambiguous. Can't we make Mono.Cecil.AssemblyResolutionException available under NET_CORE and use it here?

linker/Mono.Linker/DirectoryAssemblyResolver.cs
```diff
@@ -0,0 +1,113 @@
...
+		{
+			directories.Add (directory);
+		}
+
+		public void RemoveSearchDirectory (string directory)
+		{
+			directories.Remove (directory);
+		}
+
+		public string[] GetSearchDirectories ()
```
erozenfeld: string[] --> string []

linker/Mono.Linker/DirectoryAssemblyResolver.cs
```diff
@@ -0,0 +1,113 @@
...
+			if (name == null)
+				throw new ArgumentNullException ("name");
+			if (parameters == null)
+				parameters = new ReaderParameters ();
+
+			var assembly = SearchDirectory (name, directories, parameters);
+			if (assembly != null)
+				return assembly;
+
+			var framework_dir = Path.GetDirectoryName(typeof (object).GetTypeInfo ().Module.FullyQualifiedName);
```
erozenfeld: Add a space before (typeof.
Also, do you need GetTypeInfo() ?

linker/Mono.Linker/DirectoryAssemblyResolver.cs
sbomer: Sure, we can do that. I opened a PR on cecil: https://github.com/jbevain/cecil/pull/340. I'll update the submodule when it gets merged.

sbomer:
@erozenfeld, I've removed AssemblyResolutionException, and updated cecil now that my [change](https://github.com/mono/cecil/commit/3856cde1591cf09b006df2805ce6584643612016) is in mono/cecil. @marek-safar, could you merge this if everything looks ok to you?

marek-safar:
Merged but someone needs to update netcore CI

sbomer:
Thanks to everyone who helped with this!

@marek-safar, I believe the ci job for the merge failed because it was using the old job definitions. This change already updates the ci jobs for .NET core. I manually kicked off the new jobs that were generated by this merge, and they passed on both ubuntu and windows:
https://ci.dot.net/job/mono_linker/job/master/job/ubuntu/1/
https://ci.dot.net/job/mono_linker/job/master/job/windows_nt/8/

PR #18, opened 2017-02-23
Allow overriding the linker action for the main executable as well.

vargaz:


linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -104,7 +104,7 @@ static void MarkType (LinkContext context, TypeDefinition type)
 			context.Annotations.Pop ();
 		}
 
-		void ProcessExecutable (AssemblyDefinition assembly)
+		void ProcessExecutable (LinkContext context, AssemblyDefinition assembly)
```
kumpera: context is not used here. Typo?

vargaz: Fixed.

linker/Mono.Linker/Driver.cs
```diff
@@ -88,21 +88,24 @@ void Run ()
...
+					case "--version":
 						Version ();
 						break;
-					case 'a':
+					case "--about":
 						About ();
 						break;
+					case "--user-action":
+						context.UserAction = ParseAssemblyAction (GetParam ());
+						break;
```
lambdageek: missing a `case "--core-cation"` case?

vargaz: Fixed.

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -67,7 +67,7 @@ static void SetAction (LinkContext context, AssemblyDefinition assembly, Assembl
 		{
 			TryReadSymbols (context, assembly);
 
-			context.Annotations.SetAction (assembly, action);
+			context.SetAction (assembly, action);
 		}
```
erozenfeld: @vargaz I don't think this change is correct. Consider a case where you have -a SomeCoreAssembly -c link. With your change the action for SomeAssembly will be set to AssemblyAction.Link instead of AssemblyAction.Copy. What was the reason for changing this line?

vargaz: context.SetAction () knows about the per-assembly actions, context.Annotations.SetAction () doesn't.

erozenfeld: With this change the call to `SetAction (context, assembly, AssemblyAction.Copy);` from ProcessLibrary will not always set the action to AssemblyAction.Copy for core assemblies, which seems wrong.

vargaz:
Reverted it until this is resolved.

PR #19, opened 2017-03-01
Fix AssemblyDefinition caching in AssemblyResolver.

erozenfeld:
AssemblyDefinition caching should use string-insensitive string comparisons
so that only a single AssemblyDefinition is created when assembly name is specified
with a different case (e.g., in -a, in -x file, reference in metadata, etc.).

### DIFF
--- a/linker/Mono.Linker/AssemblyResolver.cs
+++ b/linker/Mono.Linker/AssemblyResolver.cs
@@ -28,6 +28,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.IO;
 using Mono.Cecil;
 
@@ -42,7 +43,7 @@ namespace Mono.Linker {
 		}
 
 		public AssemblyResolver ()
-			: this (new Hashtable ())
+			: this (new Dictionary<string, AssemblyDefinition> (StringComparer.OrdinalIgnoreCase))
 		{
 		}
 


erozenfeld:
@marek-safar PTAL

joncham:
Should this be dependent on the file system being case insensitive?

erozenfeld:
This is caching assembly names, not file names.

I checked the ECMA spec (https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf) and I actually see two conflicting statements on this.

In II.6.3: `The dotted name used in .assembly extern shall exactly match the name of the assembly as declared with an .assembly directive, in a case-sensitive manner. (So, even though an assembly might be stored within a file, within a file system that is case-insensitive, the names stored internally within metadata are case-sensitive, and shall match exactly.)`

In II.22: `Matching AssemblyRefs and ModuleRefs to their corresponding Assembly and Module shall be
performed case-blind (see Partition I). However, all other name matches (type, field, method, property,
event) shall be exact – so that this level of resolution is the same across all platforms, whether their OS
is case-sensitive or not.`

marek-safar:
I think the change is sort of correct as it mimics csc behaviour but I am wondering where did you hit this case

erozenfeld:
@marek-safar There are two indexers on Dictionary`2<TKey,Tvalue> and they have different behavior. One is get_Item and it throws on missing keys; the other one is System.Collections.IDictionary.get_Item (interface implementation) and it returns null on missing keys: https://msdn.microsoft.com/en-us/library/bb302460(v=vs.110).aspx
Here we are using the latter indexer.

erozenfeld:
Regarding where I hit this case: if I specify `-a someCoreAssembly` but all the references inside metadata are `SomeCoreAssembly` (note different case on the first letter), two AssemblyDefinitions will be created, one for someCoreAssembly and one for SomeCoreAssembly. That's bad, the effect is that `-a someCoreAssembly` is not honored.

erozenfeld:
@marek-safar Do you have any other concerns with this change?

PR #20, opened 2017-03-03
Fix crlf line endings that have crept in

mrvoorhe:
Some crlf line endings have crept into the code.  Change them back to lf

Add a .gitattributes file that should force lf line endings.  I think you have to re-clone or do `rm .git/index` then `git reset` to get your repo to respect the new settings.


marek-safar:
@akoeplinger is that in line with what we do elsewhere?

akoeplinger:
@mareks we set .sln/.csproj to explicit CLRF in the Mono repo because I think at one point VS couldn't handle LF in those files. If that's no longer the case then I'm fine with forcing LF.

mrvoorhe:
@akoeplinger @marek-safar The csproj is using LF currently.  I have the project open in VS right now and everything is fine with LF

MichalStrehovsky:
@mrvoorhe does Unity have a dependency on the LF line endings on Windows?

This setting is the opposite of how the other dotnet repos are configured (that go with whatever is natural for the platform): https://github.com/dotnet/runtime/blob/2d30f341394f6e1eb8cf9779143ae1709ab83c4e/.gitattributes#L28

LF on Windows still causes annoyances from time to time although some tools finally started accepting it (like Notepad.exe)

mrvoorhe:
It's been awhile.  I can't remember why I did this.  I'd so go with whatever you guys want to do and if it causes Unity issues I'll figure it out on our end.

PR #21, opened 2017-03-06
Fixes & changes and from UnityLinker

mrvoorhe:
This PR contains the changes from rebasing UnityLinker on the latest mono/linker.  On our first pass of rebasing we focused on upstreaming the following
1) Bug fixes
2) Error reporting fixes
3) Extension points & scope changes for UnityLinker classes to derive from and/or get access to members that were previously private.

Regarding (1),  We have tests in our repo for all of these.  Given that we are about to redo the test framework, I don't think its worth the effort to convert our tests into the upstream test format.

Regarding (3), 
* We didn't want to upstream larger features on the first pass.  (ex: Windows Runtime support) 
* We want to handle some situations differently.  (Ex:  For the newly added `ContinueWith` , upstream throws, but UnityLinker will continue & stub the missing type/method)


mrvoorhe:
@marek-safar any thoughts on this PR?

marek-safar:
This is very hard to review. Could you split it to multliple small PRs with rationale for each change so we can discuss them individually.

mrvoorhe:
@marek-safar   I tried to group up changes into meaningful commits so that each commit would be reviewable.  A 1/3rd of the commits are formatting fixes, another 1/3rd are scope changes or other harmless refactorings.  The behavior changes are all prefixed with "Fix" in the commit message, plus the "Undo the fix for [#30488]" commit which is also a behavior change.

I know the size isn't ideal.  I suspect breaking up the PR is going to result in tedious redoing of formatting changes since I didn't make the formatting fix commits correspond with the original commits.  Hindsight is 20/20.

marek-safar:
@mrvoorhe I cannot merge only singe commit or group of commits from PR. Please send 3 PRs instead of 1 big one for easier review

PR #23, opened 2017-03-10
Scope and refactoring changes

mrvoorhe:
This PR contains extension points & scope changes so that our UnityLinker classes can derive from and/or get access to members that were previously inaccessible


marek-safar:
Please split this PR. There are formatting, exception handling, types and method changes mixed together

PR #26, opened 2017-03-14
Don't embed descriptors in netcore builds.

erozenfeld:


erozenfeld:
@sbomer @marek-safar PTAL

sbomer:
LGTM

PR #27, opened 2017-03-14
Fixes for processing interop parameters.

erozenfeld:
This change adds marking of fields for types of parameters of pinvokeimpl and interop methods.
Before this change this was done only for byref types. Also, the code in MarkType marks fields of
types with sequential or explicit layout; however, sometimes auto-layout is used for types of parameters of
internal or pinvokeimpl calls (e.g., when there is just one field).

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -1044,8 +1044,8 @@ namespace Mono.Linker.Steps {
 			MarkCustomAttributes (method.MethodReturnType);
 			MarkMarshalSpec (method.MethodReturnType);
 
-			if (method.IsPInvokeImpl) {
-				ProcessPInvokeImplMethod (method);
+			if (method.IsPInvokeImpl || method.IsInternalCall) {
+				ProcessInteropMethod (method);
 			}
 
 			if (ShouldParseMethodBody (method))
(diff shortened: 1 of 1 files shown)


erozenfeld:
@sbomer @swaroop-sridhar @marek-safar PTAL

sbomer:
LGTM

swaroop-sridhar:
👍

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1079,19 +1079,30 @@ void MarkBaseMethods (MethodDefinition method)
...
+				MarkFields (method.DeclaringType, includeStaticFields);
+			}
 
 			foreach (ParameterDefinition pd in method.Parameters) {
-				if (pd.ParameterType.IsByReference) {
-					TypeReference paramTypeReference = ((ByReferenceType)pd.ParameterType).ElementType;
-					TypeDefinition paramTypeDefinition = ResolveTypeDefinition (paramTypeReference);
-					MarkDefaultConstructor (paramTypeDefinition);
+				TypeReference paramTypeReference = pd.ParameterType;
+				if (paramTypeReference is TypeSpecification) {
```
marek-safar: Shouldn't this be recursive?

mrvoorhe:
@erozenfeld @marek-safar I wrote some new tests for these changes and added them to the new testing infrastructure (still a work in progress).
Here are the tests in case you are curious  https://github.com/Unity-Technologies/linker/commit/d70a1cd8318fbc5920af65c47fc2f228dcbf6429

erozenfeld:
@mrvoorhe Thanks for adding the tests. Looking forward to using the new testing infrastructure!
You have InternalCalls/UnusedFieldsOfTypesWhenHasThisAreNotRemoved.cs but not PInvoke/UnusedFieldsOfTypesWhenHasThisAreNotRemoved.cs. Is that intentional?

mrvoorhe:
@erozenfeld What would the test look like?  I tried this 

```
    class UnusedFieldsOfTypesWhenHasThisAreNotRemoved
    {
        public static void Main()
        {
            A a = new A();
            a.SomeMethod();
        }

        class A
        {
            [Kept]
            private int field1;

            [DllImport("Unused")]
            public extern void SomeMethod();
        }
    }
```
But that doesn't compile because             [DllImport("Unused")] must be on a static extern method.

erozenfeld:
That makes sense, you can't have instance pinvokeimpl method. Thank you for the clarification.

PR #28, opened 2017-03-14
Fix DirectoryAssemblyResolver.

erozenfeld:
DirectoryAssemblyResolver shouldn't add the directory containing the core assembly
for the linker itself to the list of search directories. That core assembly is not necessarily
used by the app being linked so a wrong assembly may be found.


erozenfeld:
@sbomer PTAL

sbomer:
LGTM, thanks!

PR #30, opened 2017-03-15
Sweep assemblies fix

mrvoorhe:
Sweep all of the assemblies that were linked, not all of the assemblies that were resolved.  There may be some assemblies that are resolved, but are not linked.  We ran into this problem with the WinRT support in the UnityLinker.

### DIFF
--- a/linker/Mono.Linker/LinkContext.cs
+++ b/linker/Mono.Linker/LinkContext.cs
@@ -243,9 +243,9 @@ namespace Mono.Linker {
 
 		public AssemblyDefinition [] GetAssemblies ()
 		{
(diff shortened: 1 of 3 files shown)


linker/Mono.Linker/LinkContext.cs
```diff
@@ -243,9 +243,9 @@ static bool IsCore (AssemblyNameReference name)
 
 		public AssemblyDefinition [] GetAssemblies ()
 		{
-			IDictionary cache = _resolver.AssemblyCache;
```
marek-safar: I think this change is wrong GetAssemblies is mostly used to find types which this change breaks for referenced assemblies. You should introduce another method instead

PR #32, opened 2017-03-15
Improve error message

mrvoorhe:
Include the name of the method in the exception that is thrown when there is an exception marking a method

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -153,7 +153,14 @@ namespace Mono.Linker.Steps {
 			while (!QueueIsEmpty ()) {
 				MethodDefinition method = (MethodDefinition) _methods.Dequeue ();
 				Annotations.Push (method);
-				ProcessMethod (method);
+				try
+				{
+					ProcessMethod (method);
+				}
+				catch (Exception e)
+				{
+					throw new Exception (string.Format ("Error processing method: '{0}' in assembly: '{1}'", method.FullName, method.Module.Name), e);
+				}
 				Annotations.Pop ();
 			}
 		}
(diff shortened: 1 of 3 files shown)


linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -153,7 +153,11 @@ void ProcessQueue ()
 			while (!QueueIsEmpty ()) {
 				MethodDefinition method = (MethodDefinition) _methods.Dequeue ();
 				Annotations.Push (method);
-				ProcessMethod (method);
+				try {
+					ProcessMethod (method);
+				} catch (Exception e) {
+					throw new Exception (string.Format ("Error processing method: '{0}' in assembly: '{1}'", method.FullName, method.Module.Name), e);
```
marek-safar: This should be different Exception than just Exception

mrvoorhe: @marek-safar Do you know if there is an existing exception type that would make sense to use?  Or should I introduce a "FatalLinkerException" ?

marek-safar: I think there is no existing exception you could use. What about `MarkerException` ?

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -153,7 +153,11 @@ void ProcessQueue ()
...
+				}
 				Annotations.Pop ();
```
marek-safar: This should go to `finally`

mrvoorhe:
@marek-safar Pop moved into finally and now throwing MarkException

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -153,8 +153,15 @@ void ProcessQueue ()
 			while (!QueueIsEmpty ()) {
 				MethodDefinition method = (MethodDefinition) _methods.Dequeue ();
 				Annotations.Push (method);
-				ProcessMethod (method);
-				Annotations.Pop ();
+				try {
```
marek-safar: You don't need double try, do try/catch/finally instead

PR #33, opened 2017-03-15
Expose some methods

mrvoorhe:
Expose some methods that the unity linker needs to call or override

### DIFF
--- a/linker/Mono.Linker/Driver.cs
+++ b/linker/Mono.Linker/Driver.cs
@@ -172,7 +172,7 @@ namespace Mono.Linker {
 			p.Process (context);
 		}
 
-		static void AddCustomStep (Pipeline pipeline, string arg)
+		protected void AddCustomStep (Pipeline pipeline, string arg)
 		{
 			int pos = arg.IndexOf (":");
 			if (pos == -1) {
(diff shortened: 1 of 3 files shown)


linker/Mono.Linker/Driver.cs
```diff
@@ -172,7 +172,7 @@ void Run ()
 			p.Process (context);
 		}
 
-		static void AddCustomStep (Pipeline pipeline, string arg)
+		protected void AddCustomStep (Pipeline pipeline, string arg)
```
marek-safar: Why did you remove static?

linker/Mono.Linker/Driver.cs
```diff
@@ -233,7 +233,7 @@ static IStep ResolveStep (string type)
 			return (string []) lines.ToArray (typeof (string));
 		}
 
-		static I18nAssemblies ParseI18n (string str)
+		protected I18nAssemblies ParseI18n (string str)
```
marek-safar: Why did you remove static?

linker/Mono.Linker/Driver.cs
mrvoorhe: Side effect of some other changes.  Looks like I can shift some things around and change this back to static.

I will add back static

mrvoorhe:
@marek-safar Fixed.  static added back to both methods

PR #34, opened 2017-03-15
Some new methods

mrvoorhe:
Add some new methods.

* AnnotationsStore.Clear() we need to be able to call to reset the AnnotationsStore instance

* AddBeforeStep is a helper we use that seems like it could be useful upstream some day

### DIFF
--- a/linker/Mono.Linker/Annotations.cs
+++ b/linker/Mono.Linker/Annotations.cs
@@ -56,6 +56,18 @@ namespace Mono.Linker {
 		System.Xml.XmlWriter writer;
 		GZipStream zipStream;
 
+		public void Clear ()
+		{
+			method_actions.Clear ();
+			marked.RemoveWhere (m => m is MethodReference);
+			processed.Clear ();
+			preserved_types.Clear ();
+			preserved_methods.Clear ();
+			public_api.Clear ();
+			override_methods.Clear ();
+			base_methods.Clear ();
+		}
+
 		public void PrepareDependenciesDump ()
 		{
 			PrepareDependenciesDump ("linker-dependencies.xml.gz");
--- a/linker/Mono.Linker/Pipeline.cs
+++ b/linker/Mono.Linker/Pipeline.cs
@@ -65,6 +65,16 @@ namespace Mono.Linker {
 			throw new InvalidOperationException (msg);
 		}
 
+		public void AddStepBefore (IStep target, IStep step)
+		{
+			for (int i = 0; i < _steps.Count; i++) {
+				if (_steps [i] == target) {
+					_steps.Insert (i, step);
+					return;
+				}
+			}
+		}
+
 		public void ReplaceStep (Type target, IStep step)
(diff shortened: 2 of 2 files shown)


linker/Mono.Linker/Annotations.cs
```diff
@@ -56,6 +56,18 @@ public class AnnotationStore {
 		System.Xml.XmlWriter writer;
 		GZipStream zipStream;
 
+		public void Clear ()
```
marek-safar: Why do you need Clear method when you can create a new instance, or is it not really clear method if it clears only half of the state?

mrvoorhe: We have an experimental feature where we do multiple passes.  The way it's implemented is to construct a Pipeline that registers steps multiple times.  We need to clear the annotations between each pass.  I'm guessing the reasoning was that clearing the annotations was preferable over replace the Annotations instance stored on the LinkContext.

Here is what our GetStandardPipeline logic looks like to help show what we are doing.

```
        private static Pipeline GetStandardPipeline()
        {
            Pipeline p = new Pipeline();
            // TODO: Don't run repeatedly. Causes issues
            //for (int i=0; i!=3; i++)
            {
                p.AppendStep(new ClearAnnotationStoreStep());
                p.AppendStep(new LoadReferencesStep());
                p.AppendStep(new BlacklistImplicitlyFromFilesStep());
                p.AppendStep(new ResolveFromMonoBehaviours(LinkerOptions.Api));
                p.AppendStep(new ResolveFromPreserveAttribute());
                p.AppendStep(new PreserveIntPtrFields());

                // The upstream black list step will black list based on the .xml resources embedded in the Mono.Linker
                // executable.  We don't want this as it results in preserving a bunch of stuff that we don't care about
                //p.AppendStep(new BlacklistStep());

                p.AppendStep(new TypeMapStep());
                //p.AppendStep(new DevirtualizeStep());
                p.AppendStep(new UnityMarkStep());
                p.AppendStep(new UnitySweepStep());
                p.AppendStep(new CleanStep());
            }

            // Keep this outside the multi-pass loop.  It only needs to run once at the end,
            // running it multiple times results in duplicate types being added
            p.AppendStep(new AddUnresolvedStubsStep());

            p.AppendStep(new RegenerateGuidStep());
            p.AppendStep(new UnityOutputStep());
            return p;
        }
```

As you can see, we have the loop commented out right now.  I don't know what the issues were, it's on my list of things to revisit.

We can close this PR if you would prefer we wait until we have a need for it.

marek-safar: I think I'll close it and you can create a new one when you know why/how it will be used

marek-safar:
I'll close it you can reopen or create a new one once you how how/why this is going to be used

mrvoorhe:
Sounds good.  I'm going to open a new PR with just AddStepBefore.  That method we do make use of today and we need it defined on Pipeline

PR #36, opened 2017-03-15
Introduce DependenciesFor

mrvoorhe:
So that collection of dependencies can be customized by other linkers

### DIFF
--- a/linker/Mono.Linker/LinkContext.cs
+++ b/linker/Mono.Linker/LinkContext.cs
@@ -30,7 +30,6 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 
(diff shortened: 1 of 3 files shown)


linker/Mono.Linker/LinkContext.cs
```diff
@@ -194,6 +195,19 @@ public void SafeReadSymbols (AssemblyDefinition assembly)
 			} catch {}
 		}
 
+		public virtual ICollection<AssemblyDefinition> DependenciesFor (AssemblyDefinition assembly)
```
marek-safar: Use name ResolveReferences instead

linker/Mono.Linker/LinkContext.cs
```diff
@@ -28,8 +28,9 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.IO;
-
+using System.Linq;
```
marek-safar: Why do you need this using?

mrvoorhe: Removed

mrvoorhe:
@marek-safar Using removed and method renamed

PR #37, opened 2017-03-15
Mark step continue with

mrvoorhe:
In the UnityLinker, we don't error on unresolved types/method.  Instead we have a custom step that stubs in these missing types/methods.  Introducing ContinueWith gives us the extension point that we need to customize the behavior of the MarkStep.

There's no behavior change here.

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -226,21 +226,23 @@ namespace Mono.Linker.Steps {
 		protected virtual void MarkCustomAttribute (CustomAttribute ca)
 		{
 			Annotations.Push (ca);
-			MarkMethod (ca.Constructor);
+			try {
+				MarkMethod (ca.Constructor);
 
-			MarkCustomAttributeArguments (ca);
+				MarkCustomAttributeArguments (ca);
 
-			TypeReference constructor_type = ca.Constructor.DeclaringType;
-			TypeDefinition type = constructor_type.Resolve ();
+				TypeReference constructor_type = ca.Constructor.DeclaringType;
+				TypeDefinition type = constructor_type.Resolve ();
 
-			if (!ContinueWith (type, constructor_type)) {
+				if (!ContinueWith (type, constructor_type)) {
+					return;
+				}
+
+				MarkCustomAttributeProperties (ca, type);
+				MarkCustomAttributeFields (ca, type);
+			} finally {
 				Annotations.Pop ();
-				return;
 			}
-
-			MarkCustomAttributeProperties (ca, type);
-			MarkCustomAttributeFields (ca, type);
-			Annotations.Pop ();
 		}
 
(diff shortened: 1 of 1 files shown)


marek-safar:
I like try/finally change but the rest looks like a hack. You are handling only 2 cases and I still don't know why do you need to stub there methods. Wouldn't be better to have MarkStep with throw/no-throw mode?

mrvoorhe:
There are a few reasons why we stub methods.  Here is an example of one scenario.

User has a project, that is using some asset.  The asset has code which calls Unity's Ping method.  Their project never calls ping at run time.  The user builds their project for WebGL.  Unity Engine doesn't define Ping on WebGL.  Then we go to link the managed code and Ping isn't there, but the user's project is never going to call it.  So we add in a stub method that throws if it is called.

There are a couple other situations, but they basically all revolve around making linking a little more forgiving

Here is what our ContinueWith overrides look like
```
        protected override bool ContinueWith(TypeDefinition type, TypeReference reference)
        {
            if (type == null)
            {
                // Need to get this type's assembly marked right away so that it does not get deleted if there are no other uses of the assembly
                // The add unresolved stub step runs after the SweepStep, so the type will not get added back until after the SweepStep has already flagged
                // any unused assemblies for Delete
                AddUnresolvedStubsStep.MarkAssemblyOfType(((UnityLinkContext)_context), reference);
                ((UnityLinkContext)_context).AddUnresolvedType(reference);
                return false;
            }

            return true;
        }

        protected override bool ContinueWith(MethodDefinition method, MethodReference reference)
        {
            if (method == null)
            {
                ((UnityLinkContext)_context).AddUnresolvedMethod(reference);
                return false;
            }

            return true;
        }

```

Could you elaborate more on what you mean by a throw/no-throw mode?

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1198,5 +1202,23 @@ protected virtual void MarkInstruction (Instruction instruction)
 				break;
 			}
 		}
+
+		protected virtual bool ContinueWith (TypeDefinition type, TypeReference reference)
```
marek-safar: Change this to be `HandleUnresolvedType (TypeReference reference)` and just throw

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1198,5 +1202,23 @@ protected virtual void MarkInstruction (Instruction instruction)
...
+		{
+			if (type == null) {
+				throw new ResolutionException (reference);
+			}
+
+			return true;
+		}
+
+		protected virtual bool ContinueWith (MethodDefinition method, MethodReference reference)
```
marek-safar: Change this to be `HandleUnresolvedMethod (MethodReference reference)` and just throw

mrvoorhe: Done

PR #40, opened 2017-03-15
Scope change extension point

mrvoorhe:
For our WinRT support there are types that we don't want to change the scope on.  This PR gives us the extension point we need to skip scope changing on certain types

### DIFF
--- a/linker/Mono.Linker.Steps/SweepStep.cs
+++ b/linker/Mono.Linker.Steps/SweepStep.cs
@@ -176,6 +176,8 @@ namespace Mono.Linker.Steps {
 			foreach (TypeReference tr in assembly.MainModule.GetTypeReferences ()) {
 				if (hash.ContainsKey (tr))
 					continue;
+				if (tr.IsWindowsRuntimeProjection)
+					continue;
 				var td = tr.Resolve ();
 				IMetadataScope scope = tr.Scope;
 				// at this stage reference might include things that can't be resolved


linker/Mono.Linker.Steps/SweepStep.cs
```diff
@@ -175,6 +175,8 @@ bool ResolveAllTypeReferences (AssemblyDefinition assembly)
 			foreach (TypeReference tr in assembly.MainModule.GetTypeReferences ()) {
 				if (hash.ContainsKey (tr))
 					continue;
+				if (!CanChangeScope (tr))
```
marek-safar: If you don't want to change the scope you should move all the logic to the virtual method and override it with no-op

mrvoorhe: By "all of the logic " do you mean all of the logic in the for loop?  I.e. move the logic in the for loop into a `protected virtual void ChangeTypeScope(TypeReference tr)` ?

mrvoorhe: @marek-safar I looked into creating a "ChangeTypeScope" method and it gets kind of ugly.

First you have the `changes` variable to deal with.  Would probably need to become an out param.

Next there's the `hash` dictionary.  That would have to be passed in.

And lastly, there's the question of what to do with the second scope changing loop for modules with HasExportedTypes.  That loop has almost the same code, but it's not exactly the same.   Different enough that a `ChangeTypeScope` method might become ugly if we tried to reuse it in both places.

I'm not against creating a `ChangeTypeScope`, but I'm not going to go forward with it until you confirm that's the path you want to go down.  This `CanChangeType`  approach has it's advantages, low risk, we don't have to do any code duplication.

marek-safar: Hmm, why do you want to have `CanChangeScope` in one of the loops changing the scope only?

TautvydasZilys: So the reason why this added added was because we had this piece of code instead of that condition:

`if (type.IsWindowsRuntimeProjection) continue;`

Michael assumed that you don't want any of the "Windows Runtime stuff" in upstream so he factored it out into a virtual method that we now overwrite. I think the cleanest solution would be to just keep the original condition to continue if a type is a Windows Runtime projection.

The reason why we cannot change scope when a type is windows runtime projection is because cecil just throws if you try to do that (after all, a projection is not a real reference):

https://github.com/jbevain/cecil/blob/master/Mono.Cecil/TypeReference.cs#L145

marek-safar: I agree that adding the condition is better solution

mrvoorhe:
@marek-safar Updated the PR with the approach suggested by @TautvydasZilys

PR #42, opened 2017-03-16
Add an option to use only entry points visible outside of an assembly as roots.

erozenfeld:
This change adds an option -r that is similar to -a except only types, methods, and
fields visible from outside of an assembly are initially marked. This is useful for running the linker to
remove dead privates.

### DIFF
--- a/linker/Mono.Linker.Steps/LoadI18nAssemblies.cs
+++ b/linker/Mono.Linker.Steps/LoadI18nAssemblies.cs
@@ -77,7 +77,7 @@ namespace Mono.Linker.Steps {
 		void LoadAssembly (AssemblyNameReference name)
 		{
 			AssemblyDefinition assembly = Context.Resolve (name);
-			ResolveFromAssemblyStep.ProcessLibrary (Context, assembly);
+			ResolveFromAssemblyStep.ProcessLibrary (Context, assembly, ResolveFromAssemblyStep.RootVisibility.Any);
 		}
 
 		AssemblyNameReference GetAssemblyName (I18nAssemblies assembly)
--- a/linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
+++ b/linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
@@ -26,6 +26,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
+using System;
 using System.Collections;
 using Mono.Cecil;
 
@@ -35,10 +36,20 @@ namespace Mono.Linker.Steps {
 
 		AssemblyDefinition _assembly;
 		string _file;
+		RootVisibility _rootVisibility;
 
-		public ResolveFromAssemblyStep (string assembly)
+		public enum RootVisibility
+		{
+			Any = 0,
+			PublicAndFamily = 1,
+			PublicAndFamilyAndAssembly = 2
+		}
+
+
+		public ResolveFromAssemblyStep (string assembly, RootVisibility rootVisibility = RootVisibility.Any)
 		{
 			_file = assembly;
+			_rootVisibility = rootVisibility;
 		}
 
 		public ResolveFromAssemblyStep (AssemblyDefinition assembly)
@@ -53,9 +64,13 @@ namespace Mono.Linker.Steps {
 
 			AssemblyDefinition assembly = _assembly ?? Context.Resolve (_file);
 
+			if (_rootVisibility != RootVisibility.Any && HasInternalsVisibleTo (assembly)) {
+				_rootVisibility = RootVisibility.PublicAndFamilyAndAssembly;
+			}
+
 			switch (assembly.MainModule.Kind) {
 			case ModuleKind.Dll:
-				ProcessLibrary (Context, assembly);
+				ProcessLibrary (Context, assembly, _rootVisibility);
 				break;
 			default:
 				ProcessExecutable (assembly);
@@ -75,14 +90,15 @@ namespace Mono.Linker.Steps {
 			context.SafeReadSymbols (assembly);
 		}
 
-		public static void ProcessLibrary (LinkContext context, AssemblyDefinition assembly)
+		public static void ProcessLibrary (LinkContext context, AssemblyDefinition assembly, RootVisibility rootVisibility)
 		{
-			SetAction (context, assembly, AssemblyAction.Copy);
+			var action = rootVisibility == RootVisibility.Any ? AssemblyAction.Copy : AssemblyAction.Link;
+			SetAction (context, assembly, action);
 
 			context.Annotations.Push (assembly);
 
 			foreach (TypeDefinition type in assembly.MainModule.Types)
-				MarkType (context, type);
+				MarkType (context, type, rootVisibility);
 
 			if (assembly.MainModule.HasExportedTypes) {
 				foreach (var exported in assembly.MainModule.ExportedTypes) {
@@ -103,7 +119,7 @@ namespace Mono.Linker.Steps {
 						continue;
 					}
 					context.Resolve (resolvedExportedType.Scope);
-					MarkType (context, resolvedExportedType);
+					MarkType (context, resolvedExportedType, rootVisibility);
 					context.Annotations.Mark (exported);
 					if (context.KeepTypeForwarderOnlyAssemblies) {
 						context.Annotations.Mark (assembly.MainModule);
@@ -114,19 +130,34 @@ namespace Mono.Linker.Steps {
 			context.Annotations.Pop ();
 		}
 
-		static void MarkType (LinkContext context, TypeDefinition type)
+		static void MarkType (LinkContext context, TypeDefinition type, RootVisibility rootVisibility)
 		{
+			bool markType;
+			switch (rootVisibility) {
+				default:
+					markType = true;
+					break;
+
+				case RootVisibility.PublicAndFamily:
+					markType = type.IsPublic || type.IsNestedFamily || type.IsNestedFamilyOrAssembly;
+					break;
+			}
+
+			if (!markType) {
+				return;
+			}
+
 			context.Annotations.Mark (type);
 
 			context.Annotations.Push (type);
 
 			if (type.HasFields)
-				MarkFields (context, type.Fields);
+				MarkFields (context, type.Fields, rootVisibility);
 			if (type.HasMethods)
-				MarkMethods (context, type.Methods);
+				MarkMethods (context, type.Methods, rootVisibility);
 			if (type.HasNestedTypes)
 				foreach (var nested in type.NestedTypes)
-					MarkType (context, nested);
+					MarkType (context, nested, rootVisibility);
 
 			context.Annotations.Pop ();
 		}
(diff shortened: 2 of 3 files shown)


linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -75,14 +77,19 @@ static void TryReadSymbols (LinkContext context, AssemblyDefinition assembly)
 			context.SafeReadSymbols (assembly);
 		}
 
-		public static void ProcessLibrary (LinkContext context, AssemblyDefinition assembly)
+		public static void ProcessLibrary (LinkContext context, AssemblyDefinition assembly, bool publicRoots)
 		{
-			SetAction (context, assembly, AssemblyAction.Copy);
+			if (publicRoots) {
```
marek-safar: No need to duplicate whole SetAction, just conditionally set AssemblyAction

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -138,27 +149,34 @@ void ProcessExecutable (AssemblyDefinition assembly)
 			Annotations.Push (assembly);
 
 			Annotations.Mark (assembly.EntryPoint.DeclaringType);
-			MarkMethod (Context, assembly.EntryPoint, MethodAction.Parse);
+
+			const bool publicRoots = false;
+			MarkMethod (Context, assembly.EntryPoint, MethodAction.Parse, publicRoots);
```
marek-safar: Named argument is preferred instead of const

linker/Mono.Linker/Driver.cs
```diff
@@ -294,6 +296,7 @@ static void Usage (string msg)
 			Console.WriteLine ("                 default is all");
 			Console.WriteLine ("   -x          Link from an XML descriptor");
 			Console.WriteLine ("   -a          Link from a list of assemblies");
+			Console.WriteLine ("   -r          Link from a list of assemblies using public roots");
```
marek-safar: I am not sure how useful this really is. We are not doing it on global scale because it can break a lot of code. Perhaps it'd be more useful to do it for all types by default when linker detects it won't break user code

linker/Mono.Linker.Steps/LoadI18nAssemblies.cs
```diff
@@ -77,7 +77,8 @@ void LoadI18nAssembly (I18nAssemblies asm)
 		void LoadAssembly (AssemblyNameReference name)
 		{
 			AssemblyDefinition assembly = Context.Resolve (name);
-			ResolveFromAssemblyStep.ProcessLibrary (Context, assembly);
+			const bool publicRoots = false;
+			ResolveFromAssemblyStep.ProcessLibrary (Context, assembly, publicRoots);
```
erozenfeld: Done.

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
erozenfeld: Done.

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
erozenfeld: Done.

linker/Mono.Linker/Driver.cs
erozenfeld: @marek-safar Do you mean it can break a lot of code that uses reflection? Sometimes we can make an assumption that a library will only be accessed via public entry points and then we can use this option. We intend to use it to remove dead code from portions of shared libraries.

marek-safar: I am all for removing dead code but the intention of linker is to be smart and not just kill everything. Marshal.SizeOf result value is one of example how what can break badly.

erozenfeld: Sure, but this can break with -a foo.exe (where just Main is used as an entry point) as well if you remove fields from an argument to Marshal.SizeOf. And, of course reflection can also make field removal breaking. -r is just another option that can be supplemented with extra roots in xml files or extra plugin analyses.

marek-safar: I looked again and this will basically skip almost everything, every protected field/method, every nested public type, every destructor is this really intentional ?

erozenfeld: I made several changes to the PR:
1. Family and FamilyOrAssembly methods and fields are marked. That addresses your comment about protected fields/methods. Also, this keeps destructors since they get emitted as protected Finalize methods.
2. NestedFamily and NestedFamilyOrAssembly nested classes are marked. Public nested classes from internal enclosing classes are still not marked since they can't be accessed from other assemblies.
3. If assembly has InternalsVisibleTo attribute, then internal classes, fields, and methods are marked.

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -35,10 +36,20 @@ public class ResolveFromAssemblyStep : ResolveStep {
...
 		AssemblyDefinition _assembly;
 		string _file;
+		RootVisibility _rootVisibility;
 
-		public ResolveFromAssemblyStep (string assembly)
+		public enum RootVisibility
+		{
+			PublicAndFamily,
+			PublicAndFamilyAndAssembly,
+			All
```
marek-safar: Rename it to `Any` with value of 0

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -35,10 +36,20 @@ public class ResolveFromAssemblyStep : ResolveStep {
...
+		}
+
+
+		public ResolveFromAssemblyStep (string assembly, bool visibleRoots)
```
marek-safar: - This breaks existing integration, either create another overload or make the argument default.
- Bool is quite bad it should just use your new enum

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -53,9 +64,15 @@ protected override void Process ()
 
 			AssemblyDefinition assembly = _assembly ?? Context.Resolve (_file);
 
+			bool hasInternalsVisibleTo = HasInternalVisibleTo (assembly);
```
marek-safar: This needs to run only for _rootVisibility != RootVisibility.All

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -53,9 +64,15 @@ protected override void Process ()
...
+
+			if (hasInternalsVisibleTo && (_rootVisibility != RootVisibility.All)) {
```
marek-safar: Redundant braces, please fix it everywhere

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -75,14 +92,15 @@ static void TryReadSymbols (LinkContext context, AssemblyDefinition assembly)
 			context.SafeReadSymbols (assembly);
 		}
 
-		public static void ProcessLibrary (LinkContext context, AssemblyDefinition assembly)
+		public static void ProcessLibrary (LinkContext context, AssemblyDefinition assembly, RootVisibility rootVisibility)
 		{
-			SetAction (context, assembly, AssemblyAction.Copy);
+			var action = (rootVisibility == RootVisibility.All) ? AssemblyAction.Copy : AssemblyAction.Link;
```
marek-safar: Redundant braces

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -114,19 +132,35 @@ public static void ProcessLibrary (LinkContext context, AssemblyDefinition assem
 			context.Annotations.Pop ();
 		}
 
-		static void MarkType (LinkContext context, TypeDefinition type)
+		static void MarkType (LinkContext context, TypeDefinition type, RootVisibility rootVisibility)
 		{
+			bool markType = false;
+			switch (rootVisibility) {
+				case RootVisibility.All:
```
marek-safar: Replace this with `default:` and remove `false` initialization above

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -114,19 +132,35 @@ public static void ProcessLibrary (LinkContext context, AssemblyDefinition assem
...
+				case RootVisibility.PublicAndFamilyAndAssembly:
+					markType = true;
+					break;
+
+				case RootVisibility.PublicAndFamily:
+					markType = (type.IsPublic || type.IsNestedFamily || type.IsNestedFamilyOrAssembly);
```
marek-safar: Redundant braces

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -138,27 +172,73 @@ void ProcessExecutable (AssemblyDefinition assembly)
...
 
-		static void MarkFields (LinkContext context, ICollection fields)
+		static void MarkFields (LinkContext context, ICollection fields, RootVisibility rootVisibility)
 		{
-			foreach (FieldDefinition field in fields)
-				context.Annotations.Mark (field);
+			foreach (FieldDefinition field in fields) {
+				bool markField = false;
+				switch (rootVisibility) {
+					case RootVisibility.All:
```
marek-safar: Replace this with `default:` and remove false initialization above

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -138,27 +172,73 @@ void ProcessExecutable (AssemblyDefinition assembly)
...
+						markField = true;
+						break;
+
+					case RootVisibility.PublicAndFamily:
+						markField = (field.IsPublic || field.IsFamily || field.IsFamilyOrAssembly);
```
marek-safar: Redundant braces

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -138,27 +172,73 @@ void ProcessExecutable (AssemblyDefinition assembly)
...
+						break;
+
+					case RootVisibility.PublicAndFamilyAndAssembly:
+						markField = (field.IsPublic || field.IsFamily || field.IsFamilyOrAssembly || (field.IsAssembly || field.IsFamilyAndAssembly));
```
marek-safar: Redundant braces

linker/Mono.Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -138,27 +172,73 @@ void ProcessExecutable (AssemblyDefinition assembly)
...
 		}
 
-		static void MarkMethod (LinkContext context, MethodDefinition method, MethodAction action)
+		static void MarkMethod (LinkContext context, MethodDefinition method, MethodAction action, RootVisibility rootVisibility)
 		{
-			context.Annotations.Mark (method);
-			context.Annotations.SetAction (method, action);
+			bool markMethod = false;
+			switch (rootVisibility) {
+				case RootVisibility.All:
```
marek-safar: Replace this with `default:`

erozenfeld:
@marek-safar I addressed all of your comments. PTAL

PR #43, opened 2017-03-16
Linker resolver interface

mrvoorhe:
We need to use our own resolver.  This PR introduces an interface for the resolver that the linker uses.  It adds a few extra things to cecil's IAssemblyResolver

There's also a change here to change the return value of CacheAssembly from void to AssemblyDefinition.  I have another PR I will post after this one lands that will make use of the return value.  There is a minor issue today where CacheAssembly is called and then the caller returns the newly loaded assembly rather than the assembly that was already in the cache.

### DIFF
--- a/linker/Mono.Linker/AssemblyResolver.cs
+++ b/linker/Mono.Linker/AssemblyResolver.cs
@@ -67,10 +67,11 @@ namespace Mono.Linker {
 			return asm;
 		}
 
-		public void CacheAssembly (AssemblyDefinition assembly)
+		public AssemblyDefinition CacheAssembly (AssemblyDefinition assembly)
 		{
 			_assemblies [assembly.Name.Name] = assembly;
 			base.AddSearchDirectory (Path.GetDirectoryName (assembly.MainModule.FileName));
+			return assembly;
 		}
 	}
 }
--- a/linker/Mono.Linker/AssemblyResolver.cs
+++ b/linker/Mono.Linker/AssemblyResolver.cs
@@ -67,7 +67,7 @@ namespace Mono.Linker {
 			return asm;
 		}
 
-		public AssemblyDefinition CacheAssembly (AssemblyDefinition assembly)
+		public virtual AssemblyDefinition CacheAssembly (AssemblyDefinition assembly)
 		{
 			_assemblies [assembly.Name.Name] = assembly;
 			base.AddSearchDirectory (Path.GetDirectoryName (assembly.MainModule.FileName));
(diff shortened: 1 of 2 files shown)


marek-safar:
Please rebase your PR I am not sure why why you simple not derive from AssemblyResolver

mrvoorhe:
I used an interface because we have our own base class that implements IAssemblyResolver that we derive from.  Deriving from AssemblyResolver would mean we have to override half of the members of AssemblyResolver.  Introducing an interface seemed easier.

marek-safar:
I am not sure I understand, you change your base class to be AssemblyResolver and you add override to your 2 or 3 methods that's it. An interface is not very good fit here as it'll cause problems later.

mrvoorhe:
@marek-safar I can refactor things on our end so that we can derive from AssemblyResolver.

I removed ILinkerAssemblyResolver and all that remains is the signature change we need to CacheAssembly

marek-safar:
@mrvoorhe I am ok with the change but do you really need it when it's the value you pass in?

mrvoorhe:
Opps.  We'll also need it to be virtual now since we will need to override it.

In our implementation we don't necessarily return the value we passed in.  We check a cache and if the assembly was already in the cache, we return the already cached one instead.

mrvoorhe:
@marek-safar  Sorry, stumbled on some more related changes.  I was reviewing where I left off before paternity leave and found another change on my list of things still needing to go upstream.

https://github.com/Unity-Technologies/linker/commit/34b13cc825271841493d3897088d0631eda93f01

The changes to CacheAssembly were part of my efforts to fix up the file access exceptions we were hitting during our unit tests after we updated our version of cecil (this was awhile ago, before our effort to sync up with the upstream linker).  The other more important fix was https://github.com/mono/linker/pull/71.  I don't recall if this  change to CacheAssembly was necessary to upstream, or if it only happened with our implementation of CacheAssembly, or maybe it wasn't necessary at all.  In any case, I think it's a cleaner approach to give the CacheAssembly call control over which assembly is cached and which assembly is used going forward.

Originally I left https://github.com/Unity-Technologies/linker/commit/34b13cc825271841493d3897088d0631eda93f01 out of this PR because I wanted a PR for just ILinkerAssemblyResolver.  With ILinkerAssemblyResolver gone now, what do you think about me cherry-picking this changeset into this PR?

marek-safar:
It'd makes sense to add it to this PR

mrvoorhe:
Thanks.  Cherry-pick done.

PR #49, opened 2017-03-23
Fix ResolveFromAssemblyStep.MarkType processing of nested types.

erozenfeld:
Pulbic nested types have IsNestedPublic set to true but IsPublic set to false.
This change fixes ResolveFromAssemblyStep.MarkType to take that into account.

Also, NestedPrivate types shouldn't be marked when root visibility is set to
PublicAndFamilyAndAssembly.


erozenfeld:
@marek-safar PTAL

PR #51, opened 2017-03-30
Build Mono.Cecil.Pdb.dll in netcore illink builds.

erozenfeld:
Build Mono.Cecil.Pdb.dll in netcore illink builds to be able to handle native pdbs.


erozenfeld:
@sbomer @marek-safar PTAL

sbomer:
LGTM

PR #52, opened 2017-03-30
Mark methods and fields needed for types with debugger attributes.

erozenfeld:
1. Keep all fields and methods in types with DebuggerDisplayAttribute.
2. Keep all fields and methods in proxy types from DebuggerTypeProxyAttribute.

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -30,6 +30,7 @@
 using System;
 using System.Collections;
 using System.Linq;
+using System.Text.RegularExpressions;
 
 using Mono.Cecil;
 using Mono.Cecil.Cil;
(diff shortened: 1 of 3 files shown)


erozenfeld:
Details about these attributes are here: https://msdn.microsoft.com/en-us/library/x810d419(v=vs.110).aspx
These attributes can also be specified at assembly level, I will address that in a separate PR.

erozenfeld:
@marek-safar PTAL
/cc: @ericstj, @sbomer

joncham:
Do we want to always preserve all of these? Should we have a switch to indicate if a debug/release build is occurring to allow for better debugability vs smallest size?

erozenfeld:
@joncham Yes, that makes sense. I added a -v switch to control this.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -623,6 +629,27 @@ void MarkXmlSchemaProvider (TypeDefinition type, CustomAttribute attribute)
 			MarkNamedMethod (type, method_name);
 		}
 
+		void MarkTypeWithDebuggerDisplayAttribute (TypeDefinition type, CustomAttribute attribute)
+		{
+			if (_context.KeepMembersForDebuggerAttributes) {
+				MarkMethods (type);
```
marek-safar: I think this is just too aggressive it'll persist all methods/fields even if they are unused. You could try to evaluate content of {  } and only fallback to all if it's not simple property/field/method call

erozenfeld: @marek-safar I implemented what you suggested. PTAL

PR #54, opened 2017-03-31
Build improvements

sbomer:
Significant changes:
- Add a fix that allows us to restore linker once, and have this properly restore dependencies.
  - Pick up a similar change in cecil that makes this possible.
  - This allows us to simplify the restore script.

- Add a linker solution for the .NET Core build
  - This contains configuration information, letting us use this as a default in the core build scripts.

- Remove buildtools infrastructure
  - The default configuration to build is now set in the solution, rather than in config.json
  - We still bootstrap the cli.

- Update the cli version to 2.0. 
  - @swaroop-sridhar As per https://github.com/dotnet/cli/issues/6169, this version should be supported on Ubuntu 16.10. 

@erozenfeld, please review.
@russellhadley


sbomer:
@chcosta PTAL

corebuild/dotnet.ps1
```diff
@@ -10,12 +10,11 @@ if ((Test-Path $bootStrapperPath) -eq 0)
         mkdir $toolsLocalPath | Out-Null
     }
 
-    cp "bootstrap.ps1" $bootStrapperPath
+    cp (Join-Path $PSScriptRoot "bootstrap.ps1") $bootStrapperPath
 }
 
-
 # now execute it
-& $bootStrapperPath (Get-Location) $toolsLocalPath | Out-File (Join-Path (Get-Location) "bootstrap.log")
```
chcosta: Change `Join-Path (Get-Location) "bootstrap.log") ` to `Join-Path $PSScriptRoot "bootstrap.log"` for consistency

chcosta: How does a dev build just one vertical instead of all configurations.  ie, just build netcore_release|x64?  I'm not familiar with your teams product / workflow, so maybe it's not an issue.  Do they just use Visual Studio?

sbomer: fixed, thanks!

sbomer: We've been using the dotnet cli to build. "dotnet build linker.sln" would build the default configuration, in this case debug. Extra arguments are used to specify the configuration, and can be used to specify other msbuild properties.

PR #60, opened 2017-04-05
Add an option to use reflection heuristics during marking step.

erozenfeld:
This change adds a -h option that can specify heuristics for keeping
types/methods/fields that may be needed for reflection. Three simple
and coarse-grained heuristic options are added initially (all are off by default):

1. "LdtokenTypeMethods": mark all methods of types whose token is used
in an ldtoken instruction.
2. "LdtokenTypeFields": mark all fields of types whose token is used
in an ldtoken instruction.
3. "InstanceConstructors": mark all instance constructors in types
where an instance member has been marked but none of the instance
constructors have been marked. (The type is likely to be instantiated via CreateInstance).

When -h is specified MarkStep is replaced with MarkStepWithReflectionHeuristics that
derives from MarkStep. The names of heuristics are specified in a comma-separated list.
More heuristics will be added over time and any subset of them can be used for a particular
linker invocation.

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -115,6 +115,7 @@ namespace Mono.Linker.Steps {
 			while (!QueueIsEmpty ()) {
 				ProcessQueue ();
 				ProcessVirtualMethods ();
+				DoAdditionalProcessing ();
 			}
 
 			// deal with [TypeForwardedTo] pseudo-attributes
@@ -599,6 +600,11 @@ namespace Mono.Linker.Steps {
 			return type;
 		}
 
+		// Allow subclassers to mark additional things in the main processing loop
+		protected virtual void DoAdditionalProcessing()
+		{
+		}
+
 		// Allow subclassers to mark additional things when marking a method
 		protected virtual void DoAdditionalTypeProcessing (TypeDefinition method)
 		{
@@ -841,9 +847,9 @@ namespace Mono.Linker.Steps {
 				parameters [1].ParameterType.Name == "StreamingContext";
 		}
 
-		delegate bool MethodPredicate (MethodDefinition method);
+		protected delegate bool MethodPredicate (MethodDefinition method);
 
-		void MarkMethodsIf (ICollection methods, MethodPredicate predicate)
+		protected void MarkMethodsIf (ICollection methods, MethodPredicate predicate)
 		{
 			foreach (MethodDefinition method in methods)
 				if (predicate (method)) {
@@ -860,7 +866,9 @@ namespace Mono.Linker.Steps {
 			return IsConstructor (method) && !method.HasParameters;
 		}
 
-		static bool IsConstructor (MethodDefinition method)
+		protected static MethodPredicate IsConstructorPredicate = new MethodPredicate (IsConstructor);
+
+		protected static bool IsConstructor (MethodDefinition method)
 		{
 			return method.IsConstructor && !method.IsStatic;
 		}
--- a/linker/Mono.Linker.Steps/MarkStepWithReflectionHeuristics.cs
+++ b/linker/Mono.Linker.Steps/MarkStepWithReflectionHeuristics.cs
@@ -1,0 +1,98 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+
+namespace Mono.Linker.Steps {
+
+	public class MarkStepWithReflectionHeuristics : MarkStep {
+
+		protected ICollection<string> _reflectionHeuristics;
+
+		public MarkStepWithReflectionHeuristics (ICollection<string> reflectionHeuristics)
+		{
+			_reflectionHeuristics = reflectionHeuristics;
+		}
+
+		protected override void MarkInstruction (Instruction instruction)
+		{
+			base.MarkInstruction (instruction);
+
+			if (instruction.OpCode == OpCodes.Ldtoken) {
+				object token = instruction.Operand;
+				if (token is TypeReference) {
+					TypeDefinition type = ResolveTypeDefinition (GetOriginalType (((TypeReference) token)));
+					if (type != null) {
+						if (_reflectionHeuristics.Contains ("LdtokenTypeMethods")) {
+							MarkMethods (type);
+						}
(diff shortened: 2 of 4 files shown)


erozenfeld:
@marek-safar PTAL 
/cc:@russellhadley @sbomer

erozenfeld:
@marek-safar ping

marek-safar:
I am not convinced this is good approach. What we have done in past is we fine tuned the linker/tuner to include members which are needed based on used code. I still need to bring this code to the common linker but it basically checks for the context and then manually marks required dependencies.

With your approach I cannot image the end user to use this effectively as there is no easy way to find out why the member/type was removed where it should not therefore the users will probably just enable everything reducing linker power significantly.

There is no easy fix for this but I'd like to see examples of code you are trying to fix with this approach so we can work on more robust solution.

erozenfeld:
@marek-safar 
We used two of these heuristics (LdtokenTypeMethods and InstanceConstructors) when linking the corefx framework. There are hundreds of places that have a pattern similar to this in corefx:

`PropertyInfo f = typeof(TestClass).GetTypeInfo().GetDeclaredProperty(nameof(TestClass.MyProp));`
https://github.com/dotnet/corefx/blob/2117bae3a6c9736033e6d1659821ff3a1069920b/src/System.Runtime/tests/System/Reflection/PropertyInfoTests.cs#L22

There are other places where typeof is called in one method and passed to another method where the name of the member is queried.

LdtokenTypeMethods heuristics helped with these places with a very modest increase in the size of the linked assemblies. It would be a lot of work to identify precisely all of the members that need to be kept and list them in an xml file.

I do intend to add finer grained heuristics that would match various permutations of the pattern above but it wasn't worth doing for linking corefx assemblies since the increase with LdtokenTypeMethods was so small. Even with finer grained heuristics some of the members would have to be specified manually.

When the users get a failure they can try to turn on all or some of the heuristics and see if the errors go away and how much the size of the result is affected. If the increase is not acceptable they have to figure out the exact reason for the failure just like they have to do today. This just gives them additional options. It's a trade-off between getting the maximum possible win and the effort to get there. Either one can be more important depending on the user's goal.

I'll give you an example of a case for InstanceConstructors in a bit.

erozenfeld:
Here is a better example (from the framework itself rather than a test): 
`private static readonly MethodInfo s_getExportFactoryDefinitionsMethod = typeof(ExportFactoryExportDescriptorProvider).GetTypeInfo().GetDeclaredMethod("GetExportFactoryDescriptors");`
https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/System.Composition.Hosting/src/System/Composition/Hosting/Providers/ExportFactory/ExportFactoryExportDescriptorProvider.cs#L16

erozenfeld:
Here is an example where InstanceConstructor heuristic is helpful:

https://github.com/dotnet/corefx/blob/f0ccd9742065ee2aa296448fe3dd38ef6ff77af9/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs#L837

```
public static PropertyFetch FetcherForProperty(PropertyInfo propertyInfo)
{
    if (propertyInfo == null)
        return new PropertyFetch();     // returns null on any fetch.

    var typedPropertyFetcher = typeof(TypedFetchProperty<,>);
    var instantiatedTypedPropertyFetcher = typedPropertyFetcher.GetTypeInfo().MakeGenericType(
        propertyInfo.DeclaringType, propertyInfo.PropertyType);
    return (PropertyFetch)Activator.CreateInstance(instantiatedTypedPropertyFetcher, propertyInfo);
}
```

The constructor for the type created by MakeGenericType is never explicitly called but some instance members are used. The heuristic is to keep all constructors for such types.

marek-safar:
Here is what I think we should do instead. Both examples are trivial and they could be fixed more efficiently by using linker attribute than marking e.g. marking every constructor in every type. We could add private attribute to CoreFX which help us track weak dependencies.

For example in your Activator [example](https://github.com/dotnet/corefx/blob/f0ccd9742065ee2aa296448fe3dd38ef6ff77af9/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticSourceEventSource.cs#L837) the code would look like 

```csharp

[LinkerDependency (typeof (TypedFetchProperty<,>))]
public static PropertyFetch FetcherForProperty(PropertyInfo propertyInfo)
{
  ..activator call to create type of TypedFetchProperty ...
}

```

similarly in your first [example](https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/System.Composition.Hosting/src/System/Composition/Hosting/Providers/ExportFactory/ExportFactoryExportDescriptorProvider.cs#L16)

```csharp

[LinkerDependency (typeof (ExportFactoryExportDescriptorProvider), "GetExportFactoryDescriptors")]
public override IEnumerable<ExportDescriptorPromise> GetExportDescriptors(CompositionContract exportKey, DependencyAccessor definitionAccessor)
{
...used reflection call...
}


```

We then teach linker to recognise this attribute and when marked method has one or more such attributes it marks the dependant types/methods/fields as well

We could also make linker much smarted and recognise call to GetDeclaredMethod and similar and extract the first argument then detect called type from the stack but that's much more complicated solution.

erozenfeld:
What you are suggesting with attributes is equivalent to just listing the types and members to keep in an xml file. While this solution will get the largest size benefit we don't think it's always practical. With corefx we would have to list hundreds of types and members and as I mentioned above that would bring a very small size improvement over using the suggested heuristics.

We do plan to add more fine-grained heuristics (e.g., recognizing all calls to TypeInfo members) eventually. Note that they won't help if the arguments are not constant strings without more sophisticated global analysis.

Since you are pushing back on adding these optional heuristics to the core linker, we'll keep them in our version of the linker for now. I'll close this PR and open another one to merge just MarkStep changes to allow the needed extension points.

marek-safar:
Suggested approach is very different to using .xml descriptions (which btw is intended for very limited scenario of core types). You don't need to list all members but you need to scan the code and adjust it to be linker friendly. That's how it's and you won't avoid that in long run anyway.

The output size differences will be massive if you compare these 2 approaches. Where in your case you include all constructors and all dependencies and so on in cause of linker friendly code you include only what is needed which is the linker goal.

PR #62, opened 2017-04-06
AssemblyCache: Index by Assemblyname, instead of file name

### LINKED ISSUE #61: Linker doesn't like Assembly names to be different from file names
If we have a file containing a module with a different name, the linker crashes in the mark step.

For example, in `A.exe` if we have 
```
.assembly B {}
.class B
       extends [mscorlib]System.Object
{
     ...
     .method private hidebysig static int32  Main(string[] args) cil managed
    {
     ...
    }
```
When A.exe is processed by the linker, the assembly is entered AssemblyCache as `_assemblies["A"]`
However, when the linker is marking the methods of class B, it looks for methods in assembly scope `B`.
It cannot find it in the assembly-cache or a file named `B.exe`/`B.dll` and therefore fails.

I think the correct fix for this is to insert assemblies to the cache by assembly-name, and not file name.

swaroop-sridhar:
Currently, if we have a file containing a module with a different name, the linker crashes in the mark step.

If A.exe contains a module B, when A.exe is processed by the linker, the assembly is entered AssemblyCache as _assemblies["A"]. However, when the linker is marking the methods, it looks for methods in assembly scope `B`. It cannot find it in the assembly-cache or a file named B.exe/B.dll and therefore fails.

This change fixes the problem by inserting assemblies to the cache by assembly-name, and not file name.

Fixes #61

### DIFF
--- a/linker/Mono.Linker/AssemblyResolver.cs
+++ b/linker/Mono.Linker/AssemblyResolver.cs
@@ -61,7 +61,7 @@ namespace Mono.Linker {
 			AssemblyDefinition asm = (AssemblyDefinition) _assemblies [name.Name];
 			if (asm == null) {
 				asm = base.Resolve (name, parameters);
-				_assemblies [name.Name] = asm;
+				_assemblies [asm.Name.Name] = asm;
 			}
 
 			return asm;


swaroop-sridhar:
cc: @erozenfeld

PR #64, opened 2017-04-12
Add missing exclusions for SRE types.

vargaz:


kumpera:
@marek-safar @jonpryor this change is required for 15.2

vargaz:
Note that this needs to be merged to all the various copies of mscorlib.xml in the products.

marek-safar:
This needs to reflected in XA only @jonpryor could you do it?

BTW, I am working on linker unification so this won't be needed in the future

PR #65, opened 2017-04-13
Preserve System.Runtime.CompilerServices. AsyncTaskMethodBuilder:SetN…

vargaz:
…otificationForWaitCompletion () which is called by the debugger. Fixes #55037.


vargaz:
Note that this needs to be merged to all the various copies of mscorlib.xml in the products.

marek-safar:
This is not the way how the linker persists the methods. Please re-assign the bug to me.

PR #68, opened 2017-04-21
Switch non-generic collections to generic

marek-safar:


linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -28,21 +28,21 @@
...
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
 
 using Mono.Cecil;
 using Mono.Cecil.Cil;
+using Mono.Collections.Generic;
```
erozenfeld: This file now has both 
`using System.Collections.Generic;`
and
`using Mono.Collections.Generic;`
Is that intentional?

marek-safar: Yes it's, Cecil has its own set of collections but these don't have to use used for linker code

PR #69, opened 2017-04-24
Fix marking of FunctionPointer Types

swaroop-sridhar:
The MarkStep doesn't handle function pointer types correctly.
For example, the linker crashes while marking the method Fnxn.

  .method int32 Fnxn(method explicit instance int32 *(int32), int32)
  {
    ...
  }

This change fixes the bug in MarkStep.

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -544,11 +544,14 @@ namespace Mono.Linker.Steps {
 
 			reference = GetOriginalType (reference);
 
+			if (reference is FunctionPointerType)
+				return null;
+
 			if (reference is GenericParameter)
 				return null;
 
 //			if (IgnoreScope (reference.Scope))
-//				return;
+//				return null;
 
 			TypeDefinition type = ResolveTypeDefinition (reference);
 
@@ -1029,10 +1032,29 @@ namespace Mono.Linker.Steps {
 				if (mod != null)
 					MarkModifierType (mod);
 
-				type = ((TypeSpecification) type).ElementType;
+				var fnptr = type as FunctionPointerType;
+				if (fnptr != null) {
+					MarkParameters (fnptr);
+					MarkType (fnptr.ReturnType);
+					break; // FunctionPointerType is the original type
+				}
+				else {
+					type = ((TypeSpecification) type).ElementType;
+				}
 			}
 
 			return type;
+		}
+
+		void MarkParameters (FunctionPointerType fnptr)
+		{
+			if (!fnptr.HasParameters)
+				return;
+
+			for (int i = 0; i < fnptr.Parameters.Count; i++)
+			{
+				MarkType (fnptr.Parameters[i].ParameterType);
+			}
 		}
 
 		void MarkModifierType (IModifierType mod)


swaroop-sridhar:
CC @sbomer  @marek-safar

erozenfeld:
Where was the linker crashing? For FunctionPointerTypes ElementType returns null, which should end the loop in GetOriginalType.

swaroop-sridhar:
Yes, GetOriginalType() would return null, which is passed to ResolveTypeDefinition() which crashed.
Also, if GetOriginalType() just returns null, how will the parameter/return types within the FunctionType be marked?

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1029,12 +1029,30 @@ protected TypeReference GetOriginalType (TypeReference type)
 				if (mod != null)
 					MarkModifierType (mod);
 
-				type = ((TypeSpecification) type).ElementType;
+				if ( type is FunctionPointerType) {
```
erozenfeld: Use `as` instead of `is` to avoid casting twice.

erozenfeld: Remove space after `(`.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1029,12 +1029,30 @@ protected TypeReference GetOriginalType (TypeReference type)
...
+					var fnptr = (FunctionPointerType)type;
```
erozenfeld: Add space after cast.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1029,12 +1029,30 @@ protected TypeReference GetOriginalType (TypeReference type)
...
+					MarkParameters(fnptr);
```
erozenfeld: Add space before `(`.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1029,12 +1029,30 @@ protected TypeReference GetOriginalType (TypeReference type)
...
+					type = fnptr.ReturnType;
+				}
+				else {
+					type = ((TypeSpecification)type).ElementType;
```
erozenfeld: Add space after cast.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1029,12 +1029,30 @@ protected TypeReference GetOriginalType (TypeReference type)
...
+				}
 			}
 
 			return type;
 		}
 
+		void MarkParameters(FunctionPointerType fnptr)
```
erozenfeld: Add space before `(`.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1029,12 +1029,30 @@ protected TypeReference GetOriginalType (TypeReference type)
...
+		{
+			if (!fnptr.HasParameters)
+				return;
+
+			for (int i = 0; i < fnptr.Parameters.Count; i++)
+			{
+				MarkType(fnptr.Parameters[i].ParameterType);
```
erozenfeld: Add space before `(`.

swaroop-sridhar:
@erozenfeld updated. Thanks.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1029,12 +1029,30 @@ protected TypeReference GetOriginalType (TypeReference type)
...
-				type = ((TypeSpecification) type).ElementType;
+                var fnptr = type as FunctionPointerType;
+                if (fnptr != null) {
+					MarkParameters (fnptr);
+					type = fnptr.ReturnType;
+				}
+				else {
+					type = ((TypeSpecification) type).ElementType;
+				}
 			}
```
erozenfeld: Indentation is off in the above if/else block.

swaroop-sridhar: Fixed now. Thanks,.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1029,12 +1029,30 @@ protected TypeReference GetOriginalType (TypeReference type)
 				if (mod != null)
 					MarkModifierType (mod);
 
-				type = ((TypeSpecification) type).ElementType;
+				var fnptr = type as FunctionPointerType;
+				if (fnptr != null) {
+					MarkParameters (fnptr);
+					type = fnptr.ReturnType;
```
erozenfeld: I think it would be better to call MarkType (fnptr.ReturnType) here and just return. Function's return type is not the original type of the function.

erozenfeld: Hmm, that won't work because FunctionPointerType.Resolve will return null.

erozenfeld:
BTW, are you encountering this on a mixed-mode assembly?

linker/Mono.Linker.Steps/MarkStep.cs
swaroop-sridhar: See if you like what I have now -- checked for the null case back in MarkType()

swaroop-sridhar:
No, the failure is in a managed assembly.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -544,11 +544,14 @@ protected virtual TypeDefinition MarkType (TypeReference reference)
 
 			reference = GetOriginalType (reference);
 
+			if (reference == null)
+				return null;
```
erozenfeld: I suggest returning unchanged FunctionPointerType from GetOriginalType and changing this to 
```
if (reference is FunctionPointerType)`
          return null;
```

swaroop-sridhar: Done

PR #70, opened 2017-04-26
Support a custom annotation store

mrvoorhe:
We have our own AnnotationsStore that we want to pass in to the LinkContext, this change exposes the ability to create our own instance and pass it in when we construct the LinkContext


marek-safar:
Please rebase on top of master

mrvoorhe:
Done

PR #71, opened 2017-04-26
Dispose of loaded assembly definitions

mrvoorhe:
Not disposing of the assembly definitions leads to file access
exceptions.

This is not causing any issues today since there is nothing that runs
the linker multiple times within the same process.

However, the new test framework does, and the lack of assembly def
disposal leads to file access exceptions during all tests after the
first.

### DIFF
--- a/linker/Mono.Linker/Driver.cs
+++ b/linker/Mono.Linker/Driver.cs
@@ -77,109 +77,110 @@ namespace Mono.Linker {
 		void Run ()
 		{
 			Pipeline p = GetStandardPipeline ();
-			LinkContext context = GetDefaultContext (p);
-			I18nAssemblies assemblies = I18nAssemblies.All;
-			var custom_steps = new List<string> ();
+			using (LinkContext context = GetDefaultContext (p)) {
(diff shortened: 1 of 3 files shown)


linker/Mono.Linker/Driver.cs
```diff
@@ -77,109 +77,110 @@ bool HaveMoreTokens ()
...
-				if (token.Length < 2)
-					Usage ("Option is too short");
-
-				if (! (token [0] == '-' || token [1] == '/'))
-					Usage ("Expecting an option, got instead: " + token);
-
-				if (token [0] == '-' && token [1] == '-') {
-
-					if (token.Length < 3)
+			using (LinkContext context = GetDefaultContext (p)) {
```
mrvoorhe: All I did was add this using and then indent everything.   I double checked the indents and end of lines, they are all good.  I'm not sure why the diff is so ugly.

The code style is fine.  There was one change on line 90 with the space after the !, but I'd argue that was wrong to begin with since the more common pattern used in the code base seems to be not using a space.

PR #72, opened 2017-04-26
FIX - GenericParam count on method not taken into consideration

mrvoorhe:
Could result in a base class method with the same name as a derived method, but a different number of generic parameters , not being stripped when it should have been.

I brought the test we wrote for this fix over to the new test framework branch.  The link to the change set is https://github.com/Unity-Technologies/linker/commit/1f3e74972be7aed20b65ffa4ffe0214d0d0d0fa3.

### DIFF
--- a/linker/Mono.Linker.Steps/TypeMapStep.cs
+++ b/linker/Mono.Linker.Steps/TypeMapStep.cs
@@ -215,6 +215,9 @@ namespace Mono.Linker.Steps {
 			if (cp.Count != mp.Count)
 				return false;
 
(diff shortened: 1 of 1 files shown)


marek-safar:
Could you please add test for it to https://github.com/mono/linker/tree/master/linker/Tests/TestCases/Linker. I am working on running this in CI

mrvoorhe:
I'd rather not.  I think my time is better spent bringing the new test framework upstream.  This is one test of many that upstream does not have that exists in the new test framework.

marek-safar:
Then share it with me and I'll see if I can quickly port them

PR #73, opened 2017-04-27
Changes to allow tracking of all removals during sweep step

mrvoorhe:
Our sweep step does some information gathering and outputs a report at the end.  We need to add some callbacks so that our sweep step can track all removals.

### DIFF
--- a/linker/Mono.Linker.Steps/SweepStep.cs
+++ b/linker/Mono.Linker.Steps/SweepStep.cs
@@ -94,6 +94,8 @@ namespace Mono.Linker.Steps {
 
 				if (type.Name == "<Module>")
 					types.Add (type);
+				else
+					ElementRemoved (type);
 			}
 
 			assembly.MainModule.Types.Clear ();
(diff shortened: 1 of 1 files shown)


linker/Mono.Linker.Steps/SweepStep.cs
```diff
@@ -307,5 +313,13 @@ static bool AreSameReference (AssemblyNameReference a, AssemblyNameReference b)
 
 			return true;
 		}
+
+		protected virtual void ElementRemoved (object element)
```
marek-safar: Change object to IMetadataTokenProvider

marek-safar:
Is it your specific code enhancement or would linker benefit from the report as well?

mrvoorhe:
@marek-safar Fixed.

There is nothing Unity specific about our report logic.  I think it could be useful to others.  It's relatively simplistic which is why I didn't go straight to upstreaming it.

I will send you a sample of the report and the code.  If it's something you'd like upstream, I can bring it upstream.

PR #77, opened 2017-04-28
Don't preserve default constructor and fields when IsImport is true

mrvoorhe:
I updated our fork with the latest upstream changes and one of our Window Runtime related tests started failing.  The behavior change that lead to the failure was the introduction of processing methods with IsInternalCall in https://github.com/mono/linker/pull/27

For ComImport types, we don't need to preserve the default constructor and fields.  I don't fully understand this one, so I'm going to let others who know more about it provide additional commentary :smile:

If this PR ends up landing, I will add new unit tests

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -1328,7 +1328,7 @@ namespace Mono.Linker.Steps {
 		{
 			TypeDefinition returnTypeDefinition = ResolveTypeDefinition (method.ReturnType);
 			const bool includeStaticFields = false;
-			if (returnTypeDefinition != null) {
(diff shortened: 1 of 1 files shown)


TautvydasZilys:
The reason for not needing to preserve default constructors for ComImport types is because those constructors don't get called when RCWs get created as a result of out parameters/return values.

marek-safar:
build

marek-safar:
build

marek-safar:
Why are fields not preserved ?

TautvydasZilys:
I'm not even sure if it's legal to have fields on ComImport types, but in case it is, they cannot be accessed from native code anyway (since the native code gets only the interface pointer with a vtable).

PR #88, opened 2017-05-02
Test framework check and collect tweaking

mrvoorhe:
I'm converting some of our tests to the new scheme that have non-linking related Assertion requirements.

For example, the tests that verify our Sweep Report I made a reference to last week need to Assert a SweepReport.log file is written and do a little light weight processing of the file to verify that it is correct

For tests like this, I want to leverage the TestCaseCollector & TestRunner, but the [Test] itself is going to end up doing all of the Assert's rather than relying on ResultChecker

This PR moves some things around to make writing [Test]'s like this easier.

* Shuffle around collector public API's to make it more reusable

* Pull result checking out of TestRunner.


mrvoorhe:
Conflicts fixed

PR #95, opened 2017-05-04
Test framework xml tests

mrvoorhe:
This PR contains new tests plus fixes for two tests that were failing

* Fix for all members of a type being preserved when a type is preserved with nothing.  This fixes the test UnusedTypeWithPreserveNothingHasMembersRemoved

* Fix for backing fields being preserved when a type is preserved with "methods".  If all of the properties were removed, then there's no reason to keep the backing fields


marek-safar:
build

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1150,14 +1150,20 @@ void ApplyPreserveMethods (MethodDefinition method)
...
-		protected void MarkFields (TypeDefinition type, bool includeStatic)
+		protected void MarkFields (TypeDefinition type, bool includeStatic, bool markBackingFieldsOnlyIfPropertyMarked = false)
 		{
 			if (!type.HasFields)
 				return;
 
 			foreach (FieldDefinition field in type.Fields) {
 				if (!includeStatic && field.IsStatic)
 					continue;
+				if (markBackingFieldsOnlyIfPropertyMarked && field.Name.EndsWith ("k__BackingField")) {
```
marek-safar: Should check for `">k__BackingField"`

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1150,14 +1150,20 @@ void ApplyPreserveMethods (MethodDefinition method)
...
+					var propertyName = field.Name.Substring (1, field.Name.IndexOf ('>') - 1);
+					var propertyDefinition = type.Properties.First (p => p.Name == propertyName);
```
marek-safar: This would throw if there isn't one

mrvoorhe: If the backing field was there, is there a case where the property would not be there?

linker/Mono.Linker.Steps/MarkStep.cs
mrvoorhe: Done

linker/Mono.Linker.Steps/MarkStep.cs
marek-safar: Yes, compiler could reuse this name for something else or some other tool could generated only field with the name.

Also `field.Name.IndexOf ('>')` check fails with code like

```
interface I<T>
{
	int Foo { get; set; }
}

class X : I<int>
{
	int I<int>.Foo { get; set; }
}

```

marek-safar:
build

mrvoorhe:
Fixed up issues with the backing field logic in the MarkStep.  
* Don't throw if the property cannot be found
* Punting on handling cases where the property is an explicit implementation and the type has system types in the generic arguments.  The backing field will have a name of <IFoo<int>.Bar>k__BackingField while obtaining the property definition by string name requires the format IFoo<System.Int32>.Bar.  With some extra effort we could deal with this situation, but it's not worth the effort at the moment.

Added a new test called TypeWithPreserveFieldsHasBackingFieldsOfPropertiesRemoved to cover more complicated backing field situations.

Made dealing with backing fields in test cases easier.  Depending on the [KeptMember] attributes was getting out of control in some of the more complicated cases.  Writing new tests should be easy and getting the exact format of some of these complex signatures right is not easy.  Let the testing infrastructure figure out the names for us.

marek-safar:
whitelist

mrvoorhe:
build

mrvoorhe:
build

mrvoorhe:
@marek-safar I went ahead and implemented support for finding the PropertyDefinitions of backing fields such as `<IFoo<int>.Bar>k__BackingField`.  That was the easiest way to get the tests passing with both mcs & csc.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1150,18 +1150,52 @@ void ApplyPreserveMethods (MethodDefinition method)
...
+					}
+
+					if (propertyDefinition != null && !Annotations.IsMarked (propertyDefinition))
+						continue;
+				}
 				MarkField (field);
 			}
 		}
 
+		private PropertyDefinition SearchPropertiesForMatchingFieldDefinition (FieldDefinition field)
```
marek-safar: could be static

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -1150,18 +1150,52 @@ void ApplyPreserveMethods (MethodDefinition method)
...
+					// However for assemblies compiled with csc we may not be able to because some system types
+					// use a short name in the backing field name.
+					// For example:
+					// Field Name = <IFoo<int>.Bar>k__BackingField
+					// Property Name = IFoo<System.Int32>.Bar
+					// 
+					// Parsing the string and checking for these system types and converting the names would be ugly and error prone.
+					// Instead, let's search the properties and find the one that makes use of the current backing field
+					if (propertyDefinition == null) {
+						propertyDefinition = SearchPropertiesForMatchingFieldDefinition (field);
```
marek-safar: Perhaps it'd be better to maintain only one way of propertyDefinition look up and this one should cover both cases

mrvoorhe:
build

mrvoorhe:
@marek-safar  changes made

PR #96, opened 2017-05-08
Test framework - No test framework attributes in linker input assembly

mrvoorhe:
Compile the input test assembly with all expectation attributes removed so that they could not cause any unintended consequences during linking.

In order to make the assertions we now compile the test case assembly twice.  Once in a new "expectations"
directory with all attributes compiled in.  And then a second time in the "input" directory without the expecation attributes compiled in

The "expectations" assembly is used by the ResultChecker to determine what the expected results are

And the "input" assembly is the one passed to the linker for processing


mrvoorhe:
build

PR #98, opened 2017-05-08
Add .NET Core SDK tasks and targets

sbomer:
This adds a project that produces the NuGet package ILLink.Tasks, which contains MSBuild tasks and targets that integrate the linker with the .NET Core SDK. Using this package in a project file augments the .NET Core publish pipeline with a step that runs the linker.

Eventually these tasks and targets should end up in https://github.com/dotnet/sdk, but for now we're developing them as an extension to the SDK for flexibility. We've been developing them in our internal fork of the linker, but we would like to make them open-source, and it seems like the easiest way is to put them upstream for the moment. I've put everything under the corebuild directory so as not to clutter the repo.


sbomer:
Forgot to mention that this also includes a change that introduces the option to only remove full assemblies, in the first commit.

marek-safar:
Please split not related changes to separate PRs (new options, integration tests)

marek-safar:
whitelist

sbomer:
Split the first commit and opened a new PR: https://github.com/mono/linker/pull/102. Thanks!

PR #99, opened 2017-05-08
Add support for writing pure IL for R2R assemblies

sbomer:
Whenever the linker writes an assembly, we now check the attributes of
each module in the assembly to determine whether it has been
ready-to-run crossgen'd. We detect this by checking the module
attributes: R2R images have the flag COMIMAGE_FLAGS_IL_LIBRARY
(0x00000004) set in the CLI header.

To write back valid IL images, we unset the IL_LIBRARY flag, and set
the IL_ONLY flag. We also adjust the target architecture in the COFF
header, which for R2R images contains an xor of OS values with the
normal architecture values. From coreclr:
```
 #define IMAGE_FILE_MACHINE_NATIVE_NI (IMAGE_FILE_MACHINE_NATIVE ^
  IMAGE_FILE_MACHINE_NATIVE_OS_OVERRIDE)
```
We determine the target architecture from this value and write it back
without the os override used in R2R images.

@swaroop-sridhar please review

### DIFF
--- a/linker/Mono.Linker.Steps/OutputStep.cs
+++ b/linker/Mono.Linker.Steps/OutputStep.cs
@@ -27,14 +27,38 @@
 //
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 
 using Mono.Cecil;
 using Mono.Cecil.Cil;
+using Mono.Cecil.PE;
 
 namespace Mono.Linker.Steps {
 
 	public class OutputStep : BaseStep {
+
+		private Dictionary<UInt16, TargetArchitecture> architectureMap;
+
+		private enum NativeOSOverride {
+			Apple = 0x4644,
+			FreeBSD = 0xadc4,
+			Linux = 0x7b79,
+			NetBSD = 0x1993,
+			Default = 0
+		}
+
+		public OutputStep ()
+		{
+			architectureMap = new Dictionary<UInt16, TargetArchitecture> ();
+			foreach (var os in Enum.GetValues (typeof (NativeOSOverride))) {
+				ushort osVal = (ushort) (NativeOSOverride) os;
+				foreach (var arch in Enum.GetValues (typeof (TargetArchitecture))) {
+					ushort archVal = (ushort) (TargetArchitecture)arch;
+					architectureMap.Add ((ushort) (archVal ^ osVal), (TargetArchitecture) arch);
+				}
+			}
+		}
 
 		protected override void Process ()
 		{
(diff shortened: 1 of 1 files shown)


sbomer:
@erozenfeld and @jbevain if you have a chance, please take a look as well.

linker/Mono.Linker.Steps/OutputStep.cs
```diff
@@ -27,15 +27,39 @@
...
+			FreeBSD = 0xadc4,
+			Linux = 0x7b79,
+			NetBSD = 0x1993,
+			Default = 0
+		}
+
+		public OutputStep ()
+		{
+			architectureMap = new Dictionary<UInt16, TargetArchitecture> ();
+			foreach (var os in Enum.GetValues (typeof(NativeOSOverride))) {
```
erozenfeld: Please add a space between `typeof` and `(` everywhere.

linker/Mono.Linker.Steps/OutputStep.cs
```diff
@@ -27,15 +27,39 @@
...
+				ushort osVal = (ushort)(NativeOSOverride)os;
```
erozenfeld: Add a space after each cast everywhere.

linker/Mono.Linker.Steps/OutputStep.cs
sbomer: Fixed.

linker/Mono.Linker.Steps/OutputStep.cs
sbomer: Fixed.

linker/Mono.Linker.Steps/OutputStep.cs
```diff
@@ -27,15 +27,39 @@
...
 
 using Mono.Cecil;
 using Mono.Cecil.Cil;
+using Mono.Cecil.PE;
 
 namespace Mono.Linker.Steps {
 
 	public class OutputStep : BaseStep {
 
+		private Dictionary<UInt16, TargetArchitecture> architectureMap;
```
erozenfeld: Can we make this map static so that it's computed just once for all assemblies?

linker/Mono.Linker.Steps/OutputStep.cs
```diff
@@ -55,6 +79,29 @@ protected override void ProcessAssembly (AssemblyDefinition assembly)
...
+		}
+
+		void WriteAssembly (AssemblyDefinition assembly, string directory)
+		{
+			foreach (var module in assembly.Modules) {
+				// Write back pure IL even for R2R assemblies
+				if (isR2R (module)) {
+					module.Attributes |= ModuleAttributes.ILOnly;
+					module.Attributes ^= (ModuleAttributes) (uint) 0x04;
+					if (!architectureMap.ContainsKey ((ushort) module.Architecture)) {
```
erozenfeld: Use TryGetValue instead of ContainsKey to avoid repeated lookups.

linker/Mono.Linker.Steps/OutputStep.cs
```diff
@@ -55,6 +79,29 @@ protected override void ProcessAssembly (AssemblyDefinition assembly)
 			OutputAssembly (assembly);
 		}
 
+		private bool isR2R(ModuleDefinition module) {
```
marek-safar: Please use existing coding style.
- no private
- Useful name IsRreadyToRun
- Space between name and parens
- Redundant parens in return statement
- { at new line
- method should be static

linker/Mono.Linker.Steps/OutputStep.cs
```diff
@@ -27,15 +27,39 @@
...
+			Apple = 0x4644,
+			FreeBSD = 0xadc4,
+			Linux = 0x7b79,
+			NetBSD = 0x1993,
+			Default = 0
+		}
+
+		public OutputStep ()
+		{
+			architectureMap = new Dictionary<UInt16, TargetArchitecture> ();
```
marek-safar: Why do you do that when you can compute this on demand in single line?

e.g. `static TargetArchitecture CalculateReadyToRunArchitecture (TargetArchitecture src)`

linker/Mono.Linker.Steps/OutputStep.cs
sbomer: Fixed.

linker/Mono.Linker.Steps/OutputStep.cs
sbomer: I made the suggested change. Thanks!

linker/Mono.Linker.Steps/OutputStep.cs
sbomer: Fixed, thanks!

linker/Mono.Linker.Steps/OutputStep.cs
sbomer: Changed.

PR #102, opened 2017-05-09
Add an option to remove only completely unused assemblies.

sbomer:
Partially used assemblies are kept intact.

The new monolinker flag is -f.

Split from https://github.com/mono/linker/pull/98
@erozenfeld PTAL

### DIFF
--- a/linker/Mono.Linker.Steps/SweepStep.cs
+++ b/linker/Mono.Linker.Steps/SweepStep.cs
@@ -83,6 +83,10 @@ namespace Mono.Linker.Steps {
 				return;
 			}
 
+			if (Context.FullAssemblyOnlyRemoval) {
+				return;
+			}
+
 			var types = new List<TypeDefinition> ();
 
 			foreach (TypeDefinition type in assembly.MainModule.Types) {
--- a/linker/Mono.Linker/Driver.cs
+++ b/linker/Mono.Linker/Driver.cs
@@ -117,6 +117,9 @@ namespace Mono.Linker {
 					case 'o':
 						context.OutputDirectory = GetParam ();
 						break;
+					case 'f':
+						context.FullAssemblyOnlyRemoval = true;
+						break;
 					case 'c':
 						context.CoreAction = ParseAssemblyAction (GetParam ());
 						break;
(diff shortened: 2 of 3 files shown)


marek-safar:
I'd like to understand what is the rationale behind and why does it ignore type forwarders? Also this looks like very specific setting which should not be using simple -f option

sbomer:
Thanks @marek-safar. I believe the rationale is to provide a very conservative mode that keeps an entire assembly whenever any part of it is marked. @erozenfeld may know more.

When you say it ignores type forwarders, I think you're asking why the exported types are still swept in this mode. I don't know if that was the original intention, but for type forwarders I think it comes down to a question of how bad it is to retain dangling type forwards that point to assemblies that have been removed. It looks like this is also still sweeping unmarked exported types (even those that aren't forwarders), which seems incorrect. I think it would make sense not to sweep exported types at all in this mode. @erozenfeld, do you have any thoughts?

Do you have any suggestions for what the option should be named?

erozenfeld:
@marek-safar Can you clarify your question about ignoring type forwarders?

marek-safar:
If I understood the intentions it'd make more sense to extend -c options to have a new value to avoid problems when -c is combined with -f. Perhaps add a new -c option of `preserve` which means keep original assembly when required. Thoughts ?

erozenfeld:
-c option specified the action on core assemblies. The new option applies to all assemblies: do the normal linking analysis but at the end only remove completely unused assemblies and don't modify used assemblies.

marek-safar:
@erozenfeld that was just example, you'll have same problem with -p. I'd rather try to come up with new AssemblyAction than dealing with conflicting option settings like `-c/-p copy -f`. The behaviour is something between AssemblyAction.Link and AssemblyAction.Copy. What about new action called CopyUsed?

erozenfeld:
@marek-safar That sounds reasonable except we don't have a way to specify a default action for all non-core assemblies, i.e., we would have to list each non-core assembly with /p copyused. Should we have another switch to specify the action for all non-core assemblies?

marek-safar:
@erozenfeld I think that makes sense, right now we hardcode it to Link but it seems to me to be useful extension.

erozenfeld:
@swaroop-sridhar @sbomer Can you please follow up on this?

erozenfeld:
#174 and #177 implement the approach discussed here so I'm closing this PR.

PR #103, opened 2017-05-09
Get assemblies from Annotations instead of from the Resolver

mrvoorhe:
The situation we have is that we end up resolving new assemblies during the MarkStep as part of our WinRT support.  That leads to problems during the CleanStep when ProcessAssemblies makes a call to GetAction on one of these assemblies that was resolved during marking because the assembly is not in the `assembly_actions` from Annoations.


mrvoorhe:
There is some concern about the behavior change this PR may introduce.  Instead of upstreaming this, we will keep it in the Unity Linker

PR #105, opened 2017-05-09
Implement core link tests

mrvoorhe:
Implemented expectation attributes for verifying the outcome of types & methods in other assemblies.

Used these new attributes to implement the core link tests which verify that core assemblies are linked when linking of core libraries is requested

Also implemented a test to verify the behavior when the 'copy' mode is used for core libraries.

I didn't write many tests for a variety of reasons, but one is that tests like this will be a bit fragile.  A few tests will give us some level of coverage on the linker behavior.  And they will also verify that the *InAssemblyAttributes remain working.

Unity's plans for the *InAssemblyAttributes is to use them for some more integration type tests that verify a few special preservations we make.

### DIFF
--- a/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/KeptMemberInAssemblyAttribute.cs
+++ b/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/KeptMemberInAssemblyAttribute.cs
@@ -1,0 +1,18 @@
+﻿using System;
+
+
+namespace Mono.Linker.Tests.Cases.Expectations.Assertions {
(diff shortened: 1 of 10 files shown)


mrvoorhe:
The issue with the failing test is that CodeDomProvider does not support the [ConditionalAttribute] with the mono class libraries, but it does with .NET Framework.  So when the tests ran on .NET they were fine.  When the tests ran on OSX the [RemovedTypeInAssembly] attributes are left in the input assembly and then they cause the type to be preserved

After speaking with @marek-safar the [Conditional] bug was recently fixed in mcs but we need to wait for mono 5.2 to get the fix.  For now we will ignore this failing test.

PR #106, opened 2017-05-10
Fix ReferencesAreRemovedWhenAllUsagesAreRemoved, was incorrectly passing

mrvoorhe:
ReferencesAreRemovedWhenAllUsagesAreRemoved was passing when it should have been failing due to a bug in ResultChecker treating [RemovedAssembly] the same as [KeptAssembly].  Once the ResultChecker was fixed, it took a couple more changes before the test would pass.

Add a command line option to disable the blacklist step

For the test cases, disable options that tend to pollute the linked output with a bunch of preservations that are not needed for our test cases

Make assembly and module attribute marking slightly less conservative. Currently we mark them all.  With this change we will not mark assembly and module attributes when their defining module has not been marked after processing all types.

This fixes the test ReferencesAreRemovedWhenAllUsagesAreRemoved when ran against the .NET Framework.  When this test was ran against the .NET Framework, System.dll was being kept around because of the assembly attribute BitmapSuffixInSatelliteAssemblyAttribute, which is itself defined in System.dll.

I did not write a unit test for this case because (a) it would be tricky to do at the moment and (b) we are overly conservative with attributes as a whole and attribute linking need to be reviewed.  When that happens, we can pin down the exact behavior we want.  The goal of this PR is to get ReferencesAreRemovedWhenAllUsagesAreRemoved passing.

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -45,6 +45,7 @@ namespace Mono.Linker.Steps {
 		protected List<MethodDefinition> _virtual_methods;
 		protected Dictionary<TypeDefinition, CustomAttribute> _assemblyDebuggerDisplayAttributes;
 		protected Dictionary<TypeDefinition, CustomAttribute> _assemblyDebuggerTypeProxyAttributes;
+		protected Queue<CustomAttribute> _topLevelAttributes;
 
 		public AnnotationStore Annotations {
 			get { return _context.Annotations; }
@@ -54,6 +55,7 @@ namespace Mono.Linker.Steps {
 		{
 			_methods = new Queue<MethodDefinition> ();
 			_virtual_methods = new List<MethodDefinition> ();
+			_topLevelAttributes = new Queue<CustomAttribute> ();
 
 			_assemblyDebuggerDisplayAttributes = new Dictionary<TypeDefinition, CustomAttribute> ();
 			_assemblyDebuggerTypeProxyAttributes = new Dictionary<TypeDefinition, CustomAttribute> ();
@@ -118,11 +120,13 @@ namespace Mono.Linker.Steps {
 			if (QueueIsEmpty ())
 				throw new InvalidOperationException ("No entry methods");
 
-			while (!QueueIsEmpty ()) {
-				ProcessQueue ();
-				ProcessVirtualMethods ();
-				DoAdditionalProcessing ();
-			}
+			ProcessEntireQueue ();
+
+			// After all types have been processed then we can process the lazily marked attributes
+			ProcessLazyAttributes ();
+
+			// We need to process the queue again in case marking the attributes enqueued more work
+			ProcessEntireQueue ();
 
 			// deal with [TypeForwardedTo] pseudo-attributes
 			foreach (AssemblyDefinition assembly in _context.GetAssemblies ()) {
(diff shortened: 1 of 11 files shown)


mrvoorhe:
Note to self:  Add attribute flags for disabling il8n and blacklist.  Then have only these tests disable them rather than disabling these across the board.

Don't merge yet.

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -45,6 +45,7 @@ public class MarkStep : IStep {
 		protected List<MethodDefinition> _virtual_methods;
 		protected Dictionary<TypeDefinition, CustomAttribute> _assemblyDebuggerDisplayAttributes;
 		protected Dictionary<TypeDefinition, CustomAttribute> _assemblyDebuggerTypeProxyAttributes;
+		protected Queue<CustomAttribute> _lazyAttributes;
```
marek-safar: Please rename it to `topLevelAttributes`

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -461,12 +483,12 @@ protected void MarkAssembly (AssemblyDefinition assembly)
 
 			MarkAssemblySpecialCustomAttributes (assembly);
 
-			MarkCustomAttributes (assembly);
+			LazyMarkCustomAttributes (assembly);
```
marek-safar: For an assembly this should be done in MarkAssemblySpecialCustomAttributes (and possibly rename the method) as we can skip known attributes there

linker/Mono.Linker.Steps/MarkStep.cs
```diff
@@ -483,6 +505,20 @@ void ProcessModule (AssemblyDefinition assembly)
...
 		}
 
+		void ProcessLazyAttributes ()
+		{
+			while (_lazyAttributes.Count != 0) {
+				var customAttribute = _lazyAttributes.Dequeue ();
+
+				// If an attribute's module has not been marked after processing all types in all assemblies,
+				// then surely nothing is using this attribute and there is no need to mark it
+				if (!Annotations.IsMarked (customAttribute.AttributeType.Resolve ().Module))
```
marek-safar: Why check module and not AttributeType ?

mrvoorhe: I wanted to make this change as conservative as possible.  What I'm trying to do is catch the case where a reference to an assembly is preserved simply because that assembly defined & used an assembly or module attribute.

I think how we handle attributes needs to be reviewed and improved.  It might involve different modes, i.e. aggressive vs safe.  I don't know.  Someone needs to sit down and give it some thought.  I may at some point, but it's not a priority at the moment.

Now that you mention it though, maybe I should also check `&& !Annotations.IsMarked(customAttribute.AttributeType)`?  I don't think adding that would hurt, but I don't think it's necessary.  If another assembly was using the attribute, it would cause the attribute to be marked...I think..

linker/Mono.Linker.Steps/MarkStep.cs
mrvoorhe: Do you not see this scheme being useful to attributes on other elements?  I kept the naming generic because I thought we may want to process other attributes after all types have been processed.

linker/Mono.Linker.Steps/MarkStep.cs
mrvoorhe: @marek-safar Is this what you had in mind?  https://github.com/Unity-Technologies/linker/commit/d2837cdead9f41461f80f50ab455ab87b743cf71

linker/Mono.Linker.Steps/MarkStep.cs
mrvoorhe: I went ahead and renamed it

mrvoorhe:
@marek-safar ready for review again.  I made the changes you requested

PR #111, opened 2017-05-12
Improve assembly resolution in ResultChecker

mrvoorhe:
Add support for string type name in *InAssemblyAttributes so that non-public types can be checked

### DIFF
--- a/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/KeptMemberInAssemblyAttribute.cs
+++ b/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/KeptMemberInAssemblyAttribute.cs
@@ -5,13 +5,20 @@ namespace Mono.Linker.Tests.Cases.Expectations.Assertions {
 	[AttributeUsage (AttributeTargets.Class | AttributeTargets.Delegate, AllowMultiple = true, Inherited = false)]
 	public class KeptMemberInAssemblyAttribute : BaseExpectedLinkedBehaviorAttribute {
 		public readonly string AssemblyFileName;
-		public readonly Type Type;
+		public readonly string TypeName;
 		public readonly string [] MemberNames;
 
 		public KeptMemberInAssemblyAttribute (string assemblyFileName, Type type, params string [] memberNames)
 		{
 			AssemblyFileName = assemblyFileName;
-			Type = type;
+			TypeName = type.ToString ();
+			MemberNames = memberNames;
+		}
+
+		public KeptMemberInAssemblyAttribute (string assemblyFileName, string typeName, params string [] memberNames)
+		{
+			AssemblyFileName = assemblyFileName;
+			TypeName = typeName;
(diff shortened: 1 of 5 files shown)


linker/Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -7,31 +7,68 @@
...
-			readerParams.AssemblyResolver = resolver;
-			resolver.AddSearchDirectory (assemblyPath.Parent.ToString ());
-			return new AssemblyContainer (AssemblyDefinition.ReadAssembly (assemblyPath.ToString (), readerParams), resolver);
+			return _originalsResolver.Resolve (new AssemblyNameReference (assemblyPath.FileNameWithoutExtension, null));
+		}
+
+		AssemblyDefinition ResolveOriginalsAssembly (string assemblyName)
+		{
+			var cleanAssemblyName = assemblyName;
+			if (assemblyName.EndsWith (".exe") || assemblyName.EndsWith (".dll"))
```
marek-safar: Use Path.GetFileNameWithoutExtension

linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/KeptMemberInAssemblyAttribute.cs
```diff
@@ -5,13 +5,20 @@ namespace Mono.Linker.Tests.Cases.Expectations.Assertions {
 	[AttributeUsage (AttributeTargets.Class | AttributeTargets.Delegate, AllowMultiple = true, Inherited = false)]
 	public class KeptMemberInAssemblyAttribute : BaseExpectedLinkedBehaviorAttribute {
 		public readonly string AssemblyFileName;
```
marek-safar: Maybe we could just drop the fields completely, we are not using them anyway

mrvoorhe: I've been on the fence about what to do as well.  My concern with dropping the fields is that then the constructor arguments will be highlighted in the IDE as not used and that would be annoying in it's own way.

linker/Tests/TestCasesRunner/ResultChecker.cs
mrvoorhe: Done

linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/KeptMemberInAssemblyAttribute.cs
marek-safar: I am not really concern about IDE but we do something like Debug.Assert (...) or null check with throw

mrvoorhe: Ok.  I will make the change to all of the attributes in a separate PR

PR #118, opened 2017-05-18
Produce linked symbols during publish

sbomer:
- set an illink flag to emit linked pdbs

- rewrite the itemgroup _DebugSymbolsIntermediatePath to include
  linked pdbs in the publish output

@swaroop-sridhar PTAL


sbomer:
I should have mentioned:
Before this change, the linker wasn't creating pdbs when invoked from the msbuild target, and the publish pipeline was just copying the pre-linked dlls to the publish directory. This issue was found while hooking up ILLink.Tasks with our debugger tests.

PR #121, opened 2017-05-30
Base attribute for InAssembly attrs

mrvoorhe:
I have an upcoming PR that will make use of this change.  The upcoming PR will change ResultChecker.IsTypeInOtherAssemblyAssertion to leverage this new base attribute instead of explicitly checking for each of the attributes.  It wasn't worth the hassle of pulling the change to IsTypeInOtherAssemblyAssertion out of my other PR and into this one.

### DIFF
--- a/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/BaseInAssemblyAttribute.cs
+++ b/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/BaseInAssemblyAttribute.cs
@@ -1,0 +1,5 @@
+﻿namespace Mono.Linker.Tests.Cases.Expectations.Assertions
+{
+	public abstract class BaseInAssemblyAttribute : BaseExpectedLinkedBehaviorAttribute {
+	}
+}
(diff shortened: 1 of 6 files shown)


linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/BaseInAssemblyAttribute.cs
```diff
@@ -0,0 +1,5 @@
+﻿namespace Mono.Linker.Tests.Cases.Expectations.Assertions
+{
+	public abstract class BaseInAssemblyAttribute : BaseExpectedLinkedBehaviorAttribute {
```
marek-safar: What's difference between BaseInAssemblyAttribute and BaseExpectedLinkedBehaviorAttribute ?

mrvoorhe: In an up coming PR I will change the following method in ResultChecker from

```
static bool IsTypeInOtherAssemblyAssertion (CustomAttribute attr)
{
    return attr.AttributeType.Name == nameof (RemovedTypeInAssemblyAttribute)
        || attr.AttributeType.Name == nameof (KeptTypeInAssemblyAttribute)
        || attr.AttributeType.Name == nameof (RemovedMemberInAssemblyAttribute)
        || attr.AttributeType.Name == nameof (KeptMemberInAssemblyAttribute);
}
```
to
```
bool IsTypeInOtherAssemblyAssertion (CustomAttribute attr)
{
    return attr.AttributeType.Resolve ().DerivesFrom (nameof (BaseInAssemblyAttribute));
}
```

Then in our Unity.Linker.Tests.Cases.Expectations project we have an attribute [StubbedMemberInAssembly] that derives from BaseInAssemblyAttribute.  With this change ResultChecker.IsTypeInOtherAssemblyAssertion is able to catch our attribute as well.  My next PR will add a few extensibility points to ResultChecker so that having ResultChecker.IsTypeInOtherAssemblyAssertion catch our attribute is actually useful.

PR #123, opened 2017-05-31
Test framework extensibility improvements

mrvoorhe:
I have finished moving over all of our old tests to the new test framework.  The last of the tests were the most complex and I had to make some adjustments to the core of the test framework so that we could extend it to meet our needs.

In ResultChecker

* AdditionalChecking - We use this to do a few extra things before the AssemblyDefinitions are disposed of.

* UnhandledOtherAssemblyAssertion - This is where we put the logic to handle our [StubbedMemberInAssembly] attribute.

* The change to GetOriginalTypeFromInAssemblyAttribute is simply for improved error reporting when you write a bad test

In TestCaseCompiler I had to shift a few things around  to make TestCaseCompiler a bit more reusable.  We have some tests now that use TestCaseCompiler to compile supporting .dll's so we needed the ability to pass in the output name instead of having it hard coded to "test.exe"


mrvoorhe:
build

PR #129, opened 2017-06-06
Set up task package to build for multiple tfms

sbomer:
This will create a package that includes tools for both netcoreapp2.0
and net46, enabling the linker tasks to be used from the desktop
msbuild used in corefx.

@swaroop-sridhar please review.

/cc @ericstj


linker/NetCore.props
```diff
@@ -1,13 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net46</TargetFrameworks>
     <RuntimeFrameworkVersion>2.0.0-beta-001509-00</RuntimeFrameworkVersion>
     <DefineConstants>$(DefineConstants);NET_CORE</DefineConstants>
```
ericstj: Are you sure you want to define `NET_CORE` when building for desktop?

sbomer: Good question. I think the `NET_CORE` here is a bit misleading - it enables the behavior we want in illink but not monolinker. At the moment it just enables simplified assembly resolution logic and changes the usage message. Maybe in the future if we want different behavior on net46 compared to netcoreapp2.0, we should use different defines for the two purposes, but for now I think this is what we want.

sbomer:
@swaroop-sridhar, with this change we won't be able to build ILLink.Tasks on unix (you should still be able to build Mono.Linker). Will that cause problems in any of the ci jobs you set up?

ericstj:
> with this change we won't be able to build ILLink.Tasks on unix 

FWIW the SDK handles this by conditioning TargetFrameworks, then ensuring the package they ship comes from a windows machine.  https://github.com/dotnet/sdk/blob/98b0d1049bd5b140ea29be534b886a6d5653fe89/src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj#L7

linker/NetCore.props
ericstj: I see.  That's unfortunate naming.  I would at least put a comment there for now, or rename the define if its not too much trouble.  Something like "ILLINK" or "FEATURE_ILLINK" etc.

swaroop-sridhar:
We don't use ILLink.tasks on Unix for testing, only the IlLink.dll directly.

linker/NetCore.props
sbomer: Renamed to `FEATURE_ILLINK` - thanks. I also renamed NetCore.props, and the project configurations (currently netcore_Debug/netcore_Release).

sbomer:
I made the change to enable building from Unix. I had to make a copy of the nuspec that only includes the netcoreapp2.0 files, because I couldn't find a good way to pass in the package files that worked for both builds. (I had originally tried to use a single `<file src="publishOutputForAllTFMs/**/*.dll" target="tools" />` for both tfms, but that was putting the dlls in `tools/publishOutputForAllTFMs/<TFM>`, instead of `tools/<TFM>`.)

ericstj:
Consider just disabling the package from Unix since that would be producing an incomplete package.

sbomer:
Right - our netci jobs will just build the tasks without producing a package. Packaging only happens from windows in our internal builds. Thanks!

corebuild/integration/ILLink.Tasks/ILLink.Tasks.csproj
```diff
@@ -41,8 +42,9 @@
          references from being marked as dependencies. -->
     <!-- TODO: Remove the custom .nuspec once the P2P PrivateAssets
          issue is fixed. -->
-   <NuspecFile>ILLink.Tasks.nuspec</NuspecFile>
-    <NuspecProperties>id=$(AssemblyName);authors=$(AssemblyName);description=linker tasks;tfm=$(TargetFramework);</NuspecProperties>
+    <NuspecFile>ILLink.Tasks.nuspec</NuspecFile>
+    <NuspecFile Condition=" '$(OS)' != 'Windows_NT' ">ILLink.Tasks.netcoreapp2.0.nuspec</NuspecFile>
```
ericstj: I wouldn't bother with this, and instead just disable packaging on Unix.

sbomer: Removed this and the .nuspec copy.

sbomer:
@marek-safar, please merge this when you get a chance. Thanks!

PR #130, opened 2017-06-06
Verify behavior with no preserve attribute in XML

joshpeterson:
If a link.xml file has an assembly or type element that does not have
a preserve atttribute, treat that as if it is preserve="all". These two
tests verify the existing behavior.


linker/Tests/Mono.Linker.Tests.Cases/Mono.Linker.Tests.Cases.csproj
```diff
@@ -16,15 +16,15 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <OutputPath>bin\Debug\</OutputPath>
+    <OutputPath>bin/Debug/</OutputPath>
```
mrvoorhe: @joshpeterson Can you undo the slash changes?

joshpeterson: I've corrected this.

marek-safar:
whitelist

PR #133, opened 2017-07-14
Run peverify on the output assemblies

mrvoorhe:
It's relatively easy to change the linker in such a way that produces an assemblies that satisfies the test, but is actually invalid output.

By running peverify/pedump on the output assemblies we get additional coverage to make sure the linker is not producing invalid assemblies.

Note : I took a stab at where to find peverify/pedump.  If the tests fail, I might need direction on where these executables should be on the CI machines

### DIFF
--- a/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/SkipPeVerifyAttribute.cs
+++ b/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/SkipPeVerifyAttribute.cs
@@ -1,0 +1,25 @@
+using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Assertions {
+
+	public enum SkipPeVerifyForToolchian
+	{
+		Pedump
+	}
+
+	[AttributeUsage (AttributeTargets.Class, AllowMultiple = true)]
+	public class SkipPeVerifyAttribute : BaseExpectedLinkedBehaviorAttribute
+	{
+		public SkipPeVerifyAttribute ()
+		{
+		}
+
+		public SkipPeVerifyAttribute (SkipPeVerifyForToolchian toolchain)
+		{
+		}
+
+		public SkipPeVerifyAttribute (string assemblyName)
+		{
+		}
+	}
+}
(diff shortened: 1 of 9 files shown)


mrvoorhe:
Looking into the osx failures such as 

```
Error: Invalid CustomAttribute content row 1 Could not load ctor due to Could not load file or assembly 'Mono.Linker.Tests.Cases.Expectations, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. assembly:Mono.Linker.Tests.Cases.Expectations, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null type:<unknown type> member:<none>
Error count: 1
``` 
I've so far found that,

* These tests pass for our UnityLinker test suite.  In our test suite we use a build of Roslyn to compile the test cases.  When I switch our test suite over to using mcs, I see the same error

* I manually ran pedump on the input & expectation test.exe assemblies and got the same error for both, which means the linker isn't responsible for this error.

* I took the input test.exe over to windows and ran peverify on it and peverify says it's fine.

To me this seems to indicate that it's a bug in pedump that happens to be triggered by code generated from mcs, but not roslyn.

The pedump bug I filed already is about CustomAttributes https://bugzilla.xamarin.com/show_bug.cgi?id=58168 maybe this is a similar issue?

@marek-safar What is your assessment?

marek-safar:
The actual error now is `Invalid CustomAttribute content row 1 Could not load ctor due to Could not load file or assembly 'Mono.Linker.Tests.Cases.Expectations, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. assembly:Mono.Linker.Tests.Cases.Expectations, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null type:<unknown type> member:<none>
Error count: 1` which looks like an issue with missing dependency.

Perhaps we are not copying everything (pedump loads even code/assembly which is not executed)

mrvoorhe:
Mono.Linker.Tests.Cases.Expectations.dll is in the same directory as test.exe.  

Mono.Linker.Tests.Cases.Expectations.dll only depends on mscorlib, so I don't see how a dependency could be missing.

I'm out of ideas and I don't view having peverify working upstream for OSX as a particularly high priority for us at the moment.  Given that peverify.exe says the executable is fine, I don't see how it's impossible to rule out a pedump bug/quirk which has more worried I could end up spinning my wheels and wasting a bunch of time for little gain.

I see a couple paths forward.
1) Someone else looks into this error to see if they can figure it out
2) Punt.  I'll change the PR so that this new peverification step is windows only (or maybe just not on OSX, it looks like it worked on linux).  We can worry about fixing this for OSX at a later point.
3) Decline the PR and I'll let this extra verification live in the UnityLinker test suite.

Let me know how you would like to proceed.

marek-safar:
build

marek-safar:
I rebased it manually

PR #138, opened 2017-08-04
Work around GetAssemblyName issue

sbomer:
GetAssemblyName currently sometimes fails for crossgen'd assemblies for a different platform (see https://github.com/dotnet/coreclr/issues/13230). Until that is fixed, we use cecil to determine whether an assembly is managed.

Also includes some formatting fixes for the CompareSizes output.

@swaroop-sridhar, @erozenfeld please review.


erozenfeld:
LGTM

PR #141, opened 2017-08-17
Testing support for adding the KeepTypeForwarderOnlyAssemblies

mrvoorhe:


linker/Tests/Mono.Linker.Tests.Cases.Expectations/Metadata/KeepTypeForwarderOnlyAssembliesAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Metadata {
+	[AttributeUsage (AttributeTargets.Class)]
+	public sealed class KeepTypeForwarderOnlyAssembliesAttribute : BaseMetadataAttribute {
```
marek-safar: I think it's be easier to have one attribute with enum value for all boolean like settings instead of X individual attributes

mrvoorhe: Agreed, the current pattern is tedious.  I'd rather not deal with it in this PR.  I have a rather large PR in progress to the UnityLinker and these are the upstream changes it depends on.  I'd rather not be forced into adding refactoring changes into my UnityLinker PR.

How about I improve the options in a separate PR?  That way I also have chance to sync up our linker with your recent BlackList change and this PR before doing refactoring.

mrvoorhe:
build

mrvoorhe:
I do not know how these changes could have caused the ` Invalid CustomAttribute content row 1 Could not load ctor due to Could not load file or assembly 'Mono.Linker.Tests.Cases.Expectations,....` error.  This feels like the same issue we hit when I originally added the PeVerify stuff.  @marek-safar has anything changed on the build machine?

linker/Tests/Mono.Linker.Tests.Cases.Expectations/Metadata/KeepTypeForwarderOnlyAssembliesAttribute.cs
marek-safar: Agreed

marek-safar:
Where do you see `Invalid CustomAttribute content row 1 Could not load ctor due to Could not load file or assembly 'Mono.Linker.Tests.Cases.Expectations,....` error?

PR #147, opened 2017-08-21
Adds more type-forwarder tests

marek-safar:


linker/Tests/Mono.Linker.Tests.Cases/TypeForwarding/MissingTargetReference.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.TypeForwarding
+{
+	[Define ("IL_ASSEMBLY_AVAILABLE")]
+	[SetupCompileBefore ("TypeForwarderMissingReference.dll", new [] { "Dependencies/TypeForwarderMissingReference.il" })]
+	[SetupLinker ("-p link TypeForwarderMissingReference.dll")]
```
mrvoorhe: What do you think of going with `SetupLinkerAttribute(string flag, params string[] values)` and accepting multiple attributes, instead of a single string?  There are a few reasons why I think that will work better.

* You added a call to .Split(' ') in the test framework code, that's not going to work correct if there's ever a space in a file path (admittedly a low chance).  But in generally I'd prefer to avoid getting the test framework involved in linker argument parsing.

* Our UnityLinker doesn't use the same command line argument parsing logic as upstream.  The upstream logic is quite bare bones.  Instead we use our own options parsing logic, which is itself based off https://github.com/mono/mono/blob/master/mcs/class/Mono.Options/Mono.Options/Options.cs.  Anyways, long story short, our testing infrastructure for the UnityLinker needs to translate these arguments to the equivalent for our linker.  I think translation will be easier & less brittle if we minimize the amount of string parsing needed. 

* By separating out the flag into it's on parameter, it will make it easier for us to translate the arguments.  The idea would be that there is 1 [SetupLinker] attribute per flag, that way our UnityLinker test framework code doesn't have to scan an array of strings for flags, figure out how many arguments can/should follow in order to translate the arguments.

linker/Tests/Mono.Linker.Tests.Cases/TypeForwarding/TypeForwarderOnlyAssembliesKept.cs
```diff
@@ -13,15 +13,15 @@ namespace Mono.Linker.Tests.Cases.TypeForwarding {
...
 	[KeptAssembly ("Forwarder.dll")]
 	[KeptMemberInAssembly ("Library.dll", typeof (LibraryUsingForwarder), "GetValueFromOtherAssembly()")]
 	[KeptMemberInAssembly ("Implementation.dll", typeof (ImplementationLibrary), "GetSomeValue()")]
 	public class TypeForwarderOnlyAssembliesKept {
-		static void Main ()
+		public static void Main ()
 		{
-			Console.WriteLine (new ImplementationLibrary ().GetSomeValue ());
+			if (new ImplementationLibrary ().GetSomeValue () != "Hello")
+				throw new ApplicationException ("1");
```
mrvoorhe: What's the purpose of this change to throw an ApplicationException?

linker/Tests/TestCasesRunner/LinkerArgumentBuilder.cs
```diff
@@ -29,6 +29,12 @@ public virtual void AddCoreLink (string value)
 			Append (value);
 		}
 
+		public void LinkFromAssembly (string fileName)
```
mrvoorhe: Can you make virtual please

linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/RemovedForwarder.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+namespace Mono.Linker.Tests.Cases.Expectations.Assertions
+{
+	[AttributeUsage (AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
+	public class RemovedForwarder : BaseInAssemblyAttribute
```
mrvoorhe: Shouldn't the class name be RemovedForwarderAttribute?

linker/Tests/Mono.Linker.Tests.Cases.Expectations/Metadata/SetupLinker.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Metadata
+{
+	[AttributeUsage (AttributeTargets.Class, AllowMultiple = false)]
+	public class SetupLinker : BaseMetadataAttribute
```
mrvoorhe: Shouldn't the class name be SetupLinkerAttribute?

linker/Tests/Mono.Linker.Tests.Cases/TypeForwarding/TypeForwarderOnlyAssembliesKept.cs
marek-safar: Leftover from different change

linker/Tests/TestCasesRunner/LinkerArgumentBuilder.cs
marek-safar: done

linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/RemovedForwarder.cs
marek-safar: Indeed

linker/Tests/Mono.Linker.Tests.Cases/TypeForwarding/MissingTargetReference.cs
marek-safar: I don't like X number of custom attribute to mirror all input arguments. I want the setup to be as simple and readable as possible. Already with few of existing ones it's pain to track down to see what they. Why do you need to translate the arguments, why not just call existing code you have and pass the string[] as it's

mrvoorhe: Let me give you an example following the pattern you have here

[SetupLinkerOptions("-b true -t false -g false")]

So first, my problem with that.  
* Unless you grok the linker args, that is beyond meaning less. 
* Given that the linker args seem to be just arbitrary letters of the alphabet paired with some functionality, it's very hard to remember what they all do.

On the other hand, if I looked at a test case and saw something like [LinkSymbols].  That quickly communicates that the test case is going to link the symbol files. 

Now back to argument translation, given the above set of arguments, I need to pass the following to the UnityLinker
"--link-symbols --disable-keep-facades --disable-regen-guid"

Giving the example used in this PR "-p link TypeForwarderMissingReference.dll", I'd need to translate that to "-p=link,TypeForwarderMissingReference.dll"

None of this is impossible, but it's going to be tedious and error prone for us to parse whole strings and convert them.

The current test case attribute scheme came to be for a reason.  It let's us avoid arg parsing entirely in our test code and helps cover up what is in my opinion a set of command line arguments that do nothing to communicate their meaning.  This approach admittedly has it's short comings, one being it's very tedious to extend to support for new options. 

We need to find a new approach that meets in the middle.  I think an attribute accepting a raw string of args is too far in the open direction.  It will make our life harder, and aside from the 2-3 people who have memorized the linker args (I'm not one of them) it's not more readable.

PR #157, opened 2017-09-13
Improve logging

radekdoulik:
Added the ILogger interface for logging messages from the
linker. There is a new default implementation, the DefaultLogger
class, which just continues to use simple S.C.WriteLine as before.

The LinkerContext newly contains Logger property, which can be set to
instance of class implementing ILogger interface. As an example,
Xamarin.Android linker will be able to provide own logger, forwarding
the messages to Microsoft.Build.Utilities.TaskLoggingHelper. This way
it will be possible to see linker messages, depending on the actual
msbuild's verbosity level, in the XA build output.


marek-safar:
build

marek-safar:
build

PR #161, opened 2017-09-27
Add native trimming logic

sbomer:
This adds logic to scan managed dependencies for module references to
native libraries. The logic is taken from
Microsoft.Packaging.Tools.Trimming (https://github.com/dotnet/standard/blob/9a99f024343e6a36d15e7073890c07daa81606fc/Microsoft.Packaging.Tools.Trimming/docs/trimming.md).

As part of this change, we also stop rewriting the publish dependency
file, instead specifying the excluded managed and native assets as an
input to the target in the SDK that generates the deps file in the
first place.

@ericstj, @erozenfeld, please review.
/cc @swaroop-sridhar, @russellhadley


corebuild/integration/ILLink.Tasks/FindNativeDeps.cs
```diff
@@ -0,0 +1,109 @@
...
+				{
+					if (peReader.HasMetadata)
+					{
+						var reader = peReader.GetMetadataReader();
+						for (int i = 1, count = reader.GetTableRowCount(TableIndex.ModuleRef); i <= count; i++)
+						{
+							var moduleRef = reader.GetModuleReference(MetadataTokens.ModuleReferenceHandle(i));
+							var moduleName = reader.GetString(moduleRef.Name);
+
+							var moduleRefCandidates = new[] { moduleName, moduleName + ".dll", moduleName + ".so", moduleName + ".dylib" };
```
erozenfeld: Is moduleName ever specified without an extension in ModuleRef table?

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -109,7 +109,7 @@
   <UsingTask TaskName="CompareAssemblySizes" AssemblyFile="$(LinkTaskDllPath)" />
   <Target Name="_CompareLinkedAssemblySizes"
           AfterTargets="ComputeLinkedFilesToPublish"
-          DependsOnTargets="_ComputeManagedAssembliesToLink;_ComputeLinkedAssemblies"
+          DependsOnTargets="_ComputeManagedAssembliesToLink;_ComputeLinkedAssemblies;_FindNativeDeps"
```
erozenfeld: Isn't it enough to specify just _FindNativeDeps here? It depends on _ComputeLinkedAssemblies, which depends on ILLink, which depends on _ComputeManagedAssembliesToLink.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -287,38 +317,27 @@
                               RootDescriptorFilePath="$(_IntermediateRootDescriptorPath)" />
   </Target>
 
+  <!-- Uses _PublishConflictPackageFiles, an input to the publish deps
+       file generation target, to exclude the removed managed
+       assemblies and native dependencies from the generated deps
+       file. -->
+  <UsingTask TaskName="ComputeRemovedAssemblies" AssemblyFile="$(LinkTaskDllPath)" />
+  <Target Name="_ExcludeRemovedFilesFromDepFileGeneration"
+          DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish;_ComputeLinkedAssemblies;_FindNativeDeps;_HandlePublishFileConflicts"
```
erozenfeld: Some of the dependencies here are also redundant.

corebuild/integration/ILLink.Tasks/FindNativeDeps.cs
```diff
@@ -0,0 +1,109 @@
...
+				if (!allNative.ContainsKey(fileName))
+				{
+					allNative.Add(fileName, n);
+				}
+			}
+			var keptNative = new List<ITaskItem> ();
+			var managedAssemblies = ManagedAssemblyPaths.Select (i => i.ItemSpec).ToArray();
+			foreach (string managedAssembly in managedAssemblies)
+			{
+				using (var peReader = new PEReader(new FileStream(managedAssembly, FileMode.Open, FileAccess.Read, FileShare.Read)))
```
ericstj: Did you look into doing this with Cecil?  It's a bit unusual for a single task DLL to use two different metadata readers.  Not a blocking issue, but something to look into.

corebuild/integration/ILLink.Tasks/FindNativeDeps.cs
sbomer: Looking through ECMA-335, I found the following:

**ModuleRef** (page 238)
Name should match an entry in the Name column of the File table.  Moreover, that entry shall enable the CLI to locate the target module (typically it might name the file used to hold the module)  [ERROR]

**File** (page 228)
Name shall index a non-empty string in the String heap.  It shall be in the format <filename>.<extension>  (e.g., “foo.dll”, but not “c:\utils\foo.dll”)  [ERROR]

So I think we should expect moduleName to always have an extension. However, I tried this in musicstore and it turns out that most of the moduleRefs that are coming up in practice don't have extensions, so this logic is necessary unless we want to change that. @ericstj, is this expected? Do you know whether leaving out the extensions from moduleRefs was an intentional choice?

corebuild/integration/ILLink.Tasks/FindNativeDeps.cs
sbomer: I haven't looked into that.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
sbomer: That would have the same behavior, though I think it doesn't hurt to be explicit in case `_FindNativeDeps` changes. I've been trying to add dependencies on exactly the targets that produce the itemgroups that a given target uses. In this case, `_FindNativeDeps` was left over because I had been considering also printing a comparison of the removed native assemblies, so it should actually be removed here. Thanks!

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
sbomer: Here I'd prefer to be explicit. For each itemgroup that this task uses, I added a dependency on the target that produces it, so that the inputs will be available even if some of the dependencies between the dependent targets change.

corebuild/integration/ILLink.Tasks/FindNativeDeps.cs
```diff
@@ -0,0 +1,109 @@
...
+							{
+								// DLLImport that wasn't satisfied
+								Log.LogMessage(MessageImportance.High, "unsatisfied DLLImport: " + managedAssembly + " -> " + moduleName);
+							}
+						}
+					}
+				}
+			}
+
+			var nativeWhitelist = new HashSet<string> (NativeDepsToKeep.Select(i => i.ItemSpec));
```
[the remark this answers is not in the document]
sbomer: Fixed, thanks.

sbomer:
The failures appear unrelated to my change. I think they were run on a merge commit, so I don't think I'm missing any changes from master: https://jenkins.mono-project.com/job/test-linker-pull-request/143/consoleFull.  I saw that this recent PR also had a failed mono build, though with a different error (seems the disk was full): https://github.com/mono/linker/pull/164. @marek-safar, do you know what would be causing these failures?

marek-safar:
build

PR #166, opened 2017-10-12
Update to 2.0 tooling.

sbomer:
This changes updates the SDK we use to build illink to the released
2.0 version. Previously, we were using a prerelease version that had
some bugs, for which we can now eliminate the workarounds.

This upgrade has some consequences for musicstore. The prerelease SDK
was happy to build musicstore (which currently targets netcoreapp2.1),
but the 2.0 SDK correctly complains that it doesn't support targeting
this TFM. To fix this, it's necessary to build musicstore with a
prerelease version of the 2.1 SDK, as described by the instructions on
https://github.com/aspnet/JitBench/tree/86ac48203060598e60722a5f6a96112af22d069b.

This prerelease SDK enables razor view compilation by default, but the
precompiled views are added to the publish output after the linker
runs. Precompiling views also results in the reference assemblies
being omitted from the publish output. There is also some new dynamic
code generation in MusicStore that fails because the linker is missing
roots.

@erozenfeld, @swaroop-sridhar please review


corebuild/integration/test/IntegrationTestBase.cs
```diff
@@ -101,38 +111,37 @@ protected int RunCommand(string command, string args, string workingDir, string
...
 			string config = context.Configuration;
 			string demoRoot = Path.GetDirectoryName(csproj);
 
-			int ret = Dotnet($"restore -r {rid}", demoRoot);
-			if (ret != 0) {
-				output.WriteLine("restore failed");
-				Assert.True(false);
-				return;
-			}
-
```
erozenfeld: Why is restore no longer needed?

corebuild/integration/test/MusicStoreTest.cs
```diff
@@ -32,10 +34,73 @@ public void RunMusicStore()
...
 			AddLinkerReference(csproj);
 
-			BuildAndLink(csproj, rootFiles);
+			Dictionary<string, string> extraPublishArgs = new Dictionary<string, string>();
+			extraPublishArgs.Add("JITBENCH_FRAMEWORK_VERSION", netcoreappVersion);
+			BuildAndLink(csproj, rootFiles, extraPublishArgs);
+
+			// int ret = RunApp(csproj, out string commandOutput);
+			// Assert.True(ret == 0);
+		}
```
erozenfeld: Can the commented-out lines be removed?

corebuild/integration/test/IntegrationTestBase.cs
sbomer: With the new SDK, publish implicitly does a restore.

corebuild/integration/test/MusicStoreTest.cs
sbomer: Removed.

sbomer:
@erozenfeld I've addressed your feedback. Could you take another look?

corebuild/integration/test/IntegrationTestBase.cs
```diff
@@ -101,38 +111,39 @@ protected int RunCommand(string command, string args, string workingDir, string
 		///   linker task package.
 		///   Optionally takes a list of root descriptor files.
 		/// </summary>
-		public void BuildAndLink(string csproj, List<string> rootFiles = null)
+		public void BuildAndLink(string csproj, List<string> rootFiles = null, Dictionary<string, string> extraPublishArgs = null)
```
marek-safar: It'd be better to use List instead of Dictionary

PR #167, opened 2017-10-18
Allow link xml files to start with comments

joncham:
Find 'linker' child element rather than assuming it's the first element in the file. This allows link xml files to contain comments at the start.


marek-safar:
build

PR #172, opened 2017-10-24
Implement removal of link xml resources that were processed

mrvoorhe:


linker/Mono.Linker.Steps/BlacklistStep.cs
```diff
@@ -96,17 +96,17 @@ bool IsReferenced (string name)
 			return false;
 		}
 
-		static ResolveFromXmlStep GetExternalResolveStep (EmbeddedResource resource, AssemblyDefinition assembly)
+		protected virtual IStep GetExternalResolveStep (EmbeddedResource resource, AssemblyDefinition assembly)
```
mrvoorhe: I had to expose this due to some tech debt we have left over.  The UnityLinker has it's own "UnityResolveFromXmlStep " which is largely a copy of ResolveFromXmlStep with some changes I have not had a chance to unify.  Anyways, I had to introduce a "UnityBlacklistStep" so that we can use our "UnityResolveFromXmlStep"

linker/Tests/Mono.Linker.Tests.Cases/TestFramework/VerifyResourceInAssemblyAttributesBehavior.cs
```diff
@@ -0,0 +1,18 @@
...
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.TestFramework {
+	[SetupCompileResource ("Dependencies/VerifyResourceInAssemblyAttributesBehavior.txt")]
+	[KeptResource ("VerifyResourceInAssemblyAttributesBehavior.txt")]
+	// These are technically redundant, but the goal of this test is to verify the attributes are working which we can do
+	// by using them on the test case assembly even though you would normally use these attributes for other checking other
+	// supporting assemblies
+	[KeptResourceInAssembly ("test.exe", "VerifyResourceInAssemblyAttributesBehavior.txt")]
```
mrvoorhe: In our UnityLinker test suite we will have some tests that make more meaningful use of these 'KeptResourceInAssembly' and 'RemoveResourceInAssembly' attributes since our test suites have tighter control over what class libraries we use.

I thought it would be better to support them in the mono linker tests, as opposed to putting them in our unity linker tests, in case we were wanted to write a test that utilized them.

linker/Mono.Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -59,11 +59,15 @@ public class ResolveFromXmlStep : ResolveStep {
 
 		XPathDocument _document;
 		string _xmlDocumentLocation;
+		string _resourceName;
+		AssemblyDefinition _resourceAssembly;
 
-		public ResolveFromXmlStep (XPathDocument document, string xmlDocumentLocation = "<unspecified>")
+		public ResolveFromXmlStep (XPathDocument document, string xmlDocumentLocation = "<unspecified>", string resourceName = null, AssemblyDefinition resourceAssembly = null)
```
marek-safar: This should be different ctor as passing null to only one of the new values does not make sense

mrvoorhe: Done

mrvoorhe:
build

PR #174, opened 2017-11-03
Add a switch to specify action on user assemblies.

erozenfeld:
Add -u switch to specify action on user (non-core) assemblies similar
to -c switch for core assemblies.

### DIFF
--- a/linker/Mono.Linker/Driver.cs
+++ b/linker/Mono.Linker/Driver.cs
@@ -120,6 +120,9 @@ namespace Mono.Linker {
 					case 'c':
 						context.CoreAction = ParseAssemblyAction (GetParam ());
 						break;
+					case 'u':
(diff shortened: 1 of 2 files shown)


erozenfeld:
@marek-safar PTAL

linker/Mono.Linker/Driver.cs
```diff
@@ -294,6 +298,7 @@ static void Usage (string msg)
 			Console.WriteLine ("   --version   Print the version number of the {0}", _linker);
 			Console.WriteLine ("   -out        Specify the output directory, default to `output'");
 			Console.WriteLine ("   -c          Action on the core assemblies, skip, copy or link, default to skip");
+			Console.WriteLine("    -u          Action on the user assemblies, skip, copy or link, default to link");
```
erozenfeld: This line will need to be updated once #177 is merged.

PR #175, opened 2017-11-06
Prevent SDK roll-forward using global.json

sbomer:
Building illink with a machine-wide install of the SDK was causing
problems because the SDK resolver looks in global locations and rolls
forward to the newest available SDK. (The behavior is described in
https://github.com/dotnet/designs/issues/2.) We want to use a
repo-local SDK to build illink for reproducibility, so we use a
global.json to prevent roll-forward. Testing shows that this results
in the local SDK being preferred when a matching version is also
available machine-wide.

Roll-forward behavior was causing extra validation logic in the new
SDK to be enabled. This logic was tripping over a workaround we have
for nuget restore issues. We will need to investigate this interaction
when we next update the repo-local SDK.

@erozenfeld please review. Could you also confirm that this fixes the issue on your machine?


erozenfeld:
I verified that this fixes the issue on my machine. Thank you for the fix!

PR #176, opened 2017-11-07
Add an option to ignore unresolved types and methods.

erozenfeld:
With this change the linker won't throw if it encounters an unresolved
type or method during MarkStep and -n true was specified.

### DIFF
--- a/linker/Mono.Linker/Driver.cs
+++ b/linker/Mono.Linker/Driver.cs
@@ -95,6 +95,11 @@ namespace Mono.Linker {
 						if (token.Length < 3)
 							Usage ("Option is too short");
 
+						if (token == "--skip-unresolved") {
+							context.IgnoreUnresolved = bool.Parse (GetParam ());
+							continue;
+						}
+
 						switch (token [2]) {
 						case 'v':
 							Version ();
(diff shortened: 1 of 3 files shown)


erozenfeld:
@marek-safar PTAL

mrvoorhe:
We have a feature in the UnityLinker which "Stubs" unresolved types and methods with empty implementations.   That's why HandleUnresolvedType and HandleUnresolvedMethods exist.  I could bring this feature, and it's extensive test suite, upstream if that would be a more desired behavior than ignoring.

linker/Mono.Linker/Driver.cs
```diff
@@ -301,6 +304,7 @@ static void Usage (string msg)
 			Console.WriteLine ("   -b          Generate debug symbols for each linked module (true or false)");
 			Console.WriteLine ("   -g          Generate a new unique guid for each linked module (true or false)");
 			Console.WriteLine ("   -v          Keep memebers needed by debugger attributes (true or false)");
+			Console.WriteLine("    -n          Ignore unresolved types and methods (true or false)");
```
marek-safar: Looks good but for such special cases we should do advanced option e.g. `--skip-unresolved` only

erozenfeld: Sure, I can change that but wouldn't it look odd that this will be the only such option (not counting --version and --about)?

marek-safar: Don't worry the more will follow in the future

erozenfeld:
@marek-safar I addressed your comment. PTAL.

PR #180, opened 2017-11-21
Make interface marking overridable by derived classes.

rolfbjarne:
Make interface marking overridable by derived classes, so that derived classes
can decide whether an interface implementation should be marked or not.

Also remove any non-marked interface implementations.

### DIFF
--- a/linker/Mono.Linker.Steps/MarkStep.cs
+++ b/linker/Mono.Linker.Steps/MarkStep.cs
@@ -647,8 +647,7 @@ namespace Mono.Linker.Steps {
 
 			if (type.HasInterfaces) {
 				foreach (var iface in type.Interfaces) {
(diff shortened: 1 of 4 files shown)


marek-safar:
Please add a test case to https://github.com/mono/linker/tree/master/linker/Tests/Mono.Linker.Tests.Cases for the change

rolfbjarne:
@marek-safar it looks easy enough to add a test for the default case (no interfaces are removed) and I'll add that, but there doesn't seem to be obvious support for testing something that requires subclassing/overriding an existing step - am I missing something? Can you point me to an existing test that does this?

rolfbjarne:
I've added a new test now: 5ed14a738419ea1eab1da014d25056367c2b97b3

PR #182, opened 2017-11-27
Add two new assembly actions: AddBypassNGen and AddBypassNGenUsed.

erozenfeld:
AddBypassNGen specifies that instead of removing unreached methods
they should be marked with System.Runtime.BypassNGenAttribute. Types and fields
should be preserved.

AddBypassNGenUsed is the same as AddBypassNGen except unmarked assemblies should
be deleted.

ngen/crossgen will skip compilation of methods marked with
System.Runtime.BypassNGenAttribute so these actions can be used for conservative
linking: all msil is preserved but ngen/crossgen is run only on reachable methods.

System.Runtime.BypassNGenAttribute may not be present in the core library.
If that's the case, the linker will inject it.

### DIFF
--- a/linker/Mono.Linker.Steps/AddBypassNGenStep.cs
+++ b/linker/Mono.Linker.Steps/AddBypassNGenStep.cs
@@ -1,0 +1,120 @@
+﻿//
+// AddBypassNGenStep.cs
+//
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using Mono.Cecil;
+using Mono.Collections.Generic;
+using Mono.Cecil.Cil;
+
+namespace Mono.Linker.Steps {
+
+	public class AddBypassNGenStep : BaseStep{
+
+		AssemblyDefinition coreLibAssembly;
+		CustomAttribute bypassNGenAttribute;
+		
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			if (Annotations.GetAction (assembly) == AssemblyAction.AddBypassNGen) {
+				coreLibAssembly = Context.Resolve (assembly.MainModule.TypeSystem.CoreLibrary);
+				bypassNGenAttribute = null;
+				if (assembly == coreLibAssembly) {
+					EnsureBypassNGenAttribute (assembly.MainModule);
+				}
+
+				foreach (var type in assembly.MainModule.Types)
+					ProcessType (type);
+			}
+		}
+
+		void ProcessType (TypeDefinition type)
+		{
+			if (type.HasMethods)
+				ProcessMethods (type.Methods, type.Module);
+
+			if (type.HasNestedTypes)
+				ProcessNestedTypes (type);
+		}
+
+		void ProcessMethods (Collection<MethodDefinition> methods, ModuleDefinition module)
+		{
+			foreach (var method in methods)
+				if (!Annotations.IsMarked (method)) {
+					EnsureBypassNGenAttribute (module);
+					method.CustomAttributes.Add (bypassNGenAttribute);
+				}
+		}
+
+		void ProcessNestedTypes (TypeDefinition type)
+		{
+			for (int i = 0; i < type.NestedTypes.Count; i++) {
+				var nested = type.NestedTypes [i];
+				ProcessType (nested);
+			}
+		}
+
+		private void EnsureBypassNGenAttribute (ModuleDefinition targetModule)
+		{
+			if (bypassNGenAttribute != null) {
+				return;
+			}
+			ModuleDefinition corelibMainModule = coreLibAssembly.MainModule;
+			TypeReference bypassNGenAttributeRef = new TypeReference ("System.Runtime", "BypassNGenAttribute", corelibMainModule, targetModule.TypeSystem.CoreLibrary);
+			TypeDefinition bypassNGenAttributeDef = corelibMainModule.MetadataResolver.Resolve (bypassNGenAttributeRef);
+			MethodDefinition bypassNGenAttributeDefaultConstructor = null;
+
+			if (bypassNGenAttributeDef == null) {
+				// System.Runtime.BypassNGenAttribute is not found in corelib. Add it.
+				TypeReference systemAttributeRef = new TypeReference ("System", "Attribute", corelibMainModule, targetModule.TypeSystem.CoreLibrary);
+				TypeReference systemAttribute = corelibMainModule.MetadataResolver.Resolve (systemAttributeRef);
+				systemAttribute = corelibMainModule.ImportReference (systemAttribute);
+
+				if (systemAttribute == null)
+					throw new System.ApplicationException ("System.Attribute is not found in " + targetModule.TypeSystem.CoreLibrary.Name);
+
+				MethodReference systemAttributeDefaultConstructorRef = new MethodReference (".ctor", corelibMainModule.TypeSystem.Void, systemAttributeRef);
+				MethodReference systemAttributeDefaultConstructor = corelibMainModule.MetadataResolver.Resolve (systemAttributeDefaultConstructorRef);
+				systemAttributeDefaultConstructor = corelibMainModule.ImportReference (systemAttributeDefaultConstructor);
+
+				if (systemAttributeDefaultConstructor == null)
+					throw new System.ApplicationException ("System.Attribute has no default constructor");
+
+				bypassNGenAttributeDef = new TypeDefinition ("System.Runtime", "BypassNGenAttribute", TypeAttributes.NotPublic | TypeAttributes.Sealed, systemAttribute);
+
+				coreLibAssembly.MainModule.Types.Add (bypassNGenAttributeDef);
+
+				if (Annotations.GetAction (coreLibAssembly) == AssemblyAction.Copy) {
+					Annotations.SetAction (coreLibAssembly, AssemblyAction.Save);
+				}
+
+				const MethodAttributes ctorAttributes = MethodAttributes.Public | MethodAttributes.HideBySig | MethodAttributes.SpecialName | MethodAttributes.RTSpecialName;
+				bypassNGenAttributeDefaultConstructor = new MethodDefinition (".ctor", ctorAttributes, coreLibAssembly.MainModule.TypeSystem.Void);
+				var instructions = bypassNGenAttributeDefaultConstructor.Body.Instructions;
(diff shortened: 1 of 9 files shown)


erozenfeld:
@marek-safar @sbomer PTAL

linker/Mono.Linker.Steps/SweepStep.cs
```diff
@@ -336,13 +355,79 @@ void SweepDebugInfo (Collection<MethodDefinition> methods)
...
+
+				if (systemAttributeDefaultConstructor == null)
+					throw new System.ApplicationException ("System.Attribute has no default constructor");
+
+				bypassNGenAttribute = new TypeDefinition (
+										"System.Runtime",
+										"BypassNGenAttribute",
+										TypeAttributes.NotPublic | TypeAttributes.Sealed,
+										systemAttribute);
+
```
sbomer: Strange formatting - maybe put the parameters just one indentation level past the previous line?

linker/Mono.Linker.Steps/SweepStep.cs
```diff
@@ -336,13 +355,79 @@ void SweepDebugInfo (Collection<MethodDefinition> methods)
...
+				coreLibAssembly.MainModule.Types.Add (bypassNGenAttribute);
```
sbomer: What happens if we add the attribute to corelib, but specify a corelib action "Copy"? Will we end up with dangling references to this attribute?

erozenfeld:
BypassNGenAttribute is currently not defined in any assembly even though ngen/crossgen checks for it. We may (or may not) add it to SPC or mscorlib in the future. I'd like the linker to be able to work regardless of whether the attribute already present in the core library.

linker/Mono.Linker.Steps/SweepStep.cs
erozenfeld: I fixed formatting by just moving the args to the call line.

linker/Mono.Linker.Steps/SweepStep.cs
erozenfeld: Good catch. I made a change to modify corlib action from Copy to Save here.

erozenfeld:
@marek-safar I added a new AddBypassNGenStep that is inserted into the pipeline after SweepStep when necessary. PTAL.

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
```diff
@@ -0,0 +1,133 @@
+﻿//
+// AddBypassNGenStep.cs
+//
+// Author:
+//   Jb Evain (jbevain@gmail.com)
+//
+// (C) 2006 Jb Evain
+// (C) 2007 Novell, Inc.
+//
```
erozenfeld: Is this correct header for new files?

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
```diff
@@ -0,0 +1,133 @@
...
+namespace Mono.Linker.Steps {
+
+	public class AddBypassNGenStep : BaseStep{
+
+		CustomAttribute bypassNgenAttribute;
+		
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			if (Annotations.GetAction (assembly) == AssemblyAction.AddBypassNGen) {
+				bypassNgenAttribute = GetBypassNGenAttribute (assembly.MainModule);
```
marek-safar: This looks slightly wrong, you will add the attribute to CoreLibrary library even if it possibly never used (e.g. no methods in any type). It'd probably be better to create the CA on demand

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
marek-safar: You should just use a header with your name and Microsoft copyright

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
```diff
@@ -0,0 +1,133 @@
...
+						break;
+					}
+				}
+
+				if (bypassNGenAttributeDefaultConstructor == null) {
+					throw new System.ApplicationException ("System.Runtime.BypassNGenAttribute has no default constructor");
+				}
+			}
+
+			MethodReference defaultConstructorReference = targetModule.ImportReference (bypassNGenAttributeDefaultConstructor);
```
marek-safar: Did you test this with CoreLibrary uses as AddBypassNGen target ?

erozenfeld: Yes, I did. It works fine. In this case ImportReference simply returns the argument: https://github.com/mono/cecil/blob/master/Mono.Cecil/ModuleDefinition.cs#L876

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
erozenfeld: We don't normally put the author's name or Microsoft copyright on open source files under .NET foundation. For example: https://github.com/dotnet/coreclr/blob/master/src/jit/morph.cpp#L1 or
https://github.com/dotnet/corefx/blob/master/src/System.Collections/src/System/Collections/Generic/LinkedList.cs#L1 . Should we use a similar pattern for linker sources?

erozenfeld: @russellhadley Any opinion on this?

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
erozenfeld: I made a change to add the attribute on demand. I still do it eagerly if AddBypassNGen is specified for corelib because I don't want to modify corelib's type collection while iterating on it.

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
marek-safar: .NET foundation works too

erozenfeld: Updated the header.

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
```diff
@@ -0,0 +1,121 @@
...
+
+				bypassNGenAttributeDef = new TypeDefinition ("System.Runtime", "BypassNGenAttribute", TypeAttributes.NotPublic | TypeAttributes.Sealed, systemAttribute);
+
+				coreLibAssembly.MainModule.Types.Add (bypassNGenAttributeDef);
+
+				if (Annotations.GetAction (coreLibAssembly) == AssemblyAction.Copy) {
+					Annotations.SetAction (coreLibAssembly, AssemblyAction.Save);
+				}
+
+				var ctorAttributes = MethodAttributes.Public | MethodAttributes.HideBySig | MethodAttributes.SpecialName | MethodAttributes.RTSpecialName;
```
marek-safar: Should be const

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
```diff
@@ -0,0 +1,121 @@
...
+	public class AddBypassNGenStep : BaseStep{
+
+		AssemblyDefinition coreLibAssembly;
+		CustomAttribute bypassNGenAttribute;
+		
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			if (Annotations.GetAction (assembly) == AssemblyAction.AddBypassNGen) {
+				coreLibAssembly = Context.Resolve (assembly.MainModule.TypeSystem.CoreLibrary);
+				bypassNGenAttribute = null;
```
marek-safar: This is still quite confusing. Would be much better to split GetBypassNGenAttribute into 2 methods (e.g. DefineBypassNGenAttribute and GetBypassNGenAttribute) and call DefineBypassNGenAttribute one only here ?

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
```diff
@@ -0,0 +1,121 @@
...
+
+	public class AddBypassNGenStep : BaseStep{
+
+		AssemblyDefinition coreLibAssembly;
+		CustomAttribute bypassNGenAttribute;
+		
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			if (Annotations.GetAction (assembly) == AssemblyAction.AddBypassNGen) {
+				coreLibAssembly = Context.Resolve (assembly.MainModule.TypeSystem.CoreLibrary);
```
marek-safar: You could check for null no to repeat this all the time

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
erozenfeld: Added const.

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
erozenfeld: BypassNGenAttribute may already be defined in corelib so we may not have to define it and the name DefineBypassNGenAttribute would be confusing.

I changed the name of the method to EnsureByapssNGenAttribute and made its return type void. Do you like this version better?

linker/Mono.Linker.Steps/AddBypassNGenStep.cs
erozenfeld: No, I can't. The reason is that assembly1.MainModule.TypeSystem.CoreLibrary and assembly2.MainModule.TypeSystem.CoreLibrary may not be the same. For example, assembly1 may reference System.Runtime.dll while assembly2 may reference System.Private.CoreLib.dll. This may happen in the context of the same app. We actually have such scenarios with ASP.NET apps.

In general, currently there may be up to 4 corelib dlls: https://github.com/mono/cecil/blob/master/Mono.Cecil/TypeSystem.cs#L282

I considered 3 solutions for this:

1. Figure out the assembly that defines System.Object (i.e., System.Private.CoreLib.dll), add BypassNGenAttribute there, and add missing assembly references to it (e.g., to assembly1 in the example above). This may cause problems for assemblies that will reference multiple core assemblies.

2. Figure out the assembly that defines System.Object, add BypassNGenAttribute there, and add type forwarders to it from other core assemblies (e.g., from System.Runtime.dll in the example above).

3. Add BypassNGenAttribute to each of the corelib assemblies that needs it. 

The code currently implements option 3. That's simpler than option 2 and should be fine since we have at most 4 core assemblies.

marek-safar: Yeah, not sure it's all worth the trouble and injecting private attribute to assemblies which need it wouldn't be enough

PR #186, opened 2017-12-02
Add assembly name to the private types when dumping dependencies

radekdoulik:
So that it can be used in the linker analyzer as a key to the type
sizes dictionary in the space analyzer.

The `IsAssemblyBound` method checks whether a type is visible outside
of an assembly. We need it for classes like
`<PrivateImplementationDetails>/__StaticArrayInitTypeSize=120`.

In this case the class hierarchy looks like this:

```
.class private auto ansi sealed '<PrivateImplementationDetails>'
       extends System.Object
{
 .class explicit ansi sealed nested private '__StaticArrayInitTypeSize=120'
        extends System.ValueType
 {
   ...
 }
}
```

### DIFF
--- a/linker/Linker/Annotations.cs
+++ b/linker/Linker/Annotations.cs
@@ -74,7 +74,7 @@ namespace Mono.Linker {
 			writer.WriteStartDocument ();
 			writer.WriteStartElement ("dependencies");
 			writer.WriteStartAttribute ("version");
-			writer.WriteString ("1.0");
+			writer.WriteString ("1.1");
 			writer.WriteEndAttribute ();
 		}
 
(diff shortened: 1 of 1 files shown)


linker/Mono.Linker/Annotations.cs
```diff
@@ -349,11 +349,40 @@ public void Pop ()
...
+
+			return true;
+		}
+
 		string TokenString (object o)
 		{
 			if (o == null)
 				return "N:null";
 
+			if (o is TypeReference) {
```
marek-safar: `o is TypeReference t` will give you `t` immediately

linker/Mono.Linker/Annotations.cs
```diff
@@ -349,11 +349,40 @@ public void Pop ()
 			dependency_stack.Pop ();
 		}
 
+		bool IsNestedInPrivate (TypeDefinition td)
```
marek-safar: I am not sure what is the intention here, what is nested in private?
You should also move td.IsNestedPrivate check here to make this work with 2+ levels of nesting

linker/Mono.Linker/Annotations.cs
radekdoulik: I knew there's faster way, forgot to check it before pushing. Thanks!

linker/Mono.Linker/Annotations.cs
radekdoulik: I have added explanation to the commit message and updated the patch.

PR #190, opened 2017-12-08
Add a task for computing System.Private.CoreLib roots needed by the coreclr runtime.

erozenfeld:
The task will run during coreclr build and compute the roots from coreclr header
files. The roots xml will be embedded as resource into System.Private.CoreLib.dll.

Example of xml computed by this task: https://gist.github.com/erozenfeld/b005d315c2f8e7da791bd40a125c07d9

### DIFF
--- a/corebuild/integration/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
+++ b/corebuild/integration/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
@@ -1,0 +1,326 @@
+using System;
+using System.IO;
+using System.Xml;
+using System.Collections.Generic;
+using Microsoft.Build.Utilities; // Task
+using Microsoft.Build.Framework; // ITaskItem
+
+namespace ILLink.Tasks
+{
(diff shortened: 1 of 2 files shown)


erozenfeld:
@sbomer PTAL

corebuild/integration/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -0,0 +1,326 @@
...
+		public ITaskItem NamespaceFilePath { get; set; }
+
+		/// <summary>
+		///   The path to mscorlib.h.
+		/// </summary>
+		[Required]
+		public ITaskItem MscorlibFilePath { get; set; }
+
+		/// <summary>
+		///   The path to cortype.h.
```
sbomer: Should this be cortypeinfo.h?

erozenfeld: Yes. Updated the comment.

PR #191, opened 2017-12-08
Update MusicStore roots.

erozenfeld:
Add several roots from System.Private.Corelib.
Remove "required" = true since it's the default.


erozenfeld:
@sbomer PTAL

PR #192, opened 2017-12-08
Add crossgen targets

sbomer:
This change adds an optional (on by default) phase to the publish pipeline that runs crossgen on a specified set of assemblies. The default is to crossgen the platform assemblies, but the default can be overridden from other targets. Files that are already R2R will not be crossgen'd again, and support is included for incremental build. The new tasks and targets will be part of ILLink.Tasks. They could potentially be separated out into a new task if the crossgen functionality is desirable on its own.

The logic uses SDK targets from [Microsoft.NET.CrossGen.targets](https://github.com/dotnet/sdk/blob/cb20a5fb09b3ea65193a098ba5e770cc8b34ec1b/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.NET.CrossGen.targets) to restore crossgen. The SDK targets have some interdependency with the ComposeStore targets, so this change includes a few workarounds that enable us to reuse the SDK targets for restore. The SDK also has a target that runs crossgen, but I found this target too difficult to adapt for our use case.

Caveats: this only supports cases in which the target platform is the same as the host platform.

@erozenfeld, @swaroop-sridhar, please review. /cc @russellhadley


sbomer:
@ericstj, would you mind taking a look as well?

corebuild/integration/ILLink.Tasks/ComputeReadyToRunAssemblies.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Xml;
+using System.Xml.Linq;
+using System.Linq;
+using System.Reflection;
+using System.Collections.Generic;
+using Microsoft.Build.Utilities;
+using Microsoft.Build.Framework;
+using Microsoft.NET.Build.Tasks;
+using Mono.Cecil;
+
```
erozenfeld: Looks like some of the `using` declaration are unnecessary (e.g., `using System.Xml`). Please remove them.

corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
```diff
@@ -0,0 +1,231 @@
...
+    </FilterByMetadata>
+
+    <ItemGroup>
+      <FilesToCrossGen Include="@(IntermediateAssembly)" />
+      <FilesToCrossGen Include="@(_PlatformLibrariesForCrossGen)" />
+    </ItemGroup>
+
+    <Error Text="System.Private.CoreLib should already be crossgen compiled."
+           Condition=" '%(FilesToCrossGen.Filename)' == 'System.Private.CoreLib' " />
+  </Target>
```
erozenfeld: Can you change the logic so that System.Private.CoreLib is not required to be already crossgen compiled? We will soon allow linking of System.Private.CoreLib and post-link System.Private.CoreLib will have to be cross-gened.

corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
```diff
@@ -0,0 +1,231 @@
...
+  <Target Name="ComputeCrossGenFilesToPublish"
+          BeforeTargets="$(ComputeCrossGenFilesToPublishBefore)"
+          AfterTargets="$(ComputeCrossGenFilesToPublishAfter)"
+          Condition=" '$(CrossGenDuringPublish)' == 'true' "
+          DependsOnTargets="_CrossGenForPublish">
+
+    <!-- Rewrite ResolvedAssembliesToPublish, preserving metadata. -->
+    <ItemGroup>
+      <__CrossGenResolvedAssembliesToPublish Include="@(ResolvedAssembliesToPublish->'$(IntermediateOptimizedDir)/%(Filename)%(Extension)')" />
+      <_CrossGenResolvedAssembliesToPublish Include="@(__CrossGenResolvedAssembliesToPublish)" Condition="Exists('%(Identity)')" />
```
erozenfeld: Can you make the two items names above differ in something other than the number of leading underscores? Do you need the two items?

corebuild/integration/ILLink.Tasks/Utils.cs
```diff
@@ -12,4 +12,15 @@ public static bool IsManagedAssembly (string fileName)
...
 	}
-}
\ No newline at end of file
+
+	public static bool IsReadyToRunAssembly (string fileName)
+	{
+		try {
+			ModuleDefinition module = ModuleDefinition.ReadModule (fileName);
+			return (module.Attributes & ModuleAttributes.ILOnly) == 0 &&
+				(module.Attributes & (ModuleAttributes) 0x04) != 0;
```
erozenfeld: Can you use a descriptive name for this constant? Should we update ModuleDefinition enum in Cecil? Also, this logic will need to be generalized to handle re-crossgen-ing of System.Private.CoreLib, which is not ReadyToRun compiled.

corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
```diff
@@ -0,0 +1,231 @@
...
+      <!-- FilesToCrossGen doesn't contain System.Private.CoreLib, so
+           it will still be published. -->
+      <ResolvedAssembliesToPublish Remove="@(FilesToCrossGen)" />
+      <ResolvedAssembliesToPublish Include="@(_CrossGenResolvedAssembliesToPublish)" />
+    </ItemGroup>
+
+    <!-- Rewrite IntermediateAssembly, preserving metadata. -->
+    <ItemGroup>
+      <__CrossGenIntermediateAssembly Include="@(IntermediateAssembly->'$(IntermediateOptimizedDir)/%(Filename)%(Extension)')" />
+      <_CrossGenIntermediateAssembly Include="@(__CrossGenIntermediateAssembly)" Condition="Exists('%(Identity)')" />
```
erozenfeld: Can you make the two items names above differ in something other than the number of leading underscores? Do you need the two items?

corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
```diff
@@ -0,0 +1,231 @@
...
+    <MSBuild Projects="@(CrossGenInvocations)"
+             Targets="RunCrossGenForPublish" />
+
+  </Target>
+
+  <!-- Workarounds for SDK issues around the interdependency between
+       ComposeStore targets and CrossGen targets. These let us reuse
+       PrepOptimizer from the SDK to obtain a crossgen executable for
+       the target RID. -->
+
```
erozenfeld: Are these SDK issues expected to be fixed or will we get stuck with these workarounds?

corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
```diff
@@ -0,0 +1,231 @@
...
+    </ItemGroup>
+  </Target>
+
+  <UsingTask TaskName="FilterByMetadata" AssemblyFile="$(LinkTaskDllPath)" />
+  <!-- This computes the default set of files that we want to be
+       crossgen'd. Some of these may already be R2R images, and these
+       will not be crossgen'd again. The default is to crossgen the
+       app and platform libraries. Defaults will be used only if
+       FilesToCrossGen hasn't been set elsewhere, allowing users and
+       other props/targets to select what will be crossgen'd. -->
```
erozenfeld: We need to ensure that when the linker part of the publish step, we crossgen  those and only those assemblies that were  crossgen-ed before linking. I don't think that default ensures that.

corebuild/integration/ILLink.Tasks/Utils.cs
sbomer: We discussed making cecil aware of these flags in https://github.com/jbevain/cecil/issues/390, but Jb didn't think cecil was the right place. Let's make the generalization for SPC along with the other linker changes for linking SPC.

sbomer: On second thought, maybe Jb would be open to adding that enum value as long as we have the logic to handle it. I'll make a PR in cecil. I'm also going to factor this so that the logic lives in one place (right now it's duplicated in https://github.com/mono/linker/blob/master/linker/Mono.Linker.Steps/OutputStep.cs#L90).

sbomer: Cecil pr: https://github.com/jbevain/cecil/pull/475

erozenfeld:
Looks good pending ModuleAttribute cecil change and corresponding changes here.

erozenfeld:
build

corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
erozenfeld: These PRs enable linking of SPC: https://github.com/mono/linker/pull/212, https://github.com/mono/linker/pull/211, https://github.com/dotnet/coreclr/pull/15525.

marek-safar:
Could you rebase the PR

sbomer:
Thanks! Rebased.

PR #200, opened 2017-12-13
Don't catch and wrap if already an XmlResolutionException

mrvoorhe:
We have some error cases we would like to bubble up as the culprit, this behavior of catching all exceptions and throwing a new exception was interfering with that.

I changed our code to throw XmlResolutionExceptions in cases we don't want caught


marek-safar:
Please rebase

mrvoorhe:
Done

PR #202, opened 2017-12-15
Add mark attribute, avoid some deps duplicating

radekdoulik:
Add `mark` attribute to the dependencies edge element, for these edges
added when marking.

Introduce `MarkAndPush` method. It simplifies common case, where we
called `Annotations::Mark` and `Annotations::Push`. It also avoids
adding duplicate edge in this case.

### DIFF
--- a/linker/Linker/Annotations.cs
+++ b/linker/Linker/Annotations.cs
@@ -74,7 +74,7 @@ namespace Mono.Linker {
 			writer.WriteStartDocument ();
 			writer.WriteStartElement ("dependencies");
 			writer.WriteStartAttribute ("version");
-			writer.WriteString ("1.1");
+			writer.WriteString ("1.2");
 			writer.WriteEndAttribute ();
 		}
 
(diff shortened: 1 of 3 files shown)


linker/Linker/Annotations.cs
```diff
@@ -316,7 +322,7 @@ public void CloseSymbolReader (AssemblyDefinition assembly)
 			return slots;
 		}
 
-		public void AddDependency (object o)
+		public void AddDependency (object o, bool mark=false)
```
marek-safar: What does mark mean in this case?

marek-safar:
build

linker/Linker/Annotations.cs
radekdoulik: The mark value indicates whether the dependency was added when o was marked by the linker.

marek-safar: could you rename the parameter to be `marked` then ?

radekdoulik: ok, updated

mrvoorhe:
@radekdoulik Could you elaborate on why it's useful to have mark=1 on an <edge>?

radekdoulik:
@mrvoorhe it is added for the analyzer tool to be able to distinguish between mark induced dependencies and the other dependencies. I plan to add information about the stack as well, so that I can reconstruct the stack state in the analyzer, to be able to have dependency stack information around the mark time.

I also hope it might help you in your graph visualization, where you might just use `mark` edges with `*Def` and `*Ref` vertices and ignore these from push operations, which would simplify the graph.

mrvoorhe:
I don't know if this is going to produce meaningful information.  For example, in the MarkStep, there are places where we do `if(Annotations.Processed(obj)) return` and if that statement isn't true, we proceed to code that calls Annotations.Mark.  If you shift something around and things get processed in a slightly different order, _I think_ the mark=1 attribute could end up on a different edge.

radekdoulik:
That should be OK as the information still means, that this edge was added by Mark operation.

radekdoulik:
rebased

PR #204, opened 2017-12-18
Break Dependency stuff out into it's own class

mrvoorhe:
I feel like the dependency stuff is going to see some more changes coming soon.  In preparation for that, I think it would be good to pull it out into it's own class

### DIFF
--- a/linker/Linker.Steps/BaseStep.cs
+++ b/linker/Linker.Steps/BaseStep.cs
@@ -44,6 +44,8 @@ namespace Mono.Linker.Steps {
 
 		public MarkingHelpers MarkingHelpers => _context.MarkingHelpers;
 
+		public Dependencies Dependencies => _context.Dependencies;
+
 		public void Process (LinkContext context)
 		{
 			_context = context;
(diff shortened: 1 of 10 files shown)


mrvoorhe:
@radekdoulik Is there a reason why the `--dump-dependencies` option uses reflection to invoke PrepareDumpDependencies on AnnotationsStore instead of just calling it directly?

mrvoorhe:
Don't merge this PR yet.

I need to update the following code block from Driver.cs

```
						if (token == "--dump-dependencies")
						{
							var prepareDependenciesDump = context.Annotations.GetType ().GetMethod ("PrepareDependenciesDump", new Type [] { });
							if (prepareDependenciesDump != null)
								prepareDependenciesDump.Invoke (context.Annotations, null);
							continue;
						}```

I'm waiting to hear back from @radekdoulik regarding why reflection was used.  I'd like to get rid of the reflection if I can so that this option doesn't get accidentally broken (like I almost did here)

marek-safar:
@mrvoorhe Radek needs the reflection stuff and he suggested to do the refactoring himself. Closing the PR due to that.

PR #205, opened 2017-12-18
Add attribute to allow arbitrary linker option definition in tests

mrvoorhe:
Plus add a test for --skip-unresolved which utilizes this new attribute

### DIFF
--- a/linker/Tests/TestCases/IndividualTests.cs
+++ b/linker/Tests/TestCases/IndividualTests.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using Mono.Linker.Tests.Cases.References.Individual;
+using Mono.Linker.Tests.TestCases;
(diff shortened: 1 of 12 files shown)


linker/Tests/TestCases/IndividualTests.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using Mono.Linker.Tests.Cases.References.Individual;
+using Mono.Linker.Tests.TestCases;
+using Mono.Linker.Tests.TestCasesRunner;
+using NUnit.Framework;
+
+namespace Mono.Linker.Tests.TestCases
+{
+	[TestFixture]
+	public class IndividualTests
```
mrvoorhe: I don't necessarily love this organization scheme.  But I think it works until we accumulate more tests that need to work like this.

PR #207, opened 2017-12-19
Extracted dependency dumping into the new Tracer class

radekdoulik:
Kept the Annotations::PrepareDependenciesDump methods, as they are
still needed for the mobile linkers. They will be removed once we
update them.

### DIFF
--- a/linker/Linker/LinkContext.cs
+++ b/linker/Linker/LinkContext.cs
@@ -34,6 +34,10 @@ using Mono.Cecil.Cil;
 
 namespace Mono.Linker {
 
+	public interface IAnnotationStoreFactory {
+		AnnotationStore Create (Tracer tracer);
+	}
+
 	public class LinkContext : IDisposable {
 
 		Pipeline _pipeline;
@@ -129,6 +133,8 @@ namespace Mono.Linker {
 
 		public MarkingHelpers MarkingHelpers { get; private set; }
 
+		public Tracer Tracer { get; private set; } = new Tracer ();
+
 		public LinkContext (Pipeline pipeline)
 			: this (pipeline, new AssemblyResolver ())
 		{
@@ -138,20 +144,19 @@ namespace Mono.Linker {
 			: this(pipeline, resolver, new ReaderParameters
 			{
 				AssemblyResolver = resolver
-			},
-			new AnnotationStore ())
+			})
 		{
 		}
 
-		public LinkContext (Pipeline pipeline, AssemblyResolver resolver, ReaderParameters readerParameters, AnnotationStore annotations)
+		public LinkContext (Pipeline pipeline, AssemblyResolver resolver, ReaderParameters readerParameters, IAnnotationStoreFactory storeFactory = null)
 		{
 			_pipeline = pipeline;
 			_resolver = resolver;
 			_actions = new Dictionary<string, AssemblyAction> ();
 			_parameters = new Dictionary<string, string> ();
-			_annotations = annotations;
 			_readerParameters = readerParameters;
 			MarkingHelpers = CreateMarkingHelpers ();
+			_annotations = storeFactory != null ? storeFactory.Create (Tracer) : new AnnotationStore (Tracer);
 		}
 
 		public TypeDefinition GetType (string fullName)
--- a/linker/Linker.Steps/BaseStep.cs
+++ b/linker/Linker.Steps/BaseStep.cs
@@ -42,6 +42,10 @@ namespace Mono.Linker.Steps {
 			get { return _context.Annotations; }
 		}
 
+		public Tracer Tracer {
+			get { return _context.Tracer; }
+		}
+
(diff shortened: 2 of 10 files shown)


linker/Linker/LinkContext.cs
```diff
@@ -138,18 +140,17 @@ public LinkContext (Pipeline pipeline, AssemblyResolver resolver)
...
 
-		public LinkContext (Pipeline pipeline, AssemblyResolver resolver, ReaderParameters readerParameters, AnnotationStore annotations)
+		public LinkContext (Pipeline pipeline, AssemblyResolver resolver, ReaderParameters readerParameters)
 		{
 			_pipeline = pipeline;
 			_resolver = resolver;
 			_actions = new Dictionary<string, AssemblyAction> ();
 			_parameters = new Dictionary<string, string> ();
-			_annotations = annotations;
+			_annotations = new AnnotationStore (this);
```
mrvoorhe: Can you use the Create pattern I used in https://github.com/mono/linker/pull/208 ?

Our linker has our own AnnotationsStore that we need to create

marek-safar: That's very bad practice, could you pass the `annotations` value as an argument instead ?

mrvoorhe: I understand it's not great.  Why is it `very bad`?

We can (and do today) pass in AnnotationsStore as an argument to LinkContext constructor.  However, once AnnotationsStore needs LinkContext passed to it at instantiation you can no longer do that.  An alternative would be to expose a pubic property to set `annotations.Context = this` inside LinkContext .ctor().  However, I'd argue a public setter is worse

We could do a more invasive refactor to do something nicer, but is it worth the effort?

radekdoulik: I wasn't sure how you use the virtual methods, so I rather left it out from this PR. I can add it in another one.

mrvoorhe: As long as a solution follows relatively soon after.  This is an API break for us.  I can't keep our linker in sync with upstream until I have some way to create our UnityAnnotationStore instead of the default AnnotationStore.  My virtual method recommendation is my "good enough suggestion".  If you want to go with a different approach, feel free.

radekdoulik:
This might work:

```
diff --git a/linker/Linker/LinkContext.cs b/linker/Linker/LinkContext.cs
index 46267e3..244c7a1 100644
--- a/linker/Linker/LinkContext.cs
+++ b/linker/Linker/LinkContext.cs
@@ -34,6 +34,10 @@ using Mono.Cecil.Cil;
 
 namespace Mono.Linker {
 
+       public interface IAnnotationStoreFactory {
+               AnnotationStore CreateAnnotationStore (LinkContext context);
+       }
+
        public class LinkContext : IDisposable {
 
                Pipeline _pipeline;
@@ -144,15 +148,15 @@ namespace Mono.Linker {
                {
                }
 
-               public LinkContext (Pipeline pipeline, AssemblyResolver resolver, ReaderParameters readerParameters)
+               public LinkContext (Pipeline pipeline, AssemblyResolver resolver, ReaderParameters readerParameters, IAnnotationStoreFactory storeFactory = null)
                {
                        _pipeline = pipeline;
                        _resolver = resolver;
                        _actions = new Dictionary<string, AssemblyAction> ();
                        _parameters = new Dictionary<string, string> ();
-                       _annotations = new AnnotationStore (this);
                        _readerParameters = readerParameters;
                        MarkingHelpers = CreateMarkingHelpers ();
+                       _annotations = storeFactory != null ? storeFactory.CreateAnnotationStore (this) : new AnnotationStore (this);
                }
 
                public TypeDefinition GetType (string fullName)
```

radekdoulik:
Or even better pass just Tracer to the AnnotationStore constructor. Will update the PR.

radekdoulik:
@mrvoorhe would the updated patch work in your case?

mrvoorhe:
@radekdoulik Yes.  One request.  Can you make IAnnotationStoreFactory more general purpose so that 
 CreateMarkingHelpers can move there as well?

Might be a little more future proof if it was an abstract class instead of an interface and the abstract class had the default implementation of `new AnnotationStore(this)` instead of putting a `storeFactory == null` check here.  That way if a new Create* method is added in the future then there will already be a virtual method with the default implementation.

mrvoorhe:
`Or even better pass just Tracer to the AnnotationStore constructor. Will update the PR.`

I think it would be more flexible and consistent to store LinkContext.

Why have Tracer and MarkingHelpers that need LinkContext while AnnotationStore just has Tracer?

radekdoulik:
`Why have Tracer and MarkingHelpers that need LinkContext while AnnotationStore just has Tracer?`

That would mean passing reference to potentially incomplete context instance around, so it will be similar to having virtual create methods. Thus passing just the Tracer instead. To the MarkingHelpers we might just pass AnnotationStore instead of context. (until you derive from it and need the context there)

PR #208, opened 2017-12-19
Only record dependency info for nodes in assemblies that will be linked

mrvoorhe:
There's a lot of output in the linker-dependencies.xml.  I find it hard to reason about what is going on with so much noise.

I was thinking of ways to reduce the noise, mainly to start writing some tests for the xml output and I thought, if the assembly won't be modified, then do we really care about reporting these edges?

This PR makes it so that edges are only written to the xml when the assembly of one, or both, of the items will be modified.  This greatly reduces the output size of linker-dependencies.xml during tests and it makes it much easier to read the file and understand what is going on.

Sure there are some edge cases that would be odd, for example, code that calls into an assembly that is not linked and then a callback calls back into an assembly that is linked.  You would loose some useful edges there, but I think for all real world cases you will link all assemblies, so it would mainly only be tests that run into that scenario.

Having said that, I can see the argument for putting this behavior behind an opt-in command line flag.  If you would like me to add that, let me know.

### DIFF
--- a/linker/Linker/Driver.cs
+++ b/linker/Linker/Driver.cs
@@ -113,6 +113,11 @@ namespace Mono.Linker {
 							continue;
 						}
 
+						if (token == "--reduced-tracing") {
+							context.EnableReducedTracing = bool.Parse (GetParam ());
+							continue;
+						}
+
 						switch (token [2]) {
 						case 'v':
 							Version ();
@@ -328,6 +333,7 @@ namespace Mono.Linker {
 			Console.WriteLine ("   --skip-unresolved   Ignore unresolved types and methods (true or false)");
 			Console.WriteLine ("   --dependencies-file Specify the dependencies file path, if unset the default path is used: <output directory>/linker-dependencies.xml.gz");
 			Console.WriteLine ("   --dump-dependencies Dump dependencies for the linker analyzer tool");
+			Console.WriteLine ("   --reduced-tracing   Reduces dependency output related to assemblies that will not be modified");
 			Console.WriteLine ("   -out                Specify the output directory, default to `output'");
 			Console.WriteLine ("   -c                  Action on the core assemblies, skip, copy, copyused, addbypassngen, addbypassngenused or link, default to skip");
 			Console.WriteLine ("   -u                  Action on the user assemblies, skip, copy, copyused, addbypassngen, addbypassngenused or link, default to link");
--- a/linker/Linker/LinkContext.cs
+++ b/linker/Linker/LinkContext.cs
@@ -107,6 +107,8 @@ namespace Mono.Linker {
 			set { _ignoreUnresolved = value; }
 		}
 
+		public bool EnableReducedTracing { get; set; }
+
 		public System.Collections.IDictionary Actions {
 			get { return _actions; }
 		}
--- a/linker/Linker/Tracer.cs
+++ b/linker/Linker/Tracer.cs
@@ -29,20 +29,22 @@
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
-
+using System.Xml;
 using Mono.Cecil;
+using Mono.Linker.Steps;
 
 namespace Mono.Linker
 {
 	public class Tracer {
+		public const string DefaultDependenciesFileName = "linker-dependencies.xml.gz";
 
-		public string DependenciesFileName { get; set; } = "linker-dependencies.xml.gz";
+		public string DependenciesFileName { get; set; } = DefaultDependenciesFileName;
 
 		protected readonly LinkContext context;
 
 		Stack<object> dependency_stack;
 		System.Xml.XmlWriter writer;
-		GZipStream zipStream;
+		Stream stream;
 
 		public Tracer (LinkContext context) => this.context = context;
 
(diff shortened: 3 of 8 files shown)


radekdoulik:
I would indeed want that as optional, as the AssemblyAction can be (and is) set during the linking.

I have another change work-in-progress, which reduces the "noise" too. Not sure whether it would help in your case. What are the edges you want to remove with this patch?

My change should get rid of many edges like:

```
<edge b="Other:Mono.Tuner.SubStepDispatcher" e="Other:MonoDroid.Tuner.ApplyPreserveAttribute" />
```

where nothing was marked after we got to the place where mark might have potentially happen.

mrvoorhe:
My end goal is to write a test case, such as 

    [EnableReport]
    public class MethodCallDependency
    {
        static void Main()
        {
            UsedMethod();
        }

        public static void UsedMethod()
        {
            Cat();
        }

        public static void Cat() { Dog();}

        public static void Dog() { }
    }
`

And then in the linker-dependencies.xml all I want to see is the information relevant to this test case

`<?xml version="1.0" encoding="utf-8"?>
<dependencies version="1.0">
	<edge b="Other:Mono.Linker.Steps.ResolveFromAssemblyStep" e="Assembly:test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null" />
	<edge b="Assembly:test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null" e="TypeDef:Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency" />
	<edge b="Assembly:test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null" e="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::Main()" />
	<edge b="Assembly:test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null" e="TypeDef:System.Security.Permissions.SecurityPermissionAttribute" />
	<edge b="Assembly:test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null" e="TypeDef:Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency" />
	<edge b="TypeDef:Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency" e="Module:test.exe" />
	<edge b="Assembly:test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null" e="TypeDef:Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency" />
	<edge b="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::Main()" e="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::UsedMethod()" />
	<edge b="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::Main()" e="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::UsedMethod()" />
	<edge b="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::UsedMethod()" e="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::Cat()" />
	<edge b="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::UsedMethod()" e="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::Cat()" />
	<edge b="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::Cat()" e="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::Dog()" />
	<edge b="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::Cat()" e="Method:System.Void Unity.Linker.Tests.Cases.Special.Report.MethodCallDependency::Dog()" />
</dependencies>
`

radekdoulik:
I see a lot of I18N stuff preserved by `Mono.Linker.Steps.LoadI18nAssemblies` in the dump for such a case. Wonder if that's another opportunity to improve the linker itself.

mrvoorhe:
Sounds like it.

In our Linker we always turn off I18N stuff so I didn't notice that.

mrvoorhe:
I will add an opt-in command line flag to turn this behavior on after your Tracer PR lands.

linker/Linker/Annotations.cs
```diff
@@ -423,5 +434,60 @@ public void SaveDependencies ()
 			zipStream = null;
 			dependency_stack = null;
 		}
+
+		bool WillAssemblyBeModified (AssemblyDefinition assembly)
+		{
+			switch (_context.Annotations.GetAction (assembly))
```
mrvoorhe: @radekdoulik How do you want me to get to Annotations once I move this method to Tracer?

mrvoorhe:
PR is NOT ready to be merged yet

Changes made:

* This behavior is now opt-in.  You must pass `--reduced-tracing` to enable it

* Add a test to verify `--reduced-tracing` works.  This test currently fails.  It requires https://github.com/mono/linker/pull/210 to land first.

* Add a test to verify `--dump-dependencies` works.

linker/Linker/Driver.cs
```diff
@@ -109,6 +109,11 @@ void Run ()
 							continue;
 						}
 
+						if (token == "--reduced-tracing") {
```
radekdoulik: I think we need also a description of it in the `Usage` method.

mrvoorhe:
PR updated with changes

* Add new test for dumping full dependencies to uncompressed file

* Add support for dumping to uncompressed xml

* Address undesired behavior of ~6k edges from MarkStep => <Queued Method> when ReducedTracing is enabled

mrvoorhe:
build

mrvoorhe:
I don't think the build failure is related to my changes.

marek-safar:
build

marek-safar:
build

mrvoorhe:
Sorry some how I accidentally included solution file changes.

@marek-safar Fixed now

PR #209, opened 2017-12-20
Added UnintializedContextFactory class

radekdoulik:
This is a helper class, which is used in LinkContext to create
AnnotationStore, MarkingHelpers and Tracer. As the name suggest, it
uses uninitialized LinkContext. It is a slightly better than using
virtual methods in the LinkContext constructor as the class name hints
about the state of the context it uses.

It is added for Unity developers to be able to create their own
AnnotationStore, MarkingHelpers and Tracer. It is also needed for
https://github.com/mono/linker/pull/208

The class itself might be replaced later by interface with default
implementations.

### DIFF
--- a/linker/Linker/Annotations.cs
+++ b/linker/Linker/Annotations.cs
@@ -36,7 +36,7 @@ namespace Mono.Linker {
 
 	public class AnnotationStore {
 
-		readonly Tracer tracer;
+		protected readonly LinkContext context;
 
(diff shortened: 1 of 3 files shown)


radekdoulik:
@mrvoorhe would this work for your linker and for #208?

mrvoorhe:
@radekdoulik Looks good!  That will do the trick.  Thank you

PR #211, opened 2017-12-21
Don't root System.Private.CoreLib if it has an embedded root xml resource.

erozenfeld:
Once we start computing System.Private.CoreLib roots during coreclr build (https://github.com/dotnet/coreclr/pull/15525)
we won't need to root all of System.Private.CoreLib if the embedded xml
root file is present.

### DIFF
--- a/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
+++ b/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
@@ -293,20 +293,29 @@
        be part of the build step or pack step, OR pack could be made
        to work on the publish output. -->
   <Target Name="_ComputeLinkerRootAssemblies"
-          DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish;_ComputePlatformLibraries">
+          DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish;_ComputePlatformLibraries;_CheckSystemPrivateCorelibEmbeddedRoots">
     <!-- By default, roots are everything minus the framework
-         assemblies (except for System.Private.CoreLib, which we
-         always root for now). This doesn't include the intermediate
+         assemblies. This doesn't include the intermediate
          assembly, because we root it separately using an xml file,
-         which lets us explicitly root everything. -->
+         which lets us explicitly root everything. 
+         System.Private.CoreLib is rooted unless it has an embedded
+         xml descriptor file. -->
     <ItemGroup>
       <_LinkerRootAssemblies Include="@(_ManagedResolvedAssembliesToPublish->'%(Filename)')" />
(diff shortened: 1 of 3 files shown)


erozenfeld:
@sbomer PTAL

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -293,20 +293,29 @@
        be part of the build step or pack step, OR pack could be made
        to work on the publish output. -->
   <Target Name="_ComputeLinkerRootAssemblies"
-          DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish;_ComputePlatformLibraries">
+          DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish;_ComputePlatformLibraries;_CheckSystemPrivateCorelibEmbeddedRoots">
     <!-- By default, roots are everything minus the framework
-         assemblies (except for System.Private.CoreLib, which we
-         always root for now). This doesn't include the intermediate
+         assemblies.This doesn't include the intermediate
```
sbomer: nit: could you add a space after the period?

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -293,20 +293,29 @@
...
       <LinkerRootAssemblies Include="@(_LinkerRootAssemblies)" />
     </ItemGroup>
   </Target>
 
+    
+  <UsingTask TaskName="CheckEmbeddedRootDescriptor" AssemblyFile="$(LinkTaskDllPath)" />
+  <Target Name="_CheckSystemPrivateCorelibEmbeddedRoots"
+          DependsOnTargets="_ComputeManagedAssembliesToLink">
+    <CheckEmbeddedRootDescriptor AssemblyPath="@(_ManagedAssembliesToLink->WithMetadataValue('Filename', 'System.Private.CoreLib'))">
+      <Output TaskParameter="HasEmbeddedRootDescriptor" PropertyName="HasEmbeddedRootDescriptor" />
```
sbomer: I'd suggest making the output property name HasEmbeddedRootDescriptor "private" (prefix it with an underscore, by convention), and maybe rename it to something like _SPCHasEmbeddedRootDescriptor.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
erozenfeld: Done.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
erozenfeld: Done.

PR #212, opened 2017-12-23
Expose assembly actions in ILLink.

erozenfeld:
This change adds 4 linker publish options:
UsedApplicationAssemblyAction, UnusedApplicationAssemblyAction,
UsedNetCoreFrameworkAssemblyAction, and UnusedNetCoreFrameworkAssemblyAction.

ILLink task takes two sets of assemblies as inputs: application assemblies and
framework assemblies.
ILLink task emits the appropriate -p switches for each assembly.
ILLink task no longer emits -c switch.

### DIFF
--- a/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
+++ b/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
@@ -16,6 +16,10 @@
     <LinkDuringPublish Condition=" '$(LinkDuringPublish)' != 'true' ">false</LinkDuringPublish>
     <ShowLinkerSizeComparison Condition=" '$(ShowLinkerSizeComparison)' == '' ">false</ShowLinkerSizeComparison>
     <LinkerDumpDependencies Condition=" '$(LinkerDumpDependencies)' == '' ">false</LinkerDumpDependencies>
+    <UsedApplicationAssemblyAction Condition=" '$(UsedApplicationAssemblyAction)' == '' ">Copy</UsedApplicationAssemblyAction>
+    <UnusedApplicationAssemblyAction Condition=" '$(UnusedApplicationAssemblyAction)' == '' ">Delete</UnusedApplicationAssemblyAction>
+    <UsedPlatformAssemblyAction Condition=" '$(UsedPlatformAssemblyAction)' == '' ">Link</UsedPlatformAssemblyAction>
+    <UnusedPlatformAssemblyAction Condition=" '$(UnusedPlatformAssemblyAction)' == '' ">Delete</UnusedPlatformAssemblyAction>
   </PropertyGroup>
 
   <!-- This depends on LinkDuringPublish, so it needs to be imported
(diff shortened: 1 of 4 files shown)


erozenfeld:
@sbomer PTAL

corebuild/integration/ILLink.Tasks/LinkTask.cs
```diff
@@ -12,16 +12,42 @@ namespace ILLink.Tasks
...
+		///   Action to perform on used .NET Core Framework assemblies.
+		/// </summary>
+		[Required]
+		public ITaskItem UsedNetCoreFrameworkAssemblyAction { get; set; }
+
+		/// <summary>
+		///   Action to perform on unused .NET Core Framework assemblies.
+		/// </summary>
+		[Required]
+		public ITaskItem UnusedNetCoreFrameworkAssemblyAction { get; set; }
```
sbomer: In order to keep the ILLink task easy to use, I don't think we should introduce separate inputs for used/unused app/framework assemblies. The idea is that someone familiar with the mono linker could easily use LinkTask without the rest of the publish targets to implement a custom linker workflow.

I'd suggest computing the action per assembly outside of this msbuild task, somewhere in the publish logic in ILLink.Tasks.targets. Then I can think of two good ways to pass the actions through to this task:
- add an AssemblyAction metadata to the input AssemblyPaths
- introduce an input itemgroup AssemblyActions that contains assembly names with Action metadata

The first feels cleaner and less redundant to me, so that would be my suggestion.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -226,15 +235,18 @@
...
-          DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish">
+          DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish;_ComputePlatformLibraries">
     <ItemGroup>
       <_ManagedAssembliesToLink Include="@(IntermediateAssembly)" />
       <_ManagedAssembliesToLink Include="@(_ManagedResolvedAssembliesToPublish)" />
+      <_NetCoreFrameworkAssembliesToLink Include="@(PlatformLibraries)" />
+      <_NetCoreFrameworkAssembliesToLink Include="@(_ManagedResolvedAssembliesToPublish->WithMetadataValue('Filename', 'System.Private.CoreLib'))" />
+      <_ApplicationAssembliesToLink Include="@(_ManagedAssembliesToLink)" />
+      <_ApplicationAssembliesToLink Remove="@(_NetCoreFrameworkAssembliesToLink)" />
     </ItemGroup>
```
sbomer: Instead of computing separate itemgroups here, I would suggest adding metadata to ManagedAssembliesToLink with the assembly action. See my comment below for an explanation.

corebuild/integration/ILLink.Tasks/LinkTask.cs
```diff
@@ -59,51 +85,116 @@ public class ILLink : Task
...
+				}
+				assemblyAction = "skip";
+				break;
+
+			default:
+				return false;
+			}
+
+			return true;
+		}
```
sbomer: Can you explain the rationale for this validation? This is just setting reasonable default modes, right? I think this should live in the msbuild targets to make it easy to override.

erozenfeld:
@sbomer I made changes to address your comments. PTAL.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -166,22 +170,41 @@
...
+                          UsedApplicationAssemblyAction="$(UsedApplicationAssemblyAction)"
+                          UnusedApplicationAssemblyAction="$(UnusedApplicationAssemblyAction)"
+                          UsedPlatformAssemblyAction="$(UsedPlatformAssemblyAction)"
+                          UnusedPlatformAssemblyAction="$(UnusedPlatformAssemblyAction)">
+        <Output TaskParameter="AssemblyPathsWithActions" ItemName="_ManagedAssembliesToLinkWithActions" />
+      </SetAssemblyActions>
+
+      <ItemGroup>
+          <_ManagedAssembliesToLink Remove="@(_ManagedAssembliesToLink)"/>
+          <_ManagedAssembliesToLink Include ="@(_ManagedAssembliesToLinkWithActions)"/>
```
sbomer: Extra space after `Include`. Please also put spaces before the closing `/>`.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -226,15 +249,18 @@
...
   <Target Name="_ComputeManagedAssembliesToLink"
-          DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish">
+          DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish;_ComputePlatformLibraries">
     <ItemGroup>
       <_ManagedAssembliesToLink Include="@(IntermediateAssembly)" />
       <_ManagedAssembliesToLink Include="@(_ManagedResolvedAssembliesToPublish)" />
+      <_PlatformAssembliesToLink Include="@(PlatformLibraries->'%(Filename)')" />
+      <_PlatformAssembliesToLink Include="System.Private.CoreLib" />
+      <_ApplicationAssembliesToLink Include="@(_ManagedAssembliesToLink->'%(Filename)')" />
+      <_ApplicationAssembliesToLink Remove="@(_PlatformAssembliesToLink)" />
```
sbomer: Could you compute these itemgroups in the target `_SetAssemblyActions` (and make `_ComputePlatformLibraries` a dependency of `_SetAssemblyActions` instead), since they're only needed by that logic? There's at least one case in which `_ComputeManagedAssembliesToLink` is called, but `_SetAssemblyActions` isn't (during an incremental build when the linker inputs are unchanged, `_ManagedAssembliesToLink` is used for the comparison report).

corebuild/integration/ILLink.Tasks/LinkTask.cs
```diff
@@ -59,51 +57,62 @@ public class ILLink : Task
...
 			if (ExtraArgs != null) {
-				args.AddRange(ExtraArgs.Split(' '));
+				args.AddRange (ExtraArgs.Split (' '));
 			}
 
 			if (DumpDependencies)
 				args.Add ("--dump-dependencies");
 
-			return args.ToArray();
+			return args.ToArray ();
```
sbomer: Thanks for fixing the conventions in this file.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
erozenfeld: Done.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
erozenfeld: Done.

erozenfeld:
@sbomer I addressed all comments and also rebased. PTAL.

PR #214, opened 2018-01-04
Update JitBench git revision for MusicStore testing.

erozenfeld:
This revision points to the updated aspnet dev feed.


erozenfeld:
@sbomer PTAL

PR #215, opened 2018-01-05
Expose RootAllApplicationAssemblies property.

erozenfeld:
The default is to root all application assemblies (and SPC if necessary).
The new property can be used to alter that behavior and
root only the Main method of the application exe (and SPC if necessary).

### DIFF
--- a/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
+++ b/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
@@ -20,6 +20,8 @@
     <UnusedApplicationAssemblyAction Condition=" '$(UnusedApplicationAssemblyAction)' == '' ">Delete</UnusedApplicationAssemblyAction>
     <UsedPlatformAssemblyAction Condition=" '$(UsedPlatformAssemblyAction)' == '' ">Link</UsedPlatformAssemblyAction>
     <UnusedPlatformAssemblyAction Condition=" '$(UnusedPlatformAssemblyAction)' == '' ">Delete</UnusedPlatformAssemblyAction>
+    <RootAllApplicationAssemblies Condition=" '$(RootAllApplicationAssemblies)' == '' ">true</RootAllApplicationAssemblies>
+    <RootAllApplicationAssemblies Condition=" '$(RootAllApplicationAssemblies)' != 'true' ">false</RootAllApplicationAssemblies>
   </PropertyGroup>
 
   <!-- This depends on LinkDuringPublish, so it needs to be imported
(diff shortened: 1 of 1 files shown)


erozenfeld:
@sbomer PTAL

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -328,21 +330,26 @@
...
     <ItemGroup>
       <_LinkerRootAssemblies Include="@(_ManagedResolvedAssembliesToPublish->'%(Filename)')" />
       <_LinkerRootAssemblies Remove="@(PlatformLibraries->'%(Filename)')" />
       <_LinkerRootAssemblies Remove="System.Private.CoreLib" Condition=" '$(_SPCHasEmbeddedRootDescriptor)' == 'true' "/>
-      <LinkerRootAssemblies Include="@(_LinkerRootAssemblies)" />
+      <LinkerRootAssemblies Include="@(_LinkerRootAssemblies)" Condition=" '$(RootAllApplicationAssemblies)' == 'true' " />
+
+      <LinkerRootAssemblies Include="@(IntermediateAssembly)" Condition=" '$(RootAllApplicationAssemblies)' == 'false' " />
+      <LinkerRootAssemblies Include="System.Private.CoreLib" Condition=" '$(RootAllApplicationAssemblies)' == 'false' And '$(_SPCHasEmbeddedRootDescriptor)' == 'false' "/>
     </ItemGroup>
```
sbomer: I think we should only compute `_LinkerRootAssemblies` when rooting all app assemblies, by doing something like this:
```
<ItemGroup Condition=" '$(RootAllApplicationAssemblies)' == 'true' ">
  <_LinkerRootAssemblies Include="@(_ManagedResolvedAssembliesToPublish->'%(Filename)')" />
  <_LinkerRootAssemblies Remove="@(PlatformLibraries->'%(Filename)')" />
  <_LinkerRootAssemblies Remove="System.Private.CoreLib" Condition=" '$(_SPCHasEmbeddedRootDescriptor)' == 'true' "/>
  <LinkerRootAssemblies Include="@(_LinkerRootAssemblies)" />
</ItemGroup>

<ItemGroup Condition=" '$(RootAllApplicationAssemblies)' != 'true' ">
  <LinkerRootAssemblies Include="@(IntermediateAssembly)" />
  <LinkerRootAssemblies Include="System.Private.CoreLib" Condition=" '$(_SPCHasEmbeddedRootDescriptor)' == 'false' "/>
</ItemGroup>
```

(Separating `_LinkerRootAssemblies` from `LinkerRootAssemblies` is useful in the first case because it makes it possible to manually root a platform assembly or SPC.

erozenfeld: @sbomer I made the change, PTAL.

PR #222, opened 2018-01-12
Fix case where assembly attribute wouldn't be correctly marked

mrvoorhe:
Top level CustomAttributes are marked late, after all of the main marking happens.  This lets us remove attributes when the attribute type would be the only thing used from a reference.  After marking these attributes we would process the queue one more time.  However, if that second processing pass gave us a reason to keep another top level attribute, we wouldn't circle back and process the top level attributes again

Now, we will process the main queues and the late marked attributes repeatedly until no more items are being marked

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -131,13 +131,7 @@ namespace Mono.Linker.Steps {
 			if (QueueIsEmpty ())
 				throw new InvalidOperationException ("No entry methods");
 
-			ProcessEntireQueue ();
-
-			// After all types have been processed then we can process the lazily marked attributes
-			ProcessLazyAttributes ();
-
-			// We need to process the queue again in case marking the attributes enqueued more work
-			ProcessEntireQueue ();
+			while (ProcessPrimaryQueue () || ProcessLazyAttributes ())
 
 			// deal with [TypeForwardedTo] pseudo-attributes
 			foreach (AssemblyDefinition assembly in _context.GetAssemblies ()) {
@@ -173,13 +167,18 @@ namespace Mono.Linker.Steps {
 			}
 		}
 
-		void ProcessEntireQueue ()
+		bool ProcessPrimaryQueue ()
 		{
+			if (QueueIsEmpty ())
+				return false;
+
 			while (!QueueIsEmpty ()) {
 				ProcessQueue ();
 				ProcessVirtualMethods ();
 				DoAdditionalProcessing ();
 			}
+
+			return true;
 		}
 
 		void ProcessQueue ()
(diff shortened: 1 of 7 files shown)


linker/Linker.Steps/MarkStep.cs
```diff
@@ -131,13 +131,7 @@ void Process ()
...
 				throw new InvalidOperationException ("No entry methods");
 
-			ProcessEntireQueue ();
-
-			// After all types have been processed then we can process the lazily marked attributes
-			ProcessLazyAttributes ();
-
-			// We need to process the queue again in case marking the attributes enqueued more work
-			ProcessEntireQueue ();
+			while (ProcessAllQueues ())
```
marek-safar: I think `while (ProcessPrimaryQueue () || ProcessLazyAttributes ())` would be better

linker/Linker.Steps/MarkStep.cs
```diff
@@ -541,13 +572,23 @@ void ProcessLazyAttributes ()
...
 
 				MarkCustomAttribute (customAttribute);
 			}
+
+			// requeue the items we skipped in case we need to make another pass
+			foreach (var item in skippedItems)
+				_topLevelAttributes.Enqueue (item);
+
+			// If we have the same number of items afterwards, then we didn't do any additional marking
+			if (startingQueueCount == _topLevelAttributes.Count)
```
marek-safar: I am not certain this will always work. Could the queue have an item removed and added at the same time?

linker/Linker.Steps/MarkStep.cs
mrvoorhe: This PR is serving two purposes, (1) fix an obscure bug, (2) set the stage for another enhancement I'm working on.   In this next change I'm working on, I may end up introducing another queue that needs to be processed.  I thought it would be nice to establish a method for handling how we process the queues in this simple PR.  

Would you still prefer the `||` approach if there were 3 methods?

linker/Linker.Steps/MarkStep.cs
mrvoorhe: You're right.  `MarkCustomAttribute` could call  `ShouldParseMethodBody`, which could call `ResolveAssembly`.  And if that happened `MarkAssembly` would be called and new assembly or module attributes could be enqueued.

I will come up with at test case and switch to a more robust way of tracking if anything new was marked.

mrvoorhe: I tried to write a test case and I wasn't able to figure out a way to cause a problem.  We call `LazyMarkCustomAttributes` on all assemblies returned by `_context.GetAssemblies()` before the first call to `ProcessAllQueues` even takes place. 

To cause an issue, you'd have to get the `ResolveAssembly` call in `ShouldParseMethodBody` to be called on an assembly that was not returned by `_context.GetAssemblies()`

I'm not sure how you'd do that and I don't think it's worth the effort to figure out.    So I will not be including an additional test case.

There was a straightward way to make the return value more robust, so I will still update the PR with that change.

linker/Linker.Steps/MarkStep.cs
marek-safar: I'll see but in this PR `||` approach is much better

mrvoorhe: Done

PR #223, opened 2018-01-12
Allow the linker to detect basic reflection usage

joshpeterson:
The linker can now understand simple cases where reflection is used, via
GetMethod, GetProperty, etc. along with a ldtoken opcode for a type and
a ldstr opcode for a name. The linker will mark that named thing in that
type.

This helps us handle cases in the class libraries that use reflection,
e.g. CreateLambda in System.Linq.Expressions.Expression.

Generally, the approach here is pretty conservative, attempting to keep
more code when we are unsure about what to do. For example, all
overloads of a given method are kept, even if only one is invoked. From
the IL instruction stream, it is too difficult now to understand which
overload is invoked.


linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+					}
+				}
+			}
+		}
+
+		void MarkConstructorUsedViaReflection (Collection<Instruction> instructions, string unused1, TypeDefinition declaringTypeOfConstructorInvokedViaReflection, BindingFlags bindingFlags)
+		{
+			if (declaringTypeOfConstructorInvokedViaReflection != null) {
+				foreach (var method in declaringTypeOfConstructorInvokedViaReflection.Methods) {
+					if ((bindingFlags == BindingFlags.Default || bindingFlags.HasFlag(BindingFlags.Public) == method.IsPublic) && method.Name == ".ctor")
```
mrvoorhe: Space after `HasFlag ()`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+			for (var i = 0; i < instructions.Count; i++) {
+				var instruction = instructions [i];
+				if (instruction.OpCode == OpCodes.Call || instruction.OpCode == OpCodes.Callvirt) {
+					var methodBeingCalled = instruction.Operand as MethodReference;
+					if (methodBeingCalled != null && methodBeingCalled.DeclaringType.Name == "Type" && methodBeingCalled.DeclaringType.Namespace == "System")
+					{
+						if (methodBeingCalled.Name == reflectionMethod) {
+							_context.Tracer.Push ($"Reflection-{methodBeingCalled}");
+							var nameOfThingUsedViaReflection = OperandOfNearestInstructionBefore<string> (i, OpCodes.Ldstr, instructions);
+							var bindingFlags = (BindingFlags)OperandOfNearestInstructionBefore<sbyte> (i, OpCodes.Ldc_I4_S, instructions);
```
mrvoorhe: Space after `(BindingFlags)`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+
+							// There might be more than one ldtoken opcode above the call in the IL stream. Be conservative and check all of
+							// the types which were loaded for the method being used.
+							var declaringTypesOfThingInvokedViaReflection = OperandsOfInstructionsBefore<TypeReference> (i, OpCodes.Ldtoken, instructions);
+							foreach (var declaringTypeOfThingInvokedViaReflection in declaringTypesOfThingInvokedViaReflection) {
+								var typeDefinition = declaringTypeOfThingInvokedViaReflection?.Resolve ();
+								markMethod(instructions, nameOfThingUsedViaReflection, typeDefinition, bindingFlags);
```
mrvoorhe: space after `markMethod`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+				}
+			}
+		}
+
+		void MarkMethodUsedViaReflection (Collection<Instruction> instructions, string nameOfMethodInvokedViaReflection, TypeDefinition declaringTypeOfMethodInvokedViaReflection, BindingFlags bindingFlags)
+		{
+			if (nameOfMethodInvokedViaReflection != null && declaringTypeOfMethodInvokedViaReflection != null) {
+				foreach (var method in declaringTypeOfMethodInvokedViaReflection.Methods)
+				{
+					if ((bindingFlags == BindingFlags.Default || bindingFlags.HasFlag(BindingFlags.Public) == method.IsPublic && bindingFlags.HasFlag(BindingFlags.Static) == method.IsStatic)
```
mrvoorhe: Space after `HasFlag`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+						MarkEvent (eventInfo);
+					}
+				}
+			}
+		}
+
+		static TOperand OperandOfNearestInstructionBefore<TOperand> (int startingInstructionIndex, OpCode opCode, IList<Instruction> instructions)
+		{
+			for (var i = startingInstructionIndex; i >= 0; i--)
+			{
```
mrvoorhe: `{` on same line as for

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+				if (instructions [i].OpCode == opCode)
+					return (TOperand)instructions[i].Operand;
```
mrvoorhe: Space after `(TOperand)`.  Space before `[i]`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+			}
+
+			return default (TOperand);
+		}
+
+		static List<TOperand> OperandsOfInstructionsBefore<TOperand>(int startingInstructionIndex, OpCode opCode, IList<Instruction> instructions) where TOperand : class
+		{
+			var operands = new List<TOperand>();
+			for (var i = startingInstructionIndex; i >= 0; i--)
+			{
```
mrvoorhe: `{` on same line as for

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+				if (instructions [i].OpCode == opCode)
+					return (TOperand)instructions[i].Operand;
+			}
+
+			return default (TOperand);
+		}
+
+		static List<TOperand> OperandsOfInstructionsBefore<TOperand>(int startingInstructionIndex, OpCode opCode, IList<Instruction> instructions) where TOperand : class
+		{
+			var operands = new List<TOperand>();
```
mrvoorhe: Space before `()`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+			for (var i = startingInstructionIndex; i >= 0; i--)
+			{
+				if (instructions [i].OpCode == opCode)
+					return (TOperand)instructions[i].Operand;
+			}
+
+			return default (TOperand);
+		}
+
+		static List<TOperand> OperandsOfInstructionsBefore<TOperand>(int startingInstructionIndex, OpCode opCode, IList<Instruction> instructions) where TOperand : class
```
mrvoorhe: Space after `<TOperand>`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+		{
+			var operands = new List<TOperand>();
+			for (var i = startingInstructionIndex; i >= 0; i--)
+			{
+				if (instructions [i].OpCode == opCode)
+				{
+					var type = instructions[i].Operand as TOperand;
```
mrvoorhe: Space before `[i]`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,120 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+					if (type != null)
+						operands.Add(type);
```
mrvoorhe: Space after `Add`

linker/Tests/TestCases/TestSuites.cs
```diff
@@ -72,8 +72,14 @@ public void TypeForwardingTests (TestCase testCase)
 			Run (testCase);
 		}
 
-		[TestCaseSource(typeof (TestDatabase), nameof (TestDatabase.TestFrameworkTests))]
-		public void TestFrameworkTests (TestCase testCase)
+		[TestCaseSource(typeof(TestDatabase), nameof(TestDatabase.TestFrameworkTests))]
```
mrvoorhe: Avoid formatting changes TestFrameworkTests

linker/Tests/Mono.Linker.Tests.Cases/Reflection/ConstructorUsedViaReflection.cs
```diff
@@ -0,0 +1,28 @@
...
+		private class OnlyUsedViaReflection {
+			[Kept]
+			public OnlyUsedViaReflection ()
+			{ }
+
+			[Kept]
+			public OnlyUsedViaReflection(string bar)
+			{ }
+
+			private OnlyUsedViaReflection (int foo)
```
marek-safar: please also add protected and internal cases

linker/Linker.Steps/MarkStep.cs
```diff
@@ -30,11 +30,16 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
```
marek-safar: What do you need this for? This adds hard version dependency (e.g for type casts) which is very problematic for general IL tool.

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,116 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
 		}
+
+
+		void MarkSomethingUsedViaReflection (string reflectionMethod, Action<Collection<Instruction>, string, TypeDefinition, BindingFlags> markMethod, Collection<Instruction> instructions)
+		{
+			for (var i = 0; i < instructions.Count; i++) {
+				var instruction = instructions [i];
+				if (instruction.OpCode == OpCodes.Call || instruction.OpCode == OpCodes.Callvirt) {
+					var methodBeingCalled = instruction.Operand as MethodReference;
+					if (methodBeingCalled != null && methodBeingCalled.DeclaringType.Name == "Type" && methodBeingCalled.DeclaringType.Namespace == "System")
```
marek-safar: Please use `continue` instead of deep nesting

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,116 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+					{
+						if (methodBeingCalled.Name == reflectionMethod) {
+							_context.Tracer.Push ($"Reflection-{methodBeingCalled}");
+							var nameOfThingUsedViaReflection = OperandOfNearestInstructionBefore<string> (i, OpCodes.Ldstr, instructions);
+							var bindingFlags = (BindingFlags) OperandOfNearestInstructionBefore<sbyte> (i, OpCodes.Ldc_I4_S, instructions);
```
marek-safar: You cannot cast it to BindingFlags as it can be a different type, e.g mscorlib and netcore or version mismatch etc

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,116 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+									markMethod (instructions, nameOfThingUsedViaReflection, typeDefinition, bindingFlags);
+							}
+							_context.Tracer.Pop ();
+						}
+					}
+				}
+			}
+		}
+
+		void MarkConstructorUsedViaReflection (Collection<Instruction> instructions, string unused1, TypeDefinition declaringTypeOfConstructorInvokedViaReflection, BindingFlags bindingFlags)
```
marek-safar: * It marks more than one
* Repeating the name of method in parameters is not the convention we use

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,116 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+		{
+			for (var i = 0; i < instructions.Count; i++) {
+				var instruction = instructions [i];
+				if (instruction.OpCode == OpCodes.Call || instruction.OpCode == OpCodes.Callvirt) {
+					var methodBeingCalled = instruction.Operand as MethodReference;
+					if (methodBeingCalled != null && methodBeingCalled.DeclaringType.Name == "Type" && methodBeingCalled.DeclaringType.Namespace == "System")
+					{
+						if (methodBeingCalled.Name == reflectionMethod) {
+							_context.Tracer.Push ($"Reflection-{methodBeingCalled}");
+							var nameOfThingUsedViaReflection = OperandOfNearestInstructionBefore<string> (i, OpCodes.Ldstr, instructions);
```
marek-safar: Should skip when it's `null`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,116 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+		{
+			foreach (var method in declaringTypeOfConstructorInvokedViaReflection.Methods) {
+				if ((bindingFlags == BindingFlags.Default || bindingFlags.HasFlag (BindingFlags.Public) == method.IsPublic) && method.Name == ".ctor")
+					MarkMethod (method);
+			}
+		}
+
+		void MarkMethodUsedViaReflection (Collection<Instruction> instructions, string nameOfMethodInvokedViaReflection, TypeDefinition declaringTypeOfMethodInvokedViaReflection, BindingFlags bindingFlags)
+		{
+			if (nameOfMethodInvokedViaReflection != null) {
```
marek-safar: Caller should handle that

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1598,116 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+		{
+			for (var i = startingInstructionIndex; i >= 0; i--) {
+				if (instructions [i].OpCode == opCode)
+					return (TOperand )instructions [i].Operand;
+			}
+
+			return default (TOperand);
+		}
+
+		static List<TOperand> OperandsOfInstructionsBefore<TOperand> (int startingInstructionIndex, OpCode opCode, IList<Instruction> instructions) where TOperand : class
```
marek-safar: I think this is unnecessary generalization, it should check for ldtoken with TypeReference

linker/Linker.Steps/MarkStep.cs
joshpeterson: This is used for the `BindingFlags` enum. I can instead fully qualify `BindingFlags` where it is used. Is that preferred?

linker/Linker.Steps/MarkStep.cs
joshpeterson: We can't skip with the `nameOfThingUsedViaReflection`, because we want to handle constructors. We don't have a previous `ldstr` instruction in the IL stream for constructors. Hence we need to pass a possibly null value to the method which does the work, and let it decide what to do.

linker/Linker.Steps/MarkStep.cs
joshpeterson: > It marks more than one
I'll rename this method to be `MarkConstructorsUsedViaReflection`

> Repeating the name of method in parameters is not the convention we use
I'm unsure what you mean by this, sorry. Are you asking me to rename `declaringTypeOfConstructorInvokedViaReflection`?

linker/Linker.Steps/MarkStep.cs
joshpeterson: I would prefer that the caller not handle this null check, because of the constructor case mentioned above.

linker/Linker.Steps/MarkStep.cs
joshpeterson: I'm unclear about what you mean here, can you elaborate?

linker/Linker.Steps/MarkStep.cs
joshpeterson: I agree, I'll remove the generic here.

linker/Tests/Mono.Linker.Tests.Cases/Reflection/ConstructorUsedViaReflection.cs
joshpeterson: I've added these.

linker/Linker.Steps/MarkStep.cs
joshpeterson: I'll make this change.

linker/Linker.Steps/MarkStep.cs
marek-safar: My preference (if you really need it) is to use `SR` alias for System.Reflection

linker/Linker.Steps/MarkStep.cs
marek-safar: You are casting the value to BindingFlags type which is used when linker is compiled and that's wrong. Imagine you are linking 2.0 mscorlib but BindingFlags type used for linker compilation is 4.0 mscorlib (there are more variations of this)

linker/Linker.Steps/MarkStep.cs
joshpeterson: Thanks. I guess this comes down to the use of `BindingsFlags` then. If we need to keep using it then, I'll switch this to `SR`. Otherwise, I'll remove it.

linker/Linker.Steps/MarkStep.cs
marek-safar: I'd prefer to copy the BindingFlags enum into linker for this purposes instead of reusing System.Reflection

joshpeterson: Ok, that makes sense. So does that mean the linker cannot assume a given `sbyte` value loaded on the stack corresponds to a given value in the `BindingsFlags` enum?

linker/Linker.Steps/MarkStep.cs
marek-safar: yes `declaringTypeOfConstructorInvokedViaReflection` is just awfully long and it repeats already know information

joshpeterson: Ok, I'll rename this parameter.

linker/Linker.Steps/MarkStep.cs
joshpeterson: Why should we copy the `BindingFlags` enum into the linker code? We would be grabbing the value at a specific point in time. I'm not sure this would help with a possible future change to the enum values.

joshpeterson:
I've updated the code here based on the feedback, and I've squashed the changes down to one commit. Please have another look at this PR. Thanks!

linker/Linker.Steps/MarkStep.cs
```diff
@@ -35,6 +35,10 @@
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using Mono.Collections.Generic;
+using CustomAttributeNamedArgument = Mono.Cecil.CustomAttributeNamedArgument;
```
marek-safar: These are not needed

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1597,143 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+				var methodBeingCalled = instruction.Operand as MethodReference;
+				if (methodBeingCalled == null || methodBeingCalled.DeclaringType.Name != "Type" || methodBeingCalled.DeclaringType.Namespace != "System")
+					continue;
+
+				if (methodBeingCalled.Name != reflectionMethod)
+					continue;
+
+				_context.Tracer.Push ($"Reflection-{methodBeingCalled}");
+				var nameOfThingUsedViaReflection = OperandOfNearestInstructionBefore<string> (i, OpCodes.Ldstr, instructions);
+				var bindingFlags = (BindingFlags) OperandOfNearestInstructionBefore<sbyte> (i, OpCodes.Ldc_I4_S, instructions);
```
marek-safar: You could use `BindingFlags` instead of `<sbyte>`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1597,143 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+		{
+			foreach (var method in declaringType.Methods) {
+				if ((bindingFlags == BindingFlags.Default || bindingFlags.HasFlag (BindingFlags.Public) == method.IsPublic) && method.Name == ".ctor")
+					MarkMethod (method);
+			}
+		}
+
+		void MarkMethodsUsedViaReflection (Collection<Instruction> instructions, string name, TypeDefinition declaringType, BindingFlags bindingFlags)
+		{
+			if (name != null) {
```
marek-safar: Use `if (name == null) return` instead here and bellow for same pattern

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1597,143 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+				foreach (var method in declaringType.Methods)
+				{
+					if ((bindingFlags == BindingFlags.Default || bindingFlags.HasFlag (BindingFlags.Public) == method.IsPublic && bindingFlags.HasFlag(BindingFlags.Static) == method.IsStatic)
```
marek-safar: Don't use HasFlags it's slow

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1597,143 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+			foreach (var method in declaringType.Methods) {
+				if ((bindingFlags == BindingFlags.Default || bindingFlags.HasFlag (BindingFlags.Public) == method.IsPublic) && method.Name == ".ctor")
+					MarkMethod (method);
+			}
+		}
+
+		void MarkMethodsUsedViaReflection (Collection<Instruction> instructions, string name, TypeDefinition declaringType, BindingFlags bindingFlags)
+		{
+			if (name != null) {
+				foreach (var method in declaringType.Methods)
```
marek-safar: Inconsistent `{` formatting

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1597,143 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+			}
+
+			return default (TOperand);
+		}
+
+		static List<TypeReference> OperandsOfInstructionsBefore (int startingInstructionIndex, OpCode opCode, IList<Instruction> instructions)
+		{
+			var operands = new List<TypeReference> ();
+			for (var i = startingInstructionIndex; i >= 0; i--)
+			{
```
marek-safar: Following coding style for `{`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1587,5 +1597,143 @@ protected virtual void MarkInterfaceImplementation (TypeDefinition type, Interfa
...
+				if (instructions [i].OpCode == opCode)
+				{
```
marek-safar: Following coding style for `{`

linker/Linker.Steps/MarkStep.cs
joshpeterson: I've removed these `using` statements.

linker/Linker.Steps/MarkStep.cs
joshpeterson: This is not possible, since the operand for the `Ldc_I4_S` opcode is an `sbyte`. We need to find the `sbyte`, then cast the result to `BindingFlags`.

linker/Linker.Steps/MarkStep.cs
joshpeterson: I've fixed this.

linker/Linker.Steps/MarkStep.cs
joshpeterson: Fixed.

linker/Linker.Steps/MarkStep.cs
joshpeterson: Fixed

linker/Linker.Steps/MarkStep.cs
joshpeterson: Fixed

linker/Linker.Steps/MarkStep.cs
joshpeterson: Why is `HasFlag` slow? I much prefer it to manual bitwise operations, which can be tricky to get correct in my experience.

joshpeterson: Ok, after some more discussion, I'll not use `HasFlag` and roll my own instead.

linker/Linker.Steps/MarkStep.cs
marek-safar: It does only casts of the result so `OperandOfNearestInstructionBefore<BindingFlags> (i, OpCodes.Ldc_I4_S, instructions);` should do exactly same

marek-safar: Now looking at this again i4_s will be used only for low values, if you write a sample with `ExactBinding` you'd need to tweak this code

joshpeterson: I don't think we need to worry about `ExactBindings`. Maybe that is something we could add in the future, but I'd only like to support limited bindings flags now.

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1540,6 +1540,12 @@ protected virtual void MarkMethodBody (MethodBody body)
 
 			foreach (Instruction instruction in body.Instructions)
 				MarkInstruction (instruction);
+
+			MarkSomethingUsedViaReflection ("GetConstructor", MarkConstructorsUsedViaReflection, body.Instructions);
+			MarkSomethingUsedViaReflection ("GetMethod", MarkMethodsUsedViaReflection, body.Instructions);
+			MarkSomethingUsedViaReflection ("GetProperty", MarkPropertyUsedViaReflection, body.Instructions);
+			MarkSomethingUsedViaReflection ("GetField", MarkFieldUsedViaReflection, body.Instructions);
+			MarkSomethingUsedViaReflection ("GetEvent", MarkEventUsedViaReflection, body.Instructions);
 		}
```
marek-safar: It'd be also nice to extend this to support extension methods on RuntimeReflectionExtensions (e.g. GetRuntimeMethod)

joshpeterson: I agree, although I think that is outside the scope of this PR.

PR #225, opened 2018-01-15
Add missing null checks to TypeReferenceExtensions

spouliot:
Cecil's MetadataResolver returns null if it cannot resolve something.

This requires the callers, like TypeReferenceExtensions, to do null
checks, otherwise a  `NullReferenceException` can happens in the linker
code and it's not possible to give a useful error message to the
developer.

This is what's happening in
https://github.com/xamarin/xamarin-macios/issues/3215 (with test case)
and likely in
https://github.com/xamarin/xamarin-macios/issues/3210 (without test case)

### DIFF
--- a/linker/Linker/TypeReferenceExtensions.cs
+++ b/linker/Linker/TypeReferenceExtensions.cs
@@ -28,7 +28,7 @@ namespace Mono.Linker
 
 			var genericInstance = type as GenericInstanceType;
 			if (genericInstance != null) {
-				var baseType = type.Resolve ().BaseType;
+				var baseType = type.Resolve ()?.BaseType;
 				var baseTypeGenericInstance = baseType as GenericInstanceType;
 
 				if (baseTypeGenericInstance != null)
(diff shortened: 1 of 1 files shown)


linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -37,14 +37,14 @@ public static TypeReference GetInflatedBaseType (this TypeReference type)
...
-			return type.Resolve ().BaseType;
+			return type.Resolve ()?.BaseType;
 		}
 
 		public static IEnumerable<TypeReference> GetInflatedInterfaces (this TypeReference typeRef)
 		{
 			var typeDef = typeRef.Resolve ();
 
-			if (!typeDef.HasInterfaces)
+			if ((typeDef == null) || !typeDef.HasInterfaces)
```
marek-safar: You could use `if (typeDef?.HasInterfaces != true)` syntax

linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -166,7 +166,7 @@ public static IEnumerable<MethodReference> GetMethods (this TypeReference type)
 		{
 			var typeDef = type.Resolve ();
 
-			if (!typeDef.HasMethods)
+			if ((typeDef == null) || !typeDef.HasMethods)
```
marek-safar: Same as above

marek-safar:
Replaced by #234

spouliot:
thanks! 👍

PR #230, opened 2018-01-17
Add more delegate and event test cases

mrvoorhe:
I added new `[KeptEventAddMethod]` and `[KeptEventRemoveMethod]` attributes so that the expected result could be different for the two in the future.


marek-safar:
There are few tests failures

mrvoorhe:
Thanks.  I started looking into them yesterday but didn't get a chance to fix them.

PR #232, opened 2018-01-18
Add more delegate and event test cases

mrvoorhe:
I added new [KeptEventAddMethod] and [KeptEventRemoveMethod] attributes so that the expected result could be different for the two in the future.

A new [KeptDelegateCacheField] was introduced to help with asserting that the `<>f__mg$cache` fields generated by mcs are kept.

### DIFF
--- a/linker/Tests/TestCasesRunner/AssemblyChecker.cs
+++ b/linker/Tests/TestCasesRunner/AssemblyChecker.cs
@@ -12,7 +12,7 @@ namespace Mono.Linker.Tests.TestCasesRunner {
 		readonly AssemblyDefinition originalAssembly, linkedAssembly;
 
 		HashSet<string> linkedMembers;
-		HashSet<string> verifiedBackingFields = new HashSet<string> ();
+		HashSet<string> verifiedGeneratedFields = new HashSet<string> ();
 		HashSet<string> verifiedEventMethods = new HashSet<string>();
 
 		public AssemblyChecker (AssemblyDefinition original, AssemblyDefinition linked)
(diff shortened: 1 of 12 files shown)


linker/Tests/TestCasesRunner/AssemblyChecker.cs
```diff
@@ -371,6 +372,30 @@ protected virtual IEnumerable<string> FilterLinkedSecurityAttributes (ISecurityD
 				.Select (attr => attr.AttributeType.ToString ());
 		}
 
+		void VerifyDelegateBackingFields (TypeDefinition src, TypeDefinition linked)
+		{
+			// This behavior is unique to mcs.  If mcs was not used to compile the test case we can return
+			if (!UsingMcsCompiler)
```
marek-safar: I don't think this is necessary

linker/Tests/TestCasesRunner/AssemblyChecker.cs
```diff
@@ -11,8 +11,10 @@ namespace Mono.Linker.Tests.TestCasesRunner {
 	public class AssemblyChecker {
 		readonly AssemblyDefinition originalAssembly, linkedAssembly;
 
+		private const string McsDelegateCacheFieldPrefix = "<>f__mg$cache";
```
marek-safar: Should go to the score where it's used

linker/Tests/TestCasesRunner/AssemblyChecker.cs
mrvoorhe: It's for our UnityLinker test suites.  We always use mcs, even on Windows.

linker/Tests/TestCasesRunner/AssemblyChecker.cs
mrvoorhe: Done

linker/Tests/TestCasesRunner/AssemblyChecker.cs
marek-safar: Just remove the check and it's still going to work for you

mrvoorhe: Ah, I see what you mean now.  Because I'm looping over the src type's fields, the fields won't be there, so it will never try to verify they are there.

mrvoorhe: Done

PR #236, opened 2018-01-23
Add functionality to terminate a test process upon output

sbomer:
The webapi test starts a process that does not terminate
automatically, but stops when it hits the timeout. We then expect the
output to contain some well-known string.

This was causing problems in ci because the ci machines were slow,
causing the test to frequently time out before printing the
string. Increasing the timeout to a large value caused the test to run
for too long.

This fix makes it possible to specify a termination condition. Now the
test process is killed when it prints the expected string, or hits a
timeout, currently set to 60 seconds.

As part of the change, the core process running logic has been moved
to a new class.

@erozenfeld, @JosephTremoulet PTAL

### DIFF
--- a/corebuild/integration/test/CommandRunner.cs
+++ b/corebuild/integration/test/CommandRunner.cs
@@ -1,0 +1,145 @@
+using System;
+using System.Diagnostics;
+using System.Text;
+using Xunit;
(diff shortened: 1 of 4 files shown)


erozenfeld:
LGTM

PR #245, opened 2018-01-29
OutputStep refactoring to make it more extensible and consistent

mrvoorhe:
We have our own UnityOutputStep that was not unified with OutputStep.  Ours is mostly a copy of OutputStep, plus some WinRT logic, plus some different symbol handling where we tried to fix symbol issues.

This PR does the following
* Refactors and exposes the hooks we need to maintain our winrt logic
* Cleans up an inconsistency with how SaveAssembly and CopyAssembly check if symbols should be included
* Moves the CloseSymbol() call from the default clause of the switch into a finally.   It's best to make sure you always close the symbols.  If a test were to hit an exception and CloseSymbol not be called, all subsequent tests would fail.  I didn't want to remove the `CloseSymbol` call in the Copy and Delete cases.     In some cases I think it's important that the symbols be closed prior to copying/deleting

### DIFF
--- a/linker/Linker.Steps/OutputStep.cs
+++ b/linker/Linker.Steps/OutputStep.cs
@@ -93,7 +93,7 @@ namespace Mono.Linker.Steps {
 				(module.Attributes & (ModuleAttributes) 0x04) != 0;
 		}
 
-		void WriteAssembly (AssemblyDefinition assembly, string directory)
+		protected virtual void WriteAssembly (AssemblyDefinition assembly, string directory)
 		{
 			foreach (var module in assembly.Modules) {
 				// Write back pure IL even for R2R assemblies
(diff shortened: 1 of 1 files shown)


linker/Linker.Steps/OutputStep.cs
```diff
@@ -116,31 +116,38 @@ void OutputAssembly (AssemblyDefinition assembly)
...
+					CopyAssembly (assembly, directory);
+					break;
+				case AssemblyAction.Delete:
+					CloseSymbols (assembly);
+					DeleteAssembly (assembly, directory);
+					break;
 				}
-				break;
-			default:
+			} finally {
```
marek-safar: I am not sure I see any benefit of this. what was the intention to use it for ?

mrvoorhe: Here's the scenario
* You start running all unit tests
* Tests are running fine
* Then one fails because of an exception thrown and Dispose doesn't get called on the SymbolReader
* All subsequent tests now fail because the temporary directory can't be cleaned up.

I didn't run into this with this code, but I did hit this scenario with changes coming to ResultChecker to verify the symbols.

marek-safar: Sure but the symbols are already closed, e.g. at https://github.com/mono/linker/pull/245/files#diff-78e7cd35fdf21ba1a7924c0def208816R134

mrvoorhe: Ok.  I'll undo this change.

mrvoorhe: Done

PR #246, opened 2018-01-29
Support for explicitly using csc to compile test case

mrvoorhe:
A requirement for symbol tests that are coming is access to a roslyn compiler so that we can generate portable and embedded pdbs.

This PR adds a general mechanism for tests to define which compiler they need to use when compiling a test case.

### DIFF
--- a/linker/Tests/TestCasesRunner/TestCaseCompiler.cs
+++ b/linker/Tests/TestCasesRunner/TestCaseCompiler.cs
@@ -1,13 +1,16 @@
 ﻿﻿using System;
 using System.CodeDom.Compiler;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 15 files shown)


linker/Tests/TestCasesRunner/TestCaseCompiler.cs
```diff
@@ -1,13 +1,16 @@
...
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text;
 using Mono.Linker.Tests.Extensions;
+using NUnit.Framework;
 
 namespace Mono.Linker.Tests.TestCasesRunner {
 	public class TestCaseCompiler {
+		static string _cachedWindowsCscPath = null;
```
marek-safar: Nit: redundant initialization

PR #255, opened 2018-02-07
.gitignore more corebuild build/test artifacts

JosephTremoulet:
Also add a comment header for the corebuild section, and normalize the
prefix used in it.


JosephTremoulet:
@sbomer @marek-safar PTAL

sbomer:
LGTM

PR #256, opened 2018-02-07
Allow custom loggers in Driver, use MSBuild log in ILLink

JosephTremoulet:
Currently, attempts to log messages when the linker is invoked by the ILLink MSBuild task get swallowed because MSBuild swallows tasks' stdout.  The linker already has an ILogger interface, so enable passing one to the Driver, and update the ILLink task to pass one to the Driver which will forward messages to any attached MSBuild loggers.

### DIFF
--- a/linker/Linker/Driver.cs
+++ b/linker/Linker/Driver.cs
@@ -45,13 +45,18 @@ namespace Mono.Linker {
 
 		public static int Main (string [] args)
 		{
+			return Execute (args);
+		}
+
+		public static int Execute (string[] args, ILogger customLogger = null)
+		{
 			if (args.Length == 0)
 				Usage ("No parameters specified");
 
 			try {
 
 				Driver driver = new Driver (args);
-				driver.Run ();
+				driver.Run (customLogger);
 
 			} catch (Exception e) {
 				Console.WriteLine ("Fatal error in {0}", _linker);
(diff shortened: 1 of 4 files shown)


JosephTremoulet:
@sven @marek-safar PTAL

sbomer:
LGTM

linker/Linker/ILogger.cs
```diff
@@ -12,4 +12,19 @@ public interface ILogger
 	{
 		void LogMessage (MessageImportance importance, string message, params object[] values);
 	}
+
+	public class LambdaLogger : ILogger
```
marek-safar: This does not make much sense to me. Why do have a double redirect? Please just implement ILogger in ILLinker the way you want

linker/Linker/Driver.cs
```diff
@@ -75,10 +80,14 @@ bool HaveMoreTokens ()
 			return _queue.Count > 0;
 		}
 
-		public void Run ()
+		public void Run (ILogger customLogger = null)
 		{
 			Pipeline p = GetStandardPipeline ();
 			using (LinkContext context = GetDefaultContext (p)) {
+				if (customLogger != null)
```
marek-safar: nit: follow the coding style

JosephTremoulet:
@marek-safar thanks for the review, updated per your suggestions (I think -- if "follow the coding style" meant something other than "remove the braces around the single-line `if` body", please LMK)

PR #257, opened 2018-02-07
Update managed app action defaults

JosephTremoulet:
Use a set of defaults chosen to provide the key benefit for
self-contained apps while still being conservative enough to have low
risk of introducing resolution failures.  App assemblies will all be
considered roots, framework assemblies will be removed if unreferenced
at assembly granularity, and at method granularity will have prejitting
suppressed for unmarked code.


JosephTremoulet:
@sbomer PTAL

sbomer:
These defaults look good to me for self-contained. Portable publish isn't working yet, is it?
I assume you're specifying the portable defaults in advance of the changes that will enable it. If so, LGTM, otherwise I might be missing something.

JosephTremoulet:
> Portable publish isn't working yet, is it?

It is not.

> I assume you're specifying the portable defaults in advance of the changes that will enable it.

That is correct.

PR #258, opened 2018-02-07
Make native trimming optional

JosephTremoulet:
Setting LinkerTrimNativeDeps=false will now bypass the FindNativeDeps
task and preserve all native dependencies.


JosephTremoulet:
@sbomer PTAL

PR #260, opened 2018-02-23
Tolerate assembly resolution errors

sbomer:
Previously, we swallowed AssemblyResolutionException in a few places
in various steps. We need an option to be generally more tolerant of
AssemblyResolutionExceptions, because there can be cases during
portable publish in which referenced assemblies are not present during
publish because they're part of the shared framework.

Instead of handling AssemblyResolutionException in every place in the
linker that tries to resolve methods/types/etc, this change adds an
option that enables the resolver to return null when an assembly is
not found. The null result is also cached to avoid multiple lookups.

Only a few places in the linker need to be modified to handle null
results, and the existing places where we used to catch
AssemblyResolutionException now only need to deal with a null result
instead. The downside of this approach is that these places cannot
distinguish between types not found in an existing assembly, and
assemblies that were not found.

The new behavior is grouped under the option --skip-unresolved, which
already handles unresolved types and methods in some cases. When this
option is enabled, a warning is logged, and no exception is thrown.

@erozenfeld, please review

### DIFF
--- a/linker/Linker/AssemblyResolver.cs
+++ b/linker/Linker/AssemblyResolver.cs
@@ -40,8 +40,9 @@ namespace Mono.Linker {
 #endif
 
 		readonly Dictionary<string, AssemblyDefinition> _assemblies;
-		readonly HashSet<string> _unresolvedAssemblies;
-		bool _ignoreUnresolved = false;
+		HashSet<string> _unresolvedAssemblies;
+		bool _ignoreUnresolved;
+		LinkContext _context;
 
 		public IDictionary<string, AssemblyDefinition> AssemblyCache {
 			get { return _assemblies; }
@@ -55,30 +56,32 @@ namespace Mono.Linker {
 		public AssemblyResolver (Dictionary<string, AssemblyDefinition> assembly_cache)
 		{
 			_assemblies = assembly_cache;
-                        _unresolvedAssemblies = new HashSet<string>();
 		}
 
-		public bool IgnoreUnresolved
-		{
+		public bool IgnoreUnresolved {
 			get { return _ignoreUnresolved; }
 			set { _ignoreUnresolved = value; }
+		}
+
+		public LinkContext Context {
+			get { return _context; }
+			set { _context = value; }
 		}
 
 		public override AssemblyDefinition Resolve (AssemblyNameReference name, ReaderParameters parameters)
 		{
 			AssemblyDefinition asm = null;
-			if (!_assemblies.TryGetValue (name.Name, out asm) && !_unresolvedAssemblies.Contains (name.Name)) {
-				try
-				{
+			if (!_assemblies.TryGetValue (name.Name, out asm) && (_unresolvedAssemblies == null || !_unresolvedAssemblies.Contains (name.Name))) {
+				try {
 					asm = base.Resolve (name, parameters);
 					_assemblies [name.Name] = asm;
-				}
-				catch (AssemblyResolutionException)
-				{
-					if (!_ignoreUnresolved) {
+				} catch (AssemblyResolutionException) {
+					if (!_ignoreUnresolved)
 						throw;
-					}
-					Console.WriteLine($"warning: unresolved assembly {name.Name}");
+
+					_context.LogMessage ($"warning: unresolved assembly {name.Name}");
+					if (_unresolvedAssemblies == null)
+						_unresolvedAssemblies = new HashSet<string> ();
 					_unresolvedAssemblies.Add (name.Name);
 				}
 			}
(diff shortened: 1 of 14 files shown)


linker/Linker/AssemblyResolver.cs
```diff
@@ -40,6 +40,8 @@ public class AssemblyResolver : BaseAssemblyResolver {
 #endif
 
 		readonly Dictionary<string, AssemblyDefinition> _assemblies;
+		readonly HashSet<string> _unresolvedAssemblies;
+		bool _ignoreUnresolved = false;
```
marek-safar: redundant initialization

linker/Linker/AssemblyResolver.cs
```diff
@@ -53,14 +55,32 @@ public AssemblyResolver ()
...
+				{
+					asm = base.Resolve (name, parameters);
+					_assemblies [name.Name] = asm;
+				}
+				catch (AssemblyResolutionException)
+				{
+					if (!_ignoreUnresolved) {
+						throw;
+					}
+					Console.WriteLine($"warning: unresolved assembly {name.Name}");
```
marek-safar: Use LinkContext::LogMessage instead

linker/Linker/AssemblyResolver.cs
```diff
@@ -53,14 +55,32 @@ public AssemblyResolver ()
 		public AssemblyResolver (Dictionary<string, AssemblyDefinition> assembly_cache)
 		{
 			_assemblies = assembly_cache;
+                        _unresolvedAssemblies = new HashSet<string>();
```
marek-safar: I'd prefer to delay initialization of this to the catch block

linker/Linker/AssemblyResolver.cs
```diff
@@ -53,14 +55,32 @@ public AssemblyResolver ()
...
 		public override AssemblyDefinition Resolve (AssemblyNameReference name, ReaderParameters parameters)
 		{
-			AssemblyDefinition asm;
-			if (!_assemblies.TryGetValue (name.Name, out asm)) {
-				asm = base.Resolve (name, parameters);
-				_assemblies [asm.Name.Name] = asm;
+			AssemblyDefinition asm = null;
+			if (!_assemblies.TryGetValue (name.Name, out asm) && !_unresolvedAssemblies.Contains (name.Name)) {
+				try
+				{
```
marek-safar: Follow existing coding style

linker/Linker/LinkContext.cs
```diff
@@ -104,7 +104,10 @@ public bool KeepMembersForDebuggerAttributes
 		public bool IgnoreUnresolved
 		{
 			get { return _ignoreUnresolved; }
-			set { _ignoreUnresolved = value; }
+			set {
+				_ignoreUnresolved = value;
+				_resolver.IgnoreUnresolved = value;
```
marek-safar: I don't like this, _resolver could be a shared/global instance and you are changing it under the hood

sbomer:
@marek-safar I addressed your feedback, thanks!

marek-safar:
build

sbomer:
Hold off on merging this for a minute - just discovered a bug. I need to check for null in AssemblyResolver.Dispose.

linker/Linker/AssemblyResolver.cs
```diff
@@ -80,6 +103,7 @@ protected override void Dispose (bool disposing)
 			}
 
 			_assemblies.Clear ();
+			_unresolvedAssemblies.Clear ();
```
marek-safar: _unresolvedAssemblies can be null, setting it to null should do it

linker/Linker/AssemblyResolver.cs
```diff
@@ -55,12 +58,32 @@ public AssemblyResolver (Dictionary<string, AssemblyDefinition> assembly_cache)
...
+			get { return _ignoreUnresolved; }
+			set { _ignoreUnresolved = value; }
+		}
+
+		public LinkContext Context {
+			get { return _context; }
+			set { _context = value; }
+		}
+
 		public override AssemblyDefinition Resolve (AssemblyNameReference name, ReaderParameters parameters)
```
marek-safar: I think it'd be better to pass LinkContext Context as the parameter

linker/Linker/AssemblyResolver.cs
```diff
@@ -80,6 +103,8 @@ protected override void Dispose (bool disposing)
 			}
 
 			_assemblies.Clear ();
+			if (_unresolvedAssemblies != null)
```
marek-safar: Just set it to null

linker/Linker/AssemblyResolver.cs
sbomer: The problem is that the Resolve method signature is part of IAssemblyResolver, defined in cecil. I want warnings to be logged when an assembly can't be resolved while trying to resolve TypeReferences to TypeDefinitions using cecil, for example. Do you have any suggestions?

marek-safar: hmm, I see

PR #262, opened 2018-02-26
Expose option to log internal exceptions

sbomer:
Also enable it for ILLink.Tasks, which will result in resolution
errors being logged as warnings during publish.

@erozenfeld PTAL.

### DIFF
--- a/linker/Linker/Driver.cs
+++ b/linker/Linker/Driver.cs
@@ -133,6 +133,11 @@ namespace Mono.Linker {
 							continue;
 						}
 
+						if (token == "--log-warnings") {
+							context.LogInternalExceptions = true;
+							continue;
+						}
+
 						switch (token [2]) {
 						case 'v':
 							Version ();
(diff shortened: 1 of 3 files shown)


linker/Linker/Driver.cs
```diff
@@ -346,6 +351,7 @@ static void Usage (string msg)
 			Console.WriteLine ("   --about             About the {0}", _linker);
 			Console.WriteLine ("   --version           Print the version number of the {0}", _linker);
 			Console.WriteLine ("   --skip-unresolved   Ignore unresolved types, methods, and assemblies (true or false)");
+			Console.WriteLine ("   --log-warnings      Log warnings for internal exceptions");
```
erozenfeld: If I'm reading the code correctly turning on this option will enable all LogMessage output, not just warnings. I think both the option name and the description are misleading. Maybe --log-messages or --output-log?

sbomer: They are a bit misleading. I changed them to LogMessages and --log-messages.

linker/Linker/Driver.cs
```diff
@@ -346,6 +350,7 @@ static void Usage (string msg)
 			Console.WriteLine ("   --about             About the {0}", _linker);
 			Console.WriteLine ("   --version           Print the version number of the {0}", _linker);
 			Console.WriteLine ("   --skip-unresolved   Ignore unresolved types, methods, and assemblies (true or false)");
+			Console.WriteLine ("   --log-messages      Log messages indicating progress and warnings");
```
marek-safar: I'd prefer `--verbose`

linker/Linker/Driver.cs
```diff
@@ -112,8 +112,12 @@ public void Run (ILogger customLogger = null)
 							continue;
 						}
 
-						if (token == "--dependencies-file")
-						{
+                                                if (token == "--log-messages") {
```
marek-safar: Indentation is quite off

linker/Linker/Driver.cs
sbomer: Changed.

linker/Linker/Driver.cs
sbomer: Fixed.

PR #263, opened 2018-02-26
Enable link during portable publish

sbomer:
This enables the linker to run during a portable publish.

During portable publish, the framework is shared. The set of assemblies to be excluded from publish is determined by a platform manifest that's part of the NETCore.App package. The portable publish set can include references to parts of the framework that are filtered by this mechanism, and aren't exposed as reference assemblies (for example, platform-specific assemblies that we don't want developers to compile against directly). With [tolerate resolution errors](https://github.com/mono/linker/pull/260), the linker can tolerate such missing references.

However, we still want to pass ReferencePath to the linker to prevent a large number of resolution warnings for parts of the framework that don't get published. This doesn't cover everything (as noted above), but limits the number of warnings. These reference assemblies are made part of the linker "platform", with a default action of skip during portable publish. This way the ref assemblies are used to prevent resolution errors, but they aren't part of the linker output.

Portable publish also depends on [find more native dependencies](https://github.com/mono/linker/pull/261).

In addition to the mentioned changes, this includes integration tests that exercise portable publish on helloworld, webapi, and musicstore.

@erozenfeld, @ericstj PTAL.
/cc @richlander @petermarcu

### DIFF
--- a/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
+++ b/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
@@ -281,6 +281,11 @@
       <_ManagedAssembliesToLink Include="@(IntermediateAssembly)" />
       <_ManagedAssembliesToLink Include="@(_ManagedResolvedAssembliesToPublish)" />
     </ItemGroup>
+    <ItemGroup Condition=" '$(SelfContained)' != 'true' ">
+      <!-- Portable publish: need to supply the linker with any needed reference assemblies as well -->
+      <_ReferencedLibraries Include="@(ReferencePath->'%(ResolvedPath)')" Exclude="@(_ManagedAssembliesToLink)" />
+      <_ManagedAssembliesToLink Include="@(_ReferencedLibraries)" />
+    </ItemGroup>
   </Target>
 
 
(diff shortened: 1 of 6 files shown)


corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -281,6 +281,27 @@
       <_ManagedAssembliesToLink Include="@(IntermediateAssembly)" />
       <_ManagedAssembliesToLink Include="@(_ManagedResolvedAssembliesToPublish)" />
     </ItemGroup>
+
+    <!-- Portable publish: need to supply the linker with any needed
+         reference assemblies as well.
+
+         We want to exclude assemblies that are going to be published
```
erozenfeld: Did you mean "assemblies that are not going to be published" ?

corebuild/integration/test/CommandRunner.cs
```diff
@@ -67,14 +67,15 @@ public int Run(out string commandOutput)
 				RedirectStandardOutput = true,
 				RedirectStandardError = true,
 			};
+                        outputHelper.WriteLine($"from caller working directory {Environment.CurrentDirectory}");
```
erozenfeld: Indentation is off. Also I would remove "from" and add ":" after directory for consistency with the message on line 78.

Petermarcu:
Cool, tell me when I can reference a new package to try it out.

In the current linker, only things in netcoreapp are linked. In this world, will linking happen on the things in my app?

erozenfeld:
By default, things in used assemblies in your app are untouched (unused assemblies are deleted). If you want things in your app to be linked, you can override this with /p:UsedApplicationAssemblyAction=Link.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
sbomer: I meant implementation assemblies that are going to be published, and also exist as a reference assembly. That sentence was unclear - I'll just remove it entirely.

corebuild/integration/test/CommandRunner.cs
sbomer: Those spaces always seem to sneak in... thanks! Updating the message per your suggestion as well.

PR #264, opened 2018-02-26
Add ClearInitLocalsAssemblies parameter for ClearInitLocalsStep.

erozenfeld:
Modified ClearInitLocalsStep to accept ClearInitLocalsAssemblies
custom parameter that can be specified via -m. The parameter
is a comma-separated list of assembly simple names that the the
step should be applied to. If the parameter is not specified ClearInitLocals
runs on all assemblies.

### DIFF
--- a/corebuild/integration/ILLink.CustomSteps/ClearInitLocals.cs
+++ b/corebuild/integration/ILLink.CustomSteps/ClearInitLocals.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 
 using Mono.Linker;
 using Mono.Linker.Steps;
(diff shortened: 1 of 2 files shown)


erozenfeld:
@sbomer PTAL

corebuild/integration/ILLink.CustomSteps/ClearInitLocals.cs
```diff
@@ -8,17 +9,39 @@ namespace ILLink.CustomSteps
...
+			}
+
+			bool changed = false;
+
 			foreach (ModuleDefinition module in assembly.Modules) {
 				foreach (TypeDefinition type in module.Types) {
 					foreach (MethodDefinition method in type.Methods) {
 						if (method.Body != null) {
 							method.Body.InitLocals = false;
+							changed = true;
```
sbomer: Maybe check whether InitLocals is already false before setting changed to true? Not sure how likely we are to encounter an assembly that already has this set to false for all methods, so it may not make much of a difference. It's your call.

erozenfeld: Done.

erozenfeld:
@sbomer I also pushed a change to this PR to expose ClearInitLocals and ClearInitLocalsAssemblies as ILLink.Tasks properties. PTAL.

sbomer:
LGTM

PR #265, opened 2018-02-26
Add --strip-security option to remove security attributes

mrvoorhe:
Add an option that will remove SecurityDeclarations from Types and Methods along with a few security attributes.

When this new option is used with `--strip-security`  all attributes from System.Security and System.Security.Permissions that are not used will be linked away.

### DIFF
--- a/linker/Linker.Steps/RemoveSecurityStep.cs
+++ b/linker/Linker.Steps/RemoveSecurityStep.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Linq;
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps {
+	public class RemoveSecurityStep : BaseStep {
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			if (Annotations.GetAction (assembly) == AssemblyAction.Link) {
+				ClearSecurityDeclarations (assembly);
+				RemoveCustomAttributesThatAreForSecurity (assembly);
+
+				foreach (var type in assembly.MainModule.Types)
+					ProcessType (type);
+			}
+		}
+
+		static void ProcessType (TypeDefinition type)
+		{
+			RemoveCustomAttributesThatAreForSecurity (type);
+			ClearSecurityDeclarations (type);
+			type.HasSecurity = false;
+
+			foreach (var field in type.Fields)
+				RemoveCustomAttributesThatAreForSecurity (field);
+
+			foreach (var method in type.Methods) {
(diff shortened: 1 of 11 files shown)


mrvoorhe:
I'll look into the test failures tomorrow.  Something strange is going on.  The output mscorlib.dll from `NoSecurityPlusOnlyKeepUsedRemovesAllSecurityAttributesFromCoreLibraries` on Windows has 3 types in it.  I need to take a closer look.

linker/Linker.Steps/RemoveSecurityStep.cs
```diff
@@ -0,0 +1,68 @@
...
+		/// <summary>
+		/// We have to remove some security attributes, otherwise pe verify will complain that a type has HasSecurity = false
+		/// </summary>
+		/// <param name="provider"></param>
+		static void RemoveCustomAttributesThatAreForSecurity (ICustomAttributeProvider provider)
+		{
+			if (!provider.HasCustomAttributes)
+				return;
+
+			var attrsToRemove = provider.CustomAttributes.Where (IsCustomAttributeForSecurity).ToArray ();
```
marek-safar: Redundant ToArray

linker/Linker/Driver.cs
```diff
@@ -340,6 +346,7 @@ static void Usage (string msg)
 			Console.WriteLine ("   --dump-dependencies Dump dependencies for the linker analyzer tool");
 			Console.WriteLine ("   --reduced-tracing   Reduces dependency output related to assemblies that will not be modified");
 			Console.WriteLine ("   --used-attrs-only   Attributes on types, methods, etc will be removed if the attribute type is not used");
+			Console.WriteLine ("   --no-security-attrs Attributes on assemblies, types, and methods related to security will be removed");
```
marek-safar: My preference would be `--strip-security` and you should note somewhere it works only when assembly is linked

linker/Linker.Steps/RemoveSecurityStep.cs
mrvoorhe: I don't believe it is.  The next line down removes these items from the very same collection.  Without capturing the result, you get into a situation where you are modifying the collection while iterating it

mrvoorhe:
@marek-safar changes and test fixes made.

linker/Linker.Steps/RemoveSecurityStep.cs
marek-safar: you are right

PR #267, opened 2018-02-27
Add ClearInitLocals and ClearInitLocalsAssemblies optional ILLink task parameters.

erozenfeld:
This change allows ClearInitLocals and ClearInitLocalsAssemblies to be specified
directly on ILLink task, which is useful when the task  used outside of publishing (e.g., in corefx).


erozenfeld:
@sbomer PTAL

sbomer:
LGTM

PR #268, opened 2018-02-28
Fix a regression introduced in f9c0d99, putting back the try/catch.

baulig:
`LinkContext.SafeReadSymbols` used to have a try / catch block around the actual symbol reading.

This got removed in [f9c0d99](https://github.com/mono/linker/commit/f9c0d99fd8630c508f2c801552282b8218ff9d3c) and is now breaking the monthly Xamarin.Android integration.

### DIFF
--- a/linker/Linker/LinkContext.cs
+++ b/linker/Linker/LinkContext.cs
@@ -233,16 +233,18 @@ namespace Mono.Linker {
 
 			if (_symbolReaderProvider == null)
 				throw new ArgumentNullException (nameof (_symbolReaderProvider));
-			
-			var symbolReader = _symbolReaderProvider.GetSymbolReader (
-				assembly.MainModule,
-				assembly.MainModule.FileName);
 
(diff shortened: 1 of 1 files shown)


mrvoorhe:
@baulig Out of curiosity, what is causing an exception in your case?  Is it a scenario we can add a test for?

Ironically, an hour or so ago I realized this same thing was causing a problem for our windows runtime tests.  We don't have our windows runtime support upstream, so I can't bring one of these tests upstream to get coverage.

baulig:
@mrvoorhe It fails during the 2018-02 integration build for Xamarin.Android.
https://jenkins.mono-project.com/job/xamarin-android-pr-builder/2587/console (look around `04:47:18.706` in the console log).
This is from https://github.com/xamarin/xamarin-android/pull/1289.

mrvoorhe:
@baulig  Thanks.  That looks similar to what I saw.

PR #269, opened 2018-03-01
Set crossgen defaults when linker is enabled

sbomer:
This introduces behavior that will crossgen any assemblies modified by the linker that were crossgen'd before linking.

The defaults for SPC have been updated so that it will never be rooted via -a. If it has an embedded root descriptor, this is the same as before - the root descriptor will specify what to root from SPC. If it doesn't have one, then its action will be AddBypassNGenUsed as it is part of the platform.

@erozenfeld PTAL.

### DIFF
--- a/corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
+++ b/corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
@@ -210,17 +210,17 @@
 
   </Target>
 
-  <UsingTask TaskName="ComputeReadyToRunAssemblies" AssemblyFile="$(LinkTaskDllPath)" />
+  <UsingTask TaskName="ComputeCrossgenedAssemblies" AssemblyFile="$(LinkTaskDllPath)" />
   <Target Name="_ComputeFilesToCrossGen"
           DependsOnTargets="_ComputeDefaultFilesToCrossGen">
 
-    <ComputeReadyToRunAssemblies Assemblies="@(FilesToCrossGen)">
-      <Output TaskParameter="ReadyToRunAssemblies" ItemName="_ReadyToRunFiles" />
-    </ComputeReadyToRunAssemblies>
+    <ComputeCrossgenedAssemblies Assemblies="@(FilesToCrossGen)">
+      <Output TaskParameter="CrossgenedAssemblies" ItemName="_CrossgenedFiles" />
+    </ComputeCrossgenedAssemblies>
 
     <!-- Don't try to run crossgen on assemblies that are already R2R. -->
     <ItemGroup>
-      <FilesToCrossGen Remove="@(_ReadyToRunFiles)" />
+      <FilesToCrossGen Remove="@(_CrossgenedFiles)" />
     </ItemGroup>
 
     <Message Text="files to crossgen: @(FilesToCrossGen)" />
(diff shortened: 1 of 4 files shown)


erozenfeld:
If SPC has no embedded root descriptor and we treat it like a normal framework assembly with AddBypassNGen action we may skip crossgen-ing parts of SPC that the native runtime depends on, which may affect performance. I think we either should root everything in SPC or set the action to Copy.

sbomer:
As discussed, I made the default to root SPC and keep the action AddBypassNGenUsed when there's no embedded descriptor. PTAL.

erozenfeld:
SPC is crossgen-ed but not R2R so ComputeReadyToRunAssemblies task will not include SPC in the list of crossgen-ed assemblies. We need to make sure SPC is re-crossgened after linking if it's been modified.

erozenfeld:
Ideally we should remember which assemblies were R2R crossgen-ed and which assemblies were fragile crossgen-ed and do the same after linking. In practice today it doesn't matter since the only thing that's fragile crossgen-ed is SPC and it can't be R2R crossgen-ed.

sbomer:
@erozenfeld PTA(nother)L. I renamed things to clarify that SPC will be detected by the logic in ComputeReadyToRunAssemblies (now ComputeCrossgenedAssemblies). I made a note about the fragile vs R2R distinction, but I'm not going to add support for that in this PR.

corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
```diff
@@ -208,21 +207,19 @@
...
           DependsOnTargets="_ComputeDefaultFilesToCrossGen">
 
-    <ComputeReadyToRunAssemblies Assemblies="@(FilesToCrossGen)">
-      <Output TaskParameter="ReadyToRunAssemblies" ItemName="_ReadyToRunFiles" />
-    </ComputeReadyToRunAssemblies>
+    <ComputeCrossgenedAssemblies Assemblies="@(FilesToCrossGen)">
+      <Output TaskParameter="CrossgenedAssemblies" ItemName="_CrossgenedFiles" />
+    </ComputeCrossgenedAssemblies>
 
     <!-- Don't try to run crossgen on assemblies that are already R2R. -->
```
erozenfeld: Please remove R2R from the comment.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -382,14 +411,17 @@
...
     </ItemGroup>
 
     <ItemGroup Condition=" '$(RootAllApplicationAssemblies)' == 'false' ">
       <LinkerRootAssemblies Include="@(IntermediateAssembly)" />
-      <LinkerRootAssemblies Include="System.Private.CoreLib" Condition=" '$(_SPCHasEmbeddedRootDescriptor)' == 'false' "/>
     </ItemGroup>
+
+    <ItemGroup Condition=" '$(_SPCHasEmbeddedRootDescriptor)' != 'true' ">
+      <LinkerRootAssemblies Include="@(_ManagedResolvedAssembliesToPublish->'%(Filename)')" Condition=" '%(Filename)' == 'System.Private.CoreLib' " />
+    </ItemGroup>
```
erozenfeld: This is correct but looks odd. Why not just include 'System.Private.CoreLib' like we did before, without using the second Condition?

sbomer: This way, if something changes and SPC isn't in the set of assemblies to publish, it won't be included as a root. This also does the right thing for portable publish.

corebuild/integration/ILLink.Tasks/ILLink.CrossGen.targets
sbomer: Done, thanks!

erozenfeld:
LGTM

PR #270, opened 2018-03-01
Clean up ready-to-run detection

sbomer:
- Use cecil's ILLibrary attribute (https://github.com/jbevain/cecil/pull/475)
- Factor common logic into a new function

@erozenfeld PTAL

### DIFF
--- a/linker/Linker/AssemblyUtilities.cs
+++ b/linker/Linker/AssemblyUtilities.cs
@@ -5,7 +5,7 @@ namespace Mono.Linker {
 
 	public static class AssemblyUtilities {
 
-		public static bool IsReadyToRun (this ModuleDefinition module)
+		public static bool IsCrossgened (this ModuleDefinition module)
 		{
 			return (module.Attributes & ModuleAttributes.ILOnly) == 0 &&
 				(module.Attributes & ModuleAttributes.ILLibrary) != 0;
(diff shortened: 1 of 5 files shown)


linker/Linker/AssemblyUtilities.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using Mono.Cecil;
+
+namespace Mono.Linker {
+
+	public class AssemblyUtilities {
```
marek-safar: Doesn't need to be public but should be static

linker/Linker/AssemblyUtilities.cs
```diff
@@ -0,0 +1,16 @@
...
+
+		public static bool IsReadyToRun (ModuleDefinition module)
```
marek-safar: Make this an extension method

sbomer: Done, thanks!

linker/Linker/AssemblyUtilities.cs
sbomer: Made it static, but need to keep it public so I can use it from ILLink.Tasks.

sbomer:
Eugene pointed out that SPC, while detected by this logic, isn't considered R2R. So I need to update the name of this check to something like IsCrossgend. Please hold off on merging this until I get a chance to do so.

erozenfeld:
Yes, this logic detects all crossgen-ed assemblies not just R2R.

sbomer:
I've renamed the check. I'll rename the tasks and msbuild items that depend on the check in https://github.com/mono/linker/pull/269, to keep these PRs somewhat independent.

erozenfeld:
LGTM

PR #273, opened 2018-03-05
Don't attempt to link duplicate managed assemblies

sbomer:
Portable apps can have platform-specific managed assemblies, but the
linker isn't able to handle multiple assemblies with the same
identity. We detect such duplicates, and prevent them from being
linked at all. This will prevent analysis of some code paths, but
seems to be fairly rare.

I've observed the following cases:
- System.Data.SqlClient
- System.Text.Encoding.CodePages

Both were pulled in by the ASP.NET metapackage. Some of these
duplicate assemblies may have native dependencies, so the native
trimming logic does scan them.

@erozenfeld, @ericstj please review.

### DIFF
--- a/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
+++ b/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
@@ -502,7 +502,7 @@
           DependsOnTargets="_ComputeManagedResolvedAssembliesToPublish;_ComputeLinkedAssemblies;_FindNativeDeps;_HandlePublishFileConflicts"
           BeforeTargets="GeneratePublishDependencyFile"
           Condition=" '$(LinkDuringPublish)' == 'true' ">
-    <ComputeRemovedAssemblies InputAssemblies="@(_ManagedResolvedAssembliesToPublish)"
+    <ComputeRemovedAssemblies InputAssemblies="@(_ManagedAssembliesToLink)"
                               KeptAssemblies="@(_LinkedResolvedAssemblies)">
       <Output TaskParameter="RemovedAssemblies" ItemName="_RemovedManagedAssemblies" />
     </ComputeRemovedAssemblies>
--- a/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
+++ b/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
@@ -319,11 +319,11 @@
       <_ManagedAssembliesToLink Include="@(_ManagedResolvedAssembliesToPublish)" />
     </ItemGroup>
 
-    <!-- Remove all implementation dlls that have duplicates. This can
-         happen if a portable app has platform-specific managed
-         assemblies. The linker isn't currently able to analyze
-         multiple assemblies with the same identity, so we skip
-         these. -->
+    <!-- Remove duplicate implementation dlls. This can happen if a
+         portable app has platform-specific managed assemblies. The
+         linker isn't currently able to analyze multiple assemblies
+         with the same identity, so we skip these, keeping only one
+         representative assembly from each set of duplicates. -->
     <FindDuplicatesByMetadata Items="@(_ManagedAssembliesToLink)"
                               MetadataName="Filename">
       <Output TaskParameter="DuplicateItems" ItemName="_DuplicateManagedAssemblies" />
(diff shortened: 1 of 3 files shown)


ericstj:
Doesn't this mean you'll miss exclusive dependencies from these?  EG: Suppose you have a duplicate copy of SqlClient and it uses Registry, but's the only thing that uses Registry in the entire app.  By excluding it won't you end up trimming Registry?

ericstj:
FWIW I hit the same problem in the assembly level trimmer.  To address I made sure to handle *both* implementations: https://github.com/dotnet/standard/commit/c77c9b666852b2fe0334ac7c6a339578f9b5f903.

sbomer:
@ericstj it does, when the exclusive dependency isn't part of the platform, and I'll make sure to point out this limitation in documentation. It would take significant changes to the linker for it to handle multiple assemblies correctly, since it makes assumptions about assemblies (caching them by name, for example) at a pretty deep level. We might be able to address this properly in wave 2.

For now, we've decided do this simple thing. It happens to work for the apps I've tried. If we encounter more problems with this, we can address them one-by-one by hardcoding dependencies in wave 1.
/cc @jeffschwMSFT

ericstj:
I see.  Yeah I can see that it would be more difficult in the linker.  I believe when I looked at this for the trimmer, simply using `Microsoft.AspNetCore.All/2.0.0` was enough to bring in a case where it made a difference.

erozenfeld:
I suggest we analyze one of the duplicate assemblies. It's safer than not analyzing any of them.

jeffschwMSFT:
When @sbomer and I talked about this offline, we discussed what the worst potential down side was.  At the time we identified that if any of these duplicate managed assemblies had dependencies we could incorrectly trim them.  (Leaving too much in the publish is an acceptable downside of stopping at wave 1)  Do we know if these assemblies have that potential?  If they do, does @erozenfeld 's proposal help reduce the risk for that?

sbomer:
@jeffschwMSFT I believe you mean leaving too little in the publish output.
@erozenfeld's suggestion would reduce the risk of that happening for the one platform that we happen to pick. I don't think it makes a difference for the specific cases I've encountered.

sbomer:
@erozenfeld I've updated the logic to analyze one of the duplicates. PTAL.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -284,13 +285,30 @@
...
     <ItemGroup>
       <_ManagedAssembliesToLink Include="@(IntermediateAssembly)" />
       <_ManagedAssembliesToLink Include="@(_ManagedResolvedAssembliesToPublish)" />
     </ItemGroup>
 
+    <!-- Remove all implementation dlls that have duplicates. This can
+         happen if a portable app has platform-specific managed
+         assemblies. The linker isn't currently able to analyze
+         multiple assemblies with the same identity, so we skip
+         these. -->
```
erozenfeld: Please update the comment since we no longer skip all duplicates.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
```diff
@@ -383,6 +401,7 @@
       <_LinkerRootAssemblies Include="@(_ManagedResolvedAssembliesToPublish->'%(Filename)')" />
       <_LinkerRootAssemblies Remove="@(PlatformLibraries->'%(Filename)')" />
       <_LinkerRootAssemblies Remove="System.Private.CoreLib" Condition=" '$(_SPCHasEmbeddedRootDescriptor)' == 'true' "/>
+      <_LinkerRootAssemblies Remove="@(_DuplicateManagedAssembliesToSkip->'%(Filename)')" />
```
erozenfeld: Will this keep the duplicate representative assemblies that we will link as roots?

sbomer: It will if they're not part of the platform.

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
sbomer: Updated, thanks!

corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
sbomer: You're right - this was actually removing the representative assemblies completely. Fixed in latest commit.

sbomer:
@marek-safar I rebased again. Please merge this when you get a chance. Thanks!

erozenfeld:
build Linux - Mono

erozenfeld:
build

PR #274, opened 2018-03-05
Root Internal methods when resolving from assembly.

erozenfeld:
Internal methods should be rooted regardless of their visibility.


erozenfeld:
@marek-safar PTAL

mrvoorhe:
@erozenfeld Can you elaborate on why this is necessary?

erozenfeld:
It turns out the reason I thought it was necessary is not valid so I'm closing this PR.

PR #275, opened 2018-03-05
Remove BlackList step precondition.

erozenfeld:
Remove BlackList step precondition so that embedded
root lists mechanism can be used in more scenarios.


erozenfeld:
@marek-safar PTAL

erozenfeld:
build

erozenfeld:
@mrvoorhe I removed the test that verified the existing behavior. PTAL

erozenfeld:
Some motivation for the change: CoreAction is not really meaningful here since the action for each core assembly could've been overwritten. One possibility is checking the action for each assembly whose linker descriptor is found and processing only those with Link action. I'm not sure that's worth doing but let me know if you think that's preferable.

mrvoorhe:
@erozenfeld Looking back on my comment in `EmbeddedLinkXmlFileIsNotProcessedWhenCoreLinkActionIsSkip` I think my wording implies a little to strongly that the behavior the test is asserting is odd.  Thinking about what it means to "Skip" an assembly now, I think that should mean "don't do anything with this assembly".  If we accept that is what "Skip" means, then I think the test as it exists today is asserting the desired behavior.

Thinking about the impact of this change on other cases, I noticed there's no test for when CoreAction is "Copy".  I wrote a test for that case locally.  In the case of "Copy" I would expect the resource to exist in the output assembly.  The test ended up being a copy of `EmbeddedLinkXmlFileIsNotProcessedWhenCoreLinkActionIsSkip` with the core action changed.  On master it passes, with your changes it fails because the resource is removed from the output assembly.

My feeling is that there should be tests asserting the desired behavior.

As for what the desired behavior is, I think we need to decide what AssemblyAction.Copy and what AssemblyAction.Skip means.  Thinking out loud, here is my attempt at pinning down a meaning

* Skip - Completely ignore the assembly
* Copy - Copy the assembly, as is, to the output directory.

If we agreed on those definitions then I think the sensible behavior with regards to embedded link xml resources would be

* Skip - Don't process embedded link xml files.  You are ignoring the assembly.

* Copy - The copied assembly should retain it's embedded link xml resource.  Whether or not embedded link xml resources are processed, I would lean toward no.

@marek-safar what do you think?

erozenfeld:
@mrvoorhe There are two problems with the CoreAction precondition:
1. CoreAction should have no effect on what we do with non-core assemblies.
2. Even for core assemblies CoreAction can be overwritten with -p on individual assemblies.

I pushed an update that uses individual assembly actions to decide on whether embedded resource should be processed. PTAL.

I tried to add a test where a dependent assembly has an embedded resource and action Skip (since the main assembly will have action Link set by ResolveFromAssembly) but couldn't figure out a way to embed a resource into a dependent assembly. Something like
```
[SetupCompileBefore ("LibraryWithEmbeddedResource.dll", new [] { "Dependencies/LibraryWithEmbeddedResource.cs" }, additionalArguments: "/resource:./Dependencies/LibraryWithEmbeddedResource.xml", compilerToUse: "csc")]
```
doesn't work because the /resource relative path isn't processed correctly.

mrvoorhe:
I like the new approach based on the assembly action.  

I will open a PR with the testing improvement you need tomorrow.

If you are not in a hurry and can wait for my PR, I'd like to see the tests included with this PR so that there are no loose ends.

marek-safar:
@erozenfeld merging but please add the test case

PR #276, opened 2018-03-05
Updates for CreateRuntimeRootDescriptorFile msbuild task.

erozenfeld:
1. Include the contents of coreclr\src\mscorlib\ILLinkTrim.xml
(that will be added shortly to coreclr) into the descriptor computed
by CreateRuntimeRootDescriptorFile.

2. Preserve constructors in exceptions declared in rexcep.h.


erozenfeld:
@sbomer PTAL

PR #277, opened 2018-03-06
Make removal of processed resources optional.

erozenfeld:
We have a scenario when the linker is invoked on an assembly
more than once so we need to preserve the embedded xml resources.

### DIFF
--- a/linker/Linker/Driver.cs
+++ b/linker/Linker/Driver.cs
@@ -143,6 +143,11 @@ namespace Mono.Linker {
 							continue;
 						}
 
+						if (token == "--strip-resources") {
+							context.StripResources = bool.Parse (GetParam ());
+							continue;
+						}
+
 						switch (token [2]) {
 						case 'v':
 							Version ();
@@ -316,6 +321,7 @@ namespace Mono.Linker {
 			return assemblies;
 		}
 
+
 		AssemblyAction ParseAssemblyAction (string s)
 		{
 			var assemblyAction = (AssemblyAction)Enum.Parse(typeof(AssemblyAction), s, true);
@@ -362,6 +368,7 @@ namespace Mono.Linker {
 			Console.WriteLine ("   --reduced-tracing   Reduces dependency output related to assemblies that will not be modified");
 			Console.WriteLine ("   --used-attrs-only   Attributes on types, methods, etc will be removed if the attribute type is not used");
 			Console.WriteLine ("   --strip-security    In linked assemblies, attributes on assemblies, types, and methods related to security will be removed");
+			Console.WriteLine ("   --strip-resources   Remove link xml resources that were processed (true or false), default to true");
 			Console.WriteLine ("   -out                Specify the output directory, default to `output'");
 			Console.WriteLine ("   -c                  Action on the core assemblies, skip, copy, copyused, addbypassngen, addbypassngenused or link, default to skip");
 			Console.WriteLine ("   -u                  Action on the user assemblies, skip, copy, copyused, addbypassngen, addbypassngenused or link, default to link");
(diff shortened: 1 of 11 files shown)


erozenfeld:
@marek-safar @mrvoorhe PTAL

linker/Linker/Driver.cs
```diff
@@ -362,6 +368,7 @@ static void Usage (string msg)
 			Console.WriteLine ("   --reduced-tracing   Reduces dependency output related to assemblies that will not be modified");
 			Console.WriteLine ("   --used-attrs-only   Attributes on types, methods, etc will be removed if the attribute type is not used");
 			Console.WriteLine ("   --strip-security    In linked assemblies, attributes on assemblies, types, and methods related to security will be removed");
+			Console.WriteLine ("   --remove-resources  Remove link xml resources that were processed (true or false), default to true");
```
marek-safar: It'd make sense to call it `--strip-resources` to keep in line with previous options and to make the true default more meaningful

linker/Linker/LinkContext.cs
```diff
@@ -111,6 +111,8 @@ public bool IgnoreUnresolved
 
 		public bool KeepUsedAttributeTypesOnly { get; set; }
 
+		public bool RemoveResources { get; set; }
```
marek-safar: This needs to be set to true by default or the logic should be inverted

marek-safar:
Please extend existing tests to cover this case

mrvoorhe:
@erozenfeld Can you add a test verifying the new option works?

erozenfeld:
@mrvoorhe Is there a way to run the tests on Windows? If so, can you point me to the instructions?

mrvoorhe:
@erozenfeld Yes.  Any test runner that supports nunit will work.

I use VS + Resharper or Project Rider.  

Other alternatives would be

* http://nunit.org/docs/2.4/nunit-gui.html

* There's an NUnit test adapter for VS that should work

* As a last resort, you can use the nunit console runner that is downloaded from nuget.

Ex:

```packages\NUnit.ConsoleRunner.3.6.1\tools\nunit3-console.exe --test=Mono.Linker.Tests.TestCases.All.ResourcesTests linker\Tests\bin\Debug\Mono.Linker.Tests.dll --labels=All```

**Comment by erozenfeld** [comment]
@marek-safar @mrvoorhe I addressed all feedback. PTAL

File: linker/Tests/Mono.Linker.Tests.Cases.Expectations/Metadata/StripResourcesAttribute.cs [hunk 3]
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Metadata {
+	public sealed class StripResourcesAttribute : BaseMetadataAttribute {
```
mrvoorhe: It's fine that you added this, don't worry about removing it.  But FYI for the future, for simple command line flags that communicate their meaning you can use the general purpose `SetupLinkerArgumentAttribute` 

Ex: `[SetupLinkerArgument ("--strip-security", "true")]`

What `SetupLinkerArgumentAttribute` shouldn't be used for is the command line flags that communicate nothing to someone looking at a test.  For example, don't use it for:  `[SetupLinkerArgument ("-r", "library.dll")].

mrvoorhe:
As I stated in a comment, I'm not a fan of the command line flag name.  I also didn't have a great alternative suggestion so if @marek-safar is OK with it, I'm fine with it.

linker/Tests/Mono.Linker.Tests.Cases.Expectations/Metadata/StripResourcesAttribute.cs
marek-safar: @mrvoorhe I'm merging it as we are not consistent with other command line options either though it'd be nice to clean this up

PR #292, opened 2018-04-05
Fail RemovedTypeInAssemblyAttribute if the type does not exist in the original assembly

mderoy:


linker/Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -215,6 +216,14 @@ void VerifyLinkingOfOtherAssemblies (AssemblyDefinition original)
 			}
 		}
 
+		void VerifyRemovedTypeInAssembly(CustomAttribute inAssemblyAttribute, TypeDefinition linkedType)
```
mrvoorhe: Right now the name of this method doesn't make any sense.  `VerifyRemovedTypeInAssembly`, that's not what this method does.  This method tries to locate the given type in the original assembly and fails if it's not found.

Instead of the changes you have here,  what if you changed the following code in `GetOriginalTypeFromInAssemblyAttribute`

from

```
if (originalType == null)
	throw new InvalidOperationException ($"Unable to locate the original type `{expectedTypeName}`");
```

to

```
if (originalType == null)
	Assert.Fail($"Invalid test assertion.  Unable to locate the original type `{expectedTypeName}.`");
```

Would that work?

PR #303, opened 2018-04-25
Fix a bug where types marked in attributes would have their assembly deleted

mrvoorhe:
When an assembly is compiled with mcs and it has an attribute that makes the only use of a type via a typeof in a different assembly, no reference to that assembly will exist

This leads to the referenced assembly being deleted even though a type in the assembly was marked.

This issue does not happen with roslyn which adds a reference to the other assembly even though it is not strictly necessary.

Note : This PR depends on https://github.com/mono/linker/pull/302 so the history and changes will clean up a bit once that lands and I rebase this PR.

@marek-safar I wanted to open this PR before the other one lands so that I can get early feedback.   I think LoadReferenceStep is the best place for this logic, but I'm not entirely sure.


mrvoorhe:
@marek-safar linker changes removed

Tests remain, however, they are ignored since they won't pass until your change makes it into the mcs on the CI

I added csc versions of the tests.  If we are adding tests to ensure that mcs does something that the linker depends on, then I thought it made sense that we ensure that roslyn does the samething that the linker depends on.

mrvoorhe:
build

mrvoorhe:
> Please remove monobuild/TestResults.xml file
Thanks

> I'm also not convinced whether we need both csc and mcs versions.
We are moving toward using csc everywhere.  Which means in the future, the scenario I really don't want to have break is the csc scenario.  From that perspective, it seemed odd not to include csc versions.

In an experimental branch of our UnityLinker test suite I added a [ParameterizeTestCaseForCompiler("csc")] attribute to the test framework.  I didn't end up needing the mechanism for what I was doing.  If I brought that to upstream would that make having tests for the csc scenario more palatable?

> Adding something like [Requires (mcs, new Version (5.15))] would be better
I agree that would be nicer.  Can I handle this in a follow on PR?

mrvoorhe:
I removed the checked in TestResults.xml file.  The tests pass locally for me, however, the build is still failing.

This seems to be what is causing the failure, but I don't know what it means...?

```
15:07:20 [xUnit] [INFO] - [NUnit-2 (default)] - 1 test report file(s) were found with the pattern 'monobuild/TestResult*.xml' relative to '/mnt/jenkins/workspace/test-linker-pull-request' for the testing framework 'NUnit-2 (default)'.
15:07:23 [xUnit] [INFO] - Check 'Failed Tests' threshold.
15:07:23 [xUnit] [INFO] - Check 'Skipped Tests' threshold.
15:07:23 [xUnit] [INFO] - The total number of tests for this category exceeds the specified 'failure' threshold value.
15:07:23 [xUnit] [INFO] - Setting the build status to FAILURE
15:07:23 [xUnit] [INFO] - Stopping recording.
15:07:23 Build step 'Publish xUnit test result report' changed build result to FAILURE
```

mrvoorhe:
It seems like the build is failing because of the ignored tests.

In throw away PR I took these changes and used Assert.Pass instead of Assert.Ignore and the build passed. 
https://github.com/mono/linker/pull/317

Based on the error message I can only guess that something is setting a limit on the number of ignored tests allowed?

marek-safar:
build

marek-safar:
build

PR #304, opened 2018-04-30
Fix PDB handle leak in the linker.

TautvydasZilys:
This change fixes the handle leak if assembly.MainModule.ReadSymbols throws an exception.

### DIFF
--- a/linker/Linker/LinkContext.cs
+++ b/linker/Linker/LinkContext.cs
@@ -245,8 +245,15 @@ namespace Mono.Linker {
 				if (symbolReader == null)
 					return;
 
+				try {
+					assembly.MainModule.ReadSymbols (symbolReader);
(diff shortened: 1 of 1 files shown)


linker/Linker/LinkContext.cs
```diff
@@ -245,8 +245,15 @@ public virtual void SafeReadSymbols (AssemblyDefinition assembly)
 				if (symbolReader == null)
 					return;
 
+				try {
+					assembly.MainModule.ReadSymbols (symbolReader);
+				} catch {
+					symbolReader.Dispose ();
+					throw;
```
marek-safar: This is inside catch {} so just return instead of throw

TautvydasZilys: Fixed.

PR #309, opened 2018-05-04
Fix LockFileCache when SDK and ILLink  task are using different NuGet dlls

### LINKED ISSUE #308: ILLink.Task fails when used in latest SDK
/cc @sbomer 
Using the latest dotnet SDK, add the following
```
    <PackageReference Include="ILLink.Tasks" Version="0.1.5-preview-1461378" />
```
Publish the project.

It will fail with
```
2>The "GetRuntimeLibraries" task failed unexpectedly.
2>System.InvalidCastException: [A]NuGet.ProjectModel.LockFile cannot be cast to [B]NuGet.ProjectModel.LockFile. Type A originates from 'NuGet.ProjectModel, Version=4.7.0.5, Culture=neutral, PublicKeyToken=31bf3856ad364e35' in the context 'LoadFrom' at location 'C:\Program Files\dotnet\sdk\2.1.300-rc1-008673\Sdks\Microsoft.NET.Sdk\tools\net46\NuGet.ProjectModel.dll'. Type B originates from 'NuGet.ProjectModel, Version=4.3.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' in the context 'LoadFrom' at location 'C:\Users\ericstj\.nuget\packages\illink.tasks\0.1.5-preview-1461378\tools\net46\NuGet.ProjectModel.dll'.
2>   at Microsoft.NET.Build.Tasks.LockFileCache.GetLockFile(String path)
2>   at ILLink.Tasks.GetRuntimeLibraries.Execute()
2>   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
2>   at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext()
```

This is  the same bug I fixed/worked-around here: https://github.com/dotnet/standard/pull/722

@nguerrera do you have an idea about a public contract we can have around this to avoid these issues in the future?  Either we get everyone to agree on where to load NuGet from or we version the cache objects (like I did in my PR).  The latter is less good since it means we don't benefit from the cache when we get out of sync.

ericstj:
Avoid this by using a versioned prefix for the cache key. Of course we will no longer share the read if we happen to be on the same version, but this is safer.

Fixes #308

### DIFF
--- a/corebuild/integration/ILLink.Tasks/Microsoft.NET.Build.Tasks/LockFileCache.cs
+++ b/corebuild/integration/ILLink.Tasks/Microsoft.NET.Build.Tasks/LockFileCache.cs
@@ -12,6 +12,7 @@ namespace Microsoft.NET.Build.Tasks
 	internal class LockFileCache
 	{
 		private IBuildEngine4 _buildEngine;
+		private static string s_taskObjectPrefix = null;
 
 		public LockFileCache(IBuildEngine4 buildEngine)
 		{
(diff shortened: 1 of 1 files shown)


sbomer:
LGTM, thanks for fixing this!

PR #310, opened 2018-05-05
Ensure linker dependencies are properly written and closed.

erozenfeld:
Add a finally clause that ensures that dependencies file is
written and resources are disposed if there is an unhandled exception.


erozenfeld:
@radekdoulik PTAL

mrvoorhe:
Nice!

erozenfeld:
@marek-safar I made the requested change.

PR #313, opened 2018-05-07
Fix TargetTypeName processing not handling types in other assemblies

mrvoorhe:
The debugger display handling logic was not accounting for cases such as

`[assembly: DebuggerDisplay("{MyProperty}", TargetTypeName = "Library.Class1, Library")]`

This PR reuses recently added type assembly qualified name parsing code so that this case is handled


linker/Linker/Utils.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+
+namespace Mono.Linker {
+	public static class Utils {
```
marek-safar: Call it TypeNameParser (I hate Utils name)

linker/Linker/Utils.cs
```diff
@@ -0,0 +1,34 @@
...
+			}
+
+			//Filter the assembly qualified name down to the basic type by removing pointer, reference, and array markers on the type
+			//We must also convert nested types from + to / to match cecil's formatting
+			value = value
+				.Replace ('+', '/')
+				.Replace ("*", string.Empty)
+				.Replace ("&", string.Empty);
+
+			while (value.Contains ("[")) {
```
marek-safar: Use IndexOf with char value

linker/Linker/Utils.cs
mrvoorhe: Fixed and noted :smile:

linker/Linker/Utils.cs
mrvoorhe: Fixed

Issue #286, opened 2018-03-17
Fatal error in IL Linker with latest SDK


JustArchi:
```
ILLINK : warning : unresolved assembly System.Runtime [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Threading.Tasks [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Collections [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Runtime.Extensions [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Diagnostics.Debug [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Threading [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Xml.ReaderWriter [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Linq.Expressions [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Collections.Concurrent [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Threading.Timer [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.IO.FileSystem.Watcher [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.IO.Compression [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
ILLINK : warning : unresolved assembly System.Collections.Specialized [/home/archi/ArchiSteamFarm/ArchiSteamFarm/ArchiSteamFarm.csproj]
[...]

PR #324, opened 2018-06-06
Don't use ResolvedPath metadata on ReferencePath

sbomer:
This fixes https://github.com/mono/linker/issues/286, introduced by https://github.com/dotnet/sdk/pull/1857, which removes some metadata from this ItemGroup.

I tested this on a simple project. @nguerrera, can you confirm whether the ReferencePath ItemGroup should contain the correct path to any resolved assemblies in general? If not, which item metadata could we use instead?

### DIFF
--- a/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
+++ b/corebuild/integration/ILLink.Tasks/ILLink.Tasks.targets
@@ -343,14 +343,14 @@
          implementation assembly, and prevent inclusion of the ref
          assembly by filtering on filename instead of the full
          path. -->
-    <FilterByMetadata Items="@(ReferencePath->'%(ResolvedPath)')"
+    <FilterByMetadata Items="@(ReferencePath)"
                       MetadataName="Filename"
                       MetadataValues="@(_ManagedAssembliesToLink->'%(Filename)')"
                       Condition=" '$(SelfContained)' != 'true' ">
       <Output TaskParameter="FilteredItems" ItemName="_ReferencedLibrariesToExclude" />
     </FilterByMetadata>
     <ItemGroup Condition=" '$(SelfContained)' != 'true' ">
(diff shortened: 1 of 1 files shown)


nguerrera:
> @nguerrera, can you confirm whether the ReferencePath ItemGroup should contain the correct path to any resolved assemblies in general

Depends what you mean by resolved assemblies. Historically, ReferencePath is all of the assemblies that will be passed to the (C#/VB/F#) compiler. Is that what you want in this context? I ask because a linker would need to link runtime assemblies in general, which may not be in ReferencePath. 

As of recently, the compiler actually gets @(ReferencePathWithRefAssemblies), not @(ReferencePath). So for the case of CompileUsingReferencesAssemblies=true (https://github.com/Microsoft/msbuild/pull/2039), @(ReferencePath) will have implementation assemblies of user projects and @(ReferencePathWithRefAssemblies) will have the reference assemblies for user projects. However, for nuget packages with lib/ and ref/ split, @(ReferencePath) will have reference assemblies and only @(ReferenceCopyLocalPaths) will have implementation assemblies. 

(Aside: It strikes me just now that perhaps we should think about changing nuget case to use the same mechanism (or some generalization thereof) as CompileUsingReferenceAssemblies. Nuget case came before  CompileUsingReferenceAssemblies, which explains the divergence. I haven't thought it through fully, there might be an avenue for fixing issues like missing binding redirect generation for assemblies excluded from compilation, etc.)

cc @ericstj @rainersigwald

nguerrera:
> and only @(ReferenceCopyLocalPaths) will have implementation assemblies

Correcting myself, this isn't true for CopyLocalLockFileAssemblies=false (.NET Core) default.

Anyhow, I'm guessing that you must already have had to deal with all of this to link runtime assemblies from NuGet, and ReferencePath is indeed what you need in this context. But I didn't want to just answer yes because the ref/lib split case is very often missed.

sbomer:
Closing this since this is fixed with the 3.0 SDK integration.

PR #331, opened 2018-07-10
Fix nunit exceptions thrown during compile

mrvoorhe:
The recent change to compile test cases in parallel led to nunit exceptions thrown from calls such as Assert.Ignore() not propagating back out.  Instead an AggregateException would be thrown and nunit would, in the case of ignore, not fail the test instead of ignore it.

The only tests that do an Assert.Ignore during test case compilation are tests that need to use mcs when running on windows.  We don't run the tests on windows during CI so that is how this slipped by.

### DIFF
--- a/linker/Tests/TestCasesRunner/TestRunner.cs
+++ b/linker/Tests/TestCasesRunner/TestRunner.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Threading.Tasks;
 using Mono.Cecil;
 using Mono.Linker.Tests.TestCases;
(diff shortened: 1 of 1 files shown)


linker/Tests/TestCasesRunner/TestRunner.cs
```diff
@@ -93,5 +94,22 @@ private LinkedTestCaseResult Link (TestCase testCase, TestCaseSandbox sandbox, M
 		protected virtual void AddAdditionalLinkOptions (LinkerArgumentBuilder builder, TestCaseMetadaProvider metadataProvider)
 		{
 		}
+
+		private T GetResultOfTaskThatMakesNUnitAssertions<T> (Task<T> task)
+		{
+			try {
+				return task.Result;
+			} catch (AggregateException e) {
+				if (e.InnerException != null) {
```
marek-safar: Why not to check only for `ResultStateException` ?

mrvoorhe: We use an older version of nunit and ResultStateException doesn't exist in our version.  Checking for each exception is a little extra code, but it's more backward compatible with older nunit versions

PR #332, opened 2018-07-12
Fix possible race condition during test assembly compile

mrvoorhe:
If the input assembly compilation threw, for example,  due to an Assert.Ignore() , we would not wait on the expectations compilation to complete for continuing.  This could lead to a race condition with the next test that runs over files in the sandbox directory.

This PR fixes the race by waiting on the expectations compilation task to complete when an exception is thrown.


mrvoorhe:
build

PR #341, opened 2018-08-07
Add PreserveDependencyInjectorStep.

baulig:
If anything in `mscorlib`, `System`, `System.Core` or `System.Security`
references the dependency injector (via `Mono.ISystemDependencyProvider`),
then we must keep the `System.dll` side of it alive by preserving
`Mono.SystemDependencyProvider`.

### DIFF
--- a/linker/Linker.Steps/PreserveDependencyInjectorStep.cs
+++ b/linker/Linker.Steps/PreserveDependencyInjectorStep.cs
@@ -1,0 +1,107 @@
+﻿//
+// PreserveDependencyInjectorStep.cs
+//
+// Author:
+//       Martin Baulig <mabaul@microsoft.com>
+//
+// Copyright (c) 2018 Xamarin Inc. (http://www.xamarin.com)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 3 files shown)


baulig:
The only alternative to the extra step is to always preserve `Mono.SystemDependencyProvider` in `System.dll` - whether it is used or not.  This extra step is similar to what we've been doing with crypto / TLS Provider on mobile in the past.

baulig:
The reason it is failing is that once you use the `X509Certificate2` constructor instead of the obsolete `Import()` method (which will throw PNSE with "Duck Blue"), then all direct references to `Mono.SystemDependencyProvider` (or code that calls it from `System.dll`) are gone and the only "link" between the corlib-side and the System-side is the reflection load code.

Once we move into 2018-10 / 2018-12 the use case of the dependency injector is likely getting bigger as additional crypto code (RSA and friends) might be added to it.  So I don't think that always preserving the System-side is a good choice.

For mobile itself (XI and XM), I have a different solution in mind which will not require an extra linker step.  A separate PR on this issue will follow shortly - I needed the `2018-08` branch to be up for this.

marek-safar:
I have implemented it using existing PreserveDependencyAttribute logic

PR #343, opened 2018-08-07
Add search dir where Windows.winmd can be found

mrvoorhe:
With the fix merged in https://github.com/mono/linker/pull/340 many unit tests now need to be able to resolve Windows.winmd when linking against .NET Framework class libraries.  This is happening because a GetType of `System.Runtime.WindowsRuntime, System.Resources.WindowsRuntimeResourceManager` that was previously not handled is now handled.  Bringing in `System.Runtime.WindowsRuntime` eventually leads to trying to resolve `Windows` which fails.

I'm not sure if this is the best way of resolving Windows.winmd.   I copied the logic from what we do in UnityLinker for our windows runtime support and it worked.  It does seem a little fragile given that the logic expects a specific version directory.  But if it works I'm happy. 

If you know of a more robust way to locate windows.winmd I'm happy to change it.

### DIFF
--- a/linker/Linker/AssemblyResolver.cs
+++ b/linker/Linker/AssemblyResolver.cs
@@ -29,6 +29,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using Microsoft.Win32;
 using Mono.Cecil;
 
 namespace Mono.Linker {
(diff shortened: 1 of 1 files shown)


mrvoorhe:
@marek-safar Any feelings on which approach to take for fixing the .NET Core build due to the inclusion of `Registery`?

PR #348, opened 2018-08-10
Handle PreserveDependency which imports new assembly dependencies better

marek-safar:
by introducing new step to setup all hierarchical dependencies correctly


mrvoorhe:
Can you add a new test that is a variation of `PreserveDependencyMethodInAssembly` where `Main()` does not call `Dependency ();` and asserts "library.dll" is removed?

PR #363, opened 2018-09-11
Add support for asserting CustomAttributes in other assemblies

mrvoorhe:
included 2 two test attributes for asserting attributes on Assemblies, Types, and Members in other assemblies.


mrvoorhe:
@marek-safar I've dropped the change to `ShouldMarkTopLevelAttribute` from this PR.  There's really no reason that change needs to be in upstream and I was able to move it into our overridden `ShouldMarkTopLevelAttribute`

This PR now only includes new test framework attributes

PR #366, opened 2018-09-14
Extends PreserveDependencyField to work on fields too

marek-safar:


mrvoorhe:
lgtm

Out of curiosity, in what situation is preserving a method because a field is marked useful?

marek-safar:
We find that useful for special attribute types which have runtime support (DllImport, MarshalAs, etc)

PR #371, opened 2018-09-19
Optimizes marking of static constructor for beforefieldinit types.

marek-safar:
When the type is marked with beforefieldinit the static constructor
was generated from static field initializers only. Therefore we
can mark it after any of the static field is marked.

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -465,6 +465,11 @@ namespace Mono.Linker.Steps {
 			return true;
 		}
 
+		protected virtual bool ShoulMarkTypeConstructor (TypeDefinition type)
+		{
+			return !type.IsBeforeFieldInit;
+		}
+
 		protected virtual bool ShouldMarkTopLevelCustomAttribute (AttributeProviderPair app, MethodDefinition resolvedConstructor)
 		{
 			var ca = app.Attribute;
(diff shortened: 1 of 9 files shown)


mrvoorhe:
Neat.  I didn't know about `type.IsBeforeFieldInit`

Having said that, I don't think this is a safe way to implement it.  Consider these two test cases

```
	public class SideEffectCctor
	{
		[Kept]
		private static int counter;
		
		public static void Main()
		{
			counter = 0;
			Foo.Method();
			Console.WriteLine(counter);
		}

		public class Foo
		{
			[Kept]
			public static Bar o;

			[Kept]
			static Foo()
			{
				o = new Bar();
			}

			[Kept]
			public static void Method()
			{
			}
		}

		[KeptMember(".ctor()")]
		public class Bar
		{
			public Bar()
			{
				counter++;
			}
		}
	}
```

and

```
	public class SideEffectCctorWithInitializer
	{
		[Kept]
		private static int counter;
		
		public static void Main()
		{
			counter = 0;
			Foo.Method();
			Console.WriteLine(counter);
		}

		[KeptMember(".cctor()")]
		public static class Foo
		{
			[Kept]
			public static Bar o = new Bar();
			
			[Kept]
			public static void Method()
			{
			}
		}

		[KeptMember(".ctor()")]
		public class Bar
		{
			public Bar()
			{
				counter++;
			}
		}
	}
```

If you ran the linked test executable produced by both of these today, they would write `1` to stdout.

However, with this PR `SideEffectCctor` would output `1` while `SideEffectCctorWithInitializer` would output `0`.

I know the spec is more complex than normal with regards to expected behavior with static constructors, but even if this were valid according to the spec, I don't think it's a good idea for the linker to start knocking out side effects that can cause noticeable behavior changes in a program.  At least not be default.

marek-safar:
That's actually Mono bug, if you run the code on .NET (without any linking) you will also get 0

mrvoorhe:
The results I gave were running on .NET built in debug.

I saw your mono issue go by.  https://github.com/mono/mono/issues/10710

After reading your issue I think my feeling on this is, if the behavior of a program on a given run time changes for a stripped vs unstripped assembly we should proceed with caution.

I'm not sold yet that I want our UnityLinker to behave this way.  There would need to be considerable size savings to justify the risk.

mrvoorhe:
lgtm

PR #372, opened 2018-09-20
Move [PreserveDependency] tests into their own folder

mrvoorhe:
This is a growing feature and I think it's worthy of it's own directory to keep it's tests organized.


mrvoorhe:
Fixed conflicts

PR #377, opened 2018-09-26
Add tests for non required type preservations in link xml files

mrvoorhe:
We didn't have tests covering the `required="0"` xml attribute that can be used on types


marek-safar:
Could you resolve the conflict?

mrvoorhe:
Fixed.

PR #382, opened 2018-10-05
Link culture calendars preservation with -l option

marek-safar:


mrvoorhe:
Ideally I'd like to see at least 1 test to ensure this works.  Testing this is a little tricky since it's only valid when linking the mono class libraries.  We'd need to ignore the test when ran from an IDE on Windows.  I think the simplest way to do this would be to add a new group of tests for tests specific to the mono class libraries.

```
		[TestCaseSource (typeof (TestDatabase), nameof (TestDatabase.MonoClassLibraryTests))]
		public virtual void MonoClassLibraryTests (TestCase testCase)
		{
			if (Environment.OSVersion.Platform == PlatformID.Win32NT)
				Assert.Ignore("Not supported  when linking against .NET Framework");
			Run (testCase);
		}
```

I made the above virtual so that I can override it in our UnityLinker test suite and run it on Windows.  When we run the upstream tests we always supply our mono class libraries regardless of which platform we are on.

With that in place it would be simple to add tests and do some asserts to ensure this step is working as expected.

mrvoorhe:
For my own awareness, what is the motivation for this change?  Is this replacing preservations in the embedded mscorlib.xml?

marek-safar:
We have tests for that in Mono repo but could possibly copy them here just not sure how to check they work because for net_4_x mscorlib the calendars could be always there.

Motivation is to port the logic we have in XA/XI to one place where it can be shared with other targets

mrvoorhe:
That does make testing more complicated, o well.  I think it's fine to pass on adding tests then.

marek-safar:
yes, we have that covered elsewhere which is not ideal but covered.

(more changes which will require testing infrastructure are coming soon)

PR #383, opened 2018-10-08
Preserve attributes for fixed length arrays

mrvoorhe:
A couple compiler generated attributes should be kept for the runtime when only keep used attributes is enabled

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -453,6 +453,9 @@ namespace Mono.Linker.Steps {
 				case "System.ThreadStaticAttribute":
 				case "System.ContextStaticAttribute":
 					return true;
+				// Attributes related to `fixed` keyword used to declare fixed length arrays
+				case "System.Runtime.CompilerServices.FixedBufferAttribute":
+					return true;
 				case "System.Runtime.InteropServices.InterfaceTypeAttribute":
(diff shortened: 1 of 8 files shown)


linker/Linker.Steps/MarkStep.cs
```diff
@@ -453,6 +453,10 @@ protected virtual bool ShouldMarkCustomAttribute (CustomAttribute ca, ICustomAtt
 				case "System.ThreadStaticAttribute":
 				case "System.ContextStaticAttribute":
 					return true;
+				// Attributes related to `fixed` keyword used to declare fixed length arrays
+				case "System.Runtime.CompilerServices.FixedBufferAttribute":
```
marek-safar: Only FixedBufferAttribute is needed to be persisted

marek-safar: Actually, on second look. I think we should move it out of _context.KeepUsedAttributeTypesOnly as well.

mrvoorhe: Why is that?  We keep all attributes when `!_context.KeepUsedAttributeTypesOnly`

mrvoorhe: fyi, when I made the first fix I added a version of the test for when  `_context.KeepUsedAttributeTypesOnly` is not enabled.  That will give us test coverage on the behavior when `!_context.KeepUsedAttributeTypesOnly`

PR #386, opened 2018-10-10
Remove more COM attributes when the feature is disabled

marek-safar:


mrvoorhe:
Few things I don't like about this change in approach.

1) The only step that removes things today is the SweepStep.  This new step would introduce a second step that removes things.

2) One behavior of used attributes only is that if you mark the attribute type, for example by using link.xml, then all CustomAttributes of that type will be retained.  This PR as it is right now breaks that rule for these com attribute types.  As unlikely as it seems that someone would want to preserve CustomAttributes of one of these types, I'd rather not introduce exceptions to this rule.

3) The SweepStep makes virtual callbacks for most things that it remotes.  This allows us to override these methods and track what is removed.  We depend on this information for diagnostic reporting.  This new step doesn't let us track removal of these attributes.  You could similar callbacks, but then it ties back into 1 above, where it's easier to have everything in one place.

What was the motivation for pulling this out into a separate step as opposed to just adding more cases in the switch?

mrvoorhe:
> The only step that removes things today is the SweepStep. This new step would introduce a second step that removes things.

Your other security PR reminded me that my statement was not 100% accurate.  The `RemoveSecurityStep` does do removal.  I feel like there was I reason made it remove things directly, but I can't recall exactly why.  Either way, I'm still hesitant to introduce another removal step when it appears that we should be able to follow the Don't Mark & Sweep pattern.

marek-safar:
> What was the motivation for pulling this out into a separate step as opposed to just adding more cases in the switch?

The motivation is to have clear remove (or clear step). I plan to add significantly more logic here including explicit removal of method bodies, which could just make Mark too complex. I also think allowing MarkStep to remove attributes was a mistake because we want to have the pipeline customizable to include/exclude optional steps.

> The SweepStep makes virtual callbacks for most things that it removes.

Could make whatever you need virtual or you could not include this step at all in your pipeline.

> One behavior of used attributes only is that if you mark the attribute type, for example by using link.xml, then all CustomAttributes of that type will be retained.

and that's intended when you exclude a specific feature you don't want any code/metadata to be present.

mrvoorhe:
> and that's intended when you exclude a specific feature you don't want any code/metadata to be present.

I can see that argument, but I'd argue the other way.  I'd argue that link.xml is a trumps all mechanism.  If you say you want something to happen via link.xml, it shall be done.  I'd prefer not to get into a situation where we are saying "ya that <type> preservation will work, but if you used X option then we are not going to respect it".

Also link.xml has the `feature` attribute.  If you wanted something to only be excluded for com you could.

Part of the purpose of link.xml is a backdoor to override the linkers behavior.  And in my view to preserve anything a user might want to.  If we introduce a bug, link.xml should be there as a workaround.  By not respecting the type preserve we would effectively be closing off link.xml in these cases as a way to work around a mistake we've made.

mrvoorhe:
> The motivation is to have clear remove (or clear step).
Good to know.  Do you think it would be better to have 1 step per "Feature"?  That would make it easier to opt in or out of the feature by adding or not adding the step to the pipeline.

> I plan to add significantly more logic here 
What else do you plan on adding?  I'd like to try and understand where you are headed so that I can think about the best design going forward.

For example, I added a "MarkStepExtension" concept to our UnityLinker.  I piggy back on the `DoAdditional*Processing` callbacks in the MarkStep.  And then for each of those callbacks we process a list of `IMarkStepExtension` classes to see if they want to do anything.  The `IMarkStepExtension` are much simpler than an IStep would be since they can leverage the traversal of types and members performed by the `MarkStep`.

I don't know if a pattern like this would be helpful in some of your cases or not.

>  including explicit removal of method bodies
I implemented something that sounds similar in our UnityLinker awhile ago.  You can tell the linker to wipe out a methods body and replace it with a throw.

```
<method signature="System.Void MethodViaSignature()" stub="force"/>
<method signature="System.Void MethodViaSignature2()" stub="ifnotused"/>
```

Is that similar to what you are thinking of?  I can bring it upstream if it would help you.  We never ended up using or documenting it so the API could be changed if you wanted without us having to worrying about backward compatibility

mrvoorhe:
Another considerable for the pattern you are headed down here is performance.

For our method body stripping logic I added it as a step before the MarkStep because that was the easiest way to fully strip bodies and then have the stripped types, fields, methods, etc not get marked.

I knew we'd pay a price for processing bodies that would never be marked anyway, but given the complexity of stripping bodies I had to do it this way.

In hindsight this has helped highlight that as a general pattern it is more performant to only process the things you are going to keep.  Why waste time removing attributes from types, methods, fields that were never going to be marked anyways?

marek-safar:
@mrvoorhe  Quite strange why does it have an impact on FixedBufferAttribute tests

mrvoorhe:
@marek-safar PR to fix FixedBuffer tests 

https://github.com/mono/linker/pull/387

You'll need to merge that PR and rebase.

PR #388, opened 2018-10-14
Always remove empty static constructors

marek-safar:


mrvoorhe:
I wonder if there is a way to maintain the simple test writing experience we have today.  Being required to add `[AddedPseudoAttributeAttribute ((uint)TypeAttributes.BeforeFieldInit)]` to even the simplest of tests is going to be confusing and not intuitive.

Also, I'm not looking forward to the number of our tests I'd have to update as well.

Thinking out loud, maybe we could add a `CodeOptimizations` to disable this behavior and then have the test framework somehow disable it by default but allow tests to enable it.  Then we can have a small subset of tests that test this feature and then all of the other tests don't have to worry about adding `[AddedPseudoAttributeAttribute ((uint)TypeAttributes.BeforeFieldInit)]`?

Maybe there is some other way to accomplish the same goal?

marek-safar:
It should not be necessary to update many tests due to this (I had to change only 1 so far)

marek-safar:
It should not be necessary to update many tests due to this (I had to change only 1 so far)

marek-safar:
It should not be necessary to update many tests due to this (I had to change only 1 so far)

mrvoorhe:
Weird.  I saw 23 failures from CI when I originally looked at the PR.  Must have been caused by something else.

PR #389, opened 2018-10-17
Fix disable before field init not working

mrvoorhe:
* Disabling the option didn't restore the old before 

* ToLower() on the disable opt values so that all lower and/or enum value name works as you would expect

* Refactor static constructor marking in MarkStep.  Centralized logic and fixed if statements not working correctly when the optimization was disabled

* Fixed typo in method name

* Added new test to verify disabling works (test will fail without my changes to the MarkStep and Driver)

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -468,9 +468,21 @@ namespace Mono.Linker.Steps {
 			return true;
 		}
 
-		protected virtual bool ShoulMarkTypeConstructor (TypeDefinition type)
+		protected virtual bool ShouldMarkTypeStaticConstructor (TypeDefinition type)
 		{
-			return !type.IsBeforeFieldInit && _context.IsOptimizationEnabled (CodeOptimizations.BeforeFieldInit);
+			if (Annotations.HasPreservedStaticCtor (type))
+				return false;
+			
+			if (type.IsBeforeFieldInit && _context.IsOptimizationEnabled (CodeOptimizations.BeforeFieldInit))
+				return false;
+
+			return true;
+		}
+
+		protected void MarkStaticConstructor (TypeDefinition type)
+		{
+			if (MarkMethodIf (type.Methods, IsStaticConstructor))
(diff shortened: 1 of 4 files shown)


linker/Linker.Steps/MarkStep.cs
```diff
@@ -468,9 +468,27 @@ protected virtual bool ShouldMarkCustomAttribute (CustomAttribute ca, ICustomAtt
...
+			if (Annotations.HasPreservedStaticCtor (type))
+				return false;
+			
+			if (type.IsBeforeFieldInit && _context.IsOptimizationEnabled (CodeOptimizations.BeforeFieldInit))
+				return false;
+
+			return true;
+		}
+
+		protected void MarkStaticConstructor (TypeDefinition type, bool force = false)
```
marek-safar: Please keep the logic separated this should do only
```
				if (MarkMethodIf (type.Methods, IsStaticConstructor))
					Annotations.SetPreservedStaticCtor (type);
```

linker/Linker.Steps/MarkStep.cs
```diff
@@ -822,10 +834,8 @@ protected void MarkField (FieldReference reference)
 			DoAdditionalFieldProcessing (field);
 
 			var parent = reference.DeclaringType.Resolve ();
-			if (parent.IsBeforeFieldInit && !Annotations.HasPreservedStaticCtor (parent) && _context.IsOptimizationEnabled (CodeOptimizations.BeforeFieldInit)) {
-				if (MarkMethodIf (parent.Methods, IsStaticConstructor))
-					Annotations.SetPreservedStaticCtor (parent);
-			}
+			if (!Annotations.HasPreservedStaticCtor (parent))
```
marek-safar: Shouldn't this be `ShouldMarkTypeStaticConstructor` ?

mrvoorhe: All the tests pass 😄 

I wasn't sure.  I actually had `ShouldMarkTypeStaticConstructor` at first.  But then the `StaticFieldInitializer` test fails.

PR #393, opened 2018-10-23
IsImport and IsWindowsRuntime set to false when COM attrs removed

mrvoorhe:
Once all the COM attributes are removed we need to set IsImport and IsWindowsRuntime to false since the type is no longer really a com or windows runtime type.

### DIFF
--- a/linker/Linker.Steps/RemoveFeaturesStep.cs
+++ b/linker/Linker.Steps/RemoveFeaturesStep.cs
@@ -17,7 +17,11 @@ namespace Mono.Linker.Steps
 
 		void ProcessType (TypeDefinition type)
 		{
-			RemoveCustomAttributes (type);
+			if (RemoveCustomAttributes (type)) {
+				if (FeatureCOM && type.IsImport) {
+					type.IsImport = false;
+				}
+			}
 
 			foreach (var field in type.Fields)
 				RemoveCustomAttributes (field);
@@ -30,14 +34,16 @@ namespace Mono.Linker.Steps
 				ProcessType (nested);
 		}
 
-		void RemoveCustomAttributes (ICustomAttributeProvider provider)
+		bool RemoveCustomAttributes (ICustomAttributeProvider provider)
 		{
 			if (!provider.HasCustomAttributes)
-				return;
+				return false;
 
 			var attrsToRemove = provider.CustomAttributes.Where (IsCustomAttributeExcluded).ToArray ();
 			foreach (var remove in attrsToRemove)
 				provider.CustomAttributes.Remove (remove);
+			
+			return attrsToRemove.Length > 0;
 		}
 
 		bool IsCustomAttributeExcluded (CustomAttribute attr)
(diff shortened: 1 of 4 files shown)


linker/Linker.Steps/RemoveFeaturesStep.cs
```diff
@@ -17,79 +18,103 @@ protected override void ProcessAssembly (AssemblyDefinition assembly)
 
 		void ProcessType (TypeDefinition type)
 		{
-			RemoveCustomAttributes (type);
+			RemoveCom (type);
```
marek-safar: Please undo the unnecessary rename, more features than COM will be placed here

linker/Linker.Steps/RemoveFeaturesStep.cs
```diff
@@ -17,79 +18,103 @@ protected override void ProcessAssembly (AssemblyDefinition assembly)
...
-			case "SetWin32ContextInIDispatchAttribute":
-			case "ManagedToNativeComInteropStubAttribute":
-				if (!FeatureCOM || type.Namespace != "System.Runtime.InteropServices")
-					return false;
+			return Annotations.GetPreserve (definition) == TypePreserve.All;
+		}
 
-				break;
-			}
+		public static bool IsComAttribute (CustomAttribute attr)
```
marek-safar: Is this really necessary to make it public and have multiple switch statements for each feature in the future?

mrvoorhe:
> Actually, I don't think the change should be done in this step at all. It should go to SweepStep where we do such clean ups.

How should I decide when to set IsImport to false from the SweepStep?  When a type is IsImport and doesn't have a GuidAttribute anymore?  Or when the com feature is excluded?

The MarkStep has 3 usages of `IsImport`.    Our UnityMarkStep does additional marking for IsImport and IsWindowsRuntime.  Did you consider the behavior difference between setting IsImport to false in RemoveFeaturesStep vs SweepStep?  I can't say that I did initially, I was just looking to centralize the logic.

I looked at the MarkStep usages of IsImport a little bit, not sure I understand the implications of setting IsImport to false before or after the MarkStep yet.

marek-safar:
Good point. I didn't consider anyone checking IsImport before SweepStep. Let's remove it in this step but let's do it without making everything here com specific.

You could just alter ProcessType to start with something like

```
	if (RemoveCustomAttributes (type)) {
		if (!FeatureCOM && method.IsImported)
			method.IsImported = false;
	}
```

This is the second time I am looking at ProcessInteropMethod and it looks quite suspicious and don't we have any coverage for that odd logic there

linker/Linker.Steps/RemoveFeaturesStep.cs
mrvoorhe: I don't know if we are going to stick with using the `RemoveFeatureStep` or not yet, but in case we didn't, I thought it would be nice to have a helper I could use for checking if a CustomAttribute is COM related.

Not a big deal though.  I can just copy the switch if we don't use `RemoveFeatureStep`

mrvoorhe:
@marek-safar updated following the approach you suggested

PR #394, opened 2018-10-24
Implement interface sweeping

mrvoorhe:
This PR introduces two new behavior changes

* Interfaces on a type will no longer be marked simply because the type is marked. Instead, marking of interfaces on types is deferred until that interface type is marked for some other reason.  Except in a few situations

1) COM and WindowsRuntime interface implementations will be marked

2) If a type has TypePreserve.All, we will mark all of it's interface.

* When a type has no instance constructors marked, we will remove all interfaces from that type (because no instance can possibly exist).

How this works.

* Instantiation marking.  This is used to track when it is now possible an instance of this type could have been created.  (Side note : a future PR to do "base type sweeping" will build on top of this information)

* Defer marking of overrides until it is possible an instance of that type could be created.

* defer marking of interface implementations until the interface type is marked and the type is marked as instantiated.

Other notes

* I had to update a number of existing tests due to these new rules.

* I added a number of new interface tests just to increase our coverage

* ~~This PR has a test that depends on https://github.com/mono/linker/pull/393 in order to pass.  This PR will need to wait until that one lands before it can be merged.~~


mrvoorhe:
@marek-safar Don't merge this yet.  I'm still running these changes through our il2cpp test suite to provide some additional coverage.  I'll let you know when everything looks good.

linker/Linker.Steps/MarkStep.cs
```diff
@@ -811,6 +835,34 @@ bool ProcessLateMarkedAttributes ()
 			return markOccurred;
 		}
 
+		bool ProcessOnlyIfInstantiated ()
+		{
+			// Certain things on a type only need to be marked if an instance of the type could be instantiated
+			var startingQueueCount = _onlyIfInstantiated.Count;
```
marek-safar: This variable seems to be redundant

linker/Linker.Steps/MarkStep.cs
```diff
@@ -811,6 +835,34 @@ bool ProcessLateMarkedAttributes ()
...
+				}
+				
+				MarkRequirementsForInstantiatedTypes (type);
+			}
+
+			// requeue the items we skipped in case we need to make another pass
+			foreach (var item in skippedItems)
+				_onlyIfInstantiated.Enqueue (item);
+
+			return markOccurred;
```
marek-safar: Where is this set?

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1207,6 +1273,21 @@ void MarkNamedProperty (TypeDefinition type, string property_name)
...
 		}
 
+		void MarkInterfaceImplementations (TypeDefinition type)
+		{
+			if (type.HasInterfaces) {
+				foreach (var iface in type.Interfaces) {
+					// Only mark interface implementations of interface types that have been marked.
+					// This enables stripping of interfaces that are never used
+					var resolvedInterfaceType = iface.InterfaceType.Resolve ();
+					if (resolvedInterfaceType == null)
```
marek-safar: I think we generally throw on this case

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1207,6 +1273,21 @@ void MarkNamedProperty (TypeDefinition type, string property_name)
 			}
 		}
 
+		void MarkInterfaceImplementations (TypeDefinition type)
+		{
+			if (type.HasInterfaces) {
```
marek-safar:
```suggestion
if (!type.HasInterfaces)
	return;
```

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1652,6 +1733,9 @@ protected virtual void ProcessMethod (MethodDefinition method)
 
 			MarkGenericParameterProvider (method);
 
+			if (ShouldMarkAsInstancePossible (method))
```
marek-safar: Why is this extra check needed?

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1701,6 +1785,29 @@ protected virtual void DoAdditionalMethodProcessing (MethodDefinition method)
...
+
+			if (method.DeclaringType.IsInterface)
+				return true;
+
+			return false;
+		}
+
+		protected virtual void MarkRequirementsForInstantiatedTypes (TypeDefinition type)
+		{
+			Annotations.MarkInstantiated (type);
```
marek-safar: Isn't this redundant?

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1701,6 +1785,29 @@ protected virtual void DoAdditionalMethodProcessing (MethodDefinition method)
 		{
 		}
 
+		protected virtual bool ShouldMarkAsInstancePossible (MethodDefinition method)
+		{
+			// We don't need to mark it multiple times
+			if (Annotations.IsInstantiated (method.DeclaringType))
+				return false;
+
+			if (method.IsConstructor && !method.IsStatic)
```
marek-safar: If the type ctor it not marked we could convert all instance methods to throw NSE

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1701,6 +1785,29 @@ protected virtual void DoAdditionalMethodProcessing (MethodDefinition method)
...
+				return true;
+
+			if (method.DeclaringType.IsInterface)
+				return true;
+
+			return false;
+		}
+
+		protected virtual void MarkRequirementsForInstantiatedTypes (TypeDefinition type)
```
marek-safar: Call it based on what it actually does, it should be something like `MarkInterfaceImplementation`

linker/Linker.Steps/MarkStep.cs
```diff
@@ -1895,10 +2002,30 @@ protected virtual void HandleUnresolvedMethod (MethodReference reference)
...
+
+			if (Annotations.IsMarked (resolvedInterfaceType) && !Annotations.IsMarked (iface))
+				return true;
+
+			// It's hard to know if a com or windows runtime interface will be needed from managed code alone,
+			// so as a precaution we will mark these interfaces once the type is instantiated
+			if (resolvedInterfaceType.IsImport || resolvedInterfaceType.IsWindowsRuntime)
+				return true;
+
+			if (!Annotations.IsPreserved (type))
```
marek-safar: Looks like a redundant check

linker/Linker.Steps/MarkStep.cs
mrvoorhe: It's on my radar.

I hacked in logging for this scenario in the SweepStep in my `master-base-sweeping` branch where I'm doing base type sweeping (sorting out final details now, it will follow this PR).   Once you have interface + base sweeping the number of methods that fall into this situation is small for `Console.Writeline("Hello world")` program.  Still worth doing, but I wouldn't expect anything significant to come from it.

Here's the list of instance methods, with bodies, on types with no instance ctor that survive sweeping.  And keep in mind we apply some some additional mscorlib.dll preservations.  So this list would be the same or smaller for normal mono + monolinker.
```
T1 System.Tuple`4::get_Item1()
T2 System.Tuple`4::get_Item2()
T3 System.Tuple`4::get_Item3()
T4 System.Tuple`4::get_Item4()
System.Reflection.MethodInfo System.Delegate::get_Method()
System.Object System.Delegate::get_Target()
System.Reflection.MethodInfo System.Delegate::GetMethodImpl()
System.Delegate[] System.Delegate::GetInvocationList()
System.Delegate System.Delegate::CombineImpl(System.Delegate)
System.Delegate System.Delegate::RemoveImpl(System.Delegate)
System.Delegate System.DelegateSerializationHolder/DelegateEntry::DeserializeDelegate(System.Runtime.Serialization.SerializationInfo,System.Int32)
System.MarshalByRefObject System.Runtime.Remoting.Proxies.ProxyAttribute::CreateInstance(System.Type)
System.Runtime.Remoting.Proxies.RealProxy System.Runtime.Remoting.Proxies.ProxyAttribute::CreateProxy(System.Runtime.Remoting.ObjRef,System.Type,System.Object,System.Runtime.Remoting.Contexts.Context)
System.RuntimeType System.Runtime.Remoting.Proxies.TransparentProxy::GetProxyType()
System.Boolean System.Runtime.Remoting.Proxies.TransparentProxy::get_IsContextBoundObject()
System.Runtime.Remoting.Contexts.Context System.Runtime.Remoting.Proxies.TransparentProxy::get_TargetContext()
System.Boolean System.Runtime.Remoting.Proxies.TransparentProxy::InCurrentContext()
System.Object System.Runtime.Remoting.Proxies.TransparentProxy::LoadRemoteFieldNew(System.IntPtr,System.IntPtr)
System.Void System.Runtime.Remoting.Proxies.TransparentProxy::StoreRemoteField(System.IntPtr,System.IntPtr,System.Object)
System.Runtime.InteropServices.MarshalAsAttribute System.Runtime.InteropServices.MarshalAsAttribute::Copy()
System.Void System.Runtime.InteropServices.SafeBuffer::AcquirePointer(System.Byte*&)
System.Void System.Runtime.InteropServices.SafeBuffer::ReleasePointer()
```

linker/Linker.Steps/MarkStep.cs
mrvoorhe: It will be doing more soon.  I add to this in my base type sweeping branch that is coming.

linker/Linker.Steps/MarkStep.cs
mrvoorhe: The next method call, `Annotations.GetPreserve (type)` throws if `!Annotations.IsPreserved (type)`.   I'd be more comfortable with the call to `Annotations.GetPreserve` guarded by the 

```
if (!Annotations.IsPreserved (type))
    return false;
```

I'm not 100% confident this is 100% redundant.  Maybe I'm just missing it.  I would rather err on the side of caution than depend on implicit conditions based on other state we track.

Do you want to elaborate on why you think it's redundant?

The fact that `Annotations.GetPreserve` throws is a constant annoyance to me.  I understand why it does, but a lot of the times I find myself querying this information I do not want anything to throw.  Maybe the solution is to add a `TryGetPreserve` method?

linker/Linker.Steps/MarkStep.cs
mrvoorhe: In some cases yes.

When `MarkRequirementsForInstantiatedTypes ` is called from `MarkType`

```
			if (type.IsInterface) {
				// There's no benefit to deferring processing of an interface type until we know a type implementing that interface is marked
				MarkRequirementsForInstantiatedTypes (type);
			} else if (type.IsValueType) {
				// Note : Technically interfaces could be removed from value types in some of the same cases as reference types, however, it's harder to know when
				// a value type instance could exist.  You'd have to track initobj and maybe locals types.  Going to punt for now.
				MarkRequirementsForInstantiatedTypes (type);
			} else {
				// There are a number of markings we can defer until later when we know it's possible a reference type could be instantiated
				// For example, if no instance of a type exist, then we don't need to mark the interfaces on that type
				_onlyIfInstantiated.Enqueue (type);
			}
```

`Annotations.MarkInstantiated (type)` is *not* redundant.

When `MarkRequirementsForInstantiatedTypes ` is called from `ProcessMethod`

```
if (ShouldMarkAsInstancePossible (method))
    Annotations.MarkInstantiated (method.DeclaringType);
```

Yes, it *is* redundant.

Now that you point it out.  If I unify this behavoir, I think I can get rid of the entire `_onlyIfInstantiated` queue processing.....I'll try it.

mrvoorhe: To my slight surprise, by changing the usage in `Processmethod` to be 

```
if (ShouldMarkAsInstancePossible (method))
    MarkRequirementsForInstantiatedTypes (method.DeclaringType);
```

I was able to remove the entire `ProcessOnlyIfInstantiated()` mechanism.

linker/Linker.Steps/MarkStep.cs
mrvoorhe: I was able to eliminate this entire method.

linker/Linker.Steps/MarkStep.cs
mrvoorhe: I was able to eliminate this entire method.

linker/Linker.Steps/MarkStep.cs
mrvoorhe: FYI

I stumbled on a tricky aspect of instance methods that are marked when no ctor is marked.  It's going to be a bit more complicated than I initially realized.  Ideally, we will defer method body processing until an instance ctor is marked.  Otherwise you end up marking things in a body that you are going to stub out with a throw

mrvoorhe:
@marek-safar Status update on testing these changes in the il2cpp repo.

* I hit an issue with our Windows Runtime support.  We have logic expecting some interfaces to be on a type and they've all been stripped away.  Could be windows runtime specific or could be another com scenario that needs handled.   Probably safest to wait until I know what the issue is to merge.

* One of our il2cpp tests hangs if we don't have `<type fullname="System.Runtime.Remoting.Messaging.AsyncResult" preserve="all" />`.  Given that we are using mono class libs and il2cpp runtime is heavily based on mono runtime I'd like to figure out what is going on here before landing this PR.

linker/Linker.Steps/MarkStep.cs
marek-safar: Let's clean this in separate PR by renaming GetPreserve to TryGetPreserve. The uses of `GetPreserve` are not always correct either.

marek-safar:
Windows Runtime support is in the same bucket as COM and will have some hard assumptions which will probably have to be special cased.

AsyncResult is something we are happy to fix on Mono side once you track it down ;-)

mrvoorhe:
The windows runtime issue I hit is specific to COM/Windows runtime.  I will address this on our end rather than in monolinker.

@marek-safar this PR is ready for final review & merging.  Just remember to have https://github.com/mono/mono/pull/11459 any place you use mono + monolinker with these changes.

mrvoorhe:
There is one test failing `CanEnableReducedTracing`.  Nothing is really broken, these changes just send it over a threshold that was used to indicate something was probably wrong.  I can raise the threshold and get the test passing again.

However, while looking at the failure I noticed that I overlooked how these changes impact tracing.

There a new entries showing up in the `linker-dependencies.xml`

There are a few new elements that end up being slightly useless.
Ex:
```
<edge mark="1" b="Method:System.Void System.Reflection.FieldInfo::.ctor()" e="InterfaceImpl:Mono.Cecil.InterfaceImplementation" />
```

And then there are about 3x more that are really useless.

```
<edge mark="1" b="Other:Mono.Linker.Steps.MarkStep" e="InterfaceImpl:Mono.Cecil.InterfaceImplementation" />
```

I'm not sure what to do about `e="InterfaceImpl:Mono.Cecil.InterfaceImplementation"`.  It seems like I would have to adjust the `Annotations.Mark` API to put the InterfaceImplementation types into a more meaningful string format.

Then there is the issue of `b="Other:Mono.Linker.Steps.MarkStep"`.  That's happening because of interfaces marked from `ProcessMarkedTypesWithInterfaces`.  I don't know how to easily deal with that either.

@marek-safar @radekdoulik Is it OK if I leave this less than ideal dependency information as a loose end and land this PR the way it is?  I'm open to easy fixes if either of you have any ideas.

PR #396, opened 2018-10-30
Compile ExplicitStaticCtor with /optimize+

mrvoorhe:
This eliminates a source of instability in the test where it could fail if compiled with a compiler that put a `nop` into the cctor

### DIFF
--- a/linker/Tests/Mono.Linker.Tests.Cases/Statics/ExplicitStaticCtor.cs
+++ b/linker/Tests/Mono.Linker.Tests.Cases/Statics/ExplicitStaticCtor.cs
@@ -1,9 +1,11 @@
 ﻿using System.Reflection;
 using System.Runtime.CompilerServices;
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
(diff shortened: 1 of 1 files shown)


marek-safar:
How does /o- alter the test result?

mrvoorhe:
The cctor ends up looking like

```
.method private hidebysig specialname rtspecialname static 
	void .cctor () cil managed 
{
	// Method begins at RVA 0x2073
	// Code size 2 (0x2)
	.maxstack 8

	IL_0000: nop
	IL_0001: ret
} // end
```

And `IsNonEmptyStaticConstructor` returns true, which leads to the test failing

mrvoorhe:
1) If you built in debug, you're probably not that concerned about size

2) The method body stripping logic we use to get our smallest sizes doesn't go out of it's way to handle nops either.

3) Last but not least, I want to get us synced up with `master` so that I can focus on getting the interface sweeping and base type sweeping PR's landed and into UnityLinker.  This was the quickest fix for a test that is failing with my latest sync up attempt.

PR #397, opened 2018-11-05
Mark constructors only in MarkGenericArgumentConstructors ().

vargaz:


linker/Linker.Steps/MarkStep.cs
```diff
@@ -1429,7 +1429,7 @@ void MarkGenericArgumentConstructors (IGenericInstance instance)
 				if (argument_definition == null)
 					continue;
 
-				MarkMethodsIf (argument_definition.Methods, ctor => !ctor.IsStatic && !ctor.HasParameters);
+				MarkMethodsIf (argument_definition.Methods, ctor => ctor.IsConstructor && !ctor.IsStatic && !ctor.HasParameters);
```
marek-safar: Should just call
```suggestion
				MarkDefaultConstructor (argument_definition);
```

and remove the if above

PR #399, opened 2018-11-13
Add tests cover using statement difference between mcs and csc.

mrvoorhe:
@marek-safar I could use your opinion on whether or not you think this is a bug in the linker.

This PR contains just tests to reproduce what I think is a bug.  I did not implement a fix yet.  I wanted to confirm whether or not you think it is a bug before I spend time on a fix.

`AssemblyOnlyUsedByUsingWithCsc` is the test that currently fails and I think it should pass.

What it does is
* Compile an assembly called `library.dll`

* Compile an assembly called `copied.dll` that makes use of `library.dll` with only the following code

```
// This is what triggers the behavior difference between Roslyn and mcs.  Roslyn will keep the reference
// to this assembly because of this whereas mcs will not
using ImportantForBug = Mono.Linker.Tests.Cases.References.Dependencies.AssemblyOnlyUsedByUsing_Lib;
```

And `copied.dll` is setup to have `AssemblyAction.Copy`

And most importantly, compile `copied.dll` with `csc`

Note that `csc` will include a reference to `library.dll` in `copied.dll` where as `mcs` will not.

`AssemblyOnlyUsedByUsingWithCsc` fails because I have an assert expecting that `library.dll` survives stripping.  I think this is a valid assertion given that `copied.dll` references `library.dll` and `copied.dll` had `AssemblyAction.Copy`.

It seems to me that there is a bug in that references of an assembly with AssemblyAction.Copy should always exist after stripping, even if they've been stripped down to empty assemblies.  Would you agree?  What are your thoughts?

Note that I added the test `AssemblyOnlyUsedByUsingWithMcs` just to show what happens with `mcs`.  This test passes.

### DIFF
--- a/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/KeptReferencesInAssemblyAttribute.cs
+++ b/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/KeptReferencesInAssemblyAttribute.cs
@@ -1,0 +1,15 @@
+using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Assertions {
+	[AttributeUsage (AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
+	public class KeptReferencesInAssemblyAttribute : BaseInAssemblyAttribute {
+		public KeptReferencesInAssemblyAttribute (string assemblyFileName, string [] expectedReferenceAssemblyNames)
+		{
+			if (string.IsNullOrEmpty (assemblyFileName))
+				throw new ArgumentNullException (nameof (assemblyFileName));
+
+			if (expectedReferenceAssemblyNames == null)
+				throw new ArgumentNullException (nameof (expectedReferenceAssemblyNames));
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


marek-safar:
This looks to me like csc bug (will report that to get the consensus though).

mrvoorhe:
@marek-safar looks like the roslyn issue was closed as `by design` https://github.com/dotnet/roslyn/issues/31192#issuecomment-444672673, shall I proceed to investigate a fix?

mrvoorhe:
I took a brief look at trying to detect the `using` usage but I didn't see anything in the IL that would even let me detect this scenario.

The scenario where this was causing us problems has gone away due to some other unrelated changes.  So as long as w/e the linker does results in references that exist on disk everything is fine.

It looks like SweepStep will already remove all references to assemblies that are deleted, regardless of what the assembly action was.  I will update the expected behavior when `csc is used to match the final output of when `mcs` is used.  I will also add a new test framework attribute to assert the references on the assembly just to be extra certain that the reference to `library` is removed from assemblies that have `AssemblyAction.Copy`.

mrvoorhe:
@marek-safar PR is ready for review.  It's now a test only change as the current behavior of the linker is correct

PR #411, opened 2018-11-27
Fix bug with interface sweeping, explicit impl, and xml

mrvoorhe:
If a type somehow (ex: via link.xml) has an explicit interface member marked, then we need to mark the interface implementation for that member.  Otherwise you end up with invalid IL.

The new tests that reproduced the bug were

`UnusedTypeHasExplicitInterfaceMethodPreservedViaXml `
`UnusedTypeHasExplicitInterfacePropertyPreservedViaXml `

The other tests in this PR were added to fill out coverage of expected behavior with explicitly implemented interfaces

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -1691,8 +1691,10 @@ namespace Mono.Linker.Steps {
 			}
 
 			if (method.HasOverrides) {
-				foreach (MethodReference ov in method.Overrides)
+				foreach (MethodReference ov in method.Overrides) {
 					MarkMethod (ov);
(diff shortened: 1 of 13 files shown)


mrvoorhe:
I opened a separate PR to fix the test failure. https://github.com/mono/linker/pull/412

Once that PR lands I will rebase and this PR should be green

mrvoorhe:
Rebased.  Should be green now

mrvoorhe:
rebased again to fix conflict

mrvoorhe:
One more rebase to fix conflicts again.  I knew this was coming when I opened the PR's.  😄

PR #413, opened 2018-11-27
Fix an edge case with interface sweeping, no instance ctor, and preserve all

mrvoorhe:
When a type has TypePreserve.All it should retain all of it's interfaces.  This was not happening when the type had no instance constructors.  This is a rare border line invalid edge cases, however, we ran into it and PEVerify says the assembly is valid even though the type has no instance constructors.

This PR adds a check during `InitializeType` to ensure this scenario is handled

Note : This branch won't compile until the following open PR's land

https://github.com/mono/linker/pull/407
https://github.com/mono/linker/pull/406

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -105,6 +105,12 @@ namespace Mono.Linker.Steps {
 
 			MarkType (type);
 
+			// Edge case to cover a scenario where a type has preserve all, implements interfaces, but does not have any instance ctors.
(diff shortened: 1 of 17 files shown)


mrvoorhe:
Rebased and should be green now.

mrvoorhe:
Looks like a few tests failed with

```
Invalid IL detected in /tmp/linker_tests/output/library.dll
* Assertion at metadata.c:1091, condition `index < meta->heap_blob.size' not met
```

PeVerify doesn't complain on windows so I'm not too worried about it.  Going to add [SkipPeVerify] to these tests

mrvoorhe:
Green now

PR #414, opened 2018-11-29
Add a WebAssembly specific pass to mark methods which directly or indirectly call GetCallerAssembly () with a custom attribute, so the runtime can special case them.

vargaz:


vargaz:
The pass itself will be in the mono repo compiled in only if WASM is defined.

marek-safar:
@vargaz I tweaked the code in https://github.com/mono/linker/pull/418 to allow us to keep all code at one place, let's keep it in Mono for now

PR #415, opened 2018-11-29
Add attributes for asserting changes to method bodies

mrvoorhe:
This PR brings a handful of test framework attributes that we have had in the UnityLinker test framework for some time.  It adds support to the test framework for 

* Asserting instructions
* Asserting exception handlers
* Asserting locals

For each collection there are 2 attributes that can be used.  A sequence attribute which can be used for explicit checking.  These can become tedious to maintain and so you generally don't want to use them in any test that is concerned with testing something different.  For these cases there is a general, "expect the value to be different" attribute that you can use.

Also updates some ETW tests to take advantage of these new attributes.

Added new ETW tests for checking locals and exception handlers

### DIFF
--- a/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/ExpectBodyModifiedAttribute.cs
+++ b/linker/Tests/Mono.Linker.Tests.Cases.Expectations/Assertions/ExpectBodyModifiedAttribute.cs
@@ -1,0 +1,7 @@
+using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Assertions {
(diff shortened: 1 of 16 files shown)


mrvoorhe:
Pushed a tweak to fix the test `StubbedMethodWithExceptionHandlers` which passed locally for me because I was on windows where we use `csc` and on the CI we use `mcs.  `mcs` added a local for a `catch (Exception e)` where `e` was not used, while `csc` did not introduce a local.  I just removed the `(Exception e)` since it didn't matter for the sake of the test.

marek-safar:
build

mrvoorhe:
Rebased and fixed the newly added ETW test.  It failed with these changes

PR #416, opened 2018-11-29
Don't stub ETW methods which are explicitly marked with user code attribute

marek-safar:


linker/Linker.Steps/RemoveFeaturesStep.cs
```diff
@@ -86,8 +86,24 @@ void ExcludeEventSourceImplementation (TypeDefinition type)
...
+				if (method.HasCustomAttributes) {
+					if (!method.IsPrivate) {
+						foreach (var attr in method.CustomAttributes) {
+							//
+							// [NonEvent] attribute is commonly used to mark code which calls
+							// IsEnabled we could check for that as well to be more aggresive
+							// but for now I haven't seen such code in wild
+							//
+							if (BCL.EventTracingForWindows.IsNonEventAtribute (attr.AttributeType)) {
+								skip = true;
```
mrvoorhe: If a method has `[NonEvent]` on it, do you really want to clear all the CustomAttributes on it?

marek-safar: Yes, we are killing all ETW feature bits but we don't want to crash on entry to the feature

mrvoorhe: Ok

PR #419, opened 2018-11-30
Support for detecting simple Activator.CreateInstance usages

mrvoorhe:
There are two aspects to CreateInstance detection.

1) Detect the exact type being created.

This is possible in 3 flavors.
```
CreateInstance<Foo>();
CreateInstance(typeof(Foo));
CreateInstance("assembly", "Foo");
```

A caveat to this aspect is that we can only detect the simple overloads of CreateInstance.  Anything that takes additional parameters beyond the type information cannot be detected without more complex stack evaluation logic.

2) Detect the cast type after CreateInstance

If we can't figure out the exact type, then try to figure out the type from the cast.
When we detect cast information, we will mark the ctors of all marked types that are the cast type or are derived from the cast type.

Note that types that are not marked for some other reason will *NOT* be marked by this mechanism.

Example usages
`CreateInstance(localVariable) as Foo;`

## Other Notes ##

* I've left some doors open for our UnityLinker to override the MarkStep's implementation and supply our own.  We have stack evaluation tools and we could add logic to detect more cases if we thought it would be useful.  You'll notice some of the `protected virtual` methods have unused arguments, these extract arguments give us the info we need.

* Tracing support is implemented.

* Tracing during the cast scenario will record the original method that made the connection even for derived types that are processed before/after the body with the CreateInstance.


linker/Linker.Steps/MarkStep.cs
```diff
@@ -2053,6 +2086,142 @@ bool CheckReflectionMethod (Instruction instruction, string reflectionMethod)
...
+		/// analysis to figure out the type
+		/// </summary>
+		/// <param name="body"></param>
+		/// <param name="createInstanceMethod"></param>
+		/// <param name="callInstruction"></param>
+		/// <param name="variation"></param>
+		/// <param name="ctorUsage"></param>
+		/// <param name="information"></param>
+		/// <returns></returns>
+		protected virtual bool TryProcessCastTypeOfCreateInstanceCall (MethodBody body, MethodReference createInstanceMethod, Instruction callInstruction,
```
marek-safar: I am sure I understand this fully, what pattern is this trying to match

mrvoorhe:
Fixed test failure.  For some reason there is no `Activator.CreateInstance(ActivationContext)` overload in mono profile.

mrvoorhe:
This PR has a compilation conflict with https://github.com/mono/linker/pull/421 that I don't believe is going to cause a merge conflict.  You can merge either PR first, but don't merge the second one until I've had a chance to rebase and fix up whichever PR goes second.

mrvoorhe:
@marek-safar I rebased and fixed the compile error.  This PR is ready.

linker/Linker.Steps/MarkStep.cs
```diff
@@ -2231,6 +2504,21 @@ public AttributeProviderPair (CustomAttribute attribute, ICustomAttributeProvide
...
+			{
+				CallingBody = callingBody;
+				CastType = castType;
+				ConstructorCollector = constructorCollector;
+				DefaultCtorOnly = defaultCtorOnly;
+			}
+
+			public MethodBody CallingBody { get; private set; }
+			public TypeDefinition CastType { get; private set; }
+			public Func<TypeDefinition, bool, MethodDefinition[]> ConstructorCollector { get; private set; }
```
marek-safar: Is it necessary to have this customizable for each instance?

linker/Linker.Steps/MarkStep.cs
```diff
@@ -46,6 +46,7 @@ public partial class MarkStep : IStep {
 		protected Queue<AttributeProviderPair> _assemblyLevelAttributes;
 		protected Queue<AttributeProviderPair> _lateMarkedAttributes;
 		protected List<TypeDefinition> _typesWithInterfaces;
+		protected List<ActivatorCreateInstanceMarkingInformation> _activatorCreateInstanceTypes;
```
marek-safar: Why is it not HashSet?

linker/Linker.Steps/MarkStep.cs
```diff
@@ -2053,6 +2086,246 @@ bool CheckReflectionMethod (Instruction instruction, string reflectionMethod)
...
+			if (nextInstruction.OpCode.Code != Code.Isinst && nextInstruction.OpCode.Code != Code.Castclass)
+				return false;
+			
+			var instanceBeingCastedToType = nextInstruction.Operand as TypeReference;
+			if (instanceBeingCastedToType == null)
+				return false;
+
+			var castType = BCL.Activator.ResolveCastType (instanceBeingCastedToType);
+
+			if (castType == null)
```
marek-safar: Please add a test castType is a class

linker/Linker.Steps/MarkStep.cs
```diff
@@ -217,6 +220,35 @@ void ProcessVirtualMethods ()
 			}
 		}
 
+		void ProcessActivatorCreateInstanceType ()
+		{
+			var markedTypes = Annotations.GetMarkedTypes ();
+			foreach (var information in _activatorCreateInstanceTypes) {
```
marek-safar: Where is _activatorCreateInstanceTypes cleared?

linker/Linker.Steps/MarkStep.cs
```diff
@@ -217,6 +220,35 @@ void ProcessVirtualMethods ()
...
+				_context.Tracer.Push (information.CallingBody.Method);
+				try {
+					foreach (var markedType in markedTypes) {
```
marek-safar: Why do you have to process all markedTypes when you already have a possible candidate in CastType?

linker/Linker.Steps/MarkStep.cs
mrvoorhe: Because two `Activator.CreateInstance` usages related to the same type could cause different marking behavior.  The first usage may only mark the `.ctor()` while a later usage may require all `.ctor` be marked.

See the test `CreateAndThenCastAfterAnyCtorSameType` https://github.com/mono/linker/pull/419/files#diff-a25cda93d2795e5473fb811034eda4f4

Another similar scenario is that in our UnityLinker we can detect all stack variables types of a `Activator.CreateInstance` usage and are able to register a `ActivatorCreateInstanceMarkingInformation.ConstructorCollector ` method that will precisely mark just the ctors that are needed.  This is a hypothetical example, but we could do it if a worth while scenario arose and if I were to do it I'd rather not have to debug and figure out why this code failed me.

I could have implemented custom Equals, GetHashCode, ==, etc on `ActivatorCreateInstanceMarkingInformation` and then used a HashSet, but given the following I decided that wasn't worth the effort.

* `ActivatorCreateInstanceMarkingInformation` has a number of properties to consider in determining if two `Activator.CreateInstance` usages should be considered equal.

* Maintaining the overrides for comparing can be error prone.

* I'm assuming the number of detectable `Activator.CreateInstance` usages that we add to this list will be relatively low.  Maybe that's not a fair assumption?

linker/Linker.Steps/MarkStep.cs
mrvoorhe: Yes.  Each `Activator.CreateInstance` could trigger 1 of a number of marking scenarios.

1) We know only the default ctor needs to be marked.

2) We you are create some instance of a type, but we have no idea which ctor you are trying to use, so mark them all.

3) We know you want a default ctor, but only if it's public or non-public (I didn't implement this distinction because it didn't seem worth the effort, but if a need ever arose this plumbing would be ready).  This scenario is encountered by the `CreateInstance(Type type, bool nonPublic)` overload

4) When you combine UnityLinker's stack evaluation with the more complex overloads of `CreateInstance` that let you specify `BindingFlags` and pass the ctor parameters, it's conceivable that if we had a very targeted case that we wanted to detect we could go so far as to match the types in the ctor params object[] array with the parameters of the ctor's on the type and only mark the ctor(s) that full fill the same signature.

1 & 2 are implemented in this PR.  3 & 4 may go unimplemented until a need arises, but I'd like to have the paradigm that is in place be ready to support these scenarios so that I (or someone else) doesn't have to revisit this

linker/Linker.Steps/MarkStep.cs
mrvoorhe: It's not and it's can't be.

Once an `Activator.CreateInstance` usage is registered in `_activatorCreateInstanceTypes` we need to reprocess the marking that usage triggers on each pass of the primary queue.  There are a number of queue phases that could cause new types to be marked.  If a new type is marked and meets the requirements of any of the `Activator.CreateInstance` usages then we'll need to mark that types ctor's accordingly.

The following happen after `ProcessActivatorCreateInstanceType` and could easily lead to new types being marked.
```
ProcessVirtualMethods ();
ProcessMarkedTypesWithInterfaces ();
DoAdditionalProcessing ();
```

and in more obscure cases `ProcessLazyAttributes()` and `ProcessLateMarkedAttributes()` could probably trigger a new type to be marked.

linker/Linker.Steps/MarkStep.cs
mrvoorhe: In the cast scenario, the `CastType` is only 1 of N possible types that we need to mark.  We also need to apply the same marking logic to all other marked types derived from `CastType`.  We need to do this because in the cast scenario we don't actually know what type is being created.

See the test `UnusedDerivedTypesAreNotMarked`  https://github.com/mono/linker/pull/419/files#diff-b0f8eca4ab2e651d4eda74da6a827d1b

linker/Linker.Steps/MarkStep.cs
mrvoorhe: Done.  Now looks like

```
			if (castType == null || castType.IsValueType)
				return false;
```

linker/Linker.Steps/MarkStep.cs
mrvoorhe: Another approach I could have taken would have been to store information about derived types in `AnnotationsStore` during the `TypeMapStep`.  Then I could have done something like
```
foreach (var derivedType in Annotations.TypesDerivedFrom(information.CastType))
    if (Annotations.IsMark(derivedType))
        ....
```

I didn't take that approach because

* It would be more work
* It would be more complicated.
* Back to my assumption again (which could be a bad assumption to make) that the number of items in `_activatorCreateInstanceTypes` will be relatively small so I went with simple over optimal.

linker/Linker.Steps/MarkStep.cs
marek-safar: Use `IsClass` as that's the only case the code can handle

linker/Linker.Steps/MarkStep.cs
marek-safar: Please don't make the code over engineered when the need comes we can refactor it but custom ConstructorCollector for each instance seems really unnecessary for this change

linker/Linker.Steps/MarkStep.cs
```diff
@@ -2071,6 +2107,246 @@ bool CheckReflectionMethod (Instruction instruction, string reflectionMethod)
...
+			if (instanceBeingCastedToType == null)
+				return false;
+
+			var castType = BCL.Activator.ResolveCastType (instanceBeingCastedToType);
+
+			if (castType == null || castType.IsValueType)
+				return false;
+
+			var defaultCtorOnly = method.Parameters.Count == 1 || (method.Parameters.Count == 2 && method.Parameters [1].ParameterType.MetadataType == MetadataType.Boolean);
+			var information = new ActivatorCreateInstanceMarkingInformation (
```
marek-safar: Why don't use the same pattern as elsewhere and call `MarkDefaultConstructor (castType);` only for defaultCtor?

linker/Linker.Steps/MarkStep.cs
marek-safar: I don't like this approach where we could iterate all the types over and over. If I understood the code this is handling scenario where we don't know which ctor to mark exactly.

Why don't we, in this case, build the dictionary of such types and process them at the end of `while (!QueueIsEmpty ()) {` and mark all their unmarked ctors then we clean the dictionary and be done for the pass.

linker/Linker.Steps/MarkStep.cs
mrvoorhe: Switching to `IsClass` leads to the interface tests failing, which makes sense because interfaces are not classes and the code can handle interfaces.

linker/Linker.Steps/MarkStep.cs
mrvoorhe: I don't think you are understanding why it's working this way.

Once we detect an `Activator.CreateInstance` usage we need to keep track of that for the entire duration of the MarkStep.  Think of each `Activator.CreateInstance` usage as creating a new semi-dynamic mark rule that the MarkStep needs to apply to all currently marked types and to all types that are marked later in the MrakStep.  We cannot clear this list of "dynamic mark logic".

Also when we detect an `Activator.CreateInstance` usage we need to retroactively apply the marking logic to all types that have already been marked before we came across this usage.  So I don't see a way to avoid the iterate all types process unless we build up a look up table during the TypeMapStep.  Building such a table seems costly in it's own right and you'd force every run to pay that price even if that set of assemblies makes no use of `Activator.CreateInstance`

What I could maybe do is 

* Process all marked types at the time the `Activator.CreateInstance` usage is detected.
* Then iterate `_activatorCreateInstanceTypes` during `MarkType`

Then we only pay the iterate all marked types cost once for each usage.  Would you prefer that?  Seems to me whether or not it will be less costly than the current approach depends on how usages are detected.  The current approach can batch multiple usages into one pass of all marked types.

linker/Linker.Steps/MarkStep.cs
mrvoorhe: We could call `MarkDefaultConstructor (castType);`, but we'd still need to add a new `ActivatorCreateInstanceMarkingInformation` to `_activatorCreateInstanceTypes` so that

* We mark the default ctor of all derived types that have already been marked.
* We mark the default ctor of all derived types that will be marked later.

My thinking was, why special case the default ctor scenario?  Just let it work one way.

linker/Linker.Steps/MarkStep.cs
mrvoorhe: My goal isn't to over engineer, it's to leave a door open.  Each time I need to make a change upstream it leads to a series of hoops I have to jump through on our end before I can complete w/e it is I set out to do.  If I might want access to something in the future, it's simply easier to leave myself an open door.

I haven't looked at how this code plays out on a real example yet and it's probably a good idea to give it a try. I linked a simple Unity project with these changes and here is what `_activatorCreateInstanceTypes` looked like each time `ProcessActivatorCreateInstanceType` was called.

```
Call 1 - _activatorCreateInstanceTypes.Count = 3
Call 2 - _activatorCreateInstanceTypes.Count = 3
Call 3 - _activatorCreateInstanceTypes.Count = 5
Call 4 - _activatorCreateInstanceTypes.Count = 6
Call 5 - _activatorCreateInstanceTypes.Count = 7
Call 6 - _activatorCreateInstanceTypes.Count = 8
Call 7 - _activatorCreateInstanceTypes.Count = 8
Call 8 - _activatorCreateInstanceTypes.Count = 8
Call 9 - _activatorCreateInstanceTypes.Count = 8
Call 10 - _activatorCreateInstanceTypes.Count = 8
```

I don't see the worry with the 10 `ConstructorCollector` Func's that are created. 

I would like to leave a door open so that I don't have to redo this dance again in the future.  An alternative I see that keeps the door open is the Func property changes to a bool property named `DefaultCtorOnly`, but 1 or 2 new methods are added for us to hook into the creation of and processing of an `ActivatorCreateInstanceMarkingInformation`

Would you prefer that approach?

linker/Linker.Steps/MarkStep.cs
marek-safar: Didn't notice that so let's add || IsInterface

mrvoorhe:
@marek-safar brought up a good point, we shouldn't do any speculative guessing about what is needed.  This guessing does indeed cause a size increase, especially in Mono.Security.dll.

I'm going to close this PR for the time being.  At some point I will refactor it to just cover the creation type and maybe the cast type if the cast type is sealed.

Issue #59, opened 2017-04-04
Support a response file


ericstj:
Currently there is support to pass in a file for certain parameters but a pattern followed by most .NET tools is to pass in `@foo.rsp` that has all the commandline options verbatim as if they were passed on the commandline.  This is convenient for rerun and also handles command-line length limitations on various OSes.

sbomer:
The ILLink.Tasks tooling runs in-proc. Thanks!

PR #424, opened 2018-12-11
Fix issue #59

cshung:


linker/Linker/Driver.cs
```diff
@@ -72,7 +74,65 @@ public static int Execute (string[] args, ILogger customLogger = null)
...
+		Queue<String> ProcessResponseFile (string [] args)
+		{
+			const char NUL = '\0';
+			var result = new Queue<string> ();
+			foreach (string arg in args) {
+				if (arg.StartsWith ("@")) {
+					string responseFileName = arg.Substring (1);
+					try {
+						IEnumerable<string> responseFileLines = File.ReadAllLines (responseFileName).Select (line => line.Trim ());
+						string responseFileText = string.Join (" ", responseFileLines) + NUL;
```
mrvoorhe: I'd rather see length based parsing rather then tacking on null character and then assuming the contents of the arguments have no null characters.

linker/Linker/Driver.cs
```diff
@@ -72,7 +74,65 @@ public static int Execute (string[] args, ILogger customLogger = null)
 
 		public Driver (string [] args)
 		{
-			_queue = new Queue<string> (args);
+			_queue = ProcessResponseFile (args);
+		}
+
+		Queue<String> ProcessResponseFile (string [] args)
```
mrvoorhe: Can you make this public static and then add some unit tests?

cshung:
I am on it

linker/Linker/Driver.cs
cshung: Fixed

linker/Linker/Driver.cs
cshung: Fixed

linker/Tests/Tests/ParseResponseFileLinesTests.cs
```diff
@@ -0,0 +1,63 @@
...
+			Assert.That (TestParseResponseFileLines (@"""x \"" y""", new string [] { @"x "" y" }), Is.True);
+		}
+
+		[Test]
+		public void TestTwoSlashesWithQuote ()
+		{
+			Assert.That (TestParseResponseFileLines (@"""Trailing Slash\\""", new string [] { @"Trailing Slash\" }), Is.True);
+		}
+
+		private bool TestParseResponseFileLines (string v1, string [] v2)
```
mrvoorhe: This can be simplified to

```
		private void TestParseResponseFileLines (string v1, string [] v2)
		{
			var result = new Queue<string> ();
			Driver.ParseResponseFileLines (v1.Split ('\n'), result);
			Assert.That(result, Is.EquivalentTo(v2));
		}
```

And then each [Test] method is reduced down to for example,
`TestParseResponseFileLines (@"abc", new string [] { @"abc" });`

Less code and it will give better error message if a test were to fail.

linker/Tests/Tests/ParseResponseFileLinesTests.cs
```diff
@@ -0,0 +1,63 @@
+using NUnit.Framework;
+using System.Collections.Generic;
+
+namespace Mono.Linker.Tests {
+	[TestFixture]
+	public class ParseResponseFileLinesTests {
```
mrvoorhe: Can you add tests that have Windows and Linux style file paths in the response file contents?  Historically file paths are one of the most common source of command line parsing woes in my experience so having coverage of file paths is reassuring.

cshung: Fixed

linker/Tests/Tests/ParseResponseFileLinesTests.cs
cshung: Fixed

mrvoorhe:
Looks good.  Thanks!

linker/Linker/Driver.cs
```diff
@@ -72,7 +74,71 @@ public static int Execute (string[] args, ILogger customLogger = null)
...
+		{
+			var result = new Queue<string> ();
+			foreach (string arg in args) {
+				if (arg.StartsWith ("@")) {
+					try
+					{
+						string responseFileName = arg.Substring(1);
+						IEnumerable<string> responseFileLines = File.ReadAllLines(responseFileName).Select(line => line.Trim());
```
marek-safar: Please do it without reading the whole file first. Simple `while ((line = responseFileName.ReadLine ()) != null) {` would work better

linker/Linker/Driver.cs
```diff
@@ -72,7 +74,71 @@ public static int Execute (string[] args, ILogger customLogger = null)
...
+						ParseResponseFileLines(responseFileLines, result);
+					} catch (Exception) {
+						Usage ("Cannot read response file");
```
marek-safar: Include the exception message as well

linker/Linker/Driver.cs
marek-safar: Why is it not static?

linker/Linker/Driver.cs
```diff
@@ -72,7 +74,71 @@ public static int Execute (string[] args, ILogger customLogger = null)
...
+				}
+				if  (idx == responseFileText.Length) {
+					break;
+				}
+				StringBuilder argBuilder = new StringBuilder ();
+				bool inquote = false;
+				while  (true) {
+					bool copyChar = true;
+					int numBackslash = 0;
+					while  (idx < responseFileText.Length && responseFileText [idx] == '\\') {
```
marek-safar: Would be nice to keep the formatting consistent, you are using mix of 0, 1 or 2 spaces

marek-safar: Store `responseFileText [idx]` to local variable and use it instead

linker/Linker/Driver.cs
```diff
@@ -72,7 +74,71 @@ public static int Execute (string[] args, ILogger customLogger = null)
...
 		{
-			_queue = new Queue<string> (args);
+			_queue = ProcessResponseFile (args);
+		}
+
+		Queue<String> ProcessResponseFile (string [] args)
+		{
+			var result = new Queue<string> ();
+			foreach (string arg in args) {
+				if (arg.StartsWith ("@")) {
```
marek-safar: This logic is wrong, the response file value should override command line value if specified later

linker/Linker/Driver.cs
```diff
@@ -72,7 +74,71 @@ public static int Execute (string[] args, ILogger customLogger = null)
...
+				StringBuilder argBuilder = new StringBuilder ();
+				bool inquote = false;
+				while  (true) {
+					bool copyChar = true;
+					int numBackslash = 0;
+					while  (idx < responseFileText.Length && responseFileText [idx] == '\\') {
+						numBackslash++;
+						idx++;
+					}
+					if  (idx < responseFileText.Length && responseFileText [idx] == '"') {
```
marek-safar: I don't follow the logic here, are you trying to unquote `"`? Why not do simple one char look ahead?

linker/Linker/Driver.cs
cshung: My `algorithm` for producing the formatting is the following:
Format the code in Visual Studio.
Replace all instances of `" ("` to `"("`. repeat until I can't, so there is no space before any "("
Replace all instances of `"("` to `" ("`, so that each "(" is prefixed with one space.
Do the same with "["

Since the operations are done manually, it is easy to make mistake. It would be nice if there is a formatter or something such that I can use to produce the required formatting.

linker/Linker/Driver.cs
cshung: As of now - the response file content are like macro and get expanded in place. 
It looks like some command line arguments, like `--skip-unresolved`, are meant to be specified once, even though it doesn't really matter if it is specified more than once. Some arguments, such as `-d`, are meant to be specified more than once, and it doesn't make sense to override.

How about just keep the current expansion logic, and if we need to implement the override, we can choose to do that at run?

linker/Linker/Driver.cs
cshung: I am trying to be consistent with existing Microsoft tools that take response file as input. The implementation I have is based on some existing implementation, and here is I believe how the rules are derived:

Windows path use the backslash `\` a lot, it would be silly if we have to escape every backslash, so we don't. By default, `\` simply means the literal backslash.

We wanted to have space in file name as well. Conventionally, we use the double quote `"`, so we quote the whole path, all is good.

Then the question arise - how do we process a literal `"` inside an argument? Easy - just escape it with the backslash as we always do like `\"`.

Now we need to fix the problem of producing a literal `\` right before the ending double quote, which is the common case for argument that represents a directory. A plausible solution would be `\\"`. Conceptually, the first backslash escaped the second, so the remaining is just the delimiting double quote, not a literal double quote.

Inductively, we have the problem of producing a literal `\"`, so we choose `\\\"`. Again, conceptually, the first backslash escaped the second, and the third escaped the last double quote, therefore the last double quote is literal, not delimiting.

Noticing the odd even pattern, we come up with these two simple rules:

* If the sequence of backslash does not end with a double quote, treat them as literal backslash.
* Otherwise, we count the number of backslash, generate half (floored if odd) number of backslash, if it is odd, then generate a literal double quote, otherwise generate a delimiting double quote.

In order to implement the rules above, we need to look ahead until we know what follows the backslash sequence, and therefore cannot be a simple one character look ahead.

Granted, the problem of generating these could be much simplified (for example, simply use the line ending as the argument delimiter or use the Roslyn way, using a pair of double quote to mean a literal double quote. Either way the parser is much simplified. I chose this approach mostly to make sure it is consistent with other Microsoft tools.

linker/Linker/Driver.cs
marek-safar: I'd rather not introduce uncommon behaviour which we cannot change once it lands

cshung: Can you please elaborate on what do you mean by the override logic, maybe with some examples?

cshung: Any updates?

marek-safar: Consider someone passes following arguments `@foo.rsp -c skip` this should overwrite -c setting from foo.rsp

linker/Linker/Driver.cs
cshung: I think it will work in the scenario you are talking about - here is why:

1. Suppose the command line arguments are `@foo.rsp -c skip` and the content of `foo.rsp` is simply `-c copy`

2. The `ProcessResponseFile()` will get a `string[] args = { "@foo.rsp", "-c", "skip" }`

3. The `ProcessResponseFile` loop will get `@foo.rsp`, so it calls `ParseResponseFileLines(new string[]{"-c copy"})`

4. After the call, the queue will be `"-c","copy"`

5. Next, the `ProcessResponseFile` loop will get `"-c"` and then `"skip"`

6. When the loop is done, the final queue would be `{ "-c","copy", "-c", "skip" }`, it would look as if the overridding is not happening, but read on ...

7. The program would then proceed as if someone type in the command line `illink -c copy -c skip`.

8. When it comes to actually executing, the `Driver.run()` will start.

9. The `while (HaveMoreTokens())` loop will start execute, it process the queue in a first-in-first-out manner, which means it will see `'-c'` twice.

10. When the `"-c"` is processed for the second time, the `'c'` switch case would kick in and override the `context.CoreAction` to `skip`.

Because of point 6 - arguably - any overridding logic that need to happen with the presence with the response file will also need to happen in the plain command line case.

I do spot one thing though, suppose we specified `-c AddBypassNGen`, then the `ParseAssemblyAction` would set `_needAddBypassNGenStep`, but if we override it later, it will never set back to `false`, we probably should fix that, independent of the support for response file.

marek-safar: That's was just an example. It needs to work with any order of response files and individual arguments (-c skip @foo.rsp, @foo.rsp -c skip @bar.rsp, etc.) which won't because you process response arguments always first.

cshung: I think the arguments will eventually be processed in the specified order.

Suppose the command line is `A @b.rsp C @d.rsp E` and the content of `b.rsp` and `d.rsp` are simply `B` and `D`, the queue after the call `ProcessResponseFile()` should be `A,B,C,D,E`. In particular, the enqueue operation on line 94 should make sure `A,C,E` are all populated at the right spot.

After the `ProcessResponseFile()` call, the arguments will be processed in the `A,B,C,D,E` order, which I believe is what we wanted.

linker/Linker/Driver.cs
cshung: Will be fixed in the new iteration.

linker/Linker/Driver.cs
cshung: Will be fixed in the new iteration.

linker/Linker/Driver.cs
cshung: Somehow I responded to this comment somewhere else, resolving this one and continue the conversation there instead.

linker/Linker/Driver.cs
cshung: Shall we resolve this?

linker/Linker/Driver.cs
cshung: > Store `responseFileText [idx]` to local variable and use it instead

It appears to me that `responseFileText [idx]` is frequently used in loop conditions and `idx` got changed all the time. I am thinking about not fixing this comment for now. 

If you feel like we must fix it, let me know why and I will work on it in the next iteration.

PR #437, opened 2019-01-07
Preparation for base type sweeping

mrvoorhe:
This change adds plumbing that will be used by a follow on PR to change a type's base to Object when the base type is not needed.  For example, a class with a base and only a simple static method on the class is marked.

There should be no change in behavior as a result of this change.

I choose to PR these changes ahead of the changes that will actually implement support for base type sweeping in order to reduce the size of the implementation PR.

* Build up a type hierarchy table during the TypeMapStep so that a type's base classes are easily accessible for processing during the MarkStep.

* Sweep step will now change the base type of any class that does not need it's base to Object.

* MarkStep will mark all classes as needing their base type.  The follow on PR will remove this behavior and replace it with logic that will actually result in base types being changed.

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -984,6 +984,10 @@ namespace Mono.Linker.Steps {
 				MarkRequirementsForInstantiatedTypes (type);
 			}
 
+			// A future PR will remove this and introduce changes so that we only mark the base type when it's needed.
+			if (type.IsClass)
+				MarkBaseHierarchyAsRequired (type);
+
 			if (type.HasInterfaces)
 				_typesWithInterfaces.Add (type);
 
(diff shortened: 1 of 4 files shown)


linker/Linker.Steps/MarkStep.cs
```diff
@@ -1750,6 +1754,25 @@ protected virtual void ProcessMethod (MethodDefinition method)
 		protected virtual void DoAdditionalMethodProcessing (MethodDefinition method)
 		{
 		}
+		
+		void MarkBaseHierarchyAsRequired (TypeDefinition type)
+		{
+			Annotations.MarkBaseRequired (type);
```
marek-safar: I'm not ok with additional processing which is unused

mrvoorhe: Ok.

PR #438, opened 2019-01-07
Prevent duplicate keys when resolving typerefs

praeclarum:
A hash is used to detect if types have already been resolved. But this cache wasn't consulted for exported typerefs.

This caused duplicate key argument exceptions to be thrown.

I'm not sure if the exported code should be changing the hash since it wasn't used before. But I left it and just added a guard.

### DIFF
--- a/linker/Linker.Steps/SweepStep.cs
+++ b/linker/Linker.Steps/SweepStep.cs
@@ -250,7 +250,8 @@ namespace Mono.Linker.Steps {
 					IMetadataScope scope = et.Scope;
 					if ((td != null) && Annotations.IsMarked (td)) {
 						scope = assembly.MainModule.ImportReference (td).Scope;
-						hash.Add (td, scope);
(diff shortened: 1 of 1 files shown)


linker/Linker.Steps/SweepStep.cs
```diff
@@ -250,7 +250,8 @@ bool ResolveAllTypeReferences (AssemblyDefinition assembly)
 					IMetadataScope scope = et.Scope;
 					if ((td != null) && Annotations.IsMarked (td)) {
 						scope = assembly.MainModule.ImportReference (td).Scope;
-						hash.Add (td, scope);
```
marek-safar: This looks suspicious, please remove the Add completely (it's a no-op anyway)

praeclarum:
I removed the add, but I'm not sure anymore if it's actually a nop. The hash is used on line 264 when scopes are assigned. The logic is a bit hard for me to follow, so I'm not 100% sure.

PR #441, opened 2019-01-08
Add diagnostics message for unknown assembly

kant2002:
When encounter unknown assemble ruring linking phase,
which happens a lot when working with Blazor, previously IlLink
silently fails without indication what cause the error.
Now displayed name of the offending assembly

### DIFF
--- a/linker/Linker/Annotations.cs
+++ b/linker/Linker/Annotations.cs
@@ -87,7 +87,7 @@ namespace Mono.Linker {
 			if (assembly_actions.TryGetValue (assembly, out action))
 				return action;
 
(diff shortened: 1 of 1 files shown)


linker/Linker/Annotations.cs
```diff
@@ -87,7 +87,7 @@ public AssemblyAction GetAction (AssemblyDefinition assembly)
 			if (assembly_actions.TryGetValue (assembly, out action))
 				return action;
 
-			throw new NotSupportedException ();
+			throw new InvalidOperationException($"No action for the assembly {assembly.Name.Name} defined");
```
marek-safar: Use ToString to include full name

```suggestion
			throw new InvalidOperationException($"No action for the assembly {assembly.Name} defined");
```

PR #447, opened 2019-01-20
Add an '--exclude-feature collation' option to link out the unicode collation code. Allow a comma separated list of features for --exclude-feature.

vargaz:


marek-safar:
Working on small integration tweaks but it looks good overall

Issue #452, opened 2019-01-31
Use arcade to build linker


sbomer:
@marek-safar We would like the linker to participate in .NET Core product construction using https://github.com/dotnet/arcade. I'm hoping to do the work in a way that benefits mono as well as the dotnet product construction, and am looking for input on how to best accomplish that.

The arcade infrastructure uses wrapper scripts that download a bootstrap dotnet sdk and use its nuget/msbuild to restore and build a solution in the repo root. It has some conventions around directory structure, package dependencies, flowing dependencies between repos, etc. To what extent are you interested in reusing arcade infrastructure for the monolinker.exe (as opposed to illink and ILLink.Tasks) build? It seems there are options like:
- building with mono vs dotnet SDK
- arcade wrapper scripts vs Makefiles
- mono project jenkins vs Azure DevOps
- nunit vs xunit

I also noticed that https://github.com/mono/mono has a submodule dependency on mono/linker. It looks like there are dependencies on the source files in https://github.com/mono/mono/blob/master/mcs/tools/linker/monolinker.exe.sources and https://github.com/mono/mono/blob/master/mcs/tools/linker-analyzer/illinkanalyzer.exe.sources. Similarly for cecil at https://github.com/mono/mono/blob/c5b88ec4f323f2bdb7c7d0a595ece28dae66579c/mcs/class/Mono.Cecil/Mono.Cecil.dll.sources. I'm not familiar with the mono project's build.
- Are there any restrictions on what changes can be made to this repo's build? (Would you prefer the file layout to stay as-is, or may we move them around?)

I'd also like to use this as an opportunity to look into simplifying the configurations we use to build illink (and possibly monolinker). Currently, Mono.Linker/illink_Debug depends on Mono.Cecil/netstandard_Debug. The .NET SDK doesn't support configuration-specific dependencies, so we have hacks in place that explicitly inject the configuration into Mono.Linker and its dependencies. The workaround isn't so bad, but it gets worse every time we add another project (Mono.Cecil.Pdb/Mdb, an ILLink.Tasks.Tests project, etc.).
[...]

PR #453, opened 2019-01-31
[WIP] Move linker to arcade

sbomer:
I'm taking over where @cshung left off in the effort to build the linker using [arcade](https://github.com/dotnet/arcade). Thanks for the work @cshung!

I wanted to put out the WIP PR to get early feedback on the planned change. At this point most of what I've done is investigation. I've tried to summarize some of the open questions in https://github.com/mono/linker/issues/452.


mrvoorhe:
Indeed it does look like only the easiest tests were converted to xunit.  How would it look to convert https://github.com/Unity-Technologies/linker/blob/master/linker/Tests/TestCases/TestSuites.cs to xunit?

I'II need to give this more thought.

mrvoorhe:
I was thinking about this more.  Please break up and land the changes as independently as possible.  Ideally the 2nd to last PR to land would be the csproj changes and the final PR would be the switch to xunit.

I might be able to hack together something that let's us pull the csproj change into Unity's fork, but once the xunit change lands Unity will stop staying in sync with upstream.  It could take us weeks or months to make the changes in our infrastructure that would be needed to pull in the xunit change.  The more changes that land before the xunit switch, the more changes I can pull in and reduce the chances of cherry-pick problems bringing our contributions to a halt.

For example, land any file renaming, or directory restructuring by themselves first.

In an ideal world, this effort would slow down and give us a chance to prepare so that when the csproj and xunit changes come, we are better prepared to adopt both.

sbomer:
@mrvorhee, understood, thanks for the feedback! I'll focus on getting our build set up with arcade before trying to move tests to xunit. It's important to me that we retain the ability to collaborate with unity folks, and I'll do what I can to enable that.

sbomer:
I've made some progress getting the repo to build with the new project system, in a way that should also let the mono build continue to work. Now that I have a better idea of what that looks like, I'll be trying to make some independent changes that move in that direction, as you suggested. The first change is the directory restructuring: https://github.com/mono/linker/pull/466.

PR #454, opened 2019-02-01
Always mark Delegate types as instantiated

mrvoorhe:
Delegate and MulticastDelegate are instantiated from native code which causes us to not record that instances of these types could exist.  As a result, today we remove the interfaces from Delegate.  Removing the interfaces doesn't seem to cause any issues today, but it's probably not the greatest idea.

In the future with base sweeping or uninstantiated type method stubbing, this lack of recording these types as instance possible causes bigger problems.

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -982,6 +982,8 @@ namespace Mono.Linker.Steps {
 				//    and there are no other usages of that interface type, then we need to make sure the interface type is still marked because
 				//    this type is going to retain the interface implementation
 				MarkRequirementsForInstantiatedTypes (type);
+			} else if (AlwaysMarkTypeAsInstantiated (type)) {
+				MarkRequirementsForInstantiatedTypes (type);
 			}
 
 			if (type.HasInterfaces)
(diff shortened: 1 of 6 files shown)


linker/Linker.Steps/MarkStep.cs
```diff
@@ -1401,6 +1403,20 @@ static bool IsMulticastDelegate (TypeDefinition td)
 			return td.BaseType != null && td.BaseType.FullName == "System.MulticastDelegate";
 		}
 
+		protected virtual bool AlwaysMarkTypeAsInstantiated (TypeDefinition td)
+		{
+			switch (td.FullName) {
```
marek-safar: FullName is expensive property, use name + namespace instead

marek-safar:
build

linker/Tests/Mono.Linker.Tests.Cases/CoreLink/DelegateAndMulticastDelegateKeepInstantiatedReqs.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Runtime.Serialization;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.CoreLink {
+	/// <summary>
+	/// Delegate and is created from 
+	/// </summary>
+	[SetupLinkerCoreAction ("link")]
```
marek-safar: This will be problematic as the mscorlib will be used for execution.

mrvoorhe: I don’t follow.  Are you saying that it’s bad that this test is dependent on a particular mscorlib?

marek-safar: I am saying it won't be easy to run the test

PR #456, opened 2019-02-01
Override removal

mrvoorhe:
If no instance of a type can exist, then it's safe to not mark overrides as long as the method being overridden is not abstract.

This change should be good to go.  I've added `WIP` as I'm still running these changes through our il2cpp test suite as that is a valuable way to get some extra coverage on linker changes.

Do not land this branch before https://github.com/mono/linker/pull/454 otherwise there will be problems.

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -262,9 +262,14 @@ namespace Mono.Linker.Steps {
 			if (Annotations.IsMarked (method))
 				return;
 
+			var isInstantiated = Annotations.IsInstantiated (method.DeclaringType);
+
 			// We don't need to mark overrides until it is possible that the type could be instantiated
 			// Note : The base type is interface check should be removed once we have base type sweeping
-			if (!Annotations.IsInstantiated (method.DeclaringType) && @base.DeclaringType.IsInterface)
+			if (!isInstantiated && @base.DeclaringType.IsInterface)
+				return;
+
+			if (!isInstantiated && !@base.IsAbstract)
 				return;
 
 			MarkMethod (method);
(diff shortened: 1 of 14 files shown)


mrvoorhe:
@marek-safar FYI for perspective this PR is not a particularly big size for a `Console.WriteLine("Hello world")` program.  IT only saves `0.06%`.  It does better in other scenarios, for example, in the artificial `typeof(System.Xml.XmlDocument)` scenario in https://github.com/mono/linker/pull/455 this PR is able to get the same size savings.

mrvoorhe:
Fixed the test failure.  `EmbeddedLinkXmlPreservesAdditionalAssemblyWithOverriddenMethod` needed to be updated to mark a `.ctor` on some types.

I still need to check on the il2cpp test results w/ this change before I drop the `WIP`

mrvoorhe:
@marek-safar This PR is ready for final review.  After preserving the `.ctor()` of `InternalThread` our il2cpp test suites were green.  I also preserved the `.ctor()` of `MonoListItem` because looking at our mono runtime code I found that it should be, but that one doesn't seem to have been as critical as the `InternalThread`.

Do you want me to open a PR to mono master adding either of these `.ctor()` preserves to mscorlib.xml?  You had seemed to imply that `InternalThread` has been renamed?

mrvoorhe:
@marek-safar This PR is ready for final review.  After preserving the `.ctor()` of `InternalThread` our il2cpp test suites were green.  I also preserved the `.ctor()` of `MonoListItem` because looking at our mono runtime code I found that it should be, but that one doesn't seem to have been as critical as the `InternalThread`.

Do you want me to open a PR to mono master adding either of these `.ctor()` preserves to mscorlib.xml?  You had seemed to imply that `InternalThread` has been renamed?

mrvoorhe:
Looks like after rebasing and picking up the Finalize() fix I have a couple test failures.  I will take a look.

mrvoorhe:
Easy fix.  Needed to update the `NeverInstantiatedTypeWithOverridesFromObject` tests to no longer assert that the overrides from System.Object are kept 😄

PR #457, opened 2019-02-04
Don't mess with InternalThread

mrvoorhe:
Same situation as with Delegate and MulticastDelegate.

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -1412,6 +1412,9 @@ namespace Mono.Linker.Steps {
 				case "Delegate":
 				case "MulticastDelegate":
 					return td.Namespace == "System";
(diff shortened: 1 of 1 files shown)


linker/Linker.Steps/MarkStep.cs
```diff
@@ -1412,6 +1412,9 @@ protected virtual bool AlwaysMarkTypeAsInstantiated (TypeDefinition td)
 				case "Delegate":
 				case "MulticastDelegate":
 					return td.Namespace == "System";
+				// InternalThread doesn't have a ctor that is marked, but it is a very important type not to mess with!
```
marek-safar: I don't like this hardcoding of implementation details. If the linker removes it correctly the fix needs to be elsewhere

mrvoorhe:
I'm closing this PR and going to fix the issue the xml route.

PR #461, opened 2019-02-07
Fix bug with Finalize() being removed

mrvoorhe:
The CoreLink version of `InstantiatedTypeWithOverridesFromObject` is the test that triggers the bug. 

Up until this PR, the way Finalize() methods would be marked was via `IsVirtualAndHasPreservedParent`.  Which is to say, once System.Object.Finalize() was marked, all the other FInalize() methods would be marked.  However, given that Finalize() is never explicitly called from managed code, the only way I can think of that Object.FInalize would be marked is via link xml.

With this PR, once the we deem that an instance of a type could exist we will mark the Finalize() method.

### DIFF
--- a/linker/Linker.Steps/MarkStep.cs
+++ b/linker/Linker.Steps/MarkStep.cs
@@ -1790,7 +1790,14 @@ namespace Mono.Linker.Steps {
 			Annotations.MarkInstantiated (type);
 
 			MarkInterfaceImplementations (type);
-			MarkMethodsIf (type.Methods, IsVirtualAndHasPreservedParent);
+
+			foreach (var method in type.Methods) {
+				if (method.IsFinalizer ())
+					MarkMethod (method);
+				else if (IsVirtualAndHasPreservedParent (method))
+					MarkMethod (method);
+			}
+
 			DoAdditionalInstantiatedTypeProcessing (type);
 		}
 
--- a/linker/Tests/Mono.Linker.Tests.Cases/Basic/InstantiatedTypeWithOverridesFromObject.cs
+++ b/linker/Tests/Mono.Linker.Tests.Cases/Basic/InstantiatedTypeWithOverridesFromObject.cs
@@ -1,0 +1,44 @@
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+
+namespace Mono.Linker.Tests.Cases.Basic {
+	public class InstantiatedTypeWithOverridesFromObject {
+		public static void Main ()
+		{
+			var f = new Foo ();
+		}
+
+		[Kept]
+		[KeptMember (".ctor()")]
+		class Foo {
+			[Kept]
+			~Foo ()
+			{
+				// Finalize shouldn't be empty
+				DoCleanupStuff ();
+			}
+
+			[Kept]
+			void DoCleanupStuff ()
+			{
+			}
+
+			[Kept]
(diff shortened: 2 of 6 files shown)


linker/Linker.Steps/MarkStep.cs
```diff
@@ -1790,6 +1790,7 @@ protected virtual void MarkRequirementsForInstantiatedTypes (TypeDefinition type
 			Annotations.MarkInstantiated (type);
 
 			MarkInterfaceImplementations (type);
+			MarkMethodIf (type.Methods, m => m.IsFinalizer ());
```
marek-safar: Could you fold that with the next MarkMethodsIf. For example `m.IsFinalizer () || IsVirtualAndHasPreservedParent(m)`

mrvoorhe: I didn't because I think there is some chance that future improvements will change the behavior of IsVirtualAndHasPreservedParent and I think keeping it out of there would help reduce the chances of accidentally breaking `Finalize()` in the future.

Also, fundamentally, I don't believe `Finalize()` falls into the same category as `IsVirtualAndHasPreservedParent`.

`Finalize()` is a method required by the runtime (technically the GC?) once an instance of the type could exist.

Where as `IsVirtualAndHasPreservedParent` is a blunt tool to keep things that might be needed by other managed code because we don't have the information available today to know for certain if they are needed.  

I've been thinking about more improvements that can be made and one idea that I have that is relatively trivial and would impact `IsVirtualAndHasPreservedParent` but not `Finalize()` is escape analysis.  If a type never escapes the method(s) it is created in, then we can know all of the calls that were made and we wouldn't need to mark everything that `IsVirtualAndHasPreservedParent` marks today.  I realize it's a little more complicated than that and this may never be worth doing, but I bring it up as an example to highlight why I think `Finalize()` handling shouldn't be bundled with `IsVirtualAndHasPreservedParent`

mrvoorhe: Actually, thinking about this further.  `IsVirtualAndHasPreservedParent` should explicitly not mark `Finalize()`

Just because one types Finalize() is marked or because System.Object.Finalize() is marked does not mean that all types need their Finalize() marked.  Only types that are instantiated need their Finalize() marked.

mrvoorhe: > Actually, thinking about this further.  `IsVirtualAndHasPreservedParent` should explicitly not mark `Finalize()`
> 
> Just because one types Finalize() is marked or because System.Object.Finalize() is marked does not mean that all types need their Finalize() marked.  Only types that are instantiated need their Finalize() marked

Scratch that.  I got mixed up.

mrvoorhe: I got `IsVirtualAndHasPreservedParent` and `ProcessOverrides` mixed up, but my feelings remain the same.  `Finalize()` is a different case than `IsVirtualAndHasPreservedParent` and I don't want to mix the two and risk causing bugs in the future.

mrvoorhe: OK.  I adjusted the code so that we do both in 1 pass over `type.Methods`.  However, I switched to a foreach loop w/ if's.  I don't like Finalize being packed on 1 line with `IsVirtualAndHasPreservedParent`, it hides the behavior in my opinion.

PR #466, opened 2019-02-16
Adopt new directory layout

sbomer:
This organizes the source and test projects as follows:
- source projects go in `src/project/projectfile.csproj`
- test projects go in `test/project/projectfile.csproj`

The uniform layout of projects is part of the arcade onboarding (see https://github.com/mono/linker/issues/452).


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "cecil"]
-	path = cecil
+	path = src/cecil
```
marek-safar: This should either be removed (as we discussed) in favour of binary dependency or called `external` to align with namings used for submodules

sbomer: I'll move it to `external/cecil`. We discussed a binary dependency, but I believe settled on keeping the submodule to enable us to get updates more quickly and also avoid redstributing it as a third-party dependency.

PR #470, opened 2019-02-20
Fix invalid IL when inherited interfaces

mrvoorhe:
When processing overrides on a type that is never instantiated, it is only safe to skip marking of an interface method if the interface implementation corresponding to that method is *not* marked *and* no other interface implementations that implement that interface are marked


test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/NoKeptCtorButInterfaceNeeded/NestedInterfaces1.cs
```diff
@@ -0,0 +1,52 @@
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+
+namespace Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType.NoKeptCtorButInterfaceNeeded {
+	public class NestedInterfaces1 {
```
mrvoorhe: @marek-safar fyi - of all the tests I added in this PR this is the only one that failed.

mrvoorhe: `Foo` was having it's `Method()` stripped resulting in invalid IL

PR #471, opened 2019-02-20
Use SDK-style projects

sbomer:
This updates projects to use the new simplified format. ~~It depends on @jbevain's branch that I believe is a WIP: https://github.com/jbevain/cecil/tree/new-sdk.~~

~~This change will not be ready to merge until the cecil change is ready, but I wanted to share what it's going to look like and get early feedback.~~

Part of https://github.com/mono/linker/issues/452.

Edit: now that cecil has been updated to use SDK-style projects, this can move forward. Waiting on https://github.com/mono/linker/pull/489 to avoid merge conflicts in that PR.

PR #474, opened 2019-02-25
Simplify solutions used to build monolinker/illink

sbomer:
Until I can make more progress on SDK-style projects (see https://github.com/mono/linker/pull/471), I'll begin trying to set up a build of illink using arcade and the old-style projects. In this change I'm getting rid of the scattered linker.sln files in favor of monolinker.sln for monolinker, and illink.sln for the illink/ILLink.Tasks build.


sbomer:
build

sbomer:
@marek-safar the Linux - Mono build definition is failing because of the solution rename. What does it take to update the build definition?

marek-safar:
@akoeplinger I think we need to update it on Jenkins, right?

akoeplinger:
should be fixed.

@monojenkins build

PR #477, opened 2019-03-08
Run illink out of process

sbomer:
This changes illink to build as a framework-dependent app. Its runtimeconfig.json and deps.json get published in the package, alongside ILLink.Tasks.dll. The ILLink task will now run the linker in a new process, by implementing MSBuild's ToolTask.

This will make it possible to run illink more easily from the command line, and will give us a little more flexibility in the dependencies of illink (since they no longer have to unify with the dependencies of other tasks in the msbuild process).

### DIFF
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -123,14 +123,12 @@ namespace ILLink.Tasks
 
 			if (RootDescriptorFiles != null) {
 				foreach (var rootFile in RootDescriptorFiles) {
-					args.Append (" -x");
-					args.Append ($" {rootFile.ItemSpec}");
+					args.Append (" -x ").Append (rootFile.ItemSpec);
 				}
 			}
 
 			foreach (var assemblyItem in RootAssemblyNames) {
-				args.Append (" -a");
-				args.Append ($" {assemblyItem.ItemSpec}");
+				args.Append (" -a ").Append (assemblyItem.ItemSpec);
 			}
 
 			HashSet<string> directories = new HashSet<string> ();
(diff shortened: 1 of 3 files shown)


src/ILLink.Tasks/LinkTask.cs
```diff
@@ -102,42 +139,42 @@ public override bool Execute ()
...
+				args.Append (" -out");
+				args.Append ($" {OutputDirectory.ItemSpec}");
 			}
 
 			if (ClearInitLocals) {
-				args.Add ("-s");
+				args.Append (" -s");
 				// Version of ILLink.CustomSteps is passed as a workaround for msbuild issue #3016
-				args.Add ("ILLink.CustomSteps.ClearInitLocalsStep,ILLink.CustomSteps,Version=0.0.0.0:OutputStep");
+				args.Append (" LLink.CustomSteps.ClearInitLocalsStep,ILLink.CustomSteps,Version=0.0.0.0:OutputStep");
```
marek-safar: btw, it would be better to rewrite this as normal linker optimization

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -102,42 +139,42 @@ public override bool Execute ()
 				var dir = Path.GetDirectoryName (assemblyPath);
 				if (!directories.Contains (dir)) {
 					directories.Add (dir);
-					args.Add ("-d");
-					args.Add (dir);
+					args.Append (" -d");
+					args.Append ($" {dir}");
```
marek-safar: It's StringBuilder, you should avoid string concatenation (e.g. `args.Append(" ").Append(dir);` )

PR #479, opened 2019-03-08
Include a fixed GenerateDepsFile task

sbomer:
This works around a bug in the SDK's deps file generation logic: https://github.com/dotnet/sdk/issues/3010. I'm building ILLink.Tasks with a copy of the sources used by the SDK to generate deps files. It's a temporary replacement for the GenerateDepsFile task, with a fix for the issue mentioned above in https://github.com/mono/linker/commit/39a6ae7936cd72acbdb1dd42071ddb4931b6f71a. I intend to remove it as soon as the SDK issue is fixed. I'm doing this mostly to get our integration tests up and running in the interim - I definitely don't want to ship this fix in the SDK.

@nguerrera

### DIFF
--- a/src/ILLink.Tasks/ILLink.Tasks.csproj
+++ b/src/ILLink.Tasks/ILLink.Tasks.csproj
@@ -7,6 +7,7 @@
     <BaseOutputPath>bin/</BaseOutputPath>
     <PackageOutputPath>$(BaseOutputPath)nupkgs</PackageOutputPath>
     <IsPackable>true</IsPackable>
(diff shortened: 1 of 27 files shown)


sbomer:
The SDK is getting a fix once https://github.com/dotnet/sdk/pull/3086 makes it in to master via https://github.com/dotnet/sdk/pull/3113. In the meantime, https://github.com/mono/linker/pull/483 includes a GenerateDepsFile task with my workaround for the purposes of running integration tests.

PR #481, opened 2019-03-08
Write stack traces to stderr

sbomer:
This will show up in the msbuild output with a higher importance by
default, and may also align more closely with the expectations of
command line users.

@marek-safar the other option is to make this the behavior of monolinker as well. What do you think?


marek-safar:
yes, let's do it without #if-def

PR #482, opened 2019-03-08
Remove crossgen targets from package

sbomer:
Equivalent logic is going to be in the R2R tool. The FilterByMetadata line is necessary because it used to be in ILLink.CrossGen.targets.

 /cc @fadimounir


src/ILLink.Tasks/ILLink.Tasks.csproj
```diff
@@ -101,7 +100,7 @@
     <Compile Include="LinkTask.cs" />
     <Compile Include="CheckEmbeddedRootDescriptor.cs" />
     <Compile Include="CompareSizes.cs" />
-    <Compile Include="ComputeCrossgenedAssemblies.cs" />
+    <None Include="ComputeCrossgenedAssemblies.cs" />
```
fadimounir: Can we delete ComputeCrossgenedAssemblies.cs and remove the <None> entry from here?

sbomer: I'd like to keep the code around for now, in case we want to do anything special with crossgen'd assemblies. I included it in None as a reminder to myself to either compile it in or remove it entirely when we have settled on concrete linker default.

fadimounir: I think we can always get it back from the git history if we ever needed to, but it's confusing keeping it around and not compile it

marek-safar: I agree the file can be always found in the history log

sbomer: Removed along with some other files that are no longer needed.

PR #486, opened 2019-03-11
Linker Research (for review / discussion only)

baulig:
Linker changes for https://github.com/mono/mono/pull/13407.


MichalStrehovsky:
This looks similar to what I did for .NET Native last summer (see dotnet/designs#42), except it hardcodes features as integer IDs, making this only usable for framework libraries. It would be nice if this ended up with a design that third party libraries can use too.

marek-safar:
This is not the approach we'll take primary (pending discussion)

PR #487, opened 2019-03-14
Adding Sdk.props to be imported by the dotnet/sdk

fadimounir:
Adding Sdk.props to be imported by the dotnet/sdk

### DIFF
--- a/src/ILLink.Tasks/Sdk/Sdk.props
+++ b/src/ILLink.Tasks/Sdk/Sdk.props
@@ -1,0 +1,16 @@
+<!--
+***********************************************************************************************
+Sdk.props
+
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved.
+***********************************************************************************************
+-->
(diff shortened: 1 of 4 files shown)


sbomer:
LGTM, thanks!

src/ILLink.Tasks/Sdk/Sdk.props
```diff
@@ -0,0 +1,16 @@
...
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+          created a backup copy.  Incorrect changes to this file will make it
+          impossible to load or build your projects from the command-line or the IDE.
+
+Copyright (c) .NET Foundation. All rights reserved.
+***********************************************************************************************
+-->
+<Project ToolsVersion="14.0">
+
+  <UsingTask TaskName="ILLink.Tasks.ILLink" AssemblyFile="$(MSBuildThisFileDirectory)..\tools\netcoreapp2.0\ILLink.Tasks.dll" />
```
nguerrera: This can't be netcoreapp2.0 only. We have to be able to import this in msbuild running on net472.

fadimounir: @sbomer Could you look into what it takes to build the linker for net472? Once we have that, we can edit the logic in this file to import the right task dll based on which runtime msbuild is using.

PR #489, opened 2019-03-18
Enable unit tests on .NET Core

sbomer:
This enables the unit tests to run on .NET Core. CodeDom compilation doesn't exist on core, so I've added the ability to compile in-process with Roslyn APIs. This still uses NUnit, using a recent version of NUnit package references with "dotnet test". This doesn't change how tests run on mono.

Finding the right reference assemblies for compilation is a bit different on .NET Core. As described in https://github.com/dotnet/roslyn/wiki/Runtime-code-generation-using-Roslyn-compilations-in-.NET-Core-App, they can either be found using packages (what the SDK does), or we can just compile against the implementation assemblies. In this change, I'm compiling against implementation assemblies, since that's fairly similar to what the tests do on mono, where they find them in the GAC.

For mono, the common references only included mscorlib. For .NET Core, I'm getting the common references from the "Trusted Platform Assemblies" (TPA) of the test process. As a result, package references of the test project become references during testcase compilation. Explicit references specified via `ReferenceAttribute` in a testcase are looked for alongside the host process's implementations as well (since we have no way to resolve a reference from just the filename since there is no GAC).

Other changes:
- Roslyn outputs the `DebuggableAttribute` by default but mcs does not, so when running on .NET Core, we don't check for the `DebuggableAttribute`.
- Check for core types in `System.Private.CoreLib` instead of `mscorlib`.
- There are some more edge cases that I'm in the process of cleaning up and will update soon.

At the moment, I'm trying to get the tests up and running, and I'm using plenty of compile-time conditions to modify the behavior on core. It would be nice not to have to do this - I haven't given it much thought, but maybe we could provide an abstraction that contains knowledge of the platform libraries, so that we don't have to have a hard dependency on `"mscorlib.dll"` and `"System.Private.CoreLib.dll"` everywhere? @mrvoorhe please let me know if you have any thoughts. Does the approach I'm taking so far seem sane to you?

@marek-safar


mrvoorhe:
> Roslyn outputs the DebuggableAttribute by default but mcs does not, so when running on .NET Core, we don't check for the DebuggableAttribute.

When we run these tests with UnityLinker we override compilation in `TestCaseCompiler` and use roslyn in server mode instead of mcs since it is faster.  So I ran into this issue with the DebuggableAttribute before.  The way I dealt with it for us is a little easier to maintain.  What we do is filter out the DebuggableAttribute when we are asserting.  I think it would be easiest to just add this behavior.  Then you can avoid all of the ifdef's.

To do this, go to `AssemblyChecker.FilterLinkedAttributes` and add `DebuggableAttribute` to the switch statement in that method.

mrvoorhe:
> At the moment, I'm trying to get the tests up and running, and I'm using plenty of compile-time conditions to modify the behavior on core. It would be nice not to have to do this - I haven't given it much thought, but maybe we could provide an abstraction that contains knowledge of the platform libraries, so that we don't have to have a hard dependency on "mscorlib.dll" and "System.Private.CoreLib.dll" everywhere? 

We have a similar, although slightly different, situation in our test suites for unitylinker.  We run a number of unitylinker tests against our very old mono class libraries.  These class libraries a much different than latest mono and so we have a handful of tests that need to assert different things.  Here is an abbreviated example of what we end up doing.

```
    [SetupLinkerCoreAction("link")]
    [Reference("System.Core.dll")]
    [KeptAssembly("System.Core.dll")]
    [KeptTypeInAssembly("mscorlib", typeof(DSACryptoServiceProvider))]
    [KeptTypeInAssembly("mscorlib", typeof(SHA1Managed))]
    [KeptTypeInAssembly("mscorlib", typeof(SHA256Managed))]
    [KeptTypeInAssembly("mscorlib", typeof(SHA1Managed))]
    [KeptTypeInAssembly("mscorlib", typeof(SHA256Managed))]
    [KeptTypeInAssembly("mscorlib", typeof(SHA384Managed))]
    [KeptTypeInAssembly("mscorlib", typeof(SHA512Managed))]
    [KeptTypeInAssembly("mscorlib", typeof(TripleDESCryptoServiceProvider))]
    [KeptTypeInAssembly("System.Core", typeof(AesManaged))]

#if !UNITY
    [KeptTypeInAssembly("System.Core", typeof(AesCryptoServiceProvider))]
#endif

    [SkipPeVerify("System.Core.dll")] // Probably stuff we should fix, but for now lets overlook it
#if NET45
    // Can be removed once this bug is fixed https://bugzilla.xamarin.com/show_bug.cgi?id=58168
    [SkipPeVerify(SkipPeVerifyForToolchian.Pedump)]
#endif
#if UNITY || NET20
    [SkipPeVerify] // All sorts of stuff is invalid, even mscorlib
#endif
```

Looks similar.  Although, there is one notable difference I see.  We generally use this because the types or members we need to assert are different.  Where as with the netcore changes (at least so far) it looks like the types and members are exactly the same and only the expect assembly changes.

I know what I'm about to say is crude and kind of hacky, but what if we just swap out `mscorlib` for `System.Private.CoreLib` when we are asserting?  It looks like all of the test case could remain unchanged and then you could hack `TestCaseAssemblyResolver` to search for `System.Private.CoreLib` instead of `mscorlib` when `NETCOREAPP` is set.

For the few (if any) tests that need to reference assemblies beyond `mscorlib/System.Private.CoreLib` we could use the `#if` in the test cases approach.  If that ever becomes unwieldy we can revisit when that time comes.

mrvoorhe:
Can you PR the typo fixes that have their own commit separately?  If any are already mixed with real changes don't worry about it.

Some of the typo fixes are to methods that we override.  Having them as separate commits will let me pull in the changes and update our API's with reduced risk.

test/Mono.Linker.Tests/TestCasesRunner/TestCaseMetadaProvider.cs
```diff
@@ -101,7 +140,15 @@ public virtual IEnumerable<SourceAndDestinationPair> GetResponseFiles ()
 
 		public virtual IEnumerable<NPath> GetExtraLinkerSearchDirectories ()
 		{
+#if NETCOREAPP
+			var tpaDirs = GetTrustedPlatformAssemblies ().Select (p => Path.GetDirectoryName (p)).Distinct ();
+			foreach (var dir in tpaDirs)
+			{
```
mrvoorhe: Formatting

mrvoorhe:
Overall looks good and relatively painless.

mrvoorhe:
What do I need to install on Windows to open `illink.sln` and run the tests?

I installed
* Visual Studio 2019 Preview
* https://dotnet.microsoft.com/download/thank-you/dotnet-sdk-3.0.100-preview3-windows-x64-installer

I opened `illink.sln` with VS 2019 and it's hung on project 8 of 9.  Stuck at `Loading project file ....\ILLinker.CustomSteps.csproj`

I tried Rider and it's complaining about `The current .NET SDK does not support targeting .NET Core 3.0...`

test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
```diff
@@ -615,6 +615,12 @@ protected virtual IEnumerable<string> FilterLinkedAttributes (ICustomAttributePr
 					case "System.Diagnostics.DebuggerBrowsableAttribute":
 						continue;
 
+#if NETCOREAPP
```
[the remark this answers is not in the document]
sbomer: IIRC the issue is that this will prevent looking for the attribute in the linked assembly, but testcases that specify it as a KeptAttribute were still failing because it was included in GetExpectedAttributes.

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -58,7 +58,10 @@ public virtual void Check (LinkedTestCaseResult linkResult)
 
 				VerifyLinkingOfOtherAssemblies (original);
 
+#if !NETCOREAPP
+				// the PE Verifier does not know how to resolve .NET Core assemblies.
```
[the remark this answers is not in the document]
sbomer: Agreed, it's too bad. Looks like ILVerify might be a suitable replacement at some point in the future: https://github.com/dotnet/corert/tree/master/src/ILVerify.

test/Mono.Linker.Tests.Cases/Attributes/Mcs/OnlyTypeUsedInAssemblyIsTypeOnAttributeCtorOnAssembly.cs
```diff
@@ -9,6 +9,9 @@ namespace Mono.Linker.Tests.Cases.Attributes.Mcs {
 	/// In the case of attributes on assemblies, we expect both assemblies to be removed because we don't keep assembly level attributes
 	/// when that is the only type marked in the assembly
 	/// </summary>
+#if NETCOREAPP
```
marek-safar: You should exclude whole /Mcs folder conditionally with illink in csproj file

test/Mono.Linker.Tests.Cases/Attributes.Debugger/DebuggerDisplayAttributeOnAssemblyUsingTarget.cs
```diff
@@ -13,8 +13,13 @@ namespace Mono.Linker.Tests.Cases.Attributes.Debugger {
 	// Can be removed once this bug is fixed https://bugzilla.xamarin.com/show_bug.cgi?id=58168
 	[SkipPeVerify (SkipPeVerifyForToolchian.Pedump)]
 	
+#if NETCOREAPP
+	[KeptMemberInAssembly ("System.Private.CoreLib.dll", typeof (DebuggerDisplayAttribute), ".ctor(System.String)")]
```
marek-safar: Add Corlib constant instead of such #if(s)

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
```diff
@@ -185,6 +191,101 @@ protected NPath CompileAssembly (CompilerOptions options)
 
 		protected virtual NPath CompileCSharpAssemblyWithDefaultCompiler (CompilerOptions options)
 		{
+#if NETCOREAPP
```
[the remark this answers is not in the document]
sbomer: I added the in-process compiler mostly for two reasons:
- to match the default behavior of codedom
- because running csc in its own process isn't a common .NET Core scenario. In fact I wasn't sure if we even supported it, until I poked around and saw that we have a csc.dll in our SDK directory that can be run via "dotnet /path/to/csc.dll". The good news is that this dll has a runtimeconfig that specifies server GC!

I'll try to get these tests working with the out-of-process (+server GC) approach and see what the perf is like. To make it a fair comparison, I'll try using server GC for the in-process version as well.
Thanks!

test/Mono.Linker.Tests.Cases/References/AssemblyOnlyUsedByUsingWithMcs.cs
```diff
@@ -3,9 +3,12 @@
 using Mono.Linker.Tests.Cases.References.Dependencies;
 
 namespace Mono.Linker.Tests.Cases.References {
+#if NETCOREAPP
```
marek-safar: Please move such tests under /Mcs

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
```diff
@@ -185,6 +191,101 @@ protected NPath CompileAssembly (CompilerOptions options)
...
+
+#if NETCOREAPP
+		protected virtual NPath CompileCSharpAssemblyWithRoslyn (CompilerOptions options)
+		{
+			var parseOptions = new CSharpParseOptions (preprocessorSymbols: options.Defines);
+			var compilationOptions = new CSharpCompilationOptions (
+				outputKind: options.OutputPath.FileName.EndsWith (".exe") ? OutputKind.ConsoleApplication : OutputKind.DynamicallyLinkedLibrary
+			);
+			// Default debug info format for the current platform.
+			DebugInformationFormat debugType = RuntimeInformation.IsOSPlatform (OSPlatform.Windows) ? DebugInformationFormat.Pdb : DebugInformationFormat.PortablePdb;
```
marek-safar: Is this correct? I thought core uses ppdb everywhere?

test/Mono.Linker.Tests/TestCasesRunner/TestCaseMetadaProvider.cs
```diff
@@ -55,23 +55,62 @@ public virtual TestCaseLinkerOptions GetLinkerOptions ()
...
+
 		public virtual IEnumerable<string> GetCommonReferencedAssemblies (NPath workingDirectory)
 		{
 			yield return workingDirectory.Combine ("Mono.Linker.Tests.Cases.Expectations.dll").ToString ();
+#if NETCOREAPP
+			foreach (var path in GetTrustedPlatformAssemblies())
+			{
+				// Don't reference testcases dll, as these will be compiled dynamically.
+				if (Path.GetFileName (path) != "Mono.Linker.Tests.Cases.dll")
+					yield return path;
```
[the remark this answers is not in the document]
sbomer: It's a bit unfortunate, but this gives back a huge list of assemblies representing the framework and the test project dependencies (see the list: https://gist.github.com/sbomer/dc5ec8cbd2df2311817525aa2f0f882c), since the framework has been split out into many dlls for core. I could filter it down to `System.Private.CoreLib` and about 10-20 others, but that has the downside that the testcase dependency dlls need to be explicitly listed. I could add those dependencies to a the common references, or just the testcases that need them via `ReferenceAttribute`. The list I would need is approximately (in the order I encountered them):
- System.Private.CoreLib
- System.Runtime
- System
- System.ComponentModel.TypeConverter
- System.Security.Permissions (which comes from a nuget package)
- netstandard
- System.Runtime.Extensions
- System.Diagnostics.Process
- System.Console
- mscorlib
- System.Collections.NonGeneric
- System.Collections
- System.Resources.Writer
- System.Private.Uri

Whether I add these per-testcase or to the common references, I'll need to resolve them using an approach similar to the TPA list. For example, if I add `[Reference(System.Security.Permissions)]` to a testcase, then something will need to resolve that to the full path, which is something like `/Users/sven/src/dotnet/linker/test/Mono.Linker.Tests/bin/illink_Debug/netcoreapp3.0/System.Security.Permissions.dll`. Most likely this would be done in `GetReferencedAssemblies` by finding items on the TPA list that have the right filename.

What do you think? FWIW our SDK, when compiling a .NET Core project, adds references to the whole framework by default, so the current behavior isn't too different.

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
sbomer: I was matching the behavior of csc described at https://github.com/dotnet/roslyn/blob/master/docs/compilers/CSharp/CommandLine.md. However, you're right - it seems our SDK will default to passing `/debug:portable` to the compiler on all platforms. Maybe it would be better to match the SDK behavior here.

test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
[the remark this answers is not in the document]
mrvoorhe: I believe the tests that specify it as kept are the ones the explicitly use roslyn.  The tests don't actually care about the DebuggableAttribute, they just needed to assert it because roslyn was used.  You would remove the `[Kept(DebuggableAttribute)]` from tests.

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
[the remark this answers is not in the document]
mrvoorhe: Would it be so bad require that peverify or pedump be available still even when running the tests on netcore?  We run peverify/pedump out of process so there is technically no reason why we can use it.

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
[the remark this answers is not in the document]
mrvoorhe: > to match the default behavior of codedom

I only used that for convenience and to avoid having to find `csc` on Windows.  Later on it turned out we had to find `csc` anyways.

> I'll try to get these tests working with the out-of-process (+server GC) approach and see what the perf is like. To make it a fair comparison, I'll try using server GC for the in-process version as well.
👍

test/Mono.Linker.Tests/TestCasesRunner/TestCaseMetadaProvider.cs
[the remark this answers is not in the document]
mrvoorhe: Let me know after you give VS2019 a try and see if there are issues with that or if it's something on my end.  I'd like to try it locally and poke around a bit in order to get my head around it better.

test/Mono.Linker.Tests.Cases/References/AssemblyOnlyUsedByUsingWithMcs.cs
sbomer: The existing convention would be for this to go under, for example, `References/Mcs/`, and those below under `TestFramework/Mcs/`. Is that what you'd like, or are you asking for `Mcs` to be the top-level with subdirectories like `Mcs/References`, `Mcs/TestFramework`?

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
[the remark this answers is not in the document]
sbomer: The problem seems to be that it fails to find framework dependencies of the assembly it's trying to verify, and verification fails.

sbomer:
I'm seeing the same thing in VS2019. Removing the CustomSteps project from the solution at least lets it load... but there are more problems later, which seem to stem from the fact that we use different configurations of the same project, and a mixture of old-style and new-style .csproj files. I'm hoping to move the linker to use SDK-style projects now that cecil has moved (https://github.com/jbevain/cecil/pull/574), which might improve the situation in the future.
For now, you should be able to run tests by doing `.\eng\dotnet.ps1 test illink.sln`.

test/Mono.Linker.Tests.Cases/Attributes.Debugger/DebuggerDisplayAttributeOnAssemblyUsingTarget.cs
sbomer: There are some tests that look for types that used to be in mscorlib but now live in 'System.Collections', for example (see this commit: https://github.com/mono/linker/pull/489/commits/36f5e266ea9aa56c395883c4c4787c3bfaac92fc), so this isn't just a matter of replacing "mscorlib" with "System.Private.CoreLib" in all cases.

If you prefer, I could introduce a constant for the cases where the type is in mscorlib/System.Private.CoreLib and use `#if NETCOREAPP` for the rest, but that seems a bit inconsistent to me.

I'd prefer to either
- use the explicit assembly names in attributes, OR
- introduce an abstraction that knows where the expected types live on mono vs .NET Core, and remove the assembly name from these attributes.

Let me know what you think - happy to follow your suggestion (or @mrvoorhe's) on this.

test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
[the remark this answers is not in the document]
sbomer: Gotcha - I'll filter that attribute on mono as well and get rid of the attribute everywhere.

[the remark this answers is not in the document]
sbomer: Just noting that these testcases won't be checking for the attribute even though they used to (even with mcs):
- Assemblywithdefaultsymbols
- AssemblyWithDefaultSymbolsAndSymbolLinkingEnabled
- ReferencesWithMixedSymbolTypes
- CanCompileTestCaseWithDebugPdbs
- CanCompileTestCaseWithMsc

I think that's fine since they probably didn't care about the attribute either. LMK if you think otherwise.

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
sbomer: Since we're translating csc commandline flags into roslin API options, I think it makes sense to match the csc commandline behavior, so tests that request `/debug:full` or `/debug:pdbonly` should get the platform-specific behavior. At the moment we don't have any tests that request debuginfo without specifying the format (viar `/debug+`). In that case the documented behavior is a bit ambiguous, but I believe csc still defaults to the platform-specific format.

sbomer: It might make sense to disable tests that rely on anything but a portable pdb though, since those are less relevant for .NET Core. Then again, it is possible to override the SDK behavior and ask for a specifig `DebugType`.

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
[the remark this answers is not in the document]
mrvoorhe: Ah yes.  That makes sense.  O well.

test/Mono.Linker.Tests.Cases/Attributes.Debugger/DebuggerDisplayAttributeOnAssemblyUsingTarget.cs
marek-safar: Please introduce the constant for corlib name, there is no inconsistency in tests what matters is the readability

test/Mono.Linker.Tests.Cases/References/AssemblyOnlyUsedByUsingWithMcs.cs
marek-safar: I think it should be all under mcs/* that way it's clear the tests are for mcs behaviour only

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
marek-safar: Cecil supports only portable pdb cross platforms and this is not the right place to test pdb/ppdb differences as we don't care about such details in illinker. All tests should use only portable pdb for consistent behaviour.

mrvoorhe: > All tests should use only portable pdb for consistent behaviour.

fyi We have tests for pdb, mdb, ppdb, and embedded pdb in order to ensure that the linker can handle all of them correctly.

test/Mono.Linker.Tests.Cases/Attributes.Debugger/DebuggerDisplayAttributeOnAssemblyUsingTarget.cs
mrvoorhe: I agree constant for corlib and then anything else will have to use #if's and duplicate the asserts.

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
marek-safar: > We have tests for pdb, mdb, ppdb

right but these are tests which are more platform-specific not something we do for linker logic tests

test/Mono.Linker.Tests.Cases/References/AssemblyOnlyUsedByUsingWithMcs.cs
sbomer: Doing this in https://github.com/mono/linker/pull/493.

test/Mono.Linker.Tests.Cases/Attributes.Debugger/DebuggerDisplayAttributeOnAssemblyUsingTarget.cs
sbomer: The testcases get compiled both all at once into a single dll, and individually at test run time, so I need to make sure the reference containing the constant is defined in something referenced at test run time. Does Mono.Linker.Tests.Cases.Expectations seem like a reasonable place to put it @mrvoorhe?

mrvoorhe: Yes.  Add some new static class in that assembly

test/Mono.Linker.Tests/TestCasesRunner/TestCaseMetadaProvider.cs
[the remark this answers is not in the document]
mrvoorhe: I'm not going to have a chance to try this, but it seems fine.

test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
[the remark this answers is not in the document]
mrvoorhe: 👍

test/Mono.Linker.Tests.Cases/References/AssemblyOnlyUsedByUsingWithMcs.cs
sbomer: As a compromise (see discussion in https://github.com/mono/linker/pull/493), I've grouped the mcs tests together in the project file and excluded them from compilation in https://github.com/mono/linker/pull/489/commits/e239b64768549537cff339b617ba689d0deeda75 in this PR. We can always move the mcs tests around later if desired. Does that seem good to you?

Note that excluding them from `Compile` will result in messages in the output like:
```
Could not find the matching type for test case /Users/sven/src/dotnet/linker/test/Mono.Linker.Tests.Cases/Attributes/Mcs/OnlyTypeUsedInAssemblyIsTypeOnAttributePropertyOnType.cs.  Ensure the file name and class name match
```

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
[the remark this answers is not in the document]
sbomer: If it's fine with you, I'd like to do this investigation in a follow-up to this PR. At the moment my priority is to get the tests working on .NET Core and running in ci.

test/Mono.Linker.Tests.Cases/Attributes.Debugger/DebuggerDisplayAttributeOnAssemblyUsingTarget.cs
sbomer: Done in https://github.com/mono/linker/pull/489/commits/29fe2d89f6f2dbf8f518f8c03cf5a3e5b8a80dae and https://github.com/mono/linker/pull/489/commits/dac1eaaaeccf3d82937e68814018480d871f097d.

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
sbomer: Whether we run tests that use different PDB formats is orthogonal to the default compiler behavior - in this change I'm just matching csc, and we could potentially disable any pdb tests that don't pass `"/debug:portable"` on core later if there's a reason not to run them. How does that sound?

sbomer:
I've added a few more changes:
- Use a constant for the core library
- Exclude Mcs tests from compilation
- Restore an ilasm package for .NET Core
- Enable running tests when the projects were built with arcade (which creates a different output structure)
- Run the tests in ci

With this the tests run in ci, and most of them pass.
I disabled the 8 that currently fail on core (https://github.com/mono/linker/pull/489/commits/28d73a90848fb32af1972997e6f6e60e3344d0e1), and will follow up in another PR. Some of these I already have some idea of what's going wrong, but this PR is getting a bit unwieldy and it would be good to get fixes reviewed independently.

sbomer:
@marek-safar please take another look when you get a chance. Thanks!

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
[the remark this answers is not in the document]
sbomer: Filed https://github.com/mono/linker/issues/499 to track.

test/Mono.Linker.Tests.Cases/References/AssemblyOnlyUsedByUsingWithMcs.cs
mrvoorhe: That output is to help identify when a test case is named incorrectly.  So we really do not want to introduce these.

I would recommend we leave it how you had it originally.  Add the #if NETCOREAPP w/ ignores and let's worry about this later if it becomes an annoyance.  @marek-safar Are you OK with that?

Issue #465, opened 2019-02-15
Linker incorrectly removes event implementation


vargaz:
Testcase:
```
public delegate void VectorChangedEventHandler<T>();

public interface IObservableVector<T>
{
	event VectorChangedEventHandler<T> Foo;

	void bar ();
}

class BaseFoo {
	public event VectorChangedEventHandler<object> Foo;

	public BaseFoo () {
	}
}

class ClassFoo : BaseFoo, IObservableVector<object>
{
	public void bar () {
	}
}

class BaseFoo2 : IObservableVector<object>
{
	public event VectorChangedEventHandler<object> Foo;

	public BaseFoo2 () {
		Foo ();
	}

	public void bar () {
	}
}

public class Tests
{
	public static void Main (String[] args) {
		IObservableVector<object> f2 = new BaseFoo2 ();
		f2.Foo += delegate () {};
		IObservableVector<object> f = new ClassFoo ();
		f.bar ();
	}
}
```

Link with:
```
monolinker -a bug.exe -c link
```

In the resulting IL, the ClassFoo class implements IObservableVector<object>, but neither it or its base type contains the implementations for the add_Foo/remove_Foo methods from IObservableVector<object>.

vargaz:
With linker version 973587a221a10d4a73c1ea858b455880aa57e618.

mrvoorhe:
I struggled to reproduce at first.  Then I realized the original test in the github notification email didn't have generics while the latest version of your test case on the site did 😄 

https://github.com/Unity-Technologies/linker/commit/c178fa021c047e2442994e93d8741f6a69695030

I'll look into a fix soon.

mrvoorhe:
This is related to base classes supplying interface members on a derived type AND when generics are involved.  I did not finish looking into it, but it seems like a bug that has always existed and does not appear to be related to the interface sweeping changes that have landed recently.

I'm going to be out for a few weeks on paternity leave so I won't have time to finish fixing this for awhile.  In case it becomes important to fix, here are change sets that may be of help to someone.

Lots of tests to reproduce the issue.  I have some more complex cases than what was reported in this issue, but I think there are probably still some more complex edge cases that would be worth adding tests for to see if they are working correctly or not.
https://github.com/Unity-Technologies/linker/commit/747664550e4288a4e0f88018ce74cf052612cce3

Two aspects to a fix that I have identified so far are

[...]

PR #492, opened 2019-03-20
Fix interface members from base with generics

mrvoorhe:
This PR fixes https://github.com/mono/linker/issues/465

The linker is incorrectly removing base class members that fulfill an interface when generics are involved.

To fix this took a couple of steps

* Improved the TypeMapStep so that it could identify members from a base class that supply a member for a generic instance interface

* The TypeMapStep will now record the associated interface implementation for an override in the annotation store so that the `MarkStep` does not need to redo figuring out which interface implementation an override is associated with.  I had some trouble recording the interface implementation for all overrides of interface members.  This means that some interface implementation logic will still remain in the `MarkStep` (ex : `MarkStep.MarkExplicitInterfaceImplementation`).  What this means is that `OverrideInformation.MatchingInterfaceImplementation` is populated in some cases and in others it is not.  I will try to clean this up later.

Side notes.

* When looking at the new tests, the tests that hit the bug were essentially all of the tests named `BaseProvidesInterfaceMember\GenericInterface*`

* I ran across another bug while writing tests.  See the new tests `UnusedGenericInterfaceIsRemoved` and `GenericInterfaceWithMethodManyVariations`.  We are over preserving in these cases.  Fixing these looked like it would require a decent number of additional changes and this PR was already getting fairly large.  Since this is an over preservation bug, rather than invalid IL, I decided to leave the tests and punt on a fix for the time being.

* The `Inflater` class was copied from our IL2CPP code base.  If there is a simpler way to accomplish this, I'm all ears.  I thought it was better to copy a proven implementation and leave it alone rather than try to remove stuff we don't need


mrvoorhe:
Looking into test failures

mrvoorhe:
The test failures were due to problems with handling of generic methods.  I added the following new tests and added a fix for them.

```
GenericInterfaceWithGenericMethod
GenericInterfaceWithGenericMethod2
GenericInterfaceWithGenericMethod3
```

mrvoorhe:
Apparently there are more issues.   These failures don't happen when the tests are compiled with `csc`.  Investigating

mrvoorhe:
I could use your input on this one @marek-safar .  Looking at the failure of `GenericInterfaceWithGenericBaseMethod` with `mcs` only.  I see that `mcs`  is generating an explicit interface method in `FooWithBase` even though one is not declared.

```
    // method line 7
    .method private final virtual hidebysig newslot strict 
           instance default void Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType.BaseProvidesInterfaceMember.IFoo.Method (!T1 arg, !T2 arg2)  cil managed 
    {
        // Method begins at RVA 0x2085
	.override method instance void class Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType.BaseProvidesInterfaceMember.GenericInterfaceWithGenericBaseMethod/IFoo`2<!T1, !T2>::Method(!0, !1)
	// Code size 9 (0x9)
	.maxstack 8
	IL_0000:  ldarg.0 
	IL_0001:  ldarg.1 
	IL_0002:  ldarg.2 
	IL_0003:  call instance void class Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType.BaseProvidesInterfaceMember.GenericInterfaceWithGenericBaseMethod/BaseFoo`2<!T1, !T2>::Method(!0, !1)
	IL_0008:  ret 
    } // end of method FooWithBase`2::Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType.BaseProvidesInterfaceMember.IFoo.Method
```

The test fails because the explicit method doesn't have `[Kept]` on it so we expect it to be removed, which it is not.

Should I just update these 3 tests to compile with `csc` so that the IL generated by the compiler matches the assertions we define?

mrvoorhe:
> Should I just update these 3 tests to compile with csc so that the IL generated by the compiler matches the assertions we define?

I went ahead and made this change so I'm pretty sure that's going to be the best way to work around this.

I also added a new test `GenericInterfaceWithGenericBaseMethodWithExplicit` to get some coverage on the explicit interface implementation scenario that `mcs` was generating.

PR #493, opened 2019-03-20
[nomerge] Move mcs tests to a separate directory

sbomer:


mrvoorhe:
Don't merge this.  This needs additional changes.  As the PR is now these tests will no longer run.

The additional changes that are needed are:

* Namespaces of the files that were moved need to be updated to match the directory structure
* A new method with `[TestCaseSource]` needs to be added to `All` to run the tests under the `Mcs` directory

sbomer:
Thanks, will fix.

sbomer:
It looks like there are some more dependencies on the directory layout. Some of the moved tests have dependencies that are expected to be found in a subdirectory.

mrvoorhe:
So a few options are

1) update the relative paths in the tests so that they can find the files again
2) Leave the tests where they were
3) Leave the tests under the same top level test directory they were under before, but move them into a `Mcs` directory

I'm OK with any.  I'm partial to (3) or (2).  I wasn't really sold on the need to move them out to a single `Mcs` directory in the first place.

sbomer:
Thanks @mrvoorhe.
- 1 works for those that specify dependencies as file paths, but some of them also look for files matching a typeref in an attribute - for these, that resolution logic would need to change (or we would need to duplicate files).
- 3 may or may not have the same problem - I haven't tried yet.

My preference would be 2 (to leave them where they are).

sbomer:
Closing this as we are leaving them where they are (https://github.com/mono/linker/commit/27da628f5bf3e012da2841362b1a6d44b510bd8a). Feel free to reopen if you think it should be revisited.

PR #496, opened 2019-03-21
Fix overriding assembly action in ResolveFromAssemblyStep.ProcessLibrary

akoeplinger:
When a custom link action for an assembly is passed to the linker on the commandline we'd override it with `Copy` in `ProcessLibrary`. Instead we now check whether we already have an assembly action before setting it.

`LoadI18nAssemblies` used this to force `Copy` on the I18N assemblies by calling `ProcessLibrary` with `RootVisibility.Any` so we need to set the assembly action explicitly to maintain that behavior.


mrvoorhe:
Can a test be added?

mrvoorhe:
It sounds like the test in https://github.com/mono/linker/pull/497 will cover this since it will fail without this change?

PR #502, opened 2019-03-22
Mark instance method bodies lazily

mrvoorhe:
Instance method bodies do not need to be marked until an instance of the type could exist.  Any instance methods that were marked on types that are never instantiated will be converted to a throw.

* A number of existing tests needed to be updated to avoid lazy body marking

* Add a new CodeOptimization to allow disabling this mechanism.  Reasoning was, (1) it's simple to support an option.  (2) It may be helpful if it causes problems. (3) I could see using this option to make writing certain tests easier.

* Marking of some small bodies will not be deferred.  The idea is that the size cost of some methods is less than converting them to a throw.  So we might as well leave them alone.  See `IsWorthConvertingToThrow`

* Factored out `MarkAndCacheNotSupportedCtorString` since I need to call it from multiple places now.

* Expose the `RewriteBody*` methods in CodeRewriterStep for overriding.  A few motivations for this. (1) We have a runtime that doesn't support exceptions.  When we target that runtime we will need to override `ConvertToThrow` with something else.  (2) I will likely override all of the `RewriteBody*` methods so that I can record which bodies are changed and hook it up to logging mechanisms that we have.

Note.  This PR has a test that depends on https://github.com/mono/linker/pull/501 .  That PR will need to land first and I will need to rebase after that lands.  There will be 1 failing test until that happens.


mrvoorhe:
Rebased and should be ready to go pending CI results 🤞

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -93,6 +96,18 @@ protected virtual void InitializeAssembly (AssemblyDefinition assembly)
...
+		void Complete ()
+		{
+			bool anyConverted = false;
+			foreach (var body in _pendingBodies) {
+				anyConverted = true;
+				Annotations.SetAction (body.Method, MethodAction.ConvertToThrow);
+			}
+
+			if (anyConverted)
+				MarkAndCacheNotSupportedCtorString ();
```
marek-safar: I think this is in theory too late and should be called during Process. Call it in inside MarkMethodBody

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -46,6 +46,7 @@ public partial class MarkStep : IStep {
 		protected Queue<AttributeProviderPair> _assemblyLevelAttributes;
 		protected Queue<AttributeProviderPair> _lateMarkedAttributes;
 		protected List<TypeDefinition> _typesWithInterfaces;
+		protected List<MethodBody> _pendingBodies;
```
marek-safar: Please rename it to a better name, e.g. `_unreachableBodies`

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2014,6 +2046,15 @@ void MarkMethodIfNotNull (MethodReference method)
 
 		protected virtual void MarkMethodBody (MethodBody body)
 		{
+			if (_context.IsOptimizationEnabled (CodeOptimizations.LazyBodyMarking)) {
+				if (!body.Method.IsStatic
```
marek-safar: I'd like more simple method with the logic like

`if (_context.IsOptimizationEnabled (CodeOptimizations.LazyBodyMarking) && IsUnreachableBody (method.Body)) {`

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2031,6 +2072,34 @@ protected virtual void MarkMethodBody (MethodBody body)
 			PostMarkMethodBody (body);
 		}
 
+		static bool IsWorthConvertingToThrow (MethodBody body)
```
marek-safar: Hmm, this optimization would be better for all ConvertToThrow receivers. Could you move it there?

src/linker/Linker/Driver.cs
```diff
@@ -363,6 +363,9 @@ public void Run (ILogger customLogger = null)
 						case "overrideremoval":
 							context.DisabledOptimizations |= CodeOptimizations.OverrideRemoval;
 							break;
+						case "lazybodymarking":
```
marek-safar: this needs better naming with documentation for the command like option

src/linker/Linker.Steps/MarkStep.cs
mrvoorhe: Good catch.

src/linker/Linker.Steps/MarkStep.cs
mrvoorhe: Done

marek-safar:
Did you forgot to push the changes ;-)

src/linker/Linker/Driver.cs
mrvoorhe: I went with "unreachablebodies"  are you OK with that?

Added command line docs

src/linker/Linker.Steps/MarkStep.cs
mrvoorhe: I moved this out to MethodBodyScanner and I now call from a number of places

mrvoorhe:
> Did you forget to push the changes ;-)
I didn't finish making all of the changes before I had to go.

All changes are done and pushed now.

mrvoorhe:
I'm looking into the test failures with illink

mrvoorhe:
I refactored `IsWorthConvertingToThrow` a bit to handle `noop`.  A bunch of unrelated tests were failing during the illink build because roslyn likes to write empty methods as
```
noop
ret
```

I can't get the .net core tests running locally.  Hopefully everything is fixed now, but I'll have to wait for CI to finish to know for certain.

src/linker/Linker/BCL.cs
```diff
@@ -71,6 +71,10 @@ public static TypeDefinition FindPredefinedType (string ns, string name, LinkCon
 			if (cache.TryGetValue ("System.Private.CoreLib", out corlib))
 				return corlib.MainModule.GetType (ns, name);
 
+			// System.NotSupportedException lives here on .netcore
```
mrvoorhe: fyi, I had to add this so that `System.NotSupportedException` could be found.  This could become a tedious pattern to maintain

mrvoorhe:
@sbomer `BCL.FindPredefinedType` can't find `System.NotSupportedException` on .netcore.  Any idea how I should locate that type?

I tried adding
```
// System.NotSupportedException lives here on .netcore
			if (cache.TryGetValue ("System.Runtime", out corlib))
				return corlib.MainModule.GetType (ns, name);
```

To `FindPredefinedType` but that didn't work.

I can't get the .netcore tests working locally so that's making iterating on this challenging.

sbomer:
@mrvoorhe I can take a look. Could you try running the tests locally on core using `./eng/dotnet.sh test illink.sln` and let me know if you have any issues?

sbomer:
Also, it looks like `System.NotSupportedException` in `System.Runtime` is just a type forwarder to `System.Private.CoreLib`. I'm not sure why it wouldn't have been found in SPC in the first place.

sbomer:
Switching `System.Runtime` -> `System.Private.CoreLib` in `FindPredefinedType` fixes at least some of the test failures when I run locally (I haven't run them all yet). I'm not sure though exactly what the intention of `FindPredefinedType` is or why it wouldn't work with type forwarders. If it's just to resolve a set of types that we assume are present in some particular assembly for the linker to work, then it seems like hard-coding `System.PrivateCoreLib` might be ok. Maybe we could use a define for the target's core library similar to https://github.com/mono/linker/blob/31555535b160c1de7ba72fd7c51ec584ef395fdb/test/Mono.Linker.Tests.Cases.Expectations/Support/PlatformAssemblies.cs?

mrvoorhe:
@sbomer How can I debug this?  I can what you said and I can reproduce the failure locally.  However, Rider and VS 2019 are not happy with `illink.sln` and I tried to do some `Console.WriteLine` debugging but it looks like `dotnet.sh test illink.sln` is doing something other than let stdout go to the console

mrvoorhe:
I think I know what the issue is now.  The test that is failing is an IL test.

That said, I would still like some pro tips on how to debug individual tests on .netcore for future reference

sbomer:
https://github.com/mono/linker/pull/471 may help with IDE support for the .sln, but I'm not sure. I've had lots of issues with tooling support for projects that mix configurations/targetframeworks and projectreferences. Sorry it's causing you trouble.

For debugging, I've been using VS code. Here's the best way I've found to do it so far:
- Add this to the linker somewhere:
  ```
  while (!System.Diagnostics.Debugger.IsAttached)
      System.Threading.Thread.Sleep(500)
  ```
- Set breakpoints
- Run an individual test:
  ```
  ./eng/dotnet.sh test illink.sln --filter TestCaseName
  ```
  (See https://github.com/Microsoft/vstest-docs/blob/master/docs/filter.md for some docs on the filter expression)
- Wait for it to hit the while loop above, and attach to `testhost.dll`.

mrvoorhe:
@marek-safar Good to go!

alexanderkyte:
This has been shown to regress the mono harness when mono is bumped to this commit.

mrvoorhe:
@alexanderkyte Can you investigate a bit?  If any types are created from native and do not get an instance .ctor marked it will cause problems.  If this is the case it is not necessarily a problem with these changes.  Instead the fix is to properly annotate w/e instance .ctor is needed by the type.

It's certainly possible there is a bug with these changes, but I'd prefer to rule out the above before making any changes to disable this.

alexanderkyte:
@mrvoorhe I have a PR testing whether disabling the `unreachablebodies` optimization will fix the build. If so, I will triage what breaks when we enable the optimization. That way we can make use of other changes to the linker while we get `unreachablebodies` working separately.

PR #504, opened 2019-03-26
Always consider ValueType instantiated

mrvoorhe:
This wasn't causing any issues today.  ValueType is Serializable and that leads to `MarkStep.MarkSerializable` being called, which ensures `ValueType::.ctor()` is marked, which ensures `ValueType` is marked as instantiated.

However, at Unity we have a smaller class library profile that does not have the serializable code.  As a result `ValueType` is not marked as instantiated.  This leads to struct overrides of object members being removed.

Having so much hinge on `ValueType` being Serializable seems fragile.  So rather than depend on that, let's also consider `ValueType` one of the things that should always be marked instantiated if it appears.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1436,6 +1436,8 @@ namespace Mono.Linker.Steps {
 				// could exist
 				case "Delegate":
 				case "MulticastDelegate":
+				case "ValueType":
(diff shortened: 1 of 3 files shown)


mrvoorhe:
I updated the PR to also include `System.Enum`.  It's in the same situation as `ValueType`.  The only reason it gets marked as instance possible today is because we end up calling `MarkSerializable` on it.

PR #506, opened 2019-03-27
Annotate private methods kept because of reflection

alexanderkyte:
This attribute is going to be used by the AOT code generator to identify private methods that can have code generated to the visible call sites. 

These methods with the Reflected attribute will necessarily have call sites that cannot be seen, meaning that we cannot specialize the definition.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2252,6 +2252,8 @@ namespace Mono.Linker.Steps {
 					Tracer.Push ($"Reflection-{method}");
 					try {
 						MarkMethod (method);
+						if (_context.AnnotateUnseenCallers)
+							MarkMethodReflected (method);
 					} finally {
 						Tracer.Pop ();
 					}
@@ -2271,8 +2273,15 @@ namespace Mono.Linker.Steps {
 						// It is not easy to reliably detect in the IL code whether the getter or setter (or both) are used.
 						// Be conservative and mark everything for the property.
 						MarkProperty (property);
+
 						MarkMethodIfNotNull (property.GetMethod);
+						if (property.GetMethod != null && _context.AnnotateUnseenCallers)
+							MarkMethodReflected (property.GetMethod);
+
 						MarkMethodIfNotNull (property.SetMethod);
+						if (property.SetMethod != null && _context.AnnotateUnseenCallers)
+							MarkMethodReflected (property.SetMethod);
+
 					} finally {
 						Tracer.Pop ();
 					}
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1683,6 +1683,11 @@ namespace Mono.Linker.Steps {
 				MarkMethod (method);
 		}
 
+		protected void MarkMethodReflected (MethodDefinition method)
+		{
+			Annotations.MarkUnseenCallers (method);
+		}
+
 		protected virtual MethodDefinition MarkMethod (MethodReference reference)
 		{
 			reference = GetOriginalMethod (reference);
@@ -2227,6 +2232,8 @@ namespace Mono.Linker.Steps {
 					Tracer.Push ($"Reflection-{method}");
 					try {
 						MarkMethod (method);
+						if (_context.AnnotateUnseenCallers)
+							MarkMethodReflected (method);
 					} finally {
 						Tracer.Pop ();
 					}
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1683,8 +1683,21 @@ namespace Mono.Linker.Steps {
 				MarkMethod (method);
 		}
 
+		bool markedReflectionBlocked;
+
 		protected void MarkMethodReflected (MethodDefinition method)
 		{
+			if (!method.IsPrivate)
+				return;
+
+			if (!markedReflectionBlocked) {
+				markedReflectionBlocked = true;
+
+				MethodDefinition ctor = ReflectionBlockedStep.GetReflectionBlockedAttr (_context);
+				Annotations.Mark (ctor.Resolve ());
+				Annotations.Mark (ctor.DeclaringType.Resolve ());
+			}
+
 			Annotations.MarkUnseenCallers (method);
 		}
 
@@ -2232,7 +2245,7 @@ namespace Mono.Linker.Steps {
 					Tracer.Push ($"Reflection-{method}");
 					try {
 						MarkMethod (method);
-						if (_context.AnnotateUnseenCallers)
+						if (_context.NoReflectionMethods)
 							MarkMethodReflected (method);
 					} finally {
 						Tracer.Pop ();
@@ -2252,7 +2265,7 @@ namespace Mono.Linker.Steps {
 					Tracer.Push ($"Reflection-{method}");
 					try {
 						MarkMethod (method);
-						if (_context.AnnotateUnseenCallers)
+						if (_context.NoReflectionMethods)
 							MarkMethodReflected (method);
 					} finally {
 						Tracer.Pop ();
@@ -2275,11 +2288,11 @@ namespace Mono.Linker.Steps {
 						MarkProperty (property);
 
 						MarkMethodIfNotNull (property.GetMethod);
-						if (property.GetMethod != null && _context.AnnotateUnseenCallers)
+						if (property.GetMethod != null && _context.NoReflectionMethods)
 							MarkMethodReflected (property.GetMethod);
 
 						MarkMethodIfNotNull (property.SetMethod);
-						if (property.SetMethod != null && _context.AnnotateUnseenCallers)
+						if (property.SetMethod != null && _context.NoReflectionMethods)
 							MarkMethodReflected (property.SetMethod);
 
 					} finally {
--- a/src/linker/Linker.Steps/ReflectionBlockedStep.cs
+++ b/src/linker/Linker.Steps/ReflectionBlockedStep.cs
@@ -1,18 +1,19 @@
 ﻿using System;
+using System.Collections.Generic;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 
 using System.Runtime.CompilerServices;
 
 namespace Mono.Linker.Steps {
-	public class UnseenCallerAnnotateStep : BaseStep
+	public class ReflectionBlockedStep : BaseStep
 	{
 		AssemblyDefinition assembly;
 		MethodReference noOptAttr;
 
 		static MethodDefinition _reflectionMethod;
 
-		static MethodDefinition GetReflectionMethod (LinkContext context) {
+		public static MethodDefinition GetReflectionBlockedAttr (LinkContext context) {
 			if (_reflectionMethod != null)
 				return _reflectionMethod;
 
@@ -38,6 +39,9 @@ namespace Mono.Linker.Steps {
 				return;
 
 			this.assembly = assembly;
+			this.noOptAttr = assembly.MainModule.ImportReference (GetReflectionBlockedAttr (Context));
+			if (noOptAttr == null)
+				throw new Exception("Could not import System.Runtime.CompilerServices.ReflectionBlockedAttribute in BCL.");
 
 			foreach (var type in assembly.MainModule.Types)
 				ProcessType (type);
@@ -45,42 +49,51 @@ namespace Mono.Linker.Steps {
 
 		void ProcessType (TypeDefinition type)
 		{
+			bool canAnnotateAll = true;
+			List<MethodDefinition> canAnnotate = new List<MethodDefinition> ();
+
 			foreach (var method in type.Methods) {
-				if (method.HasBody)
-					ProcessMethod (method);
+				// Public methods have non-visible call sites by default, this attribute doesn't
+				// help us at all.
+				//
+				// See mono_aot_can_specialize in aot-compiler.c in mono
+				if (!method.IsPrivate)
+					continue;
+
+				if (!method.HasBody)
+					continue;
+
+				if (Annotations.HasUnseenCallers (method)) {
+					canAnnotateAll = false;
+					continue;
+				}
+
+				canAnnotate.Add (method);
+			}
+
+			if (canAnnotateAll) {
+				AnnotateType (type);
+			} else {
+				foreach (var method in canAnnotate)
+					AnnotateMethod (method);
 			}
 
 			foreach (var nested in type.NestedTypes)
 				ProcessType (nested);
 		}
 
-		void ProcessMethod (MethodDefinition method)
+		void AnnotateType (TypeDefinition type)
 		{
-			// Public methods have non-visible call sites by default, this attribute doesn't
-			// help us at all.
-			//
-			// See mono_aot_can_specialize in aot-compiler.c in mono
-			if (!method.IsPrivate)
-				return;
+			var cattr = new CustomAttribute (noOptAttr);
+			type.CustomAttributes.Add (cattr);
+			Annotations.Mark (cattr);
+		}
 
-			if (Annotations.HasUnseenCallers (method)) {
-				Console.WriteLine ("{0} has unseen callers", method.Name);
-				return;
-			} else {
-				Console.WriteLine ("{0} has no unseen callers", method.Name);
-			}
-
-			if (noOptAttr == null) {
-				noOptAttr = assembly.MainModule.ImportReference (GetReflectionMethod (Context));
-				if (noOptAttr == null)
-					throw new Exception("Could not import System.Runtime.CompilerServices.ReflectionBlockedAttribute in BCL.");
-			}
+		void AnnotateMethod (MethodDefinition method)
+		{
 			var cattr = new CustomAttribute (noOptAttr);
 			method.CustomAttributes.Add (cattr);
-
 			Annotations.Mark (cattr);
-			Annotations.Mark (cattr.AttributeType.Resolve ());
-			Annotations.Mark (cattr.Constructor.Resolve ());
 		}
 	}
 }
--- a/src/linker/Linker.Steps/ReflectionBlockedStep.cs
+++ b/src/linker/Linker.Steps/ReflectionBlockedStep.cs
@@ -8,6 +8,9 @@ using System.Runtime.CompilerServices;
 namespace Mono.Linker.Steps {
 	public class ReflectionBlockedStep : BaseStep
 	{
+		public const string AttributeNamespace = "System.Runtime.CompilerServices";
+		public const string AttributeName = "DisablePrivateReflectionAttribute";
+
 		AssemblyDefinition assembly;
 		MethodReference noOptAttr;
 
@@ -17,7 +20,7 @@ namespace Mono.Linker.Steps {
 			if (_reflectionMethod != null)
 				return _reflectionMethod;
 
-			TypeDefinition methodImpl = BCL.FindPredefinedType("System.Runtime.CompilerServices", "ReflectionBlockedAttribute", context);
+			TypeDefinition methodImpl = BCL.FindPredefinedType(AttributeNamespace, AttributeName, context);
 			if (methodImpl == null)
 				throw new Exception("Could not find System.Runtime.CompilerServices.ReflectionBlockedAttribute in BCL.");
 
--- a/src/linker/Linker.Steps/ResolveFromXmlStep.cs
+++ b/src/linker/Linker.Steps/ResolveFromXmlStep.cs
@@ -434,6 +434,8 @@ namespace Mono.Linker.Steps {
 				Context.LogMessage ($"Duplicate preserve in {_xmlDocumentLocation} of {method.FullName}"); 
 
 			Annotations.Mark (method);
+			if (Context.AnnotateUnseenCallers)
+				Annotations.MarkUnseenCallers (method);
 			Tracer.AddDirectDependency (this, method);
 			Annotations.SetAction (method, MethodAction.Parse);
 		}
--- a/src/linker/Linker.Steps/ResolveFromXmlStep.cs
+++ b/src/linker/Linker.Steps/ResolveFromXmlStep.cs
@@ -434,7 +434,7 @@ namespace Mono.Linker.Steps {
 				Context.LogMessage ($"Duplicate preserve in {_xmlDocumentLocation} of {method.FullName}"); 
 
 			Annotations.Mark (method);
-			if (Context.AnnotateUnseenCallers)
+			if (Context.NoReflectionMethods)
 				Annotations.MarkUnseenCallers (method);
 			Tracer.AddDirectDependency (this, method);
 			Annotations.SetAction (method, MethodAction.Parse);
--- a/src/linker/Linker.Steps/UnseenCallerAnnotateStep.cs
+++ b/src/linker/Linker.Steps/UnseenCallerAnnotateStep.cs
@@ -10,6 +10,28 @@ namespace Mono.Linker.Steps {
 		AssemblyDefinition assembly;
 		MethodReference noOptAttr;
 
+		static MethodDefinition _reflectionMethod;
+
+		static MethodDefinition GetReflectionMethod (LinkContext context) {
+			if (_reflectionMethod != null)
+				return _reflectionMethod;
+
+			TypeDefinition methodImpl = BCL.FindPredefinedType("System.Runtime.CompilerServices", "ReflectionBlockedAttribute", context);
+			if (methodImpl == null)
+				throw new Exception("Could not find System.Runtime.CompilerServices.ReflectionBlockedAttribute in BCL.");
+
+			foreach (var ref_method in methodImpl.Methods)
+			{
+				if (!ref_method.IsConstructor)
+					continue;
+				if (ref_method.Parameters.Count != 0)
+					continue;
+				_reflectionMethod = ref_method;
+			}
+
+			return _reflectionMethod;
+		}
+
 		protected override void ProcessAssembly (AssemblyDefinition assembly)
 		{
 			if (Annotations.GetAction (assembly) != AssemblyAction.Link)
@@ -49,20 +71,7 @@ namespace Mono.Linker.Steps {
 			}
 
 			if (noOptAttr == null) {
-				TypeDefinition methodImpl = BCL.FindPredefinedType("System.Runtime.CompilerServices", "ReflectionBlockedAttribute", Context);
-				if (methodImpl == null)
-					throw new Exception("Could not find System.Runtime.CompilerServices.ReflectionBlockedAttribute in BCL.");
-
-				MethodDefinition reflectionMethod = null;
-				foreach (var ref_method in methodImpl.Methods)
-				{
-					if (!ref_method.IsConstructor)
-						continue;
-					if (ref_method.Parameters.Count != 0)
-						continue;
-					reflectionMethod = ref_method;
-				}
-				noOptAttr = assembly.MainModule.ImportReference (reflectionMethod);
+				noOptAttr = assembly.MainModule.ImportReference (GetReflectionMethod (Context));
 				if (noOptAttr == null)
 					throw new Exception("Could not import System.Runtime.CompilerServices.ReflectionBlockedAttribute in BCL.");
 			}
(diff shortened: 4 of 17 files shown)


alexanderkyte:
Depends on https://github.com/mono/mono/pull/13588

jkotas:
How is this attribute different from the existing `PreserveAttribute` ?

alexanderkyte:
PreserveAttribute is applied when we want the linker to act as if the annotated element was a root. We don't currently apply it to elements we find when traversing the source. It seems to have different semantics.

This attribute is applied to every private method that is called directly by reflection or is marked in the XML as called by an external caller. 

It's also worth noting that PreserveAttribute is a message from the coder, made for the linker. This is a message made by the linker, for the backend code generator. They're used differently, intend to communicate different things, and probably have edge cases in which they disagree. 

Most of the time, if you set PreserveAttribute on a private method, you should see the Reflected attribute also propagated to it. The majority of the `Reflected` methods probably do not have the `PreserveAttribute` attribute on it. Merging them is definitely possible.

test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
```diff
@@ -493,6 +493,11 @@ protected virtual void VerifyCustomAttributes (ICustomAttributeProvider src, ICu
 			var expectedAttrs = GetExpectedAttributes (src).ToList ();
 			var linkedAttrs = FilterLinkedAttributes (linked).ToList ();
 
+			for (int i=0; i < linkedAttrs.Count; i++) {
```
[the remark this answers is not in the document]
alexanderkyte: I can make this optional if the addition of attributes is too much, but the codegen size savings are going to motivate it being the default for Xamarin when possible.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -434,6 +434,7 @@ void MarkMethod (MethodDefinition method)
 				Context.LogMessage ($"Duplicate preserve in {_xmlDocumentLocation} of {method.FullName}"); 
 
 			Annotations.Mark (method);
+			Annotations.MarkReflected (method);
```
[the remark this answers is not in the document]
alexanderkyte: If a method is private and accessed via a method other than the visible number of callers in the assembly, the call happens through a method other than traditional C#-compiled function calls. This would require either managed reflection or an equivalent unmanaged reflection of the method through the metadata tables.  

If the word "Reflected" portrays a false connotation that this runtime code reflection / invocation happens through the managed System.Reflection, I'm happy to take any suggestions to rename it to. 

I wanted to clarify that it was more specific than just "ExternallyVisible" though. This won't be applied to public methods, only private ones that we know can be accessed.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -391,7 +392,21 @@ protected IList<CustomAttribute> SweepCustomAttributes (ICustomAttributeProvider
 				SweepCustomAttributes (provider);
 		}
 
-		protected virtual void SweepMethods (Collection<MethodDefinition> methods)
+		void AddReflectedAttr (MethodDefinition method, AssemblyDefinition assembly)
```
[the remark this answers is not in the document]
alexanderkyte: For #2, mono embedders using the LLVM AOT backend can expect to see code size reductions when using this attribute: https://github.com/mono/mono/pull/13697

test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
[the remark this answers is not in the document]
mrvoorhe: My comment here is with regards to how tests should work.  If I look at a test file, I should be able to understand the changes the linker is expected to make.  Quietly filtering out an attribute injected by the linker makes it so that one can no longer understand all of the changes the linker will make.

Whether or not this new option is on or off is a separate discussion.

mrvoorhe:
It would be good to have some tests.  This will require adding support for a new assertion attribute.  Ex: `[ExpectInjectedAttributeAttribute(Type type)]` and `[ExpectInjectedAttributeAttribute(string typeName)]`

src/linker/Linker.Steps/ResolveFromXmlStep.cs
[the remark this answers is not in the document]
mrvoorhe: I was thinking of methods called from native code.  It wasn't clear to me if you wanted to capture these as well.

I don't have a better naming suggestion at the moment.

src/linker/Linker.Steps/SweepStep.cs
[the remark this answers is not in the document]
mrvoorhe: > For #2, mono embedders using the LLVM AOT backend can expect to see code size reductions when using this attribute: mono/mono#13697

That's good to know.

Currently, we don't use the LLVM AOT backend so for us it would only increase size.

I don't really care what the switch looks like.  Something like the following may kill two birds with one stone.
```
protected virtual GetMonoCodeReflectedAttributeCtor()
{
     return System.Type.GetType ("Mono.Codegen+Reflected")?.GetConstructor (Array.Empty<System.Type> ());
}

		void AddReflectedAttr (MethodDefinition method, AssemblyDefinition assembly)
		{
			// Public methods have non-visible call sites by default, this attribute doesn't
			// help us at all.
			//
			// See mono_aot_can_specialize in aot-compiler.c in mono
			if (!method.IsPrivate)
				return;

			var reflectionMethod = GetMonoCodeReflectedAttributeCtor();
			if (reflectionMethod == null)
				return;

			MethodReference methodRef = assembly.MainModule.Import (reflectionMethod);
			method.CustomAttributes.Add(new CustomAttribute (methodRef));
		}
```

We could override `GetMonoCodeReflectedAttributeCtor` to turn this behavior off.  And the null check would fix the tests that are failing with this PR when ran on windows against the .NET Framework assemblies.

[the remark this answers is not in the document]
mrvoorhe: Or you could pull this logic out into it's own step.  Solving the slightly confusion situation of this being in the `SweepStep` and then giving us the ability to simply not register this new step.

mrvoorhe:
Should methods marked by the `PreserveDependency` mechanism also have `MarkReflected` called on them?  See the `PreserveDependencyTests` for examples and `MarkStep.MarkDependencyMethod`

src/linker/Linker.Steps/SweepStep.cs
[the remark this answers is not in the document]
alexanderkyte: I moved this to the CodeRewriter step. Can definitely put it in it's own step instead.

test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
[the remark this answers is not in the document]
alexanderkyte: It's easy enough for me to just add a set of these tests with it disabled and a set with it enabled. I'll do that instead.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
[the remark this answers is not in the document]
jkotas: > I wanted to clarify that it was more specific than just "ExternallyVisible" though. This won't be applied to public methods, only private ones that we know can be accessed.

Are you also going to prevent the non-reflectable methods from being invoked via reflection (or other indirect means) at runtime?

Otherwise, you can get very weird crashes when the IL linker guesses wrong and the optimizations takes advantage of invariant that does not hold.

We have a very similar attribute for .NET Native, just with a opposite polarity: https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/src/System/Runtime/CompilerServices/ReflectionBlockedAttribute.cs. The AOT compiler takes advantage of this attribute for optimizations, but the runtime also prevents these methods from being invoked by reflection.

MichalStrehovsky:
I think this can be done in a way where not just the LLVM backend benefits from this (i.e. we can get a size reduction in all scenarios).

If a private method is not used from reflection, flip the accessibility of the method to `compilercontrolled` (might need to open your copy of the ECMA-335 spec on that) and set the name to null. Members with compiler controlled accessibility are only accessible within the defining module and only through the metadata token (the name/sig pair cannot be used to look them up). Might not be able to do this for all methods (e.g. the names of p/invokes are still important), but those are not that interesting anyway.

We get an immediate size reduction for the IL assembly because we no longer need to carry around names of such private methods.

At LLVM codegen time, you can assume that `compilercontrolled` methods will have all of their references visible in the containing module.

The methods will not be particularly useful for reflection without their name (I'm not sure if reflection surfaces them in `GetMethods()` at runtime, but anyone would have a hard time locating the method without the name), so the reflection blocking aspect kind of just naturally falls out.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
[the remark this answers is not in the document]
MichalStrehovsky: The opposite polarity is nice because it means that files that IL Linker didn't touch (because we didn't run it) will be set up the right way ("assume everything can be accessed by reflection").

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -391,7 +392,21 @@ protected IList<CustomAttribute> SweepCustomAttributes (ICustomAttributeProvider
...
+		{
+			// Public methods have non-visible call sites by default, this attribute doesn't
+			// help us at all.
+			//
+			// See mono_aot_can_specialize in aot-compiler.c in mono
+			if (!method.IsPrivate)
+				return;
+
+			var reflectionMethod = System.Type.GetType ("Mono.Codegen+Reflected").GetConstructor (Array.Empty<System.Type> ());
```
joncham: Won't think lookup the type in the currently running linker process?

src/linker/Linker.Steps/CodeRewriterStep.cs
```diff
@@ -43,6 +50,35 @@ void ProcessMethod (MethodDefinition method)
 				RewriteBodyToFalse (method);
 				break;
 			}
+
+			if (Annotations.IsReflected (method))
+				AddReflectedAttr (method);
+		}
+
+		void AddReflectedAttr (MethodDefinition method)
```
marek-safar: Please rename the method

src/linker/Linker.Steps/CodeRewriterStep.cs
```diff
@@ -43,6 +50,35 @@ void ProcessMethod (MethodDefinition method)
...
+		{
+			// Public methods have non-visible call sites by default, this attribute doesn't
+			// help us at all.
+			//
+			// See mono_aot_can_specialize in aot-compiler.c in mono
+			if (!method.IsPrivate)
+				return;
+
+			if (noOptAttr == null) {
+				var methodImpl = typeof (System.Runtime.CompilerServices.MethodImplAttribute);
```
marek-safar: This is wrong, you need to use target reference type. Use `FindPredefinedType` instead

src/linker/Linker.Steps/CodeRewriterStep.cs
```diff
@@ -43,6 +50,35 @@ void ProcessMethod (MethodDefinition method)
...
+				var option = typeof(System.Runtime.CompilerServices.MethodImplOptions);
+
+				noOptAttrArg = (MethodImplOptions) Enum.Parse(option, "NoOptimization");
```
marek-safar: use Mono.Cecil.MethodImplAttributes value directly

src/linker/Mono.Linker.csproj
```diff
@@ -57,6 +57,8 @@
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml" />
+    <Reference Include="System.Reflection" />
```
marek-safar: Please undo this change

src/linker/Mono.Linker.csproj
```diff
@@ -105,6 +107,7 @@
     <Compile Include="Linker.Steps\PreserveCalendarsStep.cs" />
     <Compile Include="Linker.Steps\RemoveFeaturesStep.cs" />
     <Compile Include="Linker.Steps\CodeRewriterStep.cs" />
+    <Compile Include="Linker.Steps\AccessAnnotatorStep.cs" />
```
marek-safar: Where is this file?

test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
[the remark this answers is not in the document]
marek-safar: I agree with Mike this is not the correct way to filter the attribute and in reality we want exact opposite

src/linker/Linker/Driver.cs
```diff
@@ -218,6 +219,10 @@ public void Run (ILogger customLogger = null)
 							}
 							continue;
 
+						case "--annotate-method-access":
```
marek-safar: This needs to be documented and please try to come up with more specific name (this is too vague)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1660,6 +1660,11 @@ void MarkMethodCollection (IList<MethodDefinition> methods)
 				MarkMethod (method);
 		}
 
+		protected void MarkMethodReflected (MethodDefinition method)
+		{
+				Annotations.MarkReflected (method);
```
marek-safar: This should be conditional

src/linker/Linker.Steps/CodeRewriterStep.cs
```diff
@@ -43,6 +50,35 @@ void ProcessMethod (MethodDefinition method)
...
+				AddReflectedAttr (method);
+		}
+
+		void AddReflectedAttr (MethodDefinition method)
+		{
+			// Public methods have non-visible call sites by default, this attribute doesn't
+			// help us at all.
+			//
+			// See mono_aot_can_specialize in aot-compiler.c in mono
+			if (!method.IsPrivate)
```
marek-safar: This check should happen before you populate the list

src/linker/Linker/Driver.cs
```diff
@@ -328,6 +333,9 @@ public void Run (ILogger customLogger = null)
 				foreach (string custom_step in custom_steps)
 					AddCustomStep (p, custom_step);
 
+				if (annotateAccesses)
+					p.AddStepAfter (typeof (CodeRewriterStep), new AccessAnnotatorStep ());
```
marek-safar: Please add tests for the new logic

src/linker/Linker.Steps/CodeRewriterStep.cs
alexanderkyte: I gave that a try before and Cecil reported that it couldn't import the MethodImplAttribute class when doing some emitter pass. I'll look back into it.

alexanderkyte: Still running into cecil issues, cecil doesn't see any of the constructors for that type beyond the copy constructor.

alexanderkyte: Also this: 

https://github.com/mono/linker/blob/fa9ccbdaf6907c69ef1bb117906f8f012218d57f/src/linker/Linker.Steps/CodeRewriterStep.cs#L51

doesn't actually change the emitted method's attributes at all.

alexanderkyte: It's also definitely not tested, because if I remove those lines the tests pass.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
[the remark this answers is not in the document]
alexanderkyte: Oh, nice. I didn't think there already was something out.

I was trying to avoid the size increase of tagging the majority of methods with an attribute. There's definitely soundness gains from doing it the way suggested here though.

src/linker/Linker.Steps/CodeRewriterStep.cs
mrvoorhe: Fyi, the test infrastructure does not assert `ImplAttributes` currently.  See `VerifyMethodKept.VerifyMethodKept`.  We should hook up the same pattern used for `VerifyPseudoAttributes`

src/linker/Linker.Steps/UnseenCallerAnnotateStep.cs
```diff
@@ -0,0 +1,70 @@
...
+				{
+					if (!ref_method.IsConstructor)
+						continue;
+					if (ref_method.Parameters.Count != 0)
+						continue;
+					reflectionMethod = ref_method;
+				}
+				noOptAttr = assembly.MainModule.ImportReference (reflectionMethod);
+				if (noOptAttr == null)
+					throw new Exception("Could not find System.Runtime.CompilerServices.ReflectionBlockedAttribute in BCL.");
```
mrvoorhe: Something to watch out for.

You can run the linker tests on windows.  When you do, the tests are linked against the .net framework assemblies.  I doubt they have this type which I suspect means this exception is going to be hit.

As long as `--annotate-unseen-callers` is opt-in, the solution is simple.  Any tests you add that turn on `--annotate-unseen-callers` will likely need to be ignored on Windows.

test/Mono.Linker.Tests.Cases/Reflection/UnseenCallersAnnotation.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.Reflection;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.Reflection
```
mrvoorhe: I would move this out into it's own top level directory and add a new test method w/ `[TestCaseSource]` in `TestSuites.cs`.    Two reasons

1) To date, the `Reflection` folder has been focused on tests for the detection & marking of reflection usage.  I feel like this is a separate behavior from that.
2) It will be easy to ignore all of these tests on Windows if they all fall under a single `[TestCaseSource]` method .  See https://github.com/mono/linker/pull/506/files#r270871566

src/linker/Linker/LinkContext.cs
```diff
@@ -149,6 +149,9 @@ public bool IgnoreUnresolved
 
 		public CodeOptimizations DisabledOptimizations { get; set; }
 
+        public bool AnnotateUnseenCallers { get; set; }
```
mrvoorhe: Indent

src/linker/Linker.Steps/UnseenCallerAnnotateStep.cs
```diff
@@ -0,0 +1,70 @@
...
+			}
+			var cattr = new CustomAttribute(noOptAttr);
+			method.CustomAttributes.Add (cattr);
+
+			Annotations.Mark(cattr);
+			Annotations.Mark(cattr.AttributeType);
```
mrvoorhe: I believe you will need to call `Resolve()` on `cattr.AttributeType` and `cattr.Constructor` and then mark the `TypeDefinition` and `MethodDefinition` instead.  Marking the `TypeReference` and `MethodReference` will not ensure they survive in all cases

alexanderkyte:
What we can also do @mrvoorhe is to strip out the usage of this attribute in https://github.com/mono/mono/tree/master/mcs/tools/cil-strip if you'd like to have that information available at compile-time, but not at run-time.

alexanderkyte:
@mrvoorhe looks like CI doesn't have that attribute available.

mrvoorhe:
@alexanderkyte I'm guessing the .NET Core builds fail because .NET Core doesn't have the type `System.Runtime.CompilerServices.ReflectionBlockedAttribute` either.  You'll have to ignore your new tests when the tests are running against .NET Core as well.

mrvoorhe:
> What we can also do @mrvoorhe is to strip out the usage of this attribute in https://github.com/mono/mono/tree/master/mcs/tools/cil-strip if you'd like to have that information available at compile-time, but not at run-time.

I don't entirely follow.

If this attribute injection is an opt-in command line option then that works for our needs.

test/Mono.Linker.Tests/TestCases/TestSuites.cs
```diff
@@ -144,6 +144,12 @@ public void CommandLineTests (TestCase testCase)
 			Run (testCase);
 		}
 
+		[TestCaseSource (typeof (TestDatabase), nameof (TestDatabase.CodegenAnnotationTests))]
+		public void CodegenAnnotationTests (TestCase testCase)
+		{
+			Run (testCase);
```
mrvoorhe: Into this method I would add
```
if (Windows)
    Assert.Ignore("These tests are not valid when linking against .NET Framework");

#if NETCOREAPP
    Assert.Ignore("These tests are not valid when linking against .NET Core");
#endif
```

test/Mono.Linker.Tests/TestCases/TestSuites.cs
```diff
@@ -147,6 +147,12 @@ public void CommandLineTests (TestCase testCase)
 		[TestCaseSource (typeof (TestDatabase), nameof (TestDatabase.CodegenAnnotationTests))]
 		public void CodegenAnnotationTests (TestCase testCase)
 		{
+			if (Windows)
```
mrvoorhe: Sorry, I should have been more clear.  When I said `Windows`, I meant "Replace me with the verbose way .NET makes you check if you are on windows"

```
SystemEnvironment.OSVersion.Platform == PlatformID.Win32
```

alexanderkyte: I thought you had some variables inherited from test infrastructure or something. That makes more sense. Thanks

alexanderkyte:
3 tests with CustomAttribute issues. The source for the check that does the DebuggableAttribute assertions say:
```
 // When compiling with roslyn, assemblies get the DebuggableAttribute by default.
test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs:                                      case "System.Diagnostics.DebuggableAttribute":
```

mrvoorhe:
These failures are on master.  One of my PR's that was just merged had a logical conflict with a different PR that was merged while it was open.

I will open a PR to fix the tests.

mrvoorhe:
Here's the fix PR https://github.com/mono/linker/pull/514 .  Once that lands rebase and you should be good to go.

src/linker/Linker/Driver.cs
```diff
@@ -534,6 +541,7 @@ static void Usage (string msg)
 			Console.WriteLine ("  --strip-resources         Remove XML descriptor resources for linked assemblies. Defaults to true");
 			Console.WriteLine ("  --strip-security          Remove metadata and code related to Code Access Security. Defaults to true");
 			Console.WriteLine ("  --used-attrs-only         Any attribute is removed if the attribute type is not used. Defaults to false");
+			Console.WriteLine ("  --annotate-unseen-callers Mark methods without calls through reflection, assist JIT codegen. Defaults to false");
```
marek-safar: I'd suggest this to be more explicit e.g. `--no-reflection-methods` Mark all methods never used using reflection. Defaults to false`

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2227,6 +2232,8 @@ void MarkConstructorsUsedViaReflection (Collection<Instruction> instructions, st
 					Tracer.Push ($"Reflection-{method}");
 					try {
 						MarkMethod (method);
+						if (_context.AnnotateUnseenCallers)
```
marek-safar: You need to add same to remaining UsedViaReflection* methods (e.g. for methods or accessors)

src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -434,6 +434,8 @@ void MarkMethod (MethodDefinition method)
 				Context.LogMessage ($"Duplicate preserve in {_xmlDocumentLocation} of {method.FullName}"); 
 
 			Annotations.Mark (method);
+			if (Context.AnnotateUnseenCallers)
```
marek-safar: Why is it needed?

test/Mono.Linker.Tests.Cases/CodegenAnnotation/UnseenCallersAnnotation.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Mono.Linker.Tests.Cases.CodegenAnnotation
+{
+	[SetupLinkerArgument("--annotate-unseen-callers")]
+	public class UnseenCallersAnnotation
+	{
+		public static void Main()
+		{
+			var obj = new A();
+			var method = typeof(A).GetMethod("FooPrivRefl", BindingFlags.NonPublic);
```
marek-safar: Please add tests for ctors and accessors

src/linker/Linker.Steps/ResolveFromXmlStep.cs
alexanderkyte: The methods named here are methods that may have call sites that we do not see, or they may be reflected.

src/linker/Linker.Steps/UnseenCallerAnnotateStep.cs
```diff
@@ -0,0 +1,73 @@
...
+					throw new Exception("Could not find System.Runtime.CompilerServices.ReflectionBlockedAttribute in BCL.");
+
+				MethodDefinition reflectionMethod = null;
+				foreach (var ref_method in methodImpl.Methods)
+				{
+					if (!ref_method.IsConstructor)
+						continue;
+					if (ref_method.Parameters.Count != 0)
+						continue;
+					reflectionMethod = ref_method;
```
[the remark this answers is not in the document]
alexanderkyte: Hmm, we need the context to look up the type reference, and the assembly to look up the definition. I'm not sure how much we can cache, but I'll look into it.

src/linker/Linker.Steps/UnseenCallerAnnotateStep.cs
```diff
@@ -0,0 +1,73 @@
...
+
+			this.assembly = assembly;
+
+			foreach (var type in assembly.MainModule.Types)
+				ProcessType (type);
+		}
+
+		void ProcessType (TypeDefinition type)
+		{
+			foreach (var method in type.Methods) {
```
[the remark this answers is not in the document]
alexanderkyte: If all of the private methods can have this attribute, I now hoist it and apply it to the type.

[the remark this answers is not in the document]
MichalStrehovsky: The meaning of the attribute is to completely block reflection, not just the private methods on a type - there's nothing in the attribute's name implying it would be for privates only. .NET Native and CoreRT will discard metadata completely when this attribute is applied on a type (we get like [30%](https://github.com/dotnet/corert/pull/3685#issuecomment-303579581) size on disk savings from reflection blocking in general, so aggressively discarding metadata is pretty huge).

There is an attribute with the semantic we would want for this: [DisablePrivateReflectionAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.disableprivatereflectionattribute?view=netframework-4.7.2). I would suggest using that. I know the AttributeUsage doesn't allow using it on types, but tools typically don't care. We could file a CoreFX issue to allow extending the AttributeUsage.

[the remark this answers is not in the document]
alexanderkyte: MichalStrehovsky Did that here:  https://github.com/mono/linker/pull/506/commits/7dfa1ecd4fe446792ad2e0e5779f169be02d02f3

src/linker/Linker.Steps/ReflectionBlockedStep.cs
```diff
@@ -0,0 +1,110 @@
...
+			}
+
+			foreach (var nested in type.NestedTypes)
+				ProcessType (nested);
+		}
+
+		void GetNoOptAttr ()
+		{
+			if (noOptAttr == null) {
+				noOptAttr = assembly.MainModule.ImportReference (GetReflectionBlockedAttr (Context));
```
mrvoorhe: I thought you have to call `assembly.MainModule.ImportReference` once for each `assembly`?  This caching in `noOptAttr` looks like it would prevent that from happening.

alexanderkyte: Ah, good catch. If this has to be regenerated once per assembly, I'll do that where I assign the assembly to the `assembly` object field. I'll push that change shortly.

PR #513, opened 2019-04-02
Clean up SetAction a little

mrvoorhe:
* Rename `LinkContext.SetAction` to `LinkContext.SetActionToDefault` since that's what it is actually doing.  Setting the action to the specified user or core action.

* Add a new `LinkContext.SetAction` that will ensure the assembly is registered and has SafeReadSymbols called.  I hit an issue in our linker where I had early phase steps calling `context.Annotations.SetAction`.  If you call that directly you can get into a situation where `RegisterAssembly` does not call `SafeReadSymbols` because `SeenFirstTime` is returning true.  And when that happens, you end up breaking symbol linking for that assembly.  My hope is that this new `LinkContext.SetAction` is a slightly less error prone way of setting the action on assembly in a step that runs before `LoadReferenceStep`

### DIFF
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -233,7 +233,7 @@ namespace Mono.Linker {
 		{
 			if (SeenFirstTime (assembly)) {
 				SafeReadSymbols (assembly);
-				SetAction (assembly);
+				SetDefaultAction (assembly);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/linker/Linker/LinkContext.cs
```diff
@@ -292,8 +292,14 @@ static AssemblyNameReference GetReference (IMetadataScope scope)
...
 		}
+		
+		public void SetAction(AssemblyDefinition assembly, AssemblyAction action)
+		{
+			RegisterAssembly(assembly);
+			Annotations.SetAction(assembly, action);
+		}
 
-		protected void SetAction (AssemblyDefinition assembly)
+		protected void SetActionToDefault (AssemblyDefinition assembly)
```
marek-safar: Shouldn't that be `SetDefaultAction` ?

mrvoorhe: Fixed.  I was on the fence and almost called it that originally.

mrvoorhe:
Rebased and fixed conflicts

mrvoorhe:
Rebased to pick up the fix for test failures on master.  PR is now green.

PR #514, opened 2019-04-02
Fix test failures on master

mrvoorhe:
The .netcore testing PR started filtering out the Debuggable attribute when asserting.  One of my PR's that had been open needed to be updated to account for this behavior change.


sbomer:
Fixes https://github.com/mono/linker/issues/515.

alexanderkyte:
All checks passed @mrvoorhe

PR #540, opened 2019-04-18
Bump cecil.

vargaz:


vargaz:
The bump seems to fail all CI lanes.

vargaz:
This is not going to work, cecil master has some build changes.

PR #551, opened 2019-04-26
Handle long paths in ILLink.Tasks LinkTask

swaroop-sridhar:
LinkTask invocation involving long-paths (with spaces) failed because
the arguments to dotnet.exe Illink.dll are not within quotes.
This change fixes the problem.

Ideally, the LinkTask should use a response file, instead of the long
command line. However, dotnet.exe currently doesn't support response files.
https://github.com/dotnet/cli/issues/7657

### DIFF
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -123,19 +123,24 @@ namespace ILLink.Tasks
 			set => _illinkPath = value;
 		}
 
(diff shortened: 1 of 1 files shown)


swaroop-sridhar:
CC: @marek-safar

swaroop-sridhar:
@marek-safar can you please review this change, since @sbomer is out of office. Thanks.

CC: @jeffschwMSFT @fadimounir

PR #557, opened 2019-05-01
Match sdk versions

sbomer:
This updates ILLink.Tasks to reference the same version of `System.Reflection.Metadata` and msbuild packages as the SDK, and includes SRM in the package for desktop, the same way that `Microsoft.NET.Build.Tasks` does it.
@nguerrera PTAL.

### DIFF
--- a/src/ILLink.Tasks/ILLink.Tasks.csproj
+++ b/src/ILLink.Tasks/ILLink.Tasks.csproj
@@ -71,10 +71,7 @@
 
     <MSBuild Projects="@(ProjectsToPublish)" Targets="Publish" />
     <ItemGroup>
-      <TfmSpecificPackageFile Include="$(PublishDir)/ILLink.*.dll" PackagePath="tools\$(TargetFramework)" />
-      <TfmSpecificPackageFile Include="$(PublishDir)/Mono.Cecil*.dll" PackagePath="tools\$(TargetFramework)" />
-      <TfmSpecificPackageFile Include="$(PublishDir)/illink.dll" PackagePath="tools\$(TargetFramework)"
-                              Condition=" '$(TargetFramework)' == 'netcoreapp2.0' " />
+      <TfmSpecificPackageFile Include="$(PublishDir)/*.dll" PackagePath="tools\$(TargetFramework)" />
       <TfmSpecificPackageFile Include="$(PublishDir)/*.json" PackagePath="tools\$(TargetFramework)" />
     </ItemGroup>
 
@@ -138,6 +135,7 @@
                       PrivateAssets="All" />
     <PackageReference Include="System.Reflection.Metadata" Version="$(SystemReflectionMetadataVersion)"
                       Condition=" '$(TargetFramework)' == 'net472' "
-                      PrivateAssets="All" />
+                      PrivateAssets="All"
+                      Publish="True" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


src/ILLink.Tasks/ILLink.Tasks.csproj
```diff
@@ -71,10 +71,7 @@
 
     <MSBuild Projects="@(ProjectsToPublish)" Targets="Publish" />
     <ItemGroup>
-      <TfmSpecificPackageFile Include="$(PublishDir)/ILLink.*.dll" PackagePath="tools\$(TargetFramework)" />
-      <TfmSpecificPackageFile Include="$(PublishDir)/Mono.Cecil*.dll" PackagePath="tools\$(TargetFramework)" />
-      <TfmSpecificPackageFile Include="$(PublishDir)/illink.dll" PackagePath="tools\$(TargetFramework)"
-                              Condition=" '$(TargetFramework)' == 'netcoreapp2.0' " />
+      <TfmSpecificPackageFile Include="$(PublishDir)/*.dll" PackagePath="tools\$(TargetFramework)" />
```
sbomer: The specific includes were there originally to avoid including `System.Diagnostics.DiagnosticsSource.dll` and `illink.dll` on `net472`. This is no longer necessary for `illink.dll` because the packagereference to it is conditional on the tfm, but with this change `System.Diagnostics.DiagnosticsSource.dll` will also redistributed for `net472`. I don't actually understand what this dll is for or why it is getting published. @nguerrera, do you know what's the right thing to do with it?

src/ILLink.Tasks/ILLink.Tasks.csproj
```diff
@@ -132,11 +129,13 @@
     <!-- We use private assets for the Microsoft.Build packages to
          prevent them from being published with the tasks dll, because
          these are already a part of the SDK. -->
-    <PackageReference Include="Microsoft.Build.Framework" Version="15.1.1012"
+    <PackageReference Include="Microsoft.Build.Framework" Version="$(MicrosoftBuildFrameworkVersion)"
                       PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.1.1012"
-                      PrivateAssets="All" />
-    <PackageReference Include="System.Reflection.Metadata" Version="1.3.0"
+    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="$(MicrosoftBuildUtilitiesCoreVersion)"
```
nguerrera: Put ExcludeAssets="Runtime" on both msbuild references.

nguerrera: (Probably not the issue, though).

src/ILLink.Tasks/ILLink.Tasks.csproj
nguerrera: Can you send me project.assets.json for this project and a binlog of the publish?

sbomer: I explicitly excluded assets from the `DiagnosticSource` package below. Excluding assets from `NETStandard.Library` in `ILLink.Tasks.csproj` wasn't sufficient, because the assets were coming from an indirect dependency of that package.

sbomer: I'll also try to move https://github.com/mono/linker/pull/471 forward so that we use a `netstandard2.0` version of cecil, which won't have the same problem.

sbomer:
@nguerrera I've addressed your feedback. Would you mind taking another look?

sbomer:
Thanks @nguerrera! @marek-safar this is ready to merge.

Issue #58, opened 2017-04-04
Support list of files for references


ericstj:
It's not always convenient to provide a directory for references.  Some times existing directories have too many other files and copying to a directory is excess work.  Please add the ability to specify references by file.

PR #567, opened 2019-05-14
Resolve assemblies from specific paths

sbomer:
This introduces a command-line option, "-reference <file>", which makes it possible to give the linker full assembly paths to resolve.

LinkTask now uses "-reference" instead of "-d", to prevent the directory resolution behavior from finding dependencies next to inputs. Now it takes an explicit list of inputs, which will be the only files it resolves.

When passing along full paths to each file, the command-line becomes substantially longer. To prevent hitting command length limits, arguments are now passed via a response file (using ToolTask's GenerateResponseFileCommands).

This fixes issues like that mentioned in https://github.com/dotnet/coreclr/issues/24397#issuecomment-490807112 when trying to link WPF apps, which have a `WindowsBase.dll` in the core framework directory and also in the WPF runtime pack:
```
Unhandled Exception: System.ArgumentNullException: Value cannot be null.
         Parameter name: declaringType
            at Mono.Cecil.Mixin.CheckType(Object type, Argument argument)
            at Mono.Cecil.MethodReference..ctor(String name, TypeReference returnType, TypeReference declaringType)
            at Mono.Linker.Steps.TypeMapStep.CreateGenericInstanceCandidate(GenericContext context, TypeDefinition candidateType, MethodDefinition interfaceMethod)
            at Mono.Linker.Steps.TypeMapStep.GetBaseInflatedInterfaceMethodInTypeHierarchy(GenericContext context, TypeDefinition type, MethodDefinition interfaceMethod)
            at Mono.Linker.Steps.TypeMapStep.MapInterfaceMethodsInTypeHierarchy(TypeDefinition type)
            at Mono.Linker.Steps.TypeMapStep.MapType(TypeDefinition type)
            at Mono.Linker.Steps.TypeMapStep.ProcessAssembly(AssemblyDefinition assembly)
            at Mono.Linker.Driver.Execute(String[] args, ILogger customLogger)
```

### DIFF
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -133,11 +133,11 @@ namespace ILLink.Tasks
 
 			if (RootDescriptorFiles != null) {
 				foreach (var rootFile in RootDescriptorFiles)
-					args.Append (" -x ").Append (Quote (rootFile.ItemSpec));
+					args.Append ("-x ").AppendLine (Quote (rootFile.ItemSpec));
 			}
 
 			foreach (var assemblyItem in RootAssemblyNames)
-				args.Append (" -a ").Append (Quote (assemblyItem.ItemSpec));
+				args.Append ("-a ").AppendLine (Quote (assemblyItem.ItemSpec));
 
 			HashSet<string> assemblyNames = new HashSet<string> (StringComparer.OrdinalIgnoreCase);
 			foreach (var assembly in AssemblyPaths) {
@@ -148,13 +148,13 @@ namespace ILLink.Tasks
 				if (!assemblyNames.Add (assemblyName))
 					continue;
 
-				args.Append (" --ref ").Append (Quote (assemblyPath));
+				args.Append ("--ref ").AppendLine (Quote (assemblyPath));
 
 				string action = assembly.GetMetadata ("action");
 				if ((action != null) && (action.Length > 0)) {
-					args.Append (" -p ");
+					args.Append ("-p ");
 					args.Append (action);
-					args.Append (" ").Append (Quote (assemblyName));
+					args.Append (" ").AppendLine (Quote (assemblyName));
 				}
 			}
 
@@ -168,33 +168,33 @@ namespace ILLink.Tasks
 					if (assemblyNames.Contains (assemblyName))
 						continue;
 
-					args.Append (" --ref ").Append (Quote (assemblyPath));
+					args.Append ("--ref ").AppendLine (Quote (assemblyPath));
 
 					// Treat reference assemblies as "skip". Ideally we
 					// would not even look at the IL, but only use them to
 					// resolve surface area.
-					args.Append (" -p skip ").Append (Quote (assemblyName));
+					args.Append ("-p skip ").AppendLine (Quote (assemblyName));
 				}
 			}
 
 			if (OutputDirectory != null)
-				args.Append (" -out ").Append (Quote (OutputDirectory.ItemSpec));
+				args.Append ("-out ").AppendLine (Quote (OutputDirectory.ItemSpec));
 
 			if (ClearInitLocals) {
-				args.Append (" -s ");
+				args.Append ("-s ");
 				// Version of ILLink.CustomSteps is passed as a workaround for msbuild issue #3016
-				args.Append ("LLink.CustomSteps.ClearInitLocalsStep,ILLink.CustomSteps,Version=0.0.0.0:OutputStep");
+				args.AppendLine ("ILLink.CustomSteps.ClearInitLocalsStep,ILLink.CustomSteps,Version=0.0.0.0:OutputStep");
 				if ((ClearInitLocalsAssemblies != null) && (ClearInitLocalsAssemblies.Length > 0)) {
-					args.Append (" -m ClearInitLocalsAssemblies ");
-					args.Append (ClearInitLocalsAssemblies);
+					args.Append ("-m ClearInitLocalsAssemblies ");
+					args.AppendLine (ClearInitLocalsAssemblies);
 				}
 			}
 
 			if (ExtraArgs != null)
-				args.Append (" ").Append (ExtraArgs);
+				args.AppendLine (ExtraArgs);
 
 			if (DumpDependencies)
-				args.Append (" --dump-dependencies");
+				args.AppendLine ("--dump-dependencies");
 
 			return args.ToString ();
 		}
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -12,12 +12,7 @@ namespace ILLink.Tasks
 	{
 		/// <summary>
 		///   Paths to the assembly files that should be considered as
-		///   input to the linker. Currently the linker will
-		///   additionally be able to resolve any assemblies in the
-		///   same directory as an assembly in AssemblyPaths, but this
-		///   behavior should not be relied upon. Instead, work under
-		///   the assumption that only the AssemblyPaths given will be
-		///   resolved.
+		///   input to the linker.
 		///   Each path can also have an "action" metadata,
 		///   which will set the illink action to take for
 		///   that assembly.
@@ -88,14 +83,12 @@ namespace ILLink.Tasks
 			get
 			{
 				if (!String.IsNullOrEmpty (_dotnetPath))
-				{
 					return _dotnetPath;
-				}
+
 				_dotnetPath = Environment.GetEnvironmentVariable (DotNetHostPathEnvironmentName);
 				if (String.IsNullOrEmpty (_dotnetPath))
-				{
 					throw new InvalidOperationException ($"{DotNetHostPathEnvironmentName} is not set");
-				}
+
 				return _dotnetPath;
 			}
 		}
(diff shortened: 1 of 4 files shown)


sbomer:
PTAL @swaroop-sridhar @fadimounir

sbomer:
@marek-safar PTAL.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -132,30 +124,31 @@ protected override string GenerateCommandLineCommands ()
 		{
 			var args = new StringBuilder ();
 			args.Append (Quote (ILLinkPath));
+			return args.ToString ();
+		}
+
+		protected override string GenerateResponseFileCommands ()
```
swaroop-sridhar: 👍

swaroop-sridhar: Can you please make the response file such that there is one argument/switch per line? 
Basically add newlines at the end of each option? 

This makes it easier to inspect the rsp files manually.

sbomer: Done, thanks!

sbomer:
Fixes https://github.com/mono/linker/issues/58.

src/linker/Linker/Driver.cs
```diff
@@ -240,6 +240,10 @@ public void Run (ILogger customLogger = null)
 								disabled_optimizations.Add (opt);
 
 							continue;
+
+						case "--ref":
+							context.Resolver.AddAssemblyPath (GetParam ());
```
marek-safar: I don't understand why we need a different name for something what `-d <path>` already covers

src/linker/Linker/AssemblyResolver.cs
```diff
@@ -68,12 +72,57 @@ public AssemblyResolver (Dictionary<string, AssemblyDefinition> assembly_cache)
...
+			// Try the new resolution behavior. This can't live in the cecil-owned base class.
+			asm = SearchAssemblyPaths (name, _assemblyPaths, parameters);
+			if (asm != null)
+				return asm;
+
+			// Fall back to the base class resolution logic
+			return base.Resolve (name, parameters);
+		}
+
+		private AssemblyDefinition SearchAssemblyPaths (AssemblyNameReference name, IEnumerable<string> assemblyPaths, ReaderParameters parameters)
```
marek-safar: Why base implementation cannot be used?

src/linker/Linker/AssemblyResolver.cs
```diff
@@ -68,12 +72,57 @@ public AssemblyResolver (Dictionary<string, AssemblyDefinition> assembly_cache)
...
+			if (parameters.AssemblyResolver == null)
+				parameters.AssemblyResolver = this;
+
+			return ModuleDefinition.ReadModule (file, parameters).Assembly;
+		}
+#endif
+
+		private AssemblyDefinition ResolveWithPaths (AssemblyNameReference name, ReaderParameters parameters)
+		{
+			// Validate arguments, similarly to how the base class does it.
```
marek-safar: This whole body should be better in Resolve method

src/linker/Linker/Driver.cs
sbomer: Something like `--ref` is necessary to prevent issues with`-d <path>` where different directories have dlls with the same name, but we only want to use one.

Specifically:
Our SDK passes the linker task a specific list of files as input. One of these is `WindowsBase.dll` from a netcoreapp directory. Some other inputs (for WPF apps) live in a windowsdesktop directory, which also contains a `WindowsBase.dll` (that we don't want to use). Before linking, the SDK already does conflict resolution to pick the right input. However, when the linker gets the directory of both inputs, it can resolve the wrong `WindowsBase.dll` when it scans whole directories.

src/linker/Linker/AssemblyResolver.cs
sbomer: The base classes currently only have logic to search in directories (BaseAssemblyResolver from cecil, or DirectoryAssemblyResolver from the linker). To move it to the base class, we would need to do one of:
1. Modify cecil's resolution behavior. I've attempted this in the past, but Jb suggested that the linker should own any resolution logic specific to its purposes.
2. Move it to DirectoryAssemblyResolver but not BaseAssemblyResolver, and support `--ref` only on illink, not on monolinker.
3. Keep it in the child class, extending both base classes.

I opted for the third approach, to keep the command-line consistent between illink/monolinker (I thought this option may be useful for monolinker as well, since it could in theory hit similar issues).

If you prefer, I can also make the change only for illink, and move the change to DirectoryAssemblyResolver only. This is actually what I did initially (see the first commit), before I realized that monolinker wouldn't support the change this way.

src/linker/Linker/AssemblyResolver.cs
sbomer: I've moved it into Resolve.

src/linker/Linker/Driver.cs
marek-safar: Right so the intention is to --ref behave like
`--reference /user/lib/file.dll`? It would be good if the names in the code actually reflected that

sbomer: Yup! Which names? Maybe SearchAssemblyPaths -> SearchFilePaths?

marek-safar: I'd start with the documentation, it should be
`-reference <file>  .....`

similarly, with the resolver entry, replace `AddAssemblyPath ` with `AddReferenceAssembly` or something like that

src/linker/Linker/AssemblyResolver.cs
```diff
@@ -68,12 +72,50 @@ public AssemblyResolver (Dictionary<string, AssemblyDefinition> assembly_cache)
...
+		}
+
 		public override AssemblyDefinition Resolve (AssemblyNameReference name, ReaderParameters parameters)
 		{
 			AssemblyDefinition asm = null;
 			if (!_assemblies.TryGetValue (name.Name, out asm) && (_unresolvedAssemblies == null || !_unresolvedAssemblies.Contains (name.Name))) {
 				try {
-					asm = base.Resolve (name, parameters);
+					// Validate arguments, similarly to how the base class does it.
+					if (name == null)
```
marek-safar: Move the checks to the first line

src/linker/Linker/AssemblyResolver.cs
```diff
@@ -68,12 +72,50 @@ public AssemblyResolver (Dictionary<string, AssemblyDefinition> assembly_cache)
...
+				parameters.AssemblyResolver = this;
+
+			return ModuleDefinition.ReadModule (file, parameters).Assembly;
+		}
+#endif
+
+		private AssemblyDefinition SearchAssemblyPaths (AssemblyNameReference name, IEnumerable<string> assemblyPaths, ReaderParameters parameters)
+		{
+			foreach (var assemblyPath in assemblyPaths) {
+				if (Path.GetFileName (assemblyPath) != name.Name + ".dll")
```
marek-safar: Cache `name.Name + ".dll"` outside the loop

src/linker/Linker/AssemblyResolver.cs
```diff
@@ -56,6 +59,7 @@ public AssemblyResolver ()
 		public AssemblyResolver (Dictionary<string, AssemblyDefinition> assembly_cache)
 		{
 			_assemblies = assembly_cache;
+			_assemblyPaths = new Collection<string> (10) { };
```
marek-safar:
```suggestion
			_assemblyPaths = new Collection<string> ();
```

src/linker/Linker/AssemblyResolver.cs
```diff
@@ -68,12 +72,50 @@ public AssemblyResolver (Dictionary<string, AssemblyDefinition> assembly_cache)
...
+#endif
+
+		private AssemblyDefinition SearchAssemblyPaths (AssemblyNameReference name, IEnumerable<string> assemblyPaths, ReaderParameters parameters)
+		{
+			foreach (var assemblyPath in assemblyPaths) {
+				if (Path.GetFileName (assemblyPath) != name.Name + ".dll")
+					continue;
+				try {
+					return GetAssembly (assemblyPath, parameters);
+				} catch (System.BadImageFormatException) {
```
marek-safar: nit: System. is redundant

src/linker/Linker/AssemblyResolver.cs
marek-safar: Please rename it the to `ResolveFromReferences (AssemblyNameReference name, Collection<string> references, ReaderParameters parameters)`

src/linker/Linker/DirectoryAssemblyResolver.cs
```diff
@@ -50,7 +50,7 @@ public virtual AssemblyDefinition Resolve (AssemblyNameReference name, ReaderPar
 			if (name == null)
```
marek-safar: Aren't the ArgumentNullException check now redundant?

sbomer: They're redundant in the sense that the base class performs the same check, but also necessary to prevent passing null to SearchAssemblyPaths (renamed to ResolveFromReferences).
You're right that we already deref name in Resolve,  so maybe that's not strictly necessary, but I still feel that it's safer to keep both checks (moved to the beginning of Resolve).

src/linker/Linker/Driver.cs
sbomer: Done!

cup:
https://github.com/dotnet/coreclr/issues/24397

https://github.com/mono/linker/pull/567
https://github.com/mono/linker/pull/582

PR #568, opened 2019-05-14
Remove DiagnosticSource packaging workaround

sbomer:
With the latest cecil update in https://github.com/mono/linker/pull/471, we are using SDK-style projects and building cecil for netstandard2.0, so the workaround for https://github.com/dotnet/sdk/issues/3194 is no longer necessary.


sbomer:
@marek-safar PTAL.

PR #572, opened 2019-05-20
Allow overriding assembly action on root assemblies

sbomer:
When rooting assemblies via '-a' or '-r', ResolveFromAssemblyStep sets a default action depending on whether the assembly is a library or exe. This default takes precedence over any assembly action passed for that assembly via the command line using '-p <action> <assemblyname>'.
This change makes it possible to override the default assembly action even for root assemblies.

This is useful especially in cases (like we want to do for .NET Core) where we'd like to choose a conservative default that keeps the entire app assembly, only trimming framework code.

@marek-safar PTAL

### DIFF
--- a/src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
+++ b/src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
@@ -86,8 +86,10 @@ namespace Mono.Linker.Steps
 
 		public static void ProcessLibrary (LinkContext context, AssemblyDefinition assembly, RootVisibility rootVisibility = RootVisibility.Any)
 		{
-			var action = rootVisibility == RootVisibility.Any ? AssemblyAction.Copy : AssemblyAction.Link;
-			context.SetAction (assembly, action);
+			if (!context.Actions.Contains (assembly.Name.Name)) {
+				var action = rootVisibility == RootVisibility.Any ? AssemblyAction.Copy : AssemblyAction.Link;
+				context.SetAction (assembly, action);
+			}
 
 			context.Tracer.Push (assembly);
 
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -86,8 +86,10 @@ protected virtual void ProcessLibrary (AssemblyDefinition assembly)
 
 		public static void ProcessLibrary (LinkContext context, AssemblyDefinition assembly, RootVisibility rootVisibility = RootVisibility.Any)
 		{
-			var action = rootVisibility == RootVisibility.Any ? AssemblyAction.Copy : AssemblyAction.Link;
-			context.SetAction (assembly, action);
+			if (!context.Actions.Contains (assembly.Name.Name)) {
+				var action = rootVisibility == RootVisibility.Any ? AssemblyAction.Copy : AssemblyAction.Link;
+				context.SetAction (assembly, action);
```
marek-safar: SetAction does more than just setting the action. I think better fix could be to leave this part untouched and modify SetAction to something like this

```c#
		public void SetAction (AssemblyDefinition assembly, AssemblyAction defaultAction)
		{
			RegisterAssembly (assembly);

			if (!_actions.TryGetValue (assembly.Name.Name, out AssemblyAction action))
				action = defaultAction;

			Annotations.SetAction (assembly, action);
		}
```

sbomer: Great suggestion, thanks! I had thought Context.SetAction was used elsewhere, but now I realize that we mostly call Annotations.SetAction, and *Context.SetAction* is used only in these two places.

PR #579, opened 2019-05-23
Revert "Make sure _needAddBypassNGenStep is false if the assembly act…

sbomer:
…ion is overridden by specifying more than once"

This reverts commit 7dce9327c11798b4811d2b376a8fcbc2fc399f2e.

The logic was originally changed so that parameters like `-c addbypassngen -c copy` would not result in an unnecessary AddBypassNGen step being added to the linker pipeline. However, we need to support cases like `-c addbypassngen -p copy MyAssembly`, where one assembly gets a different action from the default. In this case, we still need the addbypassngen step.

Reverting this commit will reintroduce the unnecessary AddBypassNGen step that the original change was designed to prevent for cases like `-c addbypassngen -c copy`. However, because the step itself checks the assembly action, the unnecessary step doesn't change the linker behavior.

@cshung PTAL

### DIFF
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -466,6 +466,9 @@ namespace Mono.Linker {
 		AssemblyAction ParseAssemblyAction (string s)
 		{
 			var assemblyAction = (AssemblyAction)Enum.Parse(typeof(AssemblyAction), s, true);
+			// The AddBypassNGenStep is necessary if any actions (default or per-assembly) are AddBypassNGen(Used).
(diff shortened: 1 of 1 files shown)


cshung:
I talk to @sbomer, I agree the change is the right fix.

The only change I would recommend is to add a comment in `Mono.Linker.Driver.ParseAssemblyAction()` referencing `Mono.Linker.Steps.AddBypassNGenStep.ProcessAssembly()`, so to prevent this issue get _fixed_ again.

sbomer:
Thanks @cshung, I've added a clarifying comment.
@marek-safar PTAL

PR #582, opened 2019-05-24
Use "-reference" in LinkTask

sbomer:
Left-over from https://github.com/mono/linker/pull/567/files. I believe this is what's causing https://github.com/dotnet/coreclr/issues/24397#issuecomment-495720076. @marek-safar PTAL


cup:
https://github.com/dotnet/coreclr/issues/24397
https://github.com/mono/linker/pull/567

https://github.com/mono/linker/pull/582

PR #583, opened 2019-05-25
Avoid overwriting outputs whose guid hasn't changed.

vargaz:


src/linker/Linker.Steps/OutputStep.cs
```diff
@@ -103,7 +103,34 @@ protected virtual void WriteAssembly (AssemblyDefinition assembly, string direct
...
+			var prevGuid = Guid.Empty;
+			if (File.Exists (fileName)) {
+				try {
+					var prevModule = ModuleDefinition.ReadModule (fileName);
+					prevGuid = prevModule.Mvid;
+				} catch {
+				}
+			}
+			if (assembly.MainModule.Mvid != prevGuid) {
+				if (Context.LogMessages)
```
mrvoorhe: I believe there is something like  Context.Logger.LogMessage that should be used instead of console write line.

src/linker/Linker.Steps/OutputStep.cs
```diff
@@ -103,7 +103,32 @@ protected virtual void WriteAssembly (AssemblyDefinition assembly, string direct
...
+			//
+			var prevGuid = Guid.Empty;
+			if (File.Exists (fileName)) {
+				try {
+					var prevModule = ModuleDefinition.ReadModule (fileName);
+					prevGuid = prevModule.Mvid;
+				} catch {
+				}
+			}
+			if (assembly.MainModule.Mvid != prevGuid) {
```
mrvoorhe: Could `assembly.MainModule.Mvid` ever return `Guid.Empty`?  If it's plausible, then that would cause problems when `File.Exists` returns false or when the `try` block above encounters an exception.

mrvoorhe:
Can you wait for https://github.com/mono/linker/pull/586 to land, rebase, and then add test(s)?

Here's one you could add after https://github.com/mono/linker/pull/586 lands.

```
		[Test]
		public void UnchangedAssemblyIsNotOverwritten()
		{
			var testCase = CreateIndividualCase (typeof (DeterministicMvidWorks));
			var runner = new TestRunner (new ObjectFactory ());
			var result = runner.Run (testCase);

			var lastWrite = File.GetLastWriteTime(result.OutputAssemblyPath) - new TimeSpan(0, 1, 0);
			File.SetLastWriteTime(result.OutputAssemblyPath.ToString(), lastWrite);

			var result2 = runner.Relink(result);
			
			Assert.That(result2.OutputAssemblyPath, Is.EqualTo(result.OutputAssemblyPath));
			Assert.That(File.GetLastWriteTime(result2.OutputAssemblyPath), Is.EqualTo(lastWrite));
		}
```

mrvoorhe:
I was thinking, this change would open the door to some odd behaviors if you had different symbols settings between runs.

For example. 

1) Run 1 does not have symbol linking enabled.  Run 2 has symbol linking enabled.  I suspect in this case run 2 would not end up with symbol files in the output directory.

2) Instead of symbols on/off, consider what would happen if the type of symbol file changed between runs.  I suspect the second run would not have the expected symbol file.

vargaz:
Old.

PR #587, opened 2019-05-28
Make deterministic output the default.

### LINKED ISSUE #1342: Change deterministic output  to be the default linker mode
We have already `--deterministic` option but it's not enabled by default. This is not desirable as the expectation of most developers is to produce reproducible build by default.

The logic behind the option needs to be changed to be enabled by default.

vargaz:
Fixes #1342


mrvoorhe:
@marek-safar Is there any reason to continue to maintain an option for retaining the original mvid?  We don't make use of the ability so I don't have a strong opinion either way.

mrvoorhe:
@vargaz You'll need to update the tests I landed with https://github.com/mono/linker/pull/586

marek-safar:
I am a little bit worried about changing the default but this is a quite esoteric option so if you are fine with the change we can tweak it.

@vargaz could you update the failing tests?

vargaz:
The CI lanes fail with:
##[error]eng/common/tools.sh(426,): error : Build failed (exit code '1').
##[error]Bash exited with code '1'.
##[section]Finishing: Build illink.sln Release
and no other info.

mrvoorhe:
I think there are problems with deterministic and symbols.   I'll post a PR with repro tests shortly.

Does Roslyn default to `/deterministic`?  If it doesn't, that may be an argument for not turning on deterministic by default.  Along with it maybe not being ready for prime time.

mrvoorhe:
I spoke with @jbevain and he suggested I file cecil bugs.

https://github.com/jbevain/cecil/issues/582
https://github.com/jbevain/cecil/issues/583

Here is a PR with more tests including ignored tests that encounter the cecil bugs.

https://github.com/mono/linker/pull/591

ViktorHofer:
@marek-safar what's the status of this PR?

src/linker/Linker/Driver.cs
```diff
@@ -245,14 +244,9 @@ public void Run (ILogger customLogger = null)
 							context.Resolver.AddReferenceAssembly (GetParam ());
 							continue;
 
-						case "--new-mvid":
```
vargaz: Existing options shouldn't be removed, they should be ignored.

vargaz:
Related tests fail on CI.

vargaz:
What should we do with this ?

marek-safar:
Testing change #947

vargaz:
Obsolete.

ViktorHofer:
Hi, why is this obsolete?

ViktorHofer:
@marek-safar is this now implemented?

marek-safar:
Not yet but we have more up to date PR linked to this one

PR #590, opened 2019-05-29
Add custom logic for reflection-emit to RemoveFeaturesStep.

baulig:
When `--exclude-feature sre` is provided, then we rewrite the `System.RuntimeType.MakeTypeBuilderInstantiation` to throw.

### DIFF
--- a/src/linker/Linker.Steps/RemoveFeaturesStep.cs
+++ b/src/linker/Linker.Steps/RemoveFeaturesStep.cs
@@ -59,10 +59,12 @@ namespace Mono.Linker.Steps
 			}
 
 			if (FeatureSRE) {
-				if (type.FullName == "System.RuntimeType") {
+				if (type.Namespace == "System" && type.Name == "RuntimeType") {
 					foreach (var method in type.Methods) {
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Steps/RemoveFeaturesStep.cs
```diff
@@ -57,6 +58,15 @@ void ProcessType (TypeDefinition type)
 					ExcludeEventSourceImplementation (type);
 			}
 
+			if (FeatureSRE) {
+				if (type.FullName == "System.RuntimeType") {
```
marek-safar: Don't use FullName

src/linker/Linker.Steps/RemoveFeaturesStep.cs
```diff
@@ -57,6 +58,15 @@ void ProcessType (TypeDefinition type)
...
+					foreach (var method in type.Methods) {
+						if (method.Name == "MakeTypeBuilderInstantiation")
+							Annotations.SetAction (method, MethodAction.ConvertToThrow);
```
marek-safar: I think you could break after this

PR #599, opened 2019-06-01
Bump cecil.

vargaz:


vargaz:
This picks up the determinstic guid fixes.

Issue #612, opened 2019-06-09
Linker error "member is declared in another module and needs to be imported" after PR #502


akoeplinger:
We're seeing this issue in the Xamarin.iOS integration of the latest linker sources: the linker throws an exception like `Member 'System.Boolean _unregistered' is declared in another module and needs to be imported`.

I tracked it down to #502, which also seems to have caused other issues like https://github.com/mono/linker/issues/541.

Unfortunately I don't have time to investigate this right now and talking with @mrvoorhe on Slack he doesn't either so we agreed it's best to disable this new feature by default for now.

marek-safar:
I tried to reproduce it but could not. We have also re-enabled the optimization and everything seems to be fine now so closing.

PR #613, opened 2019-06-09
Disable Unreachable Bodies feature by default

akoeplinger:
See https://github.com/mono/linker/issues/612

/cc @marek-safar

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/NoKeptCtor/UnusedTypeHasExplicitInterfacePropertyPreservedViaXml.cs
+++ b/test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/NoKeptCtor/UnusedTypeHasExplicitInterfacePropertyPreservedViaXml.cs
@@ -1,6 +1,8 @@
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
 
 namespace Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType.NoKeptCtor {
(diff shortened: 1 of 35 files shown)


test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/NoKeptCtor/UnusedTypeHasExplicitInterfacePropertyPreservedViaXml.cs
```diff
@@ -1,6 +1,8 @@
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
 
 namespace Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType.NoKeptCtor {
+	[SetupLinkerArgument ("--disable-opt", "-unreachablebodies")]
```
mrvoorhe: I don't love this double negative way of turning the option back on.  I was thinking a new `enable-opt` command line flag would be more intuitive to use.  That said, this is hopefully a one off temporary situation so I think it's OK.

marek-safar: I agree and I'm not a fan of double negation either. I'd suggest the same as @mrvoorhe. Introduce `-enable-opt` as an undocumented option for now and use it in the tests.

akoeplinger: done

PR #643, opened 2019-07-03
Fix clearinitlocals step

sbomer:
- Adjust to the renamed parameter
- Move the step to illink.dll.

When we ran illink as an msbuild task, Type.GetType would load the
custom step from a dll next to the task assembly, due to MSBuild's
assembly resolution behavior. Now that we run illink.dll as a separate
process, we don't get this behavior automatically any more.

@marek-safar @jkotas


marek-safar:
I'd much prefer this external step to be converted to standard non-default optimizations (enabled by --enable-opt)

sbomer:
@marek-safar I made the requested change - PTAL.

PR #657, opened 2019-07-12
Root interface impls

sbomer:
This makes ResolveFromAssemblyStep the source of truth for what is considered "public API" when linking a library, by marking interface implementations up-front. This needs to be done carefully since the implications of marking the interfaceimpl (marking custom attributes on it, and marking the interface type itself) are usually done later, in MarkStep, which has optional optimizations like only marking custom attributes for used attribute types.

If in the future we add features like base type sweeping, similar adjustments will probably need to be made in ResolveFromAssemblyStep to keep base types when marking public surface area.

I think this is almost ready to go - I need to do some more validation for custom attributes on the .interfaceimpls.


test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/InterfaceTypeInOtherUsedOnlyByLinkedAssembly.cs
```diff
@@ -0,0 +1,29 @@
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+using Mono.Linker.Tests.Cases.Inheritance.Interfaces.Dependencies;
+
+namespace Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType {
+	[SetupCompileBefore ("skipped.dll", new [] {typeof (InterfaceTypeInOtherUsedOnlyByCopiedAssembly_Link)})]
+	[SetupCompileBefore ("link.dll", new [] {typeof (InterfaceTypeInOtherUsedOnlyByCopiedAssembly_Copy)}, new [] {"skipped.dll"})]
+
+	[SetupLinkerAction ("link", "link")]
+	[SetupLinkerAction ("skip", "skipped")]
```
mrvoorhe: I would use `copy` instead of `skip`.  `copy` should still reproduce the problem and the assembly will actually exist in the output directory

sbomer:
This needs more work. The change https://github.com/mono/linker/pull/657/files#diff-83d9d9aa53da0b0737513a2561f82bc8L2231 seems to be causing perf issues with larger link sets (like the framework). I'm thinking about changing how we mark interfaceimpls to avoid iterating over all types with interfaces (and all impls on used types) in `ProcessMarkedTypesWithInterfaces`. Maybe we should instead track `.interfaceimpl`s that we need to revisit.

sbomer:
@marek-safar PTAL

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/InterfaceTypeInOtherUsedOnlyByLinkedAssembly.cs
```diff
@@ -0,0 +1,29 @@
...
+	[SetupCompileBefore ("link.dll", new [] {typeof (InterfaceTypeInOtherUsedOnlyByCopiedAssembly_Copy)}, new [] {"copied.dll"})]
+
+	[SetupLinkerAction ("link", "link")]
+	[SetupLinkerAction ("copy", "copied")]
+	[SetupLinkerArgument ("-r", "link")]
+
+	[KeptMemberInAssembly ("link.dll", typeof (InterfaceTypeInOtherUsedOnlyByCopiedAssembly_Copy.A), "Method()")]
+	[KeptMemberInAssembly ("link.dll", typeof (InterfaceTypeInOtherUsedOnlyByCopiedAssembly_Copy.B), "Method()")]
+	[KeptMemberInAssembly ("link.dll", typeof (InterfaceTypeInOtherUsedOnlyByCopiedAssembly_Copy.B), "Method2()")]
+	[ExpectInterfaceTypeReferenceInAssembly ("link.dll", typeof (InterfaceTypeInOtherUsedOnlyByCopiedAssembly_Copy.A), "copied.dll", typeof (InterfaceTypeInOtherUsedOnlyByCopiedAssembly_Link.IFoo))]
```
marek-safar: Why is this new logic necessary, it looks to me like you could just use existing KeptInterfaceAttribute

src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -164,24 +176,31 @@ static void MarkType (LinkContext context, TypeDefinition type, RootVisibility r
...
+			Annotations.Mark (_assembly.EntryPoint.DeclaringType);
 
-			MarkMethod (Context, assembly.EntryPoint, MethodAction.Parse, RootVisibility.Any);
+			MarkMethod (Context, _assembly.EntryPoint, MethodAction.Parse, RootVisibility.Any);
 
 			Tracer.Pop ();
 		}
 
-		static void MarkFields (LinkContext context, Collection<FieldDefinition> fields, RootVisibility rootVisibility)
+		void MarkFields (LinkContext context, Collection<FieldDefinition> fields, RootVisibility rootVisibility)
```
marek-safar: Why was this change needed?

src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -120,22 +123,29 @@ public static void ProcessLibrary (LinkContext context, AssemblyDefinition assem
...
 			}
 
 			context.Tracer.Pop ();
 		}
 
-		static void MarkType (LinkContext context, TypeDefinition type, RootVisibility rootVisibility)
+		bool MarkType (LinkContext context, TypeDefinition type, RootVisibility rootVisibility)
 		{
+			if (type.IsInterface) {
+				if (_markedInterfaces == null)
```
marek-safar: I don't like this logic hear, we are spreading Marking logic into a different step which will be hard to keep in sync

src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -164,24 +176,31 @@ static void MarkType (LinkContext context, TypeDefinition type, RootVisibility r
 			if (type.HasNestedTypes)
 				foreach (var nested in type.NestedTypes)
 					MarkType (context, nested, rootVisibility);
+			if (type.HasInterfaces)
```
marek-safar: I get this is the only change from this whole PR but I am not sure this is right. Why do you need to mark the base types explicitly here and not to do it in standard MarkStep?

sbomer: MarkStep won't mark the interfaces unless it sees that the type implementing it has been instantiated. We need some way to cause interface impls to be kept if they are considered part of the public surface area. Since we already have some logic in ResolveFromAssemblyStep to deal with marking public surface area based on visibility, I thought it made the most sense to do it here.

src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
sbomer: This was left over from a previous commit - will remove, thanks!

src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
sbomer: I think this is easier to maintain than having MarkStep deal with both the public surface area "roots" and the conditional interface marking - see my other comment. LMK what you think.

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/InterfaceTypeInOtherUsedOnlyByLinkedAssembly.cs
sbomer: Two reasons: we need a different attribute to check for the presence of an interface in a different assembly, and because checking the typeref without resolving it allows us to check for an interface impl when the interface is in a skip assembly. However since I've changed the interface assembly to "copy" per @mrvoorhe's suggestion I can use `KeptInterfaceOnTypeInAssembly` instead.

src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
marek-safar: I think we need to think about a different approach. You are trying to disable optimizations we are added and will keep adding to make linking smarter/better. With this approach, it will be constant work of undoing improvements in marking step (e.g. if we introduce base types reduction, or replacing interfaces with objects, etc).

Secondly, even this change is incomplete. What about cases where an interface is used as a constraint, or as attribute argument or typeof expression?

I think it'd be better to have for this mode a new step which replaces `ResolveFromAssemblyStep` and `MarkStep` as used now for `-r` called for example `MarkVisibleSurfaceStep`

sbomer: I'm not sure I understand the first concern - this doesn't disable any of the optimizations and they will still kick in for interface impls that aren't considered public surface area (or for all interface impls when linking apps using `-a`). For the scenario where we're linking individual corefx libraries, markstep needs to be told what to keep as public surface area.

> an interface is used as a constraint, or as attribute argument

Good point, I can work on adding those as long as we agree on the approach.

> typeof expression

I'm not sure that would be considered public surface area of a library, thoughts?

> a new step which replaces ResolveFromAssemblyStep and MarkStep as used now for -r

To me this seems like it would cause more duplication than the current approach - for example the logic for [rooting](https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/ResolveFromAssemblyStep.cs#L97) exported types up-front is shared between `-r` and `-a`.

marek-safar: My point is that when we add more optimizations which change metadata we'll have to undo them here and therefore would be easier to just have a different pipeline from this "mode".

>I'm not sure that would be considered public surface area of a library, thoughts?

That's exactly the point. I think nobody knows that right now and it's very hard to get that from the code. You are running linker as IL optimizer which I think needs more thinking about how to approach going forward.

> To me this seems like it would cause more duplication 

I don't think there will be any duplication, we can refactor any code as we need.

> for example the logic for rooting  exported types up-front is shared
We should never mark exported types upfront when in linking mode unless they are needed.

I am happy to talk about this in more details.

sbomer: I started trying the alternative we discussed in https://github.com/mono/linker/pull/793.

marek-safar:
Related to #1610

Issue #651, opened 2019-07-09
Wpf applications can't be published as a trimmed single exe


sbomer:
https://github.com/dotnet/coreclr/issues/25334#issue-459511455 from @Lakritzator

> 
> I'm not sure if this is the correct repository for this.
> 
> When using dotnet sdk 3.0.100-preview7-012564 (and a lot of builds before this) I get a crash when starting a publishing a WPF application as a trimmed single exe.
> 
> An example repository is located here: https://github.com/Lakritzator/WpfPublishTrimmedCrash
> It's a "dotnet new wpf" with a textbox added, as without it the error is not visible.
> 
> Running the following on the project:
> ```
> dotnet publish WpfTest.csproj -f netcoreapp3.0 -c Release /p:PublishSingleFile=true /p:PublishTrimmed=true /p:RuntimeIdentifier=win-x64
> ```
> 
> Than starting the WpfText.exe which is located in bin\Release\netcoreapp3.0\win-x64\publish
> will not start the application, and in the eventlog I can see the following error:
> ```
> Application: WpfTest.exe
> CoreCLR Version: 4.700.19.32201
> .NET Core Version: 3.0.0-preview7-27822-04
> Description: The process was terminated due to an unhandled exception.
> Exception Info: System.TypeInitializationException: The type initializer for 'System.Windows.Media.FontFamily' threw an exception.
>  ---> System.TypeInitializationException: The type initializer for 'MS.Internal.FontCache.DWriteFactory' threw an exception.
>  ---> System.InvalidCastException: Specified cast is not valid.
>    at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode, IntPtr errorInfo)
>    at MS.Internal.Text.TextInterface.Native.Util.ConvertHresultToException(Int32 hr)
>    at MS.Internal.Text.TextInterface.Factory.Initialize(FactoryType factoryType)
>    at MS.Internal.Text.TextInterface.Factory..ctor(FactoryType factoryType, IFontSourceCollectionFactory fontSourceCollectionFactory, IFontSourceFactory fontSourceFactory)
>    at MS.Internal.Text.TextInterface.Factory.Create(FactoryType factoryType, IFontSourceCollectionFactory fontSourceCollectionFactory, IFontSourceFactory fontSourceFactory)
>    at MS.Internal.FontCache.DWriteFactory..cctor()
>    --- End of inner exception stack trace ---
[...]

Issue #639, opened 2019-06-28
Linker removes data from "copy" assemblies


sbomer:
The "copy" action currently sometimes removes assemblyrefs and redirects references to removed type forwarders. This causes issues when there's non-IL data in an assembly that we want to copy as-is. For example, see https://github.com/dotnet/coreclr/issues/25334#issuecomment-506836765.

If we don't support this, there will be workarounds such as passing "skip" and manually copying over the file, or using '-reference' to specify all inputs and prevent resolving this reference. However this is a usability issue, so we may want to consider doing something more.

sbomer:
More context: the issue we hit wasn't actually with non-IL data - it was with C++/CLI assemblies relying on the layout of the IL. For cases like this where reading/writing the assembly with cecil will break invariants that consumers rely upon, we should consider supporting a true "copy" mode that keeps the original assembly intact. This would also provide a way for users to tell the linker to keep an assembly as-is (similar to fully-preserved types) while still analyzing it to discover dependencies.

Seems to me that "copy" was originally intended to behave like I described, but it has drifted over time. @mrvoorhe, @marek-safar what do you think of this? Also relevant is the discussion in https://github.com/mono/linker/issues/628.

sbomer:
@vatsan-madhavan asked in https://github.com/mono/linker/pull/658#issuecomment-515529461:
> Asking for my edification: Does inspecting an assembly using Mono.Cecil API's rewrite the assembly - or is the rewriting specific to the linker implementation? 

Just inspecting with Cecil won't rewrite it, but raising it to the cecil object model and writing it back to disk will. I think the fix will be in the linker - to only analyze using cecil, but not write using cecil. There might be a longer term fix to enable cecil to preserve invariants like the ordering of the IL data, but I think the immediate issue is in the linker. Specifically, the description of the "copy" action behavior is what we need, but isn't currently what actually happens.
https://github.com/mono/linker/blob/6071e0a4c8506db06b7ec1873d0b30e31b0cc870/src/linker/Linker/AssemblyAction.cs#L35

sbomer:
Won't it still rewrite copy assemblies when it removes references in https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/SweepStep.cs#L225?

PR #658, opened 2019-07-13
Skip C++/CLI assemblies

sbomer:
This will skip linking of C++/CLI assemblies. See https://github.com/mono/linker/issues/651 for context.

There are potentially some issues with skipping the C++/CLI assembly entirely - any calls that the assembly makes to .NET dlls will no longer be analyzed. This may result in some required dependencies being dropped, though I don't know how likely that is in practice. The ideal fix would be to address https://github.com/mono/linker/issues/639 (or to preserve the ordering of the relevant C++/CLI data when linking as mentioned in https://github.com/mono/linker/issues/651#issuecomment-510689413), which would still result in the assembly being analyzed.

I marked this do not merge until we have some more confidence that this is the right fix.

### DIFF
--- a/src/ILLink.Tasks/Utils.cs
+++ b/src/ILLink.Tasks/Utils.cs
@@ -1,5 +1,6 @@
 using System;
 using Mono.Cecil;
+using System.Linq;
 
 public static class Utils
 {
@@ -7,9 +8,16 @@ public static class Utils
 	{
 		try {
 			ModuleDefinition module = ModuleDefinition.ReadModule (fileName);
-			return true;
+			return !IsCPPCLIAssembly (module);
 		} catch (BadImageFormatException) {
 			return false;
 		}
+	}
+
+	private static bool IsCPPCLIAssembly (ModuleDefinition module)
+	{
+		return module.Types.Any(t =>
+			t.Namespace == "<CppImplementationDetails>" ||
+			t.Namespace == "<CrtImplementationDetails>");
 	}
 }


sbomer:
@vatsan-madhavan I'm going to try to push this change into the SDK seeing as we are late in the release and don't have much time to wait. If later we determine a better way to check for C++/CLI assemblies, I'll make the update then.

@marek-safar PTAL.

vatsan-madhavan:
@sbomer , lgtm. We should get confirmation from @tgani-msft when he is back that the C++/CLI detection being used here will remain a reliable contract/pattern over time, and if necessary, change the detection logic based on his feedback.

ericstj:
Why skip entirely?  Just don't rewrite it, but still analyze its dependencies.

sbomer:
I'd like to do that in the future, but the fix will take some more discussion: https://github.com/mono/linker/issues/639. The linker supposedly is able to do this, but in reality rewrites more often than you might expect.

vatsan-madhavan:
> I'd like to do that in the future, but the fix will take some more discussion: #639. The linker supposedly is able to do this, but in reality rewrites more often than you might expect.

Asking for my edification: Does inspecting an assembly using Mono.Cecil API's rewrite the assembly - or is the rewriting specific to the linker implementation?

sbomer:
@vatsan-madhavan I answered your question in https://github.com/mono/linker/issues/639#issuecomment-515532049.

PR #681, opened 2019-07-30
Refactor reflection marking logic to be more precise

marek-safar:


test/Mono.Linker.Tests.Cases/Reflection/MightKeepExtraThings.cs
```diff
@@ -3,7 +3,9 @@
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 
 namespace Mono.Linker.Tests.Cases.Reflection {
-	public class MightKeepExtraThings {
+	[IgnoreTestCase ("Reflection tracking for simple locals loads")]
```
mrvoorhe: Why add `IgnoreTestCase`?  I don't see any harm in asserting the current expected behavior even if we could do better in the future.

marek-safar: Oh, do we have an attribute to mark expected failure?

mrvoorhe: No we don't.  I didn't think it would fail.  You removed the some `[Kept]` attributes so I thought you had adjusted the expected results to match the new behavior.

PR #774, opened 2019-10-02
Adds processing of satellite assemblies

### LINKED ISSUE #739: Mono Linker doesn't link resources for Blazor
I'm trying to use this awesome [Humanizer](/Humanizr/Humanizer) library in my blazor project. But that library produces resource dll files into **/bin** folder like below:
- `Humanizer.dll`
- `Humanizer.resources.dll`
- `af/Humanizer.resources.dll`
- `ar/Humanizer.resources.dll`
- `bg/Humanizer.resources.dll`
...
- `{culture}/Humanizer.resources.dll`

but when I compile my blazor application, I saw only following libraries are linked:
`Humanizer.dll`
`Humanizer.resources.dll`

But none of `{culture}/Humanizer.resources.dll` are linked or added into **blazor.boot.json**

I can not solve that problem with **Linker.xml** because all the resources are in same namespace and it's `Humanizer.resources.dll`.

Is there any suggestion for solution?

marek-safar:
Fixes #739

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1821,7 +1821,10 @@ namespace Mono.Linker.Steps {
 			if (ShouldMarkAsInstancePossible (method))
 				MarkRequirementsForInstantiatedTypes (method.DeclaringType);
 
-			if (IsPropertyMethod (method))
+			if (method.IsConstructor) {
+				if (!Annotations.ProcessSatelliteAssemblies && KnownMembers.IsSatelliteAssemblyMarker (method))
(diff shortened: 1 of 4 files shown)


JustArchi:
@marek-safar I guess there is no way to upgrade mono-linker bundled in .NET Core SDK to include this change?

Is there any confirmation which .NET Core SDK version gets this fix included?

Thank you in advance for answering.

marek-safar:
There is no simple way but you could build it yourself and replace the dll but it's nontrivial. You can fill issue in core-setup to refresh linker dependency but right now I think it's bellow the bar for 3.1 release

PR #777, opened 2019-10-04
Add detection for single member lookup methods in RuntimeReflectionExtensions

### LINKED ISSUE #55: Support for RuntimeReflectionExtensions.GetRuntimeMethod
Reading on Xamarin's IOS linking (https://developer.xamarin.com/guides/ios/advanced_topics/linker/), and I'm guessing this is the tool they use to link (right?), they write that reflection is not supported by the linker, but if nameof is used (see also discussion here: https://github.com/autofac/Autofac/pull/841) can we add support for the linker to assume that the method in the nameof is being also being used and not strip it out?

marek-safar:
Fixes #55


marek-safar:
build

PR #788, opened 2019-10-14
Don't crash on unresolved assembly

mrvoorhe:
We have tests that use nunit, which leads to System.Windows.Forms being processed which leads to an attempt at resolving some type that is in `WindowsBase`.  I don't think it matters that we failed to resolve these types here.  Just ignore and continue.

### DIFF
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -281,7 +281,14 @@ namespace Mono.Linker.Steps {
 				if (tr.IsWindowsRuntimeProjection)
 					continue;
 
-				var td = tr.Resolve ();
+				TypeDefinition td;
+				try {
+					td = tr.Resolve ();
+				} catch (AssemblyResolutionException) {
+					// Don't crash on unresolved assembly
+					continue;
+				}
+
 				// at this stage reference might include things that can't be resolved
 				// and if it is (resolved) it needs to be kept only if marked (#16213)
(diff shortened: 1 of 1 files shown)


sbomer:
We already have a flag `--skip-unresolved` that is supposed to make the linker tolerate resolution failures. Do you know why this case is different? Does Resolve actually throw an exception?

mrvoorhe:
This code wasn't covered by that option.  I thought about adding an if to support that option here, but this seemed like a slightly different case and I'm not sure if there would ever be a point in wanting a throw?  I don't care either way.  I can leave that change as is, or add in a `if (Context.Resolver.IgnoreUnresolved) throw;` inside the catch block.

sbomer:
If `IgnoreUnresolved` is false, do we hit the same resolution failure in MarkStep, or does it only ever show up in SweepStep? If it shows up in MarkStep, then it seems like a resolution attempt that should be covered by `IgnoreUnresolved`, and I'd prefer to make the behavior consistent with MarkStep by throwing a `ResolutionException` if `IgnoreUnresolved` is false. If it doesn't show up in MarkStep, I'm curious how we get past MarkStep and only hit it in SweepStep. If you prefer to leave it as-is I'm fine with that too since it's a different case like you said - I'm mostly interested in understanding how we get here.

marek-safar:
This logic is correct, the exception is equal to unmarked type in this case so it's fine to skip/ignore.

PR #789, opened 2019-10-14
Don't try to change scope of windows runtime projection

mrvoorhe:
Trying to change the scope of a projection results in a exception`System.InvalidOperationException : Projected type scope can't be changed`, so let's skip these.

### DIFF
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -414,6 +414,10 @@ namespace Mono.Linker.Steps {
 
 		static void UpdateTypeScope (TypeReference type, AssemblyDefinition assembly)
 		{
+			// Can't update the scope of windows runtime projections
+			if (type.IsWindowsRuntimeProjection)
+				return;
+
 			if (type is GenericInstanceType git && git.HasGenericArguments) {
 				UpdateTypeScope (git.ElementType, assembly);
 				foreach (var ga in git.GenericArguments)


src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -414,6 +414,10 @@ static void UpdateForwardedTypesScope (CustomAttributeArgument attributeArgument
 
 		static void UpdateTypeScope (TypeReference type, AssemblyDefinition assembly)
 		{
+			// Can't update the scope of windows runtime projections
+			if (type.IsWindowsRuntimeProjection)
```
marek-safar: This is not enough as it will produce invalid assembly too. We need to either fix the limitation on Cecil side or add this logic to mark step to keep the scope assembly

marek-safar: Hmm, this is more tricky than I thought. We do already mark the dependency correctly but I think linker can currently mark only the end-type, not the forwarding type. Therefore when you link also assembly with the forwarder you end up with broken output. Another option is to fail with an error when this combination appears.

Do you have a sample I could try few combinations with?

mrvoorhe: I don't have a great sample.  These happen during our windows runtime tests.  It looks like it's trying to change the scope on CustomAttribute constructor argument types in `windows.winmd`

```
.class public auto ansi sealed windowsruntime Windows.Foundation.Metadata.WebHostHiddenAttribute
	extends [System.Runtime]System.Attribute
{
	.custom instance void [System.Runtime]System.AttributeUsageAttribute::.ctor(valuetype [System.Runtime]System.AttributeTargets) = (
		01 00 1c 14 00 00 01 00 54 02 0d 41 6c 6c 6f 77
		4d 75 6c 74 69 70 6c 65 00
	)
	.custom instance void [Windows.winmd]Windows.Foundation.Metadata.ContractVersionAttribute::.ctor(class [mscorlib]System.Type, uint32) = (
		01 00 25 57 69 6e 64 6f 77 73 2e 46 6f 75 6e 64
		61 74 69 6f 6e 2e 46 6f 75 6e 64 61 74 69 6f 6e
		43 6f 6e 74 72 61 63 74 00 00 01 00 00 00
	)
	// Methods
	.method public hidebysig specialname rtspecialname 
		instance void .ctor () runtime managed internalcall 
	{
	} // end of method WebHostHiddenAttribute::.ctor

} // end of class Windows.Foundation.Metadata.WebHostHiddenAttribute
```

It's trying to change the Scope of `[System.Runtime]System.AttributeTargets` from `System.Runtime` to `mscorlib`

Full call stack is
```
TypeReference.set_Scope() in Mono.Cecil, Mono.Cecil.dll
SweepStep.UpdateTypeScope() in Mono.Linker.Steps, monolinker.exe
SweepStep.UpdateForwardedTypesScope() in Mono.Linker.Steps, monolinker.exe
SweepStep.UpdateForwardedTypesScope() in Mono.Linker.Steps, monolinker.exe
SweepStep.SweepCustomAttributes() in Mono.Linker.Steps, monolinker.exe
SweepStep.SweepCustomAttributes() in Mono.Linker.Steps, monolinker.exe
SweepStep.SweepType() in Mono.Linker.Steps, monolinker.exe
SweepStep.SweepAssembly() in Mono.Linker.Steps, monolinker.exe
UnitySweepStep.SweepAssembly() in Unity.Linker.Steps, UnityLinker.exe
SweepStep.ProcessAssemblyAction() in Mono.Linker.Steps, monolinker.exe
SweepStep.Process() in Mono.Linker.Steps, monolinker.exe
UnitySweepStep.Process() in Unity.Linker.Steps, UnityLinker.exe
BaseStep.Process() in Mono.Linker.Steps, monolinker.exe
Pipeline.ProcessStep() in Mono.Linker, monolinker.exe
UnityPipeline.ProcessStep() in Unity.Linker, UnityLinker.exe
Pipeline.Process() in Mono.Linker, monolinker.exe
UnityDriver.Run() in Unity.Linker, UnityLinker.exe
UnityDriver.RunDriverWithoutErrorHandling() in Unity.Linker, UnityLinker.exe
UnityLinker.Link() in Unity.Linker.Tests.TestCaseRunner, Unity.Linker.Tests.dll
```

mrvoorhe: I don't fully understand this situation, but my understanding is that there is no reason to change the scope on these types.  The Scope is just there because cecil has one.

This change makes our tests green.  Even if this change isn't a perfect fix, it's much better than the current behavior (a crash).  And I believe it is no worse than the behavior we had prior to   https://github.com/mono/linker/commit/c0b569a81e7e98113c5e52722ecd417e1b916f78

This is the last issue I need to fix in order for us to sync up with `master` again.  Can we land this and worry about the ideal behavior later?  Or am I missing some aspect of this change that makes it a bigger issue than I realize?

TautvydasZilys: > This is not enough as it will produce invalid assembly too

It will not. In Windows Runtime assemblies, Scope does not exist. Cecil takes them out of thin air because a lot of code expects it to exist and be a sensible value. Windows Runtime types are fully qualified by their name and namespace. Cecil doesn't allow you changing the scope because it just doesn't make sense to do so for Windows Runtime types.

marek-safar: ok, merging to unblock @mrvoorhe but I am still not sure about the approach

Issue #676, opened 2019-07-26
Add support for C++/CLI assemblies


sbomer:
Currently the linker skips C++/CLI assemblies (see https://github.com/mono/linker/pull/658), because
1. it's not aware of native code in mixed-mode assemblies
2. even for /clr:pure assemblies, cecil may reorder IL data that C++/CLI expects to be preserved (see https://github.com/mono/linker/issues/651#issuecomment-510689413).

As a result, dependencies of the C++/CLI assemblies aren't analyzed, requiring workarounds like https://github.com/dotnet/wpf/pull/1387.

This issue tracks improving the linker's support for C++/CLI assemblies. There are a number of routes we could take:
- making the linker aware of native code (probably expensive)
- preserving the ordering of IL data (which would work for /clr:pure assemblies, and probably involves a cecil fix)
- letting the linker analyze assemblies without modifying them (see https://github.com/mono/linker/issues/639)

MichalStrehovsky:
`/clr:pure` [is now deprecated](https://docs.microsoft.com/en-us/cpp/build/reference/clr-common-language-runtime-compilation?view=msvc-160) so there's little point in adding support for that.

The only way how illink would be able to process C++/CLI is if illink was an actual, real,  linker (accepting obj files as input).

The produced EXE/DLL doesn't have enough information for illink to safely touch anything in it (the *real* linker already computed various relative offsets and threw away the relocation information for them). The .NET metadata is in the middle of the executable (in the `.rdata` section). The `.text` section precedes it and `.data` section follows. If illink resizes the .NET metadata in any way, it's going to invalidate all relative offsets within the image that follow the .NET metadata blobs.

I don't see a path to success with this, unless we accept that the file cannot be touched and just append data to it (i.e. we would bloat the file, keeping a dead copy of the old .NET metadata, with the hope that it will allow us to make non-C++ assemblies smaller).

sbomer:
> I don't see a path to success with this, unless we accept that the file cannot be touched and just append data to it

Agreed, I think the current plan is to never touch them for the reasons you pointed out, but still analyze them for dependencies (and I don't think we should try to append a metadata copy either). @mateoatr

vitek-karas:
Linker now scans C++/CLI assemblies, but it will never modify them. C++/CLI assembly is either removed as a whole, or kept untouched.

PR #796, opened 2019-10-22
Mark dependencies of assemblies with the copy and save action

mrvoorhe:
When an assembly had the action Copy, a lot of it's dependencies could go unmarked.  Things like public methods were being marked, but private methods, attributes, events, properties were not.

With this change the MarkStep will visit everything in a copy or saved assembly and mark everything.

I did not address the *Used AssemblyActions.  These will still be vulnerable to invalid IL.  Additional logic would be needed to track the first marking of these assemblies and then call `ForceFullMarkingOfAssembly`.

I did not handle `AddBypassNGen` either.  The XML comment for this action states that `Keep all types, methods, and fields but add System.Runtime.BypassNGenAttribute to unmarked methods.`  I don't know how this action is used, if it will still strip out EventDefinitions and PropertyDefinitions then additional tweaks would be needed to avoid the calls to `ForceFullMarkingOfProperties` and `ForceFullMarkingOfEvents`.  If someone would like to handle these cases they could do so in a follow on PR.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -89,8 +89,15 @@ namespace Mono.Linker.Steps {
 			try {
 				MarkAssembly (assembly);
 
-				foreach (TypeDefinition type in assembly.MainModule.Types)
-					InitializeType (type);
+				var forceFullMark = NeedToFullyMarkAssembly (assembly);
+				if (forceFullMark) {
+					ForceFullMarkingOfAssembly (assembly);
+				} else {
+					foreach (TypeDefinition type in assembly.MainModule.Types)
+						InitializeType (type);
+				}
+
+
 			} finally {
 				Tracer.Pop ();
 			}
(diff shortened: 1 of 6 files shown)


sbomer:
@mrvoorhe I think there is value in keeping copying/rooting separate - you might want to root just the entry point and keep just what's needed for that, but keep one particular assembly intact (for example to preserve ordering for https://github.com/mono/linker/issues/676) without also pulling in all of *its* dependencies. There's the possibility that the intact assembly will have dangling assemblyrefs, but I don't think that's a problem for our use case (however invalid typerefs to a removed assemblyref are problematic and should be fixed).

Would it maybe be worth fixing the invalid IL without also pulling in these extra dependencies? It sounds like we're using the options in pretty different ways - curious what your take is on the above.

mrvoorhe:
> I think there is value in keeping copying/rooting separate

I  would have liked to have put this logic in it's own step instead of in the `MarkStep` solely for the sake of organization and control.  However, the MarkStep's methods for `MarkType`, `MarkCustomAttribute`, `MarkEvent`, `MarkProperty`, etc are very useful to have access to.  Replicating their behavior in a separate step prior to the mark step is challenging and error prone.

> It sounds like we're using the options in pretty different ways - curious what your take is on the above.

It does sound like maybe we are.  Let me explain how we use it. 

We collect all the assemblies involved in a game and then we path them through the linker.  Depending on a setting(Low/Medium/High), we assign different AssemblyActions to various assemblies.  On `Low` (which is our default behavior) we give the User's assemblies the `Copy` action and we process them with the `ResolveFromAssemblyStep`.  The idea is, User's may be using who knows what sort of reflection and we'd rather not run the risk of their code not working, so just pass their entire assembly through the linker.  Other assemblies such as UnityEngine assemblies and mono class libraries will always get the action `Link`.  The idea being that we will test and ensure Unity and class library code that is undetectable is correctly annotated so that dependencies are marked correctly.

Going into the `MarkStep`, an example of what our assembly action table looks like could be
```
Copy => UserAssembly1
Copy => UserAssembly2
Copy => UserAssembly3
Link => Unity1
Link => Unity2
Link => mscorlib
```

Our expectation is that the assemblies that come out of the linker are valid.  No missing types, no unresolved methods, no unresolved attributes, etc.

Now to the problem.  The problem I ran into was calling `ResolveFromAssemblyStep` on an assembly and assigning it the action Copy would only get it's public methods assigned the MethodAction `ForceParse`.  In a lot of cases that can give the appearance that all dependencies of an assembly with the Copy action survive.  However, there are a number of cases where that is not enough.

Maybe my understanding of what the assembly Copy action means is wrong?  I think I understand your use case.  If a runtime that lazily resolves everything is used then it's OK if the dependencies of a private method that is never called are unresolvable.

If that's how you want the Copy action to work, that's fine.  I think the way we use it is reasonable as well.  Maybe we can use a flag or a new action or provide this behavior?  Unfortunately, I don't think this would change the approach I used.  I tried to put this logic in it's own step and it was tedious and error prone to ensure all dependencies were marked without the help of the methods in the `MarkStep`.

Let me know your thoughts.  I'm happy to tweak under what conditions this logic will run.

sbomer:
@mrvoorhe thanks for the detailed explanation! I think I have a better understanding of your use case now.

By keeping them separate, I meant more of a logical separation - that copying shouldn't necessarily root everything in an assembly - whether the implementation is in the same or a different step. I've been thinking of `ResolveFromAssemblyStep` or `-a` as the option that would cause an entire assembly's contents+dependencies to be kept (for libraries at least), though that may not be entirely true today as you point out.

I'm curious which cases you're hitting where what `ResolveFromAssemblyStep` marks isn't enough. The one I'm aware of is that `copy` assemblies can get assembly references removed (say if they're type forwarders), while still containing type references to the removed assembly. In this case I've seen invalid IL where the typeref points to the incorrect assembly. This case could be fixed by keeping all assembly references in `copy` assemblies, even if the assemblies they point to are removed - or do you expect that all assemblies referenced will actually be output? Did you find other issues with invalid IL?

> Maybe my understanding of what the assembly Copy action means is wrong?

I think we're both trying to understand the best way to interpret it :)

Based on my understanding, I would expect `copy` to work how I described. If it's necessary to mark all dependencies like you are doing, then I see why it would go in `MarkStep`. A new action sounds reasonable, though I'm hoping we can find a definition of the `copy` action that works for both of us so we don't have to introduce yet another one. I believe that the change as-is would ultimately be OK for .NET Core because we use `CopyUsed` by default, not `Copy`.

If there are other issues with invalid IL, that might help me understand why we'd want `Copy` to behave as you've described.

Let me know what you think!

mrvoorhe:
> I'm curious which cases you're hitting where what ResolveFromAssemblyStep marks isn't enough.

The 3 new tests in this PR all fail if you were to comment out the call to `ForceFullMarkingOfAssembly`.

CustomAttributes are the poorest handled.  All of the attribute types that appear in the copy assembly go unmarked.

> The one I'm aware of is that copy assemblies can get assembly references removed (say if they're type forwarders), while still containing type references to the removed assembly

I was tinkering around with another problematic case that sounds similar.  I had a 3rd unused assembly that I was trying to get dropped as an unused referencing which would cause the `copy` action to be changed to `save` by the SweepStep.  I kind of had something, but it was more complicated than the tests you see in the PR and the simpler tests are enough to reproduce the problem.

The original problem I hit was actually much worse.  During one of our IL2CPP test suites I was hitting a crash in cecil while writing the assembly.  Cecil has a `enum_type.CheckResolve()` call when writing CustomAttribute data and that was throwing because the enum type was unresolvable.  Our setup for these tests was more complicated.  It involved an assembly with the `copy` action that had multiple methods with an attribute defined in another assembly with the `link` action and the attribute had a constructor argument that had an enum defined in the same linked assembly.  There was also a 3rd unused assembly that was linked away.  This led to the assembly with the copy action being switched over the `save` by the SweepStep and that in turn led to the crash during `AssemblyDefinition.Write` in the OutputStep. I was never able to reproduce this exact issue in a self contained test in the upstream linker suite.  That said, once I realized what was going on, I stopped trying write a test that had the same complex setup.

Let's wait and hear what @marek-safar has to say.  I don't have a strong opinion on which direction we go in.

marek-safar:
I think the Copy action should copy the assembly intact, basically skipping SweepStep (ignoring the complication with type forwarders) and keeping any required dependencies. So in a situation where assembly A (copy) references assembly B (link) we need to do member reference marking.

The use case Sven is talking about is different as he would like to mark API shape of the assembly which actually means linking will run just not touching visible edges. I think what needs more thinking is if we can solve this for copyused to work as expected (not to overmark). 

What is interesting on this is that this is the second request which would benefit from MarkStep to be more customizable. We should think if there is a better way of splitting this up going forward.

mrvoorhe:
> So in a situation where assembly A (copy) references assembly B (link) we need to do member reference marking.

@marek-safar does that mean you are OK with these changes?

> What is interesting on this is that this is the second request which would benefit from MarkStep to be more customizable. We should think if there is a better way of splitting this up going forward.

I thought about adding an interface to expose some of the members in MarkStep that are needed.  Using this change as an example, you could then put this new code into some separate helper class.  It's not a great approach but it might help pull out a few chunks.

sbomer:
> I think the Copy action should copy the assembly intact, basically skipping SweepStep (ignoring the complication with type forwarders)

This is what I think as well, and it sounds like it would solve all of the cases arising from `copy` being changed to `save`.

> and keeping any required dependencies

To me this seems orthogonal to the issue, and I believe it would be useful to not keep all dependencies. Usually dangling references to another assembly, for example, result in pretty readable exception messages that indicate exactly what was missing.

> The use case Sven is talking about is different as he would like to mark API shape of the assembly which actually means linking will run just not touching visible edges.

I do care about that use case as well, but here I'm actually talking about the general use case (including linking an app).

> I think what needs more thinking is if we can solve this for copyused to work as expected (not to overmark).

I think we need both:
- An action that guarantees a file is copied without modification. This is useful when there are invariants not preserved by the linker, for example, as a kind of override to force it to keep things intact. On its own this action wouldn't cause any additional dependencies to be included for analysis; this could result in some dangling references that the linker predicts won't be hit at runtime, but could be hit if the analysis missed something. `Copy` action seems like the right one for this since it's already pretty close.
- A way to keep the dependencies of "everything" in an assembly, where the definition of "everything" might vary slightly (for example it might be just public surface area, and may or may not include attributes, etc.) This is where there's overlap between @mrvoorhe's change and the library-mode we've talked about. I've been thinking of `copy` as a decision made post-analysis, since we already have a different way to keep (mostly) everything in an assembly - `-r` and `-a` were designed to do this. @mrvoorhe I'm wondering if the behavior you desire for attributes could be achieved by passing `-a` instead of using the `copy` action?

mrvoorhe:
> I realize that ResolveFromAssemblyStep doesn't do what you desire currently - my question is just whether you think it's worth making the new logic kick in for -r rather than for copy, or similar.

@sbomer I would prefer the way it works in this PR.  This is the most full proof way to get the outcome that we desire (every dependency of output assemblies exists).

Here's another scenario for you to consider.  You want to only link the class library assemblies.  A super easy way to do that is to set the UserAction to `copy`.  That's it.  No need to also pass -r for every assembly and all of it's non class library references.

PR #810, opened 2019-11-01
Run full mark logic on copy and copy-used assemblies

marek-safar:
Improved #796 which used more of the existing code

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -89,8 +89,15 @@ namespace Mono.Linker.Steps {
 			try {
 				MarkAssembly (assembly);
 
-				foreach (TypeDefinition type in assembly.MainModule.Types)
-					InitializeType (type);
+				var forceFullMark = NeedToFullyMarkAssembly (assembly);
+				if (forceFullMark) {
+					ForceFullMarkingOfAssembly (assembly);
+				} else {
+					foreach (TypeDefinition type in assembly.MainModule.Types)
+						InitializeType (type);
+				}
+
+
 			} finally {
 				Tracer.Pop ();
 			}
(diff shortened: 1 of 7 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -354,6 +431,8 @@ void MarkCustomAttributes (ICustomAttributeProvider provider)
 			if (!provider.HasCustomAttributes)
 				return;
 
+			bool markOnUse = _context.KeepUsedAttributeTypesOnly && Annotations.GetAction (GetAssemblyFromCustomAttributeProvider (provider)) == AssemblyAction.Link;
```
mrvoorhe: I thought about handling the marking of CustomAttributes this way, i.e. down in   `MarkCustomAttributes` and by checking the assembly action.  I decided against this route for a few reasons.

1) I felt it was much more opaque.  If there were ever a bug again in marking everything in an assembly I'd easily stumble on `MarkEntireType` (or in my PR I had a method with the same purpose).  I'd then see that that method didn't do anything with custom attributes and I thought that could easily lead to a red herring situation.

2) Accessing the Assembly of every `ICustomAttributeProvider` object we call `MarkCustomAttributes` and then looking up that assembly in a dictionary seems much more costly than just looping back in `MarkEntireType` and calling `MarkCustomAttribute`.

3) From a tracing perspective, I think it is more accurate to mark the custom attributes back in `MarkEntireType`.  The single biggest reason the custom attribute is being marked is because the assembly has the copy action.  I'm not saying the tracing is currently setup correctly to draw a clear connection back to the Assemblies Copy action, but if we ever wanted to ensure this scenario was traced in a sensible way, I think it would be easier to wire that up if we marked the CustomAttributes directly from` MarkEntireType`

I'm not necessarily opposed to how you did it.  But I did think it was worth mentioning why I explicitly did not pick this approach.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -385,6 +461,28 @@ void MarkCustomAttributes (ICustomAttributeProvider provider)
 			}
 		}
 
+		static AssemblyDefinition GetAssemblyFromCustomAttributeProvider (ICustomAttributeProvider provider)
```
mrvoorhe: In my PR it came up that we'd all like to see less code living in the MarkStep.  One easy thing to do is to put simple static  utility type methods somewhere else.  I think this method could easily live else where.  An extension method would be nice and discoverable and would let you reduce the name down to `GetAssembly` 

i.e.
`static AssemblyDefinition GetAssembly (this ICustomAttributeProvider provider)`

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -225,6 +287,21 @@ bool QueueIsEmpty ()
 			return _methods.Count == 0;
 		}
 
+		bool ProcessCopyUsedAssemblies ()
```
mrvoorhe: I do not understand what purpose this serves.  It doesn't look like enough to handle the copyused scenario.  Can you add tests to prove that copyused works as expected?  Without tests for copyused I don't see the value in adding code to handle it.  The behavior is obscure, there are zero tests today, and this bug has gone unnoticed for years.  Any code added to handle copyused could easily break and we wouldn't know for years again.

src/linker/Linker.Steps/MarkStep.cs
marek-safar: I didn't want to copy the CA marking logic. We are heading down the path where linker has deeper knowledge about framework attributes and duplicating that for copy and link actions is something I want to avoid.

src/linker/Linker.Steps/MarkStep.cs
marek-safar: I agree but that should be done in separate PR, there are many more methods we could move from MarkStep

src/linker/Linker.Steps/MarkStep.cs
marek-safar: Fair point, will add test for this path

PR #813, opened 2019-11-06
Add callback for handling marking reflection calls

brian-taylor-unity:
Added a callback that can be overridden so we can suppress warnings on instructions that have already been handled.

Also allows for checking reflection calls like `GetMethod()` and `GetConstructor()` that include parameter arrays, so we can mark only methods with matching parameters.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2303,9 +2303,9 @@ namespace Mono.Linker.Steps {
 		}
 
 		//
-		// Callback for other mark steps to prevent normal reflection logic from running
+		// Extension point for reflection logic handling customization
 		//
-		protected virtual bool ShouldProcessReflectionDependencies (MethodBody body, Instruction instruction)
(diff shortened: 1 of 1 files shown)


mrvoorhe:
@marek-safar To add a bit more context, this is meant as a way for our us prevent the reflection logic from running for a given reflection call if we so choose.  There are 2 cases where we want to do this.

1) Our better stack evaluation abilities let's use detect usages that the mark step logic cannot.  This leads to log messages from the `MarkStep` reflection code saying that a reflection could not be analyzed even though our `UnityMarkStep` was able to successfully handle the reflection usage.

2) @brian-taylor-unity is implementing signature type matching in our reflection marking logic so that our support for `GetConstructor` can do something more sensible than marking all ctor's with matching bindings flags.  As part of that he's reusing the signature matching for `GetMethod` and found that the reflection marking logic in the `MarkStep` ends up marking all overloads even though we were able to determine exactly which method was needed.  So we need some way of preventing the mark step reflection logic from running in this case as well.

PR #817, opened 2019-11-14
Add a --output-assemblylist <filename> option to dump the final set of linked assemblies in JSON format.

vargaz:


src/linker/Linker.Steps/OutputStep.cs
```diff
@@ -38,6 +39,10 @@ namespace Mono.Linker.Steps {
 
 	public class OutputStep : BaseStep {
 
+		public static string AssemblyListFile {
```
mrvoorhe: I would prefer a new property on LinkConext that is set by the Driver and then accessed here. That is the more established pattern

PR #827, opened 2019-11-22
Add C# specific settings to .editorconfig.

vitek-karas:
When using VS these new settings make VS automatically format the code the way the rest of the linker is written.

Note that linker itself is not always 100% consistent, but it's close.


vitek-karas:
@sbomer

PR #828, opened 2019-11-22
Ignore reflection calls inside reflection implementation

vitek-karas:
Some relfection methods are implemented by calling different overloads of the same method. In these cases linker should not try to analyze the internal call.

For example:
```C#
MethodInfo GetMethod(string name) => GetMethod(name, null);
```
In this case the call to `GetMethod(name, null)` should not be analyzed and should not cause "warnings".

Added new test infra to be able to capture and validate logged messages.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2341,6 +2341,7 @@ namespace Mono.Linker.Steps {
 			if (HasManuallyTrackedDependency (body))
 				return;
 
+			var methodCalling = body.Method;
 			var instructions = body.Instructions;
 
 			//
@@ -2373,6 +2374,13 @@ namespace Mono.Linker.Steps {
 				// System.Type
 				//
 				case "Type" when methodCalledType.Namespace == "System":
+
+					// Some of the overloads are implemented by calling another overload of the same name.
+					// These "internal" calls are not interesting to analyze, the outermost call is the one
+					// which needs to be analyzed. The assumption is that all overloads have the same semantics.
+					// (for example that all overload of GetConstructor if used require the specified type to have a .ctor).
+					if (methodCalling.Name == methodCalled.Name && methodCalling.DeclaringType.Name == "Type" && methodCalling.DeclaringType.Namespace == "System")
+						break;
 
 					switch (methodCalled.Name) {
 					//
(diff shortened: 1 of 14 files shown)


vitek-karas:
@sbomer

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2374,6 +2375,13 @@ protected virtual void MarkReflectionLikeDependencies (MethodBody body)
 				//
 				case "Type" when methodCalledType.Namespace == "System":
 
+					// Some of the overloads are implemented by calling another overload of the same name.
+					// These "internal" calls are not interesting to analyze, the outermost call is the one
+					// which needs to be analyzed. The assumption is that all overloads have the same semantics.
+					// (for example that all overload of GetConstructor if used require the specified type to have a .ctor).
+					if (methodCalling.Name == methodCalled.Name && methodCalling.DeclaringType.Name == "Type" && methodCalling.DeclaringType.Namespace == "System")
```
marek-safar: What about ?
`if (methodCalledDefinition.DeclaringType == methodCalling.DeclaringType && methodCalling.Name == methodCalled.Name)` instead

test/Mono.Linker.Tests.Cases/Reflection/CoreLibMessages.cs
```diff
@@ -0,0 +1,23 @@
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.Reflection
+{
+	[SetupLinkerAction ("copy", "CoreLibEmulator")]
+	[SetupCompileBefore ("CoreLibEmulator.dll", new string [] { "Dependencies/CoreLibEmulator.cs" }, null, new string [] { "INCLUDE_CORELIB_IMPL" })]
+
+	// Validate that calls from one overload of Type.Get* to another overload of the same method don't produce warning
+	[LoggedMessageDidNotMatch ("Reflection call 'System.Reflection.ConstructorInfo System.Type::GetConstructor\\([^)]*\\)' inside 'System.Reflection.ConstructorInfo System.Type::GetConstructor\\([^)]*\\)' does not use detectable instance type extraction")]
```
marek-safar: The name is a bit weird. I'd go with something like `LogDoesNotContain`

vitek-karas: Does it matter that it does regex match instead of a simple "substring"?
But I'm perfectly fine with `LogContains` and `LogDoesNotContain` names...

marek-safar: It does not

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Thanks, fixed.

test/Mono.Linker.Tests.Cases/Reflection/CoreLibMessages.cs
vitek-karas: Thanks, renamed.

PR #848, opened 2019-12-06
Add new optimization steps to make Mark step more effective

marek-safar:
Two new steps have been introduced

### BodySubstituterStep

This step removes any conditional blocks when the condition or conditions
are evaluated as constants. This step does not do any inlining. The conditional
logic is kept but based on the known values only one branch is always taken.

A simple example which can be detected by linker.

```c#
class C
{
	static void Main ()
	{
		if (Result)
			Console.WriteLine (); // <- this call will be marked and removed
	}

	static bool Result () => false;
}

```

### RemoveUnreachableBlocksStep

A new command-line option called `--substitutions` allow external customization of any
methoda for assemblies which are linked. The syntax is same as for existing linker descriptor
XML files but it add way to control body modifications.

An example of XML file

```xml
<linker>
    <type fullname="Mono.Linker.Tests.Cases.Substitutions.StubBodyWithValue">
      <method signature="System.String TestMethod_1()" body="stub" value="abcd">
      </method>
    </type>
  </assembly>
</linker>

```

The `value` attribute is optional and only required when the method stub should not
fallback to default behaviour.

Additional to `stub` value also `remove` value is supported to forcibly remove body
of the method when the method is marked. This is useful when the conditional logic
cannot be evaluated by linker and the method will be marked but never actually reached.

### Applicability

Both steps can be combined to achieve the effect of externally customizing which
conditional branches can be removed during the linking.

I can illustrate that on IntPtr.Size property. With following substitution any code
that has compiled in conditional logic for 64 bits handling by checking IntPtr.Size will
be removed during linking.

```xml
<linker>
  <assembly fullname="mscorlib">
    <type fullname="System.IntPtr">
      <method signature="System.Int32 get_Size()" body="stub" value="4">
      </method>
    </type>
  </assembly>
</linker>

```

### DIFF
--- a/src/linker/Linker.Steps/BodySubstituterStep.cs
+++ b/src/linker/Linker.Steps/BodySubstituterStep.cs
@@ -1,0 +1,258 @@
+﻿using System;
+using System.IO;
+using System.Xml.XPath;
+using Mono.Cecil;
+using System.Globalization;
+
+namespace Mono.Linker.Steps
+{
+	public class BodySubstituterStep : BaseStep
+	{
+		protected override void Process ()
+		{
+			var files = Context.Substitutions;
+			if (files == null)
+				return;
+
+			foreach (var file in files) {
+				try {
+					ReadSubstitutionFile (GetSubstitutions (file));
+				} catch (Exception ex) when (!(ex is XmlResolutionException)) {
+					throw new XmlResolutionException ($"Failed to process XML substitution '{file}'", ex);
+				}
+
+			}
+		}
+
+		static XPathDocument GetSubstitutions (string substitutionsFile)
+		{
+			using (FileStream fs = File.OpenRead (substitutionsFile)) {
+				return GetSubstitutions (fs);
+			}
+		}
+
+		static XPathDocument GetSubstitutions (Stream substitutions)
+		{
+			using (StreamReader sr = new StreamReader (substitutions)) {
+				return new XPathDocument (sr);
+			}
+		}
+
+		void ReadSubstitutionFile (XPathDocument document)
+		{
+			XPathNavigator nav = document.CreateNavigator ();
+
+			// Initial structure check
+			if (!nav.MoveToChild ("linker", ""))
+				return;
+
+			// TODO: Add handling for feature
+
+			ProcessAssemblies (nav.SelectChildren ("assembly", ""));
+		}
+
+		void ProcessAssemblies (XPathNodeIterator iterator)
+		{
+			while (iterator.MoveNext ()) {
+				var name = GetAssemblyName (iterator.Current);
+
+				var cache = Context.Resolver.AssemblyCache;
+
+				if (!cache.TryGetValue (name.Name, out AssemblyDefinition assembly)) {
+					Context.LogMessage (MessageImportance.Low, $"Could not match assembly '{name.FullName}' for substitution");
+					continue;
+				}
+
+				ProcessAssembly (assembly, iterator);
+			}
+		}
+
+		void ProcessAssembly (AssemblyDefinition assembly, XPathNodeIterator iterator)
+		{
+			ProcessTypes (assembly, iterator.Current.SelectChildren ("type", ""));
+		}
+
+		void ProcessTypes (AssemblyDefinition assembly, XPathNodeIterator iterator)
+		{
+			while (iterator.MoveNext ()) {
+				XPathNavigator nav = iterator.Current;
+
+				string fullname = GetAttribute (nav, "fullname");
+
+				TypeDefinition type = assembly.MainModule.GetType (fullname);
+
+				if (type == null) {
+					Context.LogMessage (MessageImportance.Low, $"Could not resolve type '{fullname}' for substitution");
+					continue;
+				}
+
+				ProcessType (type, nav);
+			}
+		}
+
+		void ProcessType (TypeDefinition type, XPathNavigator nav)
+		{
+			if (!nav.HasChildren)
+				return;
+
+			XPathNodeIterator methods = nav.SelectChildren ("method", "");
+			if (methods.Count == 0)
+				return;
+
+			ProcessMethods (type, methods);
+		}
+
+		void ProcessMethods (TypeDefinition type, XPathNodeIterator iterator)
+		{
+			while (iterator.MoveNext ())
+				ProcessMethod (type, iterator);
+		}
+
+		void ProcessMethod (TypeDefinition type, XPathNodeIterator iterator)
+		{
+			string signature = GetAttribute (iterator.Current, "signature");
+			if (string.IsNullOrEmpty (signature))
+				return;
+
+			MethodDefinition method = FindMethod (type, signature);
+			if (method == null) {
+				Context.LogMessage (MessageImportance.Low, $"Could not find method '{signature}' for substitution");
+				return;
+			}
+
+			string action = GetAttribute (iterator.Current, "body");
+			switch (action) {
+			case "remove":
+				Annotations.SetAction (method, MethodAction.ConvertToThrow);
+				return;
+			case "stub":
+				string value = GetAttribute (iterator.Current, "value");
+				if (value != "") {
+					if (!TryConvertValue (value, method.ReturnType, out object res)) {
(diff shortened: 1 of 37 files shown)


src/linker/Linker.Steps/BodySubstituterStep.cs
```diff
@@ -0,0 +1,258 @@
...
+using Mono.Cecil;
+using System.Globalization;
+
+namespace Mono.Linker.Steps
+{
+	public class BodySubstituterStep : BaseStep
+	{
+		protected override void Process ()
+		{
+			var files = Context.Substitutions;
```
mrvoorhe: Why did you decide to go with a separate list of files instead of extending the ResolveFromXmlStep to be able to parse the `body` and `value` xml attributes?

mrvoorhe: I can see pros and cons to both approaches.  Just curious what your thinking was.

src/linker/Linker.Steps/BodySubstituterStep.cs
```diff
@@ -0,0 +1,258 @@
...
+			string action = GetAttribute (iterator.Current, "body");
+			switch (action) {
+			case "remove":
+				Annotations.SetAction (method, MethodAction.ConvertToThrow);
+				return;
+			case "stub":
+				string value = GetAttribute (iterator.Current, "value");
+				if (value != "") {
+					if (!TryConvertValue (value, method.ReturnType, out object res)) {
+						Context.LogMessage (MessageImportance.Normal, $"Invalid value for '{signature}' stub");
```
mrvoorhe: I'd argue for throwing instead of logging a message.  If we log a message & ignore the item then the user doesn't get the behavior they are expecting and then they have to waste time figuring out why whatever they intended to have stubbed wasn't stubbed.  Best case scenario they notice the log message and fix it, worst case they waste a lot more time fumbling around.  In either case you are wasting more of the users time than just throwing and telling them what mistake they made.

mrvoorhe: If you are not a fan of that, then my other recommendations would be

* Use `MessageImportance.High` instead.

* Put the LogMessge call into a protected virtual method so that we can choose to change the behavior if we want.

src/linker/Linker.Steps/BodySubstituterStep.cs
```diff
@@ -0,0 +1,258 @@
...
+						return;
+					}
+
+					Annotations.SetMethodStubValue (method, res);
+				}
+
+				Annotations.SetAction (method, MethodAction.ConvertToStub);
+				return;
+			default:
+				Context.LogMessage (MessageImportance.Normal, $"Unknown body modification '{action}' for '{signature}'");
```
mrvoorhe: Whatever is decided above, same here.

src/linker/Linker.Steps/CodeRewriterStep.cs
```diff
@@ -109,27 +110,90 @@ MethodBody CreateStubBody (MethodDefinition method)
...
+				il.Emit (OpCodes.Ldloc_0);
+				return;
+			}
 
-			var body = new MethodBody (method);
-			var il = body.GetILProcessor ();
-			il.Emit (OpCodes.Ldc_I4_0);
-			il.Emit (OpCodes.Ret);
-			return body;
+			throw new NotImplementedException (method.ReturnType.MetadataType.ToString ());
```
mrvoorhe: I think it would be nice to display the name of the method that was beings stubbed in the exception message.

test/Mono.Linker.Tests.Cases/Substitutions/RemoveBody.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Mono.Linker.Tests.Cases.Substitutions
+{
+	[SetupLinkerSubstitutionFile ("RemoveBody.xml")]
+	public class RemoveBody
+	{
+		public static void Main()
+		{
+			new RemoveBody ();
+            new NestedType (5);
```
mrvoorhe: Indent looks messed up in a couple places in this file.

test/Mono.Linker.Tests.Cases/UnreachableBlock/TryFinallyBlocks.cs
```diff
@@ -0,0 +1,34 @@
...
+			get {
+				try {
+					return 3;
+				} finally {
+
+				}
+			}
+		}        
+
+        [Kept]
```
mrvoorhe: Indent

test/Mono.Linker.Tests.Cases/UnreachableBlock/MultiStageRemoval.cs
```diff
@@ -0,0 +1,80 @@
...
+		{
+			TestMethod_1 ();
+			TestMethod_2 ();
+		}
+
+		[Kept]
+		[ExpectBodyModified]
+		static void TestMethod_1 ()
+		{
+			if (TestProperty_int () == 0)
```
mrvoorhe: Once we got into more complex body editing and leveraging `[ExpectBodyModified]` I noticed it wasn't that hard to have bugs that resulted in valid IL but the modified IL was wrong.  If you were really diligent with your test setup and calling methods and should be removed or kept you could prevent most bugs from slipping by.  However, for added safety we added a series of attributes to the test framework to invoke the linked code.

We a few attributes that will load the input and output assemblies in an AppDomain and invoke methods and check their behavior.
`[CheckInvokeMatchesOriginal]`
`[CheckInvoke(<expected result>)]`
`[CheckInvokeThrows]`

And we have one other attribute that you can put on the test case type itself that will tell the test framework to run the input and output `test.exe` and compare their output.
`[CheckOutput]`

If you would like me to bring these upstream let me know.

src/linker/Linker.Steps/BodySubstituterStep.cs
marek-safar: Fixed by using MessageImportance.High

src/linker/Linker.Steps/BodySubstituterStep.cs
marek-safar: Ditto as above

src/linker/Linker.Steps/CodeRewriterStep.cs
marek-safar: Changed

test/Mono.Linker.Tests.Cases/Substitutions/RemoveBody.cs
marek-safar: Fixed

test/Mono.Linker.Tests.Cases/UnreachableBlock/TryFinallyBlocks.cs
marek-safar: Fixed

src/linker/Linker.Steps/BodySubstituterStep.cs
marek-safar: There is actually not that much in common and adding if everywhere would not make it readable

test/Mono.Linker.Tests.Cases/UnreachableBlock/MultiStageRemoval.cs
marek-safar: I think peverify is able to catch such issue so no for now

src/linker/Linker.Steps/BodySubstituterStep.cs
```diff
@@ -0,0 +1,258 @@
...
+		{
+			using (FileStream fs = File.OpenRead (substitutionsFile)) {
+				return GetSubstitutions (fs);
+			}
+		}
+
+		static XPathDocument GetSubstitutions (Stream substitutions)
+		{
+			using (StreamReader sr = new StreamReader (substitutions)) {
+				return new XPathDocument (new StringReader (sr.ReadToEnd ()));
```
Therzok: Avoid loading the whole stream into a string, then making another reader on top. This should help with performance of the operation and is simpler

```suggestion
				return new XPathDocument (sr);
```

marek-safar: Thanks

PR #856, opened 2019-12-12
Disable RemoveUnreachableBlocksStep if --substitutions is not used.

vargaz:


vargaz:
This pass currently causes some regressions in wasm tests, disabling it unless --substitutions is used avoids these until the new pass is bug free.

marek-safar:
That's not what we want to achieve, these 2 passes are independent. Please open issue about the failing assembly

PR #872, opened 2019-12-18
Add abstraction between the linker and the dependency tracing/recording.

vitek-karas:
Adds ;IDependencyRecorder; which separates the linker/tracer from the actual implementation which stores the dependencies.
Implements the default XML writer for the dependencies (should produce exactly the same output as before the change).
Adds test infra to test the dependency recorder directly (as a test mock) and a simple test.

Motivation: Future analyzers need to have visibility into the dependency analysis (MarkStep basically) to report what caused a given method/type to be included. We already expose this information through the dependency XML, this just adds an API to do this programatically.

### DIFF
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -29,7 +29,6 @@
 using System;
 using System.IO;
 using System.Collections.Generic;
-using System.Linq;
 using System.Text;
 using System.Xml.XPath;
 
@@ -443,6 +442,8 @@ namespace Mono.Linker {
 				if (context.IsOptimizationEnabled (CodeOptimizations.ClearInitLocals))
 					p.AddStepBefore (typeof (OutputStep), new ClearInitLocalsStep ());
 
+				CustomizeContext (context);
+
 				PreProcessPipeline (p);
 
 				try {
@@ -456,6 +457,10 @@ namespace Mono.Linker {
 		}
 
 		partial void PreProcessPipeline (Pipeline pipeline);
+
+		protected virtual void CustomizeContext(LinkContext context)
+		{
+		}
 
 		protected static void AddCustomStep (Pipeline pipeline, string arg)
 		{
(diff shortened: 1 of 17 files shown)


vitek-karas:
@sbomer @radekdoulik

test/Mono.Linker.Tests/TestCasesRunner/LinkerCustomizations.cs
```diff
@@ -0,0 +1,22 @@
...
+	/// </summary>
+	public class LinkerCustomizations
+	{
+		public TestDependencyRecorder DependencyRecorder { get; set; }
+
+		public event Action<LinkContext> CustomizeContext;
+
+		public void CustomizeLinkContext(LinkContext context)
+		{
+			CustomizeContext (context);
```
radekdoulik: I guess this should check for `CustomizeContext` being non-null?

src/linker/Linker/Driver.cs
```diff
@@ -457,6 +458,10 @@ public void Run (ILogger customLogger = null)
 
 		partial void PreProcessPipeline (Pipeline pipeline);
 
+		protected virtual void CustomizeContext(LinkContext context)
```
marek-safar: What about overriding GetDefaultContext ?

src/linker/Linker/Tracer.cs
```diff
@@ -28,72 +28,62 @@
...
 
+		public void Start ()
+		{
 			if (string.IsNullOrEmpty (Path.GetDirectoryName (DependenciesFileName)) && !string.IsNullOrEmpty (context.OutputDirectory)) {
 				DependenciesFileName = Path.Combine (context.OutputDirectory, DependenciesFileName);
 				Directory.CreateDirectory (context.OutputDirectory);
 			}
 
-			var depsFile = File.OpenWrite (DependenciesFileName);
+			xmlDependencyRecorder = new XmlDependencyRecorder (DependenciesFileName, context);
```
marek-safar: I think xmlDependencyRecorder should not be part of this type. It should be set externally

src/linker/Linker/Driver.cs
marek-safar: What about overwriting GetDefaultContext instead?

PR #873, opened 2019-12-19
Add requirement to use LF end-of-line into editor config

vitek-karas:


.editorconfig
```diff
@@ -1,5 +1,6 @@
 # top-most EditorConfig file
 root = true
+end_of_line = "lf"
```
marek-safar: Why? I think we want native line endings between different operating systems not force LF

vitek-karas: I must admit I don't understand GIT that well - but all my clones on Windows end up having LF everywhere. And then editing such files in VS leads to mixed line endings. If somebody knows how to tell GIT to correctly handle this - and ideally do so such that simple "git clone" works correctly... I'll be happy to use that fix instead.

vitek-karas: I think I found it:
https://github.com/mono/linker/blob/08ef96864a817c5e6579bdf19a72a8287b3ee917/.gitattributes#L5

So either we change that, so that files checkout as CRLF on Windows, or if we want to keep that, then I think the editorconfig should specify the same to avoid mixed line endings in files.

marek-safar: Yep, we'd need to mess with .gitattributes. Thinking about it again I think having consistent eol is probably the least painful option

PR #907, opened 2020-01-08
Introduce IReflectionPatternRecorder to allow recording of pattern matches

vitek-karas:
The Mark step performs pattern matching for some reflection methods to be able to figure out dependencies even across reflection calls. Currently it only writes to log when it finds a pattern it can't recognize.

This change introduces a new interface IReflectionPatternRecorder which is called every time the pattern matching either successfuly recognizes a pattern or fails to do so. By default this is implemented to log on failure (just like before the change).

This allows additional tools to record the failures and or the successes in a different way.

Added support to test this functionality and modified a couple of tests to validate the basics.

The biggest change is in the MarkStep where the pattern matching has been refactored a little bit to fulfill a promise of always reporting success/failure for a given callsite. This is important for some consumers to be able to handle multiple call sites to the same reflection method in calling method body (without this we would have to somehow report the actual call site location which complicates things quite a bit on the consumption side).

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2353,8 +2353,8 @@ namespace Mono.Linker.Steps {
 			if (HasManuallyTrackedDependency (body))
 				return;
 
-			var methodCalling = body.Method;
 			var instructions = body.Instructions;
+			ReflectionPatternDetector detector = new ReflectionPatternDetector (this, body.Method);
 
 			//
 			// Starting at 1 because all patterns require at least 1 instruction backward lookup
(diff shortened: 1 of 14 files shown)


vitek-karas:
@sbomer

test/Mono.Linker.Tests.Cases.Expectations/Assertions/RecognizedReflectionAccessPatternAttribute.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Assertions
+{
+	[AttributeUsage (AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
+	public class RecognizedReflectionAccessPatternAttribute : BaseExpectedLinkedBehaviorAttribute
+	{
+		public RecognizedReflectionAccessPatternAttribute (string sourceMethod, string reflectionMethod, string accessedItem)
```
mrvoorhe: I think it would be nice to add the overloads that make maintaining and writing tests easier.  When these attributes use strings only it leads to tests that fall apart at the slightest refactoring.  And it deters us from writing tests with nested types and generics as those tend to complicate the string you need to craft.

For example, see https://github.com/mono/linker/blob/master/test/Mono.Linker.Tests.Cases.Expectations/Assertions/RemovedInterfaceOnTypeInAssemblyAttribute.cs

Supporting an overload where as much type information is supplied via `Type` rather than `string` does add additional work for supporting the new attribute, but it makes for a better and more maintainable test writing experience.

Using your assertion in the `MethodUsedViaReflection` test, it could become

```
	[RecognizedReflectionAccessPatternAttribute(typeof(MethodUsedViaReflection), nameof(MethodUsedViaReflection.Main), "()",
		typeof(Type), nameof(Type.GetMethod), "System.String,System.Reflection.BindingFlags",
		typeof(MethodUsedViaReflection), nameof(MethodUsedViaReflection.OnlyCalledViaReflection), "()")]
```

Or even
```
	[RecognizedReflectionAccessPatternAttribute(typeof(MethodUsedViaReflection), nameof(MethodUsedViaReflection.Main),new Type[0],
		typeof(Type), nameof(Type.GetMethod), new [] {typeof(BindingFlags)},
		typeof(MethodUsedViaReflection), nameof(MethodUsedViaReflection.OnlyCalledViaReflection), new Type[0])]
```

The easiest overload would be
```
		public RecognizedReflectionAccessPatternAttribute (Type sourceMethodType, string sourceMethodName, string sourceMethodSignature,
			Type reflectionMethodType, string reflectionMethodName, string reflectionMethodSignature,
			Type accessedItemType, string accessItemMemberName, string accessedItemSignature = null)
		{
		}
```
There should be enough logic in `ResultChecker` to make supporting this easy since other attributes do similar things.

Going 1 step further we could let you define the parameters by type as well.  No other attributes allow this today (I don't know why, maybe I just never thought of it).  I'd say going this far is optional but it would be cool to have.  I was thinking null could mean no signature (i.e. look for a Field or property to assert). 

```		
		public RecognizedReflectionAccessPatternAttribute (Type sourceMethodType, string sourceMethodName, Type[] sourceMethodSignature,
			Type reflectionMethodType, string reflectionMethodName, Type[] reflectionMethodSignature,
			Type accessedItemType, string accessItemMemberName, Type[] accessedItemSignature = null)
		{
		}


```

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2708,108 +2742,118 @@ protected virtual void MarkReflectionLikeDependencies (MethodBody body)
...
-							MarkMethodsFromReflectionCall (declaringType, ".ctor", 0, bindingFlags, parametersCount);
-							continue;
-						}
+							break;
+						//
+						// static T CreateInstance<T> ()
+						//
+						case "CreateInstance`1":
+							// Not sure it's worth implementing as we cannot expant T and simple cases can be rewritten
+							_context.LogMessage (MessageImportance.Low, $"Activator call '{methodCalled.FullName}' inside '{body.Method.FullName}' is not supported");
```
mrvoorhe: Shouldn't this be reported as unrecognizable?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2372,16 +2372,35 @@ protected virtual void MarkReflectionLikeDependencies (MethodBody body)
...
+			int i,
+			MethodReference methodCalled)
+		{ 
+			var methodCalledDefinition = methodCalled.Resolve ();
+			if (methodCalledDefinition == null)
+				return;
+
+			var methodCalledType = methodCalled.DeclaringType;
+
+			// IMPORTANT:
```
mrvoorhe: I'm not sold on this whole defining methods inside methods feature.  This feels horribly abused here.  A single methods with hundreds of lines and a handful of 100 line methods defined inside it.  And setting flags such as `reportedPatternRecognition` defined 600 lines above in the outer methods. 

The way this is structured I have little faith it will succeed it always reporting Recognized or Unrecognized.  It looks trivial to forget to add a call to `ReportUnrecognizedReflectionCallPattern` or `ReportRecognizedReflectionAccessPattern`

Here is the pattern we use in our reflection marking which does a much better job of guaranteeing something or recorded on every usage.  It's not perfect, but it centralizes the responsibility of ensuring everything is recorded as success or failure.

```
            foreach (var reflectionCall in reflectionCalls)
            {
                using (var reflectionContext = new ReflectionUsageContext(Context, modificationContext.Inspector, reflectionCall, _recordedInstructions))
                {
                    var stackItem = modificationContext.Inspector.StackEvaluator.Evaluate(reflectionCall.Instruction);

                    reflectionContext.SetStackEvaluationResult(stackItem);

                    // If we were unable to evaluate the stack then we won't be able to assist with marking of this reflection usage.
                    if (stackItem == null)
                        continue;

                    reflectionCall.Handler(reflectionContext, stackItem);
                }
            }
```

And then a given handler may look like
```
        void ProcessCreateInstance(ReflectionUsageContext reflectionContext, DetailedStackItem stackItem)
        {
            var reflectionMethod = reflectionContext.ReflectionCall.ReflectionMethod;
            // Activator.CreateInstance<T>()
            if (!reflectionMethod.HasParameters && reflectionMethod is GenericInstanceMethod genericInstanceMethod && genericInstanceMethod.GenericArguments.Count == 1)
            {
                // It must be the no arguments overload
                var typeToMark = genericInstanceMethod.GenericArguments[0]?.Resolve();

                if (typeToMark != null)
                {
                    reflectionContext.TraceRecordAndMark(
                        typeToMark,
                        () =>
                        {
                            MarkType(typeToMark);
                            MarkDefaultConstructor(typeToMark);
                        });
                }
                return;
            }

            var loads = stackItem.Loads.Cast<DetailedStackItem>().ToArray();

            // Activator.CreateInstance(Type)
            if (reflectionMethod.Parameters.Count == 1)
            {
                var typeToMark = DetermineType(reflectionContext.Inspector, loads[0])?.Resolve();
                if (typeToMark != null)
                {
                    reflectionContext.TraceRecordAndMark(
                        typeToMark,
                        () =>
                        {
                            MarkType(typeToMark);
                            MarkDefaultConstructor(typeToMark);
                        });
                }
                return;
            }
        }
```

And our context class looks like
```

        class ReflectionUsageContext : IDisposable
        {
            private readonly UnityLinkContext _context;
            public readonly ReflectionCall ReflectionCall;
            public readonly IBodyInspector Inspector;
            private readonly HashSet<Instruction> _recordedInstructions;

            private bool _oneOrMoreMatchesFound;
            private DetailedStackItem _stackItem;

            public ReflectionUsageContext(UnityLinkContext context, IBodyInspector inspector, ReflectionCall reflectionCall, HashSet<Instruction> recordedInstructions)
            {
                Inspector = inspector;
                _context = context;
                ReflectionCall = reflectionCall;
                _recordedInstructions = recordedInstructions;
            }

            public void SetStackEvaluationResult(DetailedStackItem stackItem)
            {
                _stackItem = stackItem;
            }

            public void TraceRecordAndMark<T>(T memberToMark, Action mark)
                where T : IMetadataTokenProvider
            {
                using (_context.Tracer.PushScoped($"Reflection-{memberToMark}"))
                {
                    _oneOrMoreMatchesFound = true;
                    mark();
                    _recordedInstructions.Add(ReflectionCall.Instruction);
                    _context.ReflectionTracker.RecordDetectedUsage(Inspector.Body.Method, memberToMark, ReflectionCall.Instruction, _stackItem);
                }
            }

            public void Dispose()
            {
                // TODO by Mike : Record as undetectable
                if (!_oneOrMoreMatchesFound)
                {
                    _context.ReflectionTracker.RecordUndetectableUsage(Inspector.Body.Method, ReflectionCall.Instruction, _stackItem);
                }
            }
        }
``` 

The important part is that the handling of every reflection usage does not call any Mark methods directly.  Instead it passes an action to `reflectionContext.TraceRecordAndMark` so that the `ReflectionContext` can know at the end if anything was marked or not.  Then we can happily bail at any point and we still ensure the reflection usage is recorded.

I'm not saying this has to structured like ours but I think assuming every break & return will come with an Unrecognized call is tedious.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2925,6 +2969,135 @@ void ProcessSystemTypeGetMemberLikeCall (System.Reflection.MemberTypes memberTyp
 					break;
 				}
 			}
+
+			//
+			// arity == null for name match regardless of arity
+			//
+			void MarkMethodsFromReflectionCall (
```
marek-safar: Existing code style does not use C-like formatting for declaration, please follow the existing code style.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2372,16 +2372,35 @@ protected virtual void MarkReflectionLikeDependencies (MethodBody body)
...
 
-				var methodCalledDefinition = methodCalled.Resolve ();
-				if (methodCalledDefinition == null)
-					continue;
+				MarkReflectionLikeDependencyForMethodCall (body, methodCalling, instructions, i, methodCalled);
+			}
+		}
 
-				var methodCalledType = methodCalled.DeclaringType;
+		private void MarkReflectionLikeDependencyForMethodCall (
```
marek-safar: Existing code prefers not to have redundant modifiers. Also, keep the declaration at one line

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2925,6 +2969,135 @@ void ProcessSystemTypeGetMemberLikeCall (System.Reflection.MemberTypes memberTyp
...
+						MarkEvent (eventInfo);
+					} finally {
+						Tracer.Pop ();
+					}
+				}
+			}
+
+			void ReportUnrecognizedReflectionCallPattern (string message)
+			{
+				reportedPatternRecognition = true;
```
marek-safar: I don't like this approach. It might be a neat way to leverage compiler-generated wrapper class but I'd much prefer to move whole reflection detection logic to worked like struct and make it explicit there

test/Mono.Linker.Tests.Cases.Expectations/Assertions/RecognizedReflectionAccessPatternAttribute.cs
vitek-karas: Thanks for the suggestion. I took it one step further and moved the attributes to the method which contains the pattern. This way the first set of parameters is not needed anymore. Makes it much cleaner.

Makes detecting when to turn on the test recorder a bit trickier, but should work well. If we need to make it work across assemblies, then we would have to either tweak that logic or introduce an explicit "turn on" attribute. But I don't see that happening any time soon (if ever).

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Good catch.

Turns out the ``CreateInstance`1`` is not the correct name as reported by Cecil. Cecil reports generic methods without any modifications to their names - so in this case it's still `CreateInstance`.

My change fixes the problem and thus introduces a new unrecognized pattern in the linker. Which is technically a behavioral change. I think it's better to fix this though.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Should be fixed now

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Removed the modifiers and I think I fixed all the multi-line declarations

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: I adapted the approach suggested by @mrvoorhe. Could not make it a struct all the way as the suggested pattern of actions requires a ref-type so that it can be passed to the closure.
But this way it should only allocate one class per method body and the per-call info is a struct passed by reference, so doesn't allocate on heap.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Thanks for the suggestion. I adapted the code to match the pattern you describe above. There's still lot of potential for improvements (like moving the entire reflection pattern detection to a separate source file), but that's for another change (this one is already too big).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2376,555 +2376,723 @@ protected virtual void MarkReflectionLikeDependencies (MethodBody body)
...
-				//
-				case "Type" when methodCalledType.Namespace == "System":
-
-					// Some of the overloads are implemented by calling another overload of the same name.
-					// These "internal" calls are not interesting to analyze, the outermost call is the one
-					// which needs to be analyzed. The assumption is that all overloads have the same semantics.
-					// (for example that all overload of GetConstructor if used require the specified type to have a .ctor).
-					if (methodCalling.DeclaringType == methodCalling.DeclaringType && methodCalling.Name == methodCalled.Name)
-						break;
+				ReflectionPatternContext reflectionContext = new ReflectionPatternContext (_context, body.Method, methodCalledDefinition);
```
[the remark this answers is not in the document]
vitek-karas: Because for some reason `using` doesn't work on structs... compiler simply says that it won't allow it. I figured that since it's only in this one place, it's OKish to do the try/finally instead.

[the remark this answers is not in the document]
vitek-karas: Sorry - that was incorrect - using(Struct) works, but it makes the struct readonly. That means I can't pass it as a ref parameter (since the callee could mutate it). I need the ref parameter to avoid copying (since I'm using the struct as a way to allocate on stack, but I still need a single instance only), so can't use `using`.

https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1657

PR #908, opened 2020-01-09
Add support for fields substitutions.

vargaz:


src/linker/Linker.Steps/BodySubstituterStep.cs
```diff
@@ -157,7 +157,7 @@ void ProcessField (TypeDefinition type, XPathNodeIterator iterator)
 
 			var field = type.Fields.FirstOrDefault (f => f.Name == name);
 			if (field == null) {
-				Context.LogMessage (MessageImportance.High, $"Could not find field '{name}' for substitution.");
+				Context.LogMessage (MessageImportance.Low, $"Could not find field '{name}' for substitution.");
```
vargaz: Shouldn't all these cases be errors ?

marek-safar: I don't think we have to be pedantic if someone wants to use single xml for multiple purposes

vargaz: The problem is that these can silently stop working etc.

marek-safar: Yes but it's not a blocker. Let's go with .Normal

Issue #913, opened 2020-01-11
AssemblyAction.Skip in MarkStep causes a crash


tlakollo:
There is no described functionality for the case in which the AssemblyAction is Skip after fix #810 , before a Skip AssemblyAction will mark the assembly and InitializeType's. 
There are errors while building the Runtime repo caused by this change. Netstandard assembly does not contain any methods since is only a type forwarder, instead it relied in other assemblies which AssemblyAction is Skip to IntializeType and have methods In the _methods MarkStep queue. Later on in MarkStep.Process() while checking if the QueueIsEmpty since only netstandard methods are taken into account and netstandard does not have methods the program fails. 

```
Unhandled exception. System.InvalidOperationException: No entry methods
     at Mono.Linker.Steps.MarkStep.Process() in D:\validation\canary\linker\src\linker\Linker.Steps\MarkStep.cs:line 215
     at Mono.Linker.Steps.MarkStep.Process(LinkContext context) in D:\validation\canary\linker\src\linker\Linker.Steps\MarkStep.cs:line 76
     at Mono.Linker.Pipeline.ProcessStep(LinkContext context, IStep step) in D:\validation\canary\linker\src\linker\Linker\Pipeline.cs:line 134
     at Mono.Linker.Pipeline.Process(LinkContext context) in D:\validation\canary\linker\src\linker\Linker\Pipeline.cs:line 126
     at Mono.Linker.Driver.Run(ILogger customLogger) in D:\validation\canary\linker\src\linker\Linker\Driver.cs:line 449
     at Mono.Linker.Driver.Execute(String[] args, ILogger customLogger) in D:\validation\canary\linker\src\linker\Linker\Driver.cs:line 60
     at Mono.Linker.Driver.Main(String[] args) in D:\validation\canary\linker\src\linker\Linker\Driver.cs:line 49
D:\dotnet\runtime\eng\illink.targets(133,5): error MSB6006: "dotnet.exe" exited with code -532462766. [D:\dotnet\runtime\src\libraries\shims\generated\netstandard.csproj]
```

Seems like fix is to include a case with AssemblyAction.Skip that executes the Mark Assembly and the InitializeType method (mimic previous behavior).

PR #914, opened 2020-01-11
Added a case statement for when AssemblyAction is Skip in MarkStep

tlakollo:
This fixes error in issue #913 by using the same behavior it had before #810 fix


marek-safar:
This is an incorrect change we don't want to run mark on ignored assemblies. The proper fix is already merged

PR #921, opened 2020-01-22
Fix a bug in the reflection pattern recorder usage

vitek-karas:
The code was written such that it would report almost all calls it didn't consider (or try to analyze) as unrecognized patterns.

This change adds the test infra to validate that no extra reflection patterns are recorded on a given test. It then uses this on two tests to validate the functionality.

The fix adds a boolean which tracks if the code is trying to analyze a given reflection pattern or not. It only reports unrecognized pattern if it did try to analyze a pattern, but didn't report any result explicitly.

Also unified the code in reflection pattern matching to always `break` (the `returns` were left-overs from previous attempts at this functionality).

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -29,6 +29,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text.RegularExpressions;
 
@@ -2380,6 +2381,7 @@ namespace Mono.Linker.Steps {
 		struct ReflectionPatternContext : IDisposable
 		{
 			readonly LinkContext _context;
+			bool _patternAnalysisAttempted;
 			bool _patternReported;
 
 			public MethodDefinition MethodCalling { get; private set; }
@@ -2391,12 +2393,20 @@ namespace Mono.Linker.Steps {
 				MethodCalling = methodCalling;
 				MethodCalled = methodCalled;
 
+				_patternAnalysisAttempted = false;
 				_patternReported = false;
+			}
+
+			public void AnalyzingPattern ()
+			{
+				_patternAnalysisAttempted = true;
 			}
 
 			public void RecordRecognizedPattern<T> (T accessedItem, Action mark)
 				where T : IMemberDefinition
 			{
+				Debug.Assert (_patternAnalysisAttempted, "To correctly report all patterns, when starting to analyze a pattern the AnalyzingPattern must be called first.");
+
 				_context.Tracer.Push ($"Reflection-{accessedItem}");
 				try {
 					_patternReported = true;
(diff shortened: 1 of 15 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2409,13 +2419,15 @@ public void RecordRecognizedPattern<T> (T accessedItem, Action mark)
 
 			public void RecordUnrecognizedPattern (string message)
 			{
+				Debug.Assert (_patternAnalysisAttempted, "To correctly report all patterns, when starting to analyze a pattern the AnalyzingPattern must be called first.");
+
 				_patternReported = true;
 				_context.ReflectionPatternRecorder.UnrecognizedReflectionAccessPattern (MethodCalling, MethodCalled, message);
 			}
 
 			public void Dispose ()
```
marek-safar: What about removing Dispose altogether and add RecordUnrecognizedPattern to places where we are missing them if there are any?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2391,12 +2393,20 @@ public ReflectionPatternContext (LinkContext context, MethodDefinition methodCal
...
 
+			public void AnalyzingPattern ()
+			{
+				_patternAnalysisAttempted = true;
+			}
+
 			public void RecordRecognizedPattern<T> (T accessedItem, Action mark)
 				where T : IMemberDefinition
 			{
+				Debug.Assert (_patternAnalysisAttempted, "To correctly report all patterns, when starting to analyze a pattern the AnalyzingPattern must be called first.");
```
mrvoorhe: Why not make this a throw?  If all code paths correctly call `AnalyzingPattern` it would never get hit.

mrvoorhe:
Why is it such a bad thing if we record every usage of a reflection call regardless of whether or not we attempted to figure it out?

From a user perspective *every* reflection usage could cause linking problems.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: I added the `RecordUnrecognizedPattern` in all places, but I kept the Dispose as a way to check that it's everywhere. Still not super clean, but should be reasonably robust now. Added a bunch of tests to cover the interesting cases.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Not sure what's the pattern in this repo - basically if this happens it's a bug in the linker - so the exception would be something like "Internal linker problem". I'm used to having asserts in such cases, but if the pattern in this repo is to throw - I'll change it.

vitek-karas:
@mrvoorhe I do agree that ideally we would report all reflection calls as "dangerous" - and I'm working on that as part of the analysis for correctness work. But this change was simply trying to expose the existing functionality to that analysis without making large changes.

If left as is, the code actually reports things like `Type.ToString` as unrecognized patterns - which is clearly wrong no matter what.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2409,14 +2421,16 @@ public void RecordRecognizedPattern<T> (T accessedItem, Action mark)
...
+
 				_patternReported = true;
 				_context.ReflectionPatternRecorder.UnrecognizedReflectionAccessPattern (MethodCalling, MethodCalled, message);
 			}
 
 			public void Dispose ()
 			{
-				if (!_patternReported) {
-					_context.ReflectionPatternRecorder.UnrecognizedReflectionAccessPattern (MethodCalling, MethodCalled, $"Call to'`{MethodCalled.FullName}' inside '{MethodCalling.FullName}' could not be recognized as a known pattern");
+				if (_patternAnalysisAttempted && !_patternReported) {
```
marek-safar: This should be `Debug.Assert`

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2380,23 +2381,34 @@ protected virtual void MarkReflectionLikeDependencies (MethodBody body)
 		struct ReflectionPatternContext : IDisposable
 		{
 			readonly LinkContext _context;
+			bool _patternAnalysisAttempted;
```
marek-safar: Should be under `#if DEBUG`

src/linker/Linker.Steps/MarkStep.cs
mrvoorhe: My philosophy is general if something is wrong throw.  A user may have the code that requires hitting some case, but they will have a release build.  If it's a debug assert we'll never seen it.

If @marek-safar has no objections I'd prefer a throw.

PR #929, opened 2020-02-03
Do not ignore unresolved assemblies added with -r option

### LINKED ISSUE #922: NullReferenceException if passed -r DoesntExist
Linker can ignore unresolved references if used with `--skip-unresolved`. At the same time `-r` is supposed to root all public APIs in the specified assembly. This is implemented by `ResolveFromAssemblyStep`. The implementation is not handling unresolved assemblies though, it fails with `NullReferenceException` in this case.

Repro:
Prepare a sample console app and build/publish it. Then:
```
linker -a consoleapp -c link -u link -d F:\\consoleapp\bin\Debug\netcoreapp3.1\publish --skip-unresolved true -r DoesntExist
```

The linker will fail with:
```
Fatal error in IL Linker

Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Mono.Linker.Steps.ResolveFromAssemblyStep.HasInternalsVisibleTo(AssemblyDefinition assembly) in F:\mono\linker2\src\linker\Linker.Steps\ResolveFromAssemblyStep.cs:line 238
   at Mono.Linker.Steps.ResolveFromAssemblyStep.Process() in F:\mono\linker2\src\linker\Linker.Steps\ResolveFromAssemblyStep.cs:line 68
   at Mono.Linker.Steps.BaseStep.Process(LinkContext context) in F:\mono\linker2\src\linker\Linker.Steps\BaseStep.cs:line 58
   at Mono.Linker.Pipeline.ProcessStep(LinkContext context, IStep step) in F:\mono\linker2\src\linker\Linker\Pipeline.cs:line 134
   at Mono.Linker.Pipeline.Process(LinkContext context) in F:\mono\linker2\src\linker\Linker\Pipeline.cs:line 126          at Mono.Linker.Driver.Run(ILogger customLogger) in F:\mono\linker2\src\linker\Linker\Driver.cs:line 449
   at Mono.Linker.Driver.Execute(String[] args, ILogger customLogger) in F:\mono\linker2\src\linker\Linker\Driver.cs:line 64
   at Mono.Linker.Driver.Main(String[] args) in F:\mono\linker2\src\linker\Linker\Driver.cs:line 49
```

There are two issues with this:
* It's not very user friendly (and the code should not null ref in normal error conditions)
* It doesn't really tell the user what's wrong - I had to debug the linker to figure it out

mateoatr:
Fixes #922.

cc @vitek-karas

### DIFF
--- a/src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
+++ b/src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
@@ -63,8 +63,13 @@ namespace Mono.Linker.Steps
 			if (_assembly != null)
 				Context.Resolver.CacheAssembly (_assembly);
 
-			AssemblyDefinition assembly = _assembly ?? Context.Resolve (_file);
+			var ignoreUnresolved = Context.Resolver.IgnoreUnresolved;
+			if (_rootVisibility == RootVisibility.PublicAndFamily) {
+				Context.Resolver.IgnoreUnresolved = false;
+			}
 
+			AssemblyDefinition assembly = _assembly ?? Context.Resolve (_file);
+			Context.Resolver.IgnoreUnresolved = ignoreUnresolved;
 			if (_rootVisibility != RootVisibility.Any && HasInternalsVisibleTo (assembly)) {
 				_rootVisibility = RootVisibility.PublicAndFamilyAndAssembly;
 			}


src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -63,8 +63,13 @@ protected override void Process ()
 			if (_assembly != null)
 				Context.Resolver.CacheAssembly (_assembly);
 
-			AssemblyDefinition assembly = _assembly ?? Context.Resolve (_file);
+			var ignoreUnresolved = Context.Resolver.IgnoreUnresolved;
```
vitek-karas: Maybe add a comment why we're making this dance

sbomer:
`--skip-unresolved` should ignore resolution failures for transitively referenced assemblies, but arguably the user would still want to know if the linker didn't find an assembly that was explicitly requested to be rooted. Are we sure that we want `IgnoreUnresolved` to apply to `-r` assemblies, rather than giving an error indicating that a requested assembly wasn't found? My vote would be for the error.

mateoatr:
That's what this change introduces: we won't be ignoring any unresolved assemblies for the assemblies that were rooted by the user (`-r`'d), instead we'll be throwing an exception that shows what went wrong (that the requested assembly wasn't found).

sbomer:
Sorry, I misread the change. That sounds good, thanks!

src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
marek-safar: I don't think this is the right place to fix it. Please change `Assembly::ResolveFromReferences` instead of this hacking

vitek-karas: I don't really like the solution either, but I don't know if there's a better way. `AssemblyResolver::ResolveFromReferences` is called by all assembly resolution everywhere. The change needs to be targetted to the `ResolveFromAssemblyStep` - basically we still want to behave as per the setting (`IgnoreUnresolved`) everywhere, except when resolving assembly names which are specified directly on the command line, so `-r`. For those we would like to fail always, regardless of the setting. Making a change in the `AssemblyResovler` would change the behavior everywhere, not just for the command line inputs.

Maybe a better way to fix this would be to let let the `Context.Resovle` return `null` and check for that explicitly. If we want custom error message we would have to also try/catch `AssemblyResolutionException` as the `Resolve` may also throw in some cases.

marek-safar: >except when resolving assembly names which are specified directly on the command line, so -r

I agree, which is what https://github.com/mono/linker/blob/6e0105e0fc16b959510480298599ca09a5509a21/src/linker/Linker/AssemblyResolver.cs#L114 does. The `_references` is populated only from `-r` settings. Moving that call into separate try/catch with different logging should resolve this issue better.

vitek-karas: As far as I can tell the `_references` is populated from `--reference` command line parameters:
https://github.com/mono/linker/blob/6e0105e0fc16b959510480298599ca09a5509a21/src/linker/Linker/Driver.cs#L290-L294

This explicitly mentions that it's different from `-r`. The `_references` contains file paths (or should contain file paths, otherwise it will break elsewhere). The `-r` specifies simple assembly name. Both should fail if they don't exist:
* for `--reference` we will eventually try to open it as a file path (`ModuleDefinition.ReadModule`)
* for `-r` we run it through the `Context.Resolve` (so it might be resolved from the `--references` or through probing)

marek-safar: You are right, I keep forgetting about the `-r` hack.

PR #937, opened 2020-02-04
Run illink tests on CI

marek-safar:
- Renamed to ILLink.Tasks.IntegrationTests because it depends on the .nupkg and arcade schedules Test projects before Pack
- Fixed path to nupgk in nuget config
- Move override of Test target for NUnit project from root Directory.Build.targets. For some reason Condition doesn't work there so the target would always be overriden.
- Disabled MusicStore tests
  They were already disabled in https://github.com/mono/linker/commit/04eb53822f98749bdf6ab47009ba7c269f504eae because of https://github.com/aspnet/JitBench/issues/96 but looks like that got reverted somehow


src/ILLink.Tasks/ILLink.Tasks.csproj
```diff
@@ -1,17 +1,15 @@
 <Project Sdk="Microsoft.NET.Sdk">
-  <Import Project="$(MSBuildThisFileDirectory)../../eng/Versions.props" Condition=" '$(ArcadeBuild)' != 'true' " />
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0;net472</TargetFrameworks>
```
sbomer: Why are we getting rid of the full framework build? In the past this was required for the task to work on desktop MSBuild.

src/analyzer/analyzer.csproj
```diff
@@ -4,26 +4,24 @@
...
-    <TargetFrameworks Condition="'$(MSBuildRuntimeType)' == 'Core' And '$(OS)' != 'Windows_NT'">netcoreapp3.0</TargetFrameworks>
   </PropertyGroup>
 
-  <PropertyGroup Condition="!$(Configuration.StartsWith('illink'))">
+  <PropertyGroup Condition="'$(MonoBuild)' != ''">
     <TargetFrameworks>net471</TargetFrameworks>
   </PropertyGroup>
 
-  <PropertyGroup Condition="$(Configuration.EndsWith('Debug'))">
+  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
```
sbomer: I think we can remove these blocks that set DebugSymbols and Optimize now that we are using standard Debug/Release configurations. They were only here because they didn't get correct defaults for the nonstandard configs.

src/linker/Mono.Linker.csproj
```diff
@@ -5,37 +5,35 @@
...
-  <PropertyGroup Condition="!$(Configuration.StartsWith('illink'))">
+  <PropertyGroup Condition="'$(MonoBuild)' != ''">
     <AssemblyName>monolinker</AssemblyName>
     <AssemblyTitle>Mono.Linker</AssemblyTitle>
     <Description>Mono CIL Linker</Description>
     <TargetFrameworks>net471</TargetFrameworks>
   </PropertyGroup>
 
-  <PropertyGroup Condition="$(Configuration.EndsWith('Debug'))">
+  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
```
sbomer: Same as above, I believe these are now unnecessary.

src/ILLink.Tasks/ILLink.Tasks.csproj
akoeplinger: I added it back.

akoeplinger:
The Windows failure is kinda weird, I don't see it locally:

https://dev.azure.com/dnceng/public/_build/results?buildId=509609&view=ms.vss-test-web.build-test-results-tab&runId=16166058&resultId=100002&paneView=attachments
```
A fatal error was encountered. The library 'hostpolicy.dll' required to execute the application was not found in 'C:\Program Files\dotnet'.
```

src/ILLink.Tasks/ILLink.Tasks.csproj
marek-safar: @sbomer what is the scenario/workload for supporting .NET 4.7.2 from .NET5?

sbomer: @ericstj do you know if we will be able to unify and build MSBuild tasks for a single tfm in .NET5?

ericstj: Only if all VS/msbuild versions you support run on net5.

marek-safar: This is purely about .net core build tasks, non-netcore don't even build this project

ericstj: Visual Studio always builds with .NETFramework (since VS itself is running on .NETFramework), regardless of what the project targets.  I believe this task can be used from VS, is that correct?

sbomer: Yes, it can be. Thanks for confirming.

marek-safar: Wouldn't then make sense make this task netstandard2.0 to avoid dealing with multiple versions of it?

ericstj: No.  MSBuild tasks need to be runnable layouts, including dependencies specific to the target framework they are running in.  `netstandard2.0` is not runnable, dependencies will not be specific to the TFM.

PR #944, opened 2020-02-11
Fix integration test configuration

sbomer:
Look for ILLink.Tasks package matching the configuration used
for "dotnet test". @vitek-karas PTAL.

### DIFF
--- a/test/ILLink.Tasks.IntegrationTests/TestContext.cs
+++ b/test/ILLink.Tasks.IntegrationTests/TestContext.cs
@@ -74,9 +74,15 @@ namespace ILLink.Tests
 			Configuration = "Release";
 			TestBin = testBin;
 
+#if DEBUG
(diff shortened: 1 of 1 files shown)


test/ILLink.Tasks.IntegrationTests/TestContext.cs
```diff
@@ -74,9 +74,15 @@ public static void SetupDefaultContext()
 			Configuration = "Release";
```
vitek-karas: This can be removed

sbomer: I'm trying to separate the illink configuration from the test project config. My intention here is to publish+link test projects in Release, using either Debug or Release illink.

vitek-karas: Got it - thanks.

PR #946, opened 2020-02-12
Fix logging of unrecognized reflection patterns

### LINKED ISSUE #945: System.FormatException: Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
`_context.LogMessage()` is a `string.Format()`-style method: https://github.com/mono/linker/blob/7ff05f2648ea566faea6be927e44df9d329ba316/src/linker/Linker/LinkContext.cs#L426-L430

Thus, this method will eventually be equivalent to `string.Format (message)`.

If `message` contains `{`, then `string.Format()` (or equivalent) will attempt to interpret `message` as a format string, which it probably is not.

This method body should instead be:

```csharp
_context.LogMessage (MessageImportance.Low, "{0}", message);
```

This is causing issues with the Xamarin-android mono/2020-02 integration:

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3458646&view=logs&j=96fd57f5-f69e-53c7-3d47-f67e6cf9b93e&t=3eafbec1-f99f-54bb-979a-54e8f7485a0b

```
"…/xamarin-android/tests/BCL-Tests/Xamarin.Android.Bcl-Tests/Xamarin.Android.Bcl-Tests.csproj" (default target) (32) ->
(_LinkAssembliesShrink target) -> 
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000: Mono.Linker.MarkException: Error processing method: 'System.Void MonoTests.System.TypeTest::GetTypeWithDelegates()' in assembly: 'monodroid_corlib_test.dll' ---> System.FormatException: Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at System.Text.StringBuilder.AppendFormatHelper (System.IFormatProvider provider, System.String format, System.ParamsArray args)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at System.String.FormatHelper (System.IFormatProvider provider, System.String format, System.ParamsArray args)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at System.String.Format (System.String format, System.Object[] args)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at Xamarin.Android.Tasks.LinkAssemblies.LogMessage (Mono.Linker.MessageImportance importance, System.String message, System.Object[] values)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at Mono.Linker.LinkContext.LogMessage (Mono.Linker.MessageImportance importance, System.String message, System.Object[] values)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at Mono.Linker.LoggingReflectionPatternRecorder.UnrecognizedReflectionAccessPattern (Mono.Cecil.MethodDefinition sourceMethod, Mono.Cecil.MethodDefinition reflectionMethod, System.String message)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at Mono.Linker.Steps.MarkStep+ReflectionPatternContext.RecordUnrecognizedPattern (System.String message)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at Mono.Linker.Steps.MarkStep+ReflectionPatternDetector.Process (Mono.Linker.Steps.MarkStep+ReflectionPatternContext& reflectionContext)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at Mono.Linker.Steps.MarkStep.MarkReflectionLikeDependencies (Mono.Cecil.Cil.MethodBody body)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at Mono.Linker.Steps.MarkStep.MarkMethodBody (Mono.Cecil.Cil.MethodBody body)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at Mono.Linker.Steps.MarkStep.ProcessMethod (Mono.Cecil.MethodDefinition method)
…/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1951,5): error XALNK7000:   at Mono.Linker.Steps.MarkStep.ProcessQueue ()
```

vitek-karas:
LinkContext.LogMessage is behaving like string.Format, so if it gets a single parameter which has `{` in it, it will break.

Fixes #945

### DIFF
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -418,15 +418,15 @@ namespace Mono.Linker {
 			return (DisabledOptimizations & optimization) == 0;
 		}
 
-		public void LogMessage (string message, params object[] values)
+		public void LogMessage (string message)
 		{
-			LogMessage (MessageImportance.Normal, message, values);
+			LogMessage (MessageImportance.Normal, message);
 		}
 
-		public void LogMessage (MessageImportance importance, string message, params object [] values)
+		public void LogMessage (MessageImportance importance, string message)
 		{
 			if (LogMessages && Logger != null)
-				Logger.LogMessage (importance, message, values);
+				Logger.LogMessage (importance, "{0}", message);
 		}
 	}
 
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -143,7 +143,7 @@ namespace Mono.Linker {
 
 		public bool LogMessages { get; set; }
 
-		public ILogger Logger { get; set; } = new ConsoleLogger ();
+		public ILogger Logger { private get; set; } = new ConsoleLogger ();
 
 		public MarkingHelpers MarkingHelpers { get; private set; }
 
--- a/src/linker/Linker.Steps/BlacklistStep.cs
+++ b/src/linker/Linker.Steps/BlacklistStep.cs
@@ -46,11 +46,11 @@ namespace Mono.Linker.Steps {
 					continue;
 
 				try {
-					Context.LogMessage ("Processing resource linker descriptor: {0}", name);
+					Context.LogMessage ($"Processing resource linker descriptor: {name}");
 					AddToPipeline (GetResolveStep (name));
 				} catch (XmlException ex) {
 					/* This could happen if some broken XML file is included. */
-					Context.LogMessage ("Error processing {0}: {1}", name, ex);
+					Context.LogMessage ($"Error processing {name}: {ex}");
 				}
 			}
 
@@ -62,12 +62,12 @@ namespace Mono.Linker.Steps {
 									.Where (res => ShouldProcessAssemblyResource (GetAssemblyName (res.Name)))
 									.Cast<EmbeddedResource> ()) {
 					try {
-						Context.LogMessage ("Processing embedded resource linker descriptor: {0}", rsc.Name);
+						Context.LogMessage ($"Processing embedded resource linker descriptor: {rsc.Name}");
 
 						AddToPipeline (GetExternalResolveStep (rsc, asm));
 					} catch (XmlException ex) {
 						/* This could happen if some broken XML file is embedded. */
-						Context.LogMessage ("Error processing {0}: {1}", rsc.Name, ex);
+						Context.LogMessage ($"Error processing {rsc.Name}: {ex}");
 					}
 				}
 			}
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -477,7 +477,7 @@ namespace Mono.Linker.Steps {
 			var args = ca.ConstructorArguments;
 			if (args.Count >= 3 && args [2].Value is string assemblyName) {
 				if (!_context.Resolver.AssemblyCache.TryGetValue (assemblyName, out assembly)) {
-					_context.Logger.LogMessage (MessageImportance.Low, $"Could not resolve '{assemblyName}' assembly dependency");
+					_context.LogMessage (MessageImportance.Low, $"Could not resolve '{assemblyName}' assembly dependency");
 					return;
 				}
 			} else {
@@ -489,7 +489,7 @@ namespace Mono.Linker.Steps {
 				td = FindType (assembly ?? context.Module.Assembly, typeName);
 
 				if (td == null) {
-					_context.Logger.LogMessage (MessageImportance.Low, $"Could not resolve '{typeName}' type dependency");
+					_context.LogMessage (MessageImportance.Low, $"Could not resolve '{typeName}' type dependency");
 					return;
 				}
 			} else {
@@ -522,7 +522,7 @@ namespace Mono.Linker.Steps {
 			if (MarkDependencyField (td, member))
 				return;
 
-			_context.Logger.LogMessage (MessageImportance.High, $"Could not resolve dependency member '{member}' declared in type '{td.FullName}'");
+			_context.LogMessage (MessageImportance.High, $"Could not resolve dependency member '{member}' declared in type '{td.FullName}'");
 		}
 
 		static TypeDefinition FindType (AssemblyDefinition assembly, string fullName)
--- a/src/linker/Linker/LoggingReflectionPatternRecorder.cs
+++ b/src/linker/Linker/LoggingReflectionPatternRecorder.cs
@@ -43,7 +43,7 @@ namespace Mono.Linker
 
 		public void UnrecognizedReflectionAccessPattern (MethodDefinition sourceMethod, MethodDefinition reflectionMethod, string message)
 		{
-			_context.LogMessage (MessageImportance.Low, message);
+			_context.LogMessage (MessageImportance.Low, "{0}", message);
 		}
 	}
 }
--- a/src/linker/Linker/LoggingReflectionPatternRecorder.cs
+++ b/src/linker/Linker/LoggingReflectionPatternRecorder.cs
@@ -43,7 +43,7 @@ namespace Mono.Linker
 
 		public void UnrecognizedReflectionAccessPattern (MethodDefinition sourceMethod, MethodDefinition reflectionMethod, string message)
 		{
-			_context.LogMessage (MessageImportance.Low, "{0}", message);
+			_context.LogMessage (MessageImportance.Low, message);
 		}
 	}
 }


jonpryor:
There should also be an audit of all the other `LogMessage()` callers, for safety & sanity.  (It doesn't necessarily need to be done as part of *this* PR, but it *should* be done.)

For example, a quick `git grep LogMessage` shows these other spots of concern:

https://github.com/mono/linker/blob/705efee5e3c108a0d51c6682c4e410b559fd15c3/src/ILLink.Tasks/FindNativeDeps.cs#L74

https://github.com/mono/linker/blob/705efee5e3c108a0d51c6682c4e410b559fd15c3/src/linker/Linker.Steps/BodySubstituterStep.cs#L63

Aside: Assembly filenames names *can* contain `{` and `}`!

```bash
% csc /out:foo{}.exe /nologo hw.cs
% ls -l foo{}.exe
-rw-r--r--  1 jon  staff  3584 Feb 12 09:32 foo{}.exe
```

https://github.com/mono/linker/blob/705efee5e3c108a0d51c6682c4e410b559fd15c3/src/linker/Linker.Steps/BodySubstituterStep.cs#L86

Aside: In the presence of IL rewriters and code obfuscators, nothing prevents type names and member names from containing `{` and `}` either!

---

*Alternatively*, it may be better to instead add `LogMessage (string)` and `LogMessage(MessageImportance, string)` overloads, to avoid the `string.Format()`-style behavior:

```csharp
partial class ConsoleLogger {
    public void LogMessage (MessageImportance importance, string message) =>
        LogMessage (importance, "{0}", message);
}

partial class LinkContext {
    public void LogMessage (string message) => LogMessage ("{0}", message);

    public void LogMessage (MessageImportance importance, string message) =>
        LogMessage (importance, "{0}", message);
}
```

This won't help with "pure" `ILogger`-based invocations, e.g. through the `LinkContext.Logger` property.  An extension method *might* be better?

```csharp
public static class ILoggerExtensions {
    public static void LogMessage (this ILogger self, MessageImportance importance, string message) =>
        self.LogMessage (importance, "{0}", message);
}
```

vitek-karas:
I removed the params from the context methods completely - almost all call sites were not actually using it.

vitek-karas:
I also removed the `LinkContext.Logger` getter. I kept the setter public as it's possible somebody is using it to override the default logger. Removing the getter is also potentially breaking (it's a public API), so I'm opened to suggestions, but this would completely eliminate this class of issues from the linker.

src/linker/Linker/LinkContext.cs
```diff
@@ -143,7 +143,7 @@ public bool IgnoreUnresolved
 
 		public bool LogMessages { get; set; }
 
-		public ILogger Logger { get; set; } = new ConsoleLogger ();
+		public ILogger Logger { private get; set; } = new ConsoleLogger ();
```
jonpryor: set-only properties are generally frowned upon:

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property

> ❌ DO NOT provide set-only properties or properties with the setter having broader accessibility than the getter.

vitek-karas: I can definitely change it to a SetLogger method. I just wanted to avoid breaking people with this. I noticed that the error this started with has a custom logger `Xamarin.Android.Tasks.LinkAssemblies.LogMessage`. Would you know how is that logger registered?

jonpryor: > Would you know how is that logger registered?

The [`<LinkAssemblies/>` MSBuild task implements the `ILogger` interface](https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssemblies.cs#L24):

```csharp
public class LinkAssemblies : AndroidTask, Mono.Linker.ILogger
```

The [`Linker.Process()`](https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/Linker.cs#L18-L31) is invoked, passing an instance of [`LinkAssemblies` as the `ILogger` parameter](https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssemblies.cs#L126):

```csharp
Linker.Process (options, this, out link_context);
```

`Linker.Process()`, meanwhile, uses the `ILogger` value to set the [`AndroidLinkContext.Logger` property](https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AndroidLinkContext.cs#L6), and `AndroidLinkContext` IS-A `LinkContext`, i.e. it's setting `LinkContext.Logger`:

https://github.com/mono/linker/blob/0c63b9ed74130695ef270002d9d9c572201da90c/src/linker/Linker/LinkContext.cs#L146

...which is the property that I was commenting on here...

Removing this property setter ("rename to `SetLogger()`") would thus result in a compatibility change, breaking the xamarin-android build.

All of this *does* show a problem on the xamarin-android side: https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs#L29

```csharp
CheckAppDomainUsageUnconditional (assembly, (string msg) => Context.Logger.LogMessage (MessageImportance.High, msg));
```

Here we have the ~same issue I was originally complaining about -- though it's in *my* repo, so I should go fix that now -- as the `Context.Logger.LogMessge()` invocation is "really" `Context.Logger.LogMessage(MessageImportance.High, msg, new object[0])`, so if `msg` contains any `{` characters, things will still blow up.

jonpryor: xamarin-android's `CheckAppDomainUsageUnconditional()` usage fixed in:

https://github.com/xamarin/xamarin-android/commit/c683788921265f6a4ba22b2af71228bea48b05fb

src/linker/Linker/LinkContext.cs
```diff
@@ -418,15 +418,15 @@ public bool IsOptimizationEnabled (CodeOptimizations optimization)
 			return (DisabledOptimizations & optimization) == 0;
 		}
 
-		public void LogMessage (string message, params object[] values)
```
jonpryor: > Removing this property setter ("rename to SetLogger()") would thus result in a compatibility change

Speaking of "compatibility changes", this commit *broke* xamarin-android compilation, as it removed the `LinkContext.LogMessage(string, params object[])` overload, resulting in CS1501 errors in xamarin-android:

https://github.com/xamarin/xamarin-android/commit/856460dbe2ec65cb5ca1111880ab8940419fe945

This raises a somewhat important question: *what are* the compatibility constraints on this repo?  Is source compatibility supposed to be present?  Binary compatibility?  No compatibility?

If compatibility is required, this repo may want to look into using `Microsoft.DotNet.ApiCompat.exe`.  See also?

  * https://github.com/xamarin/xamarin-android/commit/07e7477126fdc3eb68cca38347d1b8698d334874
  * https://github.com/xamarin/xamarin-android/blob/master/tests/api-compatibility/README.md

marek-safar: For existing Xamarin like workloads, there is no compatibility as we integrate at the source code level. Going forward with .NET 5 the plan is to switch to binary backward compatibility.

PR #951, opened 2020-02-17
Add a --exclude-feature deserialization to link out the deserialization ctors etc.

vargaz:


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1046,7 +1046,8 @@ void MarkScope (IMetadataScope scope)
 		protected virtual void MarkSerializable (TypeDefinition type)
 		{
 			MarkDefaultConstructor (type);
-			MarkMethodsIf (type.Methods, IsSpecialSerializationConstructor);
+			if (!_context.IsFeatureExcluded ("deserialization"))
```
marek-safar: I think it'd be better to do the check at caller of `MarkSerializable`

vargaz: The default ctor needs to be retained.

marek-safar: Why?

vargaz: I get some runtime assertions without it, I think things just worked by accident because some types whose ctors are needed at runtime were marked as serializable.

vargaz: Also, not sure what the default ctor has to do with serialization at all.

marek-safar: Yeah, this looks like quirk for `XmlSerializer` support which has constructor of System.Type and requires default constructor to be present but no SerializableAttribute to be applied!

Made a new tracking issue #957

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1135,7 +1136,8 @@ protected virtual TypeDefinition MarkType (TypeReference reference)
 				if (ShouldMarkTypeStaticConstructor (type))
 					MarkStaticConstructor (type);
 
-				MarkMethodsIf (type.Methods, HasSerializationAttribute);
+				if (!_context.IsFeatureExcluded ("deserialization"))
+					MarkMethodsIf (type.Methods, HasSerializationAttribute);
```
marek-safar: This marks also serialization methods, if the feature is about deserialization only we should exclude them

PR #955, opened 2020-02-19
Code fixes

sbomer:
This applies suggested code fixes except for those (slices) that aren't supported on full framework. I decided not to do the testcases for now, since some of them are arguably more readable without the fixes - and also partially because the omnisharp analyzer only kicks in for open files, and there are a lot of testcases.

### DIFF
--- a/src/ILLink.Tasks/FindNativeDeps.cs
+++ b/src/ILLink.Tasks/FindNativeDeps.cs
@@ -53,25 +53,26 @@ namespace ILLink.Tasks
 
 			var managedAssemblies = ManagedAssemblyPaths.Select (i => i.ItemSpec).ToArray ();
 			foreach (string managedAssembly in managedAssemblies) {
-				using var peReader = new PEReader(new FileStream (managedAssembly, FileMode.Open, FileAccess.Read, FileShare.Read));
-				if (peReader.HasMetadata) {
-					var reader = peReader.GetMetadataReader ();
-					for (int i = 1, count = reader.GetTableRowCount (TableIndex.ModuleRef); i <= count; i++) {
-						var moduleRef = reader.GetModuleReference (MetadataTokens.ModuleReferenceHandle (i));
-						var moduleName = reader.GetString (moduleRef.Name);
+				using (var peReader = new PEReader(new FileStream (managedAssembly, FileMode.Open, FileAccess.Read, FileShare.Read))) {
+					if (peReader.HasMetadata) {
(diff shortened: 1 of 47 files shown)


src/ILLink.Tasks/FindNativeDeps.cs
```diff
@@ -53,26 +53,25 @@ public override bool Execute ()
 
 			var managedAssemblies = ManagedAssemblyPaths.Select (i => i.ItemSpec).ToArray ();
 			foreach (string managedAssembly in managedAssemblies) {
-				using (var peReader = new PEReader(new FileStream (managedAssembly, FileMode.Open, FileAccess.Read, FileShare.Read))) {
-					if (peReader.HasMetadata) {
-						var reader = peReader.GetMetadataReader ();
-						for (int i = 1, count = reader.GetTableRowCount (TableIndex.ModuleRef); i <= count; i++) {
-							var moduleRef = reader.GetModuleReference (MetadataTokens.ModuleReferenceHandle (i));
-							var moduleName = reader.GetString (moduleRef.Name);
+				using var peReader = new PEReader(new FileStream (managedAssembly, FileMode.Open, FileAccess.Read, FileShare.Read));
```
marek-safar: I am really not sure this using syntax is better. It really hides the scope where the file can be released.

src/analyzer/common/Mono.Options/Options.cs
```diff
@@ -212,29 +212,28 @@ private static IEnumerable<string> CreateWrappedLinesIterator (string self, IEnu
 				yield return string.Empty;
```
marek-safar: Please undo changes in this file, it's a copy of a shared file

src/linker/Linker.Steps/BlacklistStep.cs
```diff
@@ -126,16 +126,14 @@ static ResolveFromXmlStep GetResolveStep (string descriptor)
 
 		protected static XPathDocument GetExternalDescriptor (EmbeddedResource resource)
 		{
-			using (var sr = new StreamReader (resource.GetResourceStream ())) {
-				return new XPathDocument (sr);
-			}
+			using var sr = new StreamReader (resource.GetResourceStream ());
```
marek-safar: Is this syntax really better?

mrvoorhe: I agree.  I'm not a big fan of the bracketless using

sbomer:
I undid the Options.cs and using changes. PTAL.

Issue #950, opened 2020-02-14
RemoveUnreachableBlocksStep.RemoveConditions incorrectly deletes IL code


tlakollo:
RemoveUnreachableBlocksStep.RemoveConditions deletes code that later on should have been used by RemoveUnreachableBlocksStep.GetReachableInstructionsMap making GetReachableInstructionsMap logic to behave incorrectly deleting code that should have being kept.

An example of this is the following function in https://github.com/dotnet/runtime/blob/master/src/coreclr/src/System.Private.CoreLib/src/Internal/Runtime/InteropServices/ComActivator.cs 
RegisterClassForTypeInternal function
```
/// <summary>
/// Internal entry point for registering a managed COM server API from native code
/// </summary>
/// <param name="cxtInt">Reference to a <see cref="ComActivationContextInternal"/> instance</param>
[CLSCompliant(false)]
public static unsafe int RegisterClassForTypeInternal(ref ComActivationContextInternal cxtInt)
{
    if (IsLoggingEnabled())
    {
        Log(
            $@"{nameof(RegisterClassForTypeInternal)} arguments:
            {cxtInt.ClassId}
            {cxtInt.InterfaceId}
            0x{(ulong)cxtInt.AssemblyPathBuffer:x}
            0x{(ulong)cxtInt.AssemblyNameBuffer:x}
            0x{(ulong)cxtInt.TypeNameBuffer:x}
            0x{cxtInt.ClassFactoryDest.ToInt64():x}");
    }

    if (cxtInt.InterfaceId != Guid.Empty
        || cxtInt.ClassFactoryDest != IntPtr.Zero)
    {
        throw new ArgumentException();
    }

    try
    {
        var cxt = ComActivationContext.Create(ref cxtInt);
        ClassRegistrationScenarioForType(cxt, register: true);
    }
    catch (Exception e)
    {
        return e.HResult;
    }

    return 0;
}
```
I will focus at this moment on the following code 
```
    if (cxtInt.InterfaceId != Guid.Empty
        || cxtInt.ClassFactoryDest != IntPtr.Zero)
    {
        throw new ArgumentException();
    }
```
The generated IL code for this piece of code
```
....
[58]	{IL_008e: ldarg.0}
[59]	{IL_008f: ldfld System.Guid Internal.Runtime.InteropServices.ComActivationContextInternal::InterfaceId}
[60]	{IL_0094: ldsfld System.Guid System.Guid::Empty}
[...]

PR #956, opened 2020-02-19
RemoveConditions+GetReachableInstructionsMap test

tlakollo:
Added a new test that reproduces error stated in https://github.com/mono/linker/issues/950
At this moment test generates an assert 
```
Expected method `System.Void Mono.Linker.Tests.Cases.UnreachableBlock.ComplexConditions::Test_2() to have it's exception handlers unchanged, however, the exception handlers differ from the original
```
The code does remove an exception handler which is one of the problems stated in the github issue so the "Expected" flags cannot be tested and might need to be updated once the issue is solved


marek-safar:
Thank you for the test case! I converted it to more robust one in #963

Issue #769, opened 2019-09-30
Output MSBuild-friendly warnings/errors


sbomer:
Currently the linker will print a stack trace to stderr if it fails. @tlakollo changed the ILLink.Tasks stderr importance so that this will show up in msbuild logs by default: https://github.com/mono/linker/pull/719.

Warnings logged to stdout, for example warnings about unresolved assemblies (or in the future, linker-unfriendly patterns), should also be made visible by default. In particular we should ensure that they show up as warnings/errors in MSBuild. Since we run out-of-process from msbuild, it seems like the standard way to do this is to format messages with a prefix like "warning: message", which will cause it to be treated as an actual warning and surfaced in IDEs.

Rather than prepending such a prefix to individual messages, we should probably add a few helpers to make the intentions clear, so that the logged messages don't get altered unintentionally.

Related to the larger diagnostics work: https://github.com/mono/linker/issues/592

danmosemsft:
definition of the format
https://docs.microsoft.com/en-us/archive/blogs/msbuild/msbuild-visual-studio-aware-error-messages-and-message-formats
or, better, from the source
https://github.com/microsoft/msbuild/blob/master/src/Shared/CanonicalError.cs

sbomer:
Just leaving a note here that I think [unresolved `PreserveDependencyAttribute`s](https://github.com/mono/linker/blob/2818dc1d5f405819fac7caf5738b205408314202/src/linker/Linker.Steps/MarkStep.cs#L480), among other things, should be surfaced as MSBuild warnings.

PR #964, opened 2020-02-21
Change fatal error messages

tlakollo:
A small change in case of a fatal error message, this is compliant with MSBuild error messages and provide a better experience to the user

### DIFF
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -68,7 +68,8 @@ namespace Mono.Linker {
 					return 1;
 
 			} catch {
-				Console.Error.WriteLine ("Fatal error in {0}", _linker);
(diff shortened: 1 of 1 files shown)


tlakollo:
This should improve a little https://github.com/mono/linker/issues/769

src/linker/Linker/Driver.cs
```diff
@@ -67,8 +67,9 @@ public static int Execute (string[] args, ILogger customLogger = null)
 				if (!driver.Run (customLogger))
 					return 1;
 
-			} catch {
-				Console.Error.WriteLine ("Fatal error in {0}", _linker);
+			} catch (Exception e) {
+				String source = e.TargetSite.DeclaringType.FullName + "." + e.TargetSite.Name;
```
marek-safar: I think printing e.Message should be enough in this case

tlakollo:
There is a better way of handling errors in #1151

PR #967, opened 2020-02-24
Ignore missing .ctor or .cctor in PreserveDependency attribute

### LINKED ISSUE #966: Relax PreserveDependency check for .ctor and .cctor
Currently linker issues a warning if the member referenced in `PreserveDependency` attribute doesn't exist. This makes perfect sense for most cases, but in case of `.ctor` and `.cctor` it's a bit tricky. Default constructor as well as static constructor are often compiler generated and otherwise unrelated changes to the code (or potentially even the compiler) may add or remove these. The `PreserveDependency` attribute should express intent, in this case dependency on being able to instantiate the target type or be able to use static fields on it. The fact that the compiler/implementation may or may not have the default .ctor/static .cctor is secondary.

The proposal is to change the check in the linker and don't issue a warning if the dependency on `.ctor` or `.cctor` is specified and the method doesn't exist.

Related to https://github.com/dotnet/runtime/pull/31718.

vitek-karas:
Currently linker issues a warning if the member referenced in PreserveDependency attribute doesn't exist. This makes perfect sense for most cases, but in case of .ctor and .cctor it's a bit tricky. Default constructor as well as static constructor are often compiler generated and otherwise unrelated changes to the code (or potentially even the compiler) may add or remove these. The PreserveDependency attribute should express intent, in this case dependency on being able to instantiate the target type or be able to use static fields on it. The fact that the compiler/implementation may or may not have the default .ctor/static .cctor is secondary.

This change will skip emitting the warning if the dependency is for default .ctor or static .cctor.

Fixed couple of small problems in the existing tests and added the new cases.

Fixes #966

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -516,6 +516,10 @@ namespace Mono.Linker.Steps {
 			if (MarkDependencyField (td, member))
 				return;
 
+			// Don't report missing .ctor() or .cctor() as those can be compiler generated and thus hard to control their existence of
+			if (((member == ".ctor") || (member == ".cctor")) && signature.Length == 0)
+				return;
+
 			_context.LogMessage (MessageImportance.High, $"Could not resolve dependency member '{member}' declared in type '{td.FullName}'");
 		}
 
--- a/test/Mono.Linker.Tests.Cases/PreserveDependencies/PreserveDependencyMethod.cs
+++ b/test/Mono.Linker.Tests.Cases/PreserveDependencies/PreserveDependencyMethod.cs
@@ -2,6 +2,13 @@
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 
 namespace Mono.Linker.Tests.Cases.PreserveDependencies {
+	
+	[LogContains("Could not resolve 'Mono.Linker.Tests.Cases.PreserveDependencies.MissingType' type dependency")]
+	[LogContains("Could not resolve dependency member 'MissingMethod' declared in type 'Mono.Linker.Tests.Cases.PreserveDependencies.C'")]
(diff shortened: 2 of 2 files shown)


mrvoorhe:
My understanding has been that `PreserveDependency` is brutally explicit mechanism.  No smarts, no wild cards.  You say preserve 1 thing and it does it or tells you something is wrong.

What if someone was copy pasting from an existing usage, say
```
[PreserveDepdency("Foo()")]
``` 
and changed it to 
```
[PreserveDepdency("Bar()")]
```

But forgot that `Bar` doesn't have a default ctor, it's ctor takes 1 argument.  They wanted to change it to
```
[PreserveDepdency("Bar(System.Int32)")]
```

Now they will get no warning

sbomer:
As another alternative to help with https://github.com/dotnet/runtime/pull/31718, would it make sense to detect simple calls to RuntimeHelpers.RunClassConstructor?

vitek-karas:
I counted the pattern matching as a worse solution, but it would definitely work.
Potential middle ground: only allow unresovled static `.cctor` in `PreserveDependency`. That should not have the same problem as default `.ctor` (as there's no possible overloads).
But if everybody prefers the pattern match for `RunClassConstructor` I'll be happy to do that instead.

sbomer:
FWIW I personally prefer pattern matching unless there are downsides I'm not thinking of. Supporting explicit `PreserveDependency` on a `.cctor` that doesn't exist seems confusing, whereas detecting simple invocations of `RunClassConstructor` falls in line with the approach we are taking for other unsafe APIs.

marek-safar:
As there is no easy way for the linker to detect if the instance constructor is compiler-generated this is disabling too much.  It might be worth exploring how ugly it will be to have pattern detection for RunClassConstructor as we might eventually need it for RunModuleConstructor as well.

vitek-karas:
Thanks for the feedback. I created #970 to track the different solution.

PR #968, opened 2020-02-24
Output Mono.Linker warnings through MSBuild

mateoatr:
Related to #769


mateoatr:
Please take a look @vitek-karas @marek-safar I've added a simple test to give an example of how this would be used and how it'd look like and created issues #986 #987 related to this.

src/linker/Linker/LinkContext.cs
```diff
@@ -435,6 +435,17 @@ public void LogMessage (MessageImportance importance, string message)
...
 				Logger.LogMessage (importance, "{0}", message);
 		}
+
+		public void LogMessage (Message message)
+		{
+			if (Logger == null)
+				return;
+
+			if (message.Category == MessageCategory.Info && !LogMessages)
+				return;
```
vitek-karas: This is different from the existing `LogMessage (MessageImportance, string)` - that one will not log at all if `LogMessages` is false. If we're changing the behavior here I think it's worth a comment as to why.

src/linker/Linker/MessageCode.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Mono.Linker {
+	public enum MessageCode {
+		L100, // Error
+		L400, // Warning
+		L900, // Info.
```
vitek-karas: Did you talk to the Roslyn about the diagnostic IDs? We need to make sure that the codes we're about to add are unique and ideally that they follow some kind of pattern.

test/Mono.Linker.Tests.Cases/Logging/Dependencies/LogStep.cs
```diff
@@ -0,0 +1,20 @@
...
+	{
+		public void Process (LinkContext context)
+		{
+			MessageOrigin msgOrigin = new MessageOrigin ("logtest");
+			Message msgError = new Message (msgOrigin, MessageCategory.Error, MessageCode.L100, text: "Error");
+			Message msgWarning = new Message (msgOrigin, MessageCategory.Warning, MessageCode.L400, text: "Warning");
+			Message msgInfo = new Message (new MessageOrigin(), MessageCategory.Info, MessageCode.L900, text: "Info");
+			context.LogMessage (msgError);
+			context.LogMessage (msgWarning);
+			context.LogMessage (msgInfo);
```
vitek-karas: Add one more which fills in Line/Column info.

src/linker/Linker/Message.cs
```diff
@@ -0,0 +1,44 @@
...
+		
+		public string Subcategory { get; }
+		
+		public MessageCategory Category { get; }
+		
+		public MessageCode Code { get; }
+		
+		public string Text { get; }
+
+		public Message (
```
marek-safar: This constructor is not usable. Why is text optional? Why is origin first when it adds no information by default?

I'd suggest you add only data you know how to pass in. Introducing API without using them is not good start.

src/linker/Linker/Message.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Mono.Linker
+{
+	public class Message
```
marek-safar: This is not "Message" but container of more things. The name should reflect that.

src/linker/Linker/Message.cs
```diff
@@ -0,0 +1,44 @@
...
+	{
+		public MessageOrigin Origin { get; }
+		
+		public string Subcategory { get; }
+		
+		public MessageCategory Category { get; }
```
marek-safar: There is already `MessageImportance` with same purposes and would need unification

src/linker/Linker/Message.cs
```diff
@@ -0,0 +1,44 @@
...
+		
+		public MessageCode Code { get; }
```
marek-safar: What is this for?

src/linker/Linker/Message.cs
marek-safar: I don't think this should be `class`

src/linker/Linker/Message.cs
```diff
@@ -0,0 +1,44 @@
...
+			Origin = origin;
+			Category = category;
+			Code = code;
+			Subcategory = subcategory;
+			Text = text;
+		}
+
+		public override string ToString()
+		{
+			string message = string.Format ("{0}: {1}{2} {3}{4}",
```
marek-safar: Is this the output format compatible with msbuild?

src/linker/Linker/MessageOrigin.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace Mono.Linker {
+	public class MessageOrigin {
+
+#if NETCOREAPP
+		private const string DefaultOrigin = "illinker";
+#else
+		private const string DefaultOrigin = "monolinker";
+#endif
+
+		public string Origin { get; }
```
marek-safar: What is the value of this property?

src/linker/Linker/MessageOrigin.cs
```diff
@@ -0,0 +1,36 @@
...
+		
+		public int? Line { get; }
```
marek-safar: I'd fold this into other class (and make non-nullable)

test/Mono.Linker.Tests.Cases/Logging/CommonLogs.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.Logging {
+
+#if !NETCOREAPP
+	[IgnoreTestCase ("Can be enabled once MonoBuild produces a dll from which we can grab the types in the Mono.Linker namespace.")]
+#else
+	[SetupCompileBefore ("LogStep.dll", new [] { "Dependencies/LogStep.cs" }, new [] { "illink.dll" })]
```
mrvoorhe: I do not believe that referencing the linker from a test case is a good path to go down.  The test cases were intended to test controlled use cases.  linking the linker itself is not a controlled test case and is bound to result in tests that are high maintenance.

If you want to write a test like this I think it should live in `Mono.Linker.Tests` as a normal unit test.  Maybe something similar to `IndividualTests` if you still want to use a test case and the testing infrastructure to drive running of the linker.

mrvoorhe: Thinking about it more...I don't think using the test case infrastructure in any way is the way to go here.

It looks more like you want to write a unit test that sets up a `LinkContext` and then calls
```
				MessageOrigin msgOrigin = new MessageOrigin ("logtest");
			Message msgError = new Message (msgOrigin, MessageCategory.Error, MessageCode.L100, text: "Error");
			Message msgWarning = new Message (msgOrigin, MessageCategory.Warning, MessageCode.L400, text: "Warning");
			Message msgInfo = new Message (new MessageOrigin(), MessageCategory.Info, MessageCode.L900, text: "Info");
			context.LogMessage (msgError);
			context.LogMessage (msgWarning);
			context.LogMessage (msgInfo);
```

And then you assert what is in the `ILogger` class that you setup with the `LinkContext`.

If you wanted to test logging end to end via a test case then I would setup a scenario that cases a message to be logged and then assert that message is in the log.  If you don't have messages hooked up yet, then I would hold off on a test case until you do.

src/linker/Linker/Message.cs
mateoatr: Thanks. I'll unify it in another PR, to not add much noise in this one.

src/linker/Linker/Message.cs
mateoatr: MSBuild requires us to identify all warnings and errors with a code (e.g. CS0029).

src/linker/Linker/Message.cs
mateoatr: For the order of the parameters I simply followed the order in which these appear in a MSBuild message (the format is described [here](https://docs.microsoft.com/en-us/archive/blogs/msbuild/msbuild-visual-studio-aware-error-messages-and-message-formats)). `Text` only adds further information for the user about the message he/she's seeing.

src/linker/Linker/Message.cs
mateoatr: Yes.

src/linker/Linker/MessageCode.cs
mateoatr: They are still debating which way to go on this. For now, I simply used `IL` as a prefix.

test/Mono.Linker.Tests.Cases/Logging/CommonLogs.cs
mateoatr: I'll move the test to `Mono.Linker.Tests` for now. I agree that a better way to test this would be to have a scenario that triggers a warning/error to be logged, but in this PR I'm just adding the basic infrastructure for logging support through MSBuild. I'll do as you advice and write tests that trigger the warnings once I update all the places where we log warnings/errors in the linker to use this.

src/linker/Linker/ConsoleLogger.cs
```diff
@@ -8,9 +8,9 @@ public void LogMessage (MessageImportance importance, string message, params obj
 			Console.WriteLine (message, values);
 		}
 
-		public void LogMessage (Message message)
+		public void LogMessage (MSBuildMessageContainer MSBuildMessage)
```
vitek-karas: parameter names should be `msBuildMessage` - or just `message`.

src/linker/Linker/Position.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace Mono.Linker
+{
+	public class Position
```
vitek-karas: I would prefer a bit more specific name - maybe `MessageSourcePosition`

test/Mono.Linker.Tests.Cases.Expectations/Assertions/LogContainsExactAttribute.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Assertions
+{
+	[AttributeUsage (AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
+	public class LogContainsExactAttribute : EnableLoggerAttribute
```
vitek-karas: Maybe we should just add a parameter to the existing attribute:
* `LogContains("Text")` - would do `string.Contains`
* `LogContains("Test", RegularExpression = true) - would do the existing regex match

I think that is cleaner than having another attribute

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -603,6 +603,12 @@ void VerifyLoggedMessages (AssemblyDefinition original, LinkerTestLogger logger)
 							$"Expected to find logged message matching `{expectedMessagePattern}`, but no such message was found.{Environment.NewLine}Logged messages:{Environment.NewLine}{allMessages}");
 					}
 
+					if (attr.AttributeType.Resolve ().Name == nameof (LogContainsExactAttribute)) {
+						var expectedMessage = (string)attr.ConstructorArguments [0].Value;
+						Assert.That (logger.Messages.Where (s => s.Message.Contains (expectedMessage)), Is.Not.Null,
```
vitek-karas: Isn't this always true? `.Where` will always return a non-null enumerable - so the condition doesn't actually validate anything (If I read it right) - The `.Any` used above is better - or if there's an `Is` predicate for a non-empty collection, use that.

src/linker/Linker/MSBuildMessageContainer.cs
```diff
@@ -0,0 +1,58 @@
...
+		/// </summary>
+		public MessageCode Code { get; }
+		
+		/// <summary>
+		/// Optional user friendly text describing the error or warning.
+		/// </summary>
+		public string Text { get; }
+
+		public MSBuildMessageContainer (
+			MessageOrigin origin,
```
marek-safar: This think should be the last parameter

src/linker/Linker/MessageCode.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Mono.Linker {
+	public enum MessageCode {
+		IL1000, // Error
```
marek-safar: Why do we need both code and category if code range included category data?

src/linker/Linker/Position.cs
marek-safar: Is the extra type actually helpful, why not just have Line, Column in MessageOrigin?

src/linker/Linker/MessageCode.cs
vitek-karas: The message codes should actually represent specific errors/warnings. Having the numbering scheme to differentiate between errors and warnings is helpful though. Basically once we start using this system each failure should get its own code (and it should fail within a certain range, that's all).

marek-safar: I agree that having specific ranges for different kind of issues and that's why I am not sure we need to duplicate the information. If we settle down to for example

0 - 1000 error
1001 - 10000 warning
10001 - 490000 for the rest

50000 - * for custom steps

then we should either pass the code in as int with enum flag or different method name where we could verify it or build this massive enum list but why to do both?

vitek-karas: Good point - that makes sense. With one glitch: If custom steps go to 50000+ those will also need to be able to to differentiate error/warning/info...

marek-safar: Right so that's why I think
```c#
public MSBuildMessageContainer (int code, MessageCategory category, string message, string subcategory = null, MessageOrigin origin = null)
{
// chech if code matches category range
}
```

might work better

src/linker/Linker/MSBuildMessageContainer.cs
```diff
@@ -0,0 +1,82 @@
...
+		/// </summary>
+		public int Code { get; }
+		
+		/// <summary>
+		/// Optional user friendly text describing the error or warning.
+		/// </summary>
+		public string Text { get; }
+
+		public MSBuildMessageContainer (
+			int code,
```
marek-safar: Please fix the order, text is the most important and should never be empty and not last in argument

src/linker/Linker/LinkContext.cs
```diff
@@ -435,6 +435,12 @@ public void LogMessage (MessageImportance importance, string message)
 			if (LogMessages && Logger != null)
 				Logger.LogMessage (importance, "{0}", message);
 		}
+
+		public void LogMessage (MSBuildMessageContainer msBuildMessage)
```
marek-safar:
```suggestion
		public void LogMessage (MSBuildMessageContainer msbuildMessage)
```

src/linker/Linker/MSBuildMessageContainer.cs
```diff
@@ -0,0 +1,82 @@
...
+		/// Code identifier for errors and warnings reported by the IL linker.
+		/// </summary>
+		public int Code { get; }
+		
+		/// <summary>
+		/// Optional user friendly text describing the error or warning.
+		/// </summary>
+		public string Text { get; }
+
+		public MSBuildMessageContainer (
```
marek-safar:
```suggestion
		public MSBuildMessageContainer (int code, MessageCategory category, string text, string subCategory = null, MessageOrigin origin = null)
```

src/linker/Linker/MessageOrigin.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace Mono.Linker
+{
+	public class MessageOrigin {
+
+#if NETCOREAPP
+		private const string ToolName = "illinker";
```
marek-safar: Move that to MSBuildMessageContainer and use it only when MessageOrigin is null

src/linker/Linker/MSBuildMessageContainer.cs
```diff
@@ -68,14 +63,20 @@ namespace Mono.Linker
...
-		public override string ToString()
+		public override string ToString ()
 		{
 			string message = string.Format ("{0}: {1}{2} {3}{4}",
-				Origin.ToString(),
-				Subcategory != MessageSubcategory.None ? Subcategory + " " : "",
-				Category.ToString().ToLowerInvariant(),
-				"IL" + Code.ToString("D4"),
-				!String.IsNullOrEmpty(Text) ? ": " + Text : Text);
+						Origin.ToString() == null ?
```
marek-safar: This will NRE when Origin is null. Please move the logic out of the expression

marek-safar:
@mateoatr I went ahead and clean it up so there could be only one way to log from illinker (made few more comments where it still needs some investigation). The tracking issue for remaining work is #1022

src/linker/Linker/MessageContainer.cs
```diff
@@ -0,0 +1,105 @@
...
+					break;
+
+				// Warnings
+				case int _code when (_code > 2000 && _code <= 6000):
+					if (category != MessageCategory.Warning)
+						throw new ArgumentException ($"MSBuild Message Container with code ${code} was expected to be of 'Warning' category.");
+					break;
+
+				// Info.
+				case int _code when (_code > 6000 && _code <= 8000):
```
marek-safar: Should we make code optional for info and diagnostic messages? It looks like they are never displayed on UI so the code is not really useful.

src/linker/Linker/MessageContainer.cs
```diff
@@ -0,0 +1,105 @@
...
+			string cat;
+			switch (Category) {
+			case MessageCategory.Error:
+				cat = "error";
+				break;
+			case MessageCategory.Warning:
+				cat = "warning";
+				break;
+			default:
+				cat = "";
```
marek-safar: My understanding is that only error or warning are recognized. Should we use different text output pattern for info and diagnostics?

mateoatr: Yes, MSBuild only recognizes (and prints) warnings and errors through the format that we're using. I added the `info` category, so that we could still log text to console even if there isn't an error or warning on sight. I don't know how useful this would be in general, if it doesn't bring any value or there isn't any scenario where this is needed, I think I should simply remove it.

src/linker/Linker/MessageContainer.cs
mateoatr: Yes, there's no need for the code parameter for this category (If we are to keep the `info` category).

marek-safar: Please change Code property to be nullable then. I am also thinking of having factory methods instead of this constructor if only some parameters are required. Something like

```c#
static MessageContainer CreateErrorMessage (string text, int code, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null) { .. }

static MessageContainer CreateWarningMessage (string text, int code, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null) { .. }

static MessageContainer CreateInfoMessage (string text) { .. }

static MessageContainer CreateDiagnostingMessage (string text) { .. }
```

src/linker/Linker/MessageContainer.cs
marek-safar: This makes sense as I'd prefer to see this used for any linker logging. To make this effect we need to hook up the filtering to MSBuild properties. Made issue #1030 about that.

src/linker/Linker/MessageContainer.cs
```diff
@@ -0,0 +1,113 @@
...
+				sb.Append (" ")
+					.Append (cat)
+					.Append (" IL")
+					.Append (Code.Value.ToString ("D4"));
+
+				if (!string.IsNullOrEmpty (Text))
+					sb.Append (": ").Append (Text);
+			}
+			else {
+				sb.Append (" " + Text);
```
marek-safar:
```suggestion
				sb.Append (" " ).Append (Text);
```

Don't do string concatenation when you have StringBuilder

PR #974, opened 2020-02-28
Explicitly enable optimization during test

mrvoorhe:
Explicitly enable `IPConstantPropagation` during the `UnreachableBlock` tests.  This gives implementors of a linker executable the freedom to pick which optimizations are enabled by default while also ensuring that they can pass all tests.

This is consistent with other optimizations such as unreachable bodies, used attrs only, etc.


marek-safar:
@mrvoorhe test/Mono.Linker.Tests.Cases/UnreachableBlock/TryFinallyBlocks.cs is missing a using

mrvoorhe:
Fixed.

PR #976, opened 2020-02-28
Resolve custom steps from external assemblies

### LINKED ISSUE #969: Custom steps cannot be resolved from external assembly in ILLink
Currently, whenever a user tries to add a custom step to the linkers pipeline, the Driver will [try to retrieve the type](https://github.com/mono/linker/blob/b577649961239131e21288aa6b826a1e2bd92674/src/linker/Linker/Driver.cs#L646) of the custom step that was passed to the linker using the `--custom-step` flag, and fail, since there's no way for the linker to know in which assembly this type is. We should define the way in which the linker should find and resolve custom steps that are passed to it through the command line.

mateoatr:
Fixes #969: These changes add the ability to specify the assembly where a custom step is defined by either specifying its full path or its file name, which we'll be looked for into the additional directories passed to the linker through the command line (with `-d`).

### DIFF
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -816,24 +816,6 @@ namespace Mono.Linker {
 			return false;
 		}
 
-		bool GetCustomStepParams (string token, Action<string> action)
-		{
-			if (arguments.Count < 1) {
-				ErrorMissingArgument (token);
-				return false;
-			}
-
-			while (arguments.Count > 0 && !string.IsNullOrEmpty (arguments.Peek())) {
-				var step_or_assembly = arguments.Peek ();
-				if (step_or_assembly.StartsWith ("-") && !step_or_assembly.EndsWith (".dll"))
-					break;
-				action (step_or_assembly);
-				arguments.Dequeue ();
-			}
-
-			return true;
-		}
-
 		string GetNextStringValue ()
 		{
 			if (arguments.Count < 1)
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -589,27 +589,25 @@ namespace Mono.Linker {
 
 		partial void PreProcessPipeline (Pipeline pipeline);
 
-		protected static Assembly GetCustomAssembly (string arg) {
-			Assembly custom_assembly = null;
+		private static Assembly GetCustomAssembly (string arg) {
 			if (Path.IsPathRooted (arg)) {
 				var assemblyPath = Path.GetFullPath (arg);
 				if (File.Exists (assemblyPath))
 					return Assembly.Load (File.ReadAllBytes (assemblyPath));
-				else
-					Console.WriteLine ($"The assembly '{arg}' specified for '--custom-step' option could not be found");
+				Console.WriteLine ($"The assembly '{arg}' specified for '--custom-step' option could not be found");
 			}
 			else
-				Console.WriteLine ($"The path to the assembly '{arg}' specified for '--custom-step' must be rooted");
+				Console.WriteLine ($"The path to the assembly '{arg}' specified for '--custom-step' must be fully qualified");
 
-			return custom_assembly;
+			return null;
 		}
 
 		protected static bool AddCustomStep (Pipeline pipeline, string arg)
 		{
 			Assembly custom_assembly = null;
-			int pos = arg.IndexOf (",,");
+			int pos = arg.IndexOf (",");
 			if (pos != -1) {
-				custom_assembly = GetCustomAssembly (arg.Substring (pos + 2));
+				custom_assembly = GetCustomAssembly (arg.Substring (pos + 1));
 				if (custom_assembly == null)
 					return false;
 				arg = arg.Substring (0, pos);
@@ -670,10 +668,7 @@ namespace Mono.Linker {
 
 		static IStep ResolveStep (string type, Assembly assembly)
 		{
-			Type step = Type.GetType (type, false);
-			if (step == null) {
-				step = assembly.GetType (type);
-			}
+			Type step = assembly != null ? assembly.GetType(type) : Type.GetType (type, false);
 
 			if (step == null) {
 				Console.WriteLine ($"Custom step '{type}' could not be found");
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -866,9 +866,9 @@ namespace Mono.Linker {
 			Console.WriteLine ("Advanced");
 			Console.WriteLine ("  --custom-step CFG         Add a custom step <config> to the existing pipeline");
 			Console.WriteLine ("                            Step can use one of following configurations");
-			Console.WriteLine ("                            TYPE,,PATH_TO_ASSEMBLY: Add user defined type as last step to the pipeline");
-			Console.WriteLine ("                            +NAME:TYPE,,PATH_TO_ASSEMBLY: Inserts step type before existing step with name");
-			Console.WriteLine ("                            -NAME:TYPE,,PATH_TO_ASSEMBLY: Add step type after existing step");
+			Console.WriteLine ("                            TYPE,PATH_TO_ASSEMBLY: Add user defined type as last step to the pipeline");
+			Console.WriteLine ("                            +NAME:TYPE,PATH_TO_ASSEMBLY: Inserts step type before existing step with name");
+			Console.WriteLine ("                            -NAME:TYPE,PATH_TO_ASSEMBLY: Add step type after existing step");
 			Console.WriteLine ("  --ignore-descriptors      Skips reading embedded descriptors (short -z). Defaults to false");
 			Console.WriteLine ("  --keep-facades            Keep assemblies with type-forwarders (short -t). Defaults to false");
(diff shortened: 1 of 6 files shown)


src/linker/Linker/Driver.cs
```diff
@@ -568,10 +570,16 @@ public bool Run (ILogger customLogger = null)
...
-				foreach (string custom_step in custom_steps) {
-					if (!AddCustomStep (p, custom_step))
+				
+				if (custom_steps.Any()) {
+					List<Assembly> custom_step_assemblies;
+					if (!ResolveCustomStepAssemblies (custom_steps, context.Resolver.GetSearchDirectories (), out custom_step_assemblies))
 						return false;
+
+					while (custom_steps.Any()) {
+						if (!AddCustomStep (p, custom_steps.Pop (), custom_step_assemblies))
```
vitek-karas: I think this will change the order in which the steps are added - which may change the effective order of steps - which is an unnecessary breaking change.

mateoatr: You're right, thanks for spotting this. I'll change it back.

mateoatr:
I think that adding a new command line option here would be a better option, being more explicit in what we're doing. I'll make the changes and then add some examples in here and update the [current documentation](https://github.com/mono/linker/blob/master/src/linker/README.md) we have about adding custom steps.

src/linker/Linker/Driver.cs
```diff
@@ -596,39 +604,25 @@ public bool Run (ILogger customLogger = null)
...
+			Assembly custom_assembly = null;
+			if (Path.IsPathRooted (arg)) {
+				var assemblyPath = Path.GetFullPath (arg);
+				if (File.Exists (assemblyPath))
+					return Assembly.LoadFrom (assemblyPath);
+				else
+					Console.WriteLine ($"Invalid assembly path '{arg}' specified for '--custom-step' option");
+			}
+			else {
+				foreach (var directory in search_directories) {
```
vitek-karas: I'm a bit nervous about this - mostly because we're combining two worlds of assemblies here. The `search_directories` are intended for resolution of assemblies for the app being linked. Now we're also using it to search for parts of the linker (the tool). This is problematic for several reasons:
* If there are potential ambiguous matches (the app and the linker using assembly with the same name) - not very likely, but still
* The linked app may not be of the same .NET Core version nor it may be the same target - for example I can link a linux app on windows. This means I'm searching through potentially linux specific directories when looking for windows assemblies.
* This also works the other way round - if I add a search directory specifically due to custom steps now I'm opening a potential issue that assemblies from that directly will be treated as part of the app.

For now I would prefer to only support the fully specified path and don't try to implement probing in any way. For the scenario we want this should not be a problem. It doesn't match the existing functionality on .NET Framework/Mono, but I don't think that's the goal we should be after.

src/linker/Linker/Driver.cs
```diff
@@ -276,6 +278,12 @@ public bool Run (ILogger customLogger = null)
 
 							continue;
 
+						case "--custom-assembly":
```
marek-safar: Please instead of introducing option which makes sense only when used with `--custom-step` please extend `--custom-step` to take optional assembly path value

mateoatr: I've changed the way we get the custom assemblies to the other option that @vitek-karas  originally suggested (use `,,` as a separator between the custom step and the custom assembly).

marek-safar: Why `,,` ? I think a single comma should be fine. If you are concern about the name being part of type we could force some escaping in that case but don't think that will be needed at this point

test/Mono.Linker.Tests.Cases/Mono.Linker.Tests.Cases.csproj
```diff
@@ -114,6 +114,7 @@
     <Content Include="LinkXml\UnusedTypeDeclarationPreservedByLinkXmlIsKept.xml" />
   </ItemGroup>
   <ItemGroup>
+    <ProjectReference Include="..\..\src\linker\Mono.Linker.csproj" />
```
marek-safar: This does not make much sense. You need to add the reference to https://github.com/mono/linker/pull/976/files#diff-38e7184cf84a72cb81b88450ae657f74R7 not here

src/linker/Linker/Driver.cs
```diff
@@ -643,9 +668,13 @@ static IStep FindStep (Pipeline pipeline, string name)
 			return null;
 		}
 
-		static IStep ResolveStep (string type)
+		static IStep ResolveStep (string type, Assembly assembly)
 		{
 			Type step = Type.GetType (type, false);
+			if (step == null) {
```
marek-safar: The logic has to be inverted, if assembly != null it has to loop up the type in the assembly first

src/linker/Linker/Driver.cs
```diff
@@ -588,11 +589,35 @@ public bool Run (ILogger customLogger = null)
 
 		partial void PreProcessPipeline (Pipeline pipeline);
 
+		protected static Assembly GetCustomAssembly (string arg) {
```
marek-safar: Why protected?

src/linker/Linker/Driver.cs
```diff
@@ -588,11 +589,35 @@ public bool Run (ILogger customLogger = null)
...
+			Assembly custom_assembly = null;
```
marek-safar: This local variable is redundant

src/linker/Linker/Driver.cs
```diff
@@ -588,11 +589,35 @@ public bool Run (ILogger customLogger = null)
...
+			if (Path.IsPathRooted (arg)) {
+				var assemblyPath = Path.GetFullPath (arg);
+				if (File.Exists (assemblyPath))
+					return Assembly.Load (File.ReadAllBytes (assemblyPath));
+				else
+					Console.WriteLine ($"The assembly '{arg}' specified for '--custom-step' option could not be found");
+			}
+			else
+				Console.WriteLine ($"The path to the assembly '{arg}' specified for '--custom-step' must be rooted");
```
marek-safar: `rooted` is not what most developers understand, please use a common word

src/linker/Linker/Driver.cs
```diff
@@ -588,11 +589,35 @@ public bool Run (ILogger customLogger = null)
 
 		partial void PreProcessPipeline (Pipeline pipeline);
 
+		protected static Assembly GetCustomAssembly (string arg) {
+			Assembly custom_assembly = null;
+			if (Path.IsPathRooted (arg)) {
+				var assemblyPath = Path.GetFullPath (arg);
+				if (File.Exists (assemblyPath))
+					return Assembly.Load (File.ReadAllBytes (assemblyPath));
+				else
```
marek-safar: Redundant else

src/linker/Linker/Driver.cs
```diff
@@ -588,11 +589,35 @@ public bool Run (ILogger customLogger = null)
...
+					Console.WriteLine ($"The assembly '{arg}' specified for '--custom-step' option could not be found");
+			}
```
marek-safar: Please follow existing coding style

src/linker/Linker/Driver.cs
vitek-karas: Basically the only counter-example is something like `MyType,MyAssembly.dll` where on disk there's a `MyAssembly.dll.dll` - so the `.dll` is part of the assembly name and not a file extension.  If we're OK not supporting that case, I'd be perfectly fine with looking at the string after the first comma and if it has a `.dll` or `.exe` extension, then treat it as a file path. To make it even better we could say that in that case there must be just one comma - multiple commas would automatically mean to treat it as an assembly qualified type name.

marek-safar: I think we should treat anything after the comma as a file. I don't think allowing assembly-qualified name together with the full path to the file would help anyone. Using the assembly-qualified name without file path would be even stranger and in this context.

vitek-karas: I was trying to address the fact that linker today supports assembly qualified type names in this context already - so I'm trying to not break that functionality.

marek-safar: I don't think that's ever used (and we could do illink support only too)

vitek-karas: OK - I don't know how important is to keep this for mono linker - so I would vote for changing it everywhere...

test/Mono.Linker.Tests.Cases/CommandLine/AddCustomStep.cs
```diff
@@ -1,14 +1,18 @@
...
 
 namespace Mono.Linker.Tests.Cases.CommandLine
 {
-	[SetupCompileBefore ("CustomStep.dll", new [] { "Dependencies/CustomStepDummy.cs" })]
-	[SetupLinkerArgument ("--custom-step", "CustomStep.CustomStepDummy,,CustomStep.dll")]
-	[SetupLinkerArgument ("--custom-step", "-CleanStep:CustomStep.CustomStepDummy,,CustomStep.dll")]
-	[SetupLinkerArgument ("--custom-step", "+CleanStep:CustomStep.CustomStepDummy,,CustomStep.dll")]
-	[LogContains("Custom step added")]
+
+#if NET471
```
marek-safar:
```suggestion
#if !NETCOREAPP
```

PR #977, opened 2020-02-28
Reference linker via assembly path in tests

sbomer:
This takes advantage of the SDK hook in 3.x which allows overriding
the tasks package via ILLinkTasksPath. This simplifies the integration
tests, which will no longer require a "dotnet pack" before running.

The old targets file, which was used for the packagereference hook, is
being removed because it was fragile and hasn't been kept up-to-date
with SDK changes.

The portable publish scenario is no longer tested since it's not
supported.

This should unblock the arcade SDK update in https://github.com/mono/linker/pull/954.


src/ILLink.Tasks/LinkTask.cs
```diff
@@ -110,7 +110,7 @@ private string DotNetPath
 					return _illinkPath;
 
 				var taskDirectory = Path.GetDirectoryName (Assembly.GetExecutingAssembly ().Location);
-                                // The linker always runs on .NET Core, even when using desktop MSBuild to host ILLink.Tasks.
+				// The linker always runs on .NET Core, even when using desktop MSBuild to host ILLink.Tasks.
```
marek-safar: Should this be reflected in https://github.com/mono/linker/blob/e1d3c6ab33542464e00779ede541a306822a9fb9/src/ILLink.Tasks/ILLink.Tasks.csproj#L3 as well?

test/ILLink.Tasks.IntegrationTests/IntegrationTestBase.cs
```diff
@@ -58,23 +55,21 @@ protected void AddLinkerReference(string csproj)
...
-					new XElement(ns + "PackageReference",
-						new XAttribute("Include", TestContext.TasksPackageName),
-						new XAttribute("Version", TestContext.TasksPackageVersion))));
+
+			foreach (var el in xdoc.Root.Elements(ns + "PropertyGroup")) {
+				el.Add(new XElement(ns + "ILLinkTasksAssembly", TestContext.TasksAssemblyPath));
+				break;
 			}
 
+			// Workaround to avoid passing an argument that was removed (-l).
```
marek-safar: Could you please create issue about this work item

sbomer: Created https://github.com/mono/linker/issues/979.

src/ILLink.Tasks/LinkTask.cs
sbomer: No - while illink.dll runs only on .NET Core, the task that shells out to it runs on desktop or .NET Core.

marek-safar: I am still not sure how do you get transfer execution from .net desktop dll to .net5 dll but nevermind.

PR #994, opened 2020-03-12
[feature/dataflow] Add dataflow analysis prototype

MichalStrehovsky:
This can do enough dataflow analysis so that this code "just works" without any annotations (real-world pattern in e.g. System.Text.Json):

```csharp
Type t;
if (isValueType)
    t = typeof(ValueTypeAccessor<>).MakeGenericType(p);
else
    t = typeof(ReferenceTypeAccessor<>).MakeGenericType(p);

// Linker keeps the default constructor for both ValueTypeAccessor and
// ReferenceTypeAccessor.
Activator.CreateInstance(t);
```

On a very high level:
* we keep track of interesting value on the evaluation stack (interesting being constant integers, allocated array sizes, System.Type instances, RuntimeTypeHandles, string constants, parameter loads, field loads, and the likes).
* uninteresting values are simply tracked as Unknown.
* we make a single pass over the IL, modeling the value on the stack.
* in case of merge points, we model the stack slots while preserving all possible values (e.g. in the example above, there's a merge point after the `else` branch, at which point the value stored in `t` can either be ValueTypeAccessor or ReferenceTypeAccessor).
* Loading a multi-value location on the evaluation stack produces a stack slot representing multiple values. One can access the values by doing a foreach over the `UniqueValues` property of the `ValueNode`.

It's easiest to start reviewing this by looking at the diff in MarkStep.cs, which really shows the consumption part of the analysis.

ValueNode.cs and MethodBodyScanner.cs are Cecil rewrites of same-named files in the closed-source .NET Native Dependency Reducer, with _a lot_ of stuff removed. I also got rid of most CCI-sms. I kept the structure the same because we can raid the .NET Native Dependency Reducer source code for more patterns if we need to later (e.g. the .NET Native Dependency Reducer can handle dataflow across `async` state machines generated by the C# compiler for async methods - but I kind of hope we won't need those kinds of things, ever).

Unfortunately, diffing with the .NET Native codebase is impaired by non-standard code formatting rules in the Linker repo (that the new files adhere to for consistency reasons).

ValueNode.cs has barely any references to Cecil - the hope is that we would be able to reuse this file for e.g. a Roslyn analyzer, but I haven't investigated Roslyn's dataflow analysis facilities. I assume there are some for the nullable analysis, but I didn't look whether they're public or we need to roll our own.


src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Mono.Collections.Generic;
+
+namespace Mono.Linker.Dataflow
+{
+	/// <summary>
+	/// Tracks information about the contents of a stack slot
+	/// </summary>
+	class StackSlot
```
MichalStrehovsky: Kept this as `class` for now because that's what it is in .NET Native, but we should be able to make this a struct, I think.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2366,6 +2367,9 @@ protected virtual void MarkReflectionLikeDependencies (MethodBody body)
 			if (HasManuallyTrackedDependency (body))
 				return;
 
+			var scanner = new ReflectionMethodBodyScanner (this);
```
MichalStrehovsky: Eventually, this should only run when we detect stores into unsafe locations (unsafe method callsite/unsafe method return value/unsafe field) as part of the method body marking.

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -0,0 +1,1258 @@
...
+	/// best-effort representation.  Additionally, as the analysis is linear and does not account for control flow, any
+	/// given ValueNode may represent multiple values simultaneously.  (This occurs, for example, at control flow join
+	/// points when both paths yield values on the IL stack or in a local.)
+	/// </summary>
+	public abstract class ValueNode : IEquatable<ValueNode>
+	{
+		public ValueNode ()
+		{
+#if false // Helpful for debugging a cycle that has inadvertently crept into the graph
+            if (this.DetectCycle(new HashSet<ValueNode>()))
```
vitek-karas: Nit: Weird indentation

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -0,0 +1,1258 @@
...
+			//InvalidateIsOpen ();
+
+			if (node.Kind == ValueNodeKind.MergePoint) {
+				foreach (ValueNode value in ((MergePointValue)node).Values)
+					m_values.Add (value);
+			} else
+				m_values.Add (node);
+
+#if false
+            if (this.DetectCycle(new HashSet<ValueNode>()))
```
vitek-karas: Nit: indentation

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3346,6 +3358,176 @@ public AttributeProviderPair (CustomAttribute attribute, ICustomAttributeProvide
...
+				if (methodReturnValue == null) {
+					if (calledMethod.ReturnType.MetadataType != MetadataType.Void) {
+						methodReturnValue = UnknownValue.Instance;
+					}
+				}
+
+				return true;
+			}
+
+			void MarkMethodsFromReflectionCall (ref ReflectionPatternContext reflectionContext, TypeDefinition declaringType, string name, BindingFlags? bindingFlags, int? parametersCount = null)
```
MichalStrehovsky: This is copypaste of the existing method in this file. I removed the generic arity parameter because generic methods don't have mangled names. I don't think that code path is ever reachable. I expect when we're done, the existing method will be deleted in favor of this scanner, so I didn't bother unifying.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3346,6 +3358,176 @@ public AttributeProviderPair (CustomAttribute attribute, ICustomAttributeProvide
...
+								}
+								else {
+									// The overload with a single System.Type argument
+									ctorParameterCount = 0;
+									bindingFlags |= BindingFlags.Public;
+								}
+
+								// Go over all types we've seen
+								foreach (var value in methodParams[0].UniqueValues ()) {
+									if (value is SystemTypeValue systemTypeValue) {
```
vitek-karas: We'll probably have to change the reporting here
The context expects to get basically one call per analysis - either recognized or not. Since this is a foreach, it may get multiple. It probably just needs reviewing.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3346,6 +3358,176 @@ public AttributeProviderPair (CustomAttribute attribute, ICustomAttributeProvide
...
+										MarkMethodsFromReflectionCall (ref reflectionContext, systemTypeValue.TypeRepresented, ".ctor", bindingFlags, ctorParameterCount);
+									} else if (value == NullValue.Instance) {
+										// Nothing to report. This is likely just a value on some unreachable branch.
+										reflectionContext.RecordHandledPattern ();
```
vitek-karas: Why do we need a new method - why not use the existing RecordRecognizedPattern?

MichalStrehovsky:
I'm trying to figure out what's the Linux failure. It's triggering the invalid IL detection that currently throws for diagnostic purposes (it will eventually be a nop):

https://github.com/mono/linker/pull/994/files#diff-83d9d9aa53da0b0737513a2561f82bc8R3373-R3375

The stack trace of the failure looks impossible (the source for the test has a call to a static method with no parameters, but at the time of dataflow analysis, Cecil thinks the called method takes some parameters that weren't supplied).

I'm setting up a Linux environment with mcs, Mono ILASM, etc. so that I can repro.

src/linker/Linker.Steps/MarkStep.cs
mrvoorhe: Would you mind giving us some way to turn this new code off without us having to customize our fork?

One way would be refactoring `MarkReflectionLikeDependencies` a bit so that we can override the new behavior while leaving the existing reflection detection running.

```
		protected virtual void MarkReflectionLikeDependencies (MethodBody body)
		{
			if (HasManuallyTrackedDependency (body))
				return;

                       MarkReflectionWithScanner(body);
                       MarkReflectionWithPatternDetector(body);
		}
        
        protected virtual void MarkReflectionWithScanner(MethodBody body)
        {
             var scanner = new ReflectionMethodBodyScanner (this);
             scanner.Scan (body);
        }
        
        protected virtual void MarkReflectionWithPatternDetector(MethodBody body)
        {
        	       var instructions = body.Instructions;
			ReflectionPatternDetector detector = new ReflectionPatternDetector (this, body.Method);

			//
			// Starting at 1 because all patterns require at least 1 instruction backward lookup
			//
			for (var i = 1; i < instructions.Count; i++) {
				var instruction = instructions [i];

				if (instruction.OpCode != OpCodes.Call && instruction.OpCode != OpCodes.Callvirt)
					continue;

				if (ProcessReflectionDependency (body, instruction))
					continue;

				if (!(instruction.Operand is MethodReference methodCalled))
					continue;

				var methodCalledDefinition = methodCalled.Resolve ();
				if (methodCalledDefinition == null)
					continue;

				ReflectionPatternContext reflectionContext = new ReflectionPatternContext (_context, body.Method, methodCalledDefinition, i);
				try {
					detector.Process (ref reflectionContext);
				}
				finally {
					reflectionContext.Dispose ();
				}
			}
        }
```

Alternatively, some sort of bool option on `LinkContext` to disable usage of `ReflectionMethodBodyScanner ` would work as well.

There is a lot of new code here and no tests.  I'm sure it was tested else where, but I don't want this new code to get in the way of us staying in sync with upstream monolinker

MichalStrehovsky: I should have made it more clear - this is going into a branch so that we can shape it into something acceptable for master. Being able to turn this off, making sure it doesn't run unless absolutely necessary, etc. is going to be addressed before this would make it into master.

We're going to add tests too - there are existing targeted tests we can port and we're going to write new tests too.

mrvoorhe: O great that sounds good.

We have more complicated reflection detection tests in our UnityLinker test suite that are supported by stack analysis that we have.  I can put these tests on branch for you to grab if you’d like.

MichalStrehovsky:
> I'm trying to figure out what's the Linux failure

Okay, this is a pre-existing problem. The input to the linker is invalid (look for the `<WTF>` marker - this method is definitely not an instance method):

```
.method public hidebysig static void  Main() cil managed
{
  .entrypoint
  // Code size       7 (0x7)
  .maxstack  8
  IL_0000:  call       <WTF>instance</WTF> object ['library']Mono.Linker.Tests.Cases.UnreachableBlock.Dependencies.ClassA::Method_1()
  IL_0005:  pop
  IL_0006:  ret
} // end of method UninitializedLocals::Main
```

I'm going to assume this is a mcs bug because this test works fine on Windows. Would #996 help? I'll try to retrigger the CI.

MichalStrehovsky:
> I'm going to assume this is a mcs bug because this test works fine on Windows.

Nope, this is a test bug. #997 is the fix.

MichalStrehovsky:
On a second though, it is also an mcs bug because it treats the method as static in the source file (doesn't complain about the signature), but then proceeds to generate the signature as instance. Do we care about mcs bugs, or not anymore?

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+					case Code.Ldc_I4_0:
+					case Code.Ldc_I4_1:
+					case Code.Ldc_I4_2:
+					case Code.Ldc_I4_3:
+					case Code.Ldc_I4_4:
+					case Code.Ldc_I4_5:
+					case Code.Ldc_I4_6:
+					case Code.Ldc_I4_7:
+					case Code.Ldc_I4_8: {
+							int value = operation.OpCode.Code - Code.Ldc_I4_0;
```
marek-safar: I am not sure I like this. Cecil can change the values anytime

src/linker/Linker.Dataflow/Helpers.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Mono.Linker.Dataflow
+{
+	static class TypeHelper
+	{
+		public static bool TypesAreEquivalent (TypeDefinition t1, TypeDefinition t2)
+		{
+			return RuntimeHelpers.Equals (t1, t2);
+		}
+
+		public static int GetHashCode (TypeDefinition t) => t.GetHashCode ();
```
marek-safar: What is the value of this?

src/linker/Linker.Dataflow/Helpers.cs
```diff
@@ -0,0 +1,23 @@
+using System.Runtime.CompilerServices;
+
+using Mono.Cecil;
+
+namespace Mono.Linker.Dataflow
+{
+	static class TypeHelper
```
marek-safar: I don't think we need this type. If you really need the one lines add them to the existing extension types

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+			Value = value;
+			IsByRef = isByRef;
+		}
+	}
+
+	abstract partial class MethodBodyScanner
+	{
+		internal ValueNode MethodReturnValue { private set; get; }
+
+		protected virtual void WarnAboutInvalidILInMethod(MethodBody method, int ilOffset)
```
marek-safar: Please make the formatting consistent

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+
+		private void CheckForInvalidReturnStack (Stack<StackSlot> stack, MethodBody method, int ilOffset)
+		{
+			int numExpectedValuesOnStack = method.Method.ReturnType.MetadataType == MetadataType.Void ? 0 : 1;
+			if (stack.Count != numExpectedValuesOnStack) {
+				WarnAboutInvalidILInMethod (method, ilOffset);
+			}
+		}
+
+		private static void PushUnknown (Stack<StackSlot> stack, int count)
```
marek-safar: All usage uses count == 1, why do you need a loop here?

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+	class StackSlot
+	{
+		public ValueNode Value { get; set; }
+
+		/// <summary>
+		/// True if the value is on the stack as a byref
+		/// </summary>
+		public bool IsByRef { get; set; }
+
+		public StackSlot ()
```
marek-safar: Could this be a private and used as static instance value?

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+		private void StoreMethodLocalValue<KeyType> (
+			Dictionary<KeyType, ValueBasicBlockPair> valueCollection,
+			ValueNode valueToStore,
+			KeyType collectionKey,
+			int curBasicBlock)
+		{
+			ValueBasicBlockPair newValue = new ValueBasicBlockPair { BasicBlockIndex = curBasicBlock };
+
+			ValueBasicBlockPair existingValue;
+			valueCollection.TryGetValue (collectionKey, out existingValue);
```
marek-safar: Would be faster if you checked for return value

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+					case Code.Stloc:
+					case Code.Stloc_S:
+					case Code.Stloc_0:
+					case Code.Stloc_1:
+					case Code.Stloc_2:
+					case Code.Stloc_3:
+						ScanStloc (operation, currentStack, methodBody, locals, curBasicBlock);
+						break;
+
+					case Code.Break:
```
marek-safar: These should not be here if they do nothing

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+			}
+		}
+
+		private void ScanLdtoken (
+			Instruction operation,
+			Stack<StackSlot> currentStack,
+			MethodDefinition thisMethod,
+			MethodBody methodBody)
+		{
+			TypeReference typeReference = operation.Operand as TypeReference;
```
marek-safar: Use `is` instead

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+			Code code = operation.OpCode.Code;
+			if (code >= Code.Ldloc_0 && code <= Code.Ldloc_3)
+				return localVariables [code - Code.Ldloc_0];
+			if (code >= Code.Stloc_0 && code <= Code.Stloc_3)
+				return localVariables [code - Code.Stloc_0];
+
+			return (VariableDefinition)operation.Operand;
+		}
+
+		private static bool HasThisParam (MethodReference method)
```
marek-safar: Please move this to extension type

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -0,0 +1,825 @@
...
+
+		public abstract bool HandleCall (
+			MethodBody callingMethodBody,
+			MethodReference calledMethod,
+			Instruction operation,
+			ValueNodeList methodParams,
+			out ValueNode methodReturnValue);
+	}
+
+	static class Extensions
```
marek-safar: Please move it to separate file

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -0,0 +1,1258 @@
...
+				otherEnumerator.MoveNext ();
+				if (!thisEnumerator.Current.Equals (otherEnumerator.Current))
+					return false;
+			}
+			return true;
+		}
+	}
+	#endregion
+
+	static class HashUtils
```
marek-safar: Why is this needed when there is HashCode.Combine ?

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -0,0 +1,1258 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+using TYPE = Mono.Cecil.TypeDefinition;
```
marek-safar: Please don't

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2366,6 +2367,9 @@ protected virtual void MarkReflectionLikeDependencies (MethodBody body)
 			if (HasManuallyTrackedDependency (body))
 				return;
 
+			var scanner = new ReflectionMethodBodyScanner (this);
+			scanner.Scan (body);
```
marek-safar: I'd like to see some numbers as the implementation look very allocation heavy and wasteful

src/linker/Linker.Dataflow/MethodBodyScanner.cs
filipnavara: I don't think that's likely. The enum is directly mirroring the values in the CIL specification.

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: This handles:

```csharp
Type t = null;
if (condition)
    t = typeof(Foo);
else
    t = typeof(Bar);
CreateInstance(t);
```

We're going to see `null`, `Foo`, `Bar` for this. We could also report nothing for `null`, but then this is going to assert/crash on `CreateInstance(null)` (I'm sure there will be user code with this, however meaningless that pattern is).

`RecordRecognizedPattern` expects that there's going to be some accessed item. In this case there's no accessed item.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
MichalStrehovsky: > Cecil can change the values anytime

Changing the value of an enum member is a breaking source and binary change.

I could see Cecil being forced into a breaking change if we ever add new unprefixed IL instructions (Cecil currently assumes all IL instructions fit in a single byte, and folds the ones that have a twobyte encoding into a single byte crossing fingers there's enough space left). But I wouldn't expect these to be ever _reordered_ - complex lookup tables from physical opcode encoding to `Code` would introduce unnecessary inefficiencies when parsing.

src/linker/Linker.Dataflow/ValueNode.cs
MichalStrehovsky: I added this so that we can keep track of how much this file depends on Cecil (if someone ever adds `using Mono.Cecil` at the top, we would see right away).

It's connected to "ValueNode.cs has barely any references to Cecil - the hope is that we would be able to reuse this file for e.g. a Roslyn analyzer" in the pull request description.

If we don't care about reusing this file, I can inline this.

src/linker/Linker.Dataflow/Helpers.cs
MichalStrehovsky: This is also in connection with "ValueNode.cs has barely any references to Cecil - the hope is that we would be able to reuse this file for e.g. a Roslyn analyzer" - if we decide we don't care, I can inline this into ValueNode. It's only used from ValueNode.

src/linker/Linker.Dataflow/Helpers.cs
MichalStrehovsky: `Object.GetHashCode`. We'll probably want to implement something better. We'll find out when we do more detailed perf measurements, but it might not matter since this only affects merging stack slots that model System.Type/RuntimeTypeHandle and I don't expect we would see many.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
MichalStrehovsky: Leftover from .NET Native. Removed the count parameter. We can always put it back once it's needed.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
MichalStrehovsky: We'll get bigger bang for the buck if we make `StackSlot` a `struct`. I'm not clear on why it was a class in .NET Native.

We might want to also explore getting rid of `StackSlot` completely and just use a stack of `ValueNodes`. The `IsByRef` tracking might not be necessary. .NET Native also tracked the static type of the value in the slot but we don't have that here because this analysis doesn't care.

MichalStrehovsky: I'll make an issue with a running list of these things. Resolving them all would be a necessary prerequisite for merging to master.

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: I'll add perf measurements to the running list of branch exit criteria. It doesn't make sense to measure now since we can definitely improve around the `StackSlot`.

Even with the inefficient `StackSlot` class and a lot more tracking (`UnknownValue.Instance` doesn't exist in .NET Native - it actually modelled the the type of the value all the time, even if it wasn't anything interesting because the scanner served other purposes too) in .NET Native, this didn't actually show up as an issue. The traces were dominated by the inefficient way CCI models IL instructions. Cecil has a similarly bad representation. I guess everyone took inspiration from Reflection.Emit.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
marek-safar: Cecil is not API stable so even source breaking changes can happen (so just be careful)

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: We would be very grateful for any tests in this area. Thanks a lot!

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: In that the correct reporting would be "match - Foo" and "match - Bar", ignoring the `null` completely. But if we see only `null` (alone) we should report it is "unrecognized". That's how the rest of the pattern matching behaves in cases where there are potentially multiple matches (like method overrides for `GetMethod` for example).

src/linker/Linker.Dataflow/ValueNode.cs
vitek-karas: I think it's OK to not do this for now - if we get to analyzer and we think it's worth reusing, then we can reintroduce this quite easily.

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: I tried to update the `if` cascade to make it like you suggest (report CreateInstance(null) as unrecognized, even though it's recognized), but things got awkward quickly. I think we'll want to make make adjustments to the reflection context anyway to make it more convenient to use it at merge points (where we have multiple values, some of which might be known).

I added that to the running list of issues to look into.

src/linker/Linker.Dataflow/ValueNode.cs
MichalStrehovsky: HashCode is not available on net472.

marek-safar: There is https://www.nuget.org/packages/Microsoft.Bcl.HashCode for such cases

MichalStrehovsky:
CI is not trigger so I'm going to try closing a reopening...

MichalStrehovsky:
I had to get a local repro to get to the failure:

```
     1>ILLINK : error :  [C:\git\linker\artifacts\bin\ILLink.Tasks.IntegrationTests\Release\netcoreapp3.0\tests-temp\helloworld\helloworld\helloworld.csproj]
           An assembly specified in the application dependencies manifest (illink.deps.json) was not found:
             package: 'Microsoft.Bcl.HashCode', version: '1.1.0'
             path: 'lib/netcoreapp2.1/Microsoft.Bcl.HashCode.dll'
```

I'm tempted to just revert the Microsoft.Bcl.HashCode (commit 88b158e) thing to unblock work in the branch.

@vitek-karas Thoughts?

vitek-karas:
Agreed - revert that and add it to the list (with the commit hash of the reverted change). This should relatively simple fix - the test is probably not deploying all the necessary dependencies correctly.

src/linker/Linker.Steps/MarkStep.cs
mrvoorhe: @vitek-karas here are all of our reflection tests 

https://github.com/Unity-Technologies/linker/commit/1c37396edcb3edeaccf3dc96530c023a8202c892

You can cherry-pick this to a branch from master and everything should run.  Of course many tests will fail.

PR #1001, opened 2020-03-13
Fix pattern matching for Activator.CreateInstance over unrecognized parameter

### LINKED ISSUE #995: Fix Debug.Asserts related to reflection pattern analysis
I can't simply run all of the tests with a debug build on Windows anymore.  I'm hitting debug assert pop ups which I then have to click ignore on in order for the tests to continue running. 

Running the tests under `CoreLinkTests` can reproduce the problem.

There are 2 aspects I'd like to point out.

First, the Debug.Assert pop up is useless.  It just displays a stack.  When I replace the assert with an exception I get this much nicer error message that tells me what method was involved as well as the stack trace.

```
Mono.Linker.MarkException : Error processing method: 'System.Int32 System.AppDomain::ActivateApplication()' in assembly: 'CommonLanguageRuntimeLibrary'
  ----> System.InvalidOperationException : A reflection pattern was analyzed, but no result was reported.
```

And as an added benefit, the unit tests keep running.  Given that the `Debug.Assert` are inside`#if DEBUG` I don't see any benefit to using `Debug.Assert` over throwing.

Secondly, if we are going to have special asserts for debug builds, then CI should build and run tests in debug as well to ensure that we don't have debug only test failures.  I don't know if the tests are currently ran on debug or release builds but it would be worth checking.

cc @vitek-karas

vitek-karas:
On .NET Framework there is an overload of Activator.CreateInstance with takes just one parameter but it's not a Type. This should be reported as unrecognized pattern by the linker.

This change also switches in the internal infra to use exceptions instead of asserts as they play nicer with tests.

Fixes #995

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2443,7 +2443,9 @@ namespace Mono.Linker.Steps {
 				where T : IMemberDefinition
 			{
 #if DEBUG
-				Debug.Assert (_patternAnalysisAttempted, "To correctly report all patterns, when starting to analyze a pattern the AnalyzingPattern must be called first.");
+				if (!_patternAnalysisAttempted)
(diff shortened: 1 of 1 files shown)


mrvoorhe:
nice.  thanks!

PR #1002, opened 2020-03-13
Enable running monolinker tests in VS

vitek-karas:
To run tests from VS the NUnit3TestAdapter package is required, otherwise VS doesn't correctly recognize NUnit tests.


vitek-karas:
@mrvoorhe Can you please try that this change doesn't break Rider? (I don't have it available on any of my machines).

mrvoorhe:
All good.

PR #991, opened 2020-03-11
Don't enable IPConstantPropagation by default

mrvoorhe:
On windows when linking against .NET Framework class libraries every test with `[SetupCoreLinkAction("link")]` fails due to
```
System.NotImplementedException : Initialization of return value in method.....
```

Between this and https://github.com/mono/linker/pull/990 and https://github.com/mono/linker/issues/975 I don't think this option is ready to be turned on by default.

PR #1003, opened 2020-03-13
Add handling for generic value types returned from stack to RemoveUnreachableBlocksStep

marek-safar:
Fixes #991


mrvoorhe:
Thanks!

PR #1004, opened 2020-03-16
Workaround Azure DevOps PR trigger issue

akoeplinger:


eng/azure-pipelines.yml
```diff
@@ -5,6 +5,12 @@ trigger:
     - master
     - release/*
 
+pr:
+  branches:
+    include:
+    - master
+    - release/*
```
marek-safar: Could you add `mono-*` as well?

PR #1010, opened 2020-03-17
Enable CI in the feature/dataflow branch

MichalStrehovsky:
#1004 turned off the CI on all but select branches.


eng/azure-pipelines.yml
```diff
@@ -12,6 +12,7 @@ pr:
     - master
     - release/*
     - mono-*
+    - feature/dataflow
```
marek-safar:
```suggestion
    - feature/*
```

PR #1011, opened 2020-03-18
Add sealer opt-in optimization

marek-safar:


src/linker/Linker.Steps/SealerStep.cs
```diff
@@ -0,0 +1,124 @@
...
+		bool IsSubclassed (TypeDefinition type)
+		{
+			foreach (var a in Context.GetAssemblies ()) {
+				foreach (var s in a.MainModule.Types) {
+					if (s.BaseType?.IsTypeOf (type.Namespace, type.Name) == true)
+						return true;
+
+					if (s.HasNestedTypes) {
+						foreach (var ns in s.NestedTypes) {
+							if (ns.BaseType.IsTypeOf (type.Namespace, type.Name))
```
MichalStrehovsky: `BaseType` could be null if this is a nested interface

src/linker/Linker.Steps/SealerStep.cs
```diff
@@ -0,0 +1,124 @@
...
+			if (!Context.Optimizations.IsEnabled (CodeOptimizations.Sealer, assembly))
+				return;
+
+			foreach (var type in assembly.MainModule.Types)
+				ProcessType (type);
+		}
+
+		bool IsSubclassed (TypeDefinition type)
+		{
+			foreach (var a in Context.GetAssemblies ()) {
```
MichalStrehovsky: This is an [O(n^2) algorithm](https://twitter.com/BruceDawson0xB/status/1120381406700429312). Would it be better to build this map when building the caches for `GetBaseMethods`/`GetOverrides`?

src/linker/Linker.Steps/SealerStep.cs
```diff
@@ -0,0 +1,124 @@
...
+			if (type.IsInterface)
+				return;
+
+			//
+			// the code does not include any subclass for this type
+			//
+			if (!type.IsAbstract && !type.IsSealed && !IsSubclassed (type))
+				type.IsSealed = true;
+
+			if (!type.IsSealed && type.HasNestedTypes) {
```
MichalStrehovsky: Why wouldn't we seal nested types if the owning type is sealed?

src/linker/Linker.Steps/SealerStep.cs
```diff
@@ -0,0 +1,124 @@
...
+
+			foreach (var type in assembly.MainModule.Types)
+				ProcessType (type);
+		}
+
+		bool IsSubclassed (TypeDefinition type)
+		{
+			foreach (var a in Context.GetAssemblies ()) {
+				foreach (var s in a.MainModule.Types) {
+					if (s.BaseType != null && s.BaseType.IsTypeOf (type.Namespace, type.Name) == true)
```
MichalStrehovsky:
```suggestion
					if (s.BaseType != null && s.BaseType.IsTypeOf (type.Namespace, type.Name))
```

mrvoorhe:
I don't think I saw a test for when a base class is sealed because all derived types were unused and thus stripped away.

src/linker/Linker.Steps/SealerStep.cs
```diff
@@ -0,0 +1,124 @@
...
+			// interface members are virtual (and we cannot change this)
+			//
+			if (type.IsInterface)
+				return;
+
+			//
+			// the code does not include any subclass for this type
+			//
+			if (!type.IsAbstract && !type.IsSealed && !IsSubclassed (type))
+				type.IsSealed = true;
```
mrvoorhe: Would you be willing to move all type/method changes into protected virtual methods so that I can override them in our linker?  I like to wire up all changes the linker makes to some sort of diagnostic logging so that we have a way to view the changes the linker makes if we want to.

I'm thinking something like

```
protected virtual void MakeTypedSealed(TypeDefinition type)
{
    type.IsSealed = true;
}

protected virtual void MakeMethodFinal(MethodDefinition method)
{
   method.IsFinal = true;
}
```

src/linker/Linker.Steps/SealerStep.cs
marek-safar: Fixed

src/linker/Linker.Steps/SealerStep.cs
marek-safar: What would you cache? This is called only once for every type

src/linker/Linker.Steps/SealerStep.cs
marek-safar: I guess it does not hurt to seal in this case too

src/linker/Linker.Steps/SealerStep.cs
marek-safar: Done

src/linker/Linker.Steps/SealerStep.cs
MichalStrehovsky: We could e.g. go over all types in all assemblies once, build a HashSet of namespace+name pairs for all types that are base types of something.

Then for each type just query the HashSet instead of iterating over everything again.

The algorithm becomes O(n) on average instead of O(n^2).

mrvoorhe: I recently added the following to our `UnityTypeMapStep` and `UnityAnnotationStore` for a different reason.

https://gist.github.com/mrvoorhe/51c20368469666f665e46c6133b0d984

I think the API could be used to address this comment as well.

marek-safar: Local cache added

src/linker/Linker.Steps/SealerStep.cs
```diff
@@ -0,0 +1,148 @@
...
+using System.Collections.Generic;
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class SealerStep : BaseStep
+	{
+		HashSet<TypeDefinition> referencedBaseTypeCache;
+
+		public SealerStep ()
```
MichalStrehovsky: Nit: the public constructor is redundant.

src/linker/Linker.Steps/SealerStep.cs
```diff
@@ -0,0 +1,148 @@
...
+			if (type.IsInterface)
+				return;
+
+			//
+			// the code does not include any subclass for this type
+			//
+			if (!type.IsAbstract && !type.IsSealed && !IsSubclassed (type))
+				SealType (type);
+
+			if (type.HasNestedTypes) {
```
MichalStrehovsky: Maybe move this all the way up so that we don't miss classes nested under interfaces?

src/linker/Linker.Steps/SealerStep.cs
```diff
@@ -0,0 +1,148 @@
...
+		{
+			if (referencedBaseTypeCache == null) {
+				referencedBaseTypeCache = new HashSet<TypeDefinition> ();
+				foreach (var a in Context.GetAssemblies ()) {
+					foreach (var s in a.MainModule.Types) {
+						var btd = s.BaseType?.Resolve ();
+						if (btd != null)
+							referencedBaseTypeCache.Add (btd);
+
+						if (s.HasNestedTypes) {
```
joncham, line 46: It seems like this will not find types nested beyond the top level. This could be made recursive, or the `a.MainModule.Types` call could be changed to `a.MainModule.GetAllTypes()` extension method.

PR #1012, opened 2020-03-18
Add tracing reasons

sbomer:
This supplies a "reason" every time that we mark a piece of IL.

A "reason" is an instance of a new type, DependencyInfo, which tracks
a single source object and a flag (DependencyKind) indicating what
kind of dependency this is. Annotations.Mark now expects a
DependencyInfo, and the dependency reporting interface is extended to
report these reasons.

The mark sites within MarkStep and other steps have been updated with
specific reasons, and the corresponding Mark* methods have been
modified to take a reason that is passed through as many layers of the
marking logic as necessary until it reaches a call to
Annotations.Mark. The marking logic has been modified to more
frequently allow marking Cecil objects multiple times for different
reasons - while still ensuring that any expensive operations are done
only once per object (usually under an IsProcessed check, rather than
an IsMarked check).

In a few cases, the marking logic will process a Cecil object and mark
its dependencies, without marking the original Cecil object. This can
happen, for example, for custom attributes in some cases, or for
generic instantiations (since we only mark definitions). In these
cases, there are extra calls that report these intermediate
dependencies to the tracing interface without actually marking them.

Some of the DependencyKinds (for example AlreadyMarked) are only used
for internal tracking, and will not be reported to the dependency
tracing interface.

The set of DependencyKinds has been selected based on what seems like
an intuitive reason to report for each object, though sometimes the
choice is not obvious since the reasons are restricted to supplying a
single source object. Some interesting cases where potentially
non-obvious choices were made follow:

- When marking overrides on instantiated type, we choose as the reason
the instantiated type, rather than the base method (unless the
override removal optimization is disabled, in which case
instantiations are not taken into account when marking overrides - so
we blame the base method).

- When marking static constructors, we keep track of cases where they
are marked due to a field access or method call from another method,
and report these separately. For cases where we track enough
information to do so, this results in a more direct dependency chain
like: "method A triggers static constructor B", rather than "method A
access field B, field B has declaring type C, declaring type C has
static constructor D".

- When marking custom attributes on assemblies or modules, we record a
dependency from the assembly/module to the attribute, even though the
assembly/module may not have been marked or even recorded. Since the
attributes and their dependencies are kept regardless of what happens
to the assembly, these are conceptually entry points to the analysis.

This also includes a change that moves the reflection reporter from
LinkContext to Tracer, and extended to support multiple sinks, similar
to the dependency reporting.

I have also added assertions in a few places where I think it makes
sense to check invariants when we are running tests in Debug mode.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -59,6 +59,7 @@ namespace Mono.Linker.Steps {
 		static DependencyKind [] _fieldReasons = new DependencyKind [] {
 			DependencyKind.AccessedViaReflection,
 			DependencyKind.AlreadyMarked,
+			DependencyKind.Custom,
 			DependencyKind.CustomAttributeField,
 			DependencyKind.EventSourceProviderField,
 			DependencyKind.FieldAccess,
@@ -77,6 +78,7 @@ namespace Mono.Linker.Steps {
 			DependencyKind.AttributeType,
 			DependencyKind.BaseType,
 			DependencyKind.CatchType,
+			DependencyKind.Custom,
 			DependencyKind.CustomAttributeArgumentType,
 			DependencyKind.CustomAttributeArgumentValue,
 			DependencyKind.DeclaringType,
@@ -105,6 +107,7 @@ namespace Mono.Linker.Steps {
 			DependencyKind.BaseMethod,
 			DependencyKind.CctorForType,
 			DependencyKind.CctorForField,
+			DependencyKind.Custom,
 			DependencyKind.DefaultCtorForNewConstrainedGenericArgument,
 			DependencyKind.DirectCall,
 			DependencyKind.ElementMethod,
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -49,6 +49,92 @@ namespace Mono.Linker.Steps {
 		protected List<TypeDefinition> _typesWithInterfaces;
 		protected List<MethodBody> _unreachableBodies;
 
+#if DEBUG
+		static DependencyKind [] _entireTypeReasons = new DependencyKind [] {
+			DependencyKind.NestedType,
+			DependencyKind.PreservedDependency,
+			DependencyKind.TypeInAssembly,
+		};
+
+		static DependencyKind [] _fieldReasons = new DependencyKind [] {
+			DependencyKind.AccessedViaReflection,
+			DependencyKind.AlreadyMarked,
+			DependencyKind.CustomAttributeField,
+			DependencyKind.EventSourceProviderField,
+			DependencyKind.FieldAccess,
+			DependencyKind.FieldOnGenericInstance,
+			DependencyKind.InteropMethodDependency,
+			DependencyKind.Ldtoken,
+			DependencyKind.MemberOfType,
+			DependencyKind.PreservedDependency,
+			DependencyKind.ReferencedBySpecialAttribute,
+			DependencyKind.TypePreserve,
+		};
+
+		static DependencyKind [] _typeReasons = new DependencyKind [] {
+			DependencyKind.AccessedViaReflection,
+			DependencyKind.AlreadyMarked,
+			DependencyKind.AttributeType,
+			DependencyKind.BaseType,
+			DependencyKind.CatchType,
+			DependencyKind.CustomAttributeArgumentType,
+			DependencyKind.CustomAttributeArgumentValue,
+			DependencyKind.DeclaringType,
+			DependencyKind.DeclaringTypeOfCalledMethod,
+			DependencyKind.ElementType,
+			DependencyKind.FieldType,
+			DependencyKind.GenericArgumentType,
+			DependencyKind.GenericParameterConstraintType,
+			DependencyKind.InterfaceImplementationInterfaceType,
+			DependencyKind.Ldtoken,
+			DependencyKind.ModifierType,
+			DependencyKind.InstructionTypeRef,
+			DependencyKind.ParameterType,
+			DependencyKind.ReferencedBySpecialAttribute,
+			DependencyKind.ReturnType,
+			DependencyKind.UnreachableBodyRequirement,
+			DependencyKind.VariableType,
+		};
+
+		static DependencyKind [] _methodReasons = new DependencyKind [] {
+			DependencyKind.AccessedViaReflection,
+			DependencyKind.AlreadyMarked,
+			DependencyKind.AttributeConstructor,
+			DependencyKind.AttributeProperty,
+			DependencyKind.BaseDefaultCtorForStubbedMethod,
+			DependencyKind.BaseMethod,
+			DependencyKind.CctorForType,
+			DependencyKind.CctorForField,
+			DependencyKind.DefaultCtorForNewConstrainedGenericArgument,
+			DependencyKind.DirectCall,
+			DependencyKind.ElementMethod,
+			DependencyKind.EventMethod,
+			DependencyKind.EventOfEventMethod,
+			DependencyKind.InteropMethodDependency,
+			DependencyKind.KeptForSpecialAttribute,
+			DependencyKind.Ldftn,
+			DependencyKind.Ldtoken,
+			DependencyKind.Ldvirtftn,
+			DependencyKind.MemberOfType,
+			DependencyKind.MethodForInstantiatedType,
+			DependencyKind.MethodForSpecialType,
+			DependencyKind.MethodImplOverride,
+			DependencyKind.MethodOnGenericInstance,
+			DependencyKind.Newobj,
+			DependencyKind.Override,
+			DependencyKind.OverrideOnInstantiatedType,
+			DependencyKind.PreservedDependency,
+			DependencyKind.ReferencedBySpecialAttribute,
+			DependencyKind.SerializationMethodForType,
+			DependencyKind.TriggersCctorForCalledMethod,
+			DependencyKind.TriggersCctorThroughFieldAccess,
+			DependencyKind.TypePreserve,
+			DependencyKind.UnreachableBodyRequirement,
+			DependencyKind.VirtualCall,
+			DependencyKind.VirtualNeededDueToPreservedScope,
+		};
+#endif
+
 		public MarkStep ()
 		{
 			_methods = new Queue<(MethodDefinition, DependencyInfo)> ();
@@ -165,13 +251,8 @@ namespace Mono.Linker.Steps {
 		void MarkEntireType (TypeDefinition type, DependencyInfo reason)
 		{
 #if DEBUG
-			var reasons = new DependencyKind [] {
-				DependencyKind.NestedType,
-				DependencyKind.PreservedDependency,
-				DependencyKind.TypeInAssembly,
-			};
-			if (!reasons.Contains (reason.Kind))
-				throw new InvalidOperationException ($"Internal error: unsupported type dependency {reason.Kind}");
+			if (!_entireTypeReasons.Contains (reason.Kind))
+				throw new ArgumentOutOfRangeException ($"Internal error: unsupported type dependency {reason.Kind}");
 #endif
 
 			if (type.HasNestedTypes) {
@@ -1048,22 +1129,8 @@ namespace Mono.Linker.Steps {
 		void MarkField (FieldDefinition field, DependencyInfo reason)
 		{
 #if DEBUG
-			var reasons = new DependencyKind [] {
-				DependencyKind.AccessedViaReflection,
-				DependencyKind.AlreadyMarked,
-				DependencyKind.CustomAttributeField,
-				DependencyKind.EventSourceProviderField,
-				DependencyKind.FieldAccess,
-				DependencyKind.FieldOnGenericInstance,
-				DependencyKind.InteropMethodDependency,
-				DependencyKind.Ldtoken,
-				DependencyKind.MemberOfType,
-				DependencyKind.PreservedDependency,
-				DependencyKind.ReferencedBySpecialAttribute,
-				DependencyKind.TypePreserve,
-			};
-			if (!reasons.Contains (reason.Kind))
-				throw new InvalidOperationException ($"Internal error: unsupported field dependency {reason.Kind}");
+			if (!_fieldReasons.Contains (reason.Kind))
+				throw new ArgumentOutOfRangeException ($"Internal error: unsupported field dependency {reason.Kind}");
 #endif
 
 			if (CheckProcessed (field))
@@ -1111,7 +1178,7 @@ namespace Mono.Linker.Steps {
 
 		protected virtual void MarkSerializable (TypeDefinition type)
 		{
-			// TODO: This marks default ctor even for types that aren't [Serializable]. Is this necessary?
+			// Keep default ctor for XmlSerializer support. See https://github.com/mono/linker/issues/957
 			MarkDefaultConstructor (type, new DependencyInfo (DependencyKind.SerializationMethodForType, type));
 			if (!_context.IsFeatureExcluded ("deserialization"))
 				MarkMethodsIf (type.Methods, IsSpecialSerializationConstructor, new DependencyInfo (DependencyKind.SerializationMethodForType, type));
@@ -1120,32 +1187,8 @@ namespace Mono.Linker.Steps {
 		protected virtual TypeDefinition MarkType (TypeReference reference, DependencyInfo reason)
 		{
 #if DEBUG
-			var reasons = new DependencyKind [] {
-				DependencyKind.AccessedViaReflection,
-				DependencyKind.AlreadyMarked,
-				DependencyKind.AttributeType,
-				DependencyKind.BaseType,
-				DependencyKind.CatchType,
-				DependencyKind.CustomAttributeArgumentType,
-				DependencyKind.CustomAttributeArgumentValue,
-				DependencyKind.DeclaringType,
-				DependencyKind.DeclaringTypeOfCalledMethod,
-				DependencyKind.ElementType,
-				DependencyKind.FieldType,
-				DependencyKind.GenericArgumentType,
-				DependencyKind.GenericParameterConstraintType,
-				DependencyKind.InterfaceImplementationInterfaceType,
-				DependencyKind.Ldtoken,
-				DependencyKind.ModifierType,
-				DependencyKind.InstructionTypeRef,
-				DependencyKind.ParameterType,
-				DependencyKind.ReferencedBySpecialAttribute,
-				DependencyKind.ReturnType,
-				DependencyKind.UnreachableBodyRequirement,
-				DependencyKind.VariableType,
-			};
-			if (!reasons.Contains (reason.Kind))
-				throw new InvalidOperationException ($"Internal error: unsupported type dependency {reason.Kind}");
+			if (!_typeReasons.Contains (reason.Kind))
+				throw new ArgumentOutOfRangeException ($"Internal error: unsupported type dependency {reason.Kind}");
 #endif
 			if (reference == null)
 				return null;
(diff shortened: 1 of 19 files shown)


sbomer:
@mrvoorhe in this change I've modified the signatures to a number of public methods, and wanted to ask what I can do to make this change easy to pick up in unity. Would you like me to add overloads of Mark* methods that just pass in a default (`Unspecified`) reason?

src/linker/Linker/DependencyInfo.cs
```diff
@@ -0,0 +1,142 @@
...
+
+	// For use when we don't care about tracking a particular dependency
+		Unspecified, // currently unused
+	}
+
+	readonly public struct DependencyInfo : IEquatable<DependencyInfo> {
+		public DependencyKind Kind { get; }
+		public object Source { get; }
+		public DependencyInfo (DependencyKind kind, object source) => (Kind, Source) = (kind, source);
+		public DependencyInfo (DependencyKind kind) => (Kind, Source) = (kind, default);
```
marek-safar: This constructor does not make much sense to me (it should be set of static fields/properties instead)

src/linker/Linker/DependencyInfo.cs
```diff
@@ -0,0 +1,142 @@
+using System;
+
+namespace Mono.Linker
+{
+	public enum DependencyKind {
```
marek-safar: This might be better to be explicitly numbered

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -273,9 +289,9 @@ bool QueueIsEmpty ()
 			return _methods.Count == 0;
 		}
 
-		protected virtual void EnqueueMethod (MethodDefinition method)
+		protected virtual void EnqueueMethod (MethodDefinition method, DependencyInfo reason)
 		{
-			_methods.Enqueue (method);
+			_methods.Enqueue ((method, reason));
```
marek-safar: This is not good. Please make this code memory neutral to what is today when tracing is disabled

src/linker/Linker/DependencyInfo.cs
```diff
@@ -0,0 +1,142 @@
...
+	// Linker internals, requirements for certain optimizations
+		UnreachableBodyRequirement, // method -> well-known type required for unreachable bodies optimization
+		DisablePrivateReflectionRequirement, // null -> DisablePrivateReflectionAttribute type/methods (note that no specific source is reported)
+		AlreadyMarked, // null -> member that has already been marked for a particular reason (used to propagate reasons internally, not reported)
+
+	// For use when we don't care about tracking a particular dependency
+		Unspecified, // currently unused
+	}
+
+	readonly public struct DependencyInfo : IEquatable<DependencyInfo> {
```
mrvoorhe: Do you have any suggestions on how we could extend this mechanism with additional dependency information that is unity specific?

Modifying the DependencyKind enum in our fork is bound to become a maintenance nightmare.

mrvoorhe:
@sbomer let me think about.  My initial reaction is that there are a lot of api breaks and I’d rather just break the APIs and pay the price.  If you add over loads then I won’t be able to depend on the compiler to tell me what to fix and I’ll probably just end up missing a bunch of things.

Can you give me a day or two to review this? I’d like to take a closer look at the changes.

mrvoorhe:
@sbomer have you checked if there is any performance impact with these changes?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1920,52 +2055,125 @@ AssemblyDefinition ResolveAssembly (IMetadataScope scope)
...
-			MarkCustomAttributes (method);
-			MarkSecurityDeclarations (method);
+			if (!markedForCall)
+				MarkType (method.DeclaringType, new DependencyInfo (DependencyKind.DeclaringType, method));
+			MarkCustomAttributes (method, new DependencyInfo (DependencyKind.CustomAttribute, method));
+			MarkSecurityDeclarations (method, new DependencyInfo (DependencyKind.CustomAttribute, method));
 
 			MarkGenericParameterProvider (method);
 
-			if (ShouldMarkAsInstancePossible (method))
```
mrvoorhe: This makes me a little nervous. Note to self - take a closer look

src/linker/Linker/DependencyInfo.cs
marek-safar: I think it should be fine to include your values here as well

src/linker/Linker/XmlDependencyRecorder.cs
```diff
@@ -89,6 +89,12 @@ public void Dispose ()
 			stream = null;
 		}
 
+		public void RecordDependency (object target, DependencyInfo reason, bool marked)
+		{
+			// Used for the DependencyInfo-based recorders. Don't do anything as we only
```
vitek-karas: Can we simply call `RecordDependency(reason.Source, target, marked)` (and eventually adapt this to the new tracing and actually preserve the reasons as well).

src/linker/Linker/Tracer.cs
```diff
@@ -90,11 +101,24 @@ bool IsRecordingEnabled ()
...
 
+		public void AddDirectDependency (object target, DependencyInfo reason, bool marked)
+		{
+			if (IsRecordingEnabled ()) {
+				foreach (IDependencyRecorder recorder in recorders)
+					recorder.RecordDependency (target, reason, marked);
+			}
+		}
+
 		public void AddDependency (object o, bool marked = false)
```
vitek-karas: I assume we will eventually remove this method - right? So all the logic around keeping the stack and so on can be removed as well?

src/linker/Linker/Tracer.cs
```diff
@@ -37,6 +38,7 @@ public class Tracer
 
 		Stack<object> dependency_stack;
 		List<IDependencyRecorder> recorders;
+		List<IReflectionPatternRecorder> reflectionPatternRecorders;
```
vitek-karas: Why did you move the reflection pattern recorder into the Tracer? I don't see any reason for this?

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: This should not allocate new objects - I thought that `(var, var)` is implemented as value types and so this will make the queue slightly larger, but it should not cause new objects on its own. Or are you worried about the size of the queue itself? (Maybe we should measure the max size of the queue when processing some largish app to get a feel of how bad it would be).

src/linker/Linker/DependencyInfo.cs
```diff
@@ -0,0 +1,142 @@
...
+
+	// Tracking cctors
+		TriggersCctorThroughFieldAccess, // field-accessing method -> cctor of field's declaring type
+		TriggersCctorForCalledMethod, // caller method -> callee method type cctor
+		DeclaringTypeOfCalledMethod, // called method -> its declaring type (used to track when a cctor is triggered by a method call)
+		CctorForType, // type -> cctor of type
+		CctorForField, // field -> cctor of field's declaring type
+
+	// Tracking instantiations
+		InstantiatedByCtor, // ctor -> type it instantiates
```
mrvoorhe: typos.  I think you meant

`type is instantiated`

src/linker/Linker/DependencyInfo.cs
mrvoorhe: > I think it should be fine to include your values here as well
I'd rather not do that.  That means every time we add a new value to our linker we'd have to remember to take it upstream.

@vitek-karas I don't entirely follow what that would look like.  Here is an example of one thing we do.  Can you show me what that would look like with your proposal?

https://gist.github.com/mrvoorhe/4711e93688968f4a8a13dda2344c4727

mrvoorhe:
What are you planning to do with this additional data?  Are you going to add to the analyzer?  Are new visualizations coming?

src/linker/Linker.Steps/MarkStep.cs
mrvoorhe: This looks OK.

mrvoorhe:
@sbomer I'd say don't bother adding overloads.  That's going to create a lot of clutter which isn't great either.  I think it's better to just bite the bullet when I sync up next time.

src/linker/Linker/DependencyInfo.cs
vitek-karas: For example:
``` C#
class MonoBehaviourAnnotation
{
    public bool IsMainScript { get; set; }
}

        protected void ProcessType(TypeDefinition typeDefinition)
        {
            using (Tracer.PushScoped(typeDefinition))
            {
                 // Mark all types derived from MonoBehaviour
                var dependencyInfo = new DependencyInfo(DependencyKind.Unspecified, "Built in rule for MonoBehaviour");
                dependencyInfo.AddAnnotation(new MonoBehaviourAnnotation() { IsMainScript = false });
                _context.Annotations.Mark(typeDefinition, dependencyInfo);
            }
        }

        // Implementation of IDependencyResolver
        override void RecordDependency(object target, DependencyInfo reason, bool marked)
        {
            if (reason.GetAnnotation<MonoBehaviourAnnotation>()?.IsMainScript)
                 Console.WriteLine($"{target} is the main script");
        }

**Comment by vitek-karas** [comment]
The use cases are:
* The analysis needs more details to be able to provide sensible "call stacks" for issues it finds. This is important if we find an issue somewhere in a library (the developer doesn't have code for it), so we need to "bubble up" the issue to the public API which is called from the app. Otherwise the error message would be pretty much meaningless for the developer.
* I think this will be useful for better analysis of "why did linker decide to keep this thing" - as right now some of the dependencies reported are very vague or hard to reason about.

File: src/linker/Linker/Annotations.cs [hunk 10]
```diff
@@ -150,17 +151,39 @@ public bool HasSubstitutedInit (TypeDefinition type)
...
 		}
 
+		[Obsolete ("Mark token providers with a reason instead.")]
 		public void Mark (IMetadataTokenProvider provider)
 		{
 			marked.Add (provider);
 			Tracer.AddDependency (provider, true);
 		}
 
+		public void Mark (IMetadataTokenProvider provider, DependencyInfo reason)
```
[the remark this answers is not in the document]
marek-safar: Not sure if JIT copies readonly structs. I might be worth to pass is by ref if we expect the struct to grow, with now it the roughly same size as 64bit ref

src/linker/Linker.Steps/MarkStep.cs
marek-safar: I agree this needs measurement but there can be a lot of items in the queue (thousands) and this change increases the underlying array size for simple entry from 8 bytes about 3 times if tracing is disabled (default mode)

vitek-karas: 1000s would not worry me too much (we're talking KBs of memory - although if we cross the large-object limit that would not be ideal). That said it might be possible to do without passing the reason into the process method. It's really only used for better tracking and it might be possible to do that in here before we enqueue it. We might still need some kind of a hashtable of methods which were marked like this to get the desired behavior, but that should be a smaller structure. That said I don't think it's possible to get the exact same behavior without at least some additional memory.

src/linker/Linker/Annotations.cs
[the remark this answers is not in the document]
vitek-karas: If we use the annotations to make the DependencyInfo extensible, then it can't be readonly unfortunately.

src/linker/Linker.Steps/MarkStep.cs
marek-safar: Let's measure it it's possible that with netcore we get into a lot higher numbers (+ all the reallocations)

src/linker/Linker/DependencyInfo.cs
sbomer: I'm not sure I understand. How would static fields work? I avoided setters so that it could be readonly.

src/linker/Linker/DependencyInfo.cs
```diff
@@ -0,0 +1,142 @@
...
+		ReturnTypeAttribute,
+		GenericParameterCustomAttribute,
+		GenericParameterConstraintCustomAttribute,
+
+	// Dependencies of custom attributes
+		AttributeConstructor, // attribute -> its ctor
+		// used for security attributes, where we mark the type/properties directly
+		AttributeType, // attribute -> attribute type
+		AttributeProperty, // attribute -> attribute ptroperty
+		CustomAttributeArgumentType,
```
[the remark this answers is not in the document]
sbomer: Added comments to all of them, hopefully that's more clear. This one means the value of a custom attribute argument (when that value specifies a type)... it's the best name I could think of. Happy to take suggestions on any of the DependencyKind names!

src/linker/Linker/XmlDependencyRecorder.cs
sbomer: I wanted to avoid producing a lot of redundant output in the xml (which this would do since the stack-based tracer has a fair amount of overlap with the tracing that I'm adding). Or do you mean to remove the stack-based tracing?

src/linker/Linker/Tracer.cs
sbomer: Do you know if anyone is currently depending on the stack-based tracer output? I wasn't sure if I could replace it with the new tracing because the output will be different. The new tracing should capture a superset of what the old tracing does, but some of the dependencies will show up differently, and I haven't done a full analysis.

src/linker/Linker/Tracer.cs
sbomer: I was prototyping this with a graph-based IDependencyRecorder that reported paths to dangerous reflection sites. My recorder needed to know about the unanalyzed reflection patterns, so I set it up to work with a list of IReflectionPatternRecorders, at which point it seemed like it belonged in Tracer.cs - it really is another kind of tracing in my mind. I can put it in a separate PR if you prefer.

src/linker/Linker/XmlDependencyRecorder.cs
vitek-karas: I thought that the only reason we needed the "stack" in the existing tracer is the fact that we don't report source->target pairs with each "Mark" action. Now that we do, there should be no need for the stack.

src/linker/Linker/DependencyInfo.cs
sbomer: Updated the wording to make it more clear.

src/linker/Linker/Tracer.cs
vitek-karas: The only one I know for sure is the analyzer tool in this repo. We would have to adapt that one to the new layout (if it wouldn't work anyway, which is not clear). It's likely we will do some nontrivial amount of work on that tool anyway to produce something usable to hep with reducing the size of trimmed framework.

src/linker/Linker/Tracer.cs
vitek-karas: I see - I would definitely prefer a separate PR - since I still don't see why needs to live on the tracer. I can see the point of view where it is another kind of traces - I'm not sure I like it though :-)

sbomer:
We haven't decided exactly what form the analysis output should take since there are many ways to show parts of the graph, but here's an example of the kind of output we've been experimenting with @mrvoorhe : https://gist.github.com/sbomer/7e8e52f62b26c9d322b90f7fb6a9ed3d.

I didn't notice any obvious perf impact, and will collect some numbers soon.

src/linker/Linker/DependencyInfo.cs
sbomer: What if I added an optional third `object` field as you suggest, but made it public? This way the DependencyInfo struct can remain immutable (the extra info would be passed to a new ctor), and extenders can use that field however they like - including by defining a new type with generic a AddAnnotation<T> method?

src/linker/Linker/Annotations.cs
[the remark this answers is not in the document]
sbomer: Yes, I actually meant to pass them by ref but forgot... will do so.

src/linker/Linker/DependencyInfo.cs
marek-safar:
```C#
public static readonly DependencyInfo Unspecified = new DependencyInfo (DependencyKind.Unspecified, null);
```

src/linker/Linker/DependencyInfo.cs
vitek-karas: I guess it would work, but it's not as "nice" :-)
The advantage of the Type-as-a-key approach and have it directly on the info object is that anything in the system can add annotation without any other part of the system being affected by it (unless they explicitly cooperate). If we add a public object then we would have to make sure that everything which uses it knows how to cooperate with anything else which uses it. As an example if Unity would use it to store some additional info, and the MarkStep itself would want to use it store additional information about the dependency (field for example), then these two would somehow have to cooperate since they would have to share that one field to store the value in.

What are the advantages of keeping the struct readonly? Does it make it significantly easier for JIT to optimize or something else?

src/linker/Linker/DependencyInfo.cs
sbomer: Gotcha, thanks!

src/linker/Linker/DependencyInfo.cs
sbomer: I mainly used readonly to express the intention that DependencyInfo is a simple data type that only carries information about why a cecil object was kept. There might be some perf advantages like preventing copies for in parameters - I don't know enough about the JIT to say whether it can better optimize readonly structs.

I actually see readonly as enforcing that we don't need to worry about the kind of cooperation you mention. Whoever creates a new DependencyInfo must fully populate it, and shouldn't rely on any other part of the marking logic to change it by adding or removing information to/from it. It's a simpler conceptual model, and I usually like to keep things simple unless there's a good reason. :) How likely do you think it is that we'll need to support the kind of interaction where both Unity and MarkStep add information to the same dependency reason?

Another line of reasoning is that we already have AnnotationStore as the core mutable database of information we are tracking about the app's dependencies, and adding "annotations" to dependencies seems slightly redundant (there are definitely things you could do with annotations on DependencyInfo that couldn't be done with the current Annotations... I just can't think of a good example where we would want to).

It also occurs to me that we could reuse the `Source` field, maybe with a `DependencyKind.Custom`, to reduce the memory footprint.

src/linker/Linker.Steps/MarkStep.cs
sbomer: I've measured a max queue length of 29 for a small console app, 8361 for a new blazor app, and 174397 (!) for a new aspnet app. This seems to be because aspnet is rooting everything, and we put everything rooted onto the queue in one pass before processing any of it. That seems like something we might want to address separately if we're concerned about the memory consumption of the queue.

sbomer:
I measured the perf on three apps: a simple console app, a new blazor app, and a new aspnet app, and got these numbers:
|| console | blazor | aspnet |
|-|-|-|-|
| before (s) | 1.015±.004 | 4.142±.055 | 11.955±.091 |
| after (s) | 1.034±.005 | 4.239±.027 | 12.429±.150 |
| Δ% | 1.828±0.621 | 2.342±1.494 | 3.971±1.485 |

src/linker/Linker/XmlDependencyRecorder.cs
sbomer: I went ahead and removed the stack-based tracer.

src/linker/Linker/Tracer.cs
sbomer: Removed in the latest change.

src/linker/Linker/DependencyInfo.cs
mrvoorhe: Sounds like there are pros & cons to either approach. 

A 3rd field that is reserved for us would be OK.  A 3rd field that is used by both monolinker and unitylinker would be painful to maintain as @vitek-karas  points out.

Do what you can to put something in place.  Whatever you think is best.  When we pull this change in I'll try it out and see how it goes and we can refine as necessary.

mrvoorhe:
> We haven't decided exactly what form the analysis output should take since there are many ways to show parts of the graph, but here's an example of the kind of output we've been experimenting with @mrvoorhe : https://gist.github.com/sbomer/7e8e52f62b26c9d322b90f7fb6a9ed3d.

I'm more interested in visualization tools.  Raw data in a file isn't useful by itself.

src/linker/Linker/DependencyInfo.cs
vitek-karas: @sbomer you have a good point about the AnnotationStore - we could add that if needed.
I'm fine with whatever you decide here.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Can you please file a bug for the asp.net case? That number will cause all kinds of issues as it's way too big. I know that in theory the queue can be of any length, but we should try to fix the cases we impose on ourselves :-)

sbomer: Filed https://github.com/mono/linker/issues/1018.

src/linker/Linker/DependencyInfo.cs
sbomer: I'll add a `DependencyKind.Custom` with a `Source` reserved for use by extenders. If we find a case where that's too restrictive, we can change this.

src/linker/Linker/DependencyInfo.cs
```diff
@@ -0,0 +1,144 @@
...
+		EventSourceProviderField = 75, // EventSource derived type -> fields on nested Keywords/Tasks/Opcodes provider classes
+		MethodForSpecialType = 76, // type -> methods kept (currently used for MulticastDelegate)
+
+	// Linker internals, requirements for certain optimizations
+		UnreachableBodyRequirement = 77, // method -> well-known type required for unreachable bodies optimization
+		DisablePrivateReflectionRequirement = 78, // null -> DisablePrivateReflectionAttribute type/methods (note that no specific source is reported)
+		AlreadyMarked = 79, // null -> member that has already been marked for a particular reason (used to propagate reasons internally, not reported)
+
+	// For tracking any other kinds of dependencies in extensions to the core logic
+		Custom = 80, // the source is reserved to carry any dependency information tracked by the extender
```
marek-safar: Might be worth making this 0xFFFFFF or something like that

sbomer:
@marek-safar I think this is ready to go.

src/linker/Linker/Annotations.cs
```diff
@@ -151,6 +151,12 @@ public bool HasSubstitutedInit (TypeDefinition type)
 			return fieldType_init.Contains (type);
 		}
 
+		[Obsolete ("Mark token providers with a reason instead.")]
```
marek-safar: nit: should be under ifdef

src/linker/Linker/IDependencyRecorder.cs
```diff
@@ -40,5 +40,16 @@ public interface IDependencyRecorder
 		/// <remarks>The source and target are typically Cecil metadata objects (MethodDefinition, TypeDefinition, ...)
 		/// but they can also be the linker steps or really any other object.</remarks>
 		void RecordDependency (object source, object target, bool marked);
```
marek-safar: Shouldn't the be removed?

PR #1014, opened 2020-03-19
Add reference assembly build

sbomer:
This adds a new project that builds a reference assembly for some of
the surface area of the linker. The package is called illink. We
should finalize the name before releasing an official package. This
leaves a number of problems unsolved (described below), but provides
something we can use to start experimenting.

The preliminary included surface area contains a public members of
BaseStep and some of its dependencies, ~~in the FEATURE_ILLINK build flavor -
which notably makes the ref assembly incompatible with the monolinker
surface area. (illink has an AssemblyResolver derived from
Mono.Linker.DirectoryAssemblyResolver, whereas monolinker has one
derived from Mono.Cecil.BaseAssemblyResolver). Therefore the reference
assembly is currently designed for consumption by illink plugins, not
by monolinker plugins.~~

The reference assembly is built for netstandard2.0, chosen because it
is the lowest common denominator for netcoreapp3.0 and net471, the
TFMs for which we currently build Mono.Linker. The illink flavor
actually doesn't need to be built for net471 (only ILLink.Tasks does),
so we could choose a higher version of netstandard if we disabled that
build flavor.

The ref assembly package is unusual in a few ways, described below. We
may be able to change some of these quirks once we have a better
picture of how linker plugins will be produced and consumed.

- The package *does not* contain any implementation assemblies,
because it is not designed to be consumed by a project targeting a
platform TFM - rather it is only designed to be consumed
from *libraries* that will act as linker plugins. Attempting to
consume it directly from a netcoreapp3.0 project will work during
build, but will fail to locate illink.dll at runtime.

- ~~The package *does not* have a transitive package dependency on
Mono.Cecil (nor does it include Mono.Cecil) (edit: it now includes a
reference assembly for cecil). This is because the
reference assembly is built against a submodule of the cecil sources,
so we do not have direct control over the package version. This
requires the custom step developer to add a dependency on Mono.Cecil.~~
We are switching from a submodule dependency to a package dependency on
the official Mono.Cecil, which is a dependency of the reference package.
ILLink.Tasks also now ships with a copy of the official version.

To consume this package in a custom step, the custom step project
should:

- Target a netstandard version compatible with the reference
assembly's target framework (currently netstandard2.0).

- ~~Include a dependency on Mono.Cecil compatible with the version
referenced during the linker build (currently 0:11:1:0, which
incidentally ships in the official Mono.Cecil package version
11.1). Attempting to use a higher version (the current latest release
is 11.2) will result in the linker failing to load the custom step.~~


vitek-karas:
Can we generate a ref assembly Mono.Cecil from our submodule and include it in the package? I think that is what we want from the "End-user" perspective (and also for versioning reasons).

sbomer:
Updated to include a Mono.Cecil.dll reference assembly in the same package.

src/linker/ref/Linker.Steps/BaseStep.cs
```diff
@@ -0,0 +1,44 @@
+//
```
marek-safar: Use .net foundation header

src/linker/ref/Linker.Steps/BaseStep.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps {
+
+	public abstract class BaseStep : IStep {
+		public LinkContext Context { get { throw null; } }
+		public AnnotationStore Annotations { get { throw null; } }
+		public Tracer Tracer { get { throw null; } }
+		public MarkingHelpers MarkingHelpers => throw null;
```
marek-safar: Remove this one

src/linker/ref/Linker/Annotations.cs
```diff
@@ -0,0 +1,115 @@
...
+//
+
+using System;
+using System.Collections.Generic;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+
+namespace Mono.Linker {
+
+	public partial class AnnotationStore {
```
marek-safar: woho, let's start with 0 (or close to 0) members

src/linker/ref/Mono.Linker.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>Mono</RootNamespace>
+    <Version>0.2.0.0</Version>
+    <LangVersion>latest</LangVersion>
+    <AssemblyName>illink</AssemblyName>
+    <Description>IL Linker</Description>
+    <DefineConstants>$(DefineConstants);FEATURE_ILLINK</DefineConstants>
```
marek-safar: I don't think we need the define, this is illink only

src/linker/ref/Linker/AssemblyResolver.cs
```diff
@@ -0,0 +1,51 @@
+﻿//
```
marek-safar: Let's go with minimal set so APIs like that should not be exposed

src/linker/ref/Linker.Steps/BaseStep.cs
sbomer: Should the implementation files also have their headers updated?

src/linker/ref/Linker/Annotations.cs
sbomer: I removed everything but the Mark methods for now.

src/linker/ref/Linker/AssemblyResolver.cs
sbomer: Removed this.

sbomer:
I had a general question about removing public members from ref assemblies. I'm aware that it's often done when there are platform-specific implementations with larger surface area, or to forward types when they move assemblies. Neither of these apply in our case. Wouldn't it make more sense for us to use C# access modifiers to define an extensibility model? For example anything public in Annotations should probably not be public in the first place if we don't want custom steps to access it.

marek-safar:
Ideally yes but Unity hack same source code so we'd have to make everything they hack on internal which might be a bit better but not sure if it's worth the work

src/linker/ref/Mono.Linker.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>Mono</RootNamespace>
+    <Version>0.2.0.0</Version>
+    <LangVersion>latest</LangVersion>
+    <AssemblyName>illink</AssemblyName>
+    <Description>IL Linker</Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
marek-safar: Not sure about this one. It probably does not hurt just feels odd

src/linker/ref/Mono.Linker.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>Mono</RootNamespace>
+    <Version>0.2.0.0</Version>
```
marek-safar: Please extract this to common property for ILLink version

src/linker/ref/Mono.Linker.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <!-- Nuget issues a warning about missing <reference> items in the generated nuspec,
+        used for packages.config which we do not support. -->
+    <NoWarn>$(NoWarn);NU5131</NoWarn>
+    <!-- We need to create a package that includes the build output of the cecil projectreference below. -->
+    <IncludeBuildOutput>true</IncludeBuildOutput>
+    <BuildOutputTargetFolder>ref</BuildOutputTargetFolder>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Condition="'$(UseLocalCecil)' == 'false'" Include="Mono.Cecil" Version="$(MonoCecilVersion)" />
```
marek-safar: This project should be for illink only so it's fine to use package only

monolinker.sln
```diff
@@ -11,8 +11,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Mono.Linker.Tests.Cases", "
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Mono.Linker.Tests.Cases.Expectations", "test\Mono.Linker.Tests.Cases.Expectations\Mono.Linker.Tests.Cases.Expectations.csproj", "{2C26601F-3E2F-45B9-A02F-58EE9296E19E}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Mono.Cecil", "external\cecil\Mono.Cecil.csproj", "{5E89AB62-1526-41F7-AAA3-D9119C9451EA}"
```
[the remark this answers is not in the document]
sbomer: I removed it from the solutions for faster builds when using the package - I don't want it to build cecil if we're not going to use the local build. Did you want monolinker to continue using the local cecil?

eng/Version.Details.xml
```diff
@@ -7,5 +7,9 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>16d4350184cf362cd59807b589d1c93803025abc</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.DotNet.ApiCompat" Version="5.0.0-beta.20162.4">
+      <Uri>https://github.com/dotnet/arcade</Uri>
+      <Sha />
```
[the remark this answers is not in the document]
sbomer: Oops - meant to remove this. I don't think we want to put it in Version.Details.xml at all since there will be no dependency flow from jbevain/cecil. That seems to be in line with how the SDK handles Newtonsoft.Json.

src/linker/ref/Mono.Linker.csproj
sbomer: Not sure which one you mean - the TFM? IsPackable?

eng/Version.Details.xml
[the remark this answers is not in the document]
sbomer: It's empty because that's how `darc add-dependency` adds it. I think when dependency flow updates it will add the Sha.

monolinker.sln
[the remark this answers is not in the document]
marek-safar: for monolinker the setup will have to stay otherwise we get cecil mismatch in mono

src/linker/ref/Mono.Linker.csproj
marek-safar: TFM: netstandard2.0

src/linker/ref/Mono.Linker.csproj
sbomer: It doesn't make a huge difference, but I'd prefer to leave the switch here just for consistency with the implementation build if that's ok with you. (I would like to be able to build illink against the submodule locally, and in that case it'd be nice if the ref/impl built the same way and didn't require the package).

src/linker/ref/Mono.Linker.csproj
sbomer: @ericstj might have some guidance - would you be able to PTAL? In particular, do you have any guidance on picking the right TFM here? The PR description has context.

src/linker/ref/Linker/AssemblyAction.cs
```diff
@@ -0,0 +1,31 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Mono.Linker {
+
+	public enum AssemblyAction {
```
vitek-karas: Why do we need this? I don't see the `IStep` and related classes use this (or at least yet).

marek-safar: I think that's leftover from the clean up

src/linker/ref/Mono.Linker.csproj
ericstj: A netstandard2.0 ref lets folks build libraries that depend on this once and then they can run on all the frameworks that support netstandard2.0.  If you provide a netstandard2.0 ref you should make sure you cover *all* frameworks that support netstandard2.0, otherwise folks will hit a FileNotFoundException on the platforms where you don't provide an implementation.

sbomer: Thanks @ericstj. In our case the only expected way to consume a plugin that depends on this ref assembly is by loading the plugin into the linker - and we only ship the linker on netcoreapp3.0 and net471. In that case are there problems with using netstandard2.0 for the ref? (Should we be publishing netcoreapp3.0/net471 ref assemblies instead? That would seem weird to me since the plugins should in theory not depend on platform-specific APIs).

ericstj: I think plugins need to publish for a specific TFM (not netstandard) anyways so that you get all their TFM-specific dependencies.  You might be creating a false promise by giving them a netstandard2.0 ref and implying you can create a runnable netstandard layout.

marek-safar: I agree with Eric (we don't need net471) so the TFM should match to illink.task netcore tfm

sbomer: Thanks! So my understanding is there are two issues (inferring a bit here - correct me if I'm wrong):
- We can't be sure that plugins will *only* build against netstandard rather than specific TFMs, so we should instead force them into a specific TFM compatible with our implementation.
- There's an expectation that authors can create runnable layouts for any ref assembly, so we should ship an implementation to avoid violating that expectation.

Based on that, I plan to change the TFM to netcoreapp3.0, and to include the full illink.dll implementation assembly in the package (under lib) in addition to the ref. So this becomes more of a "linker library" package than a pure reference assembly.

The downside is that an author trying to create a portable netstandard plugin has to choose a specific TFM - which doesn't really make a difference since we only run on netcoreapp anyway. We might also get people trying to use this package to create their own linker drivers, which may or may not be feasible depending on the surface area we expose.

marek-safar: Focusing on netcore only is totally fine, we don't have any use case for .net framework setup in this scenario

PR #1015, opened 2020-03-20
Preserve marked children of XML descriptor

### LINKED ISSUE #965: required="false" on types doesn't preserve methods
I.e.
```
<type fullname="System.Reflection.Emit.TypeBuilder" required='false' preserve="fields" feature="sre">
			<method name="IsAssignableTo" feature="sre" />
```
doesn't keep IsAssignableTo alive.

mateoatr:
This fixes #965 by preserving all children marked in a XML descriptor regardless of the `required` attribute value.

### DIFF
--- a/src/linker/Linker.Steps/ResolveFromXmlStep.cs
+++ b/src/linker/Linker.Steps/ResolveFromXmlStep.cs
@@ -247,6 +247,7 @@ namespace Mono.Linker.Steps {
 
 			if (!IsRequired (nav)) {
 				Annotations.SetPreserve (type, preserve);
+				MarkChildren (type, nav);
 				return;
 			}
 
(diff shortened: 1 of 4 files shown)


src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -288,6 +283,16 @@ void MarkSelectedFields (XPathNavigator nav, TypeDefinition type)
 			ProcessFields (type, fields);
 		}
 
+		private void MarkChildren (TypeDefinition type, XPathNavigator nav)
```
marek-safar: nit: existing code style does not use redundant modifiers

test/Mono.Linker.Tests.Cases/LinkXml/CanPreserveMarkedChildrenOfUnrequiredType.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+using Mono.Linker.Tests.Cases.LinkXml.Dependencies;
+
+namespace Mono.Linker.Tests.Cases.LinkXml
+{
+	[SetupCompileBefore ("Library.dll", new [] { "Dependencies/CanPreserveMarkedChildrenOfUnrequiredType_Library.cs" })]
```
mrvoorhe: I don't see a reason why this test needs to setup an additional assembly.  I believe it could all be in the test assembly.

mateoatr: You're right. Thanks. Please take a look: #1019.

PR #1024, opened 2020-03-26
[feature/dataflow] Implement FlowAnnotations helper class

MichalStrehovsky:
* `IFlowAnnotationSource` lets us abstract away the source of the annotations. We expect to have an attribute-based one and JSON-based one (the JSON based one mostly for internal purposes right now - it probably won't be public).
* The `FlowAnnotations` class deals with caching and propagating the attributes from properties to the accessor methods and backing fields. The cache is per-type because the properties make this more complicated. Being per-type is also more memory efficient because we don't have to keep track of all methods/fields that we looked at and found out they're not interesting (the majority of cases). Not interesting things are not tracked and that's cached at the type level.

Vitek is working on the thing that consumes this. I already tested it a bit with Vitek's thing.

### DIFF
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -97,10 +97,9 @@ namespace Mono.Linker.Dataflow
 
 					// We convert indices from metadata space to IL space here.
 					// IL space assigns index 0 to the `this` parameter on instance methods.
-					int offset = method.HasThis && !method.ExplicitThis ? 1 : 0;
+					int offset = method.HasImplicitThis () ? 1 : 0;
 
-					for (int i = 0; i < method.Parameters.Count; i++) {
-						
+					for (int i = 0; i < method.Parameters.Count; i++) {						
 						if (!IsTypeInterestingForDataflow (method.Parameters [i].ParameterType)) {
 							continue;
 						}
@@ -165,7 +164,7 @@ namespace Mono.Linker.Dataflow
 						if (annotatedMethods.Any (a => a.Method == setMethod)) {
 							// TODO: warn: duplicate annotation. not propagating.
 						} else {
-							int offset = setMethod.HasThis && !setMethod.ExplicitThis ? 1 : 0;
+							int offset = setMethod.HasImplicitThis () ? 1 : 0;
 							if (setMethod.Parameters.Count > 0) {
 								DynamicallyAccessedMemberKinds [] paramAnnotations = new DynamicallyAccessedMemberKinds [setMethod.Parameters.Count + offset];
 								paramAnnotations [offset] = annotation;
@@ -176,7 +175,7 @@ namespace Mono.Linker.Dataflow
 
 					FieldDefinition backingFieldFromGetter = null;
 
-					// Propagate the annotation to the setter method
+					// Propagate the annotation to the getter method
 					MethodDefinition getMethod = property.GetMethod;
 					if (getMethod != null) {
 
@@ -216,7 +215,7 @@ namespace Mono.Linker.Dataflow
 		private bool ScanMethodBodyForFieldAccess (MethodBody body, bool write, out FieldDefinition found)
 		{
 			// Tries to find the backing field for a property getter/setter.
-			// Returns true if this is a method body that we can unambiguous analyze.
+			// Returns true if this is a method body that we can unambiguously analyze.
 			// The found field could still be null if there's no backing store.
 
 			// TODO: could restrict this to compiler-generated fields as well so that
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -218,10 +218,7 @@ namespace Mono.Linker.Dataflow
 			// Returns true if this is a method body that we can unambiguously analyze.
 			// The found field could still be null if there's no backing store.
 
-			// TODO: could restrict this to compiler-generated fields as well so that
-			// it only works for auto properties, but maybe that's too restrictive.
-
-			found = null;
+			FieldReference foundReference = null;
 
 			foreach (Instruction instruction in body.Instructions) {
 				switch (instruction.OpCode.Code) {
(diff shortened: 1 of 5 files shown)


src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -0,0 +1,340 @@
+﻿using Mono.Cecil;
+using Mono.Cecil.Cil;
+using System;
+using System.Collections.Generic;
+
+namespace Mono.Linker.Dataflow
+{
+	// TODO: enforce virtual methods have consistent annotations
```
vitek-karas: File a tracking issue for this - just in case :-)

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -0,0 +1,340 @@
...
+
+			return 0;
+		}
+
+		private TypeAnnotations GetAnnotations(TypeDefinition type)
+		{
+			if (!_annotations.TryGetValue(type, out TypeAnnotations value)) {
+				value = BuildTypeAnnotations (type);
+				_annotations.Add (type, value);
+			}
```
vitek-karas: This means we will end up with a `TypeAnnotations` instance (along with several empty arrays on it) for every type in the system. Given that there will be proportionally tiny number of interesting types that seems wasteful. Can we store nulls in the dictionary for the cases where the type is not interesting (no annotations anywhere on it)? That should help a lot.

vitek-karas: I see it's a struct sorry - my bad. I guess it's OK then... it's still larger than one pointer per type, but not that bad.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -0,0 +1,340 @@
...
+			if (type.HasMethods) {
+				foreach (MethodDefinition method in type.Methods) {
+					DynamicallyAccessedMemberKinds [] paramAnnotations = null;
+
+					// We convert indices from metadata space to IL space here.
+					// IL space assigns index 0 to the `this` parameter on instance methods.
+					int offset = method.HasThis && !method.ExplicitThis ? 1 : 0;
+
+					for (int i = 0; i < method.Parameters.Count; i++) {
+
```
vitek-karas: Nit: empty line

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -0,0 +1,340 @@
...
+					if (setMethod != null) {
+
+						if (!ScanMethodBodyForFieldAccess (setMethod.Body, write: true, out backingFieldFromSetter)) {
+							// TODO: warn we couldn't find a unique backing field
+						}
+
+						if (annotatedMethods.Any (a => a.Method == setMethod)) {
+							// TODO: warn: duplicate annotation. not propagating.
+						} else {
+							int offset = setMethod.HasThis && !setMethod.ExplicitThis ? 1 : 0;
```
vitek-karas: Can you please create a helper for this - the code to propagate annotations will also need this - and it would be nice to have a "better" name for this operation.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -0,0 +1,340 @@
...
+								DynamicallyAccessedMemberKinds [] paramAnnotations = new DynamicallyAccessedMemberKinds [setMethod.Parameters.Count + offset];
+								paramAnnotations [offset] = annotation;
+								annotatedMethods.Add (new MethodAnnotations (setMethod, paramAnnotations, 0));
+							}
+						}
+					}
+
+					FieldDefinition backingFieldFromGetter = null;
+
+					// Propagate the annotation to the setter method
```
vitek-karas:
```suggestion
					// Propagate the annotation to the getter method
```

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -0,0 +1,340 @@
...
+				}
+			}
+
+			return new TypeAnnotations (annotatedMethods.ToArray(), annotatedFields.ToArray());
+		}
+
+		private bool ScanMethodBodyForFieldAccess (MethodBody body, bool write, out FieldDefinition found)
+		{
+			// Tries to find the backing field for a property getter/setter.
+			// Returns true if this is a method body that we can unambiguous analyze.
```
vitek-karas:
```suggestion
			// Returns true if this is a method body that we can unambiguously analyze.
```

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -0,0 +1,340 @@
...
+						FieldDefinition field = (instruction.Operand as FieldReference)?.Resolve ();
+						if (field != null && field.IsStatic == body.Method.IsStatic) {
+							if (found != null) {
+								// This writes/reads multiple fields - can't guess which one is the backing store.
+								found = null;
+								return false;
+							}
+
+							found = field;
+						}
```
vitek-karas: Should we fail if the field can't be resolved or it's not static as expected? Because if the getter/setter has two field accesses, one which passes this check and one which doesn't we will still treat it as "yeah, we completely understand this one" - which is not true. I would rather this to be very careful and only return non-null when it's absolutely sure it got it right.

MichalStrehovsky: If we want to be absolutely sure, we need to run the flow analysis on this.

E.g.

```csharp
[DynamicallyAccessedMembers(Constructors)]
Type BaseType => _isValueType ? typeof(ValueType) : typeof(Object);
```

This would currently incorrectly propagate the annotation to the `_isValueType` field. We could check that the type of the field matches the type of the property but even that one is not certain.

That's why I left the `TODO: could restrict this to compiler-generated fields as well so that it only works for auto properties, but maybe that's too restrictive.` - maybe we should do that...

src/linker/Linker.Dataflow/FlowAnnotations.cs
MichalStrehovsky: #1028

src/linker/Linker.Dataflow/FlowAnnotations.cs
vitek-karas: Hmm - yeah - I'm starting to like that more - it is restrictive, but on the other hand it's correct - which wins it for me.

PR #1031, opened 2020-03-27
[feature/dataflow] Propagate data flow annotations across method calls

vitek-karas:
Implements propagation of data flow annotations from/to method parameters and from/to method return value.

Still lacks some functionality:
* Not all member access kinds are implemented - only ctors are
* Error reporting is pretty bad, needs more work

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -50,6 +50,8 @@ namespace Mono.Linker.Steps {
 		protected List<TypeDefinition> _typesWithInterfaces;
 		protected List<MethodBody> _unreachableBodies;
 
+		readonly FlowAnnotations _flowAnnotations;
+
 		public MarkStep ()
 		{
 			_methods = new Queue<MethodDefinition> ();
@@ -58,6 +60,8 @@ namespace Mono.Linker.Steps {
 			_lateMarkedAttributes = new Queue<AttributeProviderPair> ();
 			_typesWithInterfaces = new List<TypeDefinition> ();
 			_unreachableBodies = new List<MethodBody> ();
+
+			_flowAnnotations = new FlowAnnotations (new AttributeFlowAnnotationSource (), _context);
 		}
 
 		public AnnotationStore Annotations => _context.Annotations;
@@ -2377,7 +2381,7 @@ namespace Mono.Linker.Steps {
 				return;
 
 			var scanner = new ReflectionMethodBodyScanner (this);
-			scanner.Scan (body);
+			scanner.ScanAndProcessReturnValue (body);
 
 			var instructions = body.Instructions;
 			ReflectionPatternDetector detector = new ReflectionPatternDetector (this, body.Method);
@@ -3379,10 +3383,26 @@ namespace Mono.Linker.Steps {
 		private class ReflectionMethodBodyScanner : Dataflow.MethodBodyScanner
 		{
 			private readonly MarkStep _markStep;
+			private readonly FlowAnnotations _flowAnnotations;
 
 			public ReflectionMethodBodyScanner(MarkStep parent)
 			{
 				_markStep = parent;
+				_flowAnnotations = _markStep._flowAnnotations;
+			}
+
+			public void ScanAndProcessReturnValue (MethodBody methodBody)
+			{
+				Scan (methodBody);
+
+				if (MethodReturnValue != null) {
+					var requiredMemberKinds = _flowAnnotations.GetReturnParameterAnnotation (methodBody.Method);
+					if (requiredMemberKinds != 0) {
+						var reflectionContext = new ReflectionPatternContext (_markStep._context, methodBody.Method, methodBody.Method, 0);
+						reflectionContext.AnalyzingPattern ();
+						RequireDynamicallyAccessedMembers (ref reflectionContext, requiredMemberKinds, MethodReturnValue);
+					}
+				}
 			}
 
 			protected override void WarnAboutInvalidILInMethod (MethodBody method, int ilOffset)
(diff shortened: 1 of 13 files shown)


test/Mono.Linker.Tests.Cases/DataFlow/MethodParametersDataFlow.cs
```diff
@@ -0,0 +1,197 @@
+﻿using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Mono.Linker.Tests.Cases.DataFlow
+{
+	[SetupCSharpCompilerToUse ("csc")]
```
mrvoorhe: You shouldn't need this anymore.  I made roslyn the default compiler a week or two ago.

test/Mono.Linker.Tests.Cases/DataFlow/MethodParametersDataFlow.cs
```diff
@@ -0,0 +1,197 @@
...
+			instance.UnknownValueToUnAnnotatedParameter ();
+			instance.UnknownValueToUnAnnotatedParameterOnInterestingMethod ();
+		}
+
+		[Kept]
+		[UnrecognizedReflectionAccessPattern (typeof (MethodParametersDataFlow), nameof (RequirePublicConstructors), new Type [] { typeof (Type) })]
+		[UnrecognizedReflectionAccessPattern (typeof (MethodParametersDataFlow), nameof (RequireConstructors), new Type [] { typeof (Type) })]
+		private static void DefaultConstructorParameter (
+			[DynamicallyAccessedMembers(DynamicallyAccessedMemberKinds.DefaultConstructor)]
+			[KeptAttributeAttribute(typeof(DynamicallyAccessedMembersAttribute))]
```
mrvoorhe: Are you planning to hook up `DynamicallyAccessMembersAttribute` to https://github.com/mono/linker/blob/83a48f3ca59d9587669712291119ef730a6551d4/src/linker/Linker.Steps/MarkStep.cs#L515  the same way `PreserveDependencyAttribute`?

At the very least it would be nice to do from a test writing perspective so that you don't have to sprinkle `[KeptAttributeAttribute(typeof(DynamicallyAccessedMembersAttribute))]` all over the tests

test/Mono.Linker.Tests.Cases/DataFlow/MethodParametersDataFlow.cs
```diff
@@ -0,0 +1,197 @@
...
+{
+	[SetupCSharpCompilerToUse ("csc")]
+	[KeptMember(".ctor()")]
+	public class MethodParametersDataFlow
+	{
+		public static void Main ()
+		{
+			var instance = new MethodParametersDataFlow ();
+
+			DefaultConstructorParameter (typeof (TestType));
```
mrvoorhe: I'm not a big fan of loading up a single test case with multiple usage scenarios.  It's easy to get yourself into a situation where your test isn't providing the coverage that you think it is because multiple things lead to the same thing being marked and if one of those mechanisms broke, you'd never know.

After the first `[DynamicallyAccessedMembers(DynamicallyAccessedMemberKinds.DefaultConstructor)]` is processed any additional uses of that attribute with the same `DynamicallyAccessedMemberKinds` value and the same Type are not providing much in the way of test coverage.

I put a break point in `EnqueueMethod` for the `TestType.ctor()` and `RequireDynamicallyAccessedMembers` is marking it >3 times.  If every usage after the first broke you'd never know.

test/Mono.Linker.Tests.Cases/Reflection/ActivatorCreateInstance.cs
```diff
@@ -1,11 +1,13 @@
 using System;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 using Mono.Linker.Tests.Cases.Expectations.Metadata;
 
 namespace Mono.Linker.Tests.Cases.Reflection
 {
 	[SetupCSharpCompilerToUse ("csc")]
+	[KeptMember(".ctor()")]
```
mrvoorhe: Why is this constructor kept now?

test/Mono.Linker.Tests.Cases/DataFlow/MethodParametersDataFlow.cs
mrvoorhe: Or maybe I'm misunderstanding and the primary goal of these tests isn't whether or not `TestType` ctors are marked?  But rather that the `UnrecognizedReflectionAccessPattern` are reported correctly?

vitek-karas: The purpose of the test is to validate the `UnrecognizedReflectionAccessPattern` behavior. These will eventually turn into warnings, so maybe we should write the tests later on validating the warnings instead, but I prefer specific approach like this better myself.
The marking of the `TestType.ctor` is a "side-effect" of these tests. In fact I considered if I should add a way for the test to opt-out of the `Kept` validation as nothing in this test actually needs that.

test/Mono.Linker.Tests.Cases/DataFlow/MethodParametersDataFlow.cs
vitek-karas: Yes - we probably should. The linker should by default remove these attributes, so that would match the desired behavior nicely.

test/Mono.Linker.Tests.Cases/Reflection/ActivatorCreateInstance.cs
vitek-karas: Because the test instantiates the test class (see below in `Main`). This was the simplest way as there's a rather tricky behavior difference in Cecil calling static versus instance methods, so I wanted to make sure that the logic works correctly in both cases.

test/Mono.Linker.Tests.Cases/DataFlow/MethodParametersDataFlow.cs
vitek-karas: Filed https://github.com/mono/linker/issues/1034 to track that work

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3485,22 +3522,42 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+									if (requiredMemberKinds != 0)
+										RequireDynamicallyAccessedMembers (ref reflectionContext, requiredMemberKinds, methodParams [parameterIndex]);
+								}
+
+								reflectionContext.RecordHandledPattern ();
+							}
+
+							// To get good reporting of errors we need to track the origin of the value for all method calls
+							// but except Newobj as those are special.
+							if (calledMethodDefinition.ReturnType.MetadataType != MetadataType.Void && operation.OpCode.Code != Code.Newobj) {
```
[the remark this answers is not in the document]
vitek-karas: Removed it. Thanks.

test/Mono.Linker.Tests.Cases/DataFlow/MethodParametersDataFlow.cs
vitek-karas: I added comment to the test.

What do you think about introducing a way to disable the `Kept` validation for specific tests?

vitek-karas:
I'll merge this and will followup with improvements around the `Kept` attribute on some of these tests.
Feel free to leave other comments here and I'll react in a separate PR.

test/Mono.Linker.Tests.Cases/DataFlow/MethodParametersDataFlow.cs
mrvoorhe: 👍

PR #1032, opened 2020-03-27
Add core externally used apis to ref assembly

marek-safar:


src/linker/Linker/OverrideInformation.cs
```diff
@@ -4,17 +4,17 @@
...
 		public OverrideInformation (MethodDefinition @base, MethodDefinition @override, InterfaceImplementation matchingInterfaceImplementation = null)
 		{
 			Base = @base;
 			Override = @override;
 			MatchingInterfaceImplementation = matchingInterfaceImplementation;
 		}
 
+		public MethodDefinition Base { get; private set; }
+		public MethodDefinition Override { get; private set; }
+		public InterfaceImplementation MatchingInterfaceImplementation { get; private set; }
```
sbomer: Why add setters? It looks like we can keep these readonly since they're only set in the ctor.

marek-safar: It's more resilient, one can add or change set/get behaviour if needed without breaking the contract.

sbomer: Adding getters makes sense to me - I am specifically asking about the setters. If you only add a getter, the underlying field can still be readonly without changing the contract, and you can always introduce setters later if needed: https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyDMABGgQMIEDeOB1RhARgPYMQEBCFBA5gKbADcBAL5Ua+UgAoAlBRE0a7ALwFgAJwCu3PrKE5BQA=

marek-safar: I see what you mean

src/linker/ref/Linker/Annotations.cs
```diff
@@ -2,17 +2,27 @@
...
 using Mono.Cecil;
-using Mono.Cecil.Cil;
 
 namespace Mono.Linker {
 
 	public partial class AnnotationStore
 	{
 		internal AnnotationStore() {}
+
+		public List<OverrideInformation> GetOverrides (MethodDefinition method) { throw null; }
```
vitek-karas: Should we try to change these APIs to be a bit more restrictive - given that they are becoming an official public API.
For example this one could return `IEnumerable`. And the `GetCustomAnnotations` could be replaced by something like `TryGetCustomAnnotation(object key, IMetadataTokenProvider item, out object value)` and `SetCustomAnnotation(object key, IMetadataTokenProvider item, object value)`. Basically trying to hide the implementation details.

src/linker/ref/Linker/OverrideInformation.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	public class OverrideInformation
+	{
+		internal OverrideInformation ()
```
vitek-karas: Why do we need to declare an internal .ctor in a ref assembly?

marek-safar: There were errors about missing ctor in apidiff tool

src/linker/ref/Linker/Annotations.cs
marek-safar: That makes sense to me (will change that in follow up PR)

src/linker/ref/Linker/OverrideInformation.cs
sbomer: (Specifically if you don't declare a ctor, the ref assembly will have the implicit public ctor, which doesn't exist in the implementation.)

vitek-karas: Good to know - thanks!

PR #1033, opened 2020-03-27
Align nuget package versioning with .NET5

marek-safar:


eng/Versions.props
```diff
@@ -1,8 +1,14 @@
...
-    <VersionPrefix>0.1.6</VersionPrefix>
-    <PreReleaseVersionLabel>prerelease</PreReleaseVersionLabel>
+    <!-- File version numbers -->
+    <MajorVersion>5</MajorVersion>
+    <MinorVersion>0</MinorVersion>
+    <PatchVersion>0</PatchVersion>
+    <PreReleaseVersionLabel>preview</PreReleaseVersionLabel>
+    <PreReleaseVersionIteration>3</PreReleaseVersionIteration>
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
+    <DotNetFinalVersionKind Condition="'$(StabilizePackageVersion)' == 'true'">release</DotNetFinalVersionKind>
```
sbomer: Why use two properties for this? If we ever want to set `DotNetFinalVersionKind` to `prerelease` (for the final prerelease before the official build) we would need to change this.

marek-safar: I'm not sure it looks like this is the way for arcade to uniformly build final packages from the command line. It could be simplified but not sure we need to diverge here

sbomer: I see, it looks like `StabilizePackageVersion` is a convention followed by dotnet/runtime and others. `DotNetFinalVersionKind` is the arcade property. I'm fine with following the dotnet/runtime convention.

PR #1036, opened 2020-03-27
Add ability to turn of validation of the Kept attributes

vitek-karas:
It disables the entire AssemblyChecker, but I wanted an attribute name which is not tied to the implementation details. And currently the AssemblyChecker really only validates the Kept attributes.


marek-safar:
What is this going to be used for?

/cc @mrvoorhe

vitek-karas:
See this test for an example: https://github.com/mono/linker/blob/357bd34e2bcc83123f7b16764d18ba1ed843225e/test/Mono.Linker.Tests.Cases/DataFlow/MethodParametersDataFlow.cs#L17

The test needs to use the compile->linker infra as it's exercising linker all up, but the trimming parts are not that interesting as it only validates correct propagation of the annotations. Having the ability to disable the `Kept` verification would make it easier to write the test as it would need all the `Kept` attributes around.

test/Mono.Linker.Tests.Cases.Expectations/Metadata/DisableKeptItemsValidationAttribute.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Metadata
+{
+	[AttributeUsage (AttributeTargets.Assembly, AllowMultiple = false)]
+	public class DisableKeptItemsValidationAttribute : BaseMetadataAttribute
```
mrvoorhe: My preference would be to use `AttributeTargets.Class`.  This is how other test setup type attributes are defined today.  For example
```
[SkipPeVerify]
```

I think it's easiest to read the attributes on the test type when trying to understand what the test will do.

Also, putting this attribute at the assembly level would come with some annoyances.  For example, if multiple tests want to use this attribute they would need to #if theirs out and then use `[Define]` to include it when the test is compiled.

Now that I think about this more, what you are adding is really the same type of thing as `SkipPeVerifyAttribute`.  You are adding an attribute that disables the test frameworks default behavior.  So I would prefer to follow the same pattern as `SkipPeVerifyAttribute`.  This attribute should 

* Move to `Mono.Linker.Tests.Cases.Expectations.Assertions`
* Inherit from `BaseExpectedLinkedBehaviorAttribute`
* And have on it `[AttributeUsage (AttributeTargets.Class, AllowMultiple = false)]`

marek-safar: @vitek-karas could you address the comments?

vitek-karas: Changed the attribute as per @mrvoorhe suggestion.

vitek-karas:
@mrvoorhe Can you please take a second look if the new version aligns with what you were thinking about?

PR #1039, opened 2020-03-28
[feature/dataflow] Port Expression.* pattern recognition to new infra

mateoatr:


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3407,6 +3279,106 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+							&& calledMethod.Parameters.Count == 4
+							&& calledMethod.Parameters [0].ParameterType.FullName == "System.Type": {
+
+								reflectionContext.AnalyzingPattern ();
+
+								foreach (var value in methodParams [0].UniqueValues ()) {
+									if (value is SystemTypeValue systemTypeValue) {
+										foreach (var stringParam in methodParams [1].UniqueValues ()) {
+											if (stringParam is KnownStringValue stringValue) {
+												MarkMethodsFromReflectionCall (ref reflectionContext, systemTypeValue.TypeRepresented, stringValue.Contents, BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic);
```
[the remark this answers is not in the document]
vitek-karas: Same comment applies to `Property` and `Field`.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3407,6 +3279,106 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+											} else if (stringParam is NullValue) {
+												reflectionContext.RecordHandledPattern ();
+											} else if (stringParam is MethodParameterValue) {
+												// TODO: Check if parameter is annotated.
```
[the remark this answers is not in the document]
vitek-karas: Similar comment applies to `Field` and `Property` - although in those cases I think we're pretty sure we want to fall back to keeping all fields/properties.

test/Mono.Linker.Tests.Cases/Reflection/ExpressionCallString.cs
```diff
@@ -11,17 +11,24 @@ namespace Mono.Linker.Tests.Cases.Reflection
...
-			TestNullType ();
-			TestDataFlowType ();
 		}
 
 		[RecognizedReflectionAccessPattern (
-			typeof (Expression), nameof (Expression.Call), new Type [] { typeof (Type), typeof (string), typeof (Type[]), typeof (Expression []) },
+			typeof (Expression), nameof (Expression.Call), new Type [] { typeof (Type), typeof (string), typeof (Type []), typeof (Expression []) },
 			typeof (ExpressionCallString), nameof (OnlyCalledViaExpression), new Type [0])]
 		[Kept]
 		static void TestByName ()
```
vitek-karas: We should add test which actually validates that multiple values are resolved. So for example:
```C#
string name;
switch (somevalue) {
case 1: name = "Foo"
case 2: name = "Bar"
}
Expression.Call (typeof(T), name, Type.EmptyTypes);
```

And then validate that both `Foo` and `Bar` are kept.

test/Mono.Linker.Tests.Cases/Reflection/ExpressionFieldString.cs
```diff
@@ -22,5 +25,57 @@ public static void Main ()
...
+		[Kept]
+		static string UnknownString ()
+		{
+			return "unknownstring";
+		}
+
+		[Kept]
+		static void Branch_SystemTypeValueNode_KnownStringValue_NonStatic ()
+		{
+			var expr = Expression.Field (Expression.Parameter (typeof (int), "somename"), typeof (ExpressionFieldString), "TestName1");
```
vitek-karas: Similar comment to `Call` - test the ability to process multiple potential values for name and type.

test/Mono.Linker.Tests.Cases/Reflection/ExpressionCallString.cs
vitek-karas: And similarly for the type parameter (two different `typeof()`).

test/Mono.Linker.Tests.Cases/Reflection/ExpressionNewType.cs
```diff
@@ -20,8 +42,38 @@ class T1
...
+		[Kept]
+		static void Branch_SystemTypeValueNode ()
+		{
+			var expr = Expression.New (typeof (T1));
+		}
+
+		[Kept]
+		static void Branch_NullValueNode ()
+		{
+			var expr = Expression.New (Helper1 () + Helper2 () == 12 ? null : typeof (T1));
```
vitek-karas: This test sort of breaks the test above. The tests rely on the `T1` being referenced only from one place - and the validation then checks that linker found the dependency.
So by reusing the type here, it's hard to tell if all the cases work, since if only one of them does, the others won't be validated.

So you need to create a new type for each test case.

Same logic applies to methods/fields/properties in the other tests.

test/Mono.Linker.Tests.Cases/Reflection/ExpressionNewType.cs
```diff
@@ -20,8 +42,38 @@ class T1
...
+
+		[Kept]
+		static void Branch_SystemTypeValueNode ()
+		{
+			var expr = Expression.New (typeof (T1));
+		}
+
+		[Kept]
+		static void Branch_NullValueNode ()
+		{
```
vitek-karas: Similar comment as above - add a case where there are two potential values for the type - you kind of have it here by having null and a type, but maybe another one with two existing types (again, they all have to be only used exactly once for the validation to work).

test/Mono.Linker.Tests.Cases/Reflection/ExpressionPropertyString.cs
```diff
@@ -32,6 +35,58 @@ class Foo
...
+			[Kept]
+			private string UnknownString ()
+			{
+				return "unknownstring";
+			}
+
+			[Kept]
+			public void Branch_SystemTypeValueNode_KnownStringValue_NonStatic ()
+			{
+				var expr = Expression.Property (Expression.Parameter (typeof (int), "somename"), typeof (Foo), "TestName1");
```
vitek-karas: Same comment as above - have a test for multiple type values and multiple name values.
And also the uniqueness problem with validation.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3407,6 +3279,106 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+										reflectionContext.RecordUnrecognizedPattern ($"Expression call '{calledMethod.FullName}' inside '{callingMethodBody.Method.FullName}' was detected with 2nd argument which cannot be analyzed");
+									}
+								}
+							}
+							break;
+
+						//
+						// static New (Type)
+						//
+						case "New" when calledMethod.DeclaringType.Name == "Expression"
```
MichalStrehovsky: This can be annotated by placing `[DynamicallyAccessedMembers(DefaultConstructor)]` on the parameter - we don't need to special case in the linker.

vitek-karas: For now you can emulate that by calling something like this:
```C#
RequireDynamicallyAccessedMembers (ref reflectionContext, DynamicallyAccessedMemberKind.DefaultConstructor, methodParameters[0]);
```

That will do basically the same thing as adding the attribute would. And we can later add the attribute to the framework for real (which we should for all the intrinsic methods anyway).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3607,6 +3476,97 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+							&& calledMethod.DeclaringType.Namespace == "System.Linq.Expressions"
+							&& calledMethod.Parameters.Count == 1
+							&& calledMethod.Parameters[0].ParameterType.FullName == "System.Type":
+							{
+
+								reflectionContext.AnalyzingPattern();
+
+								foreach (var value in methodParams [0].UniqueValues ()) {
+									if (value is SystemTypeValue systemTypeValue)
+										MarkMethodsFromReflectionCall (ref reflectionContext, systemTypeValue.TypeRepresented, ".ctor", BindingFlags.Instance, parametersCount: 0);
```
vitek-karas, line 3563: Technically this should include Public | NonPublic - since the runtime version will handle both cases.

PR #1040, opened 2020-03-28
Remove F# compilation resources

### LINKED ISSUE #329: Automagically remove F# compilation resources
The linker can/should automatically remove any .NET manifest assembly resources named `FSharpSignatureData.*` and `FSharpOptimizationData.*` during linking.  

These are only ever needed for F# compilation which doesn't happen on-device and given the aims of the linker it is reasonable to assume this data is unreachable.  

My understanding is that removing these would trim about 1.5MB off the size of all applications using FSharp.Core.dll, and about 3MB off all applications using Elmish.XamarinForms.dll.

Here's an example of how the resources look in ILDASM:
```
    .mresource public FSharpSignatureData.Elmish.XamarinForms
    .mresource public FSharpOptimizationData.Elmish.XamarinForms
```

Thanks

heytherewill:
Closes #329

A few considerations for reviewers:
- The code might not be placed at the best possible step. I've added it to `RemoveUnreachableBlockStep`. I've also considered adding an entirely new step, but that sounded like an overkill.
- I will add tests to cover the newly added code once it's settled that the code is in the correct spot.


mrvoorhe:
I think https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/RemoveFeaturesStep.cs would be better. But what and see what @marek-safar thinks

marek-safar:
I'd add a new step called `RemoveResources` after MarkStep instead. Any ideas about shorter `--keep-fsharp-compilation-resources"` name ?

heytherewill:
@mrvoorhe I considered that step, but it only runs if you explicitly tell the linker that you wanna remove a feature. Adding another option there that ran unconditionally and needed a flag that indicates `keep` rather than `remove` felt very counter intuitive to me 🤔 

I think a new step indeed is what makes the most sense

Regarding naming: `--keep-fsharp-resources` might work, albeit not as descriptive

src/linker/Linker.Steps/RemoveResourcesStep.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Mono.Linker.Steps
+{
+    public class RemoveResourcesStep : IStep
```
marek-safar: Derive from BaseStep

src/linker/Linker.Steps/RemoveResourcesStep.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Mono.Linker.Steps
+{
+    public class RemoveResourcesStep : IStep 
+	{
+        public void Process(LinkContext context)
+        {
+            var assemblies = context.Annotations.GetAssemblies().ToArray();
```
marek-safar: ToArray should not be needed

src/linker/Linker.Steps/RemoveResourcesStep.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        private void RemoveFSharpCompilationResources(AssemblyDefinition assembly)
+        {
+            var resourcesInAssembly = assembly.MainModule.Resources.Select(r => r.Name);
+            foreach (var resource in resourcesInAssembly.Where(IsFSharpCompilationResource)) {
+                Annotations.AddResourceToRemove(assembly, resource);
+            }
+
+            static bool IsFSharpCompilationResource(string resourceName)
+                => resourceName.StartsWith("FSharpSignatureData")
```
marek-safar: Use ordinal comparison

src/linker/Linker/Driver.cs
```diff
@@ -289,6 +290,12 @@ public bool Run (ILogger customLogger = null)
 
 							continue;
 
+						case "--keep-fsharp-compilation-resources":
```
marek-safar:
```suggestion
						case "--keep-compilers-resources":
```

src/linker/Linker.Steps/RemoveResourcesStep.cs
```diff
@@ -0,0 +1,26 @@
...
+            var assemblies = context.Annotations.GetAssemblies().ToArray();
+
+            foreach (var assembly in assemblies) {
+                RemoveFSharpCompilationResources(assembly);
+            }
+        }
+
+        private void RemoveFSharpCompilationResources(AssemblyDefinition assembly)
+        {
+            var resourcesInAssembly = assembly.MainModule.Resources.Select(r => r.Name);
```
marek-safar: Check for `HasResources` first

src/linker/Linker.Steps/RemoveResourcesStep.cs
```diff
@@ -0,0 +1,27 @@
...
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class RemoveResourcesStep : BaseStep 
+	{
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			if (!assembly.MainModule.HasResources) return;
+			RemoveFSharpCompilationResources (assembly);
```
marek-safar, line 12: Please add a check which enables this only for assemblies which are linked or saved

src/linker/Linker/Driver.cs
```diff
@@ -941,6 +952,7 @@ static Pipeline GetStandardPipeline ()
 			p.AppendStep (new PreserveDependencyLookupStep ());
 			p.AppendStep (new TypeMapStep ());
 			p.AppendStep (new MarkStep ());
+			p.AppendStep (new RemoveResourcesStep());
```
marek-safar, line 955: Please update the comments above as well

heytherewill:
Boy, 2020 sure was a busy year 😅 

Anyways, I addressed the changes requested by this PR now that I found some time. Let me know if there are any further requests

marek-safar (approved):

marek-safar:
Thank you for the contribution!

PR #1046, opened 2020-03-30
Field access annotation propagation

vitek-karas:
Open question: How to represent the field access in current reflection pattern recording (or how to change that)

### DIFF
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -331,10 +331,6 @@ namespace Mono.Linker.Dataflow
 						PushUnknown (currentStack);
 						break;
 
-					case Code.Ldsfld:
-						ScanLdsfld (operation, currentStack, thisMethod);
-						break;
-
 					case Code.Ldarg:
 					case Code.Ldarg_0:
 					case Code.Ldarg_1:
@@ -435,6 +431,7 @@ namespace Mono.Linker.Dataflow
 						break;
 
 					case Code.Ldfld:
+					case Code.Ldsfld:
 						ScanLdfld (operation, currentStack, thisMethod, methodBody);
 						break;
 
@@ -465,11 +462,8 @@ namespace Mono.Linker.Dataflow
 						break;
 
 					case Code.Stfld:
-						ScanStfld (operation, currentStack, thisMethod, methodBody);
-						break;
-
 					case Code.Stsfld:
-						ScanStsfld (operation, currentStack, thisMethod, methodBody);
+						ScanStfld (operation, currentStack, thisMethod, methodBody);
 						break;
 
 					case Code.Cpobj:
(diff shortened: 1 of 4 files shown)


src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -695,6 +696,72 @@ private void ScanLdarg (Instruction operation, Stack<StackSlot> currentStack, Me
 			StoreMethodLocalValue (locals, valueToStore.Value, localDef, curBasicBlock);
 		}
 
+		protected abstract ValueNode GetFieldValue (MethodDefinition method, ValueNode instanceValue, FieldDefinition field);
+
+		private void ScanLdsfld (
```
MichalStrehovsky: Nit: ScanLdsfld/ScanLdfld are pretty much the same. I think we could just add a single `if` to `ScanLdfld` and have it handle both (same for the field stores).

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -385,7 +385,7 @@ public static bool DetectCycle (this ValueNode node, HashSet<ValueNode> seenNode
 
 				case ValueNodeKind.LoadField:
 					LoadFieldValue lfv = (LoadFieldValue)node;
-					foundCycle = lfv.InstanceValue.DetectCycle (seenNodes, allNodesSeen);
+					foundCycle = lfv.InstanceValue != null && lfv.InstanceValue.DetectCycle (seenNodes, allNodesSeen);
```
MichalStrehovsky: We could potentially also just delete the `InstanceValue` property - we don't model these so they will pretty much always be null. It's a leftover from .NET Native I probably should have deleted.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3370,6 +3370,25 @@ protected override ValueNode GetMethodParameterValue (MethodDefinition method, i
...
+				return new LoadFieldValue (instanceValue, field, memberKinds) {
+					SourceContext = method
+				};
+			}
+
+			protected override void HandleStoreField (MethodDefinition method, FieldDefinition field, Instruction operation, ValueNode instanceValue, ValueNode valueToStore)
+			{
+				var requiredMemberKinds = _flowAnnotations.GetFieldAnnotation (field);
+				if (requiredMemberKinds != 0) {
+					// TODO: There's no way to represent store to a field given current ReflectionPatternContext (and the underlying IReflectionPatterRecorder)
```
MichalStrehovsky: Maybe this can be handled the same as method return values? (Build a list of all assigned fields as we scan the body but process the list in `ScanAndProcessReturnValue`.)

vitek-karas: It could be, but that would require more allocations (the list of values and so on). I find the virtual method a bit cleaner personally.

MichalStrehovsky: Sounds good. Maybe when addressing this TODO, we could unify this with return value handling (the `ret` instruction sounds like "a `stfld` into a pretend return value field").

PR #1047, opened 2020-03-30
Constant property option

sbomer:
Initial implementation of https://github.com/dotnet/designs/pull/99 on the linker side. This adds an option `--set-property NAME BOOL` which will substitute the body of the specified static property so that it returns a constant boolean. The intended use makes this conceptually similar to `--exclude-feature`, but unlike that option, it makes it the framework's responsibility to define the behavior when a feature is explicitly enabled or disabled.

For now:
- The property name is the same as the feature name.
- This only supports bools
- This supports arbitrary static properties. We could also restrict the supported patterns to enforce that it's only used for static readonly backing fields, or to enforce visibility constraints.
- The backing field is left unchanged

We need to decide specifically which "feature property" patterns this should support. Do we want to enforce additional constraints on the feature-defining properties? Should we check that they are backed by readonly fields? (Or that they are backed by AppContext switches?)

### DIFF
--- a/src/linker/Linker.Steps/BodySubstituterStep.cs
+++ b/src/linker/Linker.Steps/BodySubstituterStep.cs
@@ -1,4 +1,5 @@
-﻿using System;
+using System;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Globalization;
@@ -12,16 +13,88 @@ namespace Mono.Linker.Steps
 		protected override void Process ()
 		{
 			var files = Context.Substitutions;
-			if (files == null)
+			if (files != null) {
+				foreach (var file in files) {
+					try {
+						ReadSubstitutionFile (GetSubstitutions (file));
+					} catch (Exception ex) when (!(ex is XmlResolutionException)) {
+						throw new XmlResolutionException ($"Failed to process XML substitution '{file}'", ex);
+					}
+				}
+			}
+
+			// Property substitutions passed on the command-line win over substitutions file
+			SubstituteConstantProperties ();
+		}
+
+		static PropertyDefinition FindProperty (TypeDefinition type, string propertyName)
+		{
+			if (!type.HasProperties)
+				return null;
+
+			foreach (PropertyDefinition prop in type.Properties)
+				if (propertyName == prop.Name)
+					return prop;
+
+			return null;
+		}
+
+		void SubstituteConstantProperties ()
+		{
+			if (Context.ConstantProperties == null)
 				return;
 
-			foreach (var file in files) {
-				try {
-					ReadSubstitutionFile (GetSubstitutions (file));
-				} catch (Exception ex) when (!(ex is XmlResolutionException)) {
-					throw new XmlResolutionException ($"Failed to process XML substitution '{file}'", ex);
+			foreach (var (propertyStr, propertyValue) in Context.ConstantProperties) {
+				Debug.Assert (!String.IsNullOrEmpty (propertyStr));
+
+				// Parse out type name and property name from the property string
+				var separatorIndex = propertyStr.LastIndexOf (Type.Delimiter);
+				var typeFullname = separatorIndex < 0 ? null : propertyStr.Remove (propertyStr.LastIndexOf (Type.Delimiter));
+				var propertyName = separatorIndex < 0 ? null : propertyStr.Substring (propertyStr.LastIndexOf (Type.Delimiter) + 1);
+				if (separatorIndex < 0 || String.IsNullOrEmpty (typeFullname) || String.IsNullOrEmpty (propertyName)) {
+					Context.LogMessage (MessageImportance.High, $"Invaliad property name '{propertyStr}'");
+					continue;
 				}
 
+				PropertyDefinition foundProperty = null;
+				foreach (var assembly in Context.GetAssemblies ()) {
+					// Only look in the MainModule, as we do elsewhere
+					var type = assembly.MainModule.GetType (typeFullname);
+					if (type == null)
+						continue;
+					var property = FindProperty (type, propertyName);
+					if (property == null)
+						continue;
+					if (foundProperty != null)
+						Context.LogMessage (MessageImportance.High, $"Property '{propertyStr}' exists in multiple assemblies");
+					foundProperty = property;
+
+					// Retrieve the get method and check it is static
+					var getMethod = foundProperty.GetMethod;
+					if (getMethod == null) {
+						Context.LogMessage (MessageImportance.High, $"Property '{propertyStr}' has no get method");
+						continue;
+					}
+					if (!getMethod.IsStatic) {
+						Context.LogMessage (MessageImportance.High, $"Property '{propertyStr}' is not static");
+						continue;
+					}
+					// Only support boolean properties
+					if (getMethod.ReturnType.MetadataType != MetadataType.Boolean) {
+						Context.LogMessage (MessageImportance.High, $"Property '{propertyStr}' is not a boolean property");
+						continue;
+					}
+
+					// TODO: should we enforce that the getter only reads from a readonly static field?
+					// Set the field value itself to a constant?
+
+					// Stub the getter to return a constant. The stub logic uses integers to represent bools.
+					Annotations.SetMethodStubValue (getMethod, propertyValue ? 1 : 0);
+					Annotations.SetAction (getMethod, MethodAction.ConvertToStub);
+				}
+
+				if (foundProperty == null)
+					Context.LogMessage (MessageImportance.High, $"Could not resolve property '{propertyStr}' for substitution");
 			}
 		}
 
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -319,6 +319,17 @@ namespace Mono.Linker {
 
 							continue;
 
+						case "--set-property":
+							if (!GetStringParam (token, prop => {
+								if (!GetBoolParam (token, value => {
+									context.AddConstantProperty (prop, value);
+								}))
+									return;
+							}))
+								return false;
+
+							continue;
+
 						case "--new-mvid":
 							//
 							// This is not same as --deterministic which calculates MVID
@@ -904,6 +915,7 @@ namespace Mono.Linker {
 #endif
 			Console.WriteLine ("                              sre: System.Reflection.Emit namespace");
 			Console.WriteLine ("                              globalization: Globalization data and globalization behavior");
+			Console.WriteLine ("  --set-property PROP BOOL  Rewrites the specified static property to return a constant value");
 			Console.WriteLine ("  --explicit-reflection     Adds to members never used through reflection DisablePrivateReflection attribute. Defaults to false");
 			Console.WriteLine ("  --keep-dep-attributes     Keep attributes used for manual dependency tracking. Defaults to false");
 			Console.WriteLine ("  --new-mvid                Generate a new guid for each linked assembly (short -g). Defaults to true");
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -120,6 +120,8 @@ namespace Mono.Linker {
 
 		public List<string> Substitutions { get; private set; }
 
+		public List<(string, bool)> ConstantProperties { get; private set; }
+
 		public List<PInvokeInfo> PInvokes { get; private set; }
 
 		public string PInvokesListFile;
@@ -209,6 +211,21 @@ namespace Mono.Linker {
 				CodeOptimizations.IPConstantPropagation;
 
 			Optimizations = new CodeOptimizationsSettings (defaultOptimizations);
+		}
+
+		public void AddConstantProperty (string property, bool value)
+		{
+			Debug.Assert (!String.IsNullOrEmpty (property));
+			var pair = (property, value);
+			if (ConstantProperties == null) {
+				ConstantProperties = new List<(string, bool)> { pair };
+				return;
+			}
+
+			if (ConstantProperties.Contains (pair))
+				return;
+
+			ConstantProperties.Add (pair);
 		}
 
 		public void AddSubstitutionFile (string file)
--- a/test/Mono.Linker.Tests.Cases/Substitutions/ConstantProperty.cs
+++ b/test/Mono.Linker.Tests.Cases/Substitutions/ConstantProperty.cs
@@ -1,0 +1,30 @@
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.Substitutions
+{
+	[SetupLinkerArgument ("--set-property", "Mono.Linker.Tests.Cases.Substitutions.ConstantProperty.BoolProperty", "true")]
+	public class ConstantProperty
+	{
+		[Kept]
+		static bool BoolProperty {
+			[Kept]
+			[ExpectedInstructionSequence (new [] {
+				"ldc.i4.1",
+				"ret"
+			})]
+			get;
+		}
+
+		public static void Main()
+		{
+			TestProperty_1 ();
+		}
+
+		[Kept]
+		static bool TestProperty_1 ()
+		{
+			return BoolProperty;
+		}
+	}
+}


mrvoorhe:
Is this essentially a command line version of the substitutions file?

sbomer:
Yup, exactly. This will avoid the need to generate an xml input from our MSBuild task and check that it is up-to-date, etc.

MichalStrehovsky:
> I think a much better approach would be to implement this TODO

The end to end expectation is that feature switches control both the linker and get surfaced as AppContext values and we can control them from MSBuild. It's a publish-time configuration that we dynamically build for each publish. Placing it in an XML file means that we need to create the XML during publish and throw it away the moment publish is done.

AppContext values already flow as MSBuild `Items` within the SDK (and get preserved to a JSON file that is a build output).

I expect feature switches to also flow as MSBuild `Items` within the SDK that will somehow get connected to the existing AppContext handling.

Now, once we have feature switches as MSBuild items, we could just dump them to the RSP file that we _already generate_ for the linker, or create an XML file, make sure it's up to date and put a reference to _that file_ in RSP. It feels like avoiding XML file just makes things easier. We don't have to worry about the annoyances like a virus scanner locking a newly generated file so that it can scan it.

marek-safar:
>It's a publish-time configuration that we dynamically build for each publish.

It's not, the logic is known at compilation time we don't have to re-generate anything. I made tracking item #1048 to make this even more obvious.

MichalStrehovsky:
> `<linker feature="globalization">`

I see - it's adding an indirection by naming the thing - this matches with what I was proposing in separate email threads - but I wanted to get an Fxdc-approved custom attribute that we can slap on the property getter so that people don't have to maintain XML files because that's a pain. Something like:

```csharp
bool Invariant
{
    [Feature("System.Globalization.Invariant")]
    get
    {
       // ...
    }
}

**Comment by marek-safar** [comment]
I agree even the attributes would be better than hacking this in msbuild but I'm not a big fan of attributes for this. They are harder to comprehend for non-trivial features like invariant globalization, think of removing fully ref-emit or event source support. We could slap the attribute blindly everywhere for such features but that to me is too much magic.

**Comment by vitek-karas** [comment]
To me there are two sides to this which go against each other:
* We want the feature switch to be easy to explain and adapt to - that typically means there should be one IL property (possibly even public) which is the authoritative source of the on/off state. See this comment for why this might be important: https://github.com/dotnet/designs/pull/99#discussion_r399610617
* Potential need for a complex set of rules of what to remove where, to get rid of the feature (basically a complex piece of substitution.xml)

To be honest I still don't see the need for the complex set of rules. Assuming we can change CoreFX (which we can), then we should be able to change the code such that the entire feature is driven by a single property. In that case the need for complex mapping goes away. The attribute approach in this case would basically mean adding one attribute per feature.

Note that neither approach will solve all the scenarios - for example if we want to fully remove EventSource including all the custom event sources and call sites any kind of substitution.xml will not do the job. We would need additional linker smarts (basically ability to remove callsites to methods which are provably no-ops).

/cc @eerhardt

**Comment by marek-safar** [comment]
> the entire feature is driven by a single property

I'm fine to consider only that case (or hope for it) but even with that embedding the logic about features mapping to all properties somewhere deep inside msbuild targets is just not attractive to me.

**Comment by vitek-karas** [comment]
I do agree that ideally the mapping should not be in MSBuild - so either the XML or attribute. I personally like the attribute a bit better.

**Comment by sbomer** [comment]
I consider the problem of mapping MSBuild property -> feature name solved by the mechanism described in https://github.com/dotnet/designs/pull/99#discussion_r395873121.

It sounds like there is agreement, at least for now, on
> the entire feature is driven by a single property

(an IL property specifically). Working with that understanding, we don't need 1:many mappings of feature name -> IL properties. There is a 1:1 mapping from feature name <-> IL property.

The remaining question is how to uniquely identify an IL property from a feature name. I don't see a good reason not to unify the feature name with the property name. Why use multiple names if the mapping is 1:1? That just leaves the ambiguity in case a property with the same name is defined in multiple assemblies.

Could we just make it a linker error to define that property in multiple assemblies? How common do we think it is for people to define, for example, their own `System.Globalization.GlobalizationMode.Invariant`?

Alternatively what do you think of something like

```csharp
bool Invariant
{
    [FeatureDefinition]
    get {}
}
```
which would indicate that a property is intended to represent a feature, without creating an extra mapping from feature name <-> property name? This way we could support cases where other assemblies define a property with the same name but different semantics, as long as they don't also attribute it with `FeatureDefinition`.

A final option, which would be fragile if we move types between assemblies, is to include the assembly name in the feature name - so something like `System.Private.CoreLib:System.Globalization.GlobalizationMode.Invariant`. I'm not personally a fan of this one.

eerhardt:
> Could we just make it a linker error to define that property in multiple assemblies?

Typically when sharing internal/helper code, CoreFX libraries will compile the same source into multiple IL assemblies. This way the shared code doesn't need to be public, and we don't need to use `InternalsVisibleTo` between the libraries.

Take for example, https://github.com/dotnet/runtime/blob/master/src/libraries/Common/src/System/Net/Logging/NetEventSource.Common.cs. It is included in quite a few assemblies: https://github.com/dotnet/runtime/search?l=XML&q=NetEventSource.Common.

If we were to define a feature switch that spanned multiple assemblies, the property could end up in multiple assemblies, or we would be forced to make the property public and the public property would have to be defined in a low enough assembly that all the consuming assemblies referenced it.

marek-safar:
> I consider the problem of mapping MSBuild property -> feature name solved by the mechanism described in dotnet/designs#99 (comment).

Firstly, the design talks about custom attribute which this PR completely ignores and tries to hack into metadata directly. There might be an additional mapping between user-facing name and internal feature name which could be in msbuild. 

I won't comment on the whole .runtimeconfig.json approach as that sounds like no-go solution for workloads which don't use app-host and is not really relevant here.

> Could we just make it a linker error to define that property in multiple assemblies?

That's the problem. Anyone can break our tooling with no workaround if someone introduces property with the same name or even worse we won't be able to add any property later because won't be sure there is no type with the same name if we don't fully qualify the assembly.

>Working with that understanding, we don't need 1:many mappings of feature name -> IL properties. There is a 1:1 mapping from feature name <-> IL property.

Are you suggesting we will expose for example all intrinsic IsSupported properties as individual features? So every SDK would have to set about 10 different "features" to remove intrinsic and when we add more they will have to update their SDK integration?

>How common do we think it is for people to define, for example, their own System.Globalization.GlobalizationMode.Invariant

Please don't model this enhancement based on a single property.

>Alternatively what do you think of something like

This looks better but I am concerned about limiting this to bool type only or assuming true means the feature removal should kick in.

What will the attribute syntax look like to support for example this property https://github.com/dotnet/runtime/blob/7302fa490d4da1b2b806dfeb3d778636f9cd0121/src/mono/netcore/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeFeature.Mono.cs#L9

MichalStrehovsky:
The slight problem I see with exposing property name as the switch name is that the property name is an internal implementation detail of the framework, not a public API. People could go and rename/refactor the namespace/type/property, breaking this. Hopefully we would have enough test coverage to detect when this happens, but the usual tools for this (apicompat) won't detect it because the property is not actually a public API.

I see the advantage that using the property name as the feature name avoids having to hardcode an extra string, but when we look at this E2E (where we would like to also allow emulating this through `AppContext`), the extra string is unavoidable because we need to pass something to `AppContext`.

My preference would be a solution that reduces the number of times we need to duplicate the string. That's why I kind of lean towards the custom attribute, as opposed to the XML file, because we can have everything in a single place:

```csharp
const string FeatureSwitchName = "System.Globalization.Invariant";

bool Invariant { [FeatureSwitch(FeatureSwitchName)] get; } = AppContext.TryGetSwitch(FeatureSwitchName);
```

sbomer:
Thanks all for the comments. It sounds like we have a few good reasons not to select IL properties by name alone, so we require an extra mapping from feature names to specific IL properties. I don't yet see an agreement on where that mapping goes, with an expressed preference for attributes.

>> I consider the problem of mapping MSBuild property -> feature name solved by the mechanism described in dotnet/designs#99 (comment).
>
> Firstly, the design talks about custom attribute which this PR completely ignores and tries to hack into metadata directly.

I think we are talking about different things here. Whether or not we use attributes for the feature names, the MSBuild property -> feature name mapping is a solved problem, no?

I'll try to clarify by using these specific terms:
- **MSBuild feature property**: a user-facing MSBuild property like `InvariantGlobalization` that represents a feature switch
- **Appcontext string**: string in the AppContext store, populated at startup by runtimeconfig.json
- **Feature name**: string that uniquely identifies a feature
- **IL property name**: fully qualified name of the IL property (including namespace but *not* assembly)
- **IL property**: a specific *unique* (assembly-qualified) IL property definition

Here is my understanding of the proposal https://github.com/dotnet/designs/pull/99:
1. Public *feature* switches:
  - User-facing **MSBuild feature property**, maps to a specific
  - User-facing **appcontext string** == **feature name**, maps to a specific
  - **IL property name** (I think we have agreed that there is just one per **feature name**)
  - The *intended* use is for an **IL property** backed by an **appcontext string** using `TryGetSwitch`. Some cases like globalization may not strictly follow this rule (by reading the runtimeconfig setting directly early during startup), but the semantics are the same. We don't enforce this except by code review, and we don't do anything to detect a mismatch if somebody sets the **appcontext string** using `SetSwitch` or the runtimeconfig to a value different from what is baked in.

Inferring a bit from what you wrote (I could be off - please correct me), it sounds like you have in mind something more like:

2. Hidden *platform-specific linker optimization* switches:
  - Target platform information from MSBuild, maps to
  - Non-user-facing **feature name**, maps to
  - Some set of IL, which we haven't decided on. Perhaps single or multiple **IL properties**, or even arbitrary methods specified in **substitutions.xml** or by attributes.
  - There is no corresponding **appcontext string**.

If you are talking about 2., I can understand this point
> There might be an additional mapping between user-facing name and internal feature name which could be in msbuild.

but I don't think it applies to scenario 1. It might help if you could give an example of a case where we would use such a mapping.

> Are you suggesting we will expose for example all intrinsic IsSupported properties as individual features?

I was not intending for this to be used for intrinsics, and I am not suggesting that we use such an attribute for `IsDynamicCodeSupported`. I think this needs to be clarified in the design - I have asked about it here: https://github.com/dotnet/designs/pull/99#discussion_r401802571. Based on discussion with @vitek-karas, I think the idea is for intrinsics to be handled more along the lines of 2, but the design is only addressing scenario 1.

> I am concerned about limiting this to bool type only

This was just an initial simplification to get something out - I'm happy to add support for integers when we agree on the direction.

>  or assuming true means the feature removal should kick in.

This doesn't make assumptions about which branch gets removed or what is the default value of the switch - it's up to the framework to pick a meaningful **IL property name** and default. If we say that the default value is by design always `false`, I can see the use of an extra indirection in the form of **feature name** - which could have the opposite polarity of the **IL property name** if we think that's desirable.

>  the property name is an internal implementation detail of the framework, not a public API

The **appcontext string** is already public. Should we consider supporting only features that use public **IL properties**? The use case for internal **IL properties** seems to be for platform-specific implementation details which we would not expose as a feature switch anyway, at least based on my current understanding. On the other hand I can see that changing IL is a harder breaking change to make than changing the behavior of (or adding/removing) **appcontext strings**, so having the flexibility to make the **IL properties** internal might be nice.

> the extra string is unavoidable because we need to pass something to AppContext.

This is a good point I hadn't considered. I personally think it's more important to minimize the number of "concepts" involved for our devs - particularly if we agree that this is just to be used for the public-facing feature switches. If we are intending for the same attribute to be used for the platform-specific linker optimizations, then the **feature name** is more of an implementation detail.

What I would like to see is a linker command-line option that takes **feature name** and bakes in values for the corresponding **IL properties**. I would ideally like that mapping to be carried with the code itself, so I would prefer attributes over embedded XML. I don't think we need to solve the more complex problem of platform-specific optimizations (more complex because they have many more switches, and also deal with things that aren't properties like `sizeof(IntPtr)`) in order to get the feature switches working. I'd like to clarify the goals (1. vs 2. or something in-between?), and then if we agree I can put together a change that uses attributes instead. Thoughts?

marek-safar:
@sbomer good summary!

I don't see Public feature switches and Hidden platform switches are orthogonal things. To me, the first one is more like an extension of the second one. The choice is not between 1. or 2. but how to support both with the least confusion to developers (and that includes SDK developers).

If someone contributes feature description to disable legacy TLS providers does it mean we have to always introduce a new property even if we already have `SslClientAuthenticationOptions.EnabledSslProtocols` as public API ?

sbomer:
How to support both is a great question. In my mind, the public switches are a simplified case that are worth giving an easier linker input, even if it uses the same code rewriting technology underneath. To further expand on the kinds of switches, I'm thinking in these terms:
1. Public switches
2. Hidden switches
  2.1. Constants that we bake into platform-specific library builds coming out of `dotnet/runtime`
  2.2. Constants that we bake into libraries on the developer machine at publish time, depending on their target platform - without an explicit MSBuild property to turn the switch on/off.

For 2.1, we at least have direct control of the baked-in behavior in our build, and the assemblies would go through all of the testing set up in our pipeline. 2.2 is the scarier case - and the one where we would need non-user-facing MSBuild logic to map to hidden "features" in the linker.

@marek-safar do you have an example where we would use 2.2 instead of 2.1?

Regarding `SslClientAuthenticationOptions.EnabledSslProtocols`:
I'm not familiar with that API, but based on some investigation it looks like that property is an enum that gets passed to the platform when requesting credentials. It's a public instance property, so to perhaps the most straightforward way to disable the unsupported protocols would be to rewrite it to:
```csharp
get { return EnabledProtocolsConstant; }
set { throw new NotSupportedException ("Linked away"); }
```
This has the problem that user code (or third-party dependencies) which use the setter will not be equipped to deal with the NSE, leading to potentially weird behavior. To make the linker "safe by construction" we would need an analyzer that reports calls to the setter at link time. Any non-readonly property would have this problem.

If that's too large of a breaking change, we might instead rewrite it to something like:
```csharp
SslProtocols _enabledProtocols = EnabledProtocolsConstant;
SslProtocols EnabledProtocols {
    get { return _enabledProtocols; }
    set {
        if ((value & EnabledProtocolsConstant) == 0) {
            throw new NotSupportedException ("Unsupported protocol");
        } else {
            _enabledProtocols = value;
        }
    }
}
```
which is code that should be owned by the frameworks. We could use platform-specific C# files if we only want to disable legacy protocols on some platforms, or we could make the change unconditionally in the framework, if the expected impact of the breaking change is small enough. Most likely though, I think we would provide a compat switch, something like:
```csharp
set {
    if ((value & EnabledProtocolsConstant) == 0 && !SupportLegacyTLS) {
      throw new NotSupportedException ("Unsupported protocol");
    } ...
}
```
where `System.Net.Security.SslClientAuthenticationOptions.SupportLegacyTLS` (or a property that contains the names of the specific legacy protocols) is a feature switch backed by AppContext.

So in the end I think it's likely that we would want to introduce a new property, to limit the scope of breaking changes. However that's just my speculation, and this is the kind of decision that should be made by a framework author. @eerhardt I'm curious about your thoughts on this example. The hope is that feature switches would be the right pattern in most cases, and would enable linker support for those patterns.

marek-safar:
> do you have an example where we would use 2.2 instead of 2.1?

I don't know if that's a perfect example but you could consider the following. 

For iOS there is existing API https://github.com/xamarin/xamarin-macios/blob/a6b03cd10eb5fc56b00f4999a111cfb242ed4b70/src/ObjCRuntime/Runtime.iOS.cs#L46 which is used to check if the code is running on device or simulator. The final app build would like to override the field comparison to remove the unnecessary dependencies when building for the device.

> Regarding SslClientAuthenticationOptions.EnabledSslProtocols:

What I tried to illustrate is not that nobody can set the property but that for example of Android platform the build could force the property to always return e.g. TLS 1.2 because we know even the lowest supported Android version has it and reduce the dependencies. Yes, we can change every library to use a new bool based API but I am not sure that's what we want.

sbomer:
Thank you for the pointer to `Runtime.Arch` - it helps me better understand your angle on this (at least I hope so 😃).

> the build could force the property to always return e.g. TLS 1.2 because we know even the lowest supported Android version has it and reduce the dependencies

I think we are on the same page. Ignoring the setter for a moment, we would want the getter to look like this on Android:
```csharp
public SslProtocols EnabledSslProtocols { get { return SslProtocols.Tls12; } set { ... } }
```
and like this on other platforms:
```csharp
public SslProtocols EnabledSslProtocols { get { return _enabledSslProtocols; } set { ... } }
```
right? So we could write it like this if we know the target platform at compile time:
```csharp
public SslProtocols EnabledSslProtocols {
    get {
#if ANDROID
        return SslProtocols.Tls12;
#else
        return _enabledSslProtocols;
#endif
    }
    set { ... }
}
```
which wouldn't require a feature switch or special linker support. Or, if we don't know the platform at compile-time, it would look more like:
```csharp
public SslProtocols EnabledSslProtocols {
    get {
        if (Statics.RunningOnAndroid)
            return SslProtocols.Tls12;
        else
            return _enabledSslProtocols;
    }
    set { ... }
}
```
where we would want the linker to be able to keep only the relevant branch. The setter would similarly have different behavior depending on that property, which lets framework authors decide on the per-platform behavior like what to do if an unsupported protocol is requested by the setter.

Readonly static properties seem like a good way to this, and this pattern lets framework authors own the code (in C#, not xml) for both branches. Libraries would hopefully be able to keep using the existing APIs - though they might have to adjust to the new behavior of different platforms. Framework authors would decide whether the new property is public API.

For platform-specific behavior (like `RunningOnAndroid`), we would bake in the constant either in the libraries we ship or on the developer's machine. There would be some MSBuild logic to determine the constants per platform. Or if we made it a "feature switch"  like `DisableLegacyTLSProtocols` backed by AppContext, we would have the option to bake it in (when linking), or just set it in runtimeconfig.json (when not linking). The "feature switch" pattern would likely be used for backcompat switches when we make breaking changes in the framework, or for disabling feature areas.

Either way, we would pass (**feature name**, value) to the linker, which would rewrite the corresponding **IL properties** (determined by some kind of `FeatureAttribute`) to that value. It wouldn't cover all of the platform-specific things like `sizeof(IntPtr)`, but I think it provides a simple behavior for a very common case (that should be useful for "feature switches" and many of the "hidden switches" as well).

About `Runtime.Arch`:

It looks like it behaves very similarly to what I described above, with no appcontext string. It's not readonly, but it looks like it is only assigned once during initialization. From what I can tell, it gets inlined by mtouch as a special case for iOS (please correct me if I'm wrong). It seems to me that we could unify on the same pattern by using a static property backed by a field that gets written during `InitializePlatform`, where the linker rewrites that property to return a constant and eliminates the dead branches.

It looks like the Xamarin SDK has some logic that passes `IsSimulator` and other platform info into mtouch. I assume that with the SDK unification, we will similarly map .NET SDK MSBuild inputs to the mtouch/linker arguments for this and similar platform-specific properties.

eerhardt:
The `SslClientAuthenticationOptions.EnabledSslProtocols` is probably a bad example for a few reasons:

1. It is an instance property, so an application could use different values for different connections. (And they might need to since different services could require different protocols.)
2. I don't think we'd ever want to lock it into one protocol like `Tls12` because there is `Tls13` and future versions as well.
3. This scenario doesn't help us make applications smaller in reality because the SslProtocols are implemented in external libraries (Windows SCHANNEL, OpenSsl, Apple crypto). So there isn't a lot of IL code that could be trimmed away.

If size isn't the concern here, then I don't see why the `System.Net.Security` library wouldn't do the work itself. I wouldn't put that kind of logic into the linker.

PR #1049, opened 2020-03-31
Open up ReflectionPatternRecorder more

mrvoorhe:
We implement detection of methods such as IsAssignableFrom which requires marking of an InterfaceImplementation.  An InterfaceImplementation is not an IMemberDefinition so I was unable to record this reflection pattern.

Changing RecognizedReflectionAccessPattern to accept an IMetadataTokenProvider allows us to record this detection.

### DIFF
--- a/src/linker/Linker/IReflectionPatternRecorder.cs
+++ b/src/linker/Linker/IReflectionPatternRecorder.cs
@@ -46,8 +46,8 @@ namespace Mono.Linker
 		/// <param name="sourceMethod">The method which contains the reflection access pattern.</param>
 		/// <param name="reflectionMethod">The reflection method which is at the heart of the access pattern.</param>
 		/// <param name="accessedItem">The item accessed through reflection. This can be one of:
-		///   TypeDefinition, MethodDefinition, PropertyDefinition, FieldDefinition, EventDefinition.</param>
-		void RecognizedReflectionAccessPattern (MethodDefinition sourceMethod, MethodDefinition reflectionMethod, IMemberDefinition accessedItem);
+		///   TypeDefinition, MethodDefinition, PropertyDefinition, FieldDefinition, EventDefinition, InterfaceImplementation.</param>
+		void RecognizedReflectionAccessPattern (MethodDefinition sourceMethod, MethodDefinition reflectionMethod, IMetadataTokenProvider accessedItem);
 
 		/// <summary>
 		/// Called when the linker detected a reflection access but was not able to recognize the entire pattern.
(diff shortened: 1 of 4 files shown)


mrvoorhe:
@vitek-karas 

> What are you using the Recognized pattern reporting for?

The short answer is, very few tests so far, but I'm glad it was added.  It's a useful addition because we are actively working on reporting unrecognized reflection usages to our users. 

When it comes to recognized reflection usages, yes that is probably not as important but I think it's still worth having.  For example, maybe we want to hook up analytics some today to get an idea of how helpful auto handling is.

> In the data flow work we basically don't see it as useful anymore - so if we don't really need it, we could remove it.

What don't you see as being useful?  RecordRecognizedPattern? or RecordUnrecognizedPattern?  Or both?

mrvoorhe:
> Also the data flow work will probably require some more changes to this interface - see #1053

`MethodReturnType` implements `IMetadataTokenProvider` so this change would cover that use case wouldn't it?

vitek-karas:
We don't have a use case for Recognized pattern. We absolutely will have Unrecognized pattern usage.
I'm fine keeping the Recognized pattern reporting as well. The one thing which will be problematic is the current guarantees the interface makes - with the data flow it makes somewhat less sense to guarantee that for each reflection call there will be either recognized or unrecognized reported. Would it be a problem for you if we removed that guarantee? For the typical reflection calls we would probably still keep that behavior, but it would be possible to get calls through the interface as well which would not map nicely to reflection calls.

As for the changes to the interface, we won't be able to assign a reflection method to each reported action. So for example the unrecognized call would probably have to change to something like:
```C#
void UnrecognizedReflectionAccessPattern (MethodDefinition sourceMethod, IMetadataTokenProvider target, string message)
```

For example if there's an assignment to a field in a method and the field requires more than the value assigned to it can guarantee we would call `UnrecognizedReflectionAccessPattern(method, field, "message")`.

mrvoorhe:
>   with the data flow it makes somewhat less sense to guarantee that for each reflection call there will be either recognized or unrecognized reported. Would it be a problem for you if we removed that guarantee?

I don't think so.

I don't fully have my head around the data flow stuff and how that should relate to Unity users.  It's an area I need to think about more.  I think a lot of what Unity and Unity Users are doing falls under what you guys are viewing as "should be handled by source generators".

Prior to the data flow stuff, even before some of the other reflection handling improvements recently, we had our own reflection detection logic, our own reflection recording.  At the moment, we are going down an overlapping parallel track.  We are filling out our reflection detection logic to identify and report all reflection calls as either unrecognized or recognized.

I wish we were not going to parallel paths at the same time, but we need to make reflection usages visible to our users sooner than later.

So in summary, it's not a big deal if you change this interface.  It's good to know you might, we'll try not to bind ourselves to it too much.  And at some point I may try and unify what we are doing with reflection and the dataflow branch, but now is not the time since I don't really know what that would even look like

mrvoorhe:
@marek-safar can you merge?

PR #1057, opened 2020-04-02
Expose more options in ILLink.Tasks

sbomer:
Part of https://github.com/dotnet/sdk/issues/11006.

This exposes the options `LinkSymbols`, `DefaultAction`, and `CustomSteps` on the MSBuild task. I've also added unit tests for these options and made a few fixes for little things I uncovered.

I have changed Driver to make it easier to test:
- allow setting up the context without running
- provide extension points to avoid reading/writing files

Non-whitespace diff is helpful for Driver.cs: https://github.com/mono/linker/pull/1057/files?w=1#diff-cb30baaa9eeed4528bdba667f5cd2aaf


src/ILLink.Tasks/LinkTask.cs
```diff
@@ -190,6 +222,35 @@ protected override string GenerateResponseFileCommands ()
 				}
 			}
 
+			if (LinkSymbols)
+				args.AppendLine ("-b");
+
+			if (DefaultAction != null)
```
marek-safar, line 228: I think this should be always explicitly set

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -49,17 +52,20 @@ public class ILLink : ToolTask
...
 		///   documentation for details about the format.
+		///   Maps to '-x'.
 		/// </summary>
 		public ITaskItem [] RootDescriptorFiles { get; set; }
 
 		/// <summary>
 		///   Boolean specifying whether to clear initlocals flag on methods.
+		///   Maps to '--enable-opt clearinitlocals'.
 		/// </summary>
 		public bool ClearInitLocals { get; set; }
```
marek-safar, line 63: Perhaps we should re-think about this and do all optimizations consistently not just this one

src/linker/Linker/Driver.cs
```diff
@@ -161,437 +163,453 @@ static void ErrorMissingArgument (string optionName)
 			Console.WriteLine ($"Missing argument for '{optionName}' option");
 		}
 
-		public bool Run (ILogger customLogger = null)
```
marek-safar, line 164: Could you keep the code as it and instead in the mock type override GetDefaultContext to extract the context instance?

sbomer: I don't think that works. I want to get the context after command-line arguments have been processed, but without trying to resolve assemblies/xml files from disk. So I need to separate the argument processing into something I can run separately from processing the steps.

Originally I think I misunderstood your suggestion - I thought you meant that GetDefaultContext could be used to inject mocked things into the context. So I tried to set up a pipeline with mocked steps that would do nothing when processed. That just created more problems though, requiring more extension points in LinkContext, and more machinery to get information out of the steps after `Run` finishes - especially because `Pipeline.Process` removes steps when it runs.

src/ILLink.Tasks/LinkTask.cs
sbomer: Two options come to mind:
We could expose a bool for each optimization, with a default value of `true` for the default optimizations (I would add a test to check that these stay in sync). It would be used like this:
```xml
<ILLink ClearInitLocals="true" BeforeFieldInit="false" ... />
```
This would be my recommendation - to me it feels more like the "msbuild" way to pass parameters to tasks.

Another option would be to specify enabled/disabled optimizations as a semicolon-separated string (or maybe as separate strings in an ItemGroup), used like this:
```xml
<ILLink EnableOptimizations="clearinitlocals;sealer" DisableOptimizations="beforefieldinit;ipconstprop" ... />
```
which has the advantage that we don't need to add more task parameters for new optimizations, but doesn't restrict the strings that can be passed in to the task, which IMO is an unusual way to expose feature-level functionality. This kind of pattern is more often used for things like `DefineConstants` or inputs that take lists of arbitrary file paths.

src/ILLink.Tasks/LinkTask.cs
sbomer: I plan to always pass in a DefaultAction from the SDK. I don't want to mark it `Required` because that seems like an unnecessary breaking change, and I think it would be good for the task defaults to match the command-line defaults (which are `UserAction`: link, and `CoreAction`: skip).

src/ILLink.Tasks/LinkTask.cs
marek-safar:
```
<ILLink ClearInitLocals="true" BeforeFieldInit="false" ... />
```

sounds good to me but where do you specify the assemblies the optimizations will apply to? Allowing SDKs to enable e.g. sealer for their own assembly only is something we need to support

src/linker/Linker/Driver.cs
marek-safar: The context is a reference instance you can store it and when referenced later it will give you the right data

src/ILLink.Tasks/LinkTask.cs
marek-safar: I don't see a lot of value in setting defaults to values which everyone will have to override.

sbomer: Does it make sense for the task defaults to be different than the command-line defaults? (If everyone has to override them, could we change the command-line defaults to something reasonable?)

marek-safar: it'd make sense to me to change the defaults for illink build

src/ILLink.Tasks/LinkTask.cs
sbomer: For per-assembly settings, we can choose:
1. A per-optimization input (ItemGroup of assembly names)
```xml
<ItemGroup>
   <_ClearInitLocalsAssemblies Include="Asm1;Asm2" />
</ItemGroup>
<ILLink ClearInitLocalsAssemblies="@(_ClearInitLocalsAssemblies)" />
```
2. A single list of assemblies with metadata in the form of strings
(here I'm combining it with the `AssemblyPaths` input, and the task would just take out the filename without extension to use it as the assembly name - or we could use a separate ItemGroup input):
```xml
<ItemGroup>
  <_AssemblyPaths Include="path/to/Asm1.dll">
    <enabledoptimizations>clearinitlocals;...</enabledoptimizations>
    <disabledoptimizations>beforefieldinit;...</disabledoptimizations>
  </_AssemblyPaths>
</ItemGroup>
<ILLink AssemblyPaths="@(_AssemblyPaths)" />
```
3. A single list while avoiding strings:
```xml
<ItemGroup>
  <_AssemblyPaths Include="path/to/Asm1.dll">
    <clearinitlocals>true</clearinitlocals>
    <beforefieldinit>false</beforefieldinit>
  </_AssemblyPaths>
</ItemGroup>
<ILLink AssemblyPaths="@(_AssemblyPaths)" />
```
I think this also allows the alternate syntax:
```xml
<ItemGroup>
  <_AssemblyPaths Include="path/to/Asm1.dll" ClearInitLocals="true" />
</ItemGroup>
<ILLink AssemblyPaths="@(_AssemblyPaths)" />
```

My preference would be 3, since SDK components already need to ensure that they pick the right AssemblyPaths, so metadata on that input seems like a logical place for it to go. Using separate boolean metadata options seems consistent with the booleans for global optimizations.

(as an aside, MSBuild is case-insensitive, but I'm going to switch the doc strings to use Pascal casing for the metadata since that seems to be common practice).

I can't think of a good way to combine the global switches with the per-assembly optimizations like we do on the command-line.

sbomer: I implemented option 3 - please let me know if you prefer something else.

src/ILLink.Tasks/LinkTask.cs
sbomer: Doing this in https://github.com/mono/linker/pull/1076

src/ILLink.Tasks/LinkTask.cs
marek-safar: Options 3 looks fine but I'm not sure how this will be used in practice when the assemblies full paths are resolved in other targets.

I'd also expect we use the same syntax for all other per-assembly settings probably only `-p ACTION ASM` for now, right?

sbomer: That depends:
- If each SDK component owns a different version of the assembly resolution, it could set the optimizations at the same time, depending on whether the assembly was resolved from a runtime pack, nuget package, etc...
- If the assembly resolution from the core SDK is shared by all components, but each SDK component may have different optimizations, then the existing SDK ItemGroups like `ResolvedFileToPublish` can be given metadata for the optimizations. I think that ItemGroup also has information about how the files were resolved that could be used to make the choice.

Yeah, we use it for actions, and I expect we would do the same for other per-assembly settings.

sbomer:
@marek-safar I've addressed your feedback - PTAL.

PR #1068, opened 2020-04-06
JSON annotations allow comments

vitek-karas:
Also fixes a bad merge between the this annotation and the JSON annotation changes


vitek-karas:
Merging this to fix the break due to bad merge. Let me know if there's any feedback for the JSON comments part.

PR #1069, opened 2020-04-06
[feature/dataflow] systemtype

tlakollo:
-Create new set of test for ConstructorUsedViaReflection and MethodUsedViaReflection
-Port functionality of System.Type.GetConstructor and System.Type.GetMethod from ReflectionPatternDetector.Process to new HandleCall method

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -3690,7 +3690,7 @@ namespace Mono.Linker.Steps {
 						// GetConstructor (BindingFlags, Binder, CallingConventions, Type[], ParameterModifier [])
 						//
 						case "GetConstructor" when calledMethod.DeclaringType.Name == "Type"
-						  && calledMethod.Parameters.Count >= 2
+						  && calledMethod.Parameters.Count >= 1
 						  && calledMethod.DeclaringType.Namespace == "System": {
 
 								reflectionContext.AnalyzingPattern ();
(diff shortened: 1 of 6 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3683,6 +3653,90 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+								BindingFlags bindingFlags = BindingFlags.Default;
+								if (parameters.Count > 1) {
+									if (methodParams [1].AsConstInt () != null)
+										bindingFlags |= (BindingFlags)methodParams [1].AsConstInt ();
+								}
+								// Go over all types we've seen
+								foreach (var value in methodParams [0].UniqueValues ()) {
+									if (value is SystemTypeValue systemTypeValue) {
+										MarkMethodsFromReflectionCall (ref reflectionContext, systemTypeValue.TypeRepresented, ".ctor", bindingFlags, ctorParameterCount);
+									} else if (value == NullValue.Instance) {
```
MichalStrehovsky, line 3677: Vitek added a helper to handle the non-SystemTypeValue nodes. I see that the code here is a copy of how Activator.CreateInstance was handled before Vitek's change. Can you update this to how Activator.CreateInstance is handled now?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3739,6 +3659,159 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+						// GetProperty (string, Type, Type[], ParameterModifier[])
+						// GetProperty (string, BindingFlags, Binder, Type, Type[], ParameterModifier[])
+						//
+						case var fieldPropertyOrEvent when ((fieldPropertyOrEvent == "GetField" || fieldPropertyOrEvent == "GetProperty" || fieldPropertyOrEvent == "GetEvent")
+							&& calledMethod.DeclaringType.Namespace == "System"
+							&& calledMethod.DeclaringType.Name == "Type"
+							&& calledMethod.Parameters [0].ParameterType.FullName == "System.String"): {
+								
+								reflectionContext.AnalyzingPattern ();
+								DynamicallyAccessedMemberKinds? memberKind = null;
```
vitek-karas, line 3764: The original code was able to handle `BindingFlags` we ignore them here. We should probably add that.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -4083,11 +4103,25 @@ void MarkPropertiesOnTypeHierarchy (ref ReflectionPatternContext reflectionConte
...
 							if ((property.GetMethod != null) && !property.GetMethod.IsStatic) continue;
 							if ((property.SetMethod != null) && !property.SetMethod.IsStatic) continue;
 						}
 
+						if ((bindingFlags & (BindingFlags.Instance | BindingFlags.Static)) == BindingFlags.Instance) {
+							if ((property.GetMethod != null) && property.GetMethod.IsStatic) continue;
+							if ((property.SetMethod != null) && property.SetMethod.IsStatic) continue;
+						}
+						if ((bindingFlags & (BindingFlags.Public | BindingFlags.NonPublic)) == BindingFlags.Public) {
+							if ((property.GetMethod != null) && !property.GetMethod.IsPublic) continue;
```
vitek-karas, line 4116: This is incorrect - property is considered public even if only one of set/get is public. Similarly for events.

PR #1073, opened 2020-04-07
Small fixes - flow annotations and warnings

vitek-karas:
Fix a nullref when JSON annotation is applied to an abstract property.

Record the method for which we create default return value (happens for example when the code calls `typeof(T)`)

### DIFF
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -168,7 +168,8 @@ namespace Mono.Linker.Dataflow
 					MethodDefinition setMethod = property.SetMethod;
 					if (setMethod != null) {
 
-						if (!ScanMethodBodyForFieldAccess (setMethod.Body, write: true, out backingFieldFromSetter)) {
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -168,7 +168,8 @@ private TypeAnnotations BuildTypeAnnotations (TypeDefinition type)
 					MethodDefinition setMethod = property.SetMethod;
 					if (setMethod != null) {
 
-						if (!ScanMethodBodyForFieldAccess (setMethod.Body, write: true, out backingFieldFromSetter)) {
+						// TODO: Handle abstract properties - no way to propagate the annotation to the field
```
MichalStrehovsky, line 171: Do we really need to do anything with the TODO here? I think this will fall out once we enforce that all virtuals have matching annotations - the property will be reintroduced in the derived class and the setter will be a virtual override of the abstract base.

vitek-karas: I think so as well, but I wanted to add the TODO so that we don't forget to validate that.

PR #1075, opened 2020-04-07
[feature/dataflow] Port RuntimeReflectionExtensions.* pattern recognition

mateoatr:


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3484,6 +3448,80 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+								reflectionContext.AnalyzingPattern ();
+								DynamicallyAccessedMemberKinds? memberKind = null;
+
+								switch (getRuntimeMember) {
+									case var mt when getRuntimeMember.EndsWith ("Event"):
+										memberKind = DynamicallyAccessedMemberKinds.Events;
+										break;
+
+									case var mt when getRuntimeMember.EndsWith ("Field"):
+										memberKind = DynamicallyAccessedMemberKinds.Fields;
```
vitek-karas, line 3473: I tried the runtime behavior and it only returns public members via these calls. Interestingly enough `GetRuntimeFields` (the plural) returns all private/public. But the `GetRuntimeField` returns only public ones. Probably similar for the rest of these.

test/Mono.Linker.Tests.Cases/Reflection/RuntimeReflectionExtensionsCalls.cs
```diff
@@ -1,9 +1,11 @@
 using System;
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 
 namespace Mono.Linker.Tests.Cases.Reflection
 {
+	[SkipKeptItemsValidation]
```
vitek-karas, line 8: I think we should leave the existing tests working via the Kept attributes. We need both kinds of tests - the more "End-2-end" which validate that linker correctly marks the members. Some of the data flow tests go without the Kept attribute as they are mostly validating that we correctly report warnings, in which case validating what was kept is not that interesting.

I think that to test this change we should stick to the Kept-style testing.

mateoatr:
I've updated the tests and moved to using binding flags so that we don't keep non public members, PTAL @vitek-karas.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3522,6 +3486,85 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+										memberKind = DynamicallyAccessedMemberKinds.PublicMethods;
+										break;
+
+									case var mt when getRuntimeMember.EndsWith ("Property"):
+										memberKind = DynamicallyAccessedMemberKinds.PublicProperties;
+										break;
+
+									default:
+										Debug.Fail ("Unsupported member type");
+										reflectionContext.RecordUnrecognizedPattern ($"Reflection call '{calledMethod.FullName}' inside '{callingMethodBody.Method.FullName}' is of unexpected member type.");
```
vitek-karas, line 3525: Nit: This should just throw - if we get here something very bad happened (the code is inconsistent).

PR #1076, opened 2020-04-07
Change illink default action to "link"

sbomer:
As discussed in https://github.com/mono/linker/pull/1057#discussion_r404948069

### DIFF
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // Driver.cs
 //
 // Author:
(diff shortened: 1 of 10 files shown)


src/linker/Linker/Driver.cs
```diff
@@ -821,7 +821,11 @@ string GetNextStringValue ()
 		protected virtual LinkContext GetDefaultContext (Pipeline pipeline)
 		{
 			LinkContext context = new LinkContext (pipeline) {
+#if FEATURE_ILLINK
+				CoreAction = AssemblyAction.Link,
+#else
 				CoreAction = AssemblyAction.Skip,
+#endif
 				UserAction = AssemblyAction.Link,
```
marek-safar, line 829: Should we change the to copy-used?

src/linker/Linker/Driver.cs
```diff
@@ -821,7 +821,11 @@ string GetNextStringValue ()
 		protected virtual LinkContext GetDefaultContext (Pipeline pipeline)
 		{
 			LinkContext context = new LinkContext (pipeline) {
+#if FEATURE_ILLINK
+				CoreAction = AssemblyAction.Link,
```
marek-safar, line 825: Please tweak the documentation at https://github.com/sbomer/linker/blob/illinkDefaults/src/linker/Linker/Driver.cs#L869

src/linker/Linker/Driver.cs
sbomer: Hmm, I think the defaults are most relevant for people who are using illink either from the command-line, or who are using the MSBuild task directly in their own targets (not through the SDK which probably defaults to copyused). I personally would want those defaults to have the "link" behavior similar to the monolinker command-line, but I can change it if you prefer. Note that most of our tests rely on the "link" behavior so I would need to tweak those as well.

PR #1078, opened 2020-04-08
Type.GetType data flow analysis and annotation propagation

vitek-karas:
Implements intrisics handling for Type.GetType.
Implements data flow of the return value.
Implements annotation propagation for string value and its handling as Type.GetType input.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2782,47 +2782,6 @@ namespace Mono.Linker.Steps {
 									ProcessSystemTypeGetMemberLikeCall (ref reflectionContext, System.Reflection.MemberTypes.Property, instructionIndex - 1);
 
 								break;
-
-							//
-							// GetType (string)
-							// GetType (string, Boolean)
-							// GetType (string, Boolean, Boolean)
-							// GetType (string, Func<AssemblyName, Assembly>, Func<Assembly, String, Boolean, Type>)
-							// GetType (string, Func<AssemblyName, Assembly>, Func<Assembly, String, Boolean, Type>, Boolean)
-							// GetType (string, Func<AssemblyName, Assembly>, Func<Assembly, String, Boolean, Type>, Boolean, Boolean)
-							//
-							case "GetType":
-								if (!methodCalled.IsStatic) {
-									break;
-								} else {
-									reflectionContext.AnalyzingPattern ();
-									
-									var first_arg_instr = GetInstructionAtStackDepth (_instructions, instructionIndex - 1, methodCalled.Parameters.Count);
-									if (first_arg_instr < 0) {
-										reflectionContext.RecordUnrecognizedPattern ($"Reflection call '{methodCalled.FullName}' inside '{_methodCalling.FullName}' couldn't be decomposed");
-										break;
-									}
-
-									//
-									// The next value must be string constant (we don't handle anything else)
-									//
-									var first_arg = _instructions [first_arg_instr];
-									if (first_arg.OpCode != OpCodes.Ldstr) {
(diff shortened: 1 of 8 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3575,6 +3534,47 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+							&& calledMethod.DeclaringType.Namespace == "System"
+							&& calledMethod.Parameters.Count >= 1
+							&& calledMethod.Parameters [0].ParameterType.MetadataType == MetadataType.String: {
+								reflectionContext.AnalyzingPattern ();
+
+								foreach (var typeNameValue in methodParams [0].UniqueValues ()) {
+									if (typeNameValue is KnownStringValue knownStringValue) {
+										TypeDefinition foundType = _markStep.ResolveFullyQualifiedTypeName (knownStringValue.Contents);
+										if (foundType == null) {
+											reflectionContext.RecordUnrecognizedPattern ($"Reflection call '{calledMethod.FullName}' inside '{reflectionContext.MethodCalling.FullName}' was detected with type name `{knownStringValue.Contents}` which can't be resolved.");
```
MichalStrehovsky, line 3555: This could be just a lightup that is expected to fail sometime. I think if we don't find anything, it's fine to treat it as recognized. We basically only want to guarantee the behavior before/after linking is the same, and modulo bugs in our `ResolveFullyQualifiedTypeName` implementation, this should produce the same results.

(Same comment for the `uniqueValue is KnownStringValue knownStringValue` below)

vitek-karas: I understand that if we get everything right what you say should be true. The worry I have is that we won't get everything right.

Separate discussion is: If we can determine that the code is wrong and will fail at runtime - should we warn about it? I know we gave up on nulls (which will also fail at runtime)... but this is slightly different.

vitek-karas: I think I see your point though...

One side note: This failure is the one we would get if somebody calls a Type.GetType(assemblyQualifiedName) and we don't manage to pull that assembly into the closure (the existing problem here). Without this warning, there would be no static hint that linker broke the app. So maybe we should keep it for now and only remove it once we solve the assembly closure problem?

MichalStrehovsky: > The worry I have is that we won't get everything right.

That would be a bug we need to fix.

> If we can determine that the code is wrong and will fail at runtime - should we warn about it?

.NET Native warns from some places (cough, MCG, cough) and we get support calls asking what to do about warnings like "Unresolved p/invoke to method `blah` from library `__Internal`. The code will throw at runtime". `__Internal` is Mono-specific goo and we're complaining about a Mono-specific codepath in a random NuGet package the customer is using that is never reachable at runtime.

I expect we would see similar patterns in reflection in lightups and detections. E.g.

```csharp
public static bool IsRunningOnMono ()
{
    return Type.GetType ("Mono.Runtime") != null;
}
```

Is the recommended [official way](https://www.mono-project.com/docs/gui/winforms/porting-winforms-applications/) to detect Mono. (Might change for .NET 5, but there's plenty existing code in the wild).

marek-safar: > Is the recommended official way to detect Mono

This is recommended way to detect Mono but not .NET 5 MonoVM (we broke that intentionally)

MichalStrehovsky: > So maybe we should keep it for now and only remove it once we solve the assembly closure problem?

I think it would be a poor experience if we allow things to ship that way. I filed #1079 with some ideas.

PR #1080, opened 2020-04-08
[feature/dataflow] Implement dataflow to custom attributes

### LINKED ISSUE #1072: [feature/dataflow] Data flow analysis must consider attribute instantiations as callsites for the .ctor
If an attribute `.ctor` parameter is annotated, for example
```C#
public class MyAttribute : Attribute
{
   public MyAttribute([DynamicallyAccessedMembers(DynamicallyAccessMemberKinds.DefaultConstructor]) Type type
};

    [DynamicallyAccessedMembers(DynamicallyAccessedMemberKinds.Methods)]
    public Type MyProperty { get; set; }
```

And then it's used like
```C#
[MyAttribute(typeof(MyA))]
public class MyType
{
}
```

The data flow analysis actually doesn't see the attribute instantiation as a call to the `.ctor`. As such it doesn't apply the annotations on the parameters. And so in the above sample `MyA.ctor` would not get marked.

Similar issue applies to properties on attributes, where something like:
```C#
[MyAttribute(typeof(MyA), MyProperty = typeof(MyB)]
```

would not apply the annotation from `MyProperty` to the `typeof(MyB)`.

MichalStrehovsky:
A custom attribute application should be treated the same as a method call (followed by potentially several calls to property setters and field stores). All parameters will be statically known by definition.

I plan to fix up the string-based tests before this is merged (going to merge after #1078 that is a prerequisite).

Fixes #1072

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -3386,9 +3386,11 @@ namespace Mono.Linker.Steps {
 
 			public void ProcessAttributeDataflow (MethodDefinition method, Collection<CustomAttributeArgument> arguments)
 			{
+				int paramOffset = method.HasImplicitThis () ? 1 : 0;
+
 				for (int i = 0; i < method.Parameters.Count; i++)
 				{
-					var annotation = _flowAnnotations.GetParameterAnnotation (method, i + 1 /* adjust for `this` */);
+					var annotation = _flowAnnotations.GetParameterAnnotation (method, i + paramOffset);
 					if (annotation != 0) {
 						ValueNode valueNode = GetValueNodeForCustomAttributeArgument (arguments [i]);
 						if (valueNode != null) {
@@ -3424,8 +3426,8 @@ namespace Mono.Linker.Steps {
 				} else if (argument.Type.MetadataType == MetadataType.String) {
 					valueNode = new KnownStringValue ((string)argument.Value);
 				} else {
-					Debug.Fail ("We shouldn't have gotten a non-null annotation for this from GetParameterAnnotation.");
-					valueNode = null;
+					// We shouldn't have gotten a non-null annotation for this from GetParameterAnnotation
+					throw new InvalidOperationException ();
 				}
 
 				return valueNode;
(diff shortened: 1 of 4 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3409,6 +3425,53 @@ public void ScanAndProcessReturnValue (MethodBody methodBody)
 				}
 			}
 
+			public void ProcessAttributeDataflow (MethodDefinition method, Collection<CustomAttributeArgument> arguments)
+			{
```
vitek-karas, line 3429: Should we check that the method is a .ctor?
The code will not work for this-calls correctly and it will ignore any potential return values.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3409,6 +3425,53 @@ public void ScanAndProcessReturnValue (MethodBody methodBody)
...
+			private ValueNode GetValueNodeForCustomAttributeArgument(CustomAttributeArgument argument)
+			{
+				ValueNode valueNode;
+				if (argument.Type.Name == "Type") {
+					TypeDefinition referencedType = ((TypeReference)argument.Value).Resolve ();
+					valueNode = referencedType == null ? null : new SystemTypeValue (referencedType);
+				} else if (argument.Type.MetadataType == MetadataType.String) {
+					valueNode = new KnownStringValue ((string)argument.Value);
+				} else {
+					Debug.Fail ("We shouldn't have gotten a non-null annotation for this from GetParameterAnnotation.");
```
vitek-karas, line 3468: Nit: Recently there's been feedback to now use Debug.Assert/Fail and instead just throw exceptions. Mainly due to test failures being hard to handle (asserts by default popup a dialog and so on) and also because if this got hit in the wild we might get some data back eventually.

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: I don't think reflection stack would be able to do anything at runtime if this is not a constructor, so I'm not worried about ignoring return values.

Having an `explicitthis` constructor?... maybe. Added the handling for that just in case...

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: Automation should ideally add a listener that does something sane (make the unit test fail, or something like that), but I don't think I'll have time to look into that. Made it a throw.

PR #1081, opened 2020-04-08
[feature/dataflow] Look for fields, properties and methods in base type

### LINKED ISSUE #1042: Expression.Call/Field/Property pattern match does not work on members from derived classes
The semantics of `Expression.Call(Type, string, Type [])` is to look for all static public/non-public methods on the specified type and all its base types. Internally it uses `BindingFlags.FlattenHierarchy` to achieve this: https://github.com/dotnet/runtime/blob/d3366b86bc2a1415ae1998334f6a5241b14f0c4f/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/MethodCallExpression.cs#L1172-L1173

The pattern matching in the linker does not look for methods on base types though, so it will not work in all cases.

Example:
```C#
class ABase {
    public static void OnBase() {}
}

class ADerived : ABase {
}

Expression.Call(typeof(ADerived), "OnBase", Type.EmptyTypes);
```

In this case linker will not mark `ABase.OnBase` as reachable and possibly remove it (and it will issue a "warning").

Same problem applies to `Expression.Field(Expression, Type, string)` and `Expression.Property(Expression, Type, string)`.

mateoatr:
If a field, property or method which is accessed via reflection is not found inside the callers type, look for it in it's base type (if any). This fixes #1042.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -3560,10 +3560,31 @@ namespace Mono.Linker.Steps {
 												bool staticOnly = methodParams [0].Kind == ValueNodeKind.Null;
 												// TODO: Change this as needed after deciding if we are to keep all fields/properties on a type
 												// that is accessed via reflection. For now, let's only keep the field/property that is retrieved.
-												if (calledMethod.Name [0] == 'P')
-													MarkPropertiesFromReflectionCall (ref reflectionContext, systemTypeValue.TypeRepresented, stringValue.Contents, staticOnly);
-												else
-													MarkFieldsFromReflectionCall (ref reflectionContext, systemTypeValue.TypeRepresented, stringValue.Contents, staticOnly);
+												if (calledMethod.Name [0] == 'P') {
+													MarkPropertiesOnTypeHierarchy (ref reflectionContext, systemTypeValue.TypeRepresented,
+														(property) => {
+															if (property.Name != stringValue.Contents) return false;
+
+															if (staticOnly) {
+																if (property.GetMethod != null && !property.GetMethod.IsStatic) return false;
+																if (property.SetMethod != null && !property.SetMethod.IsStatic) return false;
+															}
+
+															return true;
+														});
+												} else {
+													MarkFieldsOnTypeHierarchy (ref reflectionContext, systemTypeValue.TypeRepresented,
+														(field) => {
+															if (field.Name != stringValue.Contents) return false;
+
+															if (staticOnly) {
+																if (field != null && !field.IsStatic) return false;
+																if (field != null && !field.IsStatic) return false;
+															}
+
+															return true;
+														});
+												}
 											} else {
 												reflectionContext.RecordUnrecognizedPattern ($"Expression call '{calledMethod.FullName}' inside '{callingMethodBody.Method.FullName}' was detected with 3rd argument which cannot be analyzed");
 											}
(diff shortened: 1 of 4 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3869,32 +3869,36 @@ void MarkTypeForDynamicallyAccessedMembers (ref ReflectionPatternContext reflect
 			void MarkMethodsFromReflectionCall (ref ReflectionPatternContext reflectionContext, TypeDefinition declaringType, string name, BindingFlags? bindingFlags, int? parametersCount = null)
 			{
 				bool foundMatch = false;
-				foreach (var method in declaringType.Methods) {
-					var mname = method.Name;
+				do {
+					foreach (var method in declaringType.Methods) {
```
vitek-karas, line 3873: Note that reflection will not return private members of base type - ever. So this should also ignore private members on base class. Take a look at `MarkMethodsOnTypeHierarchy` it already implements all that logic. Maybe we could merge the two now?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3869,32 +3869,36 @@ void MarkTypeForDynamicallyAccessedMembers (ref ReflectionPatternContext reflect
...
-					var methodCalling = reflectionContext.MethodCalling;
-					reflectionContext.RecordRecognizedPattern (method, () => _markStep.MarkIndirectlyCalledMethod (method, new DependencyInfo (DependencyKind.AccessedViaReflection, methodCalling)));
-				}
+						foundMatch = true;
+						var methodCalling = reflectionContext.MethodCalling;
+						reflectionContext.RecordRecognizedPattern (method, () => _markStep.MarkIndirectlyCalledMethod (method, new DependencyInfo (DependencyKind.AccessedViaReflection, methodCalling)));
+					}
+
+					declaringType = declaringType.BaseType.Resolve ();
+				} while (!foundMatch && declaringType.FullName != "System.Object");
```
vitek-karas, line 3901: Nit: It's not necessary to check for `System.Object`, just stop at null - take a look at `MarkMethodsOnTypeHierarchy`.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3905,32 +3909,36 @@ void MarkMethodsFromReflectionCall (ref ReflectionPatternContext reflectionConte
 			void MarkPropertiesFromReflectionCall (ref ReflectionPatternContext reflectionContext, TypeDefinition declaringType, string name, bool staticOnly = false)
```
vitek-karas, line 3909: There are now similar `MarkPropertiesOnTypeHierarchy` and also for fields `MarkFieldsOnTypeHierarchy` - maybe we should merge the two - if possible.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3560,10 +3560,31 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+															return true;
+														});
+												} else {
+													MarkFieldsOnTypeHierarchy (ref reflectionContext, systemTypeValue.TypeRepresented,
+														(field) => {
+															if (field.Name != stringValue.Contents) return false;
+
+															if (staticOnly) {
+																if (field != null && !field.IsStatic) return false;
+																if (field != null && !field.IsStatic) return false;
```
vitek-karas, line 3582: Unnecessary duplicate condition

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3560,10 +3560,31 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+
+															if (staticOnly) {
+																if (property.GetMethod != null && !property.GetMethod.IsStatic) return false;
+																if (property.SetMethod != null && !property.SetMethod.IsStatic) return false;
+															}
+
+															return true;
+														});
+												} else {
+													MarkFieldsOnTypeHierarchy (ref reflectionContext, systemTypeValue.TypeRepresented,
```
vitek-karas, line 3576: Just a thought: With this the func you pass in requires a closure (the `staticOnly` variable) which means each call here will allocate a new object. If we move the check for static only inside the method and pass it as a parameter we would avoid the allocation. I know it's not as nice necessarily, but it will be faster.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -4029,11 +3987,15 @@ void MarkFieldsOnTypeHierarchy (ref ReflectionPatternContext reflectionContext,
...
 						reflectionContext.RecordRecognizedPattern (field, () => _markStep.MarkField (field, new DependencyInfo (DependencyKind.AccessedViaReflection, methodCalling)));
+						foundMatch = true;
 					}
 
 					type = type.BaseType?.Resolve ();
 					onBaseType = true;
 				}
+
+				if (!foundMatch)
+					reflectionContext.RecordUnrecognizedPattern ($"Reflection call '{reflectionContext.MethodCalled.FullName}' inside '{reflectionContext.MethodCalling.FullName}' could not resolve any field on type `{receivedType}`.");
```
vitek-karas, line 3998: This is problematic for two reasons:
* For the cases where we don't filter on name at all, we don't want to fail if there's nothing - that is the case of `GetFields` - it's not a problem calling that on a type without fields - so we should not warn.
* As we discussed we should probably not warn if we find `GetField("fieldName")` and can't find `fieldName` on the type. We did recognize the pattern, but it didn't end up marking anything. I know that's how it worked before, but I think we should change that.

/cc @MichalStrehovsky - for another view on this.

mateoatr: Should we then remove all calls to `reflectionContext.AnalyzingPattern` inside `HandleCall` and only call it inside `RequireDynamicallyAccessedMembers`?

vitek-karas: Maybe - but I would not worry about that. We will have to change the `reflectionContext` anyway. It doesn't have enough capability currently to record events which are not about methods (for example storing a value in a field can't be expressed in it now).

The other change which will have to happen is to how the `AnalyzingPattern` works. Currently the context is supposed to guarantee that if we call `AnalyzingPattern` then we will call either "recognized" or "unrecognized" methods before returning. But this is already broken by the generic `RecordHandledPattern` which is basically "Don't worry - I did the right thing", but it actually doesn't report either "recognized" or "unrecognized" (it reports nothing). I plan to simply remove this check and don't make any promises about recognized/unrecognized calls. The main "promise" will be that we report "unrecognized" whenever we find something we don't understand. And that we report "recognized" for things where we think we know what's going on. But no consistency checks.

So in short - don't worry about this - we will get rid of that call eventually :-)

mateoatr: For now I've opted for simply using the `RecordHandledPattern` whenever we mark something inside the new infra, this way we can move from kind of duplicating the old `Mark*FromReflectionCall` in the new infra to just use the `Mark*OnType(Hierarchy)` methods.

PR #1083, opened 2020-04-09
[feature/dataflow] Add support for byref operations

MichalStrehovsky:
We track indirect writes to method parameters and fields. This is necessary to support dataflow annotations on `out` and `ref` parameters.


MichalStrehovsky:
We'll probably be able to get rid of the IsByRef tracking on StackSlot. I thought we might use it in validation, but it's a giant pain because we would need to start analyzing pointer math and stuff like that (to preserve the IsByRef flag, whose only purpose would be to validate we're not messing up too much by not modelling things in detail).

PR #1084, opened 2020-04-09
Migrate SubStepsDispatcher to linker and add basic test coverage for it

marek-safar:
/cc @vitek-karas

### DIFF
--- a/src/linker/Linker.Steps/BaseSubStep.cs
+++ b/src/linker/Linker.Steps/BaseSubStep.cs
@@ -1,0 +1,48 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public abstract class BaseSubStep : ISubStep
+	{
+		public AnnotationStore Annotations => Context.Annotations;
+
+		protected LinkContext Context { get; private set; }
+
+		public abstract SubStepTargets Targets { get; }
+
+		public virtual void Initialize (LinkContext context)
+		{
+			Context = context;
+		}
+
+		public virtual bool IsActiveFor (AssemblyDefinition assembly) => true;
+
+		public virtual void ProcessAssembly (AssemblyDefinition assembly)
+		{
+		}
+
+		public virtual void ProcessType (TypeDefinition type)
+		{
+		}
+
+		public virtual void ProcessField (FieldDefinition field)
+		{
+		}
+
+		public virtual void ProcessMethod (MethodDefinition method)
+		{
+		}
+
+		public virtual void ProcessProperty (PropertyDefinition property)
+		{
+		}
+
+		public virtual void ProcessEvent (EventDefinition @event)
+		{
+		}
+	}
+}
--- a/src/linker/Linker.Steps/BaseSubStep.cs
+++ b/src/linker/Linker.Steps/BaseSubStep.cs
@@ -8,7 +8,7 @@ namespace Mono.Linker.Steps
 {
 	public abstract class BaseSubStep : ISubStep
 	{
-		public AnnotationStore Annotations => Context.Annotations;
+		protected AnnotationStore Annotations => Context.Annotations;
 
 		protected LinkContext Context { get; private set; }
 
--- a/src/linker/Linker.Steps/SubStepsDispatcher.cs
+++ b/src/linker/Linker.Steps/SubStepsDispatcher.cs
@@ -1,0 +1,219 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+using Mono.Cecil;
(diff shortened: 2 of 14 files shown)


src/linker/Linker.Steps/BaseSubStep.cs
```diff
@@ -0,0 +1,48 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public abstract class BaseSubStep : ISubStep
+	{
+		public AnnotationStore Annotations => Context.Annotations;
```
vitek-karas, line 11: Should this be `protected` just like the `Context` - it seems weird those two are not the same.

src/linker/Linker.Steps/SubStepsDispatcher.cs
```diff
@@ -0,0 +1,219 @@
...
+using Mono.Collections.Generic;
+
+namespace Mono.Linker.Steps
+{
+	//
+	// Generic steps dispatcher is intended to by used by custom linker step which
+	// consist of multiple steps. It simplifies their implementation as well as the
+	// way how to hook them into the pipeline of existing steps.
+	//
+	public abstract class SubStepsDispatcher : IStep, IEnumerable<ISubStep>
```
vitek-karas, line 17: Why is this abstract? It seems unnecessary to have to override this - I understand that custom steps will want to do that, so that they can register it as a custom step, but in general it feels unnecessary.

src/linker/Linker.Steps/SubStepsDispatcher.cs
```diff
@@ -0,0 +1,219 @@
...
+
+		void BrowseAssemblies (IEnumerable<AssemblyDefinition> assemblies)
+		{
+			foreach (var assembly in assemblies) {
+				CategorizeSubSteps (assembly);
+
+				if (HasSubSteps (on_assemblies))
+					DispatchAssembly (assembly);
+
+				if (!ShouldDispatchTypes ())
```
vitek-karas, line 60: If we think this is important to check up-front for perf reasons, then I would argue we should cache the "should" outside of the foreach. Similarly for the `HasSubSteps(on_assemblies)`.

src/linker/Linker.Steps/SubStepsDispatcher.cs
```diff
@@ -0,0 +1,219 @@
...
+
+				if (type.HasEvents && HasSubSteps (on_events))
+					BrowseEvents (type.Events);
+
+				if (type.HasNestedTypes)
+					BrowseTypes (type.NestedTypes);
+			}
+		}
+
+		void BrowseFields (Collection<FieldDefinition> fields)
```
vitek-karas, line 98: Nit: I would personally inline these into the big Browse method - I know that JIT will probably inline them anyway, but still...

src/linker/Linker.Steps/SubStepsDispatcher.cs
vitek-karas, line 17: Also curious: What is the use case for having the `IEnumerable<ISubStep>` implemented by the class?

vitek-karas: Especially since we don't expose that through the ref assemblies.

test/Mono.Linker.Tests.Cases/Extensibility/SubStepDispatcherUsage.cs
```diff
@@ -0,0 +1,28 @@
...
+
+		[Kept]
+		public class NestedType
+		{
+			public int field;
+
+			public static void SomeMethod ()
+			{
+			}
+        }
```
vitek-karas, line 26: Nit: Indentation

src/linker/Linker.Steps/BaseSubStep.cs
marek-safar: Resolved

src/linker/Linker.Steps/SubStepsDispatcher.cs
marek-safar: Removed

src/linker/Linker.Steps/SubStepsDispatcher.cs
marek-safar: That is not possible as `CategorizeSubSteps` rebuilds the logic for every assembly

src/linker/Linker.Steps/SubStepsDispatcher.cs
marek-safar: ok

src/linker/Linker.Steps/SubStepsDispatcher.cs
marek-safar: Removed

vitek-karas: OK - I now realized that even inside the linker (without custom steps) the type names of steps are used to identify them. So it makes sense.

src/linker/Linker.Steps/SubStepsDispatcher.cs
vitek-karas: hmm - OK. Not sure I like this approach, but it works.

PR #1089, opened 2020-04-09
[feature/dataflow] Only keep public DefaultConstructor

sbomer:
Posting this mostly to ask a question @MichalStrehovsky @vitek-karas

The attribute encoding sort of assumes that the default ctor is public. The logic that keeps default ctors works for private default ctors, but the logic that keeps public ctors will not keep a private default ctor. Should we do something differently to prevent the behavior illustrated in this failing testcase? (`RequireDefaultConstructor` works, but `RequireDefaultOrPublicConstructors` will not keep the default private ctor on `PrivateDefaultAndPublicConstructorsType`).

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -3843,7 +3843,8 @@ namespace Mono.Linker.Steps {
 				if (requiredMemberKinds.HasFlag (DynamicallyAccessedMemberKinds.Constructors)) {
 					MarkConstructorsOnType (ref reflectionContext, typeDefinition, filter: null);
 				} else if (requiredMemberKinds.HasFlag (DynamicallyAccessedMemberKinds.PublicConstructors)) {
-					MarkConstructorsOnType (ref reflectionContext, typeDefinition, filter: m => m.IsPublic);
+					// Also keep default ctor which may be non-public.
+					MarkConstructorsOnType (ref reflectionContext, typeDefinition, filter: m => m.IsPublic || m.Parameters.Count == 0);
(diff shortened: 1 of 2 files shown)


vitek-karas:
This is a valid concern. Basically the fact that `DefaultConstructor` is a subset (bitwise) to `PublicConstructors` is weird and the implementation doesn't actually follow that intent.

I think we should simply fix the implementation. It would be much harder to explain that the different constructor settings are not in a nice inclusion relationship. So I personally would be fine if `PublicConstructors` also included potentially private default constructors.

MichalStrehovsky:
> This is a valid concern. Basically the fact that DefaultConstructor is a subset (bitwise) to PublicConstructors is weird and the implementation doesn't actually follow that intent.

Maybe I've been doing compilers for too long, but "public" is what we consider default (we don't consider the private one default in AOT compilers because it saves disk space). This is what `Activator.CreateInstance<T>` (and `new T()` in C#) can work with and that's why compilers treat it specially and have a special name for it (the private one being called "parameterless constructors"). In the compiler we actually go even further and consider the constructor on an abstract class "non-default".

Default constructor pretty much means "the constructor that the C# compiler generates for you by default". That one is public.

If you think the distinction doesn't make sense, I'm fine making it mean "parameterless". It was trying to optimize for the form of the most popular CreateInstance overload.

sbomer:
@MichalStrehovsky thanks for the clarification! The distinction makes sense to me. Elsewhere in the linker, we aren't that precise (`IsDefaultConstructor` will return parameterless instance ctors), but I think we should be. I've changed the logic to only keep the true Default ctors (it used to keep private parameterless ctors).

PR #1090, opened 2020-04-10
[feature/dataflow] Local dataflow tests

sbomer:
I had another question @MichalStrehovsky @vitek-karas:

It looks to me like the dataflow analysis doesn't handle backwards edges, which causes holes in the analysis. I could be missing something - here is my understanding:

It looks to me like the Scan logic propagates stacks through all edges in the control flow graph, but smaller IL offsets aren't revisited, so merged stacks from backwards branches aren't taken into account. Do we care? (Is it even possible to exploit this from C#?)

The same issue exists for local tracking as well. We only update the locals store as we traverse forwards edges, so changes that might reach previous instructions aren't considered by those instructions. I wrote a testcase `TestBackwardsEdge` that shows this behavior.

Along similar lines, the local tracking also accumulates locals from all blocks with lower IL offsets, even ones that couldn't reach a given instruction. This seems overly conservative (but at least overzealous warnings are better than missing warnings like for the backwards edge). This is shown in the `TestBranchIf*` cases.


MichalStrehovsky:
Yup, backwards branching is not handled. You could get into that situation with a `goto`, a weird loop, etc.  The analysis is not bulletproof - it just tries to be fast (in .NET Native this ran on all code, not just the "interesting" code).

Another hole is in exception handling - most IL instructions can potentially be faulting (e.g. we can hit a `NullReferenceException`, or out of bounds array access) - all of those are extra edges we don't handle.

If we were doing this for any other optimizations, I'm quite sure we would quickly hit this. For reflection patterns... I don't think we will anytime soon.

sbomer:
Thanks for confirming. Do you think it would make sense to warn when we see backwards edges or potentially faulting instructions in a reflection method? That would produce extraneous warnings in some cases, but I don't see a better way to detect these cases that doesn't involve doing extra work - and hopefully complex logic like this is rare in reflection methods anyway.

vitek-karas:
For my education: Why is it bad to not track the edges caused by faulting instructions? That is not a "backward" flow - it simply "jumps" to the catch block, right?

sbomer:
Good question. I haven't been able to come up with a way to create backwards edges for exception handling in C#, since it looks like catch blocks always come after try blocks, which matches the flow of execution. Maybe there is some way to break it using IL? For example it looks like the leave instruction can jump backwards or even back into the try block.

sbomer:
BTW I realized that the conservative local tracking might not only cause extra warnings - it could keep members on types where they're not required. It might be one of the things worth measuring for size impact.

vitek-karas:
It is true that this can lead to including more than necessary. But so far that has been our general approach to this. If we can't figure out exactly what to mark, we tend to mark more to make things work. So far the rule has been:
* If we can mark some additional type members and make things work - do that
* If we can't figure out which exact type, fail - that means we won't mark additional types to make things work (unless we know exactly which ones).

I do agree that this may have size impact, but I think it comes secondary for now. If we find out that this is a problem we'll look into it. But right now the annotations don't have the ability to pick exactly the right member on a type, only large groups (public/all).

sbomer:
I agree that it's aligned with the general approach we've been taking. I just wanted to note that this adds a rule something like:
- If we have a conservative (maybe inexact) guess for the type, and we can mark some additional type members to make things work, mark them on all of the guessed types.

Which can end up marking additional types. For example this would keep `Foo`:
```csharp
string s = "Foo";
if (condition) {
    s = "Bar";
    RequireMethods(s);
}
```
I imagine such a pattern is uncommon, so hopefully this is more of an edge case we don't need to worry about - I just like to make sure that edge cases are understood.

MichalStrehovsky:
> For my education: Why is it bad to not track the edges caused by faulting instructions? That is not a "backward" flow - it simply "jumps" to the catch block, right?

I think that was just a perf optimization in the .NET Native analyzer that I'm stealing from.

Some of the simplifications that the scanner is doing help for some of the more "advanced" dataflow analysis - I'm still not sure whether we'll need to port those. One of the things that I didn't port is dataflow through `async` state machines. I'm unclear on whether that's needed by real code, but I assume there was something that motivated adding this in .NET Native - but all context is now gone with the people who wrote it.)

(The `async` state machine tracking makes it possible to track dataflow within methods like: `async Task Foo() { Type t = typeof(X); await Blah(); t.GetMethod("Yadda").Invoke(...);` - these are not actually a single method body in IL).

PR #1095, opened 2020-04-14
[feature/dataflow] Implement pattern recognition for the rest of CreateInstance overloads

vitek-karas:
This implements it for Activator.CreateInstance* with strings, AppDomain.CreateInstance* and Assembly.CreateInstance.

The implementation is basically just copying the existing behavior. Didn't try to improve it much yet.

Added tests.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -3827,8 +3827,18 @@ namespace Mono.Linker.Steps {
 							&& calledMethod.DeclaringType.Namespace == "System"
 							&& calledMethod.DeclaringType.Name == "Activator"
 							&& calledMethod.Parameters.Count == 0: {
-								// Not yet supported in any combination
 								reflectionContext.AnalyzingPattern ();
+
+								if (calledMethod is GenericInstanceMethod genericCalledMethod && genericCalledMethod.GenericArguments.Count == 1
+									&& genericCalledMethod.GenericArguments [0] is GenericParameter genericParameter) {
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3739,6 +3650,99 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+						// 
+						// static T CreateInstance<T> ()
+						//
+						case "CreateInstance" when calledMethod.ContainsGenericParameter
+							&& calledMethod.DeclaringType.Namespace == "System"
+							&& calledMethod.DeclaringType.Name == "Activator"
+							&& calledMethod.Parameters.Count == 0: {
+								// Not yet supported in any combination
+								reflectionContext.AnalyzingPattern ();
+								reflectionContext.RecordUnrecognizedPattern ($"Activator call '{reflectionContext.MethodCalled.FullName}' inside '{reflectionContext.MethodCalling.FullName}' is not supported");
```
MichalStrehovsky, line 3697: This might cause a lot of noise since this is what `new T()` in C# compiles to.

Would it make sense to at least check whether the generic argument has the `new()` constraint (in which case linker already made sure this works).

Later we can be more creative and e.g. check whether the T is a concrete type (in which case we can just keep the ctor here), or whether the T is a generic parameter that has a dataflow annotation that keeps the ctor (we don't have that yet).

vitek-karas: I added handling of the generic argument with constraint.

PR #1098, opened 2020-04-14
Fix Cecil PackageReference publishing

sbomer:
The change from `ProjectReference` to `PackageReference` was causing cecil not to be published with the full framework `ILLink.Tasks` project due to inconsistencies in how `PrivateAssets` are handled. This fixes the issue in https://github.com/dotnet/sdk/pull/11214#issuecomment-613566926.


marek-safar:
build failed

marek-safar:
@monojenkins build

PR #1099, opened 2020-04-14
[feature/dataflow] Implement detection of interesting reflection methods

vitek-karas:
This implements detection of intrinsic methods as well as annotated methods/fields so that full method body scanning can be done only when necessary.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2438,12 +2438,13 @@ namespace Mono.Linker.Steps {
 				if (eh.HandlerType == ExceptionHandlerType.Catch)
 					MarkType (eh.CatchType, new DependencyInfo (DependencyKind.CatchType, body.Method));
 
+			bool requiresReflectionMethodBodyScanner = false;
 			foreach (Instruction instruction in body.Instructions)
-				MarkInstruction (instruction, body.Method);
+				MarkInstruction (instruction, body.Method, ref requiresReflectionMethodBodyScanner);
 
 			MarkInterfacesNeededByBodyStack (body);
 
-			MarkReflectionLikeDependencies (body);
+			MarkReflectionLikeDependencies (body, requiresReflectionMethodBodyScanner);
 
 			PostMarkMethodBody (body);
 		}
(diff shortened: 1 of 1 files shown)


vitek-karas:
@MichalStrehovsky  could you please take a look if this makes sense. I need to merge in all the recent changes which are still missing (Type.Get* and the other CreateInstance methods), but as an approach it should work I think.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3259,6 +3288,296 @@ protected override void HandleStoreParameter (MethodDefinition method, int index
...
+
+					//
+					// System.Reflection.RuntimeReflectionExtensions
+					//
+					// static GetRuntimeEvent (this Type type, string name)
+					// static GetRuntimeField (this Type type, string name)
+					// static GetRuntimeMethod (this Type type, string name, Type[] parameters)
+					// static GetRuntimeProperty (this Type type, string name)
+					//
+					case var getRuntimeMember when (getRuntimeMember.StartsWith ("GetRuntime")
```
MichalStrehovsky, line 3345: Later (maybe not this pull request): making this a normal switch over a bunch of strings would get the C# compiler to emit a hashtable lookup. Switch statements with complex `when` conditions have perf characteristics that are hard for me to reason about (it's possible a lot of this is going to be cascading `if` checks). We'll need this method to be as fast as possible.

vitek-karas: I agree, I just didn't want to rewrite it too much yet.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3145,6 +3155,25 @@ private class ReflectionMethodBodyScanner : Dataflow.MethodBodyScanner
 			private readonly MarkStep _markStep;
 			private readonly FlowAnnotations _flowAnnotations;
 
+			public static bool RequiresReflectionMethodBodyScanner (FlowAnnotations flowAnnotations, MethodReference method)
```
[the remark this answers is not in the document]
vitek-karas: I think we'll need to look into the "choke point" in more detail. With the method body scanner the original design is somewhat weird. I filed https://github.com/mono/linker/issues/1102 to track this work.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3145,6 +3155,25 @@ private class ReflectionMethodBodyScanner : Dataflow.MethodBodyScanner
...
+			{
+				MethodDefinition methodDefinition = method.Resolve ();
+				if (methodDefinition != null) {
+					return GetIntrinsicIdForMethod (methodDefinition) != IntrinsicId.None || flowAnnotations.RequiresDataFlowAnalysis (methodDefinition);
```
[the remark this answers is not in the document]
vitek-karas: Fixed.

src/linker/Linker.Steps/MarkStep.cs
mrvoorhe: I wonder if there would be any advantage to refactoring this to an annotation stage step that populates a table in `AnnotationStore`.  The pros being 

* We could easily parallelize parts of the annotating process
* The annotation stage (or maybe it's own step) might be a nicer "choke point" as you talked about in a another comment?

Downside being, you end up processing methods that are never called.

vitek-karas: I was actually thinking about adding a cache similar to what you suggest. But instead of filling it by scanning everything (a separate step or similar) we could lookup the intrinsic. There's not that many of them and we even know which assemblies they should live in. That would avoid scanning all methods, would be much faster and would let us have only "positive" cache (only the intrinsic methods would be stored in the cache).

PR #1104, opened 2020-04-15
Data flow based reflection handling and annotation support

vitek-karas:
Changes how reflection calls are handled by the linker. With this change the linker implements a data flow analysis inside each method's body so it can correctly track the values of `Type` and `string` values (and some others).

On top of this infrastructure the existing reflection patterns where implemented, which almost exclusive improves the success rate of the linker figuring out the necessary Mark actions.

Having the data flow now enables the linker to implement type annotations where each `Type` value can be annotated with flags noting what kind of members are necessary on that type. This lets users add annotations throughout the code base which in turn enables linker to validate many more reflection patterns and also automatically mark the necessary members.

This is the core functionality necessary to implement the upcoming reporting which will warn users about unanalyzable reflection code usage and provide tools to solve most of the cases (via the annotations).

The change itself is relatively straightforward:
* Add the data flow functionality (in `Linker.DataFlow` subdirectory) - along with the annotation support
* Rewrite the reflection processing in the `MarkStep` using the data flow (in a new class)
* Couple of global changes to introduce a new command line option to specify annotations via a JSON file (although we expect most cases to use the attributes instead)
* Lot of tests added to cover the existing and new functionality


vitek-karas:
/cc @mrvoorhe 

@MichalStrehovsky @sbomer @mateoatr @tlakollo @eerhardt

vitek-karas:
Compared this to master (new == dataflow based analysis) by running linker on the entire framework rooting all public APIs (which basically keeps everything).
* The new produces slightly less "warnings" which are currently still reported as normal `stdout`.
* The new is slightly slower, by about 2% (29.0 seconds versus 28.4 seconds)

PR #1105, opened 2020-04-16
Use new infra for logging warnings

mateoatr:
In this PR I'm moving all messages which I think should be surfaced as warnings and that we currently log in the linker to use the new infra (the `MessageContainer`.) Although many of these could be logged and the result of the linked program would remain unaltered, all of them should cause the user to take action in some way or another (either by double checking a made assumption or by removing something in a XML descriptor / code).


src/linker/WarningCodes.md
```diff
@@ -0,0 +1,74 @@
+# IL Linker Warning Codes
```
marek-safar, line 1: Please add a section for error codes as well and link this file from the source code

src/linker/WarningCodes.md
```diff
@@ -0,0 +1,74 @@
+# IL Linker Warning Codes 
+
+### IL2001
+A type whose fields were specified to be preserved via a XML descriptor has none.
```
marek-safar, line 4: The wording is a bit strange here

src/linker/WarningCodes.md
```diff
@@ -0,0 +1,74 @@
...
+
+---
+
+### IL2005
+A dependency member that was specified via the `PreserveDependency` attribute could not be resolved.
+
+---
+
+### IL2006
+The linker found an unrecognized reflection access pattern.
```
marek-safar, line 29: What does it mean to the developer?

src/linker/WarningCodes.md
```diff
@@ -0,0 +1,74 @@
...
+
+---
+
+### IL2007
+An assembly that was specified for external customization was not found.
```
marek-safar, line 34: I don't think the developers will know what `external customization` means

src/linker/WarningCodes.md
```diff
@@ -0,0 +1,74 @@
...
+
+---
+
+### IL2009
+A method which was specified for custom substitution was not found.
+
+---
+
+### IL2010
+The value specified for a method to be substituted for could not be casted to the required type.
```
marek-safar, line 49: This is not useful. The explanation should be explicit about what is possible

src/linker/WarningCodes.md
```diff
@@ -0,0 +1,74 @@
...
+
+---
+
+### IL2014
+A field was specified for substitution but no value to be substituted for was given.
+
+---
+
+### IL2015
+The value specified for a field to be substituted for could not be casted to the required type.
```
marek-safar, line 74: Same as above be explicit about what is allowed

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -616,7 +618,8 @@ protected virtual void MarkUserDependency (MemberReference context, CustomAttrib
 			if (MarkDependencyField (td, member, new DependencyInfo (DependencyKind.PreservedDependency, ca)))
 				return;
 
-			_context.LogMessage (MessageImportance.High, $"Could not resolve dependency member '{member}' declared in type '{td.FullName}'");
+			// TODO: Add origin, basically the place where `PreserveDependencyAttribute` is used.
```
vitek-karas, line 621: Either create an issue or resolve the TODO... ideally we should not have TODOs in the code - they are not really useful (And nobody tracks them).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -632,7 +634,8 @@ protected virtual void MarkUserDependency (MemberReference context, CustomAttrib
 			if (MarkDependencyField (td, member, new DependencyInfo (DependencyKind.PreservedDependency, ca)))
 				return;
 
-			_context.LogMessage (MessageImportance.High, $"Could not resolve dependency member '{member}' declared in type '{td.FullName}'");
+			_context.LogMessage (MessageContainer.CreateWarningMessage ($"Could not resolve dependency member '{member}' declared in type '{td.FullName}'",
+				2005, origin: MessageOrigin.TryGetOrigin (context.Resolve (), 0)));
```
sbomer, line 638: I'm not sure we should report IL offset 0 for the attribute - if we can't get the line number of the attribute itself, I would rather just mention that it came from the attribute on the method.

src/linker/Linker/LoggingReflectionPatternRecorder.cs
```diff
@@ -44,7 +44,8 @@ public void RecognizedReflectionAccessPattern (MethodDefinition sourceMethod, In
 
 		public void UnrecognizedReflectionAccessPattern (MethodDefinition sourceMethod, Instruction reflectionMethodCall, IMetadataTokenProvider accessedItem, string message)
 		{
-			_context.LogMessage (MessageImportance.Low, message);
+			_context.LogMessage (MessageContainer.CreateWarningMessage (message, 2006, "Unrecognized reflection pattern",
+				MessageOrigin.TryGetOrigin (sourceMethod, reflectionMethodCall != null ? reflectionMethodCall.Offset : 0)));
```
sbomer, line 48: What are the cases where the reflectionMethodCall is null? Similar to above, I'd avoid reporting offset 0 if we don't have the actual origin.

src/linker/Linker/MessageContainer.cs
```diff
@@ -7,6 +7,8 @@
 
 namespace Mono.Linker
 {
+	// See a list of error and warning codes at https://github.com/mono/linker/blob/master/src/linker/ErrorAndWarningCodes.md
```
vitek-karas, line 10: Can you please put this into the XML comments on the specific methods? Basically we want people to know about this list whenever they're reporting warning/error.

vitek-karas: The ask was to add XML style comments onto the public methods which create the messages. So on `MessageContainer.CreateErrorMessage` and so on. The comment should look like
```C#
This should be an XML comment:
/// <summary>Creates an error message</summary>
/// <param name="text">The content of the error message</param>
/// <param name="code">The code of the error - please see https://github.com/mono/linker/blob/master/src/linker/ErrorAndWarningCodes.md for the list of error/warning codes and possibly add a new one there.</param>
public static MessageContainer CreateErrorMessage (string text, int code, ...)
```

When written like this VS and VS Code will show this in a nice way in the UI in intellisense.

vitek-karas: Referring tot he list in places where we use the codes is not really needed - I would remove those comments.

The desire is such that if somebody new wants to write an error message and types the `MessageContainer.CreateErrorMessage(` into the IDE it will lpopup descriptive text which will guide them how to create the best error message possible - including making sure to use the right code and to register it in the md file when necessary.

PR #1109, opened 2020-04-16
[feature/dataflow] Make the dataflow json file general purpose

MichalStrehovsky:
This makes the file format more general purpose and renames the command line argument to signify that.

I didn't attempt to extract the parser out into a reusable class. We can do that when there's actual need to use this for something else - this is just to make sure the outside interface and structure is general purpose.


src/linker/Linker.Dataflow/JsonFlowAnnotationSource.cs
```diff
@@ -64,10 +64,18 @@ public DynamicallyAccessedMemberKinds GetThisParameterAnnotation (MethodDefiniti
 			return 0;
 		}
 
-		static DynamicallyAccessedMemberKinds ParseKinds(string s)
+		static DynamicallyAccessedMemberKinds ParseKinds(JsonElement attributes)
 		{
-			// Enum.Parse accepts a comma as a separator for Flags
-			return (DynamicallyAccessedMemberKinds)Enum.Parse (typeof (DynamicallyAccessedMemberKinds), s);
+			foreach (var attribute in attributes.EnumerateObject ()) {
+				if (attribute.Name == "System.Runtime.CompilerServices.DynamicallyAccessedMembersAttribute") {
```
vitek-karas, line 70: I think it would be nice to not require the `Attribute` suffix, just like the C# syntax doesn't require it,

src/linker/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
```
vitek-karas, line 1: Please exclude this file from the change.... We should really add this to `.gitignore`.

MichalStrehovsky:
Addressed the feedback. When I was adding the sample, it became apparent that the existing format is super-inflexible.

* It's not possible to express assembly or module level attributes
* It's not possible to express nested types
* It's not possible to express attributes on types
* It's not possible to express attributes on methods

The JSON is short, but that's where all the advantages end.

src/linker/Linker.Dataflow/JsonFlowAnnotationSource.cs
```diff
@@ -105,7 +113,9 @@ private void Initialize(LinkContext context, string jsonFile)
 						foreach (var member in typeElement.Value.EnumerateObject ()) {
 							string memberName = member.Name;
 
-							if (member.Value.ValueKind == JsonValueKind.Object) {
+							// Let's assume fields cannot have '(' in their names and cross our fingers nobody names
```
vitek-karas, line 116: We can lookup the name on the type - if it's a field/event/property we would know right away. I don't have a problem with using parenthesis here, just that the comment feels weird.

PR #1116, opened 2020-04-17
Add a <resource> element to the substitution file to allow removal of…

vargaz:
… managed resources.


src/linker/Linker.Steps/BodySubstituterStep.cs
```diff
@@ -71,6 +71,7 @@ void ProcessAssemblies (XPathNodeIterator iterator)
 		void ProcessAssembly (AssemblyDefinition assembly, XPathNodeIterator iterator)
 		{
 			ProcessTypes (assembly, iterator.Current.SelectChildren ("type", ""));
+			ProcessResources (assembly, iterator.Current.SelectChildren ("resource", ""));
```
marek-safar, line 74: This feels a bit more like hack to me for one specific use case. Could you instead add it to a custom step we build for wasm only and remove it there?

vargaz: This is a general approach which allows moving this logic to wasm instead of putting it into the linker where it doesn't really belong.

vargaz: Also, this pass is called substitution pass, but what it really does is it makes modifications to the assembly based on a config file. Removing resources is just another kind of modification. I can add an 'action' attribute with a 'remove' value to make this more explicit.

marek-safar:
@monojenkins backport mono-2020-02

vargaz:
So is this good for master as well ?

marek-safar:
ok, let's get this in but it needs more for master
- Add tests and documentation for this feature
- Make the removal explicit by using `<resource name="FullName" content="remove">`

vargaz:
Will add those.

vargaz:
Added some docs and an action attribute. Not sure how to add a test, the existing substitution tests are all done using attributes on methods/fields.

docs/data-formats.md
```diff
@@ -204,6 +204,16 @@ The `initialize` attribute is optional and when not specified the code to set th
 </linker>
 ```
 
+### Remove embedded resources
+
+```xml
+<linker>
+  <assembly fullname="Assembly">
+    <resource name="Resource" action="remove" />
```
marek-safar, line 212: I think this will be best used with the feature condition. Could you add it to this PR?

PR #1117, opened 2020-04-17
[feature/dataflow] Deduplicate ArrayBuilder helper class

vitek-karas:


src/linker/Linker.Dataflow/ArrayBuilder.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
```
MichalStrehovsky, line 1: License header

marek-safar: This should be in Mono.Linker there is nothing .Dataflow specific and can be used elsewhere

src/linker/Linker.Dataflow/ArrayBuilder.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Mono.Linker.Dataflow
+{
+	struct ArrayBuilder<T>
+	{
+		private List<T> _list;
+
+		public void Add (T value) => (_list ??= new List<T> ()).Add (value);
+
+		public bool Any (Predicate<T> callback) => _list == null ? false : _list.Exists (callback);
```
marek-safar, line 12:
```suggestion
		public bool Any (Predicate<T> callback) => _list?.Exists (callback) == true;
```

Issue #970, opened 2020-02-27
Add pattern recognition for calls to RuntimeHelpers.RunClassConstructor


vitek-karas:
Linker may decide to trim static `.cctor` from a class if it doesn't find a need to keep it. Calling `System.Runtime.CompilerServices.RuntimeHelpers.RunClassConstructor` is in such case unsafe as it requires the static `.cctor` to be present to work correctly.

Linker should be able to analyze simple patterns calling the `RunClassConstructor` and include the necessary static `.cctor`. The minimum pattern to support should be:
``` C#
RuntimeHelpers.RunClassConstructor(typeof(MyClass));
```

Once implement we can basically revert this change in CoreLib dotnet/runtime#31718.
This is a preferred solution to the changes suggested in #966.

marek-safar:
Interestingly this API is used more often than I thought (2 hits in libraries, 1 in coreclr).

PR #1121, opened 2020-04-17
Support RunClassConstructor in the linker

tlakollo:
This fixes https://github.com/mono/linker/issues/970

-Add Support for RuntimeHelpers.RunClassConstructor
-Add test for RunClassConstructor
-Change MarkStaticConstructor to internal in order to be able to call it
from ReflectionMethodBodyScanner


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -926,6 +945,25 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
 						break;
 
+					//
+					// System.Runtime.CompilerServices.RuntimeHelpers
+					//
+					// RunClassConstructor (RuntimeTypeHandle type)
+					//
+					case IntrinsicId.RuntimeHelpers_RunClassConstructor: {
+							reflectionContext.AnalyzingPattern ();
+							var parameters = calledMethod.Parameters;
```
vitek-karas, line 955: Is this used for anything?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -926,6 +945,25 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+							foreach (var TypeHandleValue in methodParams [0].UniqueValues ()) {
+								if (TypeHandleValue is RuntimeTypeHandleValue runtimeTypeHandleValue) {
+									_markStep.MarkStaticConstructor (runtimeTypeHandleValue.TypeRepresented, new DependencyInfo (DependencyKind.AccessedViaReflection, reflectionContext.MethodCalling));
+									reflectionContext.RecordHandledPattern ();
+								} else {
+									RequireDynamicallyAccessedMembers (ref reflectionContext, DynamicallyAccessedMemberKinds.DefaultConstructor, TypeHandleValue, calledMethod.Parameters [0]);
```
vitek-karas, line 961: This is not correct - we want static .cctor - not the default .ctor. We don't have an annotation to require static .cctor. I think we should just issue a warning in this case.

test/Mono.Linker.Tests.Cases/Reflection/RunClassConstructorUsedViaReflection.cs
```diff
@@ -0,0 +1,101 @@
...
+		static Type FindType ()
+		{
+			return null;
+		}
+
+		[Kept]
+		static void TestDataFlowType ()
+		{
+			Type type = FindType ();
+			RuntimeHelpers.RunClassConstructor (type.TypeHandle);
```
vitek-karas, line 50: This should warn - we have no idea what the type is. BTW: This should warn even with your changes.

The catch here is that the test won't validate the warnings unless the method has some related attribute on it (this is to avoid having to add the `Unrecognized` attributes all over the place). So to get the necessary verification you need to either add one of the `Unrecognized` attributes or add `RecognizedReflectionAccessPattern()` - this basically means "I expect this to not issue any warnings".

test/Mono.Linker.Tests.Cases/Reflection/RunClassConstructorUsedViaReflection.cs
```diff
@@ -0,0 +1,101 @@
...
+		static void TestNonKeptStaticConstructor ()
+		{
+			var a = new NonKeptStaticConstructorClass();
+		}
+
+		[Kept]
+		static void TestNull ()
+		{
+			Type type = null;
+			RuntimeHelpers.RunClassConstructor (type.TypeHandle);
```
vitek-karas, line 37: This should also have `RecognizedReflectionAccessPattern` attribute to validate that it doesn't issue a warning.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -926,6 +945,25 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+					// System.Runtime.CompilerServices.RuntimeHelpers
+					//
+					// RunClassConstructor (RuntimeTypeHandle type)
+					//
+					case IntrinsicId.RuntimeHelpers_RunClassConstructor: {
+							reflectionContext.AnalyzingPattern ();
+							var parameters = calledMethod.Parameters;
+							foreach (var TypeHandleValue in methodParams [0].UniqueValues ()) {
+								if (TypeHandleValue is RuntimeTypeHandleValue runtimeTypeHandleValue) {
+									_markStep.MarkStaticConstructor (runtimeTypeHandleValue.TypeRepresented, new DependencyInfo (DependencyKind.AccessedViaReflection, reflectionContext.MethodCalling));
```
marek-safar, line 958: Should this really in `DependencyKind.AccessedViaReflection` category?

vitek-karas: We could create a new one I guess. But it's not that different from something like `type.GetConstructor().Invoke()`. I do think it's still reflection, even if the method/type is not in the reflection namespace.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -926,6 +948,28 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
 						break;
 
+					//
+					// System.Runtime.CompilerServices.RuntimeHelpers
+					//
+					// RunClassConstructor (RuntimeTypeHandle type)
+					//
+					case IntrinsicId.RuntimeHelpers_RunClassConstructor: {
+							reflectionContext.AnalyzingPattern ();
+							foreach (var TypeHandleValue in methodParams [0].UniqueValues ()) {
```
vitek-karas, line 958: Local variable should start with lower case `TypeHandleValue` -> `typeHandleValue`.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -926,6 +948,28 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+								if (TypeHandleValue is RuntimeTypeHandleValue runtimeTypeHandleValue) {
+									_markStep.MarkStaticConstructor (runtimeTypeHandleValue.TypeRepresented, new DependencyInfo (DependencyKind.AccessedViaReflection, reflectionContext.MethodCalling));
+									reflectionContext.RecordHandledPattern ();
+								} else if (TypeHandleValue == NullValue.Instance)
+									reflectionContext.RecordHandledPattern ();
+								else {
+									reflectionContext.RecordUnrecognizedPattern ($"A {GetValueDescriptionForErrorMessage (TypeHandleValue)} " +
+										$"is passed into the {GetMetadataTokenDescriptionForErrorMessage (reflectionContext.MethodCalling)}. " +
```
vitek-karas, line 966:
```suggestion
										$"is passed into the {GetMetadataTokenDescriptionForErrorMessage (reflectionContext.MethodCalled.Parameters[0])}. " +
```

vitek-karas: The original message was not making much sense as it was talking about the code calling into `RunClassConstructor`. The problem is passing unknown value to the first parameter of `RunClassConstructor` so that's what we should report.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -926,6 +948,28 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+										$"It's not possible to guarantee that these requirements are met by the application.");
```
vitek-karas, line 967: Maybe we should be a bit more specific - just a suggestion: "It's not possible to guarantee availability of the target static constructor." or something like that.

test/Mono.Linker.Tests.Cases/Reflection/RunClassConstructorUsedViaReflection.cs
```diff
@@ -0,0 +1,161 @@
...
+		[RecognizedReflectionAccessPattern]
+		static void TestNullWithType ()
+		{
+			Type type = null;
+			RuntimeHelpers.RunClassConstructor (type.TypeHandle);
+		}
+
+		[Kept]
+		[RecognizedReflectionAccessPattern]
+		static void TestNullWithRuntimeTypeHandle ()
```
vitek-karas, line 46: How is this different from `TestNullWithType` - I don't mind having more test cases, but these two are nearly identical.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -926,6 +948,28 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+					case IntrinsicId.RuntimeHelpers_RunClassConstructor: {
+							reflectionContext.AnalyzingPattern ();
+							foreach (var typeHandleValue in methodParams [0].UniqueValues ()) {
+								if (typeHandleValue is RuntimeTypeHandleValue runtimeTypeHandleValue) {
+									_markStep.MarkStaticConstructor (runtimeTypeHandleValue.TypeRepresented, new DependencyInfo (DependencyKind.AccessedViaReflection, reflectionContext.MethodCalling));
+									reflectionContext.RecordHandledPattern ();
+								} else if (typeHandleValue == NullValue.Instance)
+									reflectionContext.RecordHandledPattern ();
+								else {
+									reflectionContext.RecordUnrecognizedPattern ($"A {GetValueDescriptionForErrorMessage (typeHandleValue)} " +
```
MichalStrehovsky, line 965: Nit/not really necessary: If we _really_ wanted - the `MemberKinds.Constructors` dataflow annotation keeps static constructors because they can be accessed with `typeof(Foo).GetConstructors(BindingFlags.NonPublic | BindingFlags.Static)`

vitek-karas: Good point. For now I would leave it as is. Supporting the annotation would mean to flow it through the get_TypeHandle call as well. Until we hit a use case I'm perfectly fine supporting just the simple constant cases.

test/Mono.Linker.Tests.Cases/Reflection/RunClassConstructorUsedViaReflection.cs
tlakollo: Basically one of them is exercising more than one RecognizedReflectionAccessPattern but they are basically having the same behavior in code, I guess it makes more sense in the IfElse case. I will leave only those.

PR #1130, opened 2020-04-21
Speed up linker by 10%

MichalStrehovsky:
...using one weird trick.

Cecil uses file I/O to read assemblies, but for our use cases this means we're spending a ton of time context switching between kernel mode and user mode.

Memory mapped I/O is more efficient for these access patterns.

The change is bigger than I would want it to be because:
* Loading assemblies from disk is not centralized in linker (and in fact there are more extensibility points that encourage everyone to do their own loading)
* Cecil violates OOP principles by giving certain kinds of streams preferential treatment (`ModuleDefinition.FileName` is useless if the assembly wasn't opened with the blessed stream kind and Cecil doesn't have a way to provide it through other means).

### DIFF
--- a/src/linker/Linker/DirectoryAssemblyResolver.cs
+++ b/src/linker/Linker/DirectoryAssemblyResolver.cs
@@ -16,7 +16,7 @@ namespace Mono.Linker {
 
 		readonly Collection<string> directories;
 
-		public readonly Dictionary<AssemblyDefinition, string> AssemblyToPath = new Dictionary<AssemblyDefinition, string> ();
+		protected readonly Dictionary<AssemblyDefinition, string> assemblyToPath = new Dictionary<AssemblyDefinition, string> ();
 
 		readonly List<MemoryMappedViewStream> viewStreams = new List<MemoryMappedViewStream> ();
 
@@ -45,19 +45,17 @@ namespace Mono.Linker {
 			if (parameters.AssemblyResolver == null)
 				parameters.AssemblyResolver = this;
 
-			FileStream fileStream = null;
-			MemoryMappedFile mappedFile = null;
 			MemoryMappedViewStream viewStream = null;
 			try {
 				// Create stream because CreateFromFile(string, ...) uses FileShare.None which is too strict
-				fileStream = new FileStream (file, FileMode.Open, FileAccess.Read, FileShare.Read, 4096, false);
-				mappedFile = MemoryMappedFile.CreateFromFile (
+				using var fileStream = new FileStream (file, FileMode.Open, FileAccess.Read, FileShare.Read, 4096, false);
+				using var mappedFile = MemoryMappedFile.CreateFromFile (
 					fileStream, null, fileStream.Length, MemoryMappedFileAccess.Read, HandleInheritability.None, true);
 				viewStream = mappedFile.CreateViewStream (0, 0, MemoryMappedFileAccess.Read);
 
 				AssemblyDefinition result = ModuleDefinition.ReadModule (viewStream, parameters).Assembly;
 
-				AssemblyToPath.Add (result, file);
+				assemblyToPath.Add (result, file);
 
 				viewStreams.Add (viewStream);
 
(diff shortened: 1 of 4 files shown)


MichalStrehovsky:
Opening as draft mostly because I would like to get some feedback on the extensibility APIs that allow people to provide their own AssemblyDefinitions that they opened through whatever other means outside the primary AssemblyResolver. Do we need to keep supporting this or can we cut it off?

Also, I think I broke things outside `FEATURE_ILLINK`, but illink.sln has all tests passing.

src/linker/Linker/DirectoryAssemblyResolver.cs
```diff
@@ -15,6 +16,10 @@ public abstract class DirectoryAssemblyResolver : IAssemblyResolver {
 
 		readonly Collection<string> directories;
 
+		public readonly Dictionary<AssemblyDefinition, string> AssemblyToPath = new Dictionary<AssemblyDefinition, string> ();
```
marek-safar, line 19: Could this field not be exposed directly?

src/linker/Linker/DirectoryAssemblyResolver.cs
```diff
@@ -40,7 +45,34 @@ protected AssemblyDefinition GetAssembly (string file, ReaderParameters paramete
 			if (parameters.AssemblyResolver == null)
 				parameters.AssemblyResolver = this;
 
-			return ModuleDefinition.ReadModule (file, parameters).Assembly;
+			FileStream fileStream = null;
```
marek-safar, line 48: Might be more readable with just `using` syntax

src/linker/Linker/DirectoryAssemblyResolver.cs
```diff
@@ -40,7 +45,34 @@ protected AssemblyDefinition GetAssembly (string file, ReaderParameters paramete
...
+				viewStream = mappedFile.CreateViewStream (0, 0, MemoryMappedFileAccess.Read);
+
+				AssemblyDefinition result = ModuleDefinition.ReadModule (viewStream, parameters).Assembly;
+
+				AssemblyToPath.Add (result, file);
+
+				viewStreams.Add (viewStream);
+
+				// We transferred the ownership of the viewStream to the collection.
+				viewStream = null;
```
marek-safar, line 65: why is this needed?

mrvoorhe:
Nothing about the API jumps out as being problematic for us to extend.

src/linker/Linker/DirectoryAssemblyResolver.cs
MichalStrehovsky: It's needed because we want to `Dispose` the `viewStream` if `ModuleDefinition.ReadModule` above failed for any reason - such assembly is unusable and we might as well throw away the mapping now. If accessing the assembly from the module succeeded, we do not want to dispose it because it's used by Cecil and the `viewStreams` collection now manages the lifetime.

It's also a reason why I'm not using a `using` - the `viewStream` handling is odd. If you see a better way to do this, I can change it.

MichalStrehovsky:
> I'm guessing the 10% is on Windows only, right?

I didn't measure Linux because I only have a WSL 1.0 instance that is famous for bad file I/O perf; any measurements would be meaningless. If someone who has a Linux box can help out on getting numbers for this, that would save me a bunch of time.

src/linker/Linker/DirectoryAssemblyResolver.cs
marek-safar: You might consider using `using` for `fileStream` and `mappedFile` and keep the only `viewStream` in custom finally to make the logic/flow more obvious

src/linker/Linker/DirectoryAssemblyResolver.cs
```diff
@@ -89,6 +121,13 @@ public void Dispose ()
 
 		protected virtual void Dispose (bool disposing)
 		{
+			if (disposing) {
```
marek-safar, line 124: Should there be also finalizer?

MichalStrehovsky: The class is not holding unmanaged resources so I don't think we need one.

src/linker/Linker/DirectoryAssemblyResolver.cs
```diff
@@ -7,6 +7,7 @@
 using System.IO;
 using Mono.Collections.Generic;
 using Mono.Cecil;
+using System.IO.MemoryMappedFiles;
```
vitek-karas, line 10: Nit: Sort usings

PR #1133, opened 2020-04-22
Preserve MonoPInvokeCallbackAttribute.

vargaz:


marek-safar:
Why hack it this way? You could just do it to linker descriptor with required=false ?

Also, this is wrong branch for wasm

vargaz:
Considering that MonoPInvokeAttribute has no set assembly/namespace, handling it this way would be desirable.

mrvoorhe:
For what it's worth we've had this code in our linker for a long time

```
            if (ca.AttributeType.Name == "MonoPInvokeCallbackAttribute")
            {
                value = true;
                return true;
            }
```

marek-safar:
> Considering that MonoPInvokeAttribute has no set assembly/namespace, handling it this way would be desirable.

Unfortunately not, this attribute is flying around for example here https://docs.microsoft.com/en-us/dotnet/api/objcruntime.monopinvokecallbackattribute?view=xamarin-ios-sdk-12.

I think as the master is primary for netcore we won't need it here as the attribute (this or similar) will be added to BCL as public type.

PR #1136, opened 2020-04-23
Change ReflectionPatternRecorder to handle fields and parameters

mateoatr:
The introduced changes will help us to keep track of the IL instruction where a specific reflection call is made (useful for reporting warnings and errors) as well as to represent field and parameter dataflow.


src/linker/Linker/MessageOrigin.cs
```diff
@@ -19,6 +22,23 @@ public MessageOrigin (string fileName, int sourceLine = 0, int sourceColumn = 0)
 			SourceColumn = sourceColumn;
 		}
 
+		public static MessageOrigin? TryGetOrigin (IMemberDefinition sourceMethod, int reflectionCallOffset)
```
marek-safar, line 25: Should be `ilOffset` instead of `reflectionCallOffset`. Similarly `reflectionCall` should have a better name as there is nothing reflection specific on the sequence point here

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -51,7 +51,19 @@ public void ScanAndProcessReturnValue (MethodBody methodBody)
 			if (MethodReturnValue != null) {
 				var requiredMemberKinds = _flowAnnotations.GetReturnParameterAnnotation (methodBody.Method);
 				if (requiredMemberKinds != 0) {
-					var reflectionContext = new ReflectionPatternContext (_context, methodBody.Method, methodBody.Method, 0);
+					var accessedMember = methodBody.Method as IMetadataTokenProvider;
+					var offset = 0;
+					if (MethodReturnValue is LoadFieldValue field) {
+						accessedMember = field.Field;
+						var loadInstruction = methodBody.Method.Body.Instructions
```
marek-safar, line 58: This will be pretty slow

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -51,7 +51,19 @@ public void ScanAndProcessReturnValue (MethodBody methodBody)
...
+							.Where (i => (i.OpCode == OpCodes.Ldfld || i.OpCode == OpCodes.Ldsfld) && i.Operand.ToString () == field.Field.FullName);
+						if (loadInstruction != null)
+							offset = loadInstruction.First ().Offset;
+					} else if (MethodReturnValue is MethodParameterValue param) {
+						accessedMember = methodBody.Method.Parameters [param.ParameterIndex - 1];
+					}
+
+					var reflectionContext = new ReflectionPatternContext (_context, methodBody.Method, accessedMember, offset);
```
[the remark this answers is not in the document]
mateoatr: What should we do for properties with explicit accessors? Should the `sourceMethod` in these cases be `get_<Property>()` and the `accessedMember` the underlying field?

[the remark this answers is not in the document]
vitek-karas: I would not expect any special treatment for properties - those with implicit accessors should be auto-annotated correctly by the `FlowAnnotations` class  and so in theory should never produce a warning. Those with explicit accessors are just normal methods and should behave as such.

mateoatr:
The following examples illustrate the desired behavior for different reflection patterns:

```C#
[return: DynamicallyAccessedMembers (Constructors)]
Type Foo () { return T; }
```
**Source Method**: `Foo`

**Accessed Member**: method return type of `Foo`

---
```C#
class T {
    void Foo () { var x = typeof (T); }
}
```

**Source Method**: `Foo`

**Accessed Member**: method `Type.GetTypeFromHandle()`

---
```C#
[Bar (someType)]
void Foo () { }

class BarAttribute : Attribute {
    public BarAttribute ([DynamicallyAccessedMembers (Constructors)] Type t) { }
}
```

**Source Method**: method `Foo`

**Accessed Member**: parameter `t`

---
```C#
[Bar (Baz)]
void Foo () { }

class BarAttribute : Attribute {
    public BarAttribute () { }

    [DynamicallyAccessedMembers (Constructors)]
    public Type type;
}

class Baz { }
```

**Source Method**: ctor of `Bar`

**Accessed Member**: type `Baz`

---
```C#
Type Foo {
    [return: DynamicallyAccessedMembers (Constructors)]
    get {
        return _field;
    }
}
```

**Source Method**: method `get_Foo`

**Accessed Member**: field `_field`

---
```C#
void Foo () { Bar (someType); }

void Bar (Type t) {
    return t;
}
```

**Source Method**: method `Bar`

**Accessed Member**: parameter `t`

vitek-karas:
```C#
Type Foo {
    [return: DynamicallyAccessedMembers (Constructors)]
    get {
        return _field;
    }
}
```

Source Method: method `get_Foo`

Accessed Member: ~~field _field~~ return type of method `get_Foo`.

The field case is something like this instead:
```C#
[DynamicallyAccessedMembers(Constructors)]
private Type _myType;

public void Method() {
    _myType = typeof(Foo);
}
```

Source Method: method `Method`

Accessed Member: field `_myType`

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -51,14 +51,19 @@ public void ScanAndProcessReturnValue (MethodBody methodBody)
 			if (MethodReturnValue != null) {
 				var requiredMemberKinds = _flowAnnotations.GetReturnParameterAnnotation (methodBody.Method);
 				if (requiredMemberKinds != 0) {
-					var reflectionContext = new ReflectionPatternContext (_context, methodBody.Method, methodBody.Method, 0);
+					IMetadataTokenProvider accessedMember = methodBody.Method.MethodReturnType;
+					if (MethodReturnValue is LoadFieldValue field)
+						accessedMember = field.Field;
+					else if (MethodReturnValue is MethodParameterValue param)
+						accessedMember = methodBody.Method.Parameters[param.ParameterIndex - 1];
+					var reflectionContext = new ReflectionPatternContext (_context, methodBody.Method, accessedMember, null);
```
vitek-karas, line 59:
```suggestion
					var reflectionContext = new ReflectionPatternContext (_context, methodBody.Method, methodBody.Method.MethodReturnType, null);
```

The problematic spot is not the source of the value but the place where that value is used and needs to meet some requirements - in this case it's being returned from a method which has requirements on the return value.

This is sort of complicated, the experience I would like to see for this is something like:
* The warning gets reported on the return statement
* The warning message contains where the value came from
* The warning contains info about the fact that the value does not fullfill the requirements on the return value

You can try this with nullable - make a method which returns non-nullable reference type and try to return null - and see what the C# analyzer prints out - we should get something very similar.

src/linker/Linker.Dataflow/ReflectionPatternContext.cs
```diff
@@ -26,15 +26,19 @@ struct ReflectionPatternContext : IDisposable
...
+		public ReflectionPatternContext (LinkContext context, MethodDefinition sourceMethod, IMetadataTokenProvider memberWithRequirements,
+			Instruction instruction = null,
+			int? parameterIndex = null)
 		{
 			_context = context;
 			SourceMethod = sourceMethod;
-			AccessedMember = accessedMember;
+			MemberWithRequirements = memberWithRequirements;
 			Instruction = instruction;
+			ParameterIndex = parameterIndex;
```
vitek-karas, line 41: Do we actually use this for anything? I think we should remove it.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -51,12 +51,11 @@ public void ScanAndProcessReturnValue (MethodBody methodBody)
...
 				if (requiredMemberKinds != 0) {
-					IMetadataTokenProvider accessedMember = methodBody.Method.MethodReturnType;
-					if (MethodReturnValue is LoadFieldValue field)
-						accessedMember = field.Field;
-					else if (MethodReturnValue is MethodParameterValue param)
-						accessedMember = methodBody.Method.Parameters[param.ParameterIndex - 1];
-					var reflectionContext = new ReflectionPatternContext (_context, methodBody.Method, accessedMember, null);
+					int? parameterIndex = null;
+					if (MethodReturnValue is MethodParameterValue param)
+						parameterIndex = param.ParameterIndex - 1;
```
vitek-karas, line 56: I don't think we need this...

src/linker/Linker.Dataflow/ReflectionPatternContext.cs
vitek-karas, line 41: Do we actually use this for anything? I think we should remove it.

PR #1138, opened 2020-04-23
Ensure consistent sources formatting

marek-safar:
- Run `dotnet-format -f path` on the sources (modulo legacy one) and tests
- Introduce CI lane Lint to validate against any future incompatibilities


.editorconfig
```diff
@@ -4,11 +4,38 @@ end_of_line = "lf"
...
 csharp_new_line_before_else = false
 csharp_new_line_before_catch = false
 csharp_new_line_before_finally = false
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
 csharp_space_after_keywords_in_control_flow_statements = true
 csharp_space_between_method_declaration_name_and_open_parenthesis = true
+csharp_space_between_method_declaration_name_and_open_parenthesis = true
```
sbomer, line 18: Redundant line

.editorconfig
```diff
@@ -4,11 +4,38 @@ end_of_line = "lf"
 
 [*.cs]
 indent_style = tab
+indent_size = 4
 csharp_new_line_before_open_brace = types,methods
```
sbomer, line 8: Should this be just methods? https://www.mono-project.com/community/contributing/coding-guidelines/ says classes get curly braces on the same line.

.editorconfig
```diff
@@ -4,11 +4,38 @@ end_of_line = "lf"
...
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_between_query_expression_clauses = true
+
 csharp_space_after_keywords_in_control_flow_statements = true
 csharp_space_between_method_declaration_name_and_open_parenthesis = true
+csharp_space_between_method_declaration_name_and_open_parenthesis = true
 csharp_space_between_method_call_name_and_opening_parenthesis = true
-csharp_space_before_open_square_brackets = true
\ No newline at end of file
+csharp_space_before_open_square_brackets = false
```
sbomer, line 20: Should this stay `true` to follow mono coding guidelines?

.editorconfig
marek-safar: Fixed

.editorconfig
marek-safar: We haven't been very consistent on this one and both exist. I reviewed also monodevelop version which is using the same `types,methods` and it's also same to the default VS so this setting makes sense to me.

.editorconfig
marek-safar: I'm open to tweaking this but there are unfortunately no different settings for array declaration and indexes. The original Mono coding guideline has different spacing for each which makes it difficult to set.

Mono guidelines

```csharp
int[] a = new object [10];
```

I went with the minimal changes to the sources and the matching default VS settings.

MichalStrehovsky:
> which I personally don't like very much, but that's a different story

+1 :). There's some value in using formatting rules that the rest of the world uses.

PR #1143, opened 2020-04-25
Fixes regression caused by missing space after `--custom-data` argument

marek-safar:
Introduced in f8d216d8611265af051163b07bccc6fdb83f3571

### DIFF
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -331,7 +331,7 @@ namespace ILLink.Tasks
 			if (clearInitLocals) {
 				args.AppendLine ("--enable-opt clearinitlocals");
 				if (ClearInitLocalsAssemblies?.Length > 0) {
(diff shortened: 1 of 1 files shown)


src/ILLink.Tasks/LinkTask.cs
```diff
@@ -331,7 +331,7 @@ protected override string GenerateResponseFileCommands ()
 			if (clearInitLocals) {
 				args.AppendLine ("--enable-opt clearinitlocals");
 				if (ClearInitLocalsAssemblies?.Length > 0) {
-					args.AppendFormat ($"--custom-data ClearInitLocalsAssemblies={ClearInitLocalsAssemblies}");
+					args.AppendFormat ($"--custom-data ClearInitLocalsAssemblies={ClearInitLocalsAssemblies} ");
```
akoeplinger, line 334: maybe use AppendLine instead?

marek-safar: I hate the newlines in the log they are a pain to deal with (merging to unblock darc update)

PR #1154, opened 2020-04-30
Move documentation to doc/ folder and fill some gaps in the documentation

marek-safar:

PR #1155, opened 2020-04-30
Use MessageContainer for logging errors

mateoatr:
I'm moving some messages to be logged using the `MessageContainer`. I added a new exception class `LinkerErrorException` so that whenever an error is found, instead of simply printing it, we throw and gracefully stop the linker. The rationale for this is that whenever an error is found, the user should take action and solve it so that the linker can do its work. I'd expect all future errors (minus internal ones) to use this exception.


src/linker/Linker.Steps/BlacklistStep.cs
```diff
@@ -55,7 +55,7 @@ protected override void Process ()
 					steps_to_add.Push (GetResolveStep (name));
 				} catch (XmlException ex) {
 					/* This could happen if some broken XML file is included. */
-					Context.LogMessage ($"Error processing {name}: {ex}");
+					throw new LinkerErrorException (MessageContainer.CreateErrorMessage ($"Error processing {name}: {ex}", 1003));
```
sbomer, line 58: Generally speaking I think it's good to recover from errors if we have a reasonable way to do so, as long as an error still fails the build. That way you can fix more problems in one iteration.

src/linker/Linker/Driver.cs
```diff
@@ -674,6 +674,13 @@ public bool Run (ILogger customLogger = null)
...
 			try {
 				p.Process (context);
+			} catch (Exception ex) {
+				Console.Error.WriteLine ("The illinker exited with error:");
+				if (ex is LinkerErrorException le) {
+					context.LogMessage (le.MessageContainer);
+				} else {
+					Console.Error.WriteLine (ex);
+				}
 			} finally {
```
sbomer, line 684: I think we need to return false here to produce a failing exit code (which signals failure to the MSBuild task, among other things).

src/linker/Linker.Steps/BlacklistStep.cs
marek-safar: Yes, this is a bad pattern. We want to produce as many errors possible during the linker run not abort on the first one.

vitek-karas: I have to apologize to Mateo - this was my suggestion. I thought that we use warnings in places where linker can safely recover from it. If there are places where we want the build to fail, but linker can safely recover, then I agree we should not throw there.

src/linker/Linker/LinkerErrorException.cs
```diff
@@ -1,22 +1,29 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Text;
 
 namespace Mono.Linker
 {
 	public class LinkerErrorException : Exception
 	{
+		public static bool FoundErrors { get; private set; } = false;
```
vitek-karas, line 7: This is weird. We should track whether we reported errors at the place where we actually report them - which is the context/logging. It should not matter if I use `LinkerErrorException` or some other way to produce an error message, all should count the same.

mateoatr: With this I'm assuming that the `LinkerErrorException` is the only only way we produce error messages, forcing us to wrap the errors in a `MessageContainer`. I went this way to not have to add a `context.FoundError = true` everywhere we report errors.

vitek-karas: I understand, but at the same time you now have to explicitly call `FoundErrors` whenever we report non-fatal error. It would be much cleaner to do this at the context level - basically look at the `MessageContainer` and if it's an error - remember that. Nothing else.

src/linker/Linker.Steps/BodySubstituterStep.cs
```diff
@@ -52,13 +48,11 @@ bool ShouldProcessSubstitutions (XPathNavigator nav)
...
 			if (string.IsNullOrEmpty (value)) {
-				Context.LogMessage (MessageContainer.CreateErrorMessage ($"Feature {feature} does not specify a \"featurevalue\" attribute", 1001));
-				return false;
+				throw new LinkerErrorException (MessageContainer.CreateErrorMessage ($"Failed to process XML substitution: '{_xmlDocumentLocation}'. Feature {feature} does not specify a \"featurevalue\" attribute", 1001));
 			}
 
 			if (!bool.TryParse (value, out bool bValue)) {
-				Context.LogMessage (MessageContainer.CreateErrorMessage ($"Unsupported non-boolean feature definition {feature}", 1002));
-				return false;
+				throw new LinkerErrorException (MessageContainer.CreateErrorMessage ($"Failed to process XML substitution: '{_xmlDocumentLocation}'. Unsupported non-boolean feature definition {feature}", 1002));
```
sbomer, line 55: Nice improvement to the message!

src/linker/Linker/Driver.cs
```diff
@@ -674,11 +674,20 @@ public bool Run (ILogger customLogger = null)
 
 			try {
 				p.Process (context);
+			} catch (Exception ex) {
+				Console.Error.WriteLine ("The ILlinker exited with error:");
```
sbomer, line 678: I think we should use either "IL Linker" to refer to it in general, or "illink" when talking about the specific tool.

```suggestion
				Console.Error.WriteLine ("illink exited with error:");
```

src/linker/Linker.Steps/BodySubstituterStep.cs
```diff
@@ -52,13 +48,11 @@ bool ShouldProcessSubstitutions (XPathNavigator nav)
 
 			var value = GetAttribute (nav, "featurevalue");
 			if (string.IsNullOrEmpty (value)) {
-				Context.LogMessage (MessageContainer.CreateErrorMessage ($"Feature {feature} does not specify a \"featurevalue\" attribute", 1001));
```
sbomer, line 55: Is there a reason we need to throw here? I'd prefer to recover from this error if possible - same with the other one in this file.

src/linker/Linker/Driver.cs
```diff
@@ -674,11 +674,20 @@ public bool Run (ILogger customLogger = null)
...
+						Console.Error.WriteLine (le.InnerException);
+				} else {
+					Console.Error.WriteLine (ex);
+				}
 			} finally {
 				context.Tracer.Finish ();
 			}
 
-			return true;
+			return !context.FoundErrors;
```
sbomer, line 690: I think we need to make sure this returns false for `InternalErrorException`.

src/linker/ErrorAndWarningCodes.md
```diff
@@ -1,4 +1,139 @@
 # IL Linker Error Codes
+
+### IL1001
+```
+Failed to process XML substitution: 'XML document location'. Feature 'feature' does not specify a "featurevalue" attribute
```
sbomer, line 5: https://github.com/mono/linker/pull/1154 changed the format to put the error text on the same line as the error code. I personally prefer the style you have here since it avoids long lines in the heading, but I don't feel strongly about - either way, we should be consistent.

sbomer: @marek-safar is there a particular reason you prefer them to be on the same line?

sbomer: Oh, and https://github.com/mono/linker/pull/1159/files is further tweaking the format.

src/linker/Linker.Steps/BodySubstituterStep.cs
```diff
@@ -52,13 +48,11 @@ bool ShouldProcessSubstitutions (XPathNavigator nav)
...
-				return false;
+				throw new LinkerErrorException (MessageContainer.CreateErrorMessage ($"Failed to process XML substitution: '{_xmlDocumentLocation}'. Feature {feature} does not specify a \"featurevalue\" attribute", 1001));
```
marek-safar, line 51: This should be error

src/linker/Linker.Steps/BodySubstituterStep.cs
marek-safar: Should be error (or even warning)

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -479,9 +479,7 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 							break;
 
 						default:
-							Debug.Fail ("Unsupported member type");
-							reflectionContext.RecordUnrecognizedPattern ($"Reflection call '{calledMethod.FullName}' inside '{callingMethodBody.Method.FullName}' is of unexpected member type.");
-							break;
+							throw new InternalErrorException ($"Reflection call '{calledMethod.FullName}' inside '{callingMethodBody.Method.FullName}' is of unexpected member type.");
```
marek-safar, line 482: I don't think this should be fatal, probably not even error

src/linker/Linker/LinkerErrorException.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace Mono.Linker
+{
+	public class LinkerErrorException : Exception
```
marek-safar, line 5: I'd suggest to name it `LinkerFatalErrorException` or something like that to make it clear it's only for unrecoverable situations

src/linker/ErrorAndWarningCodes.md
```diff
@@ -1,4 +1,139 @@
...
+---
+
+### IL1011
+```
+Could not find constructor on 'type'
+```
+
+---
+
+### IL1012
```
marek-safar, line 96: What about sharing the same code with IL1009?

src/linker/ErrorAndWarningCodes.md
```diff
@@ -1,4 +1,139 @@
...
+```
+Missing predefined 'System.Runtime.CompilerServices.DisablePrivateReflectionAttribute' type
+```
+
+---
+
+### IL1013
+```
+Could not find constructor on '{disablePrivateReflection.FullName}'
```
marek-safar, line 105: What about sharing the same code with IL1011?

src/linker/ErrorAndWarningCodes.md
marek-safar: Yep, designer review ;-)

src/linker/Linker/Driver.cs
mateoatr: I used "IL Linker" to be consistent with the error messages.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
mateoatr: Because of how this switch-case is nested, it's an invariant that the `IntrinsicId` will be matched. In other words, if we ever hit this line is because something really weird is going on.

vitek-karas: Agreed - this is basically assert, but based on other discussions we've switched all asserts to internal errors - which are fatal.

src/linker/Linker/Driver.cs
```diff
@@ -674,11 +674,20 @@ public bool Run (ILogger customLogger = null)
 
 			try {
 				p.Process (context);
+			} catch (Exception ex) {
+				Console.Error.WriteLine ("IL Linker exited with error:");
+				if (ex is LinkerFatalErrorException le) {
+					context.LogMessage (le.MessageContainer);
```
marek-safar, line 680: Why log only `LinkerFatalErrorException` exceptions?

src/linker/Linker/Driver.cs
```diff
@@ -674,11 +674,20 @@ public bool Run (ILogger customLogger = null)
...
+					if (le.InnerException != null)
```
marek-safar, line 681: This should not be necessary if `LinkerFatalErrorException::ToString` does not work please fix it there

src/linker/Linker/Driver.cs
```diff
@@ -674,11 +674,20 @@ public bool Run (ILogger customLogger = null)
 
 			try {
 				p.Process (context);
+			} catch (Exception ex) {
+				Console.Error.WriteLine ("IL Linker exited with error:");
```
marek-safar, line 678:
```suggestion
				Console.Error.WriteLine ("IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues");
```

src/linker/Linker/InternalErrorException.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace Mono.Linker
+{
+	public class InternalErrorException : Exception
+	{
+		public InternalErrorException (string message, LinkContext context)
+			: base (message)
+		{
+			context.FoundErrors = true;
```
marek-safar, line 10: I don't think this is needed here

src/linker/Linker/InternalErrorException.cs
```diff
@@ -0,0 +1,19 @@
...
+		}
+
+		public InternalErrorException (string message, Exception innerException, LinkContext context)
+			: base (message, innerException)
+		{
+			context.FoundErrors = true;
```
marek-safar, line 16: I don't think this is needed here

src/linker/Linker/Driver.cs
```diff
@@ -674,11 +674,20 @@ public bool Run (ILogger customLogger = null)
...
+				if (ex is LinkerFatalErrorException le) {
+					context.LogMessage (le.MessageContainer);
+					if (le.InnerException != null)
+						Console.Error.WriteLine (le.InnerException);
+				} else {
+					Console.Error.WriteLine (ex);
+				}
```
marek-safar, line 685: Please `return false;` from the handler

src/linker/Linker/Driver.cs
vitek-karas: I think the difference is that `LinkerFatalErrorException` has the `MessageContainer` so we can do better job reporting the error to MSBuild (as it will have a specific code, category and so on).
For the same reason we need to handle inner exceptions as well - although maybe we should handle them by producing an error to the log as well.

src/linker/Linker/Driver.cs
```diff
@@ -674,11 +674,20 @@ public bool Run (ILogger customLogger = null)
...
 			try {
 				p.Process (context);
+			} catch (Exception ex) {
+				Console.Error.WriteLine ("IL Linker exited with error:");
+				if (ex is LinkerFatalErrorException le) {
+					context.LogMessage (le.MessageContainer);
+					if (le.InnerException != null)
+						Console.Error.WriteLine (le.InnerException);
+				} else {
+					Console.Error.WriteLine (ex);
```
vitek-karas, line 684: Why don't we write the exception to the error log. As is this would not report it as MSBuild error - right? That seems wrong. (Not counting that it breaks redirecting logs via interface).

In general I don't think linker should write directly to Console ever. But we can fix that in a separate change.
Since linker supports the `ILogger` all logging should go through that.

src/linker/Linker/Driver.cs
marek-safar: That's right but we could create MessageContainer for all other exceptions from Message or ToString and log them too, right now the text goes to Console only.

We should also add check to LinkerFatalErrorException that the MessageContainer is of Error type.

src/linker/Linker/Driver.cs
sbomer: If we use the logging interface for both `LinkerFatalErrorException` and `InternalErrorException`, what's the difference between the two? I thought the point of the internal exception was that we don't create error codes for it because they should never be seen by users (and if they are, it's a bug). We just want to see a stacktrace if the linker crashes.

Anything logged to stderr will currently go to the MSBuild output with MSBuild `MessageImportance.High`.

src/linker/Linker/LinkerFatalErrorException.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace Mono.Linker
+{
+	public class LinkerFatalErrorException : Exception
```
sbomer, line 5: Maybe it'd be good to add some comments about what this is for, the intended use, and how it's different from InternalErrorException - and same for InternalErrorException.

src/linker/Linker/Driver.cs
mateoatr: I was thinking about `InternalErrorException` as a kind of dead end in the sense that we'd only see it if something went really wrong and we needed to kill the linker (there's a bug; we reached some code that wasn't supposed to be reached). Giving it a second thought, I think we might simply rely on `Debug.Assert` and existing exceptions for this, and add a new error code asking the users to file an issue with the stack trace.

sbomer: Using `InternalErrorException` or existing exceptions+`Debug.Assert` makes sense to me for the dead end behavior that you described. I don't think we need a new code though - I think of the error codes as user-facing errors - so I wouldn't force the internal errors to go through the same path that reports errors about the inputs. That might just be me - it sounds like @vitek-karas might have a different opinion. As long as the intended use of the exceptions is clear, I'm fine with either way.

vitek-karas: I don't know what the msbuild experience will be if we don't report an error code. As long as there's a clear message saying "Internal error, please report it to us" and then the details in the "Error Window" in VS, I'm fine with whichever solution we pick.

sbomer: If the linker fails during publish, this is what it currently looks like:
Screen Shot 2020-05-06 at 12 49 12 PM
The log mentioned in that popup doesn't have anything useful, but the output has the text written by the linker ("Expecting a parameter"). I haven't checked, but I think the new errors will show up similarly - they'll be errors in MSBuild, but won't make it to the VS error window if run during publish. I'll check this behavior later.

vitek-karas: Hmm - just wondering - what is the warning experience in that case? Does it even show up?

sbomer: I checked - yeah, it's the same for warnings and errors:
Screen Shot 2020-05-06 at 3 17 07 PM
Errors will additionally produce that popup.

sbomer: We should plan as if we were running during build. If I set it up to run during build, I see:
- if the task fails (should never happen)
Screen Shot 2020-05-06 at 3 24 04 PM

- if the linker fails (should never happen)
Screen Shot 2020-05-06 at 3 26 20 PM
Screen Shot 2020-05-06 at 3 26 30 PM

docs/error-codes.md
```diff
@@ -47,6 +47,10 @@ error and warning codes.
 
 - There was an error writing the linked assembly 'output'. An exception with more details is printed.
 
+#### `IL1012`: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues
+
+- There was an unexpected error while linking. An exception with more details is printed.
```
marek-safar, line 52: Would be nice to add more info where to get the log from or what kind of details to log

src/linker/Linker/Driver.cs
```diff
@@ -675,19 +675,23 @@ public bool Run (ILogger customLogger = null)
...
-						Console.Error.WriteLine (le.InnerException);
+				if (ex is LinkerFatalErrorException lex) {
+					context.LogMessage (lex.MessageContainer);
 				} else {
-					Console.Error.WriteLine (ex);
+					string exceptionMessage = ex.Message ?? string.Empty;
+					context.LogMessage (MessageContainer.CreateErrorMessage ($"IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues \n{exceptionMessage}", 1012));
 				}
+
+				string innerMessage = ex.InnerException?.Message ?? null;
```
marek-safar, line 685: I think we need to print whole .ToString () in this case somewhere instead. Showing one level of inner-exception does not really help

src/linker/Linker/LinkerFatalErrorException.cs
```diff
@@ -6,15 +6,30 @@ public class LinkerFatalErrorException : Exception
...
 
+		/// <summary>
+		/// Represents a known error that occurred during link time which is solvable by the user.
+		/// </summary>
+		/// <param name="message">Error message with a description of what went wrong</param>
 		public LinkerFatalErrorException (MessageContainer message)
 			: base (message.ToString ())
 		{
+			if (message.Category != MessageCategory.Error)
+				throw new ArgumentException ("'LinkerFatalErrorException' ought to be used for errors only");
```
marek-safar, line 17: Use `nameof`

docs/error-codes.md
```diff
@@ -49,7 +49,7 @@ error and warning codes.
 
 #### `IL1012`: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues
 
-- There was an unexpected error while linking. An exception with more details is printed.
+- There was an unexpected error while linking. An exception with more details is printed to the MSBuild log, please share the stack trace seen in there with the IL Linker team to further investigate the cause and possible solution.
```
sbomer, line 52:
```suggestion
- There was an unexpected error while linking. An exception with more details is printed to the MSBuild log. Please share this stack trace with the IL Linker team to further investigate the cause and possible solution.
```

src/linker/Linker/Driver.cs
```diff
@@ -674,6 +673,16 @@ public bool Run (ILogger customLogger = null)
 
 			try {
 				p.Process (context);
+			} catch (Exception ex) {
+				if (ex is LinkerFatalErrorException lex) {
+					context.LogMessage (lex.MessageContainer);
+					if (ex.InnerException != null)
```
marek-safar, line 679: I'm still don't like this one. Every Exception can have InnerException why to show only this one and why to show only 1 level of depth? If we are printing to console, let's print everything and for all exceptions.

Otherwise LGTM

PR #1170, opened 2020-05-05
Adapt to the new DynamicallyAccessedMembersAttribute in the linker

### LINKED ISSUE #1149: Respect the new design of DynamicallyAccessedMembersAttribute
The API review for `DynamicallyAccessedMembersAttribute` changed a few things on the design of the attribute. For example, `Public` and `NonPublic` members were separated in the enum, and `All` and `None` values were added.

We should update the linker to support the new design.

See https://github.com/dotnet/runtime/issues/33861 for discussion and https://github.com/dotnet/runtime/pull/35387 for the new attribute.

vitek-karas:
The attribute changed namespace and the DynamicallyAccessedMemberTypes enum has been renamed and has differnet values in it now.

Most notable changes:
* The DynamicallyAccessedMemberTypes values are similar to binding flags in that they specify Public/NonPublic which are exclusive
* There's a new DynamicallyAccessedMemberTypes.All value which is supposed to be a special value (not a simple superset of all the other bits) and tells the linker to keep the type intact.

This also changes the behavior of the linker when recognizing calls to reflection methods like GetField/GetMethod. Previously the overloads which didn't take BindingFlags would consider all Public and NonPublic and Static/Instance members. That's not how the reflection APIs behave, these overloads only consider Public Static/Instance members. This change modifies the linker to do the same.
As such this changes the linker behavior to potentially include less, but to also potentially break some apps which by accident worked due to the linker including more than necessary. Note that the linker will not warn if the member being asked for was not found (or doesn't have the right binding flags). The reasoning is that without linker the app would get a null from the reflection API in such case, with the linker it will still get null, so the behavior of the app doesn't change and as such there's no reason to raise a warning.

Lot of test changes to adapt to the new enum.

Fixes #1149.


vitek-karas:
Interesting observation related to this:
`DynamicallyAccessedMemberTypes.All` will not mark the entire type, including base types and nested types. This exposes possible hole in the enum around nested types. If the `DynamicallyAccessedMemberTypes.PublicNestedTypes` is used (for example) we will only "mark" the nested types - which includes the bare minimum about the type. So it will make the `Type.GetNestedType` work, but it's very unlikely that the calling code will end up working overall as the type as marked is effectively useless (no methods, no .ctors, nothing really).

It's true that in the current state the user code would get lot of warnings anyway as any action on the type returned by `Type.GetNestedType` will result in a warning. So in that sense the behavior is still correct, but it makes the `PublicNestedTypes`/`NonPublicNestedType` enum values borderline useless.

MichalStrehovsky:
> It's true that in the current state the user code would get lot of warnings anyway as any action on the type returned by `Type.GetNestedType` will result in a warning. So in that sense the behavior is still correct, but it makes the `PublicNestedTypes`/`NonPublicNestedType` enum values borderline useless.

The semantic that would make sense is to keep the nested type at the `All` level. The `GetNested*` APIs are used rarely enough that the size overhead won't matter (close to 0% on all overloads at https://apisof.net/), the result would be usable and easy to reason about (one can call `GetNested*` recursively and still get something useful out), and it would work for the one "meaningful" use case of `GetNested` in CoreLib (event source enumerates fields on nested types with a magic name).

test/Mono.Linker.Tests.Cases.Expectations/Support/DynamicallyAccessedMembersAttribute.cs
```diff
@@ -2,34 +2,132 @@
...
+	///
+	/// This enumeration has a <see cref="FlagsAttribute"/> attribute that allows a
+	/// bitwise combination of its member values.
+	/// </summary>
+	/// <remarks>
+	/// This is a copy of the enum definition in the framework at https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/DynamicallyAccessedMemberTypes.cs
+	/// </remarks>
 	[Flags]
-	public enum DynamicallyAccessedMemberKinds
+	public enum DynamicallyAccessedMemberTypes
```
marek-safar, line 17: Please move this type to a new file and create new folder called System.Diagnostics.CodeAnalysis

src/linker/Linker.Dataflow/DynamicallyAccessedMemberTypes.cs
```diff
@@ -0,0 +1,96 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Mono.Linker.Dataflow
```
marek-safar, line 7: I think this should go into System.Diagnostics.CodeAnalysis namespace as well

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1425,45 +1432,74 @@ string GetMetadataTokenDescriptionForErrorMessage (IMetadataTokenProvider target
...
+				results.Add (DynamicallyAccessedMemberTypes.PublicNestedTypes);
 
 			if (results.Count == 0)
-				return "None";
+				return DynamicallyAccessedMemberTypes.None.ToString ();
 
 			return string.Join (" | ", results.Select (r => r.ToString ()));
 		}
+
+		DynamicallyAccessedMemberTypes GetDynamicallyAccessedMemberTypesFromBindingFlagsForConstructors (BindingFlags bindingFlags) =>
```
marek-safar, line 1485: All these should be static

src/linker/Linker.Dataflow/DynamicallyAccessedMemberTypes.cs
MichalStrehovsky: Won't we just get a conflict once the linker project is updated to .NET 5? At that point it will technically be redundant and could be deleted, but linker already carries its own copy of the `BindingFlags` enum instead of using the one from the platform... it would be nice to have a consistent approach.

src/linker/Linker.Dataflow/JsonFlowAnnotationSource.cs
```diff
@@ -61,24 +61,24 @@ public DynamicallyAccessedMemberKinds GetThisParameterAnnotation (MethodDefiniti
...
 
-			return 0;
+			return DynamicallyAccessedMemberTypes.None;
 		}
 
-		static DynamicallyAccessedMemberKinds ParseKinds (JsonElement attributes)
+		static DynamicallyAccessedMemberTypes ParseKinds (JsonElement attributes)
 		{
 			foreach (var attribute in attributes.EnumerateObject ()) {
 				if (attribute.Name == "System.Runtime.CompilerServices.DynamicallyAccessedMembers") {
```
MichalStrehovsky, line 70: Should we rename this too?

src/linker/Linker.Dataflow/DynamicallyAccessedMemberTypes.cs
marek-safar: As the number of these framework copied files is growing it would be better to have a consistent approach and I'd suggest to place them under the same name/namespace to avoid confusing why they existing and if they are different to the original version. The code won't be redundant but we will just if-def it out for NET5. I'm fine to exclude BindingFlags as well. I think the original issue was that on some old versions of .NET Framework not all members were available.

vitek-karas:
For the nested types issues I filed https://github.com/mono/linker/issues/1174.
I won't make any changes to the behavior in this PR.

PR #1172, opened 2020-05-06
Fix build and improve auto-formatting

vitek-karas:
Fixes command line build on windows in the analyzer project (latest Arcade has conflicts when building net core and net desktop projects at the same time. since we don't need desktop I removed it)

Fixes editorconfig to correctly specify LF as EOL character so that it works on Windows.

Added running the lint (auto-formatter) in command line build. This uses the same command line as the CI system, so running the comand line build should fix any problems before submitting PR.


build.sh
```diff
@@ -14,3 +14,4 @@ done
 
 scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
 "$scriptroot/eng/common/build.sh" --projects "$scriptroot/illink.sln" --build --restore $@
+"$scriptroot/eng/dotnet.sh" tool run dotnet-format --check --verbosity diagnostic -f . --exclude src/analyzer,src/tuner,external
```
akoeplinger, line 17: I think this will still run the format even if there are build issues which is confusing. would be good to only do that if the previous command succeeded.

akoeplinger: we should probably also add an option to not do `--check` so it actually fixes the issues so that I don't need to manually apply suggestions.

marek-safar: I'd suggest adding a new scripts `lint.sh` and `lint.cmd` which do the formating and call .sh one from CI script with additional `--check` argument

PR #1175, opened 2020-05-06
Cleanup the illink.sln

vitek-karas:
It had two test folders with tests distributed between them and it had a duplicate link to the main linker project.

Now all tests are in the "test" folder (just like on disk) and every project is there only once.


illink.sln
```diff
@@ -21,15 +19,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "analyzer", "analyzer", "{AA
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "analyzer", "src\analyzer\analyzer.csproj", "{96182221-C5C4-436D-9BE0-EC499F9BAF17}"
 EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{AC840E16-5507-436F-B9B1-46DB20FD395D}"
-EndProject
-Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "linker", "linker", "{80AE2045-E62C-4CD6-93E6-8ED34011B457}"
-EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Mono.Linker", "src\linker\ref\Mono.Linker.csproj", "{C7338775-E3F2-413D-82FD-C5B5B1455B19}"
```
sbomer, line 28: This is for the ref assembly - I think it should stay.

illink.sln
```diff
@@ -21,15 +19,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "analyzer", "analyzer", "{AA
...
-EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{C2969923-7BAA-4FE4-853C-F670B0D3C6C8}"
```
sbomer, line 22: Could we get rid of the entry for the test directory too?

illink.sln
```diff
@@ -21,15 +19,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "analyzer", "analyzer", "{AA
 EndProject
```
sbomer, line 19: Could we remove the analyzer directory entry?

illink.sln
vitek-karas: It would mean putting all projects in a flat list - which I would like to avoid - I like having the tests in a visual subfolder - it makes it easier to separate product from tests.

illink.sln
vitek-karas: Similar to tests - this is a supporting tool, not the product (or tests) so I personally like it in a separate subfolder. But if most people don't want it there, sure...

illink.sln
vitek-karas: Yeah - I noticed now - thanks! I'll add it back into a "ref" folder.

illink.sln
sbomer: Makes sense, thanks! (I don't use VS, didn't know this changes how folders are displayed)>

Issue #1118, opened 2020-04-17
Make the supplementary custom attribute annotations general purpose


MichalStrehovsky:
We need to make sure all things can be annotated:

* Assemblies
* Nested types
* Methods
* etc.

marek-safar:
I'd consider using xml for this. The verbosity and the flexibility (e.g. adding comments) of the format might be actually advantages in this case.

tlakollo:
The following it's a proposal on how I think the XML would look like:

```
<Annotations>
  <Assembly Name="Test">
    <Attribute Name="AssemblyAttr" Parameters="SomeParameter"/>
    <Module>
      <Attribute Name="ModuleAttr" Parameters="SomeParameter"/>
    </Module>
    <Namespace Name="System.Collections">
      <Type Name="FunctionName">
        <Attribute Name="TypeAttr" Parameters="SomeParameter"/>
        <Field Name="FieldName">
          <Attribute Name="FieldAttr" Parameters="SomeParameter"/>
        </Field>
        <Method Name="MethodName" Signature="">
          <Attribute Name="MethodAttr" Parameters="SomeParameter"/>
          <Parameter></Parameter>
            <Attribute Name="ParameterAttr" Parameters="SomeParameter"/>
          <ReturnParameter>
            <Attribute Name="ReturnParameterAttr" Parameters="SomeParameter"/>
          </ReturnParameter>
        </Method>
      </Type>
      <Type Name="NestedType">
        <Type Name="NestedType2">
          <Attribute Name="AllowMultipleAttr"/>
          <Attribute Name="AllowMultipleAttr" Parameters="SomeParameter"/>
        </Type>
      </Type>
    </Namespace>
  </Assembly>
</Annotations>
```

marek-safar:
I think the attribute description needs more thoughts. The proposed

```
        <Attribute Name="TypeAttr" Parameters="SomeParameter"/>
```

does not cover all cases (e.g. Properties and multiple parameters) and it does not allow to specify the attribute name fully. Something like this might work better

```xml
        <Attribute FullName="TypeAttr" AssemblyName="Optional">
			<Arguments>
				<Argument>Value</Argument>
			</Arguments>
			<Fields>
				<Field Name="Name">Value</Field>
			</Fields>
			<Properties>
				<Property Name="Name">Value</Property>
			</Properties>
		</Attribute>
```

vitek-karas:
Fixed by https://github.com/mono/linker/pull/1176.

PR #1176, opened 2020-05-06
XmlAnnotation

tlakollo:
Making custom attribute annotations general purpose
- Move the current implementation to xml format
- Support nested types
- Aligned to the DynamicallyAccessedMemberTypes

What is not supported yet
- Support for attributes in assemblies
- Support for attributes in modules
- Process Namespace and name instead of fullname types
- Support for attributes directly on the type
- Support for attributes in events

This is for #1118


src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+						return Annotation;
+			}
+
+			return DynamicallyAccessedMemberTypes.None;
+		}
+
+		static DynamicallyAccessedMemberTypes ParseKinds (ArrayBuilder<Attribute> attributes)
+		{
+			foreach (var attribute in attributes.ToArray ()) {
+				if (attribute.attributeName == "System.Runtime.CompilerServices.DynamicallyAccessedMembers" && attribute.arguments.Count == 1) {
```
MichalStrehovsky, line 94: The namespace for this is now `System.Diagnostics.CodeAnalysis`.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+			foreach (MethodDefinition method in type.Methods)
+				if (name == method.Name)
+					ProcessMethodChildren (type, method, iterator);
+		}
+
+		protected static MethodDefinition GetMethod (TypeDefinition type, string signature)
+		{
+			if (type.HasMethods)
+				foreach (MethodDefinition meth in type.Methods)
+					if (signature == GetMethodSignature (meth, false))
```
MichalStrehovsky, line 503: This won't match if there's e.g. a whitespace difference. It would be nice if it matched.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+		protected virtual void ProcessProperty (TypeDefinition type, XPathNodeIterator iterator)
+		{
+#if !FEATURE_ILLINK
+			if (IsExcluded (iterator.Current))
+				return;
+#endif
+
+			string value = GetSignature (iterator.Current);
+			if (!String.IsNullOrEmpty (value))
+				ProcessPropertySignature (type, value, GetAccessors (iterator.Current), iterator);
```
MichalStrehovsky, line 548: There's no test coverage for this. While yes, technically one could have properties with the same name that only differ by their type, I'm not sure if we really need to have syntax to express this. Mainstream .NET languages don't support this.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+				ProcessField (type, iterator);
+		}
+
+		protected virtual void ProcessField (TypeDefinition type, XPathNodeIterator iterator)
+		{
+#if !FEATURE_ILLINK
+			if (IsExcluded (iterator.Current))
+				return;
+#endif
+			string value = GetSignature (iterator.Current);
```
MichalStrehovsky, line 347: There's no test coverage for this. While yes, technically one could have fields with the same name that only differ by their type, I'm not sure if we really need to have syntax to express this. Mainstream .NET languages don't support this.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+		}
+
+		static bool IsTypePattern (string fullname)
+		{
+			return fullname.IndexOf ("*") != -1;
+		}
+
+		static Regex CreateRegexFromPattern (string pattern)
+		{
+			return new Regex (pattern.Replace (".", @"\.").Replace ("*", "(.*)"));
```
MichalStrehovsky, line 227: I can't find discussion that motivated adding support for `*` wildcard - did someone ask for this?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+			{
+				attributeName = _attributeName;
+				arguments = _arguments;
+				fields = _fields;
+				properties = _properties;
+			}
+		}
+
+#if !FEATURE_ILLINK
+		protected virtual bool IsExcluded (XPathNavigator nav)
```
MichalStrehovsky, line 675: For the protected/virtual members, what extensibility to you expect?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+
+namespace Mono.Linker.Dataflow
+{
+	class XmlFlowAnnotationSource : IFlowAnnotationSource
+	{
+		readonly Dictionary<MethodDefinition, AnnotatedMethod> _methods = new Dictionary<MethodDefinition, AnnotatedMethod> ();
+		readonly Dictionary<PropertyDefinition, DynamicallyAccessedMemberTypes> _properties = new Dictionary<PropertyDefinition, DynamicallyAccessedMemberTypes> ();
+		readonly Dictionary<FieldDefinition, DynamicallyAccessedMemberTypes> _fields = new Dictionary<FieldDefinition, DynamicallyAccessedMemberTypes> ();
+
+		static readonly string _signature = "signature";
```
MichalStrehovsky, line 22: I don't generally find introducing constants for things that only occur once very useful. Could you inline them, or at least make them `const string` instead of `static readonly string`?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+			PropertyDefinition property = GetProperty (type, signature);
+			if (property != null) {
+				ArrayBuilder<Attribute> attributes = ProcessAttributes (iterator.Current.SelectChildren ("attribute", _ns));
+				DynamicallyAccessedMemberTypes propertyAnnotation = ParseKinds (attributes);
+				if (propertyAnnotation != DynamicallyAccessedMemberTypes.None)
+					_properties[property] = propertyAnnotation;
+			}
+		}
+
+		void ProcessPropertyName (TypeDefinition type, string name, string[] accessors, XPathNodeIterator iterator)
```
MichalStrehovsky, line 566: The `accessors` are not used anywhere. What would we use this for?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+		}
+
+		static DynamicallyAccessedMemberTypes ParseKinds (ArrayBuilder<Attribute> attributes)
+		{
+			foreach (var attribute in attributes.ToArray ()) {
+				if (attribute.attributeName == "System.Runtime.CompilerServices.DynamicallyAccessedMembers" && attribute.arguments.Count == 1) {
+					foreach (var argument in attribute.arguments.ToArray ()) {
+						if (argument == string.Empty)
+							break;
+						return (DynamicallyAccessedMemberTypes) Enum.Parse (typeof (DynamicallyAccessedMemberTypes), argument);
```
vitek-karas, line 98: This will throw if the value is not a valid enum value - probably leading to a rather nasty error experience (not sure but we probably don't recover from this - worth a try). In any case, this should be reported as a recoverable error (or maybe even a warning? not sure - ideally it should match other error behaviors in our XML parsing logic).

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+			return DynamicallyAccessedMemberTypes.None;
+		}
+
+		void Initialize ()
+		{
+			XPathNavigator nav = _document.CreateNavigator ();
+
+			// This step can be created with XML files that aren't necessarily
+			// annotations descriptor files. So bail if we don't have a <annotations> element.
+			if (!nav.MoveToChild ("annotations", _ns))
```
vitek-karas, line 111: The comment is wrong for this - probably copy/paste leftover.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+			}
+		}
+
+		protected virtual void ProcessAssemblies (LinkContext context, XPathNodeIterator iterator)
+		{
+			while (iterator.MoveNext ()) {
+				AssemblyDefinition assembly = GetAssembly (context, GetAssemblyName (iterator.Current));
+
+				if (assembly == null) {
+					context.LogMessage ($"Assembly {GetAssemblyName (iterator.Current).Name} couldn't be resolved");
```
vitek-karas, line 127: Use the new logging APIs (`MessageContainer`).

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+					continue;
+				}
+
+				ProcessAssembly (assembly, iterator);
+			}
+		}
+
+		protected virtual void ProcessAssembly (AssemblyDefinition assembly, XPathNodeIterator iterator)
```
vitek-karas, line 135: Why are these `protected virtual` ?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+		}
+
+		ArrayBuilder<string> GetAttributeProperties (XPathNodeIterator iterator)
+		{
+			ArrayBuilder<string> properties = new ArrayBuilder<string> ();
+			while (iterator.MoveNext ()) {
+				properties.Add (iterator.Current.Value);
+			}
+			return properties;
+		}
```
vitek-karas, line 183: These methods are identical - why not have just one?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+		void ProcessTypeChildren (TypeDefinition type, XPathNavigator nav)
+		{
+			if (nav.HasChildren) {
+				ProcessSelectedFields (nav, type);
+				ProcessSelectedMethods (nav, type);
+				ProcessSelectedProperties (nav, type);
+			}
+		}
+
+		void ProcessSelectedFields (XPathNavigator nav, TypeDefinition type)
```
vitek-karas, line 297: Why so many tiny methods - For example we can easily merge `ProcessSelectedFields` and `ProcessFields` into one - no reason to keep them as separate methods.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+		void ProcessSelectedProperties (XPathNavigator nav, TypeDefinition type)
+		{
+			XPathNodeIterator properties = nav.SelectChildren ("property", _ns);
+			if (properties.Count == 0)
+				return;
+
+			ProcessProperties (type, properties);
+		}
+
+		static TypePreserve GetTypePreserve (XPathNavigator nav)
```
vitek-karas, line 324: THis is not used anywhere - same for the `_preserve` const.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+				if (field.Name == name) {
+					ArrayBuilder<Attribute> attributes = ProcessAttributes (iterator.Current.SelectChildren ("attribute", _ns));
+					DynamicallyAccessedMemberTypes fieldAnnotation = ParseKinds (attributes);
+					if (fieldAnnotation != DynamicallyAccessedMemberTypes.None)
+						_fields[field] = fieldAnnotation;
+				}
+			}
+		}
+
+		protected static FieldDefinition GetField (TypeDefinition type, string signature)
```
vitek-karas, line 382: why protected?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
vitek-karas: Let's not support this....

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+			FieldDefinition field = GetField (type, signature);
+			if (field != null) {
+				ArrayBuilder<Attribute> attributes = ProcessAttributes (iterator.Current.SelectChildren ("attribute", _ns));
+				DynamicallyAccessedMemberTypes fieldAnnotation = ParseKinds (attributes);
+				if (fieldAnnotation != DynamicallyAccessedMemberTypes.None)
+					_fields[field] = fieldAnnotation;
+			}
+		}
+
+		void ProcessFieldName (TypeDefinition type, string name, XPathNodeIterator iterator)
```
vitek-karas, line 367: Again - why separate method - this makes the code much harder to read - typically if something is extracted into a method one expects it to be called from multiple places. These are not.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+				ProcessMethodChildren (type, meth, iterator);
+		}
+
+		void ProcessMethodChildren (TypeDefinition type, MethodDefinition method, XPathNodeIterator iterator)
+		{
+			ArrayBuilder<Attribute> attributes = ProcessAttributes (iterator.Current.SelectChildren ("attribute", _ns));
+			ArrayBuilder<(string, ArrayBuilder<Attribute>)> parameterAnnotations = ProcessParameters (type,
+				method, iterator.Current.SelectChildren ("parameter", _ns));
+			ArrayBuilder<ArrayBuilder<Attribute>> returnParameterAnnotations = ProcessReturnParameters (type,
+				method, iterator.Current.SelectChildren ("returnparameter", _ns));
```
vitek-karas, line 434: I would prefer either `return` or `returnParameter` names (personally I like `return` better, but either is fine)

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+		protected virtual void ProcessAssembly (AssemblyDefinition assembly, XPathNodeIterator iterator)
+		{
+#if !FEATURE_ILLINK
+			if (IsExcluded (iterator.Current))
+				return;
+#endif
+			ProcessTypes (assembly, iterator.Current.SelectChildren ("type", _ns));
+		}
+
+		ArrayBuilder<Attribute> ProcessAttributes (XPathNodeIterator iterator)
```
vitek-karas, line 144: This seems to always be called with the `SelectChildren("attribute", _ns)` why not make that part of the method. Just accept the "parent" iterator and call the `.Current.SelectChildren("attribute", _ns)` as part of this method.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+
+			foreach (MethodDefinition method in type.Methods)
+				if (name == method.Name)
+					ProcessMethodChildren (type, method, iterator);
+		}
+
+		protected static MethodDefinition GetMethod (TypeDefinition type, string signature)
+		{
+			if (type.HasMethods)
+				foreach (MethodDefinition meth in type.Methods)
```
vitek-karas, line 502: Please rename `meth` -> `method`

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+		}
+
+		public static string GetMethodSignature (MethodDefinition meth, bool includeGenericParameters)
+		{
+			StringBuilder sb = new StringBuilder ();
+			sb.Append (meth.ReturnType.FullName);
+			sb.Append (" ");
+			sb.Append (meth.Name);
+			if (includeGenericParameters && meth.HasGenericParameters) {
+				sb.Append ("`");
```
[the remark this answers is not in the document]
vitek-karas: We should add a test for this case as well.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+					if (signature == GetMethodSignature (meth, false))
+						return meth;
+
+			return null;
+		}
+
+		public static string GetMethodSignature (MethodDefinition meth, bool includeGenericParameters)
+		{
+			StringBuilder sb = new StringBuilder ();
+			sb.Append (meth.ReturnType.FullName);
```
vitek-karas, line 512: I know this is from the existing format - but do we have to include the return type in the signature? No .NET language allows overrides on return type alone... maybe we could try both (with and without?)

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+		{
+			var value = GetAttribute (nav, "feature");
+			if (string.IsNullOrEmpty (value))
+				return false;
+
+			return _context.IsFeatureExcluded (value);
+		}
+#endif
+
+		public override string ToString ()
```
vitek-karas, line 685: Do we have a use case for this? (other than debugging which is questionable)

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,690 @@
...
+
+				foreach (var (ParamName, Annotation) in ann.ParameterAnnotations)
+					if (ParamName == "this")
+						return Annotation;
+			}
+
+			return DynamicallyAccessedMemberTypes.None;
+		}
+
+		static DynamicallyAccessedMemberTypes ParseKinds (ArrayBuilder<Attribute> attributes)
```
vitek-karas, line 91: This should be called something like GetMemberTypesForDynamicallyAccessedMemberAttribute or something like that. The method does two things - it finds the attribute with the right name and parses its value.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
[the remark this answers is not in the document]
tlakollo: Now it writes the signature without Mangled name and inserting instead <GenericParameters>, notice that XML will now need to use escape sequences to declare Methods using GenericParameters

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: Now it will match no matter the whitespaces

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: I did not change this behavior, I think it should be there at least as a way to keep the format being used in ResolveFromXmlStep

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: Now Im ussing the MessageContainer but had to modify error codes to make them more generic

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: A new warning code was introduced for this

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: I don't think there was a request commit is from 13 years ago https://github.com/mono/linker/commit/a8d7006e829e505c542a2cb77b350b3e208e2d9b

docs/data-formats.md
```diff
@@ -224,22 +224,107 @@ are applied.
...
+        <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
+          <argument>DefaultConstructor</argument>
+        </attribute>
+      </field>
+    </type>
+  </assembly>
+</linker>
+```
+
+### Custom attribute on property field
```
vitek-karas, line 243:
```suggestion
### Custom attribute on property
```

docs/data-formats.md
```diff
@@ -224,22 +224,107 @@ are applied.
...
+
+```xml
+<linker>
+  <assembly fullname="Assembly">
+    <type fullname="Assembly.A">
+      <property name="MyNumericField">
```
vitek-karas, line 249:
```suggestion
      <property name="MyTypeProperty">
```

docs/data-formats.md
```diff
@@ -224,22 +224,107 @@ are applied.
...
-    }
-  }
-}
+### Custom attribute on type field
+
+```xml
+<linker>
+  <assembly fullname="Assembly">
+    <type fullname="Assembly.A">
+      <field name="MyNumericField">
```
vitek-karas, line 233:
```suggestion
      <field name="MyTypeField">
```

docs/data-formats.md
```diff
@@ -224,22 +224,107 @@ are applied.
...
+
+### Custom attribute on method
+
+```xml
+<linker>
+  <assembly fullname="Assembly">
+    <type fullname="Assembly.A">
+      <method signature="System.Void Method1(System.Boolean)">
+        <parameter name="boolParameter">
+          <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
```
vitek-karas, line 267:
```suggestion
      <method signature="System.Void Method1(System.Type)">
        <parameter name="typeParameter">
          <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
```

vitek-karas: The sample should be realistic - we don't allow (rather ignore) these attributes on non-type members.

docs/data-formats.md
```diff
@@ -224,22 +224,107 @@ are applied.
...
+            <argument>DefaultConstructor</argument>
+          </attribute>
+        </parameter>
+      </method>
+      <method signature="System.Boolean Method2()">
```
vitek-karas, line 272:
```suggestion
      <method signature="System.Type Method2()">
```

docs/data-formats.md
```diff
@@ -224,22 +224,107 @@ are applied.
...
+```
+
+### Custom attribute on nested type
+
+```xml
+<linker>
+  <assembly fullname="Assembly">
+    <type fullname="Assembly.A">
+      <type name="NestedType">
+        <property name="MyNumericField">
```
vitek-karas, line 298:
```suggestion
        <property name="MyTypeField">
```

docs/error-codes.md
```diff
@@ -38,17 +38,17 @@ error and warning codes.
...
+#### `IL2008`: Could not resolve type 'type' specified in the 'XML document location'
 
-- The type 'type' in the substitution XML could not be resolved.
+- The type 'type' in the XML could not be resolved.
 
-#### `IL2009`: Could not find method 'method' for substitution
+#### `IL2009`: Could not find method 'method' in type 'type' specified in 'XML document location'
 
-- The method 'method' in the substitution XML could not be resolved.
+- The XML descriptor preserves method 'method' on type 'type', but the method was not found.
```
vitek-karas, line 51: > XML descriptor preserves

If we're to use this in all XML formats, then this is not true.

docs/error-codes.md
```diff
@@ -58,9 +58,9 @@ error and warning codes.
 
 - The value 'action' of the body attribute used in the substitution XML is invalid (the only supported options are `remove` and `stub`).
 
-#### `IL2012`: Could not find field 'field' for substitution
+#### `IL2012`: Could not find field 'field' in type 'type' specified in 'XML document location'
 
-- The field 'field' specified in the substitution XML was not found.
+- The XML descriptor preserves field 'field' on type 'type', but the field was not found.
```
vitek-karas, line 63: Same here - only the linktrim XML will "preserve" the field, the other XML formats do different things.

docs/error-codes.md
```diff
@@ -74,26 +74,22 @@ error and warning codes.
...
 
-#### `IL2020`: Could not find the get accessor of property 'property' in type 'type' specified in 'XML document location'
+#### `IL2018`: Could not find the get accessor of property 'property' in type 'type' specified in 'XML document location'
 
 - The XML descriptor preserves the get accessor of property 'property' on type 'type', but the accessor was not found.
 
-#### `IL2021`: Could not find the set accessor of property 'property' in type 'type' specified in 'XML document location'
+#### `IL2019`: Could not find the set accessor of property 'property' in type 'type' specified in 'XML document location'
+
+- The XML descriptor preserves the set accessor of property 'property' on type 'type', but the accessor was not found.
```
vitek-karas, line 91: And one more

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+	{
+		readonly Dictionary<MethodDefinition, AnnotatedMethod> _methods = new Dictionary<MethodDefinition, AnnotatedMethod> ();
+		readonly Dictionary<PropertyDefinition, DynamicallyAccessedMemberTypes> _properties = new Dictionary<PropertyDefinition, DynamicallyAccessedMemberTypes> ();
+		readonly Dictionary<FieldDefinition, DynamicallyAccessedMemberTypes> _fields = new Dictionary<FieldDefinition, DynamicallyAccessedMemberTypes> ();
+
+		readonly XPathDocument _document;
+		readonly string _xmlDocumentLocation;
+		readonly LinkContext _context;
+
+		public class XmlResolutionException : Exception
```
vitek-karas, line 26: What is this for - I don't see a reason to introduce a new exception type with no additional info in it.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+
+		public DynamicallyAccessedMemberTypes GetReturnParameterAnnotation (MethodDefinition method)
+		{
+			return _methods.TryGetValue (method, out var ann) ? ann.ReturnAnnotation : DynamicallyAccessedMemberTypes.None;
+		}
+
+		public DynamicallyAccessedMemberTypes GetThisParameterAnnotation (MethodDefinition method)
+		{
+			if (_methods.TryGetValue (method, out var ann) && ann.ParameterAnnotations != null) {
+
```
vitek-karas, line 73: Nit: extra new line - please remove.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+		static DynamicallyAccessedMemberTypes GetMemberTypesForDynamicallyAccessedMemberAttribute (ArrayBuilder<Attribute> attributes, LinkContext _context, string _xmlDocumentLocation)
+		{
+			foreach (var attribute in attributes.ToArray ()) {
+				if (attribute.attributeName == "System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers" && attribute.arguments.Count == 1) {
+					foreach (var argument in attribute.arguments.ToArray ()) {
+						if (argument == string.Empty)
+							break;
+						try {
+							return (DynamicallyAccessedMemberTypes) Enum.Parse (typeof (DynamicallyAccessedMemberTypes), argument);
+						} catch (ArgumentException) {
```
vitek-karas, line 91: Instead of try/catch use [TryParse](https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=netcore-3.1)

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+		{
+			XPathNavigator nav = _document.CreateNavigator ();
+
+			if (!nav.MoveToChild ("linker", string.Empty))
+				return;
+
+			try {
+				ProcessAssemblies (_context, nav.SelectChildren ("assembly", string.Empty));
+			} catch (Exception ex) when (!(ex is XmlResolutionException)) {
+				throw new XmlResolutionException (string.Format ("Failed to process XML description: {0}", _document), ex);
```
vitek-karas, line 110: Is this recoverable? I don't know how we treat the exception from the call site. If it's not recoverable I think this should use the new `LinkerFatalErrorException` from https://github.com/mono/linker/pull/1155.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+		{
+			while (iterator.MoveNext ()) {
+				AssemblyDefinition assembly = GetAssembly (context, GetAssemblyName (iterator.Current));
+
+				if (assembly == null) {
+					_context.LogMessage (MessageContainer.CreateWarningMessage ($"Could not resolve assembly {GetAssemblyName (iterator.Current).Name} specified in {_xmlDocumentLocation}", 2007));
+					continue;
+				}
+
+				if (!ShouldProcessSubstitutions (iterator.Current)) {
```
vitek-karas, line 124: This source file is about attribute annotations - not about substitutions. The helper function should be called `ShouldProcessElement` if you want something really generic.

vitek-karas: I also think this should happen before we try to resolve the assembly - it would make more sense.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+				ArrayBuilder<string> arguments = GetAttributeChilds (iterator.Current.SelectChildren ("argument", string.Empty));
+				ArrayBuilder<string> fields = GetAttributeChilds (iterator.Current.SelectChildren ("field", string.Empty));
+				ArrayBuilder<string> properties = GetAttributeChilds (iterator.Current.SelectChildren ("property", string.Empty));
+
+				attributes.Add (new Attribute (attributeName, arguments, fields, properties));
+			}
+			return attributes;
+		}
+
+		ArrayBuilder<string> GetAttributeChilds (XPathNodeIterator iterator)
```
vitek-karas, line 146: "childs" is not english :-) - it's "children".

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+				TypeDefinition type = exportedType.Resolve ();
+				if (type != null) {
+					ProcessType (type, nav);
+				}
+			}
+		}
+
+		void ProcessTypePattern (string fullname, AssemblyDefinition assembly, XPathNavigator nav)
+		{
+			Regex regex = CreateRegexFromPattern (fullname);
```
vitek-karas, line 227: There's no reason to have this inside a method - just inline the method here.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+
+			foreach (TypeDefinition type in assembly.MainModule.Types) {
+				MatchType (type, regex, nav);
+			}
+
+			if (assembly.MainModule.HasExportedTypes) {
+				foreach (var exported in assembly.MainModule.ExportedTypes) {
+					MatchExportedType (exported, assembly.MainModule, regex, nav);
```
vitek-karas, line 235: This can be inline as well.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+				}
+			}
+		}
+
+		void ProcessType (TypeDefinition type, XPathNavigator nav)
+		{
+			if (!ShouldProcessSubstitutions (nav)) {
```
vitek-karas, line 242: Why do we have to check this again - as far as I can tell we already called it on the node.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+			if (!type.HasNestedTypes)
+				return;
+
+			foreach (TypeDefinition nested in type.NestedTypes) {
+				var iterator = nav.SelectChildren ("type", string.Empty);
+				while (iterator.MoveNext ()) {
+					if (nested.Name == GetAttribute (iterator.Current, "name"))
+						ProcessTypeChildren (nested, iterator.Current);
+				}
+			}
```
vitek-karas, line 257: I would swap this around - iterate over the XML and then try to lookup the types. This way we always enumerate nested types even if the XML has none in it.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
vitek-karas: I don't see this resolved - all call sites still call this like: `ProcessAttributes (iterator.Current.SelectChildren ("attribute", string.Empty))`. Unnecessary duplication.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+
+		ArrayBuilder<Attribute> ProcessAttributes (XPathNodeIterator iterator)
+		{
+			var attributes = new ArrayBuilder<Attribute> ();
+			while (iterator.MoveNext ()) {
+				string attributeName = GetFullName (iterator.Current);
+				ArrayBuilder<string> arguments = GetAttributeChilds (iterator.Current.SelectChildren ("argument", string.Empty));
+				ArrayBuilder<string> fields = GetAttributeChilds (iterator.Current.SelectChildren ("field", string.Empty));
+				ArrayBuilder<string> properties = GetAttributeChilds (iterator.Current.SelectChildren ("property", string.Empty));
+
```
vitek-karas, line 140: This is problematic. Basically only arguments are positional (that is we can rely on their order to identify which one is which). Fields and properties are not, so we would need to know both the name and the value for each of them. I think for now it would be enough to support just arguments. If we need more, we can add it later on.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+			ProcessMethodChildren (type, method, iterator);
+		}
+
+		void ProcessMethodChildren (TypeDefinition type, MethodDefinition method, XPathNodeIterator iterator)
+		{
+			ArrayBuilder<Attribute> attributes = ProcessAttributes (iterator.Current.SelectChildren ("attribute", string.Empty));
+			ArrayBuilder<(string, ArrayBuilder<Attribute>)> parameterAnnotations = ProcessParameters (type,
+				method, iterator.Current.SelectChildren ("parameter", string.Empty));
+			ArrayBuilder<ArrayBuilder<Attribute>> returnParameterAnnotations = ProcessReturnParameters (type,
+				method, iterator.Current.SelectChildren ("return", string.Empty));
```
vitek-karas, line 386: Nit: In theory it should be an error to have more than one "return" parameter. But I guess it's OK to allow it.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+
+		static DynamicallyAccessedMemberTypes GetMemberTypesForDynamicallyAccessedMemberAttribute (ArrayBuilder<Attribute> attributes, LinkContext _context, string _xmlDocumentLocation)
+		{
+			foreach (var attribute in attributes.ToArray ()) {
+				if (attribute.attributeName == "System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers" && attribute.arguments.Count == 1) {
+					foreach (var argument in attribute.arguments.ToArray ()) {
+						if (argument == string.Empty)
+							break;
+						try {
+							return (DynamicallyAccessedMemberTypes) Enum.Parse (typeof (DynamicallyAccessedMemberTypes), argument);
```
vitek-karas, line 90: This means that if there's more than one attribute of the same name, we will use the first one and ignore the rest. It would be really nice to warn (or at least inform) about such case.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+					if (paramAnnotation != 0)
+						parameterAnnotation.Add ((parameter.Item1, paramAnnotation));
+				}
+			}
+
+			if (returnParameterAnnotations.Count == 1) {
+				foreach (var returnparameter in returnParameterAnnotations.ToArray ()) {
+					DynamicallyAccessedMemberTypes returnparamAnnotation = GetMemberTypesForDynamicallyAccessedMemberAttribute (returnparameter, _context, _xmlDocumentLocation);
+					if (returnparamAnnotation != 0)
+						returnAnnotation = returnparamAnnotation;
```
vitek-karas, line 403: This on the other hand picks the last occurrence - maybe we should stick to the "first one wins" here as well.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+		}
+
+		public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition method, int index)
+		{
+			if (_methods.TryGetValue (method, out var ann) && ann.ParameterAnnotations != null) {
+				string paramName = method.Parameters[index].Name;
+
+				foreach (var (ParamName, Annotation) in ann.ParameterAnnotations)
+					if (ParamName == paramName)
+						return Annotation;
```
vitek-karas, line 54: Similar to the comment below - this will pick the first one and ignore the rest (if there are duplicates). Would be nice to warn about duplicates.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: I'm not allowing to have more than one return. `returnParameterAnnotations` parses all the return parameters but I only try to parse if the return parameters count is 1. 
Not sure if I need to fail on these kinds of cases. If I fail people will need to fix it to continue, if I warn the return parameter will not be taken into account and we will continue, but I think that's the behavior of most of the members in the XML.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: The PR is still not merged and therefore LinkerFatalErrorException does not exist yet, but it will use the error 1004 just like the ResolveXMLStep

docs/data-formats.md
```diff
@@ -224,22 +224,107 @@ are applied.
...
+### Custom attribute on type field
+
+```xml
+<linker>
+  <assembly fullname="Assembly">
+    <type fullname="Assembly.A">
+      <field name="MyNumericField">
+        <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
+          <argument>DefaultConstructor</argument>
+        </attribute>
```
marek-safar, line 236: Please add an example with more than 1 argument. We should also document how are the values converted (e.g. how to do pass enum value, a character, null value, etc.)

docs/data-formats.md
```diff
@@ -224,22 +224,107 @@ are applied.
...
+      </property>
+    </type>
+  </assembly>
+</linker>
+```
+
+### Custom attribute on method
```
marek-safar, line 259:
```suggestion
### Custom attribute on method parameter
```

docs/data-formats.md
```diff
@@ -224,22 +224,107 @@ are applied.
...
+          <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
+            <argument>DefaultConstructor</argument>
+          </attribute>
+        </property>
+      </type>
+    </type>
+  </assembly>
+</linker>
 ```
+
```
marek-safar, line 308: Could you please add an example how to add a custom attribute to
- assembly
- event
- interface

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+			{
+			}
+		}
+
+		public XmlFlowAnnotationSource (LinkContext context, string document)
+		{
+			_xmlDocumentLocation = document;
+			_document = new XPathDocument (_xmlDocumentLocation);
+			_context = context;
+			Initialize ();
```
marek-safar, line 39: This is not a great pattern, the constructor should not do such expensive operation

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -0,0 +1,595 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Xml.XPath;
+
+using Mono.Cecil;
+
+namespace Mono.Linker.Dataflow
+{
+	class XmlFlowAnnotationSource : IFlowAnnotationSource
```
marek-safar, line 16: Would not be better to do this in a more generic way and introduce a class which convert XML attributes into a new type which implements ICustomAttribute? That way we can use this mechanism for any attribute in the linker if needed not just for `DynamicallyAccessedMembers` ?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
marek-safar: Also, the message is not correct

docs/data-formats.md
vitek-karas: I'm hesitant to try to design the solution for any attribute right now, without having a use case for it. I personally would prefer to only support the necessary minimum we need right now. As we add more attributes we can extend the format as appropriate.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
vitek-karas: This is on me - I asked @tlakollo to implement the bare minimum changes to get the `DynamicallyAccessMembers` working for now. After that we can see what would be the best option to make it more generic.

docs/data-formats.md
marek-safar: That would be a breaking change as I don't think we want to have different syntax for single argument and multiple arguments

vitek-karas: I don't see how that's the case. I was assuming that if we need more arguments we would simply accept multiple `argument` elements here - order dependent. If we need fields/properties in the future that will need to also specify names.

vitek-karas: I guess we could document what we think multiple arguments, fields and properties could look like, but I don't think we need to implement all of it right now.

marek-safar: I was thinking about something like

```xml
        <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
		  <arguments>
	          <argument>DefaultConstructor</argument>
		  <arguments>
        </attribute>
```

or 

```xml
        <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
		  <fields>
	          <field name="Name">SomeValue</field>
		  <field>
        </attribute>
```

vitek-karas: I don't see a reason to add another nesting. From XML's point of view the `arguments` doesn't really add much value. The only downside is that could write:
```XML
        <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
          <argument>DefaultConstructor</argument>
          <field name="Name">SomeValue<field>
          <argument>SecondArgumentValue</argument>
        </attribute>
```

Which would work, but it's not very nice, but that's just about writing nice code - I don't think we have to enforce it by the format itself.

marek-safar: No problem with that but I think it'd be helpful to document the intention

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: As described in the data-formats.md the feature switch can be applied to any element, so it may be the case in which the assembly doesn't have the feature switch but another element type/method/field/property does have it and it will only ignore those

vitek-karas: `ProcessTypes` call this on the type element already - this is repeating that call as far as I can tell.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -165,7 +165,11 @@ public virtual void Process (LinkContext context)
 			IFlowAnnotationSource annotationSource = new AttributeFlowAnnotationSource ();
 			if (_context.AttributeDefinitions != null && _context.AttributeDefinitions.Count > 0) {
 				annotationSource = new AggregateFlowAnnotationSource (
-					_context.AttributeDefinitions.Select (s => new JsonFlowAnnotationSource (_context, s))
+					_context.AttributeDefinitions.Select (s => {
+						var xmlAnnotations = new XmlFlowAnnotationSource (_context, s);
+						xmlAnnotations.ParseXml ();
```
vitek-karas, line 170: Nit: Cleaner design would be to parse the file name to the `ParseXml` file instead...

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: Got it! I fix it and added another one for nested types

marek-safar:
Please squash your changes next time you will be merging such PR

Issue #1147, opened 2020-04-27
Respect the new UnconditionalSuppressMessageAttribute


eerhardt:
When code wants to suppress a linker warning, for example when the author is certain it is safe, they can add an `UnconditionalSuppressMessageAttribute`. We should support this new attribute in the linker.

See https://github.com/dotnet/runtime/issues/35339 for the discussion.

vitek-karas:
There are several aspects of this which can be done in stages:
* Basic support of the attribute on a method - suppresses specified warning inside the method's body.
* Implement support for [`MessageId`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute.messageid?view=netcore-3.1#System_Diagnostics_CodeAnalysis_SuppressMessageAttribute_MessageId) - adds the ability to direct the suppression to a specific symbol in the method's body. For example if the linker raises a warning due to a call to `Type.GetField` then it should be possible to direct the suppression to just that by specifying `MessageId` with a value of `GetField` (or maybe `System.Type.GetField`, or so on - needs a detailed design)
* Support the attribute on more scopes - type, module, assembly and so on. This will be performance sensitive work
* Support targeting the attribute - that is support for the [`Scope`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute.scope?view=netcore-3.1#System_Diagnostics_CodeAnalysis_SuppressMessageAttribute_Scope) and [`Target`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute.target?view=netcore-3.1) properties. Note that this requires a parser of the string in the `Target` property. We need this to be aligned with the same functionality in roslyn, both the parser and the semantics. The parser for the `Target` property can be found [here](https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/DiagnosticAnalyzer/SuppressMessageAttributeState.TargetSymbolResolver.cs).

Performance considerations: When a warning is suppressed it should be fast - linker should not spend my time finding and suppressing the warning. This is expected to be a relatively common case (likely used inside our frameworks so all apps are affected). This means that finding and matching the suppression attributes must be fast and when a warning is suppressed linker should not spend any time computing the details of the warning (specifically constructing the warning message).

MichalStrehovsky:
>  the Target property processing is similar problem to the target of the DynamicDependnecyAttribute in #1148

[...]

PR #1185, opened 2020-05-13
Add UnconditionalSuppressMessageAttribute

mateoatr:
Related to #1147. This PR contains the first effort to support the `UnconditionalSuppressMessageAttribute`. It adds the ability to suppress a specific warning in members and types. Tests were added to show its usage.


src/linker/Linker/MessageOrigin.cs
```diff
@@ -39,6 +51,15 @@ public MessageOrigin (string fileName, int sourceLine = 0, int sourceColumn = 0)
 			return null;
 		}
 
+		internal static MessageOrigin? TryGetOrigin (IMemberDefinition sourceMethod, int ilOffset,
```
marek-safar, line 54: I don't think this is the right place. We want the developer to be able to suppress any warning message (e.g. from 3rd party custom step). I think the whole logic should be close to MessageContainer.CreateWarning which probably will need to be expanded with additional "context" state parameter

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1426,6 +1428,8 @@ bool MarkSpecialCustomAttributeDependencies (CustomAttribute ca, ICustomAttribut
 					l.Parameters.Count == 1 && l.Parameters[0].ParameterType.IsTypeOf ("System", "Type"),
 					provider);
 				return true;
+			} else if (dt.Name == "UnconditionalSuppressMessageAttribute" && dt.Namespace == "System.Diagnostics.CodeAnalysis") {
+				_context.Suppressions.AddLocalSuppression (ca, provider.MetadataToken);
```
marek-safar, line 1432: Does it really have to be MetadataToken when the attribute cannot be attached to tokens?

src/linker/Linker/MessageContainer.cs
```diff
@@ -63,6 +63,10 @@ public static MessageContainer CreateWarningMessage (string text, int code, stri
 			if (!(code > 2000 && code <= 6000))
 				throw new ArgumentException ($"The provided code '{code}' does not fall into the warning category, which is in the range of 2001 to 6000 (inclusive).");
 
+			if (origin != null && origin.Value.IsSuppressed) {
+				return CreateInfoMessage ($"Warning '{origin.Value.SuppressionInfo.Id}' was suppressed");
```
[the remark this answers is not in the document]
vitek-karas: If it's suppressed it should behave as if it wasn't there at all. So no messages.

src/linker/Linker/MessageOrigin.cs
vitek-karas: Absolutely agree.
The suppression should work without any special code from the place where we create the warning - with the sole exception that we require the "context" parameter. I would even go as far as make some strict rule that the context must be specified unless the code is somehow special (like internal errors, command line parsing errors and so on) - otherwise we should always be able to tell what the warning applies to (assembly or more specific).

I think from the API perspective this change should add the "context" attribute to the `CreateWarning` method (and maybe also `CreateError` for consistency). Then we should have another API which is something like `IsSuppressed` - I would probably put it onto the `MessageContainer` if it's possible, or if that doesn't work then the `LinkContext`. It should take the necessary parameters to decide if a warning will be suppressed or not. Notable it should not require the full warning message - the indented usage would be to avoid computing potentially expensive warning message if we know it's going to be suppressed.

mateoatr: I've moved this logic out of `MessageOrigin` by adding the mentioned `IsSuppressed` API into the `LinkContext`, which is called from within `CreateWarningMessage`. It's now mandatory to specify a `MessageOrigin` for all created warnings, which must be either a filename or an objetct of a class implementing `IMetadataTokenProvider` (modules, types and members). The way to create warnings feels a little bit cumbersome though (`Context.LogMessage (MessageContainer.CreateWarningMessage (Context, Message, Code, MessageOrigin, [optional] Subcategory))`), I think this could be enhanced by adding new methods (`LogMessage` for info/diags, `LogError` for errors, `LogWarning` for warnings) and stop passing around the `LinkContext`.

src/linker/Linker.Steps/MarkStep.cs
mateoatr: All local suppressions can be attached to a a specific md token. For suppressions which cannot be attached to tokens we'll probably be storing a string instead, but I think this case only applies for suppressions that use the target/scope argument. Support for this will be incorporated in a future PR, since the signature parser will be needed for some scenarios (global suppressions made by specifying target/scope).

marek-safar: ok but to do the decoding here is too late as the suppression on metadata tokens would work for steps starting from MarkStep only

src/linker/Linker/LinkContext.cs
```diff
@@ -463,10 +465,20 @@ public bool IsOptimizationEnabled (CodeOptimizations optimization, AssemblyDefin
 			return Optimizations.IsEnabled (optimization, context);
 		}
 
-		public void LogMessage (MessageContainer message)
+		public void LogMessage (MessageContainer? message)
```
marek-safar, line 468: That's not a great API. You could experiment with static MessageContainer.Empty sentinel.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,244 @@
...
+		{
+			if (_lazyGlobalSuppressions == null) {
+				var suppressions = new GlobalSuppressions ();
+				DecodeGlobalSuppressMessageAttributes (_assembly, suppressions);
+
+				foreach (var module in _assembly.Modules) {
+					DecodeGlobalSuppressMessageAttributes (module, suppressions);
+				}
+
+				Interlocked.CompareExchange (ref _lazyGlobalSuppressions, suppressions, null);
```
marek-safar, line 190: none of the linker code is thread safe

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,244 @@
...
+			var attributes = caProvider.CustomAttributes.Where (a => nameof (a.AttributeType) == "UnconditionalSuppressMessageAttribute");
+			DecodeGlobalSuppressMessageAttributes (globalSuppressions, attributes);
+		}
+
+		private void DecodeGlobalSuppressMessageAttributes (GlobalSuppressions globalSuppressions, IEnumerable<CustomAttribute> customAttributes)
+		{
+			foreach (var customAttribute in customAttributes) {
+				SuppressMessageInfo info;
+				if (!TryDecodeSuppressMessageAttributeData (customAttribute, out info)) {
+					return;
```
marek-safar, line 209: Returns here don't make much sense to me

src/linker/ref/Linker/LinkContext.cs
```diff
@@ -15,9 +15,10 @@ public class LinkContext
 		public TypeDefinition GetType (string fullName) { throw null; }
 		public AssemblyDefinition GetLoadedAssembly (string name) { throw null; }
 
-		public void LogMessage (MessageContainer message) { throw null; }
+		public void LogMessage (MessageContainer? message) { throw null; }
 
 		public bool HasCustomData (string key) { throw null; }
 		public bool TryGetCustomData (string key, out string value) { throw null; }
+		public bool IsSuppressed (int warningCode, MessageOrigin origin) { throw null; }
```
marek-safar, line 22: We don't want to expose this method

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,244 @@
...
+					}
+
+					if (mdTokenProvider is FieldDefinition ||
+						mdTokenProvider is PropertyDefinition ||
+						mdTokenProvider is EventDefinition ||
+						mdTokenProvider is MethodDefinition) {
+						mdTokenProvider = (mdTokenProvider as IMemberDefinition).DeclaringType;
+					} else if (mdTokenProvider is TypeDefinition type) {
+						if (type.IsNested) {
+							_ = type.BaseType.Resolve ();
```
marek-safar, line 82: What is this about?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,244 @@
...
+						}
+
+						mdTokenProvider = type.Module;
+					} else if (mdTokenProvider is ModuleDefinition module) {
+						mdTokenProvider = null;
+					}
+				} while (mdTokenProvider != null);
```
marek-safar, line 89: Does this loop make sense when we care only about members and declaration?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,244 @@
...
+					switch (p.Name) {
+					case "Scope":
+						info.Scope = p.Argument.Value as string;
+						break;
+					case "Target":
+						info.Target = p.Argument.Value as string;
+						break;
+					case "MessageId":
+						info.MessageId = p.Argument.Value as string;
+						break;
```
marek-safar, line 130: I think there should be warning if any of the values are in incorrect format

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,244 @@
...
+using System.Linq;
+using System.Threading;
+using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	public class UnconditionalSuppressMessageAttributeState
+	{
+		private readonly AssemblyDefinition _assembly;
+		private GlobalSuppressions _lazyGlobalSuppressions;
```
marek-safar, line 13: Why have a class with single field?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: I've rewritten this, we now simply walk up members up to the declaring type to check for suppressions.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: I agree, although because IL doesn't keep information about the source line where a custom attribute is, I don't see a good way to report warnings on this scenario where the CA can be put anywhere.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: We're just making a clear distinction between the global suppressions and the local ones with this class, just as it's done in Roslyn. We could of course have it all in one class but I don't see a reason to not have this.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
vitek-karas: I checked how Roslyn reacts and it basically ignores invalid values. I think we should probably do the same. Maybe check that the types match, but otherwise don't validate the actual string values.

src/linker/Linker/MessageOrigin.cs
```diff
@@ -20,29 +23,48 @@ public MessageOrigin (string fileName, int sourceLine = 0, int sourceColumn = 0)
...
+
+		private MessageOrigin (string fileName, IMetadataTokenProvider mdTokenProvider, int sourceLine = 0, int sourceColumn = 0)
+		{
+			FileName = fileName;
+			MdTokenProvider = mdTokenProvider;
+			SourceLine = sourceLine;
+			SourceColumn = sourceColumn;
+		}
+
+		public static MessageOrigin TryGetOrigin (IMetadataTokenProvider sourceMethod, int ilOffset)
```
vitek-karas, line 45: This is weird - if I have an IL offset then I must have a method, and if I don't have a method I can't have IL offset. So I think this should take the first parameter as `MethodDefinition`.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,259 @@
...
+using System.Linq;
+using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	public class UnconditionalSuppressMessageAttributeState
+	{
+		private readonly AssemblyDefinition _assembly;
+		private GlobalSuppressions _lazyGlobalSuppressions;
+		private readonly Dictionary<MetadataToken, ImmutableDictionary<string, SuppressMessageInfo>> _localSuppressionsByMdToken;
```
vitek-karas, line 12: Is there a reason we store these based on actual tokens? Why not use the Cecil OM - `IMetadataTokenProvider`? (Maybe the GetHashCode is too expensive - which it might be - in which case please add a comment describing why we do it this way)

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,259 @@
...
+				scope = default;
+				return false;
+			}
+
+			return true;
+		}
+
+		private bool HasLocalSuppressions {
+			get {
+				return _localSuppressionsByMdToken.Keys.Count != 0;
```
vitek-karas, line 48:
```suggestion
				return _localSuppressionsByMdToken.Count != 0;
```

This should be faster - no need to force allocate the Keys collection

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,259 @@
...
+			SuppressMessageInfo info;
+			if (!TryDecodeSuppressMessageAttributeData (ca, out info)) {
+				return;
+			}
+
+			if (_localSuppressionsByMdToken.ContainsKey (mdToken)) {
+				_localSuppressionsByMdToken[mdToken].Add (info.Id, info);
+			} else {
+				_localSuppressionsByMdToken[mdToken] = new Dictionary<string, SuppressMessageInfo> { { info.Id, info } }.ToImmutableDictionary ();
+			}
```
vitek-karas, line 81:
```suggestion
			if (!_localSuppressionsByMdToken.TryGetValue (mdToken, out var suppressions) {
				suppressions = new Dictionary<string, SuppressMessageInfo> { { info.Id, info } }.ToImmutableDictionary ();
				_localSuppressionsByMdToken.Add (mdToken, suppressions);
			}
			
			suppressions.Add (info.Id, info);
```

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,259 @@
...
+			var separatorIndex = info.Id.IndexOf (':');
+			if (separatorIndex != -1) {
+				info.Id = info.Id.Remove (separatorIndex);
+			}
+
+			if (attribute.HasProperties) {
+				foreach (var p in attribute.Properties) {
+					switch (p.Name) {
+					case "Scope":
+						info.Scope = p.Argument.Value as string;
```
vitek-karas, line 133: In the comparison above we seem to expect this to be lowercased - I think?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,259 @@
...
+			}
+
+			return true;
+		}
+
+		private class GlobalSuppressions
+		{
+			private readonly Dictionary<string, SuppressMessageInfo> _compilationWideSuppressions = new Dictionary<string, SuppressMessageInfo> ();
+
+			private readonly Dictionary<MetadataToken, Dictionary<string, SuppressMessageInfo>> _globalSuppressions =
```
vitek-karas, line 152: In the local suppressions you have ImmutableDictionary, here you have normal dictionary - any reason for the difference?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
marek-safar: I think it'd be more readable to have `private readonly Dictionary<MetadataToken, Dictionary<string, SuppressMessageInfo>> _globalSuppressions = ...` here and hidden inside nested class

test/Mono.Linker.Tests.Cases/Logging/Dependencies/LogStep.cs
```diff
@@ -9,7 +9,7 @@ public class LogStep : IStep
 		public void Process (LinkContext context)
 		{
 			var msgError = MessageContainer.CreateErrorMessage ("Error", 1004);
-			var msgWarning = MessageContainer.CreateWarningMessage ("Warning", 2001, origin: new MessageOrigin ("logtest", 1, 1));
+			var msgWarning = (MessageContainer) MessageContainer.CreateWarningMessage (context, "Warning", 2001, origin: new MessageOrigin ("logtest", 1, 1));
```
marek-safar, line 12: why is this change needed?

test/Mono.Linker.Tests/Tests/MessageContainerTests.cs
```diff
@@ -8,10 +8,12 @@ public class MessageContainerTests
 		[Test]
 		public void MSBuildFormat ()
 		{
+			LinkContext context = new LinkContext (new Pipeline ());
+
 			var msg = MessageContainer.CreateErrorMessage ("text", 1000);
 			Assert.AreEqual ("ILlinker: error IL1000: text", msg.ToMSBuildString ());
 
-			msg = MessageContainer.CreateWarningMessage ("message", 2001, origin: new MessageOrigin ("logtest", 1, 1));
+			msg = (MessageContainer) MessageContainer.CreateWarningMessage (context, "message", 2001, new MessageOrigin ("logtest", 1, 1));
```
marek-safar, line 16: why is this change needed?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: It made sense to me to use to most specific thing we care about when comparing members/types. I've changed this to use the md provider instead; I didn't see any time burden we should worry about.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: I think that in Roslyn the `ImmutableDictionary` was used because of thread safety reasons, I've changed it to a `Dictionary`.

mateoatr:
I've removed all things related to global suppressions from this PR. I'll work on them and put them on a following PR to keep this one short and don't delay much in merging the basic support for the `UnconditionalSuppressMessage` attribute. The code introduced here gives basic capabilities for suppressing warnings in local scopes (members and types).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1012,6 +1015,8 @@ protected void MarkAssembly (AssemblyDefinition assembly)
 
 			foreach (ModuleDefinition module in assembly.Modules)
 				LazyMarkCustomAttributes (module, module);
+
+			_context.Suppressions = new UnconditionalSuppressMessageAttributeState (assembly);
```
vitek-karas, line 1019: This is probably not right. The `_context` is global - for all assemblies, and here we're assigning a state which is for one assembly only. Also - what if there already is another state in that variable? MarkStep doesn't work one assembly at a time - it jumps randomly between assemblies.

We either need to store the suppressions per-assembly or have some kind of global storage keyed of the metadata token providers.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,115 @@
...
+			if (!TryDecodeSuppressMessageAttributeData (ca, out info)) {
+				return;
+			}
+
+			if (!_localSuppressionsByMdToken.TryGetValue (mdTokenProvider, out var suppressions)) {
+				suppressions = new Dictionary<string, SuppressMessageInfo> ();
+				_localSuppressionsByMdToken.Add (mdTokenProvider, suppressions);
+			}
+
+			suppressions.Add (info.Id, info);
```
vitek-karas, line 35: Is there a guarantee that there will be only one "info" for each ID? Otherwise this is going to throw at runtime. I think this should be either a Dictionary<string, List<>> or just a flat list which we search through each time.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,115 @@
...
+			}
+
+			return true;
+		}
+
+		private static void AddOrUpdate (SuppressMessageInfo info, IDictionary<string, SuppressMessageInfo> builder)
+		{
+			// TODO: How should we deal with multiple SuppressMessage attributes, with different suppression info/states?
+			// For now, we just pick the last attribute, if not suppressed.
+			if (!builder.TryGetValue (info.Id, out _)) {
```
vitek-karas, line 102:
```suggestion
			if (!builder.ContainsKey (info.Id)) {
```

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,115 @@
...
+						return true;
+
+					mdTokenProvider = (mdTokenProvider as IMemberDefinition).DeclaringType;
+				}
+			}
+
+			return false;
+		}
+
+		private bool TryDecodeSuppressMessageAttributeData (CustomAttribute attribute, out SuppressMessageInfo info)
```
vitek-karas, line 55: static

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,115 @@
...
+						info.MessageId = p.Argument.Value as string;
+						break;
+					}
+				}
+			}
+
+			return true;
+		}
+
+		private static void AddOrUpdate (SuppressMessageInfo info, IDictionary<string, SuppressMessageInfo> builder)
```
vitek-karas, line 98: Is this used by anything?

test/Mono.Linker.Tests.Cases/WarningSuppression/WarningMessageSuppression.cs
```diff
@@ -0,0 +1,92 @@
...
+		{
+			Expression.Call (WarningMessageSuppression.TriggerUnrecognizedPattern (), "", Type.EmptyTypes);
+
+			void LocalFunction ()
+			{
+				Expression.Call (WarningMessageSuppression.TriggerUnrecognizedPattern (), "", Type.EmptyTypes);
+			}
+		}
+
+		[UnconditionalSuppressMessage ("Test", "IL2006:UnrecognizedReflectionPattern")]
```
vitek-karas, line 81: Does this really work? The implementation walks from method to the declaring type - I don't see anything which would look at the property for attributes....

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: There's not. I've changed this to `TryAdd`, since I see no scenario in which we should store two "info" for the same (md token provider, ID) tuple.

mateoatr: Changed it once again since `TryAdd` is only supported in .NET Core. The behavior now is to keep the last `info` seen for these (provider, id) tuples.

src/linker/Linker/LinkContext.cs
```diff
@@ -465,9 +468,17 @@ public bool IsOptimizationEnabled (CodeOptimizations optimization, AssemblyDefin
...
+			if (LogMessages && message != MessageContainer.Empty)
 				Logger?.LogMessage (message);
 		}
+
+		public bool IsWarningSuppressed (int warningCode, MessageOrigin origin)
+		{
+			if (Suppressions == null)
+				return false;
+
+			return Suppressions.IsSuppressed ("IL" + warningCode, origin, out _);
```
vitek-karas, line 480: Can we avoid the "IL" + warningCode. This means allocation every time we call this method. Seems unnecessary. If we only support warnings of shape IL####, then we can probably parse the suppression attributes directly to the warning code.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,104 @@
...
+			if (!TryDecodeSuppressMessageAttributeData (ca, out info)) {
+				return;
+			}
+
+			if (!_localSuppressionsByMdToken.TryGetValue (mdTokenProvider, out var suppressions)) {
+				suppressions = new Dictionary<string, SuppressMessageInfo> ();
+				_localSuppressionsByMdToken.Add (mdTokenProvider, suppressions);
+			}
+
+			suppressions[info.Id] = info;
```
vitek-karas, line 33: Please add a comment that we intentionally use the last one (this is potentially another candidate for a warning message - or maybe just informational - generating warning message from within suppression might cause trouble

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -26,14 +29,18 @@ public void AddLocalSuppression (CustomAttribute ca, IMetadataTokenProvider mdTo
...
 
 			if (!_localSuppressionsByMdToken.TryGetValue (mdTokenProvider, out var suppressions)) {
-				suppressions = new Dictionary<string, SuppressMessageInfo> ();
+				suppressions = new Dictionary<int, SuppressMessageInfo> ();
 				_localSuppressionsByMdToken.Add (mdTokenProvider, suppressions);
 			}
 
+			if (suppressions.ContainsKey (info.Id))
+				_context.LogMessage (MessageContainer.CreateInfoMessage (
+					$"Type or member {mdTokenProvider} has more than one unconditional suppression. Note that only the last one is kept."));
```
vitek-karas, line 38:
```suggestion
					$"Type or member {mdTokenProvider} has more than one unconditional suppression. Note that only the last one is used."));
```

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -62,18 +69,13 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
 
 			// Ignore the category parameter because it does not identify the warning
 			// and category information can be obtained from warnings themselves.
-			info.Id = attribute.ConstructorArguments[1].Value as string;
-			if (info.Id == null) {
+			// We only support warnings with code pattern IL####.
+			if (!(attribute.ConstructorArguments[1].Value is string warningId) ||
+				!Regex.IsMatch (warningId, "^IL\\d{4}")) {
```
vitek-karas, line 74: Let's do this by hand - regex is too heavy weight just for this simple parse.

vitek-karas: You ca do .StartsWith("IL") and then substring and int.TryParse.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -71,11 +70,12 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
 			// and category information can be obtained from warnings themselves.
 			// We only support warnings with code pattern IL####.
 			if (!(attribute.ConstructorArguments[1].Value is string warningId) ||
-				!Regex.IsMatch (warningId, "^IL\\d{4}")) {
+				warningId.Length < 6 ||
+				!warningId.StartsWith ("IL") ||
+				!int.TryParse (warningId.Substring (2, 4), out info.Id)) {
```
vitek-karas, line 75: This means that "IL1234HiThere" will be read as 1234. Probably not what we want. Just check that Length != 6.

mateoatr: `SuppressMessageAttributes` allows the user to write anything after a colon, mainly for documentation purposes (e.g., `"IL1234:Suppress this"`). I think we should keep this capability.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,109 @@
...
+				return false;
+			}
+
+			if (warningId.Length > 6 && warningId[6] != ':')
+				return false;
+
+			if (attribute.HasProperties) {
```
marek-safar, line 82: Where are the tests for this logic?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -0,0 +1,109 @@
...
+			if (suppressions.ContainsKey (info.Id))
+				_context.LogMessage (MessageContainer.CreateInfoMessage (
+					$"Type or member {mdTokenProvider} has more than one unconditional suppression. Note that only the last one is used."));
+
+			suppressions[info.Id] = info;
+		}
+
+		public bool IsSuppressed (int id, MessageOrigin warningOrigin, out SuppressMessageInfo info)
+		{
+			info = default;
```
marek-safar, line 44: Move this before `return false;`

src/linker/Linker/MessageOrigin.cs
```diff
@@ -20,29 +23,45 @@ public MessageOrigin (string fileName, int sourceLine = 0, int sourceColumn = 0)
...
+		private MessageOrigin (string fileName, IMetadataTokenProvider mdTokenProvider, int sourceLine = 0, int sourceColumn = 0)
+		{
+			FileName = fileName;
+			MdTokenProvider = mdTokenProvider;
+			SourceLine = sourceLine;
+			SourceColumn = sourceColumn;
+		}
 
-				SequencePoint correspondingSequencePoint = methodDef.DebugInformation.SequencePoints
+		public static MessageOrigin TryGetOrigin (MethodDefinition sourceMethod, int ilOffset)
```
marek-safar, line 45: This is quite expensive and you call it before warning is rejected. Could this be rewritten to be done when the FileName is actually needed?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: For now I've only added an attribute with these properties set to `null` in the tests. Most interesting tests can be implemented after we integrate global suppressions (which will need to use the signature parser for resolving the `Target` and `MessageId`).

src/linker/Linker/MessageOrigin.cs
mateoatr: Yes, we can do this and we should. I've opened an issue for this: https://github.com/mono/linker/issues/1221

PR #1189, opened 2020-05-14
Fix a couple nits

MichalStrehovsky:
* It's not mandatory to name the single instance field on enums `value__`. C#/VB do that, but the spec only requires there is an instance field. The name check is unnecessary work.
* The `IsFinalizer` extension method was wrong: one one side it includes unnecessary methods (for people who like to fight CS0465 warning from the C# compiler and name their methods `Finalize` - and yes there are those people), and on the other side it doesn't include finalizers that are not named `Finalize` (one can just `MethodImpl` the `Finalize` method). Instead of trying to get the extension method right, just letting the general virtual method handling take care of it.

### DIFF
--- a/src/linker/Linker/MethodDefinitionExtensions.cs
+++ b/src/linker/Linker/MethodDefinitionExtensions.cs
@@ -74,15 +74,15 @@ namespace Mono.Linker
 #if !FEATURE_ILLINK
 		// This implementation is wrong. It should return true if this is a virtual override
 		// of System.Object::Finalize, but that's not what this is doing. Do not use.
-		public static bool IsFinalizer (this MethodDefinition method)	
-		{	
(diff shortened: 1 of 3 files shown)


src/linker/Linker/MethodDefinitionExtensions.cs
```diff
@@ -71,6 +71,9 @@ public static bool IsStaticConstructor (this MethodDefinition method)
 			return method.IsConstructor && method.IsStatic;
 		}
 
+#if !FEATURE_ILLINK
```
mrvoorhe, line 74: What's the rational for keeping it around if it's wrong?

MichalStrehovsky: Monolinker build break.

PR #1195, opened 2020-05-15
Implement a couple more dataflow intrinsics

MichalStrehovsky:
* `Type.GetNestedType` - should be straighforward
* `TypeDelegator..ctor` - this is used in CoreLib - we should just treat it as identity so that we don't lose dataflow annotations
* `object.GetType` - I specifically need the delegate part for `CoreLib`; the rest of the behavior is nice to have but I don't expect it to kick in too often.

Creating as draft because it obviously misses tests, but it's Friday...

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -157,6 +157,8 @@ namespace Mono.Linker.Dataflow
 			IntrospectionExtensions_GetTypeInfo,
 			Type_GetTypeFromHandle,
 			Type_get_TypeHandle,
+			Object_GetType,
+			TypeDelegator_Ctor,
 
 			// Anything above this marker will require the method to be run through
 			// the reflection body scanner.
@@ -168,6 +170,7 @@ namespace Mono.Linker.Dataflow
 			Type_GetField,
 			Type_GetProperty,
 			Type_GetEvent,
+			Type_GetNestedType,
 			Expression_Call,
 			Expression_Field,
 			Expression_Property,
@@ -298,6 +301,13 @@ namespace Mono.Linker.Dataflow
 					&& calledMethod.HasThis
 					=> IntrinsicId.Type_GetEvent,
 
+				// System.Type.GetNestedType (string)
+				// System.Type.GetNestedType (string, BindingFlags)
+				"GetNestedType" when calledMethod.IsDeclaredOnType ("System", "Type")
+					&& calledMethod.HasParameterOfType (0, "System", "String")
+					&& calledMethod.HasThis
+					=> IntrinsicId.Type_GetNestedType,
+
 				// System.Type.GetProperty (string)
 				// System.Type.GetProperty (string, BindingFlags)
 				// System.Type.GetProperty (string, Type)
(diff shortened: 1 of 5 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1483,6 +1606,10 @@ string GetDynamicallyAccessedMemberKindsDescription (DynamicallyAccessedMemberTy
 			return string.Join (" | ", results.Select (r => r.ToString ()));
 		}
 
+		static DynamicallyAccessedMemberTypes GetDynamicallyAccessedMemberTypesFromBindingFlagsForTypes (BindingFlags bindingFlags) =>
```
vitek-karas, line 1609:
```suggestion
		static DynamicallyAccessedMemberTypes GetDynamicallyAccessedMemberTypesFromBindingFlagsForNestedTypes (BindingFlags bindingFlags) =>
```

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1083,7 +1190,7 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
 			foreach (var uniqueValue in value.UniqueValues ()) {
 				if (uniqueValue is LeafValueWithDynamicallyAccessedMemberNode valueWithDynamicallyAccessedMember) {
 					if (!valueWithDynamicallyAccessedMember.DynamicallyAccessedMemberKinds.HasFlag (requiredMemberKinds)) {
-						reflectionContext.RecordUnrecognizedPattern ($"The {GetValueDescriptionForErrorMessage (valueWithDynamicallyAccessedMember)} " +
+						reflectionContext.RecordUnrecognizedPattern ($"{reflectionContext.SourceMethod} The {GetValueDescriptionForErrorMessage (valueWithDynamicallyAccessedMember)} " +
```
vitek-karas, line 1193: This should be part of a different change - there are more places where we need to add something like this.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1266,15 +1373,31 @@ void MarkFieldsOnTypeHierarchy (ref ReflectionPatternContext reflectionContext,
 			}
 		}
 
-		void MarkNestedTypesOnType (ref ReflectionPatternContext reflectionContext, TypeDefinition type, Func<TypeDefinition, bool> filter)
+		TypeDefinition[] MarkNestedTypesOnType (ref ReflectionPatternContext reflectionContext, TypeDefinition type, Func<TypeDefinition, bool> filter, BindingFlags bindingFlags = BindingFlags.Default)
```
vitek-karas, line 1376: Please change the callsite in here https://github.com/mono/linker/pull/1195/files#diff-d634a3c96357eb6bf5b24034efb1e61aR1253-R1257 to use the BIndingFlags just like all the other kinds... (no need for the filter in those cases).

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
MichalStrehovsky: Oops, that was not intentional.

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -825,7 +825,7 @@ static string GetFullMemberNameFromDefinition (IMetadataTokenProvider member)
 				member = memberReference.Resolve ();
 
 			if (member is IMemberDefinition memberDefinition) {
-				if (memberDefinition is TypeDefinition) {
+				if (memberDefinition is TypeDefinition type) {
```
vitek-karas, line 828: This seems to be unnecessary change...

PR #1197, opened 2020-05-16
Add signature parser based on Roslyn doc comments

sbomer:
This adds a parser for the format described at https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format, which we plan to use for UnconditionalSuppressMessageAttribute (https://github.com/mono/linker/issues/1147) and probably also DynamicDependencyAttribute.

It's adapted from Roslyn, with some changes to allow for resolving from metadata instead of sources. This doesn't support the legacy fxcop format, and it also doesn't allow some special cases that Roslyn has - see comments in the sources. The Roslyn implementation uses a visitor pattern, where symbols in their syntax tree implement accept methods. This seemed hard to adapt to our use case, and I don't think we need the full generality of this pattern, so this parser follows the same structure without truly using the visitor pattern.


src/linker/Linker/SignatureGenerator.PartVisitor.cs
```diff
@@ -0,0 +1,207 @@
...
+					builder.Append ('.');
+				}
+
+				if (!String.IsNullOrEmpty (typeReference.Namespace))
+					builder.Append (typeReference.Namespace).Append ('.');
+
+				// This includes '`n' for mangled generic types
+				builder.Append (typeReference.Name);
+
+				if (typeReference.HasGenericParameters || !typeReference.IsGenericInstance)
```
marek-safar, line 156: Is this condition correct?

src/linker/Linker/SignatureParser.cs
```diff
@@ -0,0 +1,661 @@
...
+		private static string? ParseTypeSymbol (string id, ref int index, IGenericParameterProvider? typeParameterContext)
+		{
+			var results = new List<string> ();
+			ParseTypeSymbol (id, ref index, typeParameterContext, results);
+			if (results.Count == 0) {
+				return null;
+			} else {
+				Debug.Assert (results.Count == 1);
+				return results[0];
+			}
```
marek-safar, line 219: I'd invert the logic if you are expecting `results.Count == 1` to be the common path

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -187,6 +224,13 @@ public static string ToCecilName (this string fullTypeName)
 			return fullTypeName.Replace ('+', '/');
 		}
 
+		public static string GetSignaturePart (this TypeReference type)
+		{
+			var builder = new StringBuilder ();
+			SignatureGenerator.PartVisitor.Instance.VisitTypeReference (type, builder);
+			return builder.ToString ();
+		}
```
marek-safar, line 232: I'd keep this method inside SignatureParser

src/linker/Linker/SignatureParser.cs
```diff
@@ -0,0 +1,661 @@
...
+	///  https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/DocumentationCommentId.cs
+	/// </summary>
+	///
+	/// Roslyn's API works with ISymbol, which represents a symbol exposed by the compiler.
+	/// a Symbol has information about the source language, name, metadata name,
+	/// containing scopes, visibility/accessibility, attributes, etc.
+	/// This API instead works with the Cecil OM. It can be used to refer to IL definitions
+	/// where the signature of a member can contain references to instantiated generics.
+	///
+	public static class SignatureParser
```
marek-safar, line 31: It'd be better to use a less ambiguous name (there are a lot of signatures inside linker)

src/linker/Linker/SignatureGenerator.PartVisitor.cs
sbomer: I think so - unless I missed a case in cecil. Non-generics don't need more work, and uninstantiated generics are already mangled. I added a clarifying comment.

src/linker/Linker/SignatureParser.cs
sbomer: Changed to DocumentationSignatureParser

test/Mono.Linker.Tests.Cases.Expectations/Assertions/ExpectGeneratedStringAttribute.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Assertions
+{
+	public class ExpectGeneratedStringAttribute : BaseMemberAssertionAttribute
```
vitek-karas, line 5: I think these need comments describing what they do (and in which tests they actually work).
And I also think the names should mention the "Documentation syntax" in some way - when I saw this I could not figure out what the attribute is for until I saw it being used.

src/linker/Linker/DocumentationSignatureGenerator.PartVisitor.cs
```diff
@@ -0,0 +1,209 @@
+using System;
```
marek-safar, line 1: Add licence header please

src/linker/Linker/DocumentationSignatureGenerator.cs
```diff
@@ -0,0 +1,53 @@
+using System;
```
marek-safar, line 1: Add licence header please

src/linker/Linker/DocumentationSignatureGenerator.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Text;
+using Mono.Cecil;
+
+#nullable enable
```
marek-safar, line 5: Why if there is nullable type used?

src/linker/Linker/DocumentationSignatureParser.cs
```diff
@@ -0,0 +1,666 @@
...
+
+			if (module == null)
+				throw new ArgumentNullException (nameof (module));
+
+			var results = new List<IMemberDefinition> ();
+			Parser.ParseDeclaredSymbolId (id, module, results);
+			return results;
+		}
+
+		public static string GetSignaturePart (this TypeReference type)
```
marek-safar, line 45: Remove `this` and make it private

src/linker/Linker/DocumentationSignatureParser.cs
```diff
@@ -0,0 +1,666 @@
...
+		{
+			var builder = new StringBuilder ();
+			DocumentationSignatureGenerator.PartVisitor.Instance.VisitTypeReference (type, builder);
+			return builder.ToString ();
+		}
+
+		private static class Parser
```
marek-safar, line 52: Not sure what is the purpose of this static nesting and I'd avoid it

src/linker/Linker/DocumentationSignatureParser.cs
```diff
@@ -0,0 +1,666 @@
...
+
+			if (PeekNextChar (id, index) == '`') {
+				// `` means this is a method type parameter
+				index++;
+				var methodTypeParameterIndex = ReadNextInteger (id, ref index);
+
+				if (typeParameterContext is MethodDefinition methodContext) {
+					var count = methodContext.HasGenericParameters ? methodContext.GenericParameters.Count : 0;
+					if (count > 0 && methodTypeParameterIndex < count) {
+						results.Add ($"``{methodTypeParameterIndex}");
```
marek-safar, line 303: `results.Add ("``" + methodTypeParameterIndex);`

src/linker/Linker/DocumentationSignatureParser.cs
```diff
@@ -0,0 +1,666 @@
...
+				index = id.Length;
+			}
+
+			return DecodeName (name);
+		}
+
+		// undoes dot encodings within names...
+		private static string DecodeName (string name)
+		{
+			if (name.IndexOf ('#') >= 0)
```
marek-safar, line 647: IndexOf is redundant here

test/Mono.Linker.Tests/Tests/DocumentationSignatureParserTests.cs
```diff
@@ -0,0 +1,582 @@
...
+			{
+			}
+
+			[ExpectUnresolvedDocumentationSignature ("T:Mono.Linker.Tests.DocumentationSignatureParserTests.Invalid.NoParameterType(Mono.Linker.Tests.DocumentationSignatureParserTests.Invalid.Generic{})")]
+			public void NoGenericParameterType (Generic<A> g)
+			{
+			}
+
+			// these work, but seem like they shouldn't.
+			// see https://github.com/dotnet/roslyn/issues/44315
```
marek-safar, line 564: Could you please add tracking issue for linker

src/linker/Linker/DocumentationSignatureParser.cs
```diff
@@ -0,0 +1,666 @@
...
+				nameBuilder.Append (name);
+
+				List<string>? typeArguments = null;
+				int arity = 0;
+
+				// type arguments
+				if (PeekNextChar (id, index) == '{') {
+					typeArguments = new List<string> ();
+					if (!ParseTypeArguments (id, ref index, typeParameterContext, typeArguments)) {
+						continue;
```
mateoatr, line 341: Should we output a warning in places where the parsing fails?

vitek-karas: Ideally yes, but I'm not sure in this case. Roslyn doesn't raise a warning when this is malformed in `SuppressMessage` - it simply ignores it in that case. So I think linker should do the same in this case.

test/Mono.Linker.Tests/Tests/DocumentationSignatureParserTests.cs
sbomer: Filed https://github.com/mono/linker/issues/1214

src/linker/Linker/DocumentationSignatureParser.cs
sbomer: Agreed it would be nice. Currently the way it will be done is that the caller of the parser gets to decide what kind of warnings/errors to log - which unfortunately doesn't distinguish between syntax errors and resolution errors. I initially tried to separate those two cases, but ended up just following the Roslyn implementation.

PR #1204, opened 2020-05-19
Implement support for RequiresUnreferencedCodeAttribute

### LINKED ISSUE #1166: Implement RequiresUnreferencedCodeAttribute in the linker
Implement the linker behaviors associated with the `RequiresUnreferencedCodeAttribute`. See https://github.com/dotnet/runtime/issues/33862 for the API design and discussion.

vitek-karas:
Introduces a cache of "linker attributes", currently only RequiresUnreferencedCodeAttribute is supported, but it's highly likely we will need this for other attributes as well (DynamicDependency, future AOT/Single-file related attributes, UnconditionalSuppressMessage, ...). The cache is on Annotations class and should also act as a natural point to unify attribute and XML based inputs (once we have the XML based attribute injection).

RequiresUnreferencedCode now causes a linker warning whenever a method with it is called.
It also auto-suppresses warnings within the method with this attribute - it actually disables data flow analysis on the method right now.

In the future when we have AOT/Single-File related analysis we will need to modify this to selectively support a subset and only auto-suppress warnings related to the relevant subsets.

Extends the LogContains/LogDoesNotContain test attributes so that they can appear on anything within the test class. Functionality is still the same as if they're on the class, but it makes it easier to structure the tests and make them more readable. Once we have solid warning origins, we could also auto-validate that the warning came from the method it is on.

Fixes #1166

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -163,6 +163,10 @@ error and warning codes.
 
 - The linker found a call to a method which is annotated with 'RequiresUnreferencedCodeAttribute' which can break functionality of a trimmed application.
 
-#### `IL2027`: Found multiple instances of attribute 'attribut'` on 'method', but only one is allowed. Only the first one will be used, the others will be ignored.
+#### `IL2027`: Attribute 'attribute' should only be used once on 'method'.
 
 - The linker found multiple instances of attribute 'attribute' on 'method'. This attribute is only allowed to have one instance, linker will only use the fist instance and ignore the rest.
+
+#### `IL2028`: Attribute 'attribute' on 'method' doesn't have a required constructor argument.
+
+- The linker found an instance of attribute 'attribute' on 'method' but it lacks a required constructor argument. Linker will ignore this attribute.
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -158,3 +158,7 @@ error and warning codes.
 #### `IL2025`: Duplicate preserve of 'member' in 'XML document location'
 
 - The XML descriptor marks for preservation the member or type 'member' more than once.
+
+#### `IL2026`: Calling method annotated with `RequiresUnreferencedCodeAttribute`
+
+- The linker found a call to a method which is annotated with `RequiresUnreferencedCodeAttribute` which can break functionality of a trimmed application.
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -161,4 +161,8 @@ error and warning codes.
 
 #### `IL2026`: Calling method annotated with `RequiresUnreferencedCodeAttribute`
 
-- The linker found a call to a method which is annotated with `RequiresUnreferencedCodeAttribute` which can break functionality of a trimmed application.
+- The linker found a call to a method which is annotated with 'RequiresUnreferencedCodeAttribute' which can break functionality of a trimmed application.
+
+#### `IL2027`: Found multiple instances of attribute 'attribut'` on 'method', but only one is allowed. Only the first one will be used, the others will be ignored.
+
+- The linker found multiple instances of attribute 'attribute' on 'method'. This attribute is only allowed to have one instance, linker will only use the fist instance and ignore the rest.
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -25,10 +25,10 @@ namespace Mono.Linker.Dataflow
 		{
 			MethodDefinition methodDefinition = calledMethod.Resolve ();
 			if (methodDefinition != null) {
-				return 
+				return
 					GetIntrinsicIdForMethod (methodDefinition) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel ||
 					flowAnnotations.RequiresDataFlowAnalysis (methodDefinition) ||
-					context.Annotations.TryGetLinkerAttribute <RequiresUnreferencedCodeAttribute> (methodDefinition, out _);
+					context.Annotations.TryGetLinkerAttribute<RequiresUnreferencedCodeAttribute> (methodDefinition, out _);
 			}
 
 			return false;
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -21,21 +21,46 @@ namespace Mono.Linker.Dataflow
 		readonly MarkStep _markStep;
 		readonly FlowAnnotations _flowAnnotations;
 
-		public static bool RequiresReflectionMethodBodyScanner (FlowAnnotations flowAnnotations, MethodReference method)
+		public static bool RequiresReflectionMethodBodyScannerForCallSite (LinkContext context, FlowAnnotations flowAnnotations, MethodReference calledMethod)
 		{
-			MethodDefinition methodDefinition = method.Resolve ();
+			MethodDefinition methodDefinition = calledMethod.Resolve ();
 			if (methodDefinition != null) {
-				return GetIntrinsicIdForMethod (methodDefinition) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel || flowAnnotations.RequiresDataFlowAnalysis (methodDefinition);
+				return 
+					GetIntrinsicIdForMethod (methodDefinition) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel ||
+					flowAnnotations.RequiresDataFlowAnalysis (methodDefinition) ||
+					context.Annotations.TryGetLinkerAttribute <RequiresUnreferencedCodeAttribute> (methodDefinition, out _);
 			}
 
 			return false;
 		}
 
-		public static bool RequiresReflectionMethodBodyScanner (FlowAnnotations flowAnnotations, FieldReference field)
+		public static bool RequiresReflectionMethodBodyScannerForMethodBody (FlowAnnotations flowAnnotations, MethodReference method)
+		{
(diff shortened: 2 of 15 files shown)


src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -0,0 +1,67 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	struct LinkerAttributesInformation
+	{
+		Dictionary<Type, object> _linkerAttributes;
```
sbomer, line 14: I think we'll need support for multiple attributes on a member

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2460,10 +2460,14 @@ protected virtual void MarkMethodBody (MethodBody body)
 				if (eh.HandlerType == ExceptionHandlerType.Catch)
 					MarkType (eh.CatchType, new DependencyInfo (DependencyKind.CatchType, body.Method));
 
-			bool requiresReflectionMethodBodyScanner = ReflectionMethodBodyScanner.RequiresReflectionMethodBodyScanner (_flowAnnotations, body.Method);
+			bool requiresReflectionMethodBodyScanner =
+				ReflectionMethodBodyScanner.RequiresReflectionMethodBodyScannerForMethodBody (_flowAnnotations, body.Method);
 			foreach (Instruction instruction in body.Instructions)
 				MarkInstruction (instruction, body.Method, ref requiresReflectionMethodBodyScanner);
 
+			if (ReflectionMethodBodyScanner.AutomaticallySuppressReflectionmethodBodyScannerForMethod (_context, body.Method))
```
sbomer, line 2468:
```suggestion
			if (ReflectionMethodBodyScanner.AutomaticallySuppressReflectionMethodBodyScannerForMethod (_context, body.Method))
```

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -0,0 +1,67 @@
...
+
+			if (_linkerAttributes != null && _linkerAttributes.TryGetValue (typeof (T), out var returnValue)) {
+				attributeValue = (T) returnValue;
+				return true;
+			}
+
+			return false;
+		}
+
+		void AddAttribute (object attribute)
```
mateoatr, line 39: Why not use `Attribute` instead of `object` in the parameter?

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -0,0 +1,67 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	struct LinkerAttributesInformation
```
marek-safar, line 12: I'd prefer to see readonly struct

src/linker/Linker/LinkerAttributesInformation.cs
marek-safar: I don't think we need to add multiple attributes at this moment but I'd change the signature to be `Dictionary<Type, Attribute> _linkerAttributes;`

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -40,6 +55,16 @@ public static bool RequiresReflectionMethodBodyScanner (FlowAnnotations flowAnno
 			return false;
 		}
 
+		public static bool AutomaticallySuppressReflectionmethodBodyScannerForMethod (LinkContext context, MethodReference method)
+		{
+			MethodDefinition methodDefinition = method.Resolve ();
+			if (methodDefinition != null) {
+				return context.Annotations.TryGetLinkerAttribute<RequiresUnreferencedCodeAttribute> (methodDefinition, out _);
```
marek-safar, line 62: What aboud adding `HasLinkerAttribute (methodDefinition)` instead?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -975,6 +999,18 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 						reflectionContext.RecordHandledPattern ();
 					}
 
+					if (_context.Annotations.TryGetLinkerAttribute<RequiresUnreferencedCodeAttribute> (calledMethodDefinition, out var requiresUnreferencedCode)) {
```
marek-safar, line 1002:
```suggestion
					if (_context.Annotations.TryGetLinkerAttribute (calledMethodDefinition, out RequiresUnreferencedCodeAttribute requiresUnreferencedCode)) {
```

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -0,0 +1,67 @@
...
+				foreach (var prop in customAttribute.Properties) {
+					if (prop.Name == "Url") {
+						url = (string) prop.Argument.Value;
+					}
+				}
+
+				return new RequiresUnreferencedCodeAttribute (message) { Url = url };
+			}
+
+			return null;
```
marek-safar, line 64: Shouldn't this assert/unreachable instead?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -40,6 +55,16 @@ public static bool RequiresReflectionMethodBodyScanner (FlowAnnotations flowAnno
 			return false;
 		}
 
+		public static bool AutomaticallySuppressReflectionmethodBodyScannerForMethod (LinkContext context, MethodReference method)
```
MichalStrehovsky, line 58:
```suggestion
		public static bool AutomaticallySuppressReflectionMethodBodyScannerForMethod (LinkContext context, MethodReference method)
```

docs/error-codes.md
```diff
@@ -161,4 +161,8 @@ error and warning codes.
 
 #### `IL2026`: Calling method annotated with `RequiresUnreferencedCodeAttribute`
 
-- The linker found a call to a method which is annotated with `RequiresUnreferencedCodeAttribute` which can break functionality of a trimmed application.
+- The linker found a call to a method which is annotated with 'RequiresUnreferencedCodeAttribute' which can break functionality of a trimmed application.
+
+#### `IL2027`: Found multiple instances of attribute 'attribut'` on 'method', but only one is allowed. Only the first one will be used, the others will be ignored.
```
marek-safar, line 166:
```suggestion
#### `IL2027`: Attribute 'attribute' should be used only once on 'method'
```

src/linker/Linker/LinkerAttributesInformation.cs
vitek-karas: I implemented support for multiple attributes for two reasons:
* It includes a check for multiple instances for cases where we expect only one - and issues a warning (this is good even for this PR, although there's no way in C# to get to that situation right now)
* Sven will need it soon anyway, and since I'm at it ...

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -5,49 +5,62 @@
...
 		}
 
-		void AddAttribute (object attribute)
+		public IEnumerable<T> GetAttributes<T> () where T : Attribute
 		{
-			if (attribute == null)
-				return;
+			if (_linkerAttributes != null && _linkerAttributes.TryGetValue (typeof (T), out var attributeList)) {
+				if (attributeList == null || attributeList.Count == 0) {
+					throw new LinkerFatalErrorException (MessageContainer.CreateErrorMessage ("IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues \nUnexpected list of attributes.", 1012));
```
marek-safar, line 54: Perhaps that should be inside the default LinkerFatalErrorException ctor

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -5,49 +5,62 @@
...
+		{
+			return _linkerAttributes != null && _linkerAttributes.ContainsKey (typeof (T));
 		}
 
-		void AddAttribute (object attribute)
+		public IEnumerable<T> GetAttributes<T> () where T : Attribute
 		{
-			if (attribute == null)
-				return;
+			if (_linkerAttributes != null && _linkerAttributes.TryGetValue (typeof (T), out var attributeList)) {
```
marek-safar, line 52: return early on `if (_linkerAttributes == null || !_linkerAttributes.TryGetValue (typeof (T), out var attributeList))`

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -5,49 +5,62 @@
...
+						attributeList.Add (attributeValue);
+					}
+				}
 			}
+		}
 
-			return false;
+		public bool HasAttribute<T> () where T : Attribute
+		{
+			return _linkerAttributes != null && _linkerAttributes.ContainsKey (typeof (T));
```
marek-safar, line 47:
```suggestion
			return _linkerAttributes?.ContainsKey (typeof (T));
```

src/linker/Linker/LinkerAttributesInformation.cs
vitek-karas: I changed it to issue a warning and ignore the attribute.

src/linker/Linker/LinkerAttributesInformation.cs
vitek-karas: Such expression returns bool? and so I would have to also add == true  - which honestly makes the whole thing less readable.

PR #1210, opened 2020-05-25
Sweep IVT references to removed assemblies

marek-safar:


akoeplinger:
This seems to cause an issue in the dotnet/runtime build: https://github.com/dotnet/runtime/pull/37006

PR #1212, opened 2020-05-26
Respect flow annotations on RuntimeType and TypeBuilder

MichalStrehovsky:
`RuntimeType` and `TypeBuilder` are both interesting from the dataflow perspective.

### DIFF
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -288,8 +288,9 @@ namespace Mono.Linker.Dataflow
 			// It's necessary to be able to implement a custom "Type" type in tests to validate
 			// the correct propagation of the annotations on "this" parameter. And tests can't really
 			// override System.Type - as it creates too many issues.
-			return (typeReference.Name.StartsWith ("Type") || typeReference.Name == "String") &&
-				typeReference.Namespace == "System";
+			// It also covers TypeBuilder and RuntimeType.
+			return typeReference.MetadataType == MetadataType.String ||
+				(typeReference.Name.Contains ("Type") && typeReference.Namespace.StartsWith ("System"));
 		}
 
 		readonly struct TypeAnnotations


vitek-karas:
I just found that there's one other place where we have this check:
https://github.com/mono/linker/blob/d97424c6bef56697c6fee9757ae6a532524f66aa/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs#L122

Not sure if it's applicable (I don't think so, as attributes can only have literals as values, and that means `System.Type`, I don't think there's a way to create a `RuntimeTypeHandle` as a literal), but worth a look.

MichalStrehovsky:
> I don't think so, as attributes can only have literals as values

Yes, it wouldn't be possible to encode descendants of System.Type in the file format. Only System.Type can be encoded in the file format. That place doesn't need updating.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -288,8 +288,9 @@ static bool IsTypeInterestingForDataflow (TypeReference typeReference)
 			// It's necessary to be able to implement a custom "Type" type in tests to validate
 			// the correct propagation of the annotations on "this" parameter. And tests can't really
 			// override System.Type - as it creates too many issues.
-			return (typeReference.Name.StartsWith ("Type") || typeReference.Name == "String") &&
-				typeReference.Namespace == "System";
+			// It also covers TypeBuilder and RuntimeType.
+			return typeReference.MetadataType == MetadataType.String ||
+				(typeReference.Name.Contains ("Type") && typeReference.Namespace.StartsWith ("System"));
```
marek-safar, line 293: Is it ok that this includes types like https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedtype?view=netcore-3.1 or https://docs.microsoft.com/en-us/dotnet/api/system.typedreference?view=netcore-3.1 ?

MichalStrehovsky: We only do this filtering to help perf a bit to avoid looking for attributes on everything.

This just means that we're going to look for the dataflow annotations on locations typed like that. And if someone puts dataflow annotations on those places, it will be basically impossible to actually satisfy them. I don't have an opinion on this - it feels dirty to hardcode `RuntimeType` knowledge in the linker, but we can be more explicit here too...

Issue #1140, opened 2020-04-24
DynamicDependencyAttribute should not disable flow analysis in the method


MichalStrehovsky:
We now have an attribute that can be used to selectively disable specific warnings. Disabling analysis of the entire method feels like a pretty big hammer - developers should ideally be able to see all places that have warning suppressions and for that, the suppression should be explicit here too.

Cc @vitek-karas for thoughts

vitek-karas:
I do agree that we should remove the "big hammer" part of it. I guess we could go with it as you describe above and see if it becomes a problem (as in too many people have to specify both `DynamicDependency` and `UnconditionalSuppressMessage` at the same time for the same reason).

sbomer:
https://github.com/mono/linker/pull/1215/files#diff-83d9d9aa53da0b0737513a2561f82bc8L2629 removed the big hammer check. Is there anything else that needs to be done here?

vitek-karas:
I think if we just add a test to validate that we still raise warnings in this case... and that should be it.

PR #1215, opened 2020-05-26
Support DynamicDependencyAttribute

sbomer:
This adds support for DynamicDependencyAttribute, which uses the doc comment signature parser from https://github.com/mono/linker/pull/1197.

The behavior of the lookup step is pretty much kept the same - it will resolve assemblies from attributes, without recursively processing attributes from those assemblies. I originally started to do the recursive processing, but we agreed not to change it here since it is related to the broader work to enable adding new assemblies to the context in MarkStep, and there were concerns about the performance of loading extra unnecessary assemblies into the context.


docs/error-codes.md
```diff
@@ -163,10 +163,34 @@ error and warning codes.
...
+
+- The type in a DynamicDependencyAttribute constructor could not be resolved. Ensure that the argument specifies a valid type name or type reference, that the type exists in the specified assembly, and that the assembly is available to the linker.
+
+### `IL2033`: Unresolved member 'memberSignature' in DynamicDependencyAttribute on 'member'
+
+- The member signature in a DynamicDependencyAttribute constructor could not be resolved. Ensure that it refers to an existing member, and that it uses the format defined at https://github.com/dotnet/csharplang/blob/master/spec/documentation-comments.md#id-string-format.
+
+### `IL2034`: No members were resolved for for 'memberTypes' in DynamicDependencyAttribute on 'member'
+
+- The DynamicallyAccessedMemberTypes passed into the DynamicDependencyAttribute constructor did not match any members on the specified type. No members will be kept for this attribute. Ensure that you are using DynamicallyAccessedMemberTypes which exist for the type.
```
vitek-karas, line 196:
```suggestion
- The DynamicallyAccessedMemberTypes passed into the DynamicDependencyAttribute constructor did not match any members on the specified type. No members will be kept for this type. Ensure that you are using DynamicallyAccessedMemberTypes which exist for the type.
```

src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,297 @@
...
+namespace Mono.Linker.Dataflow
+{
+	internal static class DynamicallyAccessedMembersBinder
+	{
+		// Returns the members of the type bound by memberTypes. For MemberTypes.All, this returts a single member
+		// which is the type itself.
+		public static IEnumerable<IMemberDefinition> GetDynamicallyAccessedMembers (TypeDefinition typeDefinition, DynamicallyAccessedMemberTypes memberTypes)
+		{
+			if (memberTypes == DynamicallyAccessedMemberTypes.All) {
+				yield return typeDefinition;
```
vitek-karas, line 16: This doesn't feel right - we're returning the type itself?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1238,286 +1238,116 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
 
 		void MarkTypeForDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionContext, TypeDefinition typeDefinition, DynamicallyAccessedMemberTypes requiredMemberKinds)
 		{
-			if (requiredMemberKinds == DynamicallyAccessedMemberTypes.All) {
-				var sourceMethod = reflectionContext.SourceMethod;
-				reflectionContext.RecordRecognizedPattern (typeDefinition, () => _markStep.MarkEntireType (typeDefinition, includeBaseTypes: true, new DependencyInfo (DependencyKind.AccessedViaReflection, sourceMethod)));
-				return;
+			foreach (var member in DynamicallyAccessedMembersBinder.GetDynamicallyAccessedMembers (typeDefinition, requiredMemberKinds)) {
+				if (member == typeDefinition) {
```
vitek-karas, line 1242: I would probably keep this a special case in the caller - and force everybody who calls the helper to handle All as a special case as well - it has somewhat special meaning compared to the other flags.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -544,24 +557,132 @@ void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo
 				MarkCustomAttribute (ca, reason);
 				MarkSpecialCustomAttributeDependencies (ca, provider);
 			}
+
+			// avoid allocating linker attributes for members that don't need them
```
vitek-karas, line 561: 1. I assume we do this because otherwise we would populate the linker attribute cache for all methods/fields everywhere. If that's the case please change the comment to note about that. Also note that the existing usage for `DynamicallyAccessedMembersAttribute` will already populate the cache for all marked methods and fields - so I don't think we need this optimization here (otherwise we would have to add it to that case as well, but there it's probably a regression). I would still keep it in the assembly resolution step as that runs on everything, not just marked members.
1. I think this is worth adding as a functionality into the cache somehow - maybe we could simply have a `GetLinkerAttributesNoEmptyCache` or something like that? Which would act just like GetLinkerAttributes but it would not add a record for cases where there are no attributes to remember. The point being that you already have this same "logic/code" in two places.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -544,24 +557,132 @@ void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo
...
+				}
+			}
+
+			MarkMembers (type, members, new DependencyInfo (DependencyKind.DynamicDependency, dynamicDependency.OriginalAttribute));
+		}
+
+		void MarkMembers (TypeDefinition typeDefinition, IEnumerable<IMemberDefinition> members, DependencyInfo reason)
+		{
+			foreach (var member in members) {
+				if (member == typeDefinition) {
```
vitek-karas, line 660: Hmm - maybe we should have a "sentinel" value for this. Could be `null` or could be an artificial member (just new up empty member from Cecil). Not sure - I definitely don't like returning the type itself again - it lends itself to easy endless recursion bugs.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2607,7 +2716,21 @@ protected virtual void MarkInterfaceImplementation (InterfaceImplementation ifac
...
+			if (!method.HasCustomAttributes)
+				return false;
+
+			// avoid allocating linker attributes for members that don't need them
+			foreach (var ca in method.CustomAttributes) {
+				if (LinkerAttributesInformation.IsAttribute<DynamicDependencyAttribute> (ca.AttributeType))
+					return true;
+#if !FEATURE_ILLINK
+				if (DynamicDependencyLookupStep.IsPreserveDependencyAttribute (ca.AttributeType))
+					return true;
```
vitek-karas, line 2730: I would very much like to avoid doing this. See https://github.com/mono/linker/issues/1140.

Since we have the suppression mechanism maybe that is the better solution for this problem. Also in most cases where we use it, the code will not generate warnings, we use the attribute as basically just a workaround for the assembly loading issue.

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -12,42 +13,63 @@ namespace Mono.Linker
 {
 	readonly struct LinkerAttributesInformation
 	{
-		readonly Dictionary<Type, List<Attribute>> _linkerAttributes;
+		readonly Dictionary<Type, List<object>> _linkerAttributes;
```
vitek-karas, line 16: I thought about this - and I'm slightly in favor of forcing the Attribute here. If we need a special version for the linker (like in your case) just make the internal linker class derive from Attribute - it doesn't really hurt anything. I think it's good to have the intent of storing structural information here - with object it feels like it's fine to store just simple values - which I think we should avoid (we would not get rid of allocation as they would be boxed anyway, and having the extra metadata is a good thing in my mind)

src/linker/Linker/Annotations.cs
```diff
@@ -56,6 +56,7 @@ public partial class AnnotationStore
 		protected readonly Dictionary<MethodDefinition, List<MethodDefinition>> base_methods = new Dictionary<MethodDefinition, List<MethodDefinition>> ();
 		protected readonly Dictionary<AssemblyDefinition, ISymbolReader> symbol_readers = new Dictionary<AssemblyDefinition, ISymbolReader> ();
 		readonly Dictionary<MethodDefinition, LinkerAttributesInformation> method_linker_attributes = new Dictionary<MethodDefinition, LinkerAttributesInformation> ();
+		readonly Dictionary<FieldDefinition, LinkerAttributesInformation> field_linker_attributes = new Dictionary<FieldDefinition, LinkerAttributesInformation> ();
```
vitek-karas, line 59: I already though about this - I think a better solution is to key the Dictionary with ICustomAttributeProvider and make the cache general. I don't see much value in having special code in the cache which would know which attributes apply to which types (it doesn't need to for anything really).

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -12,42 +13,63 @@ namespace Mono.Linker
...
+				}
+			}
+		}
 
-						Type attributeValueType = attributeValue.GetType ();
-						if (!_linkerAttributes.TryGetValue (attributeValueType, out var attributeList)) {
-							attributeList = new List<Attribute> ();
-							_linkerAttributes.Add (attributeValueType, attributeList);
-						}
+		public LinkerAttributesInformation (LinkContext context, FieldDefinition field)
```
vitek-karas, line 35: As noted above - I would simply have this based on ICustomAttributeProvider - and no special code for different types...

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -59,7 +81,13 @@ public LinkerAttributesInformation (LinkContext context, MethodDefinition method
 			return attributeList.Cast<T> ();
 		}
 
-		static Attribute ProcessRequiresUnreferencedCodeAttribute (LinkContext context, MethodDefinition method, CustomAttribute customAttribute)
+		public static bool IsAttribute<T> (TypeReference tr)
```
vitek-karas, line 84: Nice - I like this :-)

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -80,5 +108,80 @@ static Attribute ProcessRequiresUnreferencedCodeAttribute (LinkContext context,
...
+				return null;
+
+			// Third argument is assembly name
+			if (!(args[2].Value is string assemblyName))
+				return null;
+
+			return memberSignature == null ? new DynamicDependency (memberTypes.Value, typeName, assemblyName) : new DynamicDependency (memberSignature, typeName, assemblyName);
+		}
+
+		public static bool ShouldProcessDependencyAttribute (LinkContext context, CustomAttribute ca)
```
vitek-karas, line 166: Why not have this as a static member on the DynamicDependency class - I don't feel it belongs here really.

src/linker/System.Diagnostics.CodeAnalysis/DynamicDependencyAttribute.cs
```diff
@@ -0,0 +1,136 @@
...
+	/// This can be used to inform tooling of a dependency that is otherwise not evident purely from
+	/// metadata and IL, for example a member relied on via reflection.
+	/// </remarks>
+	/// <remarks>
+	/// This is a copy of the enum definition in the framework at https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/DynamicDependencyAttribute.cs
+	/// </remarks>
+	[AttributeUsage (
+		AttributeTargets.Constructor | AttributeTargets.Field | AttributeTargets.Method,
+		AllowMultiple = true, Inherited = false)]
+	internal sealed class DynamicDependencyAttribute : Attribute
```
vitek-karas, line 22: Since you have the internal version of this - why do we need a copy of the declaration now?

test/Mono.Linker.Tests.Cases/DynamicDependencies/DynamicDependencyMemberSignatureWildcard.cs
```diff
@@ -0,0 +1,23 @@
...
+	[SetupCompileBefore ("library.dll", new[] { "Dependencies/DynamicDependencyMethodInAssemblyLibrary.cs" })]
+	public class DynamicDependencyMemberSignatureWildcard
+	{
+		public static void Main ()
+		{
+			Dependency ();
+		}
+
+		[Kept]
+		[DynamicDependency ("*", "Mono.Linker.Tests.Cases.DynamicDependencies.Dependencies.DynamicDependencyMethodInAssemblyLibrary", "library")]
```
vitek-karas, line 18: I thought that we said we would not support wildecards - instead we have the DynamicallyAccessedMemberTypes for that purpose... right?

src/linker/System.Diagnostics.CodeAnalysis/DynamicDependencyAttribute.cs
sbomer: Now it's only being used for `IsAttribute<T>`. If you prefer, I can get rid of that and use `IsDynamicDependencyAttribute` instead - let me know.

test/Mono.Linker.Tests.Cases/DynamicDependencies/DynamicDependencyMemberSignatureWildcard.cs
sbomer: This test is just checking the warning.

src/linker/System.Diagnostics.CodeAnalysis/DynamicDependencyAttribute.cs
vitek-karas: I don't mind keeping it, but I would only do it as an empty shell - to make it clear - we're going to delete it once we move over to 5 anyway.

test/Mono.Linker.Tests.Cases/DynamicDependencies/DynamicDependencyMemberSignatureWildcard.cs
vitek-karas: Sorry - I missed that.

src/linker/Linker.Steps/MarkStep.cs
sbomer: I changed it to use a `null` sentinel.

src/linker/Linker/LinkerAttributesInformation.cs
sbomer: Forcing it to derive from `Attribute` feels weird, but I see your point - and agree that it doesn't really hurt. I changed it back to `Attribute`.

src/linker/Linker/LinkerAttributesInformation.cs
sbomer: Changed to `IMemberDefinition` which implements `ICustomAttributeProvider`. If we need to use the cache for attributes on non-members, we can always relax it to `ICustomAttributeProvider`.

src/linker/Linker/LinkerAttributesInformation.cs
sbomer: :)

src/linker/Linker/LinkerAttributesInformation.cs
sbomer: Moved

src/linker/Linker.Steps/MarkStep.cs
sbomer: Originally I wanted to avoid setting up the cache for members without DynamicDependencyAttribute. `GetLinkerAttributesNoEmptyCache` would avoid inserting a struct into the Annotations dictionary for every member - on the other hand, it would still do unnecessary processing of linker attributes for unmarked members, and also requires members without linker attributes to be re-scanned later. Without something to measure it's not clear to me what is better.

For now, maybe we should just avoid populating the cache for members without any attributes. There will still be entries in the Annotations dictionary for members with non-linker attributes, but prevents a lot of the unnecessary structs already and doesn't require the callsite to use a different helper. See here: https://github.com/mono/linker/pull/1215/commits/5c637fad798051546245372ac7eb6c403f0d68c5#diff-efcf63149081c7bcad6e9ab7c490ac0bR433. If we hit a case where the perf is still an issue it might be easier to optimize that when we can measure it. What do you think?

vitek-karas: I like that idea... (hopefully nullable attributes won't ruin it :-)) - we should definitely do that.

src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,296 @@
...
+					yield return e;
+			}
+
+			if (memberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicEvents)) {
+				foreach (var e in GetEventsOnTypeHierarchy (typeDefinition, filter: null, bindingFlags: BindingFlags.Public))
+					yield return e;
+			}
+		}
+
+		public static IEnumerable<MethodDefinition> GetConstructorsOnType (TypeDefinition type, Func<MethodDefinition, bool> filter, BindingFlags? bindingFlags = null)
```
marek-safar, line 85: These are local helpers, right? Why are they public?

src/linker/Linker.Steps/DynamicDependencyLookupStep.cs
```diff
@@ -0,0 +1,100 @@
...
+					continue;
+
+				var assembly = Context.Resolve (new AssemblyNameReference (assemblyName, new Version ()));
+				if (assembly == null)
+					continue;
+				ProcessReferences (assembly);
+#endif
+			}
+
+			var dynamicDependencies = Context.Annotations.GetLinkerAttributes<DynamicDependency> (member);
```
marek-safar, line 77: I don't think you should use `GetLinkerAttributes` here, you already iterate all attributes above so this is doing the same loop and also building cache for member which might not be reached at all.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -544,24 +554,123 @@ void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo
 				MarkCustomAttribute (ca, reason);
 				MarkSpecialCustomAttributeDependencies (ca, provider);
 			}
+
+			if (!(provider is MethodDefinition || provider is FieldDefinition))
+				return;
+
+			var dynamicDependencies = _context.Annotations.GetLinkerAttributes<DynamicDependency> ((IMemberDefinition) provider);
+			Debug.Assert (dynamicDependencies != null);
```
marek-safar, line 562: What is this assert for ?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -544,24 +554,123 @@ void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo
...
+			}
+
+			return true;
+		}
+
+		void MarkDynamicDependency (DynamicDependency dynamicDependency, MemberReference context)
+		{
+			Debug.Assert (context is MethodDefinition || context is FieldDefinition);
+			AssemblyDefinition assembly;
+			if (dynamicDependency.AssemblyName is string assemblyName) {
```
marek-safar, line 595: Why `is string` if the AssemblyName is string?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -544,24 +554,123 @@ void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo
...
+				assembly = _context.GetLoadedAssembly (assemblyName);
+				if (assembly == null) {
+					_context.LogMessage (MessageContainer.CreateWarningMessage ($"Unresolved assembly '{assemblyName}' in DynamicDependencyAttribute on '{context}'", 2031));
```
marek-safar, line 598: Will that ever happen when you do same check during load?

src/linker/Linker/DynamicDependency.cs
```diff
@@ -0,0 +1,140 @@
...
+#nullable enable
+
+namespace Mono.Linker
+{
+	/// Tracks dependencies created via DynamicDependencyAttribute in the linker.
+	/// This is almost identical to DynamicDependencyAttribute, but it holds a
+	/// TypeReference instead of a Type, and it has a reference to the original
+	/// CustomAttribute for dependency tracing. It is also a place for helper
+	/// methods related to the attribute.
+	internal class DynamicDependency : Attribute
```
marek-safar, line 15: Why does it derive from Attribute?

src/linker/Linker/DynamicDependency.cs
```diff
@@ -0,0 +1,140 @@
...
+		public string? AssemblyName { get; }
+
+		public string? Condition { get; set; }
+
+		public static DynamicDependency? ProcessAttribute (LinkContext context, ICustomAttributeProvider provider, CustomAttribute customAttribute)
+		{
+			if (!(provider is MethodDefinition || provider is FieldDefinition))
+				return null;
+
+			if (!DynamicDependency.ShouldProcess (context, customAttribute))
```
marek-safar, line 66: Redundant `DynamicDependency.` prefix

src/linker/Linker/DynamicDependency.cs
```diff
@@ -0,0 +1,140 @@
...
+			string? memberSignature = args[0].Value as string;
+			if (args.Count == 1)
+				return memberSignature == null ? null : new DynamicDependency (memberSignature);
+			DynamicallyAccessedMemberTypes? memberTypes = null;
+			if (memberSignature == null) {
+				var argType = args[0].Type;
+				if (!(argType.Namespace == "System.Diagnostics.CodeAnalysis" && argType.Name == "DynamicallyAccessedMemberTypes"))
+					return null;
+				try {
+					memberTypes = (DynamicallyAccessedMemberTypes) args[0].Value;
```
marek-safar, line 93: This is wrong you are casting user type to linker compiled type

src/linker/Linker/DynamicDependency.cs
```diff
@@ -0,0 +1,140 @@
...
+				return null;
+
+			// First argument is string or DynamicallyAccessedMemberTypes
+			string? memberSignature = args[0].Value as string;
+			if (args.Count == 1)
+				return memberSignature == null ? null : new DynamicDependency (memberSignature);
+			DynamicallyAccessedMemberTypes? memberTypes = null;
+			if (memberSignature == null) {
+				var argType = args[0].Type;
+				if (!(argType.Namespace == "System.Diagnostics.CodeAnalysis" && argType.Name == "DynamicallyAccessedMemberTypes"))
```
marek-safar, line 90: Use `IsTypeOf`

src/linker/Linker/AssemblyUtilities.cs
```diff
@@ -39,5 +40,16 @@ public static TypeDefinition FindType (this AssemblyDefinition assembly, string
 			var type = assembly.MainModule.GetType (fullName);
 			return type?.Resolve ();
 		}
+
+		// Takes a documentation signature (not including the documentation member type prefix) and resolves it to a type
+		// in the assembly.
+		public static TypeDefinition FindTypeByDocumentationSignature (this AssemblyDefinition assembly, string signature)
```
marek-safar, line 46: I don't think this should be here, it won't be universally used and you have to prefix it with DocumentationSignature name already

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1238,286 +1238,115 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
 
 		void MarkTypeForDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionContext, TypeDefinition typeDefinition, DynamicallyAccessedMemberTypes requiredMemberKinds)
 		{
-			if (requiredMemberKinds == DynamicallyAccessedMemberTypes.All) {
-				var sourceMethod = reflectionContext.SourceMethod;
-				reflectionContext.RecordRecognizedPattern (typeDefinition, () => _markStep.MarkEntireType (typeDefinition, includeBaseTypes: true, new DependencyInfo (DependencyKind.AccessedViaReflection, sourceMethod)));
-				return;
+			foreach (var member in DynamicallyAccessedMembersBinder.GetDynamicallyAccessedMembers (typeDefinition, requiredMemberKinds)) {
```
marek-safar, line 1241: Would be better to do such refactoring in separate PR

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -544,24 +554,123 @@ void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo
...
+				if (type == null) {
+					_context.LogMessage (MessageContainer.CreateWarningMessage ($"Unresolved type '{typeReference}' in DynamicDependencyAtribute on '{context}'", 2032));
+					return;
+				}
+			} else {
+				type = context.DeclaringType.Resolve ();
+				if (type == null) {
+					_context.LogMessage (MessageContainer.CreateWarningMessage ($"Unresolved type '{context.DeclaringType}' in DynamicDependencyAttribute on '{context}'", 2032));
+					return;
+				}
```
marek-safar, line 624: I'm tempted for this logic to respect `--skip-unresolved`

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -544,24 +554,123 @@ void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo
...
+				members = type.FindMembersByDocumentationSignature (memberSignature);
+				if (!members.Any ()) {
+					_context.LogMessage (MessageContainer.CreateWarningMessage ($"Unresolved member '{memberSignature}' in DynamicDependencyAttribute on '{context}'", 2033));
+					return;
+				}
+			} else {
+				var memberTypes = dynamicDependency.MemberTypes;
+				members = DynamicallyAccessedMembersBinder.GetDynamicallyAccessedMembers (type, memberTypes);
+				if (!members.Any ()) {
+					_context.LogMessage (MessageContainer.CreateWarningMessage ($"No members were resolved for '{memberTypes}' in DynamicDependencyAttribute on '{context}'", 2034));
```
marek-safar, line 638: Do we really need different error code/message from 2033?

src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,296 @@
...
+	{
+		// Returns the members of the type bound by memberTypes. For MemberTypes.All, this returns a single null result.
+		public static IEnumerable<IMemberDefinition> GetDynamicallyAccessedMembers (TypeDefinition typeDefinition, DynamicallyAccessedMemberTypes memberTypes)
+		{
+			if (memberTypes == DynamicallyAccessedMemberTypes.All) {
+				yield return null;
+				yield break;
+			}
+
+			if (memberTypes.HasFlag (DynamicallyAccessedMemberTypes.NonPublicConstructors)) {
```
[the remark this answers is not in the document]
MichalStrehovsky: `Enum.HasFlag` doesn't box anymore: https://github.com/dotnet/coreclr/pull/13748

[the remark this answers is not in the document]
marek-safar: Are you sure this applies to Unity and other linker users?

src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
vitek-karas: They are also used from ReflectionMethodBodyScanner - so maybe "internal" would be better.

src/linker/Linker/DynamicDependency.cs
vitek-karas: Because LinkerAttributesInformation enforces "Attribute" base class. Sven actually originally changed that to be "object" based, but in the end we didn't like it. I still think it's better for these to be attribute based - it makes it clearer what the class represents.
But if others don't like it I'll change the cache to be object based.

src/linker/Linker/DynamicDependency.cs
vitek-karas: This is enum - basically same story as with BindingFlags. Currently the type is defined in linker, once we switch over to .NET 5 it will come from the framework (on the linker side, just like BindingFlags do).

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: That is good point. Curious: do normal apps run linker with `--skip-unresolved` or is that more of a corner case option?

src/linker/Linker/DynamicDependency.cs
marek-safar: It's weird to see Attribute like without AttributeUsage. What does Attribute give us here?

src/linker/Linker/DynamicDependency.cs
marek-safar: I'm not sure this is going work like this. You probably need to cast it to int first as the type used inside linker does not have to much type in the linked assembly (basically there are 2 different mscorlibs/SPCs)

src/linker/Linker.Steps/MarkStep.cs
marek-safar: It's not a default option it's more about avoiding the aggressive context loading again

src/linker/Linker/DynamicDependency.cs
vitek-karas: You're right - I didn't realize that the runtime would do direct enum -> enum cast. We do need to go through integer.

src/linker/Linker/DynamicDependency.cs
vitek-karas: It doesn't give us anything functionality - that said we could use it - we could enforce `AttributeUsage` in the common code. But it's not very useful as it's highly unlikely we'll get an attribute on the wrong item (since C# would not allow it anyway).
I'm fine if we agree to change it to object based approach.

src/linker/Linker/DynamicDependency.cs
MichalStrehovsky: As long as the underlying type is the same, you can unbox into any enum you want.

```csharp
class Program
{
    enum SomeEnum { }
    enum OtherEnum : byte { }
    static void Main()
    {
        object o = (SomeEnum)123;
        _ = (OtherEnum)o;
    }
}
```

(I'm also pretty sure in this case the Value is a boxed integer, not an enum, but it works for that too.)

src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
sbomer: `DynamicallyAccessedMembersBinder` is already internal because our copy of `DynamicallyAccessedMemberTypes` is internal, so I think they are effectively internal already. Generally I prefer to use class-based accessibility modifiers for consistency with the rest of the codebase, especially since we have the ref assembly to control externally-visible public surface.

src/linker/Linker.Steps/DynamicDependencyLookupStep.cs
sbomer: I [originally](https://github.com/mono/linker/pull/1215/commits/a1e1c3249a90c37ece24be0fb3329ebe505fb1e6#diff-8bcf8e9be10f0ae524d90bad71f67db3R81) optimized this a little more, but removed it based on this discussion: https://github.com/mono/linker/pull/1215#discussion_r431010530. If you think it is worth it, I could add back the check for this callsite, and only call `GetLinkerAttributes` if the previous loop found any `DynamicDependencyAttribute`s.

src/linker/Linker.Steps/MarkStep.cs
sbomer: Just to assert that the API doesn't give back null. :) I can remove it.

src/linker/Linker.Steps/MarkStep.cs
sbomer: To extract the value out of nullable AssemblyName (null check and assign the member to a local).

src/linker/Linker.Steps/MarkStep.cs
sbomer: Unfortunately it could happen if an assembly added by `DynamicDependencyLookupStep` has further `DynamicDependencyAttributes` that depend on assemblies not in the context, because we don't do the lookup recursively.

src/linker/Linker/DynamicDependency.cs
sbomer: I will rename the `IsAttribute<T>` helper to `IsTypeOf<T>` and use that in both places.

src/linker/Linker/AssemblyUtilities.cs
sbomer: `DynamicDependency` doesn't require the prefix - but I will move it to the signature parser and we can always move it back if it is commonly used.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
sbomer: Done in https://github.com/mono/linker/pull/1223.

src/linker/Linker/DynamicDependency.cs
sbomer: The value is a boxed Int32 - so this is fine unless I misunderstood the concern.

src/linker/Linker/DynamicDependency.cs
sbomer: I plan to leave it as-is unless you would like me to change the cache to be object-based instead @marek-safar.

src/linker/Linker.Steps/MarkStep.cs
sbomer: We do pass it when running the linker from the SDK. As I understand it, that option enables the linker to continue from resolution failures (see https://github.com/mono/linker/pull/260) instead of crashing. I don't think we need to respect it here since these warnings won't fail the build. I think that if `--skip-unresolved` is false, the call to `Resolve()` would cause an `AssemblyResolutionException` and we wouldn't even get to log the warning.

marek-safar:
Let's land #1223 before this PR

sbomer:
@marek-safar I've addressed your feedback. PTAL

src/linker/Linker/AssemblyUtilities.cs
```diff
@@ -1,4 +1,5 @@
-using System;
+using System.Collections.Generic;
+using System.Diagnostics;
```
marek-safar, line 2: nit: some left over?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -534,24 +544,133 @@ void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo
...
+				Tracer.AddDirectDependency (ca, reason, marked: false);
+			}
+
+			return true;
+		}
+
+		void MarkDynamicDependency (DynamicDependency dynamicDependency, MemberReference context)
+		{
+			Debug.Assert (context is MethodDefinition || context is FieldDefinition);
+			var member = context as IMemberDefinition;
```
marek-safar, line 581: As it's used on unlikely paths only, move it there

src/linker/Linker/DynamicDependency.cs
```diff
@@ -0,0 +1,143 @@
...
+
+		public string? AssemblyName { get; }
+
+		public string? Condition { get; set; }
+
+		public static DynamicDependency? ProcessAttribute (LinkContext context, ICustomAttributeProvider provider, CustomAttribute customAttribute)
+		{
+			if (!(provider is MethodDefinition || provider is FieldDefinition))
+				return null;
+			var member = provider as IMemberDefinition;
```
marek-safar, line 65: Move it before it's used

src/linker/Linker/DynamicDependency.cs
```diff
@@ -0,0 +1,143 @@
...
+			}
+			Debug.Assert (args.Count == 3);
+			if (!(args[1].Value is string typeName))
+				return null;
+
+			// Third argument is assembly name
+			if (!(args[2].Value is string assemblyName))
+				return null;
+
+			var dynamicDependency = memberSignature == null ? new DynamicDependency (memberTypes!.Value, typeName, assemblyName) : new DynamicDependency (memberSignature, typeName, assemblyName);
```
marek-safar, line 116:
```suggestion
			return new DynamicDependency (memberSignature ?? memberTypes!.Value, typeName, assemblyName) {
				OriginalAttribute = ca
			};
```

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -77,5 +87,87 @@ static Attribute ProcessRequiresUnreferencedCodeAttribute (LinkContext context,
...
+
+			var dynamicDependency = memberSignature == null ? new DynamicDependency (memberTypes.Value, typeName, assemblyName) : new DynamicDependency (memberSignature, typeName, assemblyName);
+			dynamicDependency.OriginalAttribute = ca;
+
+			return dynamicDependency;
+		}
+
+		public static bool ShouldProcessDependencyAttribute (LinkContext context, CustomAttribute ca)
+		{
+			if (ca.HasProperties && ca.Properties[0].Name == "Condition") {
```
marek-safar, line 154: I'd probably not enable it for DynamicDependencyAttribute as it needs more thinking about what we want to do there

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -77,5 +87,87 @@ static Attribute ProcessRequiresUnreferencedCodeAttribute (LinkContext context,
...
+			if (dynamicDependency != null)
+				return dynamicDependency;
+
+			context.LogMessage (MessageContainer.CreateWarningMessage (context,
+				$"Invalid DynamicDependencyAttribute on '{member}'",
+				2034, MessageOrigin.TryGetOrigin (member)));
+			return null;
+		}
+
+		static DynamicDependency GetDynamicDependency (LinkContext context, CustomAttribute ca)
```
vitek-karas, line 110: Isn't this duplicated between DynamicDependency and here? I would keep the one in DynamiDependency.

src/linker/Linker/TypeDefinitionExtensions.cs
```diff
@@ -1,5 +1,6 @@
-
 using System;
+using System.Collections.Generic;
```
vitek-karas, line 2: Do we still need these?

src/linker/System.Diagnostics.CodeAnalysis/DynamicDependencyAttribute.cs
```diff
@@ -0,0 +1,14 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#nullable enable
+
+namespace System.Diagnostics.CodeAnalysis
+{
+	/// This is a n internal version of copy of the attribute in the framework at https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/DynamicDependencyAttribute.cs
```
vitek-karas, line 9:
```suggestion
	/// This is an internal version of copy of the attribute in the framework at https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/DynamicDependencyAttribute.cs
```

src/linker/Linker/LinkerAttributesInformation.cs
sbomer: Good catch, thank you.

src/linker/Linker.Steps/MarkStep.cs
sbomer: It's also used to get the `Module` when the attribute doesn't give an assembly - which is probably the most common path.

src/linker/Linker/DynamicDependency.cs
sbomer: That doesn't type-check because `memberTypes!.Value` isn't a string.

marek-safar: What type check?

sbomer:
@mateoatr could you review the last commit in particular? https://github.com/mono/linker/pull/1215/commits/82eca9c6b0e73623c2420cbf8571520d2843b2c7
I removed parameters to one of the ctors because I don't think we will have line numbers without a source file.

src/linker/Linker/DynamicDependency.cs
sbomer: It needs to call a different ctor for string vs DynamicallyAccessedMemberTypes. `memberSignature ?? memberTypes!.Value` doesn't have a well-defined type - it produces `error CS0019: Operator '??' cannot be applied to operands of type 'string' and 'DynamicallyAccessedMemberTypes'`.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -578,12 +578,11 @@ protected virtual bool ProcessLinkerSpecialAttribute (CustomAttribute ca, ICusto
 		void MarkDynamicDependency (DynamicDependency dynamicDependency, MemberReference context)
 		{
 			Debug.Assert (context is MethodDefinition || context is FieldDefinition);
-			var member = context as IMemberDefinition;
 			AssemblyDefinition assembly;
 			if (dynamicDependency.AssemblyName != null) {
 				assembly = _context.GetLoadedAssembly (dynamicDependency.AssemblyName);
 				if (assembly == null) {
-					_context.LogWarning ($"Unresolved assembly '{dynamicDependency.AssemblyName}' in DynamicDependencyAttribute on '{context}'", 2035, MessageOrigin.TryGetOrigin (member));
+					_context.LogWarning ($"Unresolved assembly '{dynamicDependency.AssemblyName}' in DynamicDependencyAttribute on '{context}'", 2035, MessageOrigin.TryGetOrigin (context as IMemberDefinition));
```
mateoatr, line 585: Can't we get the source line of this with `context.Resolve()`?

src/linker/Linker/DynamicDependency.cs
```diff
@@ -0,0 +1,147 @@
...
+			// Don't honor the Condition until we have figured out the behavior for DynamicDependencyAttribute:
+			// https://github.com/mono/linker/issues/1231
+			// if (!ShouldProcess (context, customAttribute))
+			// 	return null;
+
+			var dynamicDependency = GetDynamicDependency (context, customAttribute);
+			if (dynamicDependency != null)
+				return dynamicDependency;
+
+			context.LogMessage (MessageContainer.CreateWarningMessage (context,
```
mateoatr, line 78: Please use `LogWarning` instead.

PR #1216, opened 2020-05-27
Change XmlFlowAnnotationSource implementation to expose CustomAttributes

tlakollo:
- Change XmlFlowAnnotationSource to expose CustomAttributes
- Use ICustomAttributeProvider to simplify when asking for information
about annotations
- Use of assembly in the attribute is now mandatory
- Assembly name now could be * either in the assembly element or in the
assembly property inside the attribute element, leaving no assembly name is not
supported
- Added new error codes, swaped non used codes for new error codes
- Changed documentation to reflect the way of use


docs/data-formats.md
```diff
@@ -308,21 +308,21 @@ This allows to add a custom attribute to a class, interface, delegate, struct or
     <type fullname="Assembly.A">
       <method signature="System.Void Method1(System.Type)">
         <parameter name="typeParameter">
-          <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
+          <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers" assembly="AssemblyName">
```
vitek-karas, line 311:
```suggestion
          <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute" assembly="System.Runtime">
```

I know we accept the attribute names without the `Attribute` suffix, but for a sample code I would prefer to use the full correct type name (it makes it clear that this is a type name).

Specifically for `DynamicallyAccessedMembersAttribute` we know that this lives in `System.Runtime` assembly (the implementation is in S.P.CoreLib but System.Runtime has a typeforwarder to it and it's the public assembly which apps should refer to).

docs/data-formats.md
```diff
@@ -332,6 +332,24 @@ This allows to add a custom attribute to a class, interface, delegate, struct or
 </linker>
 ```
 
+### DynamicallyAccessedMembers attribute in implicit this parameter
```
vitek-karas, line 335: I would omit this section - this is a very special case which pretty much nobody should need but us - it's really only valid on System.Type. Also in itself this is not related to the XML format - from the XML's point of view this is just an attribute on a method - the "this" semantics is special to the `DynamicallyAccessedMembersAttribute`.

docs/error-codes.md
```diff
@@ -169,4 +169,20 @@ error and warning codes.
 
 #### `IL2028`: Attribute 'attribute' on 'method' doesn't have a required constructor argument.
 
-- The linker found an instance of attribute 'attribute' on 'method' but it lacks a required constructor argument. Linker will ignore this attribute.
\ No newline at end of file
+- The linker found an instance of attribute 'attribute' on 'method' but it lacks a required constructor argument. Linker will ignore this attribute.
+
+#### `IL2029`: There is no xml attribute 'fullname' in xml element '{attributeFullName}'
+
+- The linker found an instance of attribute 'attribute' on 'method' but it lacks a required constructor argument. Linker will ignore this attribute.
```
vitek-karas, line 176: The description doesn't match the title in this case - probably copy/paste error.

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,54 @@
...
+					annotationSources.Add (xmlAnnotations);
+				}
+			}
+			_sources = annotationSources.ToArray ();
+		}
+
+		public IEnumerable<CustomAttribute> GetCustomAttributes (ICustomAttributeProvider provider)
+		{
+			IEnumerable<CustomAttribute> aggregateAttributes = null;
+			foreach (var source in _sources) {
```
vitek-karas, line 32: We should probably optimize for case where there are no XML sources - so wrap the foreach with if _sources.Length > 0.

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Mono.Linker.Dataflow
+{
+	class CustomAttributeSource
+	{
+		private readonly XmlFlowAnnotationSource[] _sources;
+
+		public CustomAttributeSource (LinkContext _context)
+		{
+			Collection<XmlFlowAnnotationSource> annotationSources = new Collection<XmlFlowAnnotationSource> ();
```
vitek-karas, line 18: You can use ArrayBuilder here instead of Collection.

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,54 @@
...
+					aggregateAttributes = aggregateAttributes == null ? source.GetCustomAttributes (provider) : aggregateAttributes.Concat (source.GetCustomAttributes (provider));
+			}
+			if (provider.HasCustomAttributes)
+				aggregateAttributes = aggregateAttributes == null ? provider.CustomAttributes : aggregateAttributes.Concat (provider.CustomAttributes);
+			return aggregateAttributes ?? Enumerable.Empty<CustomAttribute> ();
+		}
+
+		public bool HasCustomAttributes (ICustomAttributeProvider provider)
+		{
+			foreach (var source in _sources) {
```
vitek-karas, line 43: Same as above - optimize for case with no XML sources.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -45,6 +53,25 @@ public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition m
 			return DynamicallyAccessedMemberTypes.None;
 		}
 
+		public DynamicallyAccessedMemberTypes GetMemberTypesForDynamicallyAccessedMemberAttribute (CustomAttributeSource source, ICustomAttributeProvider provider)
```
vitek-karas, line 56: Why is this public?

vitek-karas, line 56: There's no need to pass the `source` as a parameter - it's available as a field on "this".

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -28,76 +27,121 @@ public XmlFlowAnnotationSource (LinkContext context)
...
+							_context.LogMessage (MessageContainer.CreateWarningMessage ($"Could not parse argument '{xmlArguments[i]}' specified in '{_xmlDocumentLocation}' as a {constructor.Parameters[i].ParameterType.FullName}", 2021));
+							recognizedArgument = false;
+						}
+					} else {
+						switch (constructor.Parameters[i].ParameterType.MetadataType) {
+						case MetadataType.String:
+							argumentValue = xmlArguments[i];
+							break;
+						case MetadataType.Int32:
+							argumentValue = int.Parse (xmlArguments[i]);
```
vitek-karas, line 105: This will throw if the value is not an integer. Probably leading to linker failing.
We should call `TryParse` instead and log a warning if it fails.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -28,76 +27,121 @@ public XmlFlowAnnotationSource (LinkContext context)
...
-						return Annotation;
-			}
+			TypeDefinition attributeType;
+			foreach (AssemblyDefinition asm in _context.GetAssemblies ()) {
+				attributeType = asm.FindType (fullName);
+				if (attributeType != null)
+					return attributeType;
 
-			return DynamicallyAccessedMemberTypes.None;
+				foreach (AssemblyDefinition refasm in _context.ResolveReferences (asm)) {
```
vitek-karas, line 129: I would expect all referenced assemblies to be resolved by the time we get here... thus we should not need this loop.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -28,76 +27,121 @@ public XmlFlowAnnotationSource (LinkContext context)
...
-								return result;
-							} else {
-								_context.LogMessage (MessageContainer.CreateWarningMessage ($"Could not parse argument '{argument}' specified in '{_xmlDocumentLocation}' as a DynamicallyAccessedMemberTypes", 2021));
-							}
-						}
-					} else {
-						_context.LogMessage (MessageContainer.CreateWarningMessage ($"DynamicallyAccessedMembers attribute was specified but there is more than one argument", 2022));
-					}
-				}
+			ArrayBuilder<string> childs = new ArrayBuilder<string> ();
```
vitek-karas, line 140:
```suggestion
			ArrayBuilder<string> children = new ArrayBuilder<string> ();
```

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -120,44 +164,31 @@ public void ParseXml (string document)
...
 			while (iterator.MoveNext ()) {
-				if (!ShouldProcessElement (iterator.Current)) {
+				if (!ShouldProcessElement (iterator.Current))
 					return;
-				}
 
-				AssemblyDefinition assembly = GetAssembly (context, GetAssemblyName (iterator.Current));
+				if (GetFullName (iterator.Current) == "*") {
+					foreach (AssemblyDefinition assemblyIterator in context.GetAssemblies ()) {
+						ProcessTypes (assemblyIterator, iterator);
```
vitek-karas, line 172: If I read this correctly:
* Let's say I specify to add an attribute `LinkerRemoveAttribute` to `NullableAttribute` on `assembly="*"`.
* The code will do ProcessTypes on all assemblies
* If one of those assemblies doesn't have `NullableAttribute` type, this will issue a warning

I think that for the `assembly="*"` case we should intentionally ignore if the specified types are missing. Otherwise there will be lot of warnings with no go way to get rid of them. Once we do find the type I think we should warn about everything else on the type, but the lack of the type itself should not cause a warning in the "All-assemblies" case.

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,57 @@
...
+using Mono.Cecil;
+using System.Linq;
+
+namespace Mono.Linker.Dataflow
+{
+	class CustomAttributeSource
+	{
+		private readonly XmlFlowAnnotationSource[] _sources;
+
+		public CustomAttributeSource (LinkContext _context)
```
marek-safar, line 15: Don't use `_` for parameter names

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,57 @@
...
+		{
+			ArrayBuilder<XmlFlowAnnotationSource> annotationSources = new ArrayBuilder<XmlFlowAnnotationSource> ();
+			if (_context.AttributeDefinitions != null && _context.AttributeDefinitions.Count > 0) {
```
marek-safar, line 18:
```suggestion
			if (_context.AttributeDefinitions?.Count > 0) {
```

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,57 @@
...
+			_sources = annotationSources.ToArray ();
+		}
+
+		public IEnumerable<CustomAttribute> GetCustomAttributes (ICustomAttributeProvider provider)
+		{
+			IEnumerable<CustomAttribute> aggregateAttributes = null;
+			if (_sources != null) {
+				foreach (var source in _sources) {
+					if (source.HasCustomAttributes (provider))
+						aggregateAttributes = aggregateAttributes == null ? source.GetCustomAttributes (provider) : aggregateAttributes.Concat (source.GetCustomAttributes (provider));
```
marek-safar, line 34: Use `yield` instead

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,57 @@
...
+				}
+			}
+			if (provider.HasCustomAttributes)
+				aggregateAttributes = aggregateAttributes == null ? provider.CustomAttributes : aggregateAttributes.Concat (provider.CustomAttributes);
```
marek-safar, line 38: Use `yield` instead

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,57 @@
...
+			return aggregateAttributes ?? Enumerable.Empty<CustomAttribute> ();
```
marek-safar, line 39: This can be removed if you make the method iterator

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,57 @@
...
+		{
+			ArrayBuilder<XmlFlowAnnotationSource> annotationSources = new ArrayBuilder<XmlFlowAnnotationSource> ();
+			if (_context.AttributeDefinitions != null && _context.AttributeDefinitions.Count > 0) {
+				foreach (string a in _context.AttributeDefinitions) {
+					XmlFlowAnnotationSource xmlAnnotations = new XmlFlowAnnotationSource (_context);
+					xmlAnnotations.ParseXml (a);
+					annotationSources.Add (xmlAnnotations);
+				}
+			}
+			_sources = annotationSources.ToArray ();
```
marek-safar, line 25: Why does it have to array and not list if it's not exposed?

src/linker/Linker.Dataflow/CustomAttributeSource.cs
```diff
@@ -0,0 +1,57 @@
...
+		public bool HasCustomAttributes (ICustomAttributeProvider provider)
+		{
+			if (_sources != null) {
+				foreach (var source in _sources) {
+					if (source.HasCustomAttributes (provider)) {
+						return true;
+					}
+				}
+			}
+			if (provider.HasCustomAttributes) {
```
marek-safar, line 51: `return provider.HasCustomAttributes;`

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -45,6 +53,25 @@ public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition m
 			return DynamicallyAccessedMemberTypes.None;
 		}
 
+		private DynamicallyAccessedMemberTypes GetMemberTypesForDynamicallyAccessedMemberAttribute (ICustomAttributeProvider provider)
+		{
+			if (_source.HasCustomAttributes (provider)) {
```
marek-safar, line 58: Use early exit

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -45,6 +53,25 @@ public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition m
...
+				foreach (var attribute in _source.GetCustomAttributes (provider)) {
+					if (IsDynamicallyAccessedMembersAttribute (attribute)) {
```
marek-safar, line 60:
```
if (!IsDynamicallyAccessedMembersAttribute (attribute))
	continue
```

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -45,6 +53,25 @@ public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition m
...
+						if (attribute.ConstructorArguments.Count == 0) {
+							_context.LogMessage (MessageContainer.CreateWarningMessage ($"DynamicallyAccessedMembers attribute was specified but no argument was proportioned", 2020));
+						} else if (attribute.ConstructorArguments.Count == 1) {
+							var arguments = attribute.ConstructorArguments.ToArray ();
+							return (DynamicallyAccessedMemberTypes) arguments[0].Value;
```
marek-safar, line 65: You cannot cast external value to local type

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -45,6 +53,25 @@ public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition m
...
 
+		private DynamicallyAccessedMemberTypes GetMemberTypesForDynamicallyAccessedMemberAttribute (ICustomAttributeProvider provider)
+		{
+			if (_source.HasCustomAttributes (provider)) {
+				foreach (var attribute in _source.GetCustomAttributes (provider)) {
+					if (IsDynamicallyAccessedMembersAttribute (attribute)) {
+						if (attribute.ConstructorArguments.Count == 0) {
+							_context.LogMessage (MessageContainer.CreateWarningMessage ($"DynamicallyAccessedMembers attribute was specified but no argument was proportioned", 2020));
+						} else if (attribute.ConstructorArguments.Count == 1) {
+							var arguments = attribute.ConstructorArguments.ToArray ();
```
marek-safar, line 64: Why is ToArray needed?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -28,76 +26,109 @@ public XmlFlowAnnotationSource (LinkContext context)
...
+				}
+				string concatAttribute;
+				concatAttribute = attributeFullName.EndsWith ("Attribute") ? attributeFullName : String.Concat (attributeFullName, "Attribute");
+				string assemblyName = GetAttribute (iterator.Current, "assembly");
+				if (assemblyName == String.Empty) {
+					_context.LogMessage (MessageContainer.CreateWarningMessage ($"Attribute element doest not contain attribute 'assembly' in xml element '{attributeFullName}'", 2030));
+					continue;
+				} else if (assemblyName == "*")
+					attributeType = _context.GetType (concatAttribute);
+				else {
```
marek-safar, line 60: I think it'd be better to recognize missing assembly attribute as "any assembly"

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -28,76 +26,109 @@ public XmlFlowAnnotationSource (LinkContext context)
...
-		{
-			return _methods.TryGetValue (method, out var ann) ? ann.ReturnAnnotation : DynamicallyAccessedMemberTypes.None;
-		}
+				string attributeFullName = GetFullName (iterator.Current);
+				if (attributeFullName == String.Empty) {
+					_context.LogMessage (MessageContainer.CreateWarningMessage ($"Attribute element does not contain attribute 'fullname'", 2029));
+					continue;
+				}
+				string concatAttribute;
+				concatAttribute = attributeFullName.EndsWith ("Attribute") ? attributeFullName : String.Concat (attributeFullName, "Attribute");
```
marek-safar, line 53: Why is the rationale behind this logic?

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
```diff
@@ -28,76 +26,109 @@ public XmlFlowAnnotationSource (LinkContext context)
...
 
-		public DynamicallyAccessedMemberTypes GetThisParameterAnnotation (MethodDefinition method)
-		{
-			if (_methods.TryGetValue (method, out var ann) && ann.ParameterAnnotations != null) {
-				foreach (var (ParamName, Annotation) in ann.ParameterAnnotations)
-					if (ParamName == "this")
-						return Annotation;
-			}
+				ArrayBuilder<string> arguments = GetAttributeChildren (iterator.Current.SelectChildren ("argument", string.Empty));
+				MethodDefinition constructor = attributeType.Methods.Where (method => method.IsConstructor).FirstOrDefault (c => c.Parameters.Count == arguments.Count);
```
marek-safar, line 75: There is already `IsInstanceConstructor` logic

src/linker/Linker.Dataflow/FlowAnnotations.cs
vitek-karas: Same comment as in the previous PR - this is an enum type, this is basically the only case where it's "safe" since it's almost like a primitive type.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
vitek-karas: @marek-safar can you please explain in a bit more detail - I don't get what you mean by this comment.

src/linker/Linker.Dataflow/FlowAnnotations.cs
marek-safar: I don't think that's the case. If someone declares it's own type (we matching on names) then you cannot cast the enum type to "runtime" version of the type as they have different assembly identity.

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
marek-safar: The change requires to always specify assembly, for example as

```xml
        <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers" assembly="*">
          <argument>DefaultConstructor</argument>
        </attribute>
```

what I'm suggesting is instead of special '*' character use following as equivalent 

```xml
        <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembers">
          <argument>DefaultConstructor</argument>
        </attribute>
```

Either way, this needs to be documented

src/linker/Linker.Dataflow/FlowAnnotations.cs
vitek-karas: You're right - we have to go through integer:
```C#
return (DynamicallyAccessedMemberTypes)(int)arguments[0].Value;
```

Then it should be correct - as we will simply map the number to the linker understanding of the flags.

src/linker/Linker.Dataflow/CustomAttributeSource.cs
tlakollo: Now is using a List

src/linker/Linker.Dataflow/FlowAnnotations.cs
tlakollo: I was not needed, I deleted it

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: I change the behavior and document how is used on the attribute specifics

src/linker/Linker.Dataflow/XmlFlowAnnotationSource.cs
tlakollo: I was thinking about if someone would use the abbreviate names, in those cases append the Attribute at the end. But yeah, it may cause some confusion. Deleted it and now the user will need to write the actual attribute name.

vitek-karas:
I rebased to master and fixed any issues after that (mostly warning creation requiring context and location).
I moved the custom attribute source files to main linker namespace as they don't really belong to DataFlow anymore.
Small formatting fixes.

src/linker/Linker/XmlCustomAttributeSource.cs
```diff
@@ -362,66 +391,54 @@ void ProcessMethodSignature (TypeDefinition type, string signature, XPathNodeIte
...
+			iterator = iterator.Current.SelectChildren ("parameter", string.Empty);
 			while (iterator.MoveNext ()) {
-				methodParameters.Add ((GetAttribute (iterator.Current, "name"), ProcessAttributes (iterator)));
+				IEnumerable<CustomAttribute> attributes = ProcessAttributes (iterator.Current);
+				if (attributes.Count () > 0) {
+					string paramName = GetAttribute (iterator.Current, "name");
+					foreach (ParameterDefinition parameter in method.Parameters) {
+						if (paramName == parameter.Name) {
+							if (_attributes.ContainsKey (parameter))
+								_context.LogMessage (MessageContainer.CreateWarningMessage (_context, $"There are duplicate parameter names for '{paramName}' inside '{method.Name}' in '{_xmlDocumentLocation}'", 2024, _xmlDocumentLocation));
```
marek-safar, line 419: Why is it warning?

src/linker/Linker/XmlCustomAttributeSource.cs
```diff
@@ -362,66 +391,54 @@ void ProcessMethodSignature (TypeDefinition type, string signature, XPathNodeIte
...
+			bool firstAppearance = true;
 			while (iterator.MoveNext ()) {
-				methodParameters.Add (ProcessAttributes (iterator));
+				if (firstAppearance) {
+					firstAppearance = false;
+					IEnumerable<CustomAttribute> attributes = ProcessAttributes (iterator.Current);
+					if (attributes.Count () > 0)
+						_attributes[method.MethodReturnType] = attributes;
+				} else {
+					_context.LogMessage (MessageContainer.CreateWarningMessage (_context, $"There is more than one return parameter specified for '{method.Name}' in '{_xmlDocumentLocation}'", 2023, _xmlDocumentLocation));
```
marek-safar, line 439: Why is it warning?

src/linker/Linker/XmlCustomAttributeSource.cs
```diff
@@ -121,48 +142,34 @@ public void ParseXml (string document)
...
+		void ProcessAssemblies (LinkContext context, XPathNodeIterator iterator)
 		{
 			while (iterator.MoveNext ()) {
-				if (!ShouldProcessElement (iterator.Current)) {
+				if (!ShouldProcessElement (iterator.Current))
 					return;
-				}
 
-				AssemblyDefinition assembly = GetAssembly (context, GetAssemblyName (iterator.Current));
+				if (GetFullName (iterator.Current) == "*") {
```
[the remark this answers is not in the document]
vitek-karas: I added the docs for this

src/linker/Linker/XmlCustomAttributeSource.cs
vitek-karas: We found two values for the same thing - we'll pick one of them, but drop the other. I think this should be a warning (since we're warning about other issues in the XML already)

src/linker/Linker/XmlCustomAttributeSource.cs
vitek-karas: Same as above - two values for the same thing - we will intentionally drop data in this case so we should warn about it.

src/linker/Linker/XmlCustomAttributeSource.cs
marek-safar: What if we instead of warning we do something like `_attributes[parameter].Add(attributes);`

PR #1219, opened 2020-05-28
Allow types and other metadata item as source for reflection access pattern

vitek-karas:
With the upcoming support for DynamicallyAccessedMembersAttribute on generic parameters, the source of the access pattern will no longer be any code, but it can be a type declaration or field definition and so on. In such cases trying to point back to some line of code is complicated and in lot of cases confusing.

This change allows more values to be passed as the source of a reflection access pattern. Currently there's no use case for it. The one which comes close is attribute fields, but those make sense to use the attribute ctor as the source of the problem and use the field as the "accessed" item.

Corresponding changes to the test infra to allow for wider range of values and to handle missing values better.

### DIFF
--- a/src/linker/Linker/LoggingReflectionPatternRecorder.cs
+++ b/src/linker/Linker/LoggingReflectionPatternRecorder.cs
@@ -37,27 +37,26 @@ namespace Mono.Linker
 			_context = context;
 		}
 
-		public void RecognizedReflectionAccessPattern (IMetadataTokenProvider source, Instruction sourceInstruction, IMetadataTokenProvider accessedItem)
+		public void RecognizedReflectionAccessPattern (IMemberDefinition source, Instruction sourceInstruction, IMetadataTokenProvider accessedItem)
 		{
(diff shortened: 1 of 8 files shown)


src/linker/Linker/LoggingReflectionPatternRecorder.cs
```diff
@@ -37,16 +37,29 @@ public LoggingReflectionPatternRecorder (LinkContext context)
...
+				origin = new MessageOrigin (source);
+
+			if (origin.FileName == null) {
+				if (method != null)
+					location = method.DeclaringType.FullName + "::" + GetSignature (method) + ": ";
+				else
+					location = source.DeclaringType?.FullName + "::" + source.Name;
+			}
+
+			_context.LogMessage (MessageContainer.CreateWarningMessage (_context, location + message, 2006, origin, "Unrecognized reflection pattern"));
```
marek-safar, line 62: I don't think we should prefix the location to the message but fix MessageOrigin to handle this case instead.

This could be done as separate PR

PR #1223, opened 2020-05-29
Factor out binding logic for DynamicallyAccessedMemberTypes

sbomer:
This will be used for DynamicDependencyAttribute as well.

In https://github.com/mono/linker/pull/1215#discussion_r431685282 there were concerns about the performance of `HasFlag`, potentially boxing on Unity for example. Let's continue the discussion here.

### DIFF
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
@@ -1,0 +1,297 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+using Mono.Cecil;
+
+namespace Mono.Linker.Dataflow
+{
+	internal static class DynamicallyAccessedMembersBinder
+	{
+		// Returns the members of the type bound by memberTypes. For MemberTypes.All, this returts a single member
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,298 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
+using Mono.Cecil;
+
+namespace Mono.Linker.Dataflow
```
vitek-karas, line 7: Nit: Maybe we should put this into `Mono.Linker` namespace as it's not DataFlow-only anymore... but I'm fine either way

marek-safar: Yep, this should go to something like `Mono.Linker.TypeDefinitionLookup` class and possibly changed to extension methods

sbomer: Changed to `Mono.Linker` with extension methods. I kept the class name because this is specific to the `DynamicallyAccessedMemberTypes`, and doesn't have other member lookup logic.

PR #1224, opened 2020-05-29
Simplify logging methods

mateoatr:
This PR attempts to simplify a little bit the way in which we log warnings, errors and messages by adding a few methods to `LinkContext`, thus removing the need to manually create a `MessageContainer` every time we want to log a message.


src/linker/Linker/LinkContext.cs
```diff
@@ -468,10 +468,84 @@ public bool IsOptimizationEnabled (CodeOptimizations optimization, AssemblyDefin
 
 		public void LogMessage (MessageContainer message)
 		{
-			if (LogMessages && message != MessageContainer.Empty)
+			if (LogMessages)
 				Logger?.LogMessage (message);
 		}
 
+		public void LogMessage (string message, bool isDiagnostic = false)
```
marek-safar, line 475: I'd go with `LogDiagnostic` instead of extra argument which you have to type all the time to know what it's for

src/linker/Linker/LinkContext.cs
```diff
@@ -468,10 +468,84 @@ public bool IsOptimizationEnabled (CodeOptimizations optimization, AssemblyDefin
 
 		public void LogMessage (MessageContainer message)
 		{
-			if (LogMessages && message != MessageContainer.Empty)
```
vitek-karas, line 471: Why is it OK to not consider the `Empty` message anymore? I thought we used that for warning suppression, so how does it work without this?

src/linker/Linker/LinkContext.cs
```diff
@@ -468,10 +468,84 @@ public bool IsOptimizationEnabled (CodeOptimizations optimization, AssemblyDefin
...
+		/// <param name="code">Unique warning ID. Please see https://github.com/mono/linker/blob/master/doc/error-codes.md for the list of warnings and possibly add a new one</param>
+		/// <param name="origin">Filename or member where the warning is coming from</param>
+		/// <param name="subcategory">Optionally, further categorize this warning</param>
+		public void LogWarning (string text, int code, MessageOrigin origin, string subcategory = MessageSubCategory.None)
+		{
+			if (!LogMessages)
+				return;
+
+			var warning = MessageContainer.CreateWarningMessage (this, text, code, origin, subcategory);
+			Logger?.LogMessage (warning);
```
vitek-karas, line 503: Similarly here - how is warning suppression handled?

test/Mono.Linker.Tests/TestCasesRunner/LinkerTestLogger.cs
```diff
@@ -13,6 +13,9 @@ public struct MessageRecord
 
 		public void LogMessage (MessageContainer msBuildMessage)
 		{
+			if (msBuildMessage == MessageContainer.Empty)
```
vitek-karas, line 16: This only fixes the warning suppression for tests - but ConsoleLogger will still try to print out empty warnings. I think warning suppression should be handled outside of the loggers as it's a linker feature.

src/linker/Linker/LinkContext.cs
```diff
@@ -472,6 +472,82 @@ public void LogMessage (MessageContainer message)
...
+		/// <param name="subcategory">Optionally, further categorize this error</param>
+		/// <param name="origin">Filename, line, and column where the error was found</param>
+		/// <returns>New MessageContainer of 'Error' category</returns>
+		public void LogError (string text, int code, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null)
+		{
+			if (!LogMessages)
+				return;
+
+			var error = MessageContainer.CreateErrorMessage (text, code, subcategory, origin);
+			Logger?.LogMessage (error);
```
marek-safar, line 548: nit: call LogMessage as everywhere else

PR #1225, opened 2020-05-29
Fix edge case in signature parser

sbomer:
Because generic methods don't have mangled names, a generic method
signature could match a type if we don't check its arity.

### DIFF
--- a/test/Mono.Linker.Tests/Tests/DocumentationSignatureParserTests.cs
+++ b/test/Mono.Linker.Tests/Tests/DocumentationSignatureParserTests.cs
@@ -576,6 +576,14 @@ namespace Mono.Linker.Tests
 			{
 			}
 
+			public class GenericType<T, U>
+			{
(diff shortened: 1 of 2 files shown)


test/Mono.Linker.Tests/Tests/DocumentationSignatureParserTests.cs
```diff
@@ -571,6 +571,11 @@ public void NoGenericParameterType (Generic<A> g)
 			{
 			}
 
+			[ExpectUnresolvedDocumentationSignature ("T:Mono.Linker.Tests.DocumentationSignatureParserTests.Invalid.TypeWithMethodGenericParameters``1")]
```
marek-safar, line 574: Could you add test for nested generic and non-generic types inside generic types as well?

sbomer: Generic types don't hit this edge case because they already have mangled names that will not match an unmangled method name. I will add the same test for a non-generic nested in a generic type.

Note that there are already nested generic tests for the happy case: https://github.com/mono/linker/blob/master/test/Mono.Linker.Tests/Tests/DocumentationSignatureParserTests.cs#L414

PR #1228, opened 2020-06-01
Repurpose unused dependency tracing to track marking reasons

marek-safar:
- Convert DepdencyKind to MarkingInfo and include more cases for tracking
- Include marking info data in the trace output
- Include all edges in the output data and not only first hits
- Add documentation for the command line options
- Add documentation for the data format


docs/data-formats.md
```diff
@@ -419,3 +431,85 @@ For fields and properties, you need to include the name since they are not order
...
+          </xs:complexType>
+        </xs:element>
+      </xs:sequence>
+      <xs:attribute type="xs:float" name="version"/>
+    </xs:complexType>
+  </xs:element>
+</xs:schema>
+```
+
+Every edge has `b` and `e` attribute which represent vertices of the directed graph. They can have
```
vitek-karas, line 469: Just curious - the names `b`, `e` and `c` don't make sense to me - maybe explain what they stand for. For example I would choose `s` = "source", `t` = "target", `r` = "reason".

docs/data-formats.md
```diff
@@ -419,3 +431,85 @@ For fields and properties, you need to include the name since they are not order
...
+{VT}:{info}
+```
+
+VT is one of following values 2 letters acronyms describing `info` format
+
+| VT  | Vertex Type  | Info Format
+|---|---|---|
+| AS  | Assembly  | fully-qualified-name |
+| CA  | Custom Attribute  | attribute-type |
+| EV  | Event | declaring-type:property-name |
```
vitek-karas, line 492: This should probably be `declaring-type::property-name` (double `:`). Also for the ones below.

docs/data-formats.md
```diff
@@ -419,3 +431,85 @@ For fields and properties, you need to include the name since they are not order
...
+| OT  | Other | any-string-value |
+
+### Data Example
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<dependencies version="1.3">
+  <edge b="FI:System.Exception::_innerException" ba="System.Private.CoreLib" e="TD:System.Exception" c="FieldType" />
+  <edge b="ME:System.Runtime.CompilerServices.NullableAttribute::.ctor(System.Byte)" ba="System.Private.CoreLib" e="TD:System.Byte" c="ParameterType" />
+  <edge b="AS:sample, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null" e="ME:AppDelegate::Main(System.String[])" ea="sample" c="RootAssembly" />
```
vitek-karas, line 513: Nit: I think this example is slightly wrong - Assembly -> EntryPoint reason should be something like `AssemblyAction`

src/linker/Linker/MarkingReason.cs
```diff
@@ -0,0 +1,167 @@
...
+	//
+	// Try to keep the name short and self explamantory at the same time. The name
+	// is stored in the output format as text and be better easily understood.
+	//
+	public enum MarkingReason
+	{
+		//
+		// Entry points to the analysis
+		//
+		AssemblyAction = 1,
```
vitek-karas, line 16: Can we keep the comments from the original enum? I like that those include what is the edge the reason is used on - it makes it easier to understand the reason.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -410,10 +310,10 @@ void ProcessMarkedTypesWithInterfaces ()
 			foreach (var type in typesWithInterfaces) {
 				// Exception, types that have not been flagged as instantiated yet.  These types may not need their interfaces even if the
 				// interface type is marked
-				if (!Annotations.IsInstantiated (type))
+				if (!Annotations.IsInstantiated (type.Item1))
 					continue;
 
-				MarkInterfaceImplementations (type);
+				MarkInterfaceImplementations (type.Item1, type.Item2);
```
vitek-karas, line 316: Can we please name the values in the tuple?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -844,9 +745,7 @@ protected virtual void MarkSecurityAttribute (SecurityAttribute sa, in Dependenc
 				return;
 			}
 
-			// Security attributes participate in inference logic without being marked.
-			Tracer.AddDirectDependency (sa, reason, marked: false);
```
vitek-karas, line 848: This will introduce a "hole" in the graph, right?
That is we will see that the type of the attribute is marked due to the security attribute, but there won't be any edge describing what marked the security attribute. I find this weird.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2103,90 +1959,73 @@ protected virtual MethodDefinition MarkMethod (MethodReference reference, Depend
...
 		}
 
-		protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo reason)
+		protected virtual void ProcessMethod (MethodDefinition method, in MarkingInfo reason)
 		{
-#if DEBUG
-			if (!_methodReasons.Contains (reason.Kind))
-				throw new ArgumentOutOfRangeException ($"Internal error: unsupported method dependency {reason.Kind}");
-#endif
+			/*
```
vitek-karas, line 1981: I assume that the "comment out" here is to be removed... ?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2423,26 +2269,23 @@ protected virtual bool ShouldParseMethodBody (MethodDefinition method)
 			}
 		}
 
-		internal protected void MarkProperty (PropertyDefinition prop, in DependencyInfo reason)
+		internal protected void MarkProperty (PropertyDefinition prop)
```
vitek-karas, line 2272: Similar comment as above - won't this introduce a "hole" in the graph? That is we will mark the custom attribute on the property as "reason - property", but there won't be any record of why the property itself was marked...

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1128,16 +1034,10 @@ bool ProcessLateMarkedAttributes ()
...
 		{
 			if (reference.DeclaringType is GenericInstanceType) {
-				Debug.Assert (reason.Kind == DependencyKind.FieldAccess || reason.Kind == DependencyKind.Ldtoken);
-				// Blame the field reference (without actually marking) on the original reason.
-				Tracer.AddDirectDependency (reference, reason, marked: false);
-				MarkType (reference.DeclaringType, new DependencyInfo (DependencyKind.DeclaringType, reference));
-
-				// Blame the field definition that we will resolve on the field reference.
-				reason = new DependencyInfo (DependencyKind.FieldOnGenericInstance, reference);
+				MarkType (reference.DeclaringType, new MarkingInfo (MarkingReason.GenericInstance, reference));
```
sbomer, line 1040:
```csharp
public class G<T> {
  public T field;
}

public class Foo {}

public class C {
  public void M() {
    var g = new G<Foo> ();
    _ = g.field;
  }
}
```
Before the changes here and in GetOriginalType, this creates a dependency graph like:

- `M`
  - FieldAccess -> `G<Foo>.field`
    - DeclaringType (here) -> `G<Foo>`
      - ElementType (in GetOriginalType) -> `G<T>`
      - GenericArgumentType -> `Foo`
    - FieldOnGenericInstance -> `G<T>.field`
      - DeclaringType (in MarkField) -> `G<T>`

With the new logic, it looks like this:

- `M`
  - FieldAccessInstr -> `G<T>.field`
    - DeclaringType (in MarkField) -> `G<T>`
- `G<Foo>.field`
  - GenericInstance -> `G<T>`
- `G<Foo>`
  - GenericArgument -> `Foo`

which includes `G<Foo>.field` or `G<Foo>` in the dependency graph without saying why, and also seems to record `G<T>` as a generic instance of `G<Foo>.field` which doesn't make sense to me. What is the reason for the change?

Issue #1179, opened 2020-05-09
Embedded xml could gen non suppressible/fixable warnings


tlakollo:
Remove ResolveStep (#1169) changes AddUnresolveMarker calls for a warning message, this could lead to warnings that cannot be deleted in the case an XML file comes embedded in the assembly. 
Example:
You have an application that includes a nugget package, no source code for it. and contains an embedded XML on it with a bug/mismatch and talks about an unexisting member. If it warns there would be no way to get rid of the warning.
XmlAnnotation (#1176) mimics this same behavior

vitek-karas:
/cc @mateoatr 

@marek-safar Would you know if it's important to have a solution for this?
I'm sort of split on this:
* On the one hand if we report these things as just informational messages it's likely that almost nobody will notice them. So having them show up as warnings is in my mind a good thing.
* On the other hand if my app uses a NuGet which happens to have a wrong embedded XML in it, there's no way for me to suppress the warning - which is bad.

Do we have experience with this in Xamarin?

marek-safar:
I think we should have a solution for it as the embedded XML are in most cases the only solution how to reliably ship the data. We don't issue any warnings in Xamarin today (errors only) but I think this can happen on existing code.

The easy solution would be to use or extend [Scope](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute.scope?view=netcore-3.1#remarks) property which we'll have to support most likely anyway.

vitek-karas:
Just to make sure I understand: By "solution" you mean to have a way to suppress warnings coming from parsing embedded XML files, right?

As you mention we should support `Scope`, unfortunately we would probably have to extend it to some degree - I don't know if it can be used to suppress warnings in a different module/assembly. But if yes - that would be the right way to do so.

marek-safar:
Yes, `Scope = "resource"` should suppress warning related to embedded data because we would use the "resource" as the source location, right?

We could extend the name to make it more obvious or even support embedded resource name as the scope.

vitek-karas:
I missed the `resource` value, you're right. Now if only this works cross assembly (attribute in one assembly targeting a different assembly). If not we would have to rely on the XML attribute format - would work, but still.

marek-safar:
I think we issue most XML warnings at load time so that should work fine.

PR #1232, opened 2020-06-02
Add more warning suppression scenarios

mateoatr:
Continuing with the work of https://github.com/mono/linker/pull/1185 this PR adds the ability to suppress warnings throughout a module or assembly. It also adds support for using the `Target` and `Scope` properties for global suppression on types and members.

### DIFF
--- a/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
+++ b/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
@@ -1,4 +1,6 @@
 ﻿using System.Collections.Generic;
+using System.Linq;
+using System.Transactions;
 using Mono.Cecil;
 
 namespace Mono.Linker
@@ -7,6 +9,7 @@ namespace Mono.Linker
 	{
 		private readonly LinkContext _context;
 		private readonly Dictionary<ICustomAttributeProvider, Dictionary<int, SuppressMessageInfo>> _localSuppressions;
+		private GlobalSuppressions _globalSuppressions;
 
 		private bool HasLocalSuppressions {
 			get {
@@ -41,8 +44,15 @@ namespace Mono.Linker
 
 		public bool IsSuppressed (int id, MessageOrigin warningOrigin, out SuppressMessageInfo info)
 		{
+			if (IsGloballySuppressed (id, warningOrigin, out info))
+				return true;
+
 			if (HasLocalSuppressions && warningOrigin.MemberDefinition != null) {
 				IMemberDefinition memberDefinition = warningOrigin.MemberDefinition;
+				// Check if it's suppressed by a module level suppression.
+				if (IsLocallySuppressed (id, memberDefinition.DeclaringType.Module, out info))
+					return true;
+
 				while (memberDefinition != null) {
 					if (IsLocallySuppressed (id, memberDefinition, out info))
 						return true;
@@ -103,6 +113,71 @@ namespace Mono.Linker
 			info = default;
 			return _localSuppressions.TryGetValue (provider, out suppressions) &&
 				suppressions.TryGetValue (id, out info);
+		}
+
+		private bool IsGloballySuppressed (int id, MessageOrigin origin, out SuppressMessageInfo info)
+		{
+			DecodeGlobalSuppressMessageAttributes ();
+			// TODO
+			info = default;
+			return false;
+		}
+
+		private void DecodeGlobalSuppressMessageAttributes ()
+		{
+			if (_globalSuppressions == null) {
+				var suppressions = new GlobalSuppressions ();
+				HashSet<ModuleDefinition> modules = new HashSet<ModuleDefinition> ();
+				foreach (var assembly in _context.Resolver.AssemblyCache.Values) {
+					DecodeGlobalSuppressMessageAttributes (assembly, suppressions);
+					modules.Add (assembly.MainModule);
+				}
+
+				foreach (var module in modules)
+					DecodeGlobalSuppressMessageAttributes (module, suppressions);
+
+				_globalSuppressions = suppressions;
+			}
+		}
+
+		private void DecodeGlobalSuppressMessageAttributes (ICustomAttributeProvider provider, GlobalSuppressions globalSuppressions)
+		{
+			var attributes = provider.CustomAttributes.Where (a => a.AttributeType.Name == "UnconditionalSuppressMessageAttribute");
+			foreach (var instance in attributes) {
+				SuppressMessageInfo info;
+				if (!TryDecodeSuppressMessageAttributeData (instance, out info))
+					continue;
+
+				if (info.Target == null && (info.Scope == "module" || info.Scope == null)) {
+					if (provider is ModuleDefinition)
+						AddLocalSuppression (instance, provider);
+					else if (provider is AssemblyDefinition assembly)
+						AddLocalSuppression (instance, assembly.MainModule);
+
+					continue;
+				}
+
+				switch (info.Scope) {
+				case "module":
+				case "type":
+				case "method":
+				case "member":
+					// Integrate Sven's work here
+					break;
+				case "resource":
+					// TODO
+					break;
+				case "namespace":
+				case "namespaceanddescendants":
+					// TODO
+					break;
+				}
+			}
+		}
+
+		private class GlobalSuppressions
+		{
+
 		}
 	}
 }
--- a/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
+++ b/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
@@ -1,5 +1,4 @@
 ﻿using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using Mono.Cecil;
 
@@ -8,26 +7,30 @@ namespace Mono.Linker
 	public class UnconditionalSuppressMessageAttributeState
 	{
 		private readonly LinkContext _context;
-		private readonly Dictionary<ICustomAttributeProvider, Dictionary<int, SuppressMessageInfo>> _localSuppressions;
-		private readonly GlobalSuppressions _globalSuppressions;
+		private readonly Dictionary<ICustomAttributeProvider, Dictionary<int, SuppressMessageInfo>> _suppressions;
+		private HashSet<AssemblyDefinition> InitializedAssemblies { get; }
 
 		public UnconditionalSuppressMessageAttributeState (LinkContext context)
 		{
 			_context = context;
-			_localSuppressions = new Dictionary<ICustomAttributeProvider, Dictionary<int, SuppressMessageInfo>> ();
-			_globalSuppressions = new GlobalSuppressions ();
+			_suppressions = new Dictionary<ICustomAttributeProvider, Dictionary<int, SuppressMessageInfo>> ();
+			InitializedAssemblies = new HashSet<AssemblyDefinition> ();
 		}
 
-		public void AddLocalSuppression (CustomAttribute ca, ICustomAttributeProvider provider)
+		public void AddSuppression (CustomAttribute ca, ICustomAttributeProvider provider)
 		{
 			SuppressMessageInfo info;
-			if (!TryDecodeSuppressMessageAttributeData (ca, out info)) {
+			if (!TryDecodeSuppressMessageAttributeData (ca, out info))
 				return;
-			}
 
-			if (!_localSuppressions.TryGetValue (provider, out var suppressions)) {
+			AddSuppression (info, provider);
+		}
+
+		private void AddSuppression (SuppressMessageInfo info, ICustomAttributeProvider provider)
+		{
+			if (!_suppressions.TryGetValue (provider, out var suppressions)) {
 				suppressions = new Dictionary<int, SuppressMessageInfo> ();
-				_localSuppressions.Add (provider, suppressions);
+				_suppressions.Add (provider, suppressions);
 			}
 
 			if (suppressions.ContainsKey (info.Id))
@@ -43,23 +46,32 @@ namespace Mono.Linker
 			if (warningOrigin.MemberDefinition == null)
 				return false;
 
-			IMemberDefinition memberDefinition = warningOrigin.MemberDefinition;
-			while (memberDefinition != null) {
-				if (IsLocallySuppressed (id, memberDefinition, out info) ||
-					IsGloballySuppressed (id, memberDefinition, out info))
+			IMemberDefinition elementContainingWarning = warningOrigin.MemberDefinition;
+			ModuleDefinition module = GetModuleFromProvider (warningOrigin.MemberDefinition);
+			DecodeModuleLevelAndGlobalSuppressMessageAttributes (module);
+			while (elementContainingWarning != null) {
+				if (IsSuppressed (id, elementContainingWarning, out info))
 					return true;
 
-				memberDefinition = memberDefinition.DeclaringType;
+				elementContainingWarning = elementContainingWarning.DeclaringType;
 			}
 
 			// Check if there's an assembly or module level suppression.
-			memberDefinition = warningOrigin.MemberDefinition;
-			if ((memberDefinition is TypeDefinition type && IsLocallySuppressed (id, type?.Module, out info)) ||
-					IsLocallySuppressed (id, memberDefinition.DeclaringType?.Module, out info) ||
-					IsLocallySuppressed (id, memberDefinition.DeclaringType?.Module?.Assembly, out info))
+			if (IsSuppressed (id, module, out info) ||
+				IsSuppressed (id, module.Assembly, out info))
 				return true;
 
 			return false;
+		}
+
+		private bool IsSuppressed (int id, ICustomAttributeProvider provider, out SuppressMessageInfo info)
+		{
+			info = default;
+			if (provider == null)
+				return false;
+
+			return (_suppressions.TryGetValue (provider, out var suppressions) &&
+				suppressions.TryGetValue (id, out info));
 		}
 
 		private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attribute, out SuppressMessageInfo info)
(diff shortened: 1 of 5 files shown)


src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -1,4 +1,7 @@
 ﻿using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Transactions;
```
vitek-karas, line 4: Nit: I doubt we need `System.Transactions`

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,117 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
 		private bool IsLocallySuppressed (int id, ICustomAttributeProvider provider, out SuppressMessageInfo info)
 		{
-			Dictionary<int, SuppressMessageInfo> suppressions;
 			info = default;
+			if (provider == null)
+				return false;
+
+			Dictionary<int, SuppressMessageInfo> suppressions;
 			return _localSuppressions.TryGetValue (provider, out suppressions) &&
 				suppressions.TryGetValue (id, out info);
```
vitek-karas, line 116:
```suggestion
			return _localSuppressions.TryGetValue (provider, out var suppressions) &&
				suppressions.TryGetValue (id, out info);
```

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,117 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+				if (_globalSuppressions.TryGetValue (provider, out var suppressions)) {
+					AddOrUpdate (info, suppressions);
+					return;
+				}
+
+				var _suppressions = new Dictionary<int, SuppressMessageInfo> { { info.Id, info } };
+				_globalSuppressions.Add (provider, _suppressions);
+			}
+
+			private void AddOrUpdate (SuppressMessageInfo info, IDictionary<int, SuppressMessageInfo> builder)
```
vitek-karas, line 216:
```suggestion
			private static void AddOrUpdate (SuppressMessageInfo info, IDictionary<int, SuppressMessageInfo> builder)
```

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,117 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+				if (provider is ModuleDefinition _module)
+					module = _module;
+				else if (provider is AssemblyDefinition assembly)
+					module = assembly.MainModule;
+				else if (provider is TypeDefinition type)
+					module = type.Module;
+				else if (provider is IMemberDefinition member)
+					module = member.DeclaringType.Module;
+				else {
+					_context.LogMessage (MessageContainer.CreateInfoMessage ("`UnconditionalSuppressMessage` attribute was placed in an language element which is currently not supported."));
```
vitek-karas, line 160: We need to `continue` here, otherwise the assert below will fire.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -41,17 +39,27 @@ public void AddLocalSuppression (CustomAttribute ca, ICustomAttributeProvider pr
...
-					if (IsLocallySuppressed (id, memberDefinition, out info))
-						return true;
+			info = default;
+			if (warningOrigin.MemberDefinition == null)
+				return false;
 
-					memberDefinition = memberDefinition.DeclaringType;
-				}
+			IMemberDefinition memberDefinition = warningOrigin.MemberDefinition;
+			if (IsGloballySuppressed (id, memberDefinition, out info))
```
vitek-karas, line 47: Why do we need this here - the while loop below will do exactly this again...

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -41,17 +39,27 @@ public void AddLocalSuppression (CustomAttribute ca, ICustomAttributeProvider pr
...
+				return true;
+
+			// Check if there's a module level suppression.
```
vitek-karas, line 50: I think module level suppressions should be done after we do the while loop - in general I expect local suppressions to be on the method - so that should be the "fast path" through this function.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,117 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+			if (_globalSuppressions == null) {
+				var suppressions = new GlobalSuppressions ();
+				HashSet<ModuleDefinition> modules = new HashSet<ModuleDefinition> ();
+				foreach (var assembly in _context.Resolver.AssemblyCache.Values) {
+					DecodeGlobalSuppressMessageAttributes (assembly, suppressions);
+					modules.Add (assembly.MainModule);
+				}
+
+				foreach (var module in modules)
+					DecodeGlobalSuppressMessageAttributes (module, suppressions);
```
vitek-karas, line 136: Why not just do the module in the foreach above as well? There can be more than one module per assembly, but I don't think there can be a module which spans multiple assemblies.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,117 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+			DecodeGlobalSuppressMessageAttributes ();
+			return _globalSuppressions.HasGlobalSuppressions (id, provider, out info);
+		}
+
+		private void DecodeGlobalSuppressMessageAttributes ()
+		{
+			if (_globalSuppressions == null) {
+				var suppressions = new GlobalSuppressions ();
+				HashSet<ModuleDefinition> modules = new HashSet<ModuleDefinition> ();
+				foreach (var assembly in _context.Resolver.AssemblyCache.Values) {
```
vitek-karas, line 130: Currently this is "probably" OK since we preload all assemblies really early on. That said if we hit a warning before we finish resolving assemblies this will lead to us possibly missing some global suppressions.

I think a better way to do would be to have a dictionary/hashset of already "initialized" assemblies. And when `IsGloballySuppressed` is called, take the assembly from the `provider` and initialize just that assembly (if we didn't already do that). That way even if we move to lazy-loading assemblies, this will just work. And it's also "lazy" in the sense that we won't go over all assemblies, but only those which generate warnings (which is almost guaranteed to be the set which is marked).

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,116 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+				foreach (var module in modules)
+					DecodeGlobalSuppressMessageAttributes (module, suppressions);
+
+				_globalSuppressions = suppressions;
+			}
+		}
+
+		private void DecodeGlobalSuppressMessageAttributes (ICustomAttributeProvider provider, GlobalSuppressions globalSuppressions)
+		{
+			var attributes = provider.CustomAttributes.Where (a => a.AttributeType.Name == "UnconditionalSuppressMessageAttribute");
```
marek-safar, line 143: What about the namespace?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,116 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+			foreach (var instance in attributes) {
+				SuppressMessageInfo info;
+				if (!TryDecodeSuppressMessageAttributeData (instance, out info))
+					continue;
+
+				ModuleDefinition module = null;
+				if (provider is ModuleDefinition _module)
```
marek-safar, line 150: Use switch

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,116 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+				case "namespaceanddescendants":
+					_context.LogMessage (MessageContainer.CreateInfoMessage ($"Scope `{info.Scope}` used in `UnconditionalSuppressMessage` is currently not supported."));
+					break;
+				default:
+					break;
+				}
+			}
+		}
+
+		private class GlobalSuppressions
```
marek-safar, line 190: Please fold that into parent class

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,116 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+				if (_globalSuppressions.TryGetValue (provider, out var suppressions)) {
+					AddOrUpdate (info, suppressions);
+					return;
+				}
+
+				var _suppressions = new Dictionary<int, SuppressMessageInfo> { { info.Id, info } };
+				_globalSuppressions.Add (provider, _suppressions);
+			}
+
+			private static void AddOrUpdate (SuppressMessageInfo info, IDictionary<int, SuppressMessageInfo> builder)
```
marek-safar, line 215: The name says Add, where is that code?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: Wouldn't this cause a problem when checking for global suppressions? I think we need to go through all assemblies to check if there are any global suppressions specified which could potentially be targeting types or members in other assemblies. I'm thinking in a scenario where an assembly B has the following suppression: 
```C#
[module: UnconditionalSuppressMessage (..., Scope = "type", Target = "A.Foo")]
```
but there's no warning thrown from within B, thus never initializing it's assembly here. Then, if a warning is thrown in the scope of type `A.Foo`, we wouldn't know that this warning was meant to be suppressed.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: I think I don't understand what you mean by folding this into the parent class. Could you please clarify?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: We're adding or updating by using the indexer. If we use the dictionary `Add` method here, we could be throwing an exception. I can change the name of this method it it's misleading.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +107,116 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+				case "member":
+					foreach (var result in DocumentationSignatureParser.GetMembersForDocumentationSignature (info.Target, module))
+						globalSuppressions.AddGlobalSuppression (result, info);
+
+					break;
+				case "resource":
+				case "module":
+				case "namespace":
+				case "namespaceanddescendants":
+					_context.LogMessage (MessageContainer.CreateInfoMessage ($"Scope `{info.Scope}` used in `UnconditionalSuppressMessage` is currently not supported."));
```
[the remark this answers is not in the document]
mateoatr: I've removed these cases, will open an issue to keep track of this scopes, which we probably want to implement (e.g., the "resource" scope is important for scenarios like https://github.com/mono/linker/issues/1179)

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
marek-safar: This is a class with like 10 lines where most global logic is in parent class already. You could make for change `IsGloballySuppressed` to have full implementation and it'd make the code more readable.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -114,85 +116,94 @@ private bool IsLocallySuppressed (int id, ICustomAttributeProvider provider, out
...
-
-				Debug.Assert (module != null);
-				if (info.Target == null && (info.Scope == "module" || info.Scope == null)) {
-					AddLocalSuppression (instance, module);
-					continue;
-				}
+			void LookForSuppressions (ICustomAttributeProvider provider) {
+				ModuleDefinition module = provider is ModuleDefinition ? provider as ModuleDefinition : (provider as AssemblyDefinition).MainModule; 
+				var attributes = provider.CustomAttributes.
+					Where (a => a.AttributeType.Name == "UnconditionalSuppressMessageAttribute" && a.AttributeType.Namespace == "System.Diagnostics.CodeAnalysis");
```
marek-safar, line 160: Extract the check to static method. I think you are using it at multiple places already

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
marek-safar: I don't think that is what the code does. It would if you removed `ContainsKey` logic

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -41,17 +39,26 @@ public void AddLocalSuppression (CustomAttribute ca, ICustomAttributeProvider pr
...
-					if (IsLocallySuppressed (id, memberDefinition, out info))
-						return true;
+			info = default;
+			if (warningOrigin.MemberDefinition == null)
+				return false;
 
-					memberDefinition = memberDefinition.DeclaringType;
-				}
+			IMemberDefinition memberDefinition = warningOrigin.MemberDefinition;
+			while (memberDefinition != null) {
```
vitek-karas, line 47: Nit: I would use a separate local variable for the loop here - currently it's confusing.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +106,127 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+		{
+			DecodeGlobalSuppressMessageAttributes (provider);
+			return _globalSuppressions.HasGlobalSuppressions (id, provider, out info);
+		}
+
+		private void DecodeGlobalSuppressMessageAttributes (ICustomAttributeProvider provider)
+		{
+			ModuleDefinition module;
+			switch (provider.MetadataToken.TokenType) {
+			case TokenType.Module:
```
vitek-karas, line 127: I would extract this to a static helper - GetModuleForProvider - or something like that.
Then use it here and also in the parent class when we check for local suppressions on module/assembly. It would make the code more readable.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +106,127 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
 
 		private bool IsLocallySuppressed (int id, ICustomAttributeProvider provider, out SuppressMessageInfo info)
```
vitek-karas, line 107: Not for this PR but eventually we should cleanup the types here. The warning requires the "location" to be `IMemberDefinition`, but then all of the suppression logic works on `ICustomAttributeProvider`. Would be nice to unify this.

For example it's possible to suppress a warning at the assembly level, but currently it's not possible to emit a warning at the assembly level.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +106,127 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+				ModuleDefinition module = provider is ModuleDefinition ? provider as ModuleDefinition : (provider as AssemblyDefinition).MainModule; 
+				var attributes = provider.CustomAttributes.
+					Where (a => a.AttributeType.Name == "UnconditionalSuppressMessageAttribute" && a.AttributeType.Namespace == "System.Diagnostics.CodeAnalysis");
+				foreach (var instance in attributes) {
+					SuppressMessageInfo info;
+					if (!TryDecodeSuppressMessageAttributeData (instance, out info))
+						continue;
+
+					if (info.Target == null && (info.Scope == "module" || info.Scope == null)) {
+						AddLocalSuppression (instance, provider);
```
vitek-karas, line 167: Naming is weird here. The method we're in is called `DecodeGlobal...` but here we're adding local suppressions.

vitek-karas: I guess I find it weird how we store the suppressions.
Maybe a better design would be to have one storage (the one which is currently called local) and then each time check against it we first make sure that global suppressions where processed (initialized), those would all feed into the same storage at the end.

If you look at the two local/global, they store the exact same information in the end - same dictionary, same kind of lookup and so on. The only difference is how they get initialized.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +106,127 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+			if (!_globalSuppressions.InitializedAssemblies.Contains (assembly)) {
+				DecodeGlobalSuppressMessageAttributes (assembly);
+				_globalSuppressions.InitializedAssemblies.Add (assembly);
+			}
+		}
+
+		private void DecodeGlobalSuppressMessageAttributes (AssemblyDefinition assembly)
+		{
+			void LookForSuppressions (ICustomAttributeProvider provider) {
+				ModuleDefinition module = provider is ModuleDefinition ? provider as ModuleDefinition : (provider as AssemblyDefinition).MainModule;
```
vitek-karas, line 158: The caller knows this information already - why do we have to compute it again.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +106,127 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+			AssemblyDefinition assembly = module.Assembly;
+			if (!_globalSuppressions.InitializedAssemblies.Contains (assembly)) {
+				DecodeGlobalSuppressMessageAttributes (assembly);
+				_globalSuppressions.InitializedAssemblies.Add (assembly);
+			}
+		}
+
+		private void DecodeGlobalSuppressMessageAttributes (AssemblyDefinition assembly)
+		{
+			void LookForSuppressions (ICustomAttributeProvider provider) {
```
vitek-karas, line 157: Extract this to a static method outside - I don't see why this needs to be a local method.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -99,10 +106,127 @@ private static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attri
...
+			private readonly Dictionary<ICustomAttributeProvider, Dictionary<int, SuppressMessageInfo>> _globalSuppressions;
+			public HashSet<AssemblyDefinition> InitializedAssemblies { get; private set; }
+
+			public GlobalSuppressions ()
+			{
+				_globalSuppressions = new Dictionary<ICustomAttributeProvider, Dictionary<int, SuppressMessageInfo>> ();
+				InitializedAssemblies = new HashSet<AssemblyDefinition> ();
+			}
+
+			public bool HasGlobalSuppressions (int id, ICustomAttributeProvider provider, out SuppressMessageInfo info)
```
vitek-karas, line 204: Should be called `HasGlobalSuppression` - it only gets one.

PR #1234, opened 2020-06-03
Add CombineLinkerXmlFiles task

eerhardt:
This task combines multiple linker xml files into a single, combined xml file.

@sbomer @marek-safar


eerhardt:
An example of using this task is here:

https://github.com/dotnet/runtime/pull/37288

Which makes the first `System.Private.CoreLib` substitutions xml file. And then we can add more xml files to be combined together like this sample: https://github.com/dotnet/runtime/commit/9678f41842c82c5b08dccdfa646bfdd89eb446ba

PR #1236, opened 2020-06-03
Scope down PublicConstructors to DefaultConstructor if possible

### LINKED ISSUE #1222: GetConstructor(Type.EmptyTypes) should only need DynamicallyAccessedMemberTypes.DefaultConstructor
When code is calling `type.GetConstructor(Type.EmptyTypes)`, we should only need to annotate the `type` with `DynamicallyAccessedMemberTypes.DefaultConstructor` since that is the constructor being reflected upon.

Instead, today you need to use `DynamicallyAccessedMemberTypes.PublicConstructors`, which preserves too much.

See https://github.com/dotnet/runtime/pull/36532#discussion_r431930071

cc @MichalStrehovsky

MichalStrehovsky:
Also recognize `Type.EmptyTypes` and `Array.Empty<Type>`.

Fixes #1222.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -149,10 +149,18 @@ namespace Mono.Linker.Dataflow
 
 		protected override ValueNode GetFieldValue (MethodDefinition method, FieldDefinition field)
 		{
-			DynamicallyAccessedMemberTypes memberKinds = _flowAnnotations.GetFieldAnnotation (field);
-			return new LoadFieldValue (field, memberKinds) {
-				SourceContext = method
-			};
+			switch (field.Name) {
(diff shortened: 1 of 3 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -741,6 +758,14 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 							if (methodParams[1].AsConstInt () != null)
 								bindingFlags |= (BindingFlags) methodParams[1].AsConstInt ();
 						}
+						int ctorParameterCount = parameters.Count switch
+						{
+							1 => (methodParams[1] as ArrayValue)?.Size.AsConstInt () ?? -1,
```
vitek-karas, line 763: Nit I would prefer to use `int?` and use `null` as the "we don't know" value.

vitek-karas:
I agree - it's nice the data flow can figure that out.

Issue #1129, opened 2020-04-21
Clean up ClearInitLocalsStep step


marek-safar:
There are few oddities in this step which would be good to clean up.

- the step changes action on assemblies from copy to save. This is not what linker does for any other optimization and should be avoided. Instead the optimization should run only on linked assemblies.
- the integration with msbuild uses custom-data context to pass the list of assemblies in which is not in line with any other optimization which uses common per assembly optimization setting

PR #1237, opened 2020-06-03
Clean up clearinitlocals handling

sbomer:
- Avoids changing assembly action
- Uses assembly optimizations instead of custom data
- Fixes https://github.com/mono/linker/issues/1129

### DIFF
--- a/src/linker/Linker.Steps/ClearInitLocals.cs
+++ b/src/linker/Linker.Steps/ClearInitLocals.cs
@@ -6,15 +6,6 @@ namespace Mono.Linker.Steps
 {
 	public class ClearInitLocalsStep : BaseStep
 	{
-		HashSet<string> _assemblies;
-
-		protected override void Process ()
-		{
-			if (Context.TryGetCustomData ("ClearInitLocalsAssemblies", out string parameter)) {
-				_assemblies = new HashSet<string> (parameter.Split (','), StringComparer.OrdinalIgnoreCase);
-			}
-		}
-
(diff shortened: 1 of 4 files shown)


MichalStrehovsky:
Do we need to keep supporting this now that https://github.com/dotnet/runtime/pull/454 is part of .NET 5 (and the Roslyn feature is in Roslyn's master)?

sbomer:
I think once https://github.com/dotnet/runtime/issues/35527 is finished we can drop it. dotnet/runtime is the only place I know of that is using the optimization.

src/linker/Linker.Steps/ClearInitLocals.cs
```diff
@@ -30,28 +21,25 @@ private static IEnumerable<TypeDefinition> EnumerateTypesAndNestedTypes (IEnumer
...
 		{
-			if (!Context.IsOptimizationEnabled (CodeOptimizations.ClearInitLocals, assembly) &&
-				_assemblies?.Contains (assembly.Name.Name) != true) {
+			if (!Context.IsOptimizationEnabled (CodeOptimizations.ClearInitLocals, assembly))
 				return;
-			}
 
-			bool changed = false;
+			if (Annotations.GetAction (assembly) != AssemblyAction.Link) {
+				Context.LogMessage (MessageContainer.CreateErrorMessage ($"Clearinitlocals is not supported for copy assembly '{assembly.Name.Name}'.", 1014));
```
marek-safar, line 28: This can happen for any action (e.g. skip) so I'd not report it as error and just return

PR #1240, opened 2020-06-03
Support for annotations in LinkContext

tlakollo:
- Create an instance of CustomAttributeSource in LinkContext
- Support for Embedded XMLs
- Cache now uses LinkContext CustomAttributeSource to have both XML and Cecil attributes
- Attribute defs xmls are now read in the steps pipeline as XmlCustomAttributesStep
- Dictionary of ICustomAttributeProvider now lives in CustomAttributeSource and is a unified dictionary for all XMLs
- CustomAttributeSource exposes AddCustomAttribute in order to be used by the XmlCustomAttributesStep


src/linker/Linker/CustomAttributeSource.cs
```diff
@@ -30,13 +28,9 @@ public IEnumerable<CustomAttribute> GetCustomAttributes (ICustomAttributeProvide
...
 
-			if (_sources.Count > 0) {
-				foreach (var source in _sources) {
-					if (source.HasCustomAttributes (provider)) {
-						foreach (var customAttribute in source.GetCustomAttributes (provider))
-							yield return customAttribute;
-					}
-				}
+			if (_xmlCustomAttributes.ContainsKey (provider)) {
+				foreach (var customAttribute in _xmlCustomAttributes.TryGetValue (provider, out var ann) ? ann : null)
```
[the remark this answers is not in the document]
tlakollo: It doesn't work with null values, still in this case since it was after a ContainsKey is never going to be null. Still is more performant only running TryGetValue once. So Im going to change it.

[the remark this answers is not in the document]
tlakollo: I end up doing
```
if (_xmlCustomAttributes.ContainsKey (provider)) {
	foreach (var customAttribute in _xmlCustomAttributes[provider])
		yield return customAttribute;
}
```
Which is optimized for not founding the provider, since it never allocates the default in TryGetValue (https://referencesource.microsoft.com/#mscorlib/system/collections/generic/dictionary.cs,2e5bc6d8c0f21e67)
I made a small benchmark and TryGetValue is faster than doing ContainsKey and then trying to use the get indexer when the provider exists. 
So this implementation will be faster if the value is not found but will be slower in the case the value is found. In this case most of the time the value will not be found.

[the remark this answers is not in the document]
vitek-karas: Nit: It's good to add a comment into the code with this reasoning - for example I would be tempted to rewrite it with just `TryGet` 😄

src/linker/Linker/Driver.cs
```diff
@@ -633,6 +645,8 @@ protected int SetupContext (ILogger customLogger = null)
 			//   dynamically adds steps:
 			//     ResolveFromXmlStep [optional, possibly many]
 			//     BodySubstituterStep [optional, possibly many]
+			//     XmlCustomAttributesStep [optional, possibly many]
+			// XmlCustomAttributesStep [optional, possibly many]
```
vitek-karas, line 649: Why add it twice?

tlakollo: BlackList step adds XmlCustomAttributesStep based on the embedded xml on the assemblies

src/linker/Linker/Driver.cs
```diff
@@ -1033,6 +1052,7 @@ static void Usage ()
 			Console.WriteLine ("  --strip-substitutions     Remove XML substitution resources for linked assemblies. Defaults to true");
 			Console.WriteLine ("  --used-attrs-only         Attribute usage is removed if the attribute type is not used. Defaults to false");
 			Console.WriteLine ("  --attribute-defs FILE     Supplementary custom attribute definitions for attributes controlling the linker behavior.");
+			Console.WriteLine ("  --ignore-attributes       Skips reading embedded attributes. Defaults to false");
```
marek-safar, line 1055: I think `--ignore-attributes` is not the right name for this. I actually like the annotation name better but we need something that is be less ambiguous

src/linker/Linker/CustomAttributeSource.cs
```diff
@@ -30,13 +32,11 @@ public IEnumerable<CustomAttribute> GetCustomAttributes (ICustomAttributeProvide
...
-			if (_sources.Count > 0) {
-				foreach (var source in _sources) {
-					if (source.HasCustomAttributes (provider)) {
-						foreach (var customAttribute in source.GetCustomAttributes (provider))
-							yield return customAttribute;
-					}
-				}
+			// This piece of code is optimized for when _xmlCustomAttributes doesn't contain provider
+			// That's why is using ContainsKey and the get indexer instead of TryGetValue
+			if (_xmlCustomAttributes.ContainsKey (provider)) {
```
marek-safar, line 37: Why not to use TryGet?

tlakollo: Visual Studio doesn't show TryGet, neither is in https://referencesource.microsoft.com/#mscorlib/system/collections/generic/dictionary.cs,2e5bc6d8c0f21e67
Maybe I'm missing context on how to use TryGet

src/linker/Linker/Driver.cs
vitek-karas: "custom attributes" ? I know it also slightly collides with the terminology in reflection/Cecil, but it's at least a bit more specific.
I personally don't like annotations as the mechanism is all about attributes (the intent is annotations yes, but that's really fuzzy).

src/linker/Linker/CustomAttributeSource.cs
marek-safar: Sorry, I mean TryGetValue

vitek-karas: The reason is in the comment - it turns out to be faster for negative cases to call `ContainsKey` - and this will be almost always called on negative cases.

src/linker/Linker/Driver.cs
marek-safar: I think we need to capture that the intent is to control the attributes "placement" not the attributes itself. I'm not sure what most people understand by attributes but considering this is a very advanced feature it should target the people who work in this space.

vitek-karas: Two points:
* I think we need to synchronize the name of the command line argument with the name of the embedded resource - whatever it is.
* What about "additional attributes", or "additional link attributes"?

marek-safar: > I think we need to synchronize the name of the command line argument with the name of the embedded resource 

I agree but `ILLink.Attributes.xml` does not allow you to define attributes only apply existing ones, right?

src/linker/Linker/CustomAttributeSource.cs
marek-safar: I'd prefer simpler code over micro-optimization with explicit comment

src/linker/Linker/Driver.cs
vitek-karas: Yes, that's right. That said my understanding of "add attribute" is not to define a new one, but to apply an existing one... but maybe that's just me.

vitek-karas: Discussed offline with @marek-safar and we sort of agreed on calling it "LinkAttributes". So:
* command line to specify .xml externally `--link-attributes <filename>` (this is to replace the existing `--attribute-defs`)
* embedded file called `ILLink.LinkAttributes.xml`
* command line to enable/disable stripping of the embedded `ILLink.LinkAttributes.xml` will be called `--strip-link-attributes`
* command line to effectively disable the feature `--ignore-link-attributes`

src/linker/Linker/CustomAttributeSource.cs
vitek-karas: Let's go with `TryGetValue` - if this becomes a perf problem we can change it in the future.

src/linker/Linker/Driver.cs
vitek-karas: We also plan to rename the embedded descriptors to `ILLink.Descriptors.xml` - see https://github.com/mono/linker/issues/1257.

src/linker/Linker.Steps/XmlCustomAttributesStep.cs
```diff
@@ -5,35 +5,34 @@
...
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Xml.XPath;
 using Mono.Cecil;
 
-namespace Mono.Linker
+namespace Mono.Linker.Steps
 {
-	class XmlCustomAttributeSource
+	class XmlCustomAttributesStep : BaseStep
```
vitek-karas, line 15: Consider renaming this to LinkAttributesStep or something like that - to match the "feature name", the "XML" part is effectively an implementation detail.

src/linker/Linker/Driver.cs
```diff
@@ -1032,7 +1057,9 @@ static void Usage ()
 			Console.WriteLine ("  --ignore-substitutions    Skips reading embedded substitutions. Defaults to false");
 			Console.WriteLine ("  --strip-substitutions     Remove XML substitution resources for linked assemblies. Defaults to true");
 			Console.WriteLine ("  --used-attrs-only         Attribute usage is removed if the attribute type is not used. Defaults to false");
-			Console.WriteLine ("  --attribute-defs FILE     Supplementary custom attribute definitions for attributes controlling the linker behavior.");
+			Console.WriteLine ("  --link-attributes FILE    Supplementary custom attribute definitions for attributes controlling the linker behavior.");
+			Console.WriteLine ("  --ignore-link-attributes  Skips reading embedded attributes. Defaults to false");
+			Console.WriteLine ("  --strip-link-attributes   Remove XML link attributes resources for linker assemblies. Defaults to true");
```
vitek-karas, line 1062:
```suggestion
			Console.WriteLine ("  --strip-link-attributes   Remove XML link attributes resources for linked assemblies. Defaults to true");
```

src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -126,18 +129,21 @@ ArrayBuilder<string> GetAttributeChildren (XPathNodeIterator iterator)
 			return children;
 		}
 
-		public void ParseXml (string document)
+		protected override void Process ()
 		{
-			_xmlDocumentLocation = document;
-			_document = new XPathDocument (_xmlDocumentLocation);
+			if (_resource != null && Context.StripLinkAttributes)
```
marek-safar, line 134: Use single if (_resource != null) check

vitek-karas: We have the same pattern in BodySubstituterStep - can you please change it there as well?

tlakollo: Changed the pattern in LinkAttributesStep, BodySubstituterStep and ResolveFromXmlStep

PR #1245, opened 2020-06-04
Implement data flow annotations on generic parameters

### LINKED ISSUE #1086: [feature/dataflow] Allow DynamicallyAccessedMembers on generic types
In `System.Text.Json`, we have code like the [following](https://github.com/dotnet/runtime/blob/3838fb74a57d658200f28443ff2086849210d19e/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializer.Write.Utf8JsonWriter.cs#L16-L26)

```C#
        internal static void Serialize<T>(
            Utf8JsonWriter writer,
            T value,
            JsonSerializerOptions options,
            ref WriteStack state,
            string? propertyName = null)
        {
            if (options == null)
            {
                throw new ArgumentNullException(nameof(options));
            }

            JsonConverter jsonConverter = state.Current.InitializeReEntry(typeof(T), options, propertyName);
            bool success = jsonConverter.TryWriteAsObject(writer, value, options, ref state);
            Debug.Assert(success);
        }
```

Here the type `T` needs to be annotated with `DynamicallyAccessedMemberKinds.Properties` because `InitializeReEntry` will eventually need to access all the properties on `T`. However, the linker currently doesn't recognize when I try to annotate the method like the following:

```C#
        internal static void Serialize<[DynamicallyAccessedMembers(DynamicallyAccessedMemberKinds.Properties)] T>(
            Utf8JsonWriter writer,
            T value,
            JsonSerializerOptions options,
            ref WriteStack state,
            string? propertyName = null)
        {
```

cc @MichalStrehovsky @vitek-karas

vitek-karas:
The checks are done when linker calls `MarkType(TypeReference)` or `MarkMethod(MethodReference)`. The check itself is straightforward as it uses existing functionality.

Added annotations in `FlowAnnotations` for generic parameters.

Added one parameter to `MarkType` to better track the location where the type reference comes from. This enables much more precise reporting of warnings.

On test side this makes the recognized/unrecognized attributes applicable to many more things as warnings can come from types, fields and so on.

Fixes #1086

### DIFF
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -79,13 +79,9 @@ namespace Mono.Linker.Dataflow
 			}
 
 			MethodDefinition declaringMethod = genericParameter.DeclaringMethod?.Resolve ();
-			if (declaringMethod != null) {
-				if (GetAnnotations (declaringMethod.DeclaringType).TryGetAnnotation (declaringMethod, out var methodAnnotations) &&
-					methodAnnotations.TryGetAnnotation (genericParameter, out var annotation))
-					return annotation.Annotation;
-
-				return DynamicallyAccessedMemberTypes.None;
-			}
+			if (declaringMethod != null && GetAnnotations (declaringMethod.DeclaringType).TryGetAnnotation (declaringMethod, out var methodTypeAnnotations) &&
+				methodTypeAnnotations.TryGetAnnotation (genericParameter, out var methodAnnotation))
+				return methodAnnotation.Annotation;
 
 			return DynamicallyAccessedMemberTypes.None;
 		}
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -31,6 +31,11 @@ namespace Mono.Linker.Dataflow
 			return GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out _);
 		}
 
+		public bool RequiresDataFlowAnalysis (GenericParameter genericParameter)
+		{
+			return GetGenericParameterAnnotation (genericParameter) != DynamicallyAccessedMemberTypes.None;
+		}
+
 		/// <summary>
 		/// Retrieves the annotations for the given parameter.
 		/// </summary>
@@ -58,6 +63,28 @@ namespace Mono.Linker.Dataflow
 		{
 			if (GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out var annotation)) {
 				return annotation.Annotation;
+			}
+
+			return DynamicallyAccessedMemberTypes.None;
+		}
+
+		public DynamicallyAccessedMemberTypes GetGenericParameterAnnotation (GenericParameter genericParameter)
+		{
+			TypeDefinition declaringType = genericParameter.DeclaringType?.Resolve ();
+			if (declaringType != null) {
+				if (GetAnnotations (declaringType).TryGetAnnotation (genericParameter, out var annotation))
+					return annotation.Annotation;
+
+				return DynamicallyAccessedMemberTypes.None;
+			}
+
+			MethodDefinition declaringMethod = genericParameter.DeclaringMethod?.Resolve ();
+			if (declaringMethod != null) {
+				if (GetAnnotations (declaringMethod.DeclaringType).TryGetAnnotation (declaringMethod, out var methodAnnotations) &&
+					methodAnnotations.TryGetAnnotation (genericParameter, out var annotation))
+					return annotation.Annotation;
+
+				return DynamicallyAccessedMemberTypes.None;
 			}
 
 			return DynamicallyAccessedMemberTypes.None;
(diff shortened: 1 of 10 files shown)


vitek-karas:
This is still missing making `MakeGenericType` and `MakeGenericMethod` potentially linker unfriendly, but I wanted to get early feedback on the core of the changes.

MichalStrehovsky:
Looks reasonable. Would it make sense to treat `new()` constraint as an implicit `PublicParameterlessConstructor` annotation on the parameter?

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -65,6 +70,28 @@ public DynamicallyAccessedMemberTypes GetFieldAnnotation (FieldDefinition field)
 			return DynamicallyAccessedMemberTypes.None;
 		}
 
+		public DynamicallyAccessedMemberTypes GetGenericParameterAnnotation (GenericParameter genericParameter)
+		{
+			TypeDefinition declaringType = genericParameter.DeclaringType?.Resolve ();
```
marek-safar, line 75: What about a case like

```c#
class A<T>
{
    class B
    {
        static void Foo ()
        {
			T t;
        }
    }
}
```

will the generic parameter point to the right one with the annotation ?

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -693,7 +693,11 @@ private void ScanLdarg (Instruction operation, Stack<StackSlot> currentStack, Me
 			MethodDefinition thisMethod,
 			MethodBody methodBody)
 		{
-			if (operation.Operand is TypeReference typeReference) {
+			if (operation.Operand is GenericParameter genericParameter) {
+				StackSlot slot = new StackSlot (new RuntimeTypeHandleForGenericParameterValue (genericParameter));
+				currentStack.Push (slot);
+				return;
+			} else if (operation.Operand is TypeReference typeReference) {
```
marek-safar, line 700: nit: redundant else

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1863,11 +1875,11 @@ protected virtual void MarkMulticastDelegate (TypeDefinition type)
 			MarkMethodCollection (type.Methods, new DependencyInfo (DependencyKind.MethodForSpecialType, type));
 		}
 
-		protected (TypeReference, DependencyInfo) GetOriginalType (TypeReference type, DependencyInfo reason)
+		protected (TypeReference, DependencyInfo) GetOriginalType (TypeReference type, DependencyInfo reason, IMemberDefinition sourceLocationMember)
 		{
 			while (type is TypeSpecification specification) {
 				if (type is GenericInstanceType git) {
-					MarkGenericArguments (git);
+					MarkGenericArguments (git, sourceLocationMember ?? (reason.Source as IMemberDefinition) ?? (reason.Source as MemberReference)?.Resolve ());
```
marek-safar, line 1882: I'd prefer if `reason.Source` was not used in this context as I have no confidence in the value and we have no tests for that

src/linker/Linker.Dataflow/FlowAnnotations.cs
MichalStrehovsky: Yes, the C# compiler propagates the attributes on `A`'s `T` to the `B`'s `T`.

We might have reporting that doesn't make sense to end users though (because it will refer to `B`'s generic parameter). I would probably leave fixing that to user feedback - it requires hardcoding C# compiler's details that I'm not sure map to e.g. F#.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: In that case we would have to pass the new parameter on every call to MarkType and I would have to add a similar parameter to MarkMethod and also pass it everywhere.

I actually did that at first, but it turned out that in almost all cases we passed the same value to that new parameter as we do to reason.Source.

So in a way the tests for this new functionality would also validate the reason.Source.

src/linker/Linker.Dataflow/FlowAnnotations.cs
vitek-karas: I'll add a test case for this - it's not great but it's OKish. Since we report the full type name and the actual name of the generic parameter like this ``Generic parameter 'T' from 'A`1/B'`` I think it's acceptable.

vitek-karas:
@MichalStrehovsky @marek-safar can you please take a look again - it should be complete now.

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -28,6 +29,9 @@ public enum ValueNodeKind
 		MethodParameter,                // symbolic placeholder
 		MethodReturn,                   // symbolic placeholder
 
+		RuntimeTypeHandleForGenericParameter, // symbolic placeholder for generic parameter
```
MichalStrehovsky, line 32: Add the new values to `ValueNodeExtensions.DetectCycle`?

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: I propagated the source member explicitly - this spreads everywhere, so the change is large (almost all of mark step), but mostly it's just mechanical.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -65,6 +70,28 @@ public DynamicallyAccessedMemberTypes GetFieldAnnotation (FieldDefinition field)
...
+				return DynamicallyAccessedMemberTypes.None;
+			}
+
+			MethodDefinition declaringMethod = genericParameter.DeclaringMethod?.Resolve ();
+			if (declaringMethod != null) {
+				if (GetAnnotations (declaringMethod.DeclaringType).TryGetAnnotation (declaringMethod, out var methodAnnotations) &&
+					methodAnnotations.TryGetAnnotation (genericParameter, out var annotation))
+					return annotation.Annotation;
+
+				return DynamicallyAccessedMemberTypes.None;
```
marek-safar, line 89: nit: redundant return?

Issue #1220, opened 2020-05-28
Improve member signatures used in warnings and errors


marek-safar:
Most of the code uses `FullName` property to report the name of the member involved in the warning/error message. This is not a great user experience as FullName is very verbose as it includes

- return type
- field type
- property/event type
- modreq/modopt modifiers with their types
- a fully qualified name for primitive types

It also uses a syntax which is not very friendly to high-level language users
- nested types separated by `/`
- .ctor() for constructors
- set_X, get_X for accessors
- `X for generic arity
- etc

I think we should simplify that to make it more friendly to developers writing code in their language. We could start with C# for now and add other languages later.

Perhaps the best way to implement this would be inside Cecil.

vitek-karas:
This is basically fixed, we use "DisplayName" which currently tries to match C# formatting rules. There's a discussion about using Reflection formatting rules, but that's separate from this issue.

PR #1249, opened 2020-06-05
Improve method signature used in warnings and errors 

mateoatr:
Opening this as a WIP to collect some opinions before modifying all the places in which we log a method's name to the user.

This adds the `GetName` method extension for method references. The intention is to log the name of methods in a more readable fashion for the end user than using the `FullName` property, which is very verbose (see https://github.com/mono/linker/issues/1220). This reuses the `GetSignature` implementation of https://github.com/mono/linker/pull/1217. Notable changes are that we don't log the return type of the method, the full name of the declaring type, the `get_` and `set_` prefixes when dealing with getter/setter, nor use the full name of the parameter types. Here are some examples of `FullName` vs `GetName` to give a sense of how these compare:

---

Setter and getter of:
```C#
        public bool AllowMultiple
        {
            get => _allowMultiple;
            set => _allowMultiple = value;
        }
```

| FullName | GetName |
|---|---|
| `System.Void System.AttributeUsageAttribute::set_AllowMultiple(System.Boolean)` | `AttributeUssageAttribute::AllowMultiple(Boolean)` |
| `System.Void System.AttributeUsageAttribute::get_AllowMultiple()` | `AttributeUssageAttribute::AllowMultiple()` |

---

Class constructor:
```C#
    public class DebuggerDisplayAttributeOnAssemblyUsingTarget
	{
		public class Foo { ... }
	}
```

| FullName | GetName |
|---|---|
| `System.Void Mono.Linker.Tests.Cases.Attributes.Debugger.DebuggerDisplayAttributeOnAssemblyUsingTarget/Foo::.ctor()` | `Foo::Foo()` |

---
"Common" method:

```C#
    public sealed partial class String : ...
    {
        public static bool IsNullOrEmpty([NotNullWhen(false)] string? value) { ... }
    }    
```

| FullName | GetName |
|---|---|
| `System.Boolean System.String::IsNullOrEmpty(System.String)` | `String::IsNullOrEmpty(String)` |


src/linker/Linker/MethodReferenceExtensions.cs
```diff
@@ -8,6 +8,46 @@ namespace Mono.Linker
 {
 	public static class MethodReferenceExtensions
 	{
+		public static string GetName (this MethodReference method)
+		{
+			var builder = new System.Text.StringBuilder ();
+			builder.Append (method.DeclaringType.Name);
```
sbomer, line 14: I think we should include generic type parameters of the declaring type.

sbomer:
Other cases we would want to handle: arrays, byrefs, pointers. You could look at https://github.com/mono/linker/blob/master/src/linker/Linker/DocumentationSignatureGenerator.PartVisitor.cs for an example.

vitek-karas:
I think we should fully qualify the main type name, linker will report warnings from all modules in the application and so reporting just the typename may not be specific enough. Also given a "random" typename which may very well be an internal type I as a user will have a hard time associating that with an assembly.

I think it's OK to drop namespaces from types in parameters and everywhere else.

Turning `get_Property()` to `Property()` feels wrong - it looks like the property is a method, which it's not. I personally would keep the `get_Property()`, but if we think it's not C#y enough we can use something like `getter for Property` or in the worst case just `Property` (but that will have problems with being specific enough in some cases). I would prefer correctness and precision over brevity.

I would also use a different method than `GetName` it's way too generic in my mind. I think it should be `GetDisplayName` or `GetDescription` or something which would mean that I'm getting a humanly readable representation. Note that most of these types already have a `Name` property, so `GetName` would infer getting that value.

I'm fine if we implement this only for the most common cases (methods really are the big problem right now) and return `FullName` for the rest. And improve on this over time. I would rather we have some improvements now, than spend lot of time to make everything work (like array types, ref types, in/out and so on).

Note: we will need to eventually come up with a syntax for things which C# doesn't really have a syntax for, for example the generic parameter TValue in `Dictionary<TKey,TValue>`. Just saying `TValue` is not enough on its own, it needs context of the type/method it belongs to. Or "return value of method Foo", and so on.

test/Mono.Linker.Tests/Tests/GetDisplayNameTests.cs
```diff
@@ -0,0 +1,155 @@
...
+			}
+
+		}
+
+		public static IEnumerable<object[]> GetMemberAssertionsAsArray (Type type)
+		{
+			return MemberAssertionsCollector.GetMemberAssertions (type).Select (v => new object[] { v.member, v.ca });
+		}
+
+		[DisplayName ("GetDisplayNameTests.A")]
```
marek-safar, line 45: Please add tests for
- generic types definition including nested combinations
- ref/out methods
- ref return types
- generic methods inside generic types

It'd be also good to have tests for TypeReferences

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -8,6 +7,53 @@ namespace Mono.Linker
...
+
+			var builder = new System.Text.StringBuilder ();
+			if (type.DeclaringType != null)
+				builder.Append (type.DeclaringType.GetDisplayName ()).Append (".");
+
+			if (type.HasGenericParameters) {
+				builder.Append (type.Name.Substring (0, type.Name.IndexOf ('`')));
+				builder.Append ('<');
+				for (int i = 0; i < type.GenericParameters.Count - 1; i++)
+					builder.Append ($"{type.GenericParameters[i]},");
```
marek-safar, line 23: Better pattern is not to string.concant before calling into StringBuilder (you have that at many places)

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Text;
```
marek-safar, line 2: Why move it to GetDisplayName?

src/linker/Linker/MethodReferenceExtensions.cs
```diff
@@ -11,7 +11,9 @@ public static class MethodReferenceExtensions
 		public static string GetDisplayName (this MethodReference method)
 		{
 			var builder = new System.Text.StringBuilder ();
-			builder.Append (method.DeclaringType.FullName);
+			builder.Append ((method as MemberReference).GetNamespaceDisplayName ());
```
vitek-karas, line 14: Why do you need the "as" here - MethodReference should always be a MemberReference.

src/linker/Linker/MethodReferenceExtensions.cs
```diff
@@ -11,7 +11,9 @@ public static class MethodReferenceExtensions
...
+			builder.Append ('.');
+			builder.Append (method.DeclaringType.GetDisplayName ());
 			builder.Append ("::");
```
vitek-karas, line 17: Can you please change this to `.`? `::` is a C++ syntax, not C#.

src/linker/Linker/MethodReferenceExtensions.cs
```diff
@@ -10,38 +10,36 @@ public static class MethodReferenceExtensions
...
-				builder.Append (method.Parameters[method.Parameters.Count - 1].ParameterType.GetDisplayName ());
+			sb.Append (")");
+			
+			// Insert generic parameters
+			if (method.HasGenericParameters) {
+				TypeReferenceExtensions.ParseGenericParameters (method.GenericParameters, null, method.GenericParameters.Count, sb);
 			}
 
-			builder.Append (")");
+			// Insert method name
```
vitek-karas, line 31: Why do you prepend the name at the end, why not simply append it before the parameters?

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -12,46 +13,94 @@ public static string GetDisplayName (this TypeReference type)
...
+							sb.Insert (0, type.Name);
+							break;
+						}
+
+						ParseGenericParameters (type.GenericParameters.TakeLast (genericParameters).ToList (), genericArguments, genericParameters, sb);
+						sb.Insert (0, type.Name.Substring (0, type.Name.IndexOf ('`')));
+						break;
+					}
+
+					sb.Insert (0, type.Name);
```
vitek-karas, line 41: Same here - why prepend... this can easily be done by appending, right?
(The reason is that Insert is inherently slower as it has to move the existing content first).

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -12,46 +13,94 @@ public static string GetDisplayName (this TypeReference type)
...
+					break;
+
+				sb.Insert (0, '.');
+			}
+
+			return sb.ToString ();
+		}
 
-				builder.Append ($"{genericInstanceType.GenericArguments[genericInstanceType.GenericArguments.Count - 1].GetDisplayName ()}>");
+		internal static void ParseGenericParameters (IList<GenericParameter> genericParameters, Stack<TypeReference> genericArguments, int arguments, StringBuilder sb)
```
vitek-karas, line 55: I don't understand the naming - "Parse" typically stands for parsing string into OM. This is the other way round. Why not `WriteGenericParameters` ?

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -12,46 +13,94 @@ public static string GetDisplayName (this TypeReference type)
...
+			var sb = new StringBuilder ();
+			Stack<TypeReference> genericArguments = null;
+			while (true) {
+				switch (type) {
+				case ArrayType arrayType:
+					ParseArrayType (arrayType, sb);
+					break;
+				case GenericInstanceType genericInstanceType:
+					int arguments = int.Parse (genericInstanceType.Name.Substring (genericInstanceType.Name.IndexOf ('`') + 1));
+					genericArguments = new Stack<TypeReference> (genericInstanceType.GenericArguments);
```
vitek-karas, line 25: I must admit I don't really understand why we're doing this. The number of generic arguments can be determined by looking at the `GenericArguments` collection. And why do we need to allocate stack to reverse the order, only to then use prepend semantics... it feels backwards to me.

PR #1252, opened 2020-06-08
Add SetupLinkerDescriptorFile 

tlakollo:
Added SetupLinkerDescriptorFile option in the testing infrastructure to differentiate Descriptor XML from substitution and attribute XML. 
All the file names ending in .xml were treated by the testing infrastructure as a descriptor xml using the flag -x. This generated in some cases error codes from ResolveXmlStep in substitution and attribute definition xml tests or will even fail some test because ResolveFromXmlStep doesn't support the formats of the other xml files.


tlakollo:
cc @mateoatr @sbomer @vitek-karas

test/Mono.Linker.Tests.Cases/Substitutions/FeatureSubstitutionsInvalid.xml
```diff
@@ -6,12 +6,8 @@
         <method signature="System.Boolean BooleanFeatureMethod()" body="stub" value="false" feature="BooleanFeature" featurevalue="true" />
       </type>
       <type fullname="Mono.Linker.Tests.Cases.Substitutions.FeatureSubstitutionsInvalid/Foo">
-        <field signature="NonExistentField" />
+        <field name="NonExistentField" />
         <method signature="NonExistentMethod" />
-        <event signature="NonExistentEvent" />
```
vitek-karas, line 11: Why did you remove these? They seem like good tests to have.

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/NoInstanceCtor/NoInstanceCtorAndTypePreserveMethodsWithInterfacesMarked.cs
```diff
@@ -23,7 +23,8 @@ namespace Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType.NoInsta
 	[KeptMemberInAssembly ("library",
 		"Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType.NoInstanceCtor.Dependencies.NoInstanceCtorAndAssemblyPreserveAll_Lib/A",
 		"Bar()")]
-	public class NoInstanceCtorAndTypePreserveMethodsWithInterfacesMarked
+	[SetupLinkerDescriptorFile ("NoInstanceCtorAndTypePreserveMetheodsWithInterfacesMarked.xml")]
+	public class NoInstanceCtorAndTypePreserveMetheodsWithInterfacesMarked
```
mateoatr, line 27: `NoInstanceCtorAndTypePreserve`**Metheods**`WithInterfacesMarked`

test/Mono.Linker.Tests.Cases/Mono.Linker.Tests.Cases.csproj
```diff
@@ -38,84 +38,14 @@
     <Content Include="TypeForwarding\Dependencies\TypeForwarderMissingReference.il" />
     <Content Include="Attributes\OnlyKeepUsed\Dependencies\AssemblyWithUnusedAttributeOnReturnParameterDefinition.il" />
     <Content Include="Attributes\OnlyKeepUsed\Dependencies\AssemblyWithUnusedAttributeOnGenericParameter.il" />
-    <Content Include="Attributes\OnlyKeepUsed\UnusedAttributePreservedViaLinkXmlIsKept.xml" />
```
mateoatr, line 41: Why do we remove all these xml files?

test/Mono.Linker.Tests.Cases/Substitutions/FeatureSubstitutionsInvalid.xml
tlakollo: BodySubstituterStep does not have any code to handle events, properties nor field signatures. All of these were just warning because of the ResolveXmlStep

vitek-karas: I see - makes sense.

test/Mono.Linker.Tests.Cases/Mono.Linker.Tests.Cases.csproj
tlakollo: They seemed to be necessary only to tell the linker which XML's were available. The BodySubstituter and the CustomAttributes steps don't include their xml's in the Test.Cases.csproj and the test now seemed to not use them

tlakollo:
I also noticed that:
PreserveBackingFieldWhenPropertyIsKept
UnusedNonRequiredTypeIsRemoved
Pass with or without the XML file so not sure if they are valuable tests

vitek-karas:
I think the one for PreserveBackingFieldWhenPropertyIsKept can be removed - linker should figure it out on its own.
The other one is I think important to keep as it tests that the XML will only kick in if something else brings in the type - in this case it doesn't so the XML should have no effect.

PR #1255, opened 2020-06-08
Warnings during propagation of data flow annotations in properties

vitek-karas:
When we propagate data flow annotations from the property to its getter/setter/backing field there are cases which the code doesn't handle. This change will issue warnings for most of those cases:
- If we can't detect backing field
- If there already is a data flow annotation in a place where we would try to apply it

Added tests for these cases

### DIFF
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -206,13 +206,14 @@ namespace Mono.Linker.Dataflow
 					MethodDefinition setMethod = property.SetMethod;
 					if (setMethod != null) {
 
-						// TODO: Handle abstract properties - no way to propagate the annotation to the field
+						// TODO: Handle abstract properties - can't do any propagation on the base property, should rely on validation
+						//  that overrides also correctly annotate.
 						if (!setMethod.HasBody || !ScanMethodBodyForFieldAccess (setMethod.Body, write: true, out backingFieldFromSetter)) {
-							// TODO: warn we couldn't find a unique backing field
+							_context.LogWarning ($"Could not find a unique backing field for property '{property.FullName}' to propagate DynamicallyAccessedMembersAttribute. The property setter is either abstract or not a compiler generated setter.", 2042, property);
 						}
 
 						if (annotatedMethods.Any (a => a.Method == setMethod)) {
-							// TODO: warn: duplicate annotation. not propagating.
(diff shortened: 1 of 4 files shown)


vitek-karas:
Handling of abstract properties requires the overall handling of overrides/virtual methods. So I left that as is for now.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -206,13 +206,14 @@ TypeAnnotations BuildTypeAnnotations (TypeDefinition type)
 					MethodDefinition setMethod = property.SetMethod;
 					if (setMethod != null) {
 
-						// TODO: Handle abstract properties - no way to propagate the annotation to the field
+						// TODO: Handle abstract properties - can't do any propagation on the base property, should rely on validation
+						//  that overrides also correctly annotate.
 						if (!setMethod.HasBody || !ScanMethodBodyForFieldAccess (setMethod.Body, write: true, out backingFieldFromSetter)) {
-							// TODO: warn we couldn't find a unique backing field
+							_context.LogWarning ($"Could not find a unique backing field for property '{property.FullName}' to propagate DynamicallyAccessedMembersAttribute. The property setter is either abstract or not a compiler generated setter.", 2042, property);
```
MichalStrehovsky, line 212: I was thinking about this in the context of the other pull request in the runtime repo.

Would it make sense to only do propagation to the field if this is an auto property, and simply don't deal with fields otherwise?

I'm thinking that if I have:

```csharp
class Foo
{
    private Type _bar;

    [DynamicallyAccessed(XYZ)]
    public Type Bar => _bar;
}
```

It feels like it would be a pretty okay user experience to just get the usual warning that says that the return parameter of `get_Bar` requires XYZ, but field `_bar` is not annotated. The fix for that would be to just annotate `_bar`

The user experience as it is in this pull request would basically require people to remove the annotation from the property, move it to the getter's return parameter, and then annotate `_bar`, which feels fairly non-obvious.

vitek-karas: We already only handle compiler generated backing fields, but I guess your suggestion is to not warn about "unrecognized field" in the above case, and instead warn about "unannotated field". Right?
I think I like that - the "unrecognized field" is somewhat weird as it doesn't really tell the user what to do (and saying to explicitly annotate getters and field is not nice).

MichalStrehovsky: Yup

vitek-karas: OK - I changed it. It's a bit weird for properties with setter only - in that case there's nothing in the property itself which will tell the user to annotate the field (since just assigning to it is not a problem as it has no requirements). But if it was used somewhere, that would warn and also it's probably pretty unlikely to have a setter only property of type `Type`.

Issue #1257, opened 2020-06-09
Allow embedded linker descriptors to be called ILLink.Descriptors.xml


vitek-karas:
Currently linker only recognizes embedded descriptors if they're called `AssemblyName.xml`. This is rather confusing as there's nothing in the name of the resource which describes what it's for.

We should also recognize embedded resources called `ILLink.Descriptors.xml` (no assembly name in them). This will also unify the handling with existing `ILLink.Substitutions.xml` and the coming `ILLink.LinkAttributes.xml`.

It's probably not worth the trouble to try to remove the existing behavior, so linker should still recognize `AssemblyName.xml` as well.

vitek-karas:
@eerhardt once implemented we should try to switch all descriptors in the framework (dotnet/runtime) to this new resource name.

eerhardt:
Opened https://github.com/dotnet/runtime/issues/37651 to update dotnet/runtime.

PR #1258, opened 2020-06-09
Support ILLink.Descriptors.xml

sbomer:
Fixes https://github.com/mono/linker/issues/1257


sbomer:
Updated docs. I have removed the documentation for `assemblyname.xml`.

Issue #1221, opened 2020-05-28
Call `TryGetOrigin` method only when needed 


mateoatr:
Currently, `TryGetOrigin` is called every time there is a call to `CreateWarningMessage` where we are able to retrieve information of the method (and possibly line and column within the file where this method is in the source code) where something went wrong, e.g.:
```C#
_context.LogMessage (MessageContainer.CreateWarningMessage (_context, message, 2026, origin: MessageOrigin.TryGetOrigin (callingMethodBody.Method, operation.Offset)));
```
Because the warning can be suppressed by the user by using the `UnconditionalSuppressMessage` attribute, and hence no message will be logged, we should call this method only when necessary, i.e., after we make sure sure that the warning message will be logged, thus saving us some time.

PR #1261, opened 2020-06-09
Remove TryGetOrigin from MessageOrigin

mateoatr:
I'm moving the optional `ilOffset` parameter that was originally passed to `TryGetOrigin` to `LogWarning`, so that we can delay the retrieval of the source information for warnings until it's needed (i.e., after we're sure that the warning isn't suppressed).

Solves https://github.com/mono/linker/issues/1221

### DIFF
--- a/src/linker/Linker/LoggingReflectionPatternRecorder.cs
+++ b/src/linker/Linker/LoggingReflectionPatternRecorder.cs
@@ -44,10 +44,8 @@ namespace Mono.Linker
 
 		public void UnrecognizedReflectionAccessPattern (IMemberDefinition source, Instruction sourceInstruction, IMetadataTokenProvider accessedItem, string message)
 		{
-			string location = string.Empty;
-			var method = source as MethodDefinition;
-			var origin = new MessageOrigin (method, sourceInstruction?.Offset);
-			_context.LogWarning (location + message, 2006, origin, "Unrecognized reflection pattern");
+			var origin = new MessageOrigin (source, sourceInstruction?.Offset);
+			_context.LogWarning (message, 2006, origin, "Unrecognized reflection pattern");
 		}
 	}
 }
--- a/src/linker/Linker/LoggingReflectionPatternRecorder.cs
+++ b/src/linker/Linker/LoggingReflectionPatternRecorder.cs
@@ -44,14 +44,10 @@ namespace Mono.Linker
 
 		public void UnrecognizedReflectionAccessPattern (IMemberDefinition source, Instruction sourceInstruction, IMetadataTokenProvider accessedItem, string message)
 		{
-			MessageOrigin origin;
 			string location = string.Empty;
 			var method = source as MethodDefinition;
-			if (sourceInstruction != null && method != null)
-				origin = MessageOrigin.TryGetOrigin (method, sourceInstruction.Offset);
-			else
-				origin = new MessageOrigin (source);
-
+			var origin = new MessageOrigin (method, sourceInstruction?.Offset);
+			origin.TryGetSourceInfo ();
 			if (origin.FileName == null) {
 				if (method != null)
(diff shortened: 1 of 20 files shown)


src/linker/Linker/LoggingReflectionPatternRecorder.cs
```diff
@@ -44,50 +44,10 @@ public void RecognizedReflectionAccessPattern (IMemberDefinition source, Instruc
 
 		public void UnrecognizedReflectionAccessPattern (IMemberDefinition source, Instruction sourceInstruction, IMetadataTokenProvider accessedItem, string message)
 		{
-			MessageOrigin origin;
 			string location = string.Empty;
```
vitek-karas, line 47: location is effectively unused now.

src/linker/Linker/MessageContainer.cs
```diff
@@ -70,6 +70,12 @@ public static MessageContainer CreateWarningMessage (LinkContext context, string
 			if (context.IsWarningSuppressed (code, origin))
 				return Empty;
 
+			origin.TryGetSourceInfo ();
+			if (subcategory == "Unrecognized reflection pattern" && origin.FileName == null) {
```
vitek-karas, line 74: This should probably happen for every warning. If we can't figure out the filename but we know which member it belongs to, we should report the member. This is not specific to the reflection patterns.

src/linker/Linker/MessageContainer.cs
```diff
@@ -70,6 +70,12 @@ public static MessageContainer CreateWarningMessage (LinkContext context, string
 			if (context.IsWarningSuppressed (code, origin))
 				return Empty;
 
+			origin.TryGetSourceInfo ();
```
sbomer, line 73: Should we call `TryGetSourceInfo` in `CreateErrorMessage` too?

src/linker/Linker/MessageContainer.cs
```diff
@@ -49,6 +49,7 @@ public static MessageContainer CreateErrorMessage (string text, int code, string
 			if (!(code >= 1000 && code <= 2000))
 				throw new ArgumentException ($"The provided code '{code}' does not fall into the error category, which is in the range of 1000 to 2000 (inclusive).");
 
+			origin?.TryGetSourceInfo ();
```
marek-safar, line 52: This is an odd code flow, you are changing the content of struct from a constructor of container-like object. The logic is relevant to MessageOrigin::ToString only and I think it'd be better to do it there only. This would allow change MessageOrigin to be readonly and ensure that logic in methods like MessageOrigin::Equals work all the time

vitek-karas: I agree with @marek-safar here. Currently we only use the info from `TryGetSource` in two places - in the warning case below where we prepend the location from member if there's no file name and then in the ToString.
I think we should unify this:
* Same behavior between error and warning around "origin/location"
* Remove `TryGetSourceInfo` and make the struct `readonly`
* In `ToString` if we have `FileName` keep the current behavior. If we don't have `FileName` use the "member" to describe the location.

Note: I think it's OK for the linker to report warnings which don't have the typical `filename(line)` information since we're dealing with IL input. That said in most cases we know where the error/warning comes from, so we should report it. Unfortunately we have to "invent" our own syntax for this, but that's also OK.

vitek-karas:
> I'm a little concerned that putting the actual warning text at the very end makes it harder to read. My preference would be for the warning text to come before the member signature.

I'm also split on this, but I'm leaning the other way for now. Mainly because for me the location is "more" important than the warning - at least that's how I personally work. But it's subjective.
The other consideration is "sorting"/"categorization" which I personally prefer to go by location - we don't have a good categorization of the reflection warnings, so they all tend to look the same.

I think right now it doesn't matter - let's try one and we'll see how it works - as in we need to do end-2-end validation soon anyway - "How does it look in VS kind of thing" and we'll go from there.

PR #1262, opened 2020-06-10
Match DynamicDependency members by name

sbomer:
The signature parser allows a name to match different member types (method/field, etc.) with the same string, and it also allows specifying a parameterless method by name alone. This change also allows it to match methods/properties with any number of parameters by name alone.

I thought this was already the case, but didn't notice that the parser still checks parameters when it doesn't see parentheses, and didn't hit it when testing.

I noticed this while turning it on for dotnet/runtime. There we use it to pull in `GetSourceLineInfo` by name: https://github.com/dotnet/runtime/blob/144e5145453ac3885ac20bc1f1f2641523c6fcea/src/coreclr/src/System.Private.CoreLib/src/System/Diagnostics/StackFrameHelper.cs#L91.

`GetSourceLineInfo` has a lot of parameters, so we probably wouldn't want to put the full signature in the attribute: https://github.com/dotnet/runtime/blob/144e5145453ac3885ac20bc1f1f2641523c6fcea/src/libraries/System.Diagnostics.StackTrace/src/System/Diagnostics/StackTraceSymbols.cs#L53.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -613,7 +613,7 @@ namespace Mono.Linker.Steps
 
 			IEnumerable<IMemberDefinition> members;
 			if (dynamicDependency.MemberSignature is string memberSignature) {
-				members = DocumentationSignatureParser.GetMembersByDocumentationSignature (type, memberSignature);
+				members = DocumentationSignatureParser.GetMembersByDocumentationSignature (type, memberSignature, acceptName: true);
 				if (!members.Any ()) {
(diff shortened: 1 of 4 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -613,7 +613,7 @@ void MarkDynamicDependency (DynamicDependency dynamicDependency, MemberReference
 
 			IEnumerable<IMemberDefinition> members;
 			if (dynamicDependency.MemberSignature is string memberSignature) {
-				members = DocumentationSignatureParser.GetMembersByDocumentationSignature (type, memberSignature);
+				members = DocumentationSignatureParser.GetMembersByDocumentationSignature (type, memberSignature, acceptName: true);
```
marek-safar, line 616: So this is a light version of wildcard support, right? It will mark any members if `(` is not part of the signature.

Issue #1206, opened 2020-05-21
[Tracking] Document usage of new linker attributes


vitek-karas:
Create detailed documentation of the new linker attributes and their usage:
* `DynamicallyAccessedMemnbers`
* `RequiresUnreferencedCode`
* `DynamicDependency`
* `UnconditionalSuppressMessage`

There should be a high-level guide describing the problem space, and the general ideas behind the new attributes and related linker capabilities. Then a list of sample issues with the generated warnings and recommended solutions.

vitek-karas:
The main documentation for all of these is done. Various improvements can be tracked by other issues.

PR #1265, opened 2020-06-10
Document DynamicDependencyAttribute

sbomer:
And add a few more tests to match the examples.

I am wondering about https://github.com/mono/linker/blob/master/docs/design/reflection-flow.md - is that supposed to preserve the original design doc, or should we be keeping it up-to-date with the new attributes?

Contributes to https://github.com/mono/linker/issues/1206

### DIFF
--- a/docs/linker-attributes.md
+++ b/docs/linker-attributes.md
@@ -1,0 +1,78 @@
+# Linker attributes
+
+As the linker scans the referenced code, it warns about non-understood reflection patterns or other code patterns whose behavior may not be preserved by linking. The following attributes influence the pattern-detection and warning behavior.
+
+## DynamicallyAccessedMembers
+
+Gives extra information about dynamic access patterns which would not otherwise be understood. This can tell the linker to keep additional members that are reflected over, making simple reflection patterns linker-friendly.
+
+TODO: provide examples
+
+The attribute specifies which members of a type are dynamically accessed (e.g. by reflection or interop code). It can be applied to IL locations (method parameters, method return parameters, fields, and properties) that hold a representation of a type, and it indicates via DynamicallyAccessedMemberTypes which members of that type are used dynamically. The linker uses information from these attributes to keep additional members of types, which can make some reflection access patterns linker-safe, eliminating the corresponding warnings.
+
+See [Cross-method annotations](design/reflection-flow.md#Cross-method-annotations) for more information.
+
+## RequiresUnreferencedCode
+
+Specifies that a method requires code which is not statically referenced, preventing analysis of the method, and warning for calls to it.
+
+TODO: provide examples
+
+When annotating linker-unfriendly code, this attribute can be used to prevent multiple warnings originating from reflection patterns in the same method body, instead turning them into a single warning with a message that explains why the method is conceptually linker-unsafe. Calls to methods with this attribute will result in a single warning for the callsite.
+
+## DynamicDependency
+
+Indicates that a member has a dependency on other members. This results in the referenced members being kept whenever the member with the attribute is kept. It does not otherwise influence the analysis behavior.
+
+```csharp
+using System;
+using System.Reflection;
+using System.Diagnostics.CodeAnalysis;
+
+class Program
+{
+    [DynamicDependency("Helper")]
+    public static void Main()
+    {
+        var helper = Assembly.GetExecutingAssembly().GetType("Program").GetMethod("Helper");
+        helper.Invoke(null, null);
+    }
+
+    public static void Helper()
+    {
+        Console.WriteLine("Dynamic call to helper");
+    }
+}
+```
+
+Without `DynamicDependency`, linking this assembly would remove `Helper`, causing a failure at runtime. The attribute ensures that `Helper` is kept.
+
+The attribute specifies the members to keep via a `string` or via `DynamicallyAccessedMemberTypes`. The type and assembly are either implicit in the attribute context, or explicitly specified in the attribute (by `Type`, or by `string`s for the type and assembly name).
+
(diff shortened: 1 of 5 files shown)


docs/linker-attributes.md
```diff
@@ -0,0 +1,78 @@
+# Linker attributes
+
+As the linker scans the referenced code, it warns about non-understood reflection patterns or other code patterns whose behavior may not be preserved by linking. The following attributes influence the pattern-detection and warning behavior.
```
marek-safar, line 3: I think this should be a general intro about where the attributes are and how to use them. It'd not pick reflection only summary here at all

docs/linker-attributes.md
```diff
@@ -0,0 +1,78 @@
...
+Indicates that a member has a dependency on other members. This results in the referenced members being kept whenever the member with the attribute is kept. It does not otherwise influence the analysis behavior.
+
+```csharp
+using System;
+using System.Reflection;
+using System.Diagnostics.CodeAnalysis;
+
+class Program
+{
+    [DynamicDependency("Helper")]
```
marek-safar, line 34: Id' be better to use some tricky code which linker cannot automatically detect as the example

docs/linker-attributes.md
```diff
@@ -0,0 +1,78 @@
+# Linker attributes
```
marek-safar, line 1:
```suggestion
# Linker Specific Custom Attributes
```

docs/linker-attributes.md
```diff
@@ -0,0 +1,78 @@
...
+
+As the linker scans the referenced code, it warns about non-understood reflection patterns or other code patterns whose behavior may not be preserved by linking. The following attributes influence the pattern-detection and warning behavior.
+
+## DynamicallyAccessedMembers
```
marek-safar, line 5: Not sure if we should not use full attribute name

docs/linker-attributes.md
```diff
@@ -0,0 +1,78 @@
...
+
+Gives extra information about dynamic access patterns which would not otherwise be understood. This can tell the linker to keep additional members that are reflected over, making simple reflection patterns linker-friendly.
```
marek-safar, line 7: Can we add MSDN link to the documentation page (even if it does not exist yet)

docs/linker-attributes.md
```diff
@@ -0,0 +1,78 @@
...
+
+TODO: provide examples
+
+The attribute specifies which members of a type are dynamically accessed (e.g. by reflection or interop code). It can be applied to IL locations (method parameters, method return parameters, fields, and properties) that hold a representation of a type, and it indicates via DynamicallyAccessedMemberTypes which members of that type are used dynamically. The linker uses information from these attributes to keep additional members of types, which can make some reflection access patterns linker-safe, eliminating the corresponding warnings.
```
marek-safar, line 11: Can we include DynamicallyAccessedMemberTypes enum here with documentation as well

docs/linker-attributes.md
vitek-karas: Linker will not figure out the sample below - we don't handle `GetExecutingAssembly` - but I guess it "looks obvious" - so maybe something less obvious...

docs/linker-attributes.md
```diff
@@ -0,0 +1,78 @@
...
+[DynamicDependency("MethodWithGenericParameterTypes(System.Collections.Generic.List{System.String})")]
+[DynamicDependency("MethodOnGenericType(`0)", "GenericType`1", "UnreferencedAssembly")]
+[DynamicDependency("MethodOnGenericType(`0)", typeof(GenericType<>))]
+```
+
+This attribute will most often be used in cases where a method contains reflection patterns that can not be analyzed even with the help of DynamicallyAccessedMembers. In these cases, the linker will warn about non-understood patterns in the method body. Applying DynamicDependencyAttribute to the method allows the dynamically referenced members to be kept. The warnings will still show up unless silenced by another mechanism (see UnconditianalSuppressMessage).
+
+## UnconditionalSuppressMessage
+
+Silences linker warnings originating from a method.
```
vitek-karas, line 74: Not just "method" - maybe "member" is safer (even though it's not precise either, but probably good enough)

docs/linker-attributes.md
```diff
@@ -0,0 +1,78 @@
...
+
+TODO: provide examples
+
+In some cases, the linker may warn about unrecognized reflection access patterns which the developer knows to be safe. For example, the accessed members may be referenced elsewhere (such as by DynamicDependency), or they may have safe fallback behavior when the reflection targets are trimmed by the linker. This attribute can be used to suppress messages about such warnings. It can also suppress other linker warnings that are unrelated to the analysis.
```
vitek-karas, line 78: Don't use "safe" and/or "unsafe" - it has too much "security" feeling to it. Maybe "trim friendly" or "trim correct" and so on...

MichalStrehovsky:
> I am wondering about https://github.com/mono/linker/blob/master/docs/design/reflection-flow.md - is that supposed to preserve the original design doc, or should we be keeping it up-to-date with the new attributes?

I've updated the dataflow annotations there after the API review. Stale documents cause confusion.

vitek-karas:
> I am wondering about https://github.com/mono/linker/blob/master/docs/design/reflection-flow.md - is that supposed to preserve the original design doc, or should we be keeping it up-to-date with the new attributes?

+1 to @MichalStrehovsky response. There's little reason to keep around documents describing behavior we changed during implementation (we have version history for that). What we should try to keep as part of the design docs is the reasoning behind some of the decisions - that is valuable info.

sbomer:
This was documented in dotnet/docs instead.

PR #1274, opened 2020-06-16
Allow inlining of return values of instance methods which are explicitly substituted

vitek-karas:
So far we only allowed inlining of return values for methods which are static, with this change we also allow it on an instance method which is explicitly substituted.


vitek-karas:
/cc @eerhardt 

I verified that it works when applied to `EventSource.IsEnabled` - the callsites get the branch body removed - the end result being that lot of the "WriteEvent" methods are trimmed away.

Issue #1268, opened 2020-06-11
Support feature conditions on XML descriptors


sbomer:
This would allow us to use the embedded ILLink.Descriptors.xml to root "global" dependencies for which we don't have a great place to put `DynamicDependencyAttribute`, while allowing them to be linked out using feature settings. From discussion with @vitek-karas, it sounds like this will be useful for removing debugger support, COM, etc.

@marek-safar what do you think of this?

vitek-karas:
I'm all for doing this 😄

marek-safar:
I agree that this would be a useful extension of XML descriptors.

eerhardt:
If we take this approach, we should be able to close https://github.com/dotnet/runtime/issues/37352, right?

The motivation behind https://github.com/dotnet/runtime/issues/37352 was being able to "root" types globally, but also conditionally. See the discussion here: https://github.com/dotnet/runtime/pull/37288#discussion_r434018579.

So if there are scenarios where we need to preserve a type globally under certain conditions (Debug, COM, JS, etc), we can use the XML descriptor + feature switch mechanism, instead of putting a `[DynamicDependency("foo", Condition = "feature switch")]`.

Note that the advantage of having `[DynamicDependency]` on the Type vs having a global XML descriptor, is that the global XML descriptor will keep the dependency even if the `Type` isn't being used. With the `[DynamicDependency]` on the Type approach, if the whole Type could be trimmed, so could its dependencies.

marek-safar:
> So if there are scenarios where we need to preserve a type globally under certain conditions (Debug, COM, JS, etc), we can use the XML descriptor + feature switch mechanism, instead of putting a [DynamicDependency("foo", Condition = "feature switch")].

This scenario should be used only for cases where the source of the dependency is outside of src/libraries (e.g. external code, different runtime, etc) for anything else `DynamicDependency` is the preference

> global XML descriptor will keep the dependency even if the Type isn't being used.

That's not true. See https://github.com/mono/linker/blob/master/docs/data-formats.md#preserve-a-type

PR #1277, opened 2020-06-16
Support feature settings on descriptor files

sbomer:
Fixes https://github.com/mono/linker/issues/1268.

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -9,11 +9,11 @@ error and warning codes.
 
 #### `IL1001`: Failed to process 'XML document location'. Feature 'feature' does not specify a "featurevalue" attribute
 
-- The substitution in 'XML document location' with feature value 'feature' does not use the `featurevalue` attribute. These attributes have to be used together.
+- The substitution or descriptor in 'XML document location' with feature value 'feature' does not use the `featurevalue` attribute. These attributes have to be used together.
 
 #### `IL1002`: Failed to process 'XML document location'. Unsupported non-boolean feature definition 'feature'
 
-- The substitution in 'XML document location' with feature value 'feature' sets the attribute `featurevalue` to a non-boolean value. Only boolean values are supported for this attribute.
+- The substitution or descriptor in 'XML document location' with feature value 'feature' sets the attribute `featurevalue` to a non-boolean value. Only boolean values are supported for this attribute.
 
 #### `IL1003`: Error processing 'XML document name': 'XmlException'
 
@@ -54,6 +54,10 @@ error and warning codes.
 #### `IL1013`: Error processing 'XML document location': 'XmlException'
 
 - There was an error processing 'XML document location' xml file. The most likely reason for this is that the descriptor file has syntactical errors.
+
+#### `IL1014`: Failed to process 'XML document location`. Unsupported value for featuredefault attribute
+
+- The substitution or descriptor in 'XML document location' contains a 'featuredefault' attribute with an invalid value. This attribute only supports the true value, to indicate that this is the default behavior for a feature when a value is not given.
 
 ----
 ## Warning Codes
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -7,11 +7,11 @@ error and warning codes.
 
 ## Error Codes
 
-#### `IL1001`: Failed to process XML substitution: 'XML document location'. Feature 'feature' does not specify a "featurevalue" attribute
+#### `IL1001`: Failed to process 'XML document location'. Feature 'feature' does not specify a "featurevalue" attribute
 
 - The substitution in 'XML document location' with feature value 'feature' does not use the `featurevalue` attribute. These attributes have to be used together.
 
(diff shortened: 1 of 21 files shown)


src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -86,17 +89,26 @@ protected override void Process ()
...
 				throw new LinkerFatalErrorException (MessageContainer.CreateErrorMessage ($"Failed to process description file '{_xmlDocumentLocation}'", 1004), ex);
 			}
 		}
 
-		protected virtual void ProcessAssemblies (LinkContext context, XPathNodeIterator iterator)
+		protected virtual void ProcessAssemblies (XPathNodeIterator iterator)
 		{
 			while (iterator.MoveNext ()) {
-				AssemblyDefinition assembly = GetAssembly (context, GetAssemblyName (iterator.Current));
+				var name = GetAssemblyName (iterator.Current);
```
vitek-karas, line 106: Nit: I don't understand why we need to call GetAssemblyName before checking for feature switches... it will work, just weird that you're changing it...

test/Mono.Linker.Tests.Cases/FeatureSettings/Dependencies/ResourceFile.txt
```diff
@@ -0,0 +1 @@
+This resource will be embedded into an assembly and then removed by the linker.
```
vitek-karas, line 1: Is this file actually used anywhere (I probably missed it)

vitek-karas:
@tlakollo Could you please double check that we don't "return" from the attribute XML parser in places where we should "continue" instead? Also would be nice to add a test to validate at least one of those cases.

test/Mono.Linker.Tests.Cases/FeatureSettings/Dependencies/ResourceFile.txt
sbomer: Yeah - it's used in https://github.com/mono/linker/pull/1277/files#diff-7158670bf7e2f9050db3d3309f790831R7 since I moved the feature switch tests to a new directory. The substitutions tests have their own copy which is why this isn't just a rename like the others.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
sbomer: Parsing a null/empty assemblyname in AssemblyNameReference.Parse can throw an exception - I wanted that error not to be hidden by the feature switches.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -86,17 +89,30 @@ protected override void Process ()
 				if (Context.IgnoreDescriptors)
 					return;
 			}
+
+#if FEATURE_ILLINK
```
marek-safar, line 93: Could you if-def the implementation instead of many `#if FEATURE_ILLINK` for ShouldProcessElement

src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -720,5 +768,28 @@ public override string ToString ()
 		{
 			return "ResolveFromXmlStep: " + _xmlDocumentLocation;
 		}
+
+		public static bool ShouldProcessElement (XPathNavigator nav, LinkContext context, string documentLocation)
```
marek-safar, line 772: Ideally, this should be in a different class

sbomer:
@eerhardt I have added support for `featuredefault="true"` - PTAL

src/linker/Linker.Steps/ResolveFromXmlStep.cs
vitek-karas: OK - maybe add a comment so that it's not moved later on.

docs/error-codes.md
```diff
@@ -55,6 +55,10 @@ error and warning codes.
 
 - There was an error processing 'XML document location' xml file. The most likely reason for this is that the descriptor file has syntactical errors.
 
+#### `IL1014`: Failed to process 'XML document location`. Unsupported value for featuredefault attribute
+
+- The substitution or descriptor in 'XML document location' contains a 'featuredefault' attribute with an invalid value. This attribute only supports the true value, to indicate that this is the default behavior for a feature when a value is not given.
```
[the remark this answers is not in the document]
sbomer: I don't see why we would want to add `featuredefault="false"` to all of the substitutions that apply only when a feature setting is passed - which I think is most of them. I have no strong opposition to supporting it if you would prefer that.

vitek-karas:
> What happens in the case where a Substitutions.xml file doesn't specify a featuredefault, but a Descriptors.xml does for the same feature switch?

Nothing interesting 😉 . Each element in the XML is independent. So if you want to have the same behavior between substitutions and descriptors, you'll need to specify the same thing twice.

sbomer:
> What happens...

Sorry, I missed this question - thanks for answering it @vitek-karas

PR #1279, opened 2020-06-17
Fix global suppressions that target members

mateoatr:
When a module/assembly level suppression specifies a target, the provider attached to the suppression should be the member resolved by the signature parser and not the module.

Use the linker attributes cache to look for `UnconditionalSuppressMessage` attributes.

Use the initialized assemblies hash whenever a global suppression is found.


src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -26,6 +25,8 @@ public LinkerAttributesInformation (LinkContext context, ICustomAttributeProvide
 						attributeValue = ProcessRequiresUnreferencedCodeAttribute (context, provider, customAttribute);
 					else if (attributeType.IsTypeOf<DynamicDependencyAttribute> ())
 						attributeValue = DynamicDependency.ProcessAttribute (context, provider, customAttribute);
+					else if (attributeType.IsTypeOf<UnconditionalSuppressMessageAttribute> ())
```
vitek-karas, line 28: This feels VERY fragile:
* The warning suppression system doesn't call attribute cache anywhere, so it relies on others calling it - on the right items - this is... well fragile to say the least
* With this change the attribute is now stored in two caches - once in the warning suppression system and once in the linker attribute cache. The linker attribute cache version is not used for anything - it's basically wasted memory right now.

I don't really understand the reason for making such change. I know it was fragile before since we relied on correct ordering in MarkStep, but that was way less fragile than the new approach which relies on others calling the cache (also from within the MarkStep).

Given that we maintain a separate cache in the warning suppression system, there's no reason to use the generic attribute cache again - so I would not do that.

If we want to improve on the fragile behavior I would modify the warning suppression system to explicitly lookup attributes as it walks the OM from the member where warning occurred to the root. It already caches info about these, so we can rely on the cache to only do it once.

mateoatr: Using the linker attributes cache for all linker-related attributes seemed reasonable to me (although you're right, it's actually not being used for anything here). I'm undoing this change. I'll modify the warning suppression mechanism in a later PR to store the CAs that are seen when walking up the OM from a member.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -138,6 +138,7 @@ private void DecodeModuleLevelAndGlobalSuppressMessageAttributes (ModuleDefiniti
 		{
 			AssemblyDefinition assembly = module.Assembly;
 			if (!InitializedAssemblies.Contains (assembly)) {
+				InitializedAssemblies.Add (assembly);
```
marek-safar, line 141: No need for Contains when you have Add (just check the return value)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1530,6 +1530,7 @@ bool MarkSpecialCustomAttributeDependencies (CustomAttribute ca, ICustomAttribut
 				return true;
 			} else if (UnconditionalSuppressMessageAttributeState.TypeRefHasUnconditionalSuppressions (dt)) {
 				_context.Suppressions.AddSuppression (ca, provider);
+				return true;
```
marek-safar, line 1533: Why is there no test for this?

PR #1280, opened 2020-06-17
Add option to generate baselines

mateoatr:
This adds a new option `--generate-warning-suppressions` that outputs a `<AssemblyName>.WarningSuppressions.cs` file which includes the required `UnconditionalSuppressMessage` attributes to shut down the warnings that were output during the linking for each of the assemblies linked. We currently don't support `resource` suppressions (I'm working on adding it, so that we can suppress warnings like `IL2025`).

/cc @eerhardt


src/linker/Linker/WarningSuppressor.cs
```diff
@@ -0,0 +1,61 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	public class WarningSuppressor
```
vitek-karas, line 8: Nit: I don't like the name - this in itself doesn't suppress warnings. Maybe `WarningSuppressionWriter`.

src/linker/Linker/MessageContainer.cs
```diff
@@ -67,6 +67,9 @@ public static MessageContainer CreateWarningMessage (LinkContext context, string
 			if (!(code > 2000 && code <= 6000))
 				throw new ArgumentException ($"The provided code '{code}' does not fall into the warning category, which is in the range of 2001 to 6000 (inclusive).");
 
+			if (context.WarningSuppressor != null && origin.MemberDefinition != null)
```
vitek-karas, line 70: We must only write suppressions for warnings which are actually reported - so this has to go after the `IsWarningSupressed` below.

src/linker/Linker/WarningSuppressor.cs
```diff
@@ -0,0 +1,61 @@
...
+
+		public void AddWarning ((int, IMemberDefinition) warning)
+		{
+			_warnings.Add (warning);
+		}
+
+		public void OutputSuppressions ()
+		{
+			if (_warnings.Count == 0) {
+				_context.LogMessage ("");
```
vitek-karas, line 29: What is this for?

src/linker/Linker/Driver.cs
```diff
@@ -434,6 +434,12 @@ protected int SetupContext (ILogger customLogger = null)
 
 						continue;
 
+					case "--generate-warning-suppressions":
+						if (!GetStringParam (token, l => context.GenerateWarningSuppressions (l)))
```
vitek-karas, line 438: As is the generated code is "Wrong" - we generate suppressions for all assemblies into one file. But global warning suppressions don't work cross-assembly. So someone would have to go, manually decide which suppression belongs to which assembly (tricky - we only write namespace/type) and split them into their respective assemblies.

I think this should be a simple boolean option and we will write `AssemblyName.WarningSuppressions.cs` for each assembly which has at least one warning to suppress into the output directory.

src/linker/Linker/LinkContext.cs
```diff
@@ -235,6 +237,11 @@ public LinkContext (Pipeline pipeline, AssemblyResolver resolver, ReaderParamete
 			Optimizations = new CodeOptimizationsSettings (defaultOptimizations);
 		}
 
+		public void GenerateWarningSuppressions ()
```
sbomer, line 240: I would rename this so it doesn't sound like it is doing the generation here, and also prevent allocating a new writer in case it is called multiple times.

src/linker/Linker/Driver.cs
```diff
@@ -434,6 +434,10 @@ protected int SetupContext (ILogger customLogger = null)
 
 						continue;
 
+					case "--generate-warning-suppressions":
```
marek-safar, line 437: This needs documentation what is this suppose to produce. Right now it outputs a list of suppression of unsuppressed warnings. Is it intentional that it does not include any global suppression?

src/linker/Linker/LinkContext.cs
```diff
@@ -172,6 +172,8 @@ public class LinkContext : IDisposable
 
 		public KnownMembers MarkedKnownMembers { get; private set; }
 
+		public WarningSuppressionWriter WarningSuppressionWriter { get; private set; }
```
marek-safar, line 175: This should be bool property

src/linker/Linker/Driver.cs
```diff
@@ -434,6 +434,10 @@ protected int SetupContext (ILogger customLogger = null)
...
+						context.GenerateWarningSuppressions ();
```
marek-safar, line 438: This should be bool property

src/linker/Linker/MessageContainer.cs
```diff
@@ -70,6 +70,9 @@ public static MessageContainer CreateWarningMessage (LinkContext context, string
 			if (context.IsWarningSuppressed (code, origin))
 				return Empty;
 
+			if (context.WarningSuppressionWriter != null && origin.MemberDefinition != null)
```
marek-safar, line 73: I think this is an incorrect place, the act of creating the message instance does not mean it will be sent to the output

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -0,0 +1,63 @@
...
+				warnings = new HashSet<(int, IMemberDefinition)> ();
+				_warnings.Add (assemblyName, warnings);
+			}
+
+			warnings.Add (warning);
+		}
+
+		public void OutputSuppressions ()
+		{
+			foreach (var assemblyName in _warnings.Keys) {
```
marek-safar, line 32: As you store it in HashSet you need to sort it to make the output deterministic

src/linker/Linker/LinkContext.cs
```diff
@@ -513,6 +518,9 @@ public void LogWarning (string text, int code, MessageOrigin origin, string subc
 				return;
 
 			var warning = MessageContainer.CreateWarningMessage (this, text, code, origin, subcategory);
+			if (OutputWarningSuppressions && warning != MessageContainer.Empty && origin.MemberDefinition != null)
+				WarningSuppressionWriter.AddWarning ((code, origin.MemberDefinition));
```
vitek-karas, line 522: This should be in the `LogMessage` if somebody calls `CreateWarningMessage` manually and then calls `LogMessage` that warning would be missed.

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -0,0 +1,72 @@
...
+		private readonly LinkContext _context;
+		private readonly Dictionary<AssemblyNameDefinition, HashSet<(int, IMemberDefinition)>> _warnings;
+
+		public WarningSuppressionWriter (LinkContext context)
+		{
+			_context = context;
+			_warnings = new Dictionary<AssemblyNameDefinition, HashSet<(int, IMemberDefinition)>> ();
+		}
+
+		public void AddWarning ((int, IMemberDefinition) warning)
```
vitek-karas, line 20: Nit: Please make the method accept two parameters instead - using tuple here doesn't seem to make much sense.

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Linq;
+using System.Text;
+using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	public class WarningSuppressionWriter
+	{
+		private readonly LinkContext _context;
+		private readonly Dictionary<AssemblyNameDefinition, HashSet<(int, IMemberDefinition)>> _warnings;
```
vitek-karas, line 12:
```suggestion
		private readonly Dictionary<AssemblyNameDefinition, HashSet<(int Code, IMemberDefinition Member)>> _warnings;
```

You can specify names for the items in the tuple - this will make the code using it easier to read.

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -0,0 +1,72 @@
...
+		{
+			foreach (var assemblyName in _warnings.Keys) {
+				using (var sw = new StreamWriter (Path.Combine (_context.OutputDirectory, $"{assemblyName.Name}.WarningSuppressions.cs"))) {
+					StringBuilder sb = new StringBuilder ("using System.Diagnostics.CodeAnalysis;").AppendLine ().AppendLine ();
+					List<(int, IMemberDefinition)> listOfWarnings = _warnings[assemblyName].ToList ();
+					listOfWarnings.Sort ((a, b) => {
+						if (a.Item1 == b.Item1)
+							return a.Item2.Name.CompareTo (b.Item2.Name);
+
+						return a.Item1.CompareTo (b.Item1);
```
vitek-karas, line 41: Personal preference: I think it would make a bit more sense to order by member first - to have all warnings for a given member in one place. Also this would naturally order warnings by namespace and such, which is probably easier to consume.

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -0,0 +1,72 @@
...
+
+		public void OutputSuppressions ()
+		{
+			foreach (var assemblyName in _warnings.Keys) {
+				using (var sw = new StreamWriter (Path.Combine (_context.OutputDirectory, $"{assemblyName.Name}.WarningSuppressions.cs"))) {
+					StringBuilder sb = new StringBuilder ("using System.Diagnostics.CodeAnalysis;").AppendLine ().AppendLine ();
+					List<(int, IMemberDefinition)> listOfWarnings = _warnings[assemblyName].ToList ();
+					listOfWarnings.Sort ((a, b) => {
+						if (a.Item1 == b.Item1)
+							return a.Item2.Name.CompareTo (b.Item2.Name);
```
vitek-karas, line 39: I think we should sort them by `namespace.type.member` - sorting just by name will end up in stable but otherwise seemingly random order.

test/Mono.Linker.Tests/TestCases/Dependencies/WarningSuppressionExpectations1.cs
```diff
@@ -0,0 +1,6 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+[assembly: UnconditionalSuppressMessage ("", "IL2006", Scope = "member", Target = "M:Mono.Linker.Tests.Cases.WarningSuppression.Warnings.NestedType.Warning3")]
```
eerhardt, line 3: Why is the first parameter (the category) blank? This is a required parameter on the attribute, it doesn't feel correct for it to be blank.

For SuppressMessage, we have:

```C#
        [SuppressMessage("Microsoft.Performance", "CA1801:ReviewUnusedParameters", MessageId = "isChecked")]
        [SuppressMessage("Microsoft.Performance", "CA1804:RemoveUnusedLocals", MessageId = "fileIdentifier")]
```

from the docs:

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute.category?view=netcore-3.1#System_Diagnostics_CodeAnalysis_SuppressMessageAttribute_Category

> The Category property describes the tool or tool analysis category for which a message suppression attribute applies.

mateoatr: Is there any preference on what to put here? Would `ILLinkWarningSuppression` be good?

vitek-karas: My understanding is that it is the name of the tool the attribute applies to - so it's not a "warning suppression". What would be more fitting is just `ILlinker` - basically the text we prefix all our messages (including warnings) with.

vitek-karas: BTW: From a casing perspective this should be spelled `ILLinker` - also in MessageContainer.ToMSBuildString. Would also be nice to have it as a constant used from both places.

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -0,0 +1,76 @@
...
+				warnings = new HashSet<(int, IMemberDefinition)> ();
+				_warnings.Add (assemblyName, warnings);
+			}
+
+			warnings.Add ((code, memberDefinition));
+		}
+
+		public void OutputSuppressions ()
+		{
+			foreach (var assemblyName in _warnings.Keys) {
```
marek-safar, line 33: You need to sort (or order) the Keys as well

vitek-karas: We might, but it's not necessary. The keys are assembly names and we write a separate file for each assembly name. It doesn't matter in which order the files are produced.

marek-safar: Right, missed that part

Issue #1289, opened 2020-06-23
Non-actionable scenario for warning IL2025


mateoatr:
Because of how `ResolveFromXmlStep` is implemented, whenever a user specifies a member in an XML descriptor, it'll mark not only the member but also its declaring type and assembly. In the scenario in which the `ResolveFromAssemblyStep` was run and an XML descriptor is being used to keep some member from being trimmed, this can cause the linker to output warning `IL2025` ("Duplicate preserve of ..."), but there'll be nothing the user can do to resolve it.

Since all warnings should be actionable by the user, I think we should not emit [this warning](https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/ResolveFromXmlStep.cs#L250).

Linking `System.Private.Xml` reproduces a scenario like the above mentioned.

PR #1291, opened 2020-06-23
Don't warn on duplicate preserve for types

mateoatr:
Related to #1289. Whenever we mark a member with `ResolveFromXmlStep` we also have to mark its declaring type. Because there can be multiple calls to this step, each marking a different member from the same declaring type, there's a chance that the linker will output this warning multiple times without having the possibility to resolve it.

### DIFF
--- a/src/linker/Linker.Steps/ResolveFromXmlStep.cs
+++ b/src/linker/Linker.Steps/ResolveFromXmlStep.cs
@@ -274,7 +274,6 @@ namespace Mono.Linker.Steps
 
 			if (Annotations.IsMarked (type)) {
 				var duplicateLevel = preserve != TypePreserve.Nothing ? preserve : nav.HasChildren ? TypePreserve.Nothing : TypePreserve.All;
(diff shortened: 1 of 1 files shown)


src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -274,7 +274,6 @@ protected virtual void ProcessType (TypeDefinition type, XPathNavigator nav)
 
 			if (Annotations.IsMarked (type)) {
 				var duplicateLevel = preserve != TypePreserve.Nothing ? preserve : nav.HasChildren ? TypePreserve.Nothing : TypePreserve.All;
-				Context.LogWarning ($"Duplicate preserve of '{type.FullName}' in '{_xmlDocumentLocation}'", 2025, _xmlDocumentLocation);
```
MichalStrehovsky, line 277: Is the `if` branch this is nested in useful for anything?

mateoatr: Ops. It's not, we should remove it as well.

PR #1292, opened 2020-06-23
Update option names and docs

sbomer:
`Action` and `DefaultAction` become `TrimMode`, and updates documentation to match.
The documentation needs more work, but this at least replaces some incorrect information.

Part of https://github.com/dotnet/sdk/issues/12035. The SDK side of this is implemented in https://github.com/dotnet/sdk/pull/12116. Once the SDK part is in, I will update the SDK here to get the integration tests working with the change.

### DIFF
--- a/docs/illink-tasks.md
+++ b/docs/illink-tasks.md
@@ -62,13 +62,10 @@ The linker default behavior depends on the SDK, because each SDK uses linker set
 By default, only framework assemblies are trimmed, and they are trimmed in a conservative
 assembly-level mode (`copyused` action). Third-party libraries and the app will be analyzed but not trimmed.
 
-By default, the linker will operate in assembly-level mode (`copyused` action) for all framework or
-core managed assemblies. The 3rd party libraries or final app will be analyzed but not trimmed.
-
 ### Blazor SDK
 
 By default, framework assemblies are trimmed in an aggressive member-level mode (`link` action). ASP.NET assemblies
-use type-level trimming (achieved by custom root descriptors, not `TrimMode`), and third-party libraries are analyzed
+use type-level trimming (achieved by using custom root descriptors with the global `TrimMode` `link`), and third-party libraries are analyzed
 but not trimmed.
 
 ## Customizing Linking Behavior
--- a/docs/illink-tasks.md
+++ b/docs/illink-tasks.md
@@ -55,18 +55,7 @@ The linker can be invoked as an MSBuild task, `ILLink`. We recommend not using t
 
 ## Default Linking Behavior
 
-The linker default behavior depends on the SDK, because each SDK uses linker settings appropriate for the target form factor.
-
-### .NET Core SDK
-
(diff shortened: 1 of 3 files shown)


docs/illink-tasks.md
```diff
@@ -53,20 +53,33 @@ The linker can be invoked as an MSBuild task, `ILLink`. We recommend not using t
...
 
-### .NET 3.x
+By default, only framework assemblies are trimmed, and they are trimmed in a conservative
+assembly-level mode (`copyused` action). Third-party libraries and the app will be analyzed but not trimmed.
 
-By default, the linker will operate in a conservative mode that keeps
-all managed assemblies that aren't part of the framework (they are
-kept intact, and the linker simply copies them).
+By default, the linker will operate in assembly-level mode (`copyused` action) for all framework or
+core managed assemblies. The 3rd party libraries or final app will be analyzed but not trimmed.
```
vitek-karas, line 66: This looks like a different wording of the statement right above it - I guess we need only one version...

docs/illink-tasks.md
```diff
@@ -53,20 +53,33 @@ The linker can be invoked as an MSBuild task, `ILLink`. We recommend not using t
...
+
+### Blazor SDK
+
+By default, framework assemblies are trimmed in an aggressive member-level mode (`link` action). ASP.NET assemblies
+use type-level trimming (achieved by custom root descriptors, not `TrimMode`), and third-party libraries are analyzed
+but not trimmed.
```
vitek-karas, line 72: Would be interesting (from the point of view of this doc) what is the `TrimMode` for the ASP.NET assemblies - I assume it's `link` + descriptors... right?

docs/illink-tasks.md
sbomer: Yup... was left over. Thanks.

docs/illink-tasks.md
sbomer: Yes, that's my understanding - clarified this. @pranavkm could you confirm whether this section is accurate?

pranavkm: This description is accurate 👍

docs/illink-tasks.md
```diff
@@ -53,20 +53,30 @@ The linker can be invoked as an MSBuild task, `ILLink`. We recommend not using t
...
 
-By default, the linker will operate in a conservative mode that keeps
-all managed assemblies that aren't part of the framework (they are
-kept intact, and the linker simply copies them).
+### Blazor SDK
+
+By default, framework assemblies are trimmed in an aggressive member-level mode (`link` action). ASP.NET assemblies
+use type-level trimming (achieved by using custom root descriptors with the global `TrimMode` `link`), and third-party libraries are analyzed
+but not trimmed.
+
```
marek-safar, line 70: I'm not sure this is worth mentioning here. It's pretty confusing for seasonal Web Developer and most of the terms are not self-describing. I'd just remove it as we don't need it here

Issue #1167, opened 2020-05-04
Design and implement implicit warning suppression


vitek-karas:
In some cases we want the linker to suppress itself from generating warnings. Right now linker does this already (sort-of) when `PreserveDependencyAttribute` is used. In such case the linker will simply skip analyzing the method's body. That has multiple side effects one of which is that it doesn't produce warnings from that method's body.

With the new attributes we probably want implicit suppression to work in additional cases as well:
* `RequiresUnreferencedCodeAttribute` should suppress warnings about calling other methods marked with `RequiresUnreferencedCodeAttribute`.
* `RequiresUnreferencedCodeAttribute` should probably suppress warnings cause by `DynamicallyAccessedMembersAttribute` (and related reflection pattern detection)

vitek-karas:
Related to https://github.com/mono/linker/issues/1140

vitek-karas:
This is implemented, but we should probably discuss the desired exact behavior.

The implementation currently completely disables data flow analysis of any method body where the method has `RequiresUnreferencedCodeAttribute` on it. This has two effects:
* No warnings will be generated by that method's body (the desired effect)
* No data flow processing occurs and thus the method doesn't apply any of the features implemented that way. So for example if the method calls `typeof(Foo).GetMethod("bar")` linker will NOT mark method `Foo;:bar` - even if normally it would. Similarly any application of `DynamicallyAccessedMembersAttribute` is skipped.

I think this needs a change. The data flow processing of such method should still occur, we just need to make sure that no warnings are generated by it. Note that going forward once we introduce new "analysis modes" like "single-file" or "AOT" those will have similar `RequiresFeatureAttribute` which should also disable warnings, but only those specific to that "analysis mode". So having the ability to run data flow but "selectively" disable warnings in it will be useful for this as well.

marek-safar:
It'd be useful to issue a warning if reflection analysis yields same value(s) as DynamicallyAccessedMembersAttribute has or the value is not included in DynamicallyAccessedMembersAttribute

vitek-karas:
> It'd be useful to issue a warning if reflection analysis yields same value(s) as DynamicallyAccessedMembersAttribute has or the value is not included in DynamicallyAccessedMembersAttribute

@marek-safar I don't think I understand the comment. This issue really only applies to `RequiresUnreferencedCodeAttribute`.

[...]

PR #1293, opened 2020-06-24
RequiresUnreferencedCode should not suppress reflection analysis

vitek-karas:
RequiresUnreferencedCode attribute should only suppress reflection analysis warnings for a given method, but reflection analysis should still run.

Should basically fix https://github.com/mono/linker/issues/1167.


vitek-karas:
I'm struggling with naming a little bit - specifically if anybody has a better idea for name of `ShouldEnableReflectionPatternReporting` I would be very interested.

Issue #1256, opened 2020-06-08
Remove clearinitlocals optimization


sbomer:
Once https://github.com/dotnet/runtime/issues/35527 is finished, we should remove the optimization from the linker. I'm not aware of anyone else using it.

PR #1295, opened 2020-06-25
Remove unused tasks and clearinitlocals optimization

sbomer:
@eerhardt pointed out that blazor was using some logic copied from one of our unused tasks, left over from when we shipped ILLink.Tasks OOB. This removes the unused tasks.

Also removes the unused clearinitlocals optimization. Fixes https://github.com/mono/linker/issues/1256.


test/ILLink.Tasks.Tests/ILLink.Tasks.Tests.cs
```diff
@@ -206,11 +206,11 @@ public static IEnumerable<object[]> PerAssemblyOptimizationsCases ()
 			yield return new object[] {
 				new ITaskItem [] {
 					new TaskItem ("path/to/Assembly1.dll", new Dictionary<string, string> {
-						{ "ClearInitLocals", "True" },
+						{ "Sealer", "True" },
 						{ "BeforeFieldInit", "False" }
 					}),
 					new TaskItem ("path/to/Assembly2.dll", new Dictionary<string, string> {
-						{ "ClearInitLocals", "False" },
+						{ "ealer", "False" },
```
marek-safar, line 213: Is this typo?

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -85,13 +84,6 @@ public class ILLink : ToolTask
 		public bool UnreachableBodies { set => _unreachableBodies = value; }
 		bool? _unreachableBodies;
 
-		/// <summary>
-		///   Boolean specifying whether to enable clearinitlocals optimization globally.
-		///   Maps to '--enable-opt clearinitlocals' or '--disable-opt clearinitlocals'.
-		/// </summary>
-		public bool ClearInitLocals { set => _clearInitLocals = value; }
```
eerhardt, line 92: Note that once you remove this, the line in the SDK that sets this is going to fail when the new ILLink task version is brought in:

https://github.com/dotnet/sdk/blob/12d968b2be23e1d581a4a92b915ce0d61e14eb11/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L103

sbomer: Like https://github.com/mono/linker/pull/1292, I'll wait for the SDK changes to flow into the linker, to get the integration tests passing.

PR #1297, opened 2020-06-25
Support for IgnoreCase BindingFlag in reflection pattern matching

### LINKED ISSUE #1070: Reflection pattern matching doesn't handle BindingFlags.IgnoreCase
One of the possible binding options is `BindingFlags.IgnoreCase` which will perform case insensitive match for the name. This applies to all reflection APIs which take a name and binding flags (`GetMethod`, `GetField`, ...).

The current intrinsic handling in the linker always performs a case sensitive match and it ignores the binding flags (even if it would be possible to determine the value statically).

This can lead to linker incorrectly considering some patterns as handled and not issuing warnings, while still breaking the app at runtime.

tlakollo:
This PR fixes #1070
- Introduce a conditional option for filter defined by the IgnoreCasebinding flag
- Add tests


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -871,13 +871,14 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 						} else if (calledMethod.Parameters.Count > 2 && calledMethod.Parameters[2].ParameterType.Name == "BindingFlags" && methodParams[3].AsConstInt () != null) {
 							bindingFlags = (BindingFlags) methodParams[3].AsConstInt ();
 						}
+						var caseOption = (bindingFlags & (BindingFlags.IgnoreCase)) == BindingFlags.IgnoreCase ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
```
MichalStrehovsky, line 874: This will need to assume `IgnoreCase` is set if this is an overload that has binding flags and we couldn't deduce the value statically (`methodParams[3].AsConstInt () == null`).

MichalStrehovsky:
> I wonder if we should just consider unknown binding flags or binding flags that want case insensitive linker unfriendly and warn...

(Before anyone says naming identifiers with non-ASCII characters and then using IgnoreCase binding flags to look for them is an academical scenario: https://github.com/dotnet/runtime/issues/11335 and https://github.com/dotnet/runtime/issues/13416.)

vitek-karas:
I assume your "fun" scenario is effectively a bug either in the runtime or libraries - right?

Another alternative would be to fallback to include everything. If we guarantee recognize all of the binding flags - do the magic, otherwise fallback to `DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods`. This way we make the code work (basically guaranteed) without additional warnings. The cost is size obviously, but to me this feels like a better trade off.

vitek-karas:
I actually like the fallback even better - it would solve the problem for not just `IgnoreCase` but any other binding flags we might have trouble with.

MichalStrehovsky:
> I assume your "fun" scenario is effectively a bug either in the runtime or libraries - right?

It's probably a doc bug because the doc doesn't specify in what sense the case insensitive comparison is done. In this case this is an culture invariant comparison as defined by Windows.

It differs from how case insensitive invariant comparison is done in .NET. Swapping `Ordinal` for `Invariant` would produce a bug in this one:

```csharp
class Program
{
    public static void Straße() { }
    static void Main()
    {
        var mi = typeof(Program).GetMethod("Strasse", BindingFlags.Public | BindingFlags.Static | BindingFlags.IgnoreCase);
        Console.WriteLine(mi != null);
        Console.WriteLine("Straße".Equals("Strasse", StringComparison.InvariantCultureIgnoreCase));
    }
}
```

vitek-karas:
Thanks @MichalStrehovsky - this tells me the linker should not be in the business of trying to emulate the runtime behavior (also, it's quite likely that Mono has slightly different behavior as well - let alone the Mono-wasm runtime). @tlaka thanks for the PR - the tests are good so let's keep those. 

The product should basically detect which combination of `BindingFlags` it can recognize (if it can determine the value statically) - in all other cases it should fallback to the `RequireDynamicallyAccessedMembers` and mark all public and nonpublic members.

Currently we really only support `BindingFlags.Instance`, `BindingFlags.Static`, `BindingFlags.Public` and `BindingFlags.NonPublic`. We should NOT recognize `BindingFlags.IgnoreCase`.

Open question about:
* `BindingFlags.Default` - @MichalStrehovsky  is there a predictable behavior for this one - I know that last time I tried it it was rather weird and changed per-API.
* `BindingFlags.DeclaredOnly` - we can consider this one "supported" but don't change behavior (it would mean to not walk the hierarchy - we can make that change later if we think it's an interesting size win)
* `BindingFlags.FlattenHierarchy` - I think we already implement this behavior always - so similarly to `DeclaredOnly` - recognized but don't change behavior.
* everything else - unrecognized, fallback to marking all members of a given type

MichalStrehovsky:
We should also extend this to the `Type.GetType` overload with two bools (one of the bools is to specify case insensitivity).

I would think that `Default` should return no members but I don't dare to make assumptions about reflection APIs. They're a mess.

I agree with the strategy Vitek outlined.

MichalStrehovsky:
> it's quite likely that Mono has slightly different behavior as well - let alone the Mono-wasm runtime)

Yep, Mono is failing the test for this. The test is one of the hundreds of "needs triage" excluded-on-mono tests in issues.targets.

tlakollo:
> We should also extend this to the `Type.GetType` overload with two bools (one of the bools is to specify case insensitivity).

I still don't get what should be the behavior of this
```
static void TestTypeIgnoreCase ()
{
	const string reflectionTypeKeptString = "MONO.LINKER.TEST.CASES.REFLECTION.IGNORECASECLASS+privateclass, tEsT";
	var typeKept = Type.GetType (reflectionTypeKeptString, false, true);
}
```
Basically any of the elements can have a weird casing, what should this function keep in those cases?

MichalStrehovsky:
> Basically any of the elements can have a weird casing, what should this function keep in those cases?

I would stop recognizing this overload as an intrinsic. That way linker will fall back to the annotations and this API is [annotated](https://github.com/dotnet/runtime/blob/f1bf0defd31da8f60dc33a63071a8e00b036e300/src/coreclr/src/System.Private.CoreLib/src/System/Type.CoreCLR.cs#L24-L30) as linker unsafe.

tlakollo:
I will issue a different PR for the Type.GetType case, in the meantime this PR is ready for review

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -877,7 +877,10 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 							if (value is SystemTypeValue systemTypeValue) {
 								foreach (var stringParam in methodParams[1].UniqueValues ()) {
 									if (stringParam is KnownStringValue stringValue) {
-										MarkMethodsOnTypeHierarchy (ref reflectionContext, systemTypeValue.TypeRepresented, m => m.Name == stringValue.Contents, bindingFlags);
+										if ((bindingFlags & BindingFlags.IgnoreCase) == BindingFlags.IgnoreCase || (int) bindingFlags > 255)
```
vitek-karas, line 880: Nit: I would wrap this in a function - to make it more readable. Something like `BindingFlagsAreSupported` ... currently the condition looks like magic. In the separate function we can add a comment to list all of the supported cases to make it clear why we "filter" out the rest.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -909,13 +912,16 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 							if (value is SystemTypeValue systemTypeValue) {
 								foreach (var stringParam in methodParams[1].UniqueValues ()) {
 									if (stringParam is KnownStringValue stringValue) {
-										TypeDefinition[] matchingNestedTypes = MarkNestedTypesOnType (ref reflectionContext, systemTypeValue.TypeRepresented, m => m.Name == stringValue.Contents, bindingFlags);
-
-										if (matchingNestedTypes != null) {
-											for (int i = 0; i < matchingNestedTypes.Length; i++)
-												methodReturnValue = MergePointValue.MergeValues (methodReturnValue, new SystemTypeValue (matchingNestedTypes[i]));
+										if ((bindingFlags & BindingFlags.IgnoreCase) == BindingFlags.IgnoreCase || (int) bindingFlags > 255)
+											RequireDynamicallyAccessedMembers (ref reflectionContext, DynamicallyAccessedMemberTypes.PublicNestedTypes | DynamicallyAccessedMemberTypes.NonPublicNestedTypes, value, calledMethodDefinition);
```
vitek-karas, line 916: Nit: Add a comment that we've chosen to not populate the `methodReturnValue` for now (it's technically doable as we know the exact type we're working on).

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -991,13 +997,22 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 									if (stringParam is KnownStringValue stringValue) {
 										switch (fieldPropertyOrEvent) {
 										case IntrinsicId.Type_GetEvent:
-											MarkEventsOnTypeHierarchy (ref reflectionContext, systemTypeValue.TypeRepresented, filter: e => e.Name == stringValue.Contents, bindingFlags);
+											if ((bindingFlags & BindingFlags.IgnoreCase) == BindingFlags.IgnoreCase || (int) bindingFlags > 255)
+												RequireDynamicallyAccessedMembers (ref reflectionContext, DynamicallyAccessedMemberTypes.PublicEvents | DynamicallyAccessedMemberTypes.NonPublicEvents, value, calledMethodDefinition);
+											else
+												MarkEventsOnTypeHierarchy (ref reflectionContext, systemTypeValue.TypeRepresented, filter: e => e.Name == stringValue.Contents, bindingFlags);
 											break;
```
vitek-karas, line 1004: Nit: Seeing how this is repeated I wonder if it would make sense to "hide" the binding flags check inside the Mark* methods... Not a big deal though, we could also leave it as is if you like this better.

vitek-karas:
Can you please create an issue for the Type.GetType case - just so that we track it...

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
tlakollo: I think I like this way because is visible along with the other fallbacks

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -781,6 +781,7 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 				//
 				case IntrinsicId.Type_GetType: {
 						reflectionContext.AnalyzingPattern ();
+						var parameters = calledMethod.Parameters;
```
vitek-karas, line 785: This doesn't seem to be used anywhere.

Issue #1294, opened 2020-06-24
Putting a file extension in an XML descriptor assembly name causes the descriptor elements to be skipped


eerhardt:
Create a descriptors xml file like the following:

```xml
<linker>
  <assembly fullname="Microsoft.AspNetCore.Authorization.dll">
    <type fullname="*" required="false" preserve="all" />
  </assembly>
</linker>
```

Note that the assembly fullname has the file extension `.dll` at the end.

When you pass this file into the linker with `-x`, the linker isn't able to resolve the assembly (because it has a `.dll` suffix on it), and the linker will ignore the descriptors in that assembly.

This is how Blazor was generating their descriptor files, and it wasn't working. See https://github.com/dotnet/aspnetcore/blob/cd9da1c3503d89e7f42b7ee780416eb1d4036b84/src/Razor/Microsoft.NET.Sdk.Razor/src/CreateBlazorTrimmerRootDescriptorFile.cs#L35.

We should consider either making this an error/warning, or supporting assembly names with the file extension.

cc @vitek-karas @marek-safar

marek-safar:
I consider this to be a bug on Blazor side. Discussed with @vitek-karas issuing warning in such case.

eerhardt:
> I consider this to be a bug on Blazor side

Agreed. I have https://github.com/dotnet/aspnetcore/pull/23330 to fix it in Blazor. Getting a warning to help diagnose such problems in the future could help save someone's time in the future.

vitek-karas:
#1299 now makes the linker issue a warning in this case.

PR #1299, opened 2020-06-26
Refactor reading of XML linker formats to reuse lot of code.

vitek-karas:
The descriptors and linker attributes logic is now almost entirely shared in the base class. The substitutions has still a little bit of additional code since it only handles fields and methods.

This change also adds more warnings for cases where the referenced items could not be resolved.

The effect of this change is that all XML files should now report warnings whenever the referenced items can't be resolved. (Before some XML files did report warnings in some cases while others didn't).

This change also modifies `TypeReference.GetDisplayName` to return namespace qualified display name as that is the typical form we want to see in error/warning messages. The "WithoutNamespace" version is also available.


vitek-karas:
The behavior of all the XML steps should be exactly as it was with the exception of more warnings in place. The review of this is obviously a mess. The most important to get right is `ResolveFromXmlStep` as that needs to work the same as before - please focus on that one.

vitek-karas:
This effectively fixes https://github.com/mono/linker/issues/1294.

/cc @eerhardt

src/linker/Linker.Steps/ProcessLinkerXmlStepBase.cs
```diff
@@ -0,0 +1,491 @@
...
+			}
+		}
+
+		void ProcessSelectedFields (XPathNavigator nav, TypeDefinition type)
+		{
+			XPathNodeIterator fields = nav.SelectChildren (FieldElementName, XmlNamespace);
+			if (fields.Count == 0)
+				return;
+
+			ProcessFields (type, fields);
```
tlakollo, line 195: I thought we didn't want to have a bunch of small functions like ProcessSelected* calling Process* and instead, we will have the implementation directly on the code since the call doesn't add any value.

vitek-karas: I'll look into cutting it down a little bit. I went with this just because I copied most of the code from `ResolveFromXmlStep`. That said, quite a few of these methods are `protected virtual` on `ResolveFromXmlStep`, so removing them is effectively a breaking change.

vitek-karas: Thanks for the suggestion - relatively simple change and the file is now 50 lines shorter 👍

src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -132,24 +81,31 @@ protected virtual void ProcessAssembly (AssemblyDefinition assembly, XPathNodeIt
...
 				foreach (TypeDefinition type in assembly.MainModule.Types) {
 					if (type.Namespace != fullname)
 						continue;
 
+					foundMatch = true;
 					MarkAndPreserveAll (type);
 				}
+
+				if (!foundMatch) {
+					// WARN
```
tlakollo, line 107: A wild missing warning

src/linker/Linker.Steps/ProcessLinkerXmlStepBase.cs
```diff
@@ -0,0 +1,439 @@
...
+						return;
+				}
+
+				if (!ShouldProcessElement (nav))
+					return;
+
+				ProcessAssemblies (nav.SelectChildren ("assembly", ""));
+
+			} catch (Exception ex) when (!(ex is LinkerFatalErrorException)) {
+				throw new LinkerFatalErrorException (MessageContainer.CreateErrorMessage ($"Error processing '{_xmlDocumentLocation}'", 1013), ex);
```
tlakollo, line 65: This is my fault, but the description of the error still talks about an XML descriptor, maybe better to change it to XML only

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -65,16 +74,16 @@ internal static void ParseGenericParameters (IList<GenericParameter> genericPara
 			sb.Insert (0, '<');
 		}
 
-		private static void ParseGenericArguments (Stack<TypeReference> genericArguments, int argumentsToTake, StringBuilder sb)
+		static void PrependGenericArguments (Stack<TypeReference> genericArguments, int argumentsToTake, StringBuilder sb)
 		{
-			sb.Insert (0, '>').Insert (0, genericArguments.Pop ().GetDisplayName ());
+			sb.Insert (0, '>').Insert (0, genericArguments.Pop ().GetDisplayNameWithoutNamespace ().ToString ());
```
mateoatr, line 79: I thought we were dropping namespaces from types in places other than the "main member" being printed.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
vitek-karas: Good catch - thanks!

src/linker/Linker.Steps/ProcessLinkerXmlStepBase.cs
vitek-karas: I fixed the md file to only talk about XML.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -345,114 +173,35 @@ static TypePreserve GetTypePreserve (XPathNavigator nav)
...
-		{
-			while (iterator.MoveNext ()) {
-				if (!ShouldProcessElement (iterator.Current))
-					continue;
-				ProcessField (type, iterator.Current);
-			}
-		}
-
-		protected virtual void ProcessField (TypeDefinition type, XPathNavigator nav)
+		protected override void ProcessField (TypeDefinition type, XPathNavigator nav)
```
marek-safar, line 176: You could #if !FEATURE_ILLINK whole override

PR #1300, opened 2020-06-26
Use GetDisplayName when logging types or methods

mateoatr:
Use `GetDisplayName` for logging methods and types when possible.

There are still places where we log types and methods by calling `ToString` implicitly on a `IMemberDefinition` (e.g., [here](https://github.com/mono/linker/blob/master/src/linker/Linker/Annotations.cs#L463)), we can transition to start using `GetDisplayName` in these cases after we implement the method for event, field, and property references. 

Although I double checked that I didn't miss a spot where `GetDisplayName` could be used, having so many places where we log a member's name, it's quite possible that I missed some.


mateoatr:
Changing this to WIP until #1299 is merged.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1546,14 +1548,14 @@ string GetParameterDescriptionForErrorMessage (ParameterDefinition parameterDefi
 			if (string.IsNullOrEmpty (parameterDefinition.Name))
 				return $"parameter #{parameterDefinition.Index} of method '{parameterDefinition.Method}'";
 
-			return $"parameter '{parameterDefinition.Name}' of method '{parameterDefinition.Method}'";
+			return $"parameter '{parameterDefinition.Name}' of method '{(parameterDefinition.Method as MethodDefinition).GetDisplayName ()}'";
```
vitek-karas, line 1551: This is potentially dangerous. There doesn't seem to be a guarantee that the cast will succeed. So it should be conditional.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -688,7 +688,7 @@ protected virtual void MarkUserDependency (MemberReference context, CustomAttrib
 				assembly = _context.GetLoadedAssembly (assemblyName);
 				if (assembly == null) {
 					_context.LogWarning (
-						$"Could not resolve '{assemblyName}' assembly dependency specified in a `PreserveDependency` attribute that targets method '{context.FullName}'", 2003, context.Resolve ());
+						$"Could not resolve '{assemblyName}' assembly dependency specified in a `PreserveDependency` attribute that targets method '{(context as MethodReference).GetDisplayName ()}'", 2003, context.Resolve ());
```
vitek-karas, line 691: This is potentially dangerous - the cast may return null in which case this will throw.
I think we should have an extension method for `MemberReference`, which for now will handle methods nicely and everthing else as `FullName`...

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1531,10 +1533,10 @@ string GetMetadataTokenDescriptionForErrorMessage (IMetadataTokenProvider target
 			return targetContext switch
 			{
 				ParameterDefinition parameterDefinition => GetParameterDescriptionForErrorMessage (parameterDefinition),
-				MethodReturnType methodReturnType => $"return value of method '{methodReturnType.Method}'",
 				FieldDefinition fieldDefinition => $"field '{fieldDefinition}'",
+				MethodReturnType methodReturnType => $"return value of method '{(methodReturnType.Method as MethodDefinition).GetDisplayName ()}'",
```
vitek-karas, line 1537: Possible nullref here (very unlikely, but still)

Issue #1288, opened 2020-06-22
Allow constant return values from methods with byvalue parameters to be propagated to callsites


eerhardt:
See this [discussion](https://github.com/dotnet/runtime/pull/38129/files#r442997376).

We should allow the linker to propagate a constant `return false;` from `EventSource::IsEnabled(EventLevel level, EventKeywords keywords, EventChannel channel)` to the callsites to allow for branch trimming.

See the code blocking this scenario here inside `TryInlineBodyDependencies`:

https://github.com/mono/linker/blob/b0b5143ee2a59517df69d38cb6c363cdd17177a8/src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs#L188-L189

cc @marek-safar @MichalStrehovsky @vitek-karas

PR #1302, opened 2020-06-29
Support propagation of constant return value from methods with parameters

vitek-karas:
This is currently limitted only to method explicitly substituted via substitution.xml and only to methods with in parameters. Out and Ref parameters are not allowed.

Fixes https://github.com/mono/linker/issues/1288.

/cc @eerhardt

### DIFF
--- a/src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
+++ b/src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
@@ -193,13 +193,11 @@ namespace Mono.Linker.Steps
 						if (!explicitlyAnnotated)
 							break;
 
-						bool hasByRefOrOutParameter = false;
-						for (int paramIndex = 0; paramIndex < md.Parameters.Count; paramIndex++) {
-							var param = md.Parameters[paramIndex];
-							hasByRefOrOutParameter |= param.ParameterType.IsByReference || param.IsOut;
-						}
+						bool hasByRefParameter = false;
+						foreach (var param in md.Parameters)
+							hasByRefParameter |= param.ParameterType.IsByReference;
 
-						if (hasByRefOrOutParameter || md.CallingConvention == MethodCallingConvention.VarArg)
+						if (hasByRefParameter || md.CallingConvention == MethodCallingConvention.VarArg)
 							break;
 					}
 
--- a/src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
+++ b/src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
@@ -194,8 +194,12 @@ namespace Mono.Linker.Steps
 							break;
 
 						bool hasByRefParameter = false;
-						foreach (var param in md.Parameters)
-							hasByRefParameter |= param.ParameterType.IsByReference;
+						foreach (var param in md.Parameters) {
+							if (param.ParameterType.IsByReference) {
+								hasByRefParameter = true;
+								break;
+							}
+						}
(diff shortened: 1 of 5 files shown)


src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -177,16 +177,31 @@ bool TryInlineBodyDependencies (ref BodyReducer reducer)
...
+					// Allow inlining results of methods with by-value parameters which are explicitly annotated
+					// but don't allow inlining of results of any other method with parameters.
+					if (md.HasParameters) {
+						if (!explicitlyAnnotated)
+							break;
+
+						bool hasByRefOrOutParameter = false;
+						for (int paramIndex = 0; paramIndex < md.Parameters.Count; paramIndex++) {
+							var param = md.Parameters[paramIndex];
+							hasByRefOrOutParameter |= param.ParameterType.IsByReference || param.IsOut;
```
marek-safar, line 199: This could break on the first find

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -177,16 +177,31 @@ bool TryInlineBodyDependencies (ref BodyReducer reducer)
...
 
+					bool explicitlyAnnotated = Annotations.GetAction (md) == MethodAction.ConvertToStub;
+
 					// Allow inlining results of instance methods which are explicitly annotated
 					// but don't allow inling results of any other instance method.
 					// See https://github.com/mono/linker/issues/1243 for discussion as to why.
 					// Also explicitly prevent inlining results of virtual methods.
 					if (!md.IsStatic &&
-						(md.IsVirtual || Annotations.GetAction (md) != MethodAction.ConvertToStub))
+						(md.IsVirtual || !explicitlyAnnotated))
```
marek-safar, line 187: Please add check for `vararg` as well

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -177,16 +177,31 @@ bool TryInlineBodyDependencies (ref BodyReducer reducer)
...
-					if (md.HasParameters)
-						break;
+					// Allow inlining results of methods with by-value parameters which are explicitly annotated
+					// but don't allow inlining of results of any other method with parameters.
+					if (md.HasParameters) {
+						if (!explicitlyAnnotated)
+							break;
+
+						bool hasByRefOrOutParameter = false;
+						for (int paramIndex = 0; paramIndex < md.Parameters.Count; paramIndex++) {
```
MichalStrehovsky, line 197: Nit: foreach?

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
MichalStrehovsky, line 199: If you specifically want `out` in the C# sense, this should be `param.ParameterType.IsByReference && param.IsOut`, but maybe we could just focus on the IsByReference part.

(This:

```csharp
static void Blah([Out] int x) { }
```

is just some harmless odd piece of metadata.)

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
vitek-karas: I added a check higher up - basically any method with vararg will not get its return value inlined (regardless if it's static or not).

PR #1305, opened 2020-06-30
Correctness fixes to interface stripping

### LINKED ISSUE #1235: Interface stripping produces invalid outputs
Found this when I was looking at #1077 (it looks like having to declare `IsAssignableFrom` linker unsafe would be quite a pain.)

It wouldn't be completely academical to assume that an app that has some `Foo`s implementing `IFoo` could end up in a situation where we have an empty array of `Foo`s (because no `Foo` was allocated because the code that normally news them up was e.g. linked out) that we need to cast to an array of `IFoo`.

Linker currently produces invalid output for this because it strips the `IFoo` implementation off `Foo`:

```csharp
using System;

interface IFoo { }

class Foo : IFoo { }

class Program
{
    static void Main()
    {
        // need to go through object to force a cast for added dramatic effect at runtime
        // otherwise this would just silently break type safety
        IFoo[] foos = (IFoo[])(object)new Foo[0];
    }
}
```

MichalStrehovsky:
This is actually 3 separate things:

* Bugfix in linker where linker was stripping type `Foo` despite the program containing a `new Foo[x,y]` statement. I hit this while writing tests for this.
* If a type is used as an array element or as a generic argument, don't strip any of its interfaces because they might be necessary for variant casting later.
* If a type is visible from reflection, don't strip any of its interfaces.

Fixes #1235.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1288,7 +1288,8 @@ namespace Mono.Linker.Steps
 		internal protected virtual TypeDefinition MarkTypeVisibleToReflection (TypeReference reference, DependencyInfo reason, IMemberDefinition sourceLocationMember)
 		{
 			// If a type is visible to reflection, we need to stop doing optimization that could cause observable difference
-			// in reflection APIs. This includes APIs like IsAssignableFrom.
+			// in reflection APIs. This includes APIs like MakeGenericType (where variant castability of the produced type
+			// could be incorrect) or IsAssignableFrom (where assignability of unconstructed types might change).
 			Annotations.MarkRelevantToVariantCasting (reference.Resolve ());
 
 			return MarkType (reference, reason, sourceLocationMember);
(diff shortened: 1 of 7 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2652,6 +2670,9 @@ protected virtual void MarkInstruction (Instruction instruction, MethodDefinitio
 				}
 
 			case OperandType.InlineType:
+				if (instruction.OpCode.Code == Code.Newarr) {
+					Annotations.MarkRelevantToVariantCasting (((TypeReference) instruction.Operand).Resolve ());
```
vitek-karas, line 2674: `Resolve` may return null - in which case we will add null to the `HashSet` in `Annotations`. I guess it doesn't matter that much, but it might be cleaner to check for null in the `MarkRelevantToVariantCasting` before adding it to the hashset.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2652,6 +2671,9 @@ protected virtual void MarkInstruction (Instruction instruction, MethodDefinitio
 				}
 
 			case OperandType.InlineType:
+				if (instruction.OpCode.Code == Code.Newarr) {
```
marek-safar, line 2674: Are you sure this is the only relevant opcode? What about code like `var foos = (IFoo[]) Array.CreateInstance (typeof (Foo), 1);` ?

MichalStrehovsky: That pattern should be fine because the `typeof(Foo)` means "visible to reflection" and that's handled elsewhere in this pull request.

Issue #1163, opened 2020-05-04
Warning for references to explicitly removed attributes


marek-safar:
The outcome of the attribute removal should be predictable and therefore we need to issue a new warning when any of the removed attributes is still referenced in the user code. This should apply in all cases where the attributes are explicitly removed (for example as part of #1162).

vitek-karas:
We should also think about scenarios where the attribute is not referenced via type-ref. Maybe we won't be able to solve it, but we should see if there's something we can do about this. An example of code which would still be wrong:
```C#
foreach (var a in myType.GetCustomAttributes()) {
    if (a.Name == "NullableAttribute" && a.Namespace == "System.Runtime.CompilerServices") {
        // Do something which should never happen in linked app - and thus be broken
    }
}
```

Maybe we should simply mark unconditional `GetCustomAttributes` as a problematic API.

/cc @MichalStrehovsky

MichalStrehovsky:
I think you know my stance on attribute removal... It's pretty problematic.

For the nullable ones, things like EF are already reading the nullable attributes and because the identity of these attributes is based on a name match, they use the catch-all APIs to enumerate them.

There's a proposal to add official APIs to read these annotations that we could key off at some point but it will require more sophistication in the linker. https://github.com/dotnet/runtime/issues/29723

vitek-karas:
@tlakollo  - I think this was already done as part of #1307 - please close this issue if that's the case.

tlakollo:
Yup that is the case

Issue #1162, opened 2020-05-04
Support for LinkerRemovableAttribute


marek-safar:
Once #1118 is completed we should extend it further to allow specifying "always" removed attributes. This is will be used for the final app mode of linker where some of the attributes can be always removed but we don't want to hardcode them in the linked. The idea is to have them annotated in the code or inside the custom attribute annotations.

Custom Attributes to be included in the list
- all nullable attributes
- System.CLSCompliantAttribute
- System.Security.UnverifiableCodeAttribute
- etc (the full list will be collected separately and might be RID specific)

Proposal of how it can look like when building on top of #1118 

```xml
<linker>
  <assembly fullname="System.Private.CoreLib">
    <type fullname="System.CLSCompliantAttribute">
		<attribute FullName="System.LinkerRemovableAttribute" AssemblyName="System.Private.CoreLib">
		</attribute>
    </type>
  </assembly>
</linker>
```

The attribute name `System.LinkerRemovableAttribute` will be the only name hardcoded in linker which will remove all attributes usages but not the attribute definition.

To make this even more effective and to allow removing attributes which are feature linker the XML format should be extended to support `feature` and `featurevalue` attributes like in [substitutions format](https://github.com/mono/linker/blob/master/docs/data-formats.md#conditional-substitutions)

```xml
<linker>
  <assembly fullname="Assembly">
    <type fullname="System.Runtime.InteropServices.GuidAttribute" feature="COM" featurevalue="false">
		<attribute FullName="System.LinkerRemovableAttribute" AssemblyName="System.Private.CoreLib">
		</attribute>
    </type>
  </assembly>
</linker>
```

vitek-karas:
The name of the attribute can remain internal for now (as we don't expect people outside of runtime/CoreFX to need to specify this) - but we should think about the name. The typical discussion of having the name of the tool in the attribute name.

Additionally, some attributes are auto-generated into the target assembly, so they don't have a well defined "declaring" assembly. Good example are the nullable attributes `NullableAttribute`, `NullableContextAttribute`, .... They can be defined in the assembly where they are needed. From the point of view of the linker there will be several `TypeDefinition` instances all representing for example `NullableAttribute` from different assemblies. To handle this the XML syntax would need to support something like a wildcard for assembly, so for example:
```XML
<linker>
  <assembly fullname="*">
    <type fullname="System.Runtime.CompilerServices.NullableAttribute">
      <attribute fullname="System.LinkerRemovableAttribute" />
    </type>
[...]

PR #1307, opened 2020-07-01
Add a mechanism to remove custom attribute instances

tlakollo:
This PR adds functionality described by #1162 and #1163. By using:
```xml
<linker>
  <assembly fullname="*"> 
    <type fullname="CustomAttributeName">
      <attribute internal="RemoveAttributeInstances" feature="EnableOptionalFeature" featurevalue="false" />
    </type>
  </assembly>
</linker>
```
The user will be able to delete all instance usages of `CustomAttribributeName`. In case the linker still references the custom attribute after the deletion the linker will warn.


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1335,6 +1338,10 @@ internal protected virtual TypeDefinition MarkType (TypeReference reference, Dep
 			if (type.HasMethods && ShouldMarkTypeStaticConstructor (type) && reason.Kind == DependencyKind.DeclaringTypeOfCalledMethod)
 				MarkStaticConstructor (type, new DependencyInfo (DependencyKind.TriggersCctorForCalledMethod, reason.Source), type);
 
+			// Check type being used was not removed by the LinkerRemovableAttribute
+			if (_context.Annotations.HasLinkerAttribute<LinkerRemovableAttribute> (type))
+				_context.LogWarning ($"Custom Attribute {type} is being referenced in code after Custom Attribute type was removed", 2045, sourceLocationMember);
```
mateoatr, line 1343: Use `GetDisplayName` for printing the type.

docs/error-codes.md
```diff
@@ -233,3 +233,7 @@ error and warning codes.
 #### `IL2044`: Could not find any type in namespace 'namespace' specified in 'XML document location'
 
 - The 'XML document location' specifies a namespace 'namespace' but there are no types found in such namespace.
+
+#### `IL2045`: Custom Attribute 'type' is being referenced in code after Custom Attribute type was removed
```
mateoatr, line 237: I think the wording is a little bit strange here and doesn't tell me much about what should I do to resolve the warning.

vitek-karas:
```suggestion
#### `IL2045`: Attribute type 'type' is being referenced in code but the linker was instructed to remove all instances of this attribute. If the attribute instances are necessary make sure to either remove the linker attribute XML portion which removes the attribute instances, or to override this use the linker XML descriptor to keep the attribute type (which in turn keeps all of its instances).
```

Something along these lines. I know this PR doesn't implement the descriptor part, but we will do that and the warning should be actionable.

src/linker/System.Diagnostics.CodeAnalysis/LinkerRemovableAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace System.Diagnostics.CodeAnalysis
+{
+	/// <summary>
+	/// This attribute name will be the name hardcoded in linker which will remove all 
+	/// attribute usages but not the attribute definition
+	/// </summary>
+	[AttributeUsage (
+		AttributeTargets.Class, Inherited = false)]
+	public sealed class LinkerRemovableAttribute : Attribute
```
vitek-karas, line 13: The name should be very descriptive: `RemoveAttributeInstancesAttribute` - to match the actual name in the XML.

src/linker/System.Diagnostics.CodeAnalysis/LinkerRemovableAttribute.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Diagnostics.CodeAnalysis
```
vitek-karas, line 5: This is not the right namespace - I would put this into a linker namespace - it's a purely internal class. So either just the Linker namespace, or one below it.

src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -24,14 +24,23 @@ public LinkAttributesStep (XPathDocument document, EmbeddedResource resource, As
...
 			while (iterator.MoveNext ()) {
+				if (!ShouldProcessElement (iterator.Current))
+					continue;
+
 				AssemblyDefinition assembly;
 				TypeDefinition attributeType;
 
+				string internalAttribute = GetAttribute (iterator.Current, "internal");
+				if (internalAttribute == "RemoveAttributeInstances" && provider != null) {
+					Context.Annotations.AddRemovableAttribute ((IMemberDefinition) provider);
```
vitek-karas, line 40: This won't work correctly. Basically if this type has more than one attribute from the XML (or from multiple XMLs) it will lose all but the remove one. Calling this method on the annotations will do the "cache the attributes" in the linker, but at that point the CustomAttributes is not yet populated with all XML attributes - so it will only see some attributes, not all.

vitek-karas: I would probably solve it like this:
* add a new Dictionary to CustomAttributeSource - something like _internalAttributes which will be `Dictionary<ICustomAttributeProvider, IEnumerable<Attribute>>`
* add `AddInternalAttribute`, `HasInternalAttributes` and `GetInternalAttributes` to the class
* teach the `LinkerAttributesInformation` about the second set of internal attributes so that it populates the cache from those as well.

vitek-karas: Add a test for this (basically the RemoveAttributeInstances followed by some test attribute on the same type, both should take effect.

test/Mono.Linker.Tests.Cases/DataFlow/LinkerAttributeRemoval.cs
```diff
@@ -41,4 +45,13 @@ public static void TestType ()
 			var typeKept = Type.GetType (reflectionTypeKeptString, false);
 		}
 	}
+
+	[KeptBaseType (typeof (System.Attribute))]
+	class TestDontRemoveAttribute:Attribute
```
vitek-karas: This is a good test, but we need another one - in the LinkerAttributeRemoval.xml try to add another attribute which will have effect on linker onto the attribute itself. This is tricky since we don't really have anything right now... maybe suppressmessage attribute would work.

Maybe add an issue to add this test as it needs a bit more work to figure out how to test it - basically we want to test that one can combine both normal attributes and internal attribute on the same element in the XML (And that it works).

tlakollo: I opened #1323  to track that test

PR #1308, opened 2020-07-02
Keep GetInstance method on custom marshalers

### LINKED ISSUE #1281: Linker breaks custom marshalers
If a type is used as `MarshalAs.CustomMarshaler`, we need to keep the `GetInstance` method on it because the runtime is going to call it.

I'm going to mark code that relies on this within the class libraries as linker safe because this is a linker bug.

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.icustommarshaler?view=netcore-3.1

MichalStrehovsky:
Fixes #1281.

We also need a trimming directive in CoreLib that says we want to keep all methods on ICustomMarshaler to make this work E2E, but that's not a linker concern anymore (and it's not testable here...).

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -108,6 +108,9 @@ namespace Mono.Linker.Steps
 			DependencyKind.ReturnType,
 			DependencyKind.UnreachableBodyRequirement,
 			DependencyKind.VariableType,
+			DependencyKind.ParameterMarshalSpec,
+			DependencyKind.FieldMarshalSpec,
+			DependencyKind.ReturnTypeMarshalSpec,
 		};
 
 		static readonly DependencyKind[] _methodReasons = new DependencyKind[] {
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1877,6 +1885,16 @@ protected bool MarkDefaultConstructor (TypeDefinition type, in DependencyInfo re
 			return MarkMethodIf (type.Methods, MethodDefinitionExtensions.IsDefaultConstructor, reason, sourceLocationMember) != null;
 		}
 
+		protected bool MarkGetInstanceMethod (TypeDefinition type, in DependencyInfo reason, IMemberDefinition sourceLocationMember)
```
vitek-karas, line 1888: Nit: why return `bool`  (not used in here)... and also why `protected`.

MichalStrehovsky: I copy-pasted MarkDefaultConstructor above. No special reason. Undone.

PR #1309, opened 2020-07-03
Postpone marking fields on types with layout

MichalStrehovsky:
We can postpone marking fields on types with layout until we're really sure they're needed. They're needed if they're visible to reflection, or considered instantiated (valuetypes are implicitly considered instantiated).

I'm also allowing stripping fields on Auto layout structs. This is a rare sight, but there's no reason that I know of to force them to have all their fields kept.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1292,6 +1292,8 @@ namespace Mono.Linker.Steps
 			// could be incorrect) or IsAssignableFrom (where assignability of unconstructed types might change).
 			Annotations.MarkRelevantToVariantCasting (reference.Resolve ());
 
+			MarkImplicitlyUsedFields (reference.Resolve ());
+
 			return MarkType (reference, reason, sourceLocationMember);
 		}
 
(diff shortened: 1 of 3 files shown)


marek-safar:
This is causing regression in linker bump https://github.com/dotnet/runtime/pull/38880. The problem seems to be trigger by this code https://github.com/dotnet/runtime/blob/master/src/mono/netcore/System.Private.CoreLib/src/System/Array.Mono.cs#L57 which uses StructLayout type as type argument and RawData struct no longer has all fields marked

@MichalStrehovsky could please look into that or revert this change

MichalStrehovsky:
Right, linker optimizations are only valid as long as type safety is ensured (e.g. if we were to put instance methods on `RawData` and start calling them, we would quickly find out that the "unreachable bodies" optimization stubbed them out).

The fix would be to put DynamicDependency on RawData that makes RawData depend on it's ctor, but we don't support that option.

The only options I see are an ILLinkTrim entry for `RawData::.ctor`, or reverting this change. Do you have a preference?

MichalStrehovsky:
Brainstorming some more, this particular problem could also be fixed by keeping all fields on explicit/sequential types if any field is kept. It's weird, but it would work and _probably_ still let us optimize out junk.

Issue #1028, opened 2020-03-26
[feature/dataflow] Add validation that dataflow annotations match for virtuals


MichalStrehovsky:
Need to make sure that e.g. if an interface method has an annotated parameter, all the implementations have a matching annotation. And vice versa.

https://github.com/mono/linker/pull/1024#discussion_r398618497

marek-safar:
I guess this applies to implicit interface implementations only

MichalStrehovsky:
It doesn't matter whether it's explicit or implicit. If this is restricted to implicit only, we wouldn't warn for `A`, but the problem is the same (at the callsite, we don't know we need to preserve the default ctor of X/Y, but the implementation assumes that it was kept (and the CreateInstance is safe).

```csharp
interface IFoo
{
    void A(Type t);
    void B(Type t);
}

class Foo : IFoo
{
    void IFoo.A([DynamicallyAccessed(Constructors)]Type t) => Activator.CreateInstance(t);
    public void B([DynamicallyAccessed(Constructors)]Type t) => Activator.CreateInstance(t);

    static void Main()
    {
        IFoo p = new Foo();
        p.A(typeof(X));
        p.B(typeof(Y));
    }
}
```

marek-safar:
I thought you were describing case like

```c#
interface IFoo
{
    void A([DynamicallyAccessed(Constructors)]Type t);
}
```

Do we have any data which shows this can be actually implemented for at least some real-world cases (putting generics problem aside) ?

MichalStrehovsky:
If the interface is annotated but the implementation isn't, this is less of a problem - it's only weakening the annotation (we ensured linker keeps more stuff than the particular implementation needs - there still might be a different implementation that needs what the interface demands).

This is needed to achieve correctness so I didn't look whether there is real world code using the pattern - if we don't fix this, we could see linker breaking an app without generating a warning about it.

marek-safar:
> if we don't fix this, we could see linker breaking an app without generating a warning about it.

Agree but that essentially means if someone adds DynamicallyAccessed to one of the interface implementations every interface usage has to be annotated as well (kind of scary to think about it for stuff like IEnumerable)

vitek-karas:
Implemented in https://github.com/mono/linker/pull/1315.

Issue #1157, opened 2020-05-01
Improve IL2006 warning


marek-safar:
Today the users are facing with a message like

```
Unrecognized reflection pattern warning IL2006: The parameter 'attributeType' of method 'System.Attribute System.Diagnostics.Tracing.EventSource::GetCustomAttributeHelper(System.Reflection.MemberInfo,System.Type,System.Diagnostics.Tracing.EventManifestOptions)' with dynamically accessed member kinds 'None' is passed into the parameter 'type' of method 'System.Object System.Activator::CreateInstance(System.Type)' which requires dynamically accessed member kinds `DefaultConstructor`. To fix this add DynamicallyAccessedMembersAttribute to it and specify at least these member kinds 'DefaultConstructor'.
```

I'd suggest
- Add location information (at least assembly name)
- Drop return types from the method signatures
- Don't use parameter name but parameter index
- Split the text to be easier to understand
	- a first warning for missing attribute
	- a second warning for incomplete MemberKinds

An example of the proposed change

```
(foo.dll) warning IL2006: The parameter '#1' in method 'System.Diagnostics.Tracing.EventSource::GetCustomAttributeHelper(System.Reflection.MemberInfo,System.Type,System.Diagnostics.Tracing.EventManifestOptions)' is missing 'DynamicallyAccessedMembersAttribute'. Consider adding it with 'DefaultConstructor' value.
```

/cc @vitek-karas

vitek-karas:
Part of this should be to avoid printing out potentially long list of enum values. The `DynamicallyAccessedMemberType` is a flags enum, the warning message may potentially need to include representation of multiple such flags which would make it hard to read.

See https://github.com/mono/linker/pull/1315#discussion_r451314266 for more details.

vitek-karas:
@marek-safar I don't understand why this is desirable:
> Don't use parameter name but parameter index

I think that this will be rather confusing to users:
* It's unclear if the numbering starts with 0 or 1
* Parameter names are considered part of the public API surface (several languages can refer to them by name even in the callsite, not just the callee body)
* The code will fallback to index when it can't determine the parameter name (never ran into that case though)

marek-safar:
I'm not sure this is confusing. To me syntax `#N` means it starts with 1 or does `#0` make sense to you like a number? 

> Parameter names are considered part of the public API

I don't think parameter names as part of the public API surface matter here, the error handling can happen anywhere, for private stuff, compiler-generated methods (maybe not for this code but in general).

[...]

PR #1315, opened 2020-07-06
Validation of annotations across virtual method hierarchies

vitek-karas:
For the linker to work correctly all correctness annotations must be exactly the same on all methods in a single virtual method override hierarchy.

Currently this applies to DynamicallyAccessedMemberAttribute and RequiresUnreferencedCodeAttribute.

The validation is postponed after the MarkStep since it requires full knowledge of virtual method hierarchies (and interface method hierarchies). With TypeMapStep these are available during MarkStep as well, but we're trying to move away from that requirement, so the validation is done after MarkStep, this only adds a very small amount of memory (the list of methods to validate is currently cca 200 for the entire corelibe).

Refactorings to help with this change:
* Change how we store generic parameter annotations - it now copies the method parameter annotation storage - array indexed by the index of the generic parameter on the type/method. This makes it easy to map the annotation to a generic parameter on any method in the virtual hierarchy (as they all share the same index).
* FlowAnnotations are now stored directly on the Annotations class (since the new step needs them after MarkStep)
* Move some helpers into DataFlowUtilities - mostly around getting description of members for error message purposes
* Synchronize behavior of LogContains and LogDoesNotContain test attributes - both now support "contains" and "regex" versions
* Correctly detect persense of LogContains and LogDoesNotContain attributes on all members of a test, so that test logging is turned on when they're used
* Added a TestSystemTypeBase class - which is a class derived from System.Type and implements all abstract members. This is used to then write test classes which need to derive from System.Type (to test the "this" parameter annotations).

This implements https://github.com/mono/linker/issues/1028 and https://github.com/mono/linker/issues/1028 - the very strict version. We can dial it back for the case of implementation->interface if we think that's important, but that's relatively tricky (for example it's possibly OK to not annotate implementation's in parameters even if interface has it, but it's not OK to omit annotation on the return value in this case - so we would have to correctly track in/out annotations for the purposes of validation).


src/linker/Linker.Dataflow/DataFlowUtilities.cs
```diff
@@ -0,0 +1,108 @@
+﻿using System.Collections.Generic;
```
marek-safar, line 1: Missing licence header

src/linker/Linker.Dataflow/DataFlowUtilities.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using Mono.Cecil;
+
+namespace Mono.Linker.Dataflow
+{
+	static class DataFlowUtilities
```
marek-safar, line 8: Weird name ;-) 

`Mono.Linker.Dataflow.DataFlowUtilities`

src/linker/Linker.Dataflow/DataFlowUtilities.cs
```diff
@@ -0,0 +1,108 @@
...
+			if (memberTypes.HasFlag (DynamicallyAccessedMemberTypes.NonPublicNestedTypes))
+				results.Add (DynamicallyAccessedMemberTypes.NonPublicNestedTypes);
+
+			if (memberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicNestedTypes))
+				results.Add (DynamicallyAccessedMemberTypes.PublicNestedTypes);
+
+			if (results.Count == 0)
+				return DynamicallyAccessedMemberTypes.None.ToString ();
+
+			return string.Join (" | ", results.Select (r => r.ToString ()));
```
marek-safar, line 57: Do we actually want such output? It can produce almost unreadable text

src/linker/Linker/VirtualMethodAnnotationValidator.cs
```diff
@@ -0,0 +1,75 @@
...
+		{
+			if (!method.IsVirtual)
+				return;
+
+			if (_context.Annotations.FlowAnnotations.RequiresDataFlowAnalysis (method) ||
+				_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method))
+				_virtualMethodsWithAnnotationsToValidate.Add (method);
+		}
+
+		public void Validate ()
```
marek-safar, line 37: I'd rather like to see this logic inside the ValidateVirtualMethodAnnotationsStep

src/linker/Linker/VirtualMethodAnnotationValidator.cs
```diff
@@ -0,0 +1,75 @@
...
+		readonly LinkContext _context;
+		readonly HashSet<MethodDefinition> _virtualMethodsWithAnnotationsToValidate;
+
+		public VirtualMethodAnnotationValidator (LinkContext context)
+		{
+			_context = context;
+			_virtualMethodsWithAnnotationsToValidate = new HashSet<MethodDefinition> ();
+		}
+
+		public void EnqueueVirtualMethod (MethodDefinition method)
```
marek-safar, line 27: Would make sense to have this method on AnnotationStore

src/linker/Linker/VirtualMethodAnnotationValidator.cs
```diff
@@ -0,0 +1,75 @@
...
+
+			if (_context.Annotations.FlowAnnotations.RequiresDataFlowAnalysis (method) ||
+				_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method))
+				_virtualMethodsWithAnnotationsToValidate.Add (method);
+		}
+
+		public void Validate ()
+		{
+			foreach (var method in _virtualMethodsWithAnnotationsToValidate) {
+				var baseMethods = _context.Annotations.GetBaseMethods (method);
```
marek-safar, line 40: How is this going to work with if the method implements 2 different interfaces which message will win?

src/linker/Linker.Dataflow/DataFlowUtilities.cs
vitek-karas: It should recreate the text used in the attribute in source code pretty well. By far the most common case is annotations with one or two flags from the enum, so that should be fine.

I must admit I don't understand the scenario when this would produce a mess. The only potential is "I want to include everything", but for that we have the `All` value which is handled explicitly.

src/linker/Linker/VirtualMethodAnnotationValidator.cs
vitek-karas: I didn't want to pass around a reference to the step... but I guess I can put the collection onto Annotations directly and move the rest to the step.

src/linker/Linker/VirtualMethodAnnotationValidator.cs
vitek-karas: It will compare the methods against both of the interface methods. If both don't match it will produce two warnings. I'll add a test for that.

src/linker/Linker/Annotations.cs
```diff
@@ -487,5 +490,15 @@ public bool SetPreservedStaticCtor (TypeDefinition type)
 			attribute = attributes.FirstOrDefault ();
 			return attribute != null;
 		}
+
+		public void EnqueueVirtualMethod (MethodDefinition method)
+		{
+			if (!method.IsVirtual)
+				return;
+
+			if (context.Annotations.FlowAnnotations.RequiresDataFlowAnalysis (method) ||
```
marek-safar, line 499:
```suggestion
			if (FlowAnnotations.RequiresDataFlowAnalysis (method) || HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method))
```

src/linker/Linker.Steps/ValidateVirtualMethodAnnotationsStep.cs
```diff
@@ -2,13 +2,49 @@
...
+using Mono.Cecil;
+
 namespace Mono.Linker.Steps
 {
 	public class ValidateVirtualMethodAnnotationsStep : BaseStep
 	{
 		protected override void Process ()
 		{
-			Context.VirtualMethodAnnotationValidator.Validate ();
+			foreach (var method in Context.Annotations.VirtualMethodsWithAnnotationsToValidate) {
```
marek-safar, line 14: You could store Context.Annotations in local variable to have fewer prefixes here

src/linker/Linker.Dataflow/DataFlowUtilities.cs
marek-safar: We used this to dump to user two sets of enums and let him compare. This is very hard as most of the options are very similar. I don't know if the common path will use 2, 3 or more values but I think it'd be better to re-word the error message to allow us to report only single enum value instead.

(ignoring the aspect that this is C#-ish in IL-based tool)

vitek-karas: I'd like to address this in a separate PR as it's mostly unrelated to this one (the validation introduced in this PR doesn't actually use this method, I just moved it out of the already large `ReflectionMethodBodyScanner`).

I added note about this to https://github.com/mono/linker/issues/1157 which should address the warning message where this is used from.

marek-safar: ok

Issue #1158, opened 2020-05-02
Fix inconsistencies in error formatting


sbomer:
Related to https://github.com/mono/linker/issues/1153, but I made a new issue for this because it's not about the contents of the messages.

Here's an example of what gets logged today:
<img width="623" alt="Screen Shot 2020-05-01 at 8 55 20 PM" src="https://user-images.githubusercontent.com/787361/80854524-12ac9900-8bee-11ea-9c0e-eecf783738f3.png">

The first line is from the new error logging, and the second line from the old. The inconsistent prefixes will go away once all warnings/errors get moved to the new logging APIs, but I'd make two suggestions:
- Use `ILLink` instead of `illinker`
- Ensure that our warnings all use single quotes instead of double quotes

vitek-karas:
@mateoatr I assume this is fixed now - can you please double check and close?

sbomer:
Since we've been discussing naming conventions, do we feel that 'ILLink' (or 'illinker'?) is the right prefix here? @eerhardt

mateoatr:
I had already checked that we were only using `ILLink` and single quotes. I just double checked. The only inconsistency I've seen is that sometimes we write the name of custom attributes between grave accents ('`'), and sometimes we just use the name of the cas (without any sort of quotes).

e.g.:
https://github.com/mono/linker/blob/master/src/linker/Linker.Dataflow/FlowAnnotations.cs#L286
https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/MarkStep.cs#L691

vitek-karas:
> Since we've been discussing naming conventions, do we feel that 'ILLink' (or 'illinker'?) is the right prefix here?

I think it's fine - I know there's a bit of a confusion, but there's at least a system to it:
* The capability/feature is "trim" - the general tool to implement this is "trimmer"
* The actual implementation of the feature is in ILLink - the tool is "ILLinker"

So MSBuild properties are called "Trim" since they're talking about the capability. Warnings/Errors are called "ILLinker" since they are comming from "ILLinker".

It's similar to for example Compile/csc
* The capability is compilation
* The actual implementation is "csc"

So MSBuild properties are called "compile" (mostly), but the errors are all "CSC".

marek-safar:
> I think it's fine - I know there's a bit of a confusion, but there's at least a system to it:

I think this is worth documenting perhaps even on the doc entry page https://github.com/mono/linker/tree/master/docs

vitek-karas:
[...]

PR #1317, opened 2020-07-07
Add terminology description to the illink task doc

vitek-karas:
Related to https://github.com/mono/linker/issues/1158

### DIFF
--- a/docs/illink-tasks.md
+++ b/docs/illink-tasks.md
@@ -2,15 +2,15 @@
 
 ## Terminology
 
-The capability of the SDK is called "trim". Trimming allows the SDK to reduce the size of the application by removing unused code and other assets. MSBuild/SDK integration should use "trim" in all user facing names (property names, item group names, ...).
+Trimming allows a developer to reduce the final size of the application by removing unused managed code (IL) and other assets. Any publicly visible MSBuild property or item group name should use the term "trim" to refer to this capability.
 
-The current implementation of the capability is via the "ILLink" tool/task. Actual errors/warnings will come from the tool, so using the "ILLink" prefix.
+The trimming process exposed to MSBuild via the ILLink.Tasks assembly which calls into ILLink tool to perform the actual trimming. The ILLink tool produces the trimmed output and can issue its own [warnings and errors](error-codes.md). All of them use `ILLink` prefix and unique code for easier identification.
 
 *Note: This is similar to Compiler/csc: the capability is "compilation" and the related MSBuild/SDK integration uses "compile" as the term. The actual tool is csc and the errors/warnings coming from it are prefixed with CSC/CS.*
 
 ## Usage
 
-To enable ILLinker set `PublishTrimmed` property to `true` in your project and publish your app as self-contained.
+To enable ILLink set `PublishTrimmed` property to `true` in your project and publish your app as self-contained.
 
 ```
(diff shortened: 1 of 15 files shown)


docs/illink-tasks.md
```diff
@@ -1,5 +1,13 @@
 # ILLink.Tasks
 
+## Terminology
+
+The capability of the SDK is called "trim". Trimming allows the SDK to reduce the size of the application by removing unused code and other assets. General name for the tool which implements this capability is "trimmer". MSBuild/SDK integration should use these terms (trim/trimmer).
+
+The current implementation of the capability is via the "ILLink" task. The tool which the task uses is called "ILLinker". Actual errors/warnings will come from the tool, so "ILLinker".
```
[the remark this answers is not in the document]
vitek-karas: I changed the text as per your suggestion. I also renamed ILLinker to ILLink. I do agree we should make the change... and I don't think it's controversial 😄

docs/illink-tasks.md
```diff
@@ -1,5 +1,13 @@
 # ILLink.Tasks
 
+## Terminology
+
+The capability of the SDK is called "trim". Trimming allows the SDK to reduce the size of the application by removing unused code and other assets. MSBuild/SDK integration should use "trim" in all user facing names (property names, item group names, ...).
```
sbomer, line 5:
```suggestion
The capability of the SDK is called "trim". Trimming allows the SDK to reduce the size of the application by removing unused code and other assets. MSBuild/SDK integration should use "trim" in all user facing names specific to trimming (property names, item group names, ...).
```

Nit: since we do have some integration points that technically are user-facing for advanced users (ex: `PrepareForILLink`) that relate to the tool, maybe be a little more specific.

agocke, line 5: Nit: this sentence is a bit confusing because it reads like "the SDK" is the .NET SDK, and singular "capability" implies that the SDK has only one capability, namely trimming. Could you instead say: "The ILLink tasks provide the ability to "trim" a set of published assemblies"?.

marek-safar: I agree the wording is a bit confusing. My suggestion

```suggestion
Trimming allows a developer to reduce the final size of the application by removing unused managed code (IL) and other assets. When working with MSBuild properties or items groups any publicly visible name which is related to trimming starts with `Trim` prefix.

The trimming process is managed by MSBuild based ILLink.Tasks assembly which calls into ILLink tool to perform the actual trimming. The ILLink tool handles all the heavy work to produce trimmed output and can issue its own [warnings and errors](error-codes.md). All of them use `ILLink` prefix and unique code for easier identification. 
```

src/linker/Linker/Constants.cs
```diff
@@ -4,8 +4,9 @@
 
 namespace Mono.Linker
 {
-	internal static class Constants
+	static class Constants
 	{
-		public const string ILLinker = "ILLinker";
+		// Prefix used in error/warning reporting to specify the tool it comes from
+		public const string ILLink = "ILLink";
```
marek-safar, line 10: Please update also documentation references. There are a few places where we use ILLinker

docs/illink-tasks.md
vitek-karas: I ended up using a variation of what @marek-safar suggested - thanks!

PR #1320, opened 2020-07-07
Show error for unresolved root assemblies

sbomer:
Fixes https://github.com/mono/linker/issues/1319. This will also help with https://github.com/mono/linker/issues/811.

This has two parts:
1. Changes the nullref exception into the following:
```
Unhandled exception. Mono.Cecil.AssemblyResolutionException: Failed to resolve assembly: 'System.Threading.Tasks.dll, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
   ---> System.IO.FileNotFoundException: Unable to find 'System.Threading.Tasks.dll.dll' or 'System.Threading.Tasks.dll.exe' file
     at Mono.Linker.DirectoryAssemblyResolver.Resolve(AssemblyNameReference name, ReaderParameters parameters)
     at Mono.Linker.AssemblyResolver.Resolve(AssemblyNameReference name, ReaderParameters parameters)
     at Mono.Linker.LinkContext.Resolve(IMetadataScope scope)
     at Mono.Linker.LinkContext.Resolve(String name)
     at Mono.Linker.Steps.ResolveFromAssemblyStep.Process()
     at Mono.Linker.Driver.Execute(String[] args, ILogger customLogger)
```
which makes it more obvious that it is treating the extension as part of the assembly name.

2. Surfaces unhandled exceptions like the above by rethrowing them. Previously they were [passed](https://github.com/mono/linker/blob/master/src/linker/Linker/Driver.cs#L705) to `LogError`, which currently doesn't log them without `--verbose`, and even with `--verbose` will only show up with `diagnostic` MSBuild verbosity settings. With this change, it will be rethrown, logged to stderr, and shown with `High` importance by the [task](https://github.com/mono/linker/blob/master/src/ILLink.Tasks/LinkTask.cs#L198), causing it to show up with the default `minimal` MSBuild verbosity settings.

### DIFF
--- a/src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
+++ b/src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // ResolveFromAssemblyStep.cs
 //
 // Author:
@@ -64,15 +64,12 @@ namespace Mono.Linker.Steps
 				Context.Resolver.CacheAssembly (_assembly);
 
 			var ignoreUnresolved = Context.Resolver.IgnoreUnresolved;
-			if (_rootVisibility == RootVisibility.PublicAndFamily) {
-				Context.Resolver.IgnoreUnresolved = false;
-			}
-
+			Context.Resolver.IgnoreUnresolved = false;
 			AssemblyDefinition assembly = _assembly ?? Context.Resolve (_file);
 			Context.Resolver.IgnoreUnresolved = ignoreUnresolved;
-			if (_rootVisibility != RootVisibility.Any && HasInternalsVisibleTo (assembly)) {
+
+			if (_rootVisibility != RootVisibility.Any && HasInternalsVisibleTo (assembly))
 				_rootVisibility = RootVisibility.PublicAndFamilyAndAssembly;
-			}
 
 			switch (assembly.MainModule.Kind) {
 			case ModuleKind.Dll:
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -52,6 +52,7 @@ namespace Mono.Linker
 		public static int Main (string[] args)
 		{
 			if (args.Length == 0) {
+				Usage ();
 				Console.Error.WriteLine ("No parameters specified");
 				return 1;
 			}
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -28,6 +28,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Reflection;
 using System.Text;
@@ -50,11 +51,6 @@ namespace Mono.Linker
 
 		public static int Main (string[] args)
 		{
-			return Execute (args);
-		}
-
-		public static int Execute (string[] args, ILogger customLogger = null)
-		{
 			if (args.Length == 0) {
 				Console.Error.WriteLine ("No parameters specified");
 				return 1;
@@ -64,18 +60,13 @@ namespace Mono.Linker
 				return 1;
 
 			try {
-
 				using (Driver driver = new Driver (arguments)) {
-					if (!driver.Run (customLogger))
-						return 1;
+					return driver.Run ();
 				}
-
 			} catch {
 				Console.Error.WriteLine ("Fatal error in {0}", _linker);
 				throw;
 			}
-
-			return 0;
 		}
 
 		readonly Queue<string> arguments;
@@ -684,13 +675,18 @@ namespace Mono.Linker
 			return 0;
 		}
 
-		public bool Run (ILogger customLogger = null)
+		// Returns the exit code of the process. 0 indicates success.
+		// Known non-recoverable errors (LinkerFatalErrorException) set the exit code
+		// to the error code.
+		// May propagate exceptions, which will result in the process getting an
+		// exit code determined by dotnet.
+		public int Run (ILogger customLogger = null)
 		{
 			int setupStatus = SetupContext (customLogger);
 			if (setupStatus > 0)
-				return true;
+				return 0;
 			if (setupStatus < 0)
-				return false;
+				return 1;
 
 			Pipeline p = context.Pipeline;
 			PreProcessPipeline (p);
@@ -700,7 +696,9 @@ namespace Mono.Linker
 			} catch (LinkerFatalErrorException lex) {
 				context.LogMessage (lex.MessageContainer);
 				Console.Error.WriteLine (lex.ToString ());
-				return false;
+				Debug.Assert (lex.MessageContainer.Category == MessageCategory.Error);
+				Debug.Assert (lex.MessageContainer.Code != null);
+				return lex.MessageContainer.Code ?? 1;
 			} catch (Exception) {
 				// Unhandled exceptions are usually linker bugs. Ask the user to report it.
 				context.LogError ($"IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues", 1012);
@@ -711,7 +709,7 @@ namespace Mono.Linker
 				context.Tracer.Finish ();
 			}
 
-			return true;
+			return 0;
 		}
 
 		partial void PreProcessPipeline (Pipeline pipeline);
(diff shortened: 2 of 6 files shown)


src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
```diff
@@ -64,15 +64,12 @@ protected override void Process ()
 				Context.Resolver.CacheAssembly (_assembly);
 
 			var ignoreUnresolved = Context.Resolver.IgnoreUnresolved;
-			if (_rootVisibility == RootVisibility.PublicAndFamily) {
-				Context.Resolver.IgnoreUnresolved = false;
-			}
-
+			Context.Resolver.IgnoreUnresolved = false;
 			AssemblyDefinition assembly = _assembly ?? Context.Resolve (_file);
```
sbomer, line 68: This will treat `-a` the same as we treat `-r` today. There may be more room for improvement here. Ideally we would have a `Resolve` method that lets the caller decide what to do on resolution failure.

src/linker/Linker/Driver.cs
```diff
@@ -697,15 +697,16 @@ public bool Run (ILogger customLogger = null)
...
-				if (ex is LinkerFatalErrorException lex) {
-					context.LogMessage (lex.MessageContainer);
-					Console.Error.WriteLine (ex.ToString ());
-				} else {
-					context.LogError ($"IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues \n{ex}", 1012);
-				}
-
+			} catch (LinkerFatalErrorException lex) {
+				context.LogMessage (lex.MessageContainer);
+				Console.Error.WriteLine (lex.ToString ());
```
sbomer, line 702: @vitek-karas @mateoatr  Should we do the same thing for `LinkerFatalErrorException`? I don't fully understand why we have this exception type - I thought that its purpose was to carry a `MessageContainer` for non-recoverable but fixable (by the user) errors, but it also has a [ctor](https://github.com/mono/linker/blob/master/src/linker/Linker/LinkerFatalErrorException.cs#L14) that takes a string that suggests it is to be used for non-fixable errors. Why don't we just use standard .NET exception types for internal linker errors?

sbomer: There is a benefit to logging the stacktrace instead of rethrowing here - `LinkerFatalErrorException` will have an exit code of 1, instead of 134 for the unhandled exceptions, allowing us to distinguish based on the exit code (in telemetry for example). This is why I would like to clarify the meaning of `LinkerFatalErrorException`.

sbomer:
I will add tests once I get feedback - I had a couple of questions about the approach.

src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
vitek-karas: I find this weird: the IgnoreUnresolved is a user facing setting, and we're deliberately ignore it here. I guess I would have to understand why we're doing this. That said, if we get a desired behavior... I'm fine with it.

src/linker/Linker/Driver.cs
vitek-karas: I also thought that the purpose of `LinkerFatalErrorException` was to carry the `MessageContainer` (we want this so that we can for example specify the origin). But as the name suggests, it's a fatal - non-recoverable error. 

We might want to let it go unhandled, but there are pros and cons:
* Pro - it will trigger Watson and potentially give a chance to get detailed info about the failure through that
* Cons - we might lose the nice formatting - not sure how it will behave with regard to MSBuild (for example, will it get the right origin?)

If we don't rethrow, we can use the message code as the exit code for linker, that way we get nice telemetry. This is probably the best solution - if we swallow it fully we get exit code 1. If we rethrow we get exit code 134 (or whatever), but no details. So catching and using the message code as the exit code gives us most info.

mateoatr: Non-recoverable but user-fixable errors should be logged using `LogError`, the `LinkerFatalErrorException` is used to throw whenever the linker ends up in a place where it shouldn't (and the user can do nothing about it). I think there are a few places where we're using this exception without specifying an origin (and we can/should, e.g., [here](https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/MarkStep.cs#L2454)).

I like @vitek-karas idea of catching and using the error code as the exit code to get more informative telemetry about liker failures.

sbomer: I took the suggestion to return the error code - PTAL.

src/linker/Linker.Steps/ResolveFromAssemblyStep.cs
sbomer: Agreed. There was some discussion of it in https://github.com/mono/linker/pull/929. I looked into making `Resolve` more flexible, but it looks to be more effort than it's worth right now.

src/linker/Linker/Driver.cs
```diff
@@ -49,11 +50,6 @@ public partial class Driver : IDisposable
...
 
 		public static int Main (string[] args)
-		{
-			return Execute (args);
-		}
-
-		public static int Execute (string[] args, ILogger customLogger = null)
 		{
 			if (args.Length == 0) {
 				Console.Error.WriteLine ("No parameters specified");
```
mateoatr, line 55: Should we display `Usage()` in this case?

vitek-karas: Nit: Should this be the other way round? First print out the error and then usage?

src/linker/Linker/LinkerFatalErrorException.cs
```diff
@@ -31,9 +20,6 @@ public LinkerFatalErrorException (MessageContainer message)
 			MessageContainer = message;
```
vitek-karas, line 20: Nit: We should also check that `Code != null`

src/linker/Linker/Driver.cs
```diff
@@ -684,33 +676,41 @@ protected int SetupContext (ILogger customLogger = null)
...
-					context.LogError ($"IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues \n{ex}", 1012);
-				}
-
-				return false;
+			} catch (LinkerFatalErrorException lex) {
+				context.LogMessage (lex.MessageContainer);
+				Console.Error.WriteLine (lex.ToString ());
+				Debug.Assert (lex.MessageContainer.Category == MessageCategory.Error);
+				Debug.Assert (lex.MessageContainer.Code != null);
+				return lex.MessageContainer.Code ?? 1;
```
marek-safar, line 702: Do we have a check that Code is never set to 0 ?

src/linker/Linker/LinkerFatalErrorException.cs
sbomer: This should be covered by the category check since the only way to create an error `MessageContainer` is with a ctor that requires a code - but an extra check doesn't hurt and might prevent future bugs.

src/linker/Linker/Driver.cs
sbomer: The `MessageContainer` ctors that set Code ensure it's in a valid range.

src/linker/Linker/Driver.cs
sbomer: I think usage first makes sense so that the error isn't hidden by the large amount of usage output. If we want to avoid printing usage, maybe we should instead support standard `-h` and `--help` options? Right now it takes some guessing to stumble on `-?` or `-help` for someone not familiar with the tool.

vitek-karas: Sure - it was jut a nit.

Re printing out usage at all - I think we should print out something. Currently running linker without any command line doesn't provide any actionable message. "No parameters specified" -> "Well yeah, I know, I just did that, thanks for telling me... but what am I supposed to do instead?".

I can see the argument of not printing Usage when there's an error in the command line, but empty command line should be a special case. I personally would like to see Usage, but if not that, then I agree with @sbomer that we should at the very least print out how to get to the usage (although it feels like unnecessary two step process).

src/linker/Linker/LinkerFatalErrorException.cs
vitek-karas: I was asking purely because we assert on it when consuming the exception...

src/linker/Linker/Driver.cs
sbomer: I think the situation is improved with `-h` and `--help`. I opened https://github.com/mono/linker/issues/1353 where we can discuss any follow-ups.

PR #1324, opened 2020-07-08
Add nowarn option

mateoatr:
This PR adds the new `--nowarn` option that behaves similarly to the Roslyn option of the same name. The user can turn off warning messages by specifying an ILLinker warning code. Some examples:

```
--nowarn IL2001
--nowarn IL2002;IL2003;IL2004
--nowarn IL2005,IL2006;IL2007
--nowarn IL2008,IL2009;IL2010
--nowarn "IL2011 IL2012,IL2013"
```

It also adds a new warning subcategory, `TrimCorrectness`, and a new ILLinkTask property, `NoWarn`.

This is the set of warnings that will be turned off whenever `TrimmerAnalysisWarnings` is set to true:

| Code | Message                                                                                                                                         | Subcategory                   |
|------|-------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------|
| 2006 | -                                                                                                                                               | TrimCorrectness |
| 2026 | Calling method annotated with RequiresUnreferencedCodeAttribute                                                                                 | TrimCorrectness              |
| 2043 | Trying to propagate DynamicallyAccessedMemberAttribute from property 'property' to its getter 'method', but it already has such attribute.      | TrimCorrectness |

/cc @agocke @eerhardt


src/ILLink.Tasks/LinkTask.cs
```diff
@@ -303,6 +316,13 @@ protected override string GenerateResponseFileCommands ()
 			if (OutputDirectory != null)
 				args.Append ("-out ").AppendLine (Quote (OutputDirectory.ItemSpec));
 
+			if (NoWarn != null)
+				args.Append ($"--nowarn \"{NoWarn}\"");
+
+			if (_trimmerAnalysisWarnings is bool trimmerAnalysisWarnings &&
+				trimmerAnalysisWarnings == true)
+				args.Append ("--nowarn 2006;2026;2037;2043");
```
vitek-karas, line 324: The command line parser for linker below doesn't handle multiple `--nowarn` options, the last one will win. So the user won't be able to combine the analysis warning disable with any other nowarn.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -51,11 +51,24 @@ public class ILLink : ToolTask
...
 		/// </summary>
 		[Required]
 		public ITaskItem OutputDirectory { get; set; }
 
+		/// <summary>
+		/// Turns off trim correctness analysis warnings.
+		/// Maps to '--nowarn 2006;2026;2037;2043'.
+		/// </summary>
+		public bool TrimmerAnalysisWarnings { set => _trimmerAnalysisWarnings = value; }
+		bool? _trimmerAnalysisWarnings;
```
vitek-karas, line 64: I just wanted to start the discussion where should the list of "trimmer analysis warnings" be defined.
Possibilities:
* linker itself
* ILLink task
* ILLink target

I personally don't want to put it into linker itself. But I'm split on whether it should be in the task or the target. Target gives us potentially greater flexibility with regard to versioning and so on. The task is better since it lives in the same repo as the linker.

@sbomer any suggestions?

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -51,11 +51,24 @@ public class ILLink : ToolTask
...
+		///    Maps to '-out'.
 		/// </summary>
 		[Required]
 		public ITaskItem OutputDirectory { get; set; }
 
+		/// <summary>
+		/// Turns off trim correctness analysis warnings.
+		/// Maps to '--nowarn 2006;2026;2037;2043'.
+		/// </summary>
+		public bool TrimmerAnalysisWarnings { set => _trimmerAnalysisWarnings = value; }
```
vitek-karas, line 63: Regardless of where this should live the name needs to change.
Currently we have "Trim correctness" analysis. In the future I anticipate "Single-file correctness", "AOT correctness" and potentially more. The name must reflect this.

I would probably call it `EnableTrimCorrectnessAnalysis` or `TrimCorrectnessAnalysis` (whichever fits the current style in SDK better). It would be "false" by default for now.

src/ILLink.Tasks/LinkTask.cs
vitek-karas, line 324: As noted in the parser below, linker should only accept prefixed codes, so `IL2006;IL2026;IL2037;IL2043`.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -162,12 +162,14 @@ TypeAnnotations BuildTypeAnnotations (TypeDefinition type)
 								paramAnnotations[0] = methodMemberTypes;
 							}
 						} else if (methodMemberTypes != DynamicallyAccessedMemberTypes.None) {
-							_context.LogWarning ($"The DynamicallyAccessedMembersAttribute is only allowed on method parameters, return value or generic parameters.", 2041, method);
+							_context.LogWarning ($"The DynamicallyAccessedMembersAttribute is only allowed on method parameters, return value or generic parameters.",
+								2041, method, subcategory: MessageSubCategory.TrimCorrectness);
```
vitek-karas, line 166: I would argue that this warning should not be part of the TrimCorrectness subcategory. It warns about invalid usage of the attribute - in itself it doesn't affect correctness. In my mind this is similar to invalid XML in the descriptor.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -247,7 +249,8 @@ TypeAnnotations BuildTypeAnnotations (TypeDefinition type)
 						}
 
 						if (annotatedMethods.Any (a => a.Method == setMethod)) {
-							_context.LogWarning ($"Trying to propagate DynamicallyAccessedMemberAttribute from property '{property.FullName}' to its setter '{setMethod.GetDisplayName ()}', but it already has such attribute on the 'value' parameter.", 2043, setMethod);
+							_context.LogWarning ($"Trying to propagate DynamicallyAccessedMemberAttribute from property '{property.FullName}' to its setter '{setMethod.GetDisplayName ()}', but it already has such attribute on the 'value' parameter.",
+								2043, setMethod, subcategory: MessageSubCategory.TrimCorrectness);
```
vitek-karas, line 253: Same as above - this is about correct usage of the attribute, not about trim correctness.

src/ILLink.Tasks/LinkTask.cs
vitek-karas: I also find it weird that this set doesn't match the set of warnings in the new TrimCorrectness subcategory. Maybe I don't understand the intent of the subcategory...

src/linker/Linker/Driver.cs
```diff
@@ -715,6 +723,29 @@ public bool Run (ILogger customLogger = null)
...
+					return arg.Substring (1, arg.Length - 2);
+
+				return arg;
+			}
+
+			value = Unquote (value);
+			string[] values = value.Split (new char[] { ',', ';', ' ' }, StringSplitOptions.RemoveEmptyEntries);
+			HashSet<uint> noWarnCodes = new HashSet<uint> ();
+			foreach (string id in values) {
+				var warningCode = id.StartsWith ("IL") ? id.Substring (2) : id;
```
vitek-karas, line 740: Linker must ONLY recognize prefixed warning codes - so `IL1234` is recognized as linker warning. Everything else in the nowarn must be ignored and linker should not fail on it:
* The NoWarn property is shared between multiple tools, CSC being the most prominent one. It will contain stuff for multiple tools - we must filter out the ones intended for linker and ignore the rest
* For historical reasons, unprefixed numbers in NoWarn are recognized as CS warnings, linker should not "take over" these in any way

There's a separate discussion if this logic should be in the linker or the ILLink task as both could handle this. (Technically it could also be in ILLink target, but it's difficult to implement the logic in MSBuild).

I personally would prefer we don't do "preprocessing" of values in the ILLink task - so I would prefer we keep this in the linker.

src/linker/Linker/Driver.cs
```diff
@@ -715,6 +723,29 @@ public bool Run (ILogger customLogger = null)
...
+			string Unquote (string arg)
+			{
+				if (arg.Length > 1 && arg[0] == '"' && arg[arg.Length - 1] == '"')
+					return arg.Substring (1, arg.Length - 2);
+
+				return arg;
+			}
+
+			value = Unquote (value);
+			string[] values = value.Split (new char[] { ',', ';', ' ' }, StringSplitOptions.RemoveEmptyEntries);
```
vitek-karas, line 737: Can you check what CSC supports as separators? Would be good to be consistent.

src/linker/Linker/LoggingReflectionPatternRecorder.cs
```diff
@@ -45,7 +45,7 @@ public void RecognizedReflectionAccessPattern (IMemberDefinition source, Instruc
 		public void UnrecognizedReflectionAccessPattern (IMemberDefinition source, Instruction sourceInstruction, IMetadataTokenProvider accessedItem, string message)
 		{
 			var origin = new MessageOrigin (source, sourceInstruction?.Offset);
-			_context.LogWarning (message, 2006, origin, "Unrecognized reflection pattern");
+			_context.LogWarning (message, 2006, origin, MessageSubCategory.UnrecognizedReflectionPattern);
```
vitek-karas, line 48: I think this should be TrimCorrectness subcategory. Unrecognized reflection is a problem for trim correctness, but not other types of analysis.

src/ILLink.Tasks/LinkTask.cs
sbomer: My understanding of the conclusion (https://github.com/mono/linker/pull/1303#issuecomment-653433498) was that we were only going to add an option for `--nowarn`, and not to design the option for disabling categories yet. They would be disabled from the MSBuild targets via the `NoWarn` property, not in the task itself.

If we were to introduce a warning category option, I think it should go in the linker, since I have been thinking of it as enabling or disabling different "linker analyzers". For versioning, we could always pass a TFM from the SDK into the linker, which the linker can use in its decisions about what to output.

src/ILLink.Tasks/LinkTask.cs
sbomer: I think the warnings should be disabled from MSBuild [targets](https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L99) until we properly define the categories. It still might be nice to allow multiple `--nowarn` options.

src/ILLink.Tasks/LinkTask.cs
sbomer: My understanding is that this should live in the SDK only for now, not in the task. For naming, I personally find `correctness` a little redundant. What do you think of `TrimAnalysis`?

Maybe it would make sense to preemptively open an SDK PR that introduces the option to discuss naming there, and avoid getting caught up in it here? I'm happy to do that if you think it makes sense @vitek-karas @mateoatr

src/ILLink.Tasks/LinkTask.cs
mateoatr: I used the `TrimCorrectness` subcategory in all warnings related to trimming analysis and the usage of attributes related to it. But I think you're right that warnings about invalid usage of the attributes are not `TrimCorrectness` warnings, so I'm changing this. I'm also dropping the IL2037 from the list of `TrimCorrectness` warnings since I think we should always warn on unresolved members.

With this, the set of warnings we turn off by default is the set of warnings that use the `TrimCorrectness` subcategory.

src/linker/Linker/Driver.cs
mateoatr: This is the same set that they use for separators: https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/CommandLine/CSharpCommandLineParser.cs#L1956

src/ILLink.Tasks/LinkTask.cs
mateoatr: > Maybe it would make sense to preemptively open an SDK PR that introduces the option to discuss naming there, and avoid getting caught up in it here? I'm happy to do that if you think it makes sense @vitek-karas @mateoatr

This sounds good. I'm dropping this property from here then. I think `EnableTrimAnalysis` would be a good name, but we can follow up on this conversation in the SDK.

src/linker/Linker/Driver.cs
```diff
@@ -998,6 +1032,7 @@ static void Usage ()
 			Console.WriteLine ("  -out PATH           Specify the output directory. Defaults to 'output'");
 			Console.WriteLine ("  --about             About the {0}", _linker);
 			Console.WriteLine ("  --verbose           Log messages indicating progress and warnings");
+			Console.WriteLine ("  --nowarn WARN-LIST  Disable specific warning messages.");
```
marek-safar, line 1035: redundant .

src/linker/Linker/MessageSubcategory.cs
```diff
@@ -7,6 +7,7 @@ namespace Mono.Linker
 	public static class MessageSubCategory
 	{
 		public const string None = "";
+		public const string TrimCorrectness = "Trim correctness";
```
marek-safar, line 10: What is this for? I don't think this term is broadly understandable

src/linker/Linker/Driver.cs
```diff
@@ -438,6 +438,14 @@ protected int SetupContext (ILogger customLogger = null)
 						context.OutputWarningSuppressions = true;
 						continue;
 
+					case "--nowarn":
```
marek-safar, line 441: Please update the documentation page with this new option

src/linker/Linker/MessageSubcategory.cs
mateoatr: I agree it might not be very informative to all users, but its main purpose is to group all warnings related to trimming analysis. Going forward, if we ever allow subcategory mapping in `nowarn` (things like `--nowarn trimcorrectness` = `--nowarn IL2006;IL2026;IL2043`) it might come handy. What do you think could be a more informative/useful subcategory name/message?

src/linker/Linker/Driver.cs
mateoatr: This is the same message you get from `csc`.

marek-safar, line 1035:
```suggestion
			Console.WriteLine ("  --nowarn WARN-LIST  Disable specific warning messages");
```

src/linker/Linker/MessageSubcategory.cs
marek-safar, line 10:
```suggestion
		public const string Correctness = "correctness";
```

tlakollo: I think something like "trimalert" for the flag and "Trim alert" for subcategory printing is more readable.
`--nowarn trimalert` explains the purpose
`Trim alert warning IL2006 : Mono.Linker.Tests....` 
No warn on correctness sounds like the program warns when is correct and we are disabling it. Also, since it's a broader term sounds like it should have a lot more warnings inside the subcategory.

docs/illink-options.md
```diff
@@ -147,6 +147,12 @@ control options listed below.
 | Substition  | ILLink.Substitution.xml  |  --ignore-substitutions |
 | LinkAttributes  | ILLink.LinkAttributes.xml    |  --ignore-link-attributes |
 
+### Turning off warnings
+
+The `--nowarn` option prevents the linker from displaying one or more linker warnings by
+specifying its warning codes. All warning codes must be prepended with `IL` and multiple
+warnings should be separated with a comma or semicolon.
```
tlakollo, line 154: Missing separated by spaces

mateoatr: Although it's possible to separate warnings with spaces I think we shouldn't put that in the docs to favor using commas and semicolons.

src/ILLink.Tasks/LinkTask.cs
vitek-karas: @sbomer Please prepare the SDK change (you can do that fastest)... `EnableTrimAnalysis` sounds good - please include Sam in the discussion on the SDK side.

docs/illink-options.md
vitek-karas: I also think that our command line parser will probably have trouble with spaces used that way...

src/linker/Linker/Driver.cs
```diff
@@ -715,6 +723,32 @@ public bool Run (ILogger customLogger = null)
...
+					return arg.Substring (1, arg.Length - 2);
+
+				return arg;
+			}
+
+			value = Unquote (value);
+			string[] values = value.Split (new char[] { ',', ';', ' ' }, StringSplitOptions.RemoveEmptyEntries);
+			HashSet<uint> noWarnCodes = new HashSet<uint> ();
+			foreach (string id in values) {
+				if (!id.StartsWith ("IL"))
```
vitek-karas, line 740: This should be `StartsWith ("IL", StringComparison.Ordinal)`

src/linker/Linker/MessageSubcategory.cs
vitek-karas: Given the discussion on the public property - maybe "TrimAnalysis" ?

docs/illink-options.md
mateoatr: Sapces will only work if the warning list is quoted. Do you think we should remove the ability to use spaces/quote the warning list?

src/linker/Linker/MessageSubcategory.cs
mateoatr: I think `TrimAnalysis` is a good option given that we'll probably have an `EnableTrimAnlaysis` property in the SDK. `Correctness` is even broader than `TrimCorrectness`.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -51,11 +51,17 @@ public class ILLink : ToolTask
...
+		///    Maps to '-out'.
 		/// </summary>
 		[Required]
 		public ITaskItem OutputDirectory { get; set; }
 
+		/// <summary>
+		/// The subset of warnings that have to be turned off. 
+		/// Maps to '--nowarn'.
+		/// </summary>
+		public string[] NoWarn { get; set; }
```
sbomer, line 63: This should be a `string` - since `NoWarn` is a single MSBuild property.

src/ILLink.Tasks/LinkTask.cs
sbomer: https://github.com/dotnet/sdk/pull/12388 to get the discussion started. I will add tests and such as we go.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -303,6 +309,9 @@ protected override string GenerateResponseFileCommands ()
 			if (OutputDirectory != null)
 				args.Append ("-out ").AppendLine (Quote (OutputDirectory.ItemSpec));
 
+			if (NoWarn != null)
+				args.Append ($"--nowarn \"{NoWarn}\"");
```
sbomer, line 313:
```suggestion
				args.Append ("--nowarn ").AppendLine (Quote (NoWarn));
```
Mostly to follow conventions. The important bit is that it needs whitespace after the option, to prevent being combined with the next option.

Could you also add a test in `ILLink.Tasks.Tests.cs`?

docs/illink-options.md
vitek-karas: I don't mind supporting it (especially if it's the same as csc)...

PR #1325, opened 2020-07-08
Add handling of calli in MethodBodyScanner

MichalStrehovsky:
We now need this to link CoreLib.

https://github.com/dotnet/runtime/pull/32520/files#r451338721

### DIFF
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -506,8 +506,22 @@ namespace Mono.Linker.Dataflow
 					NewKnownStack (knownStacks, ((Instruction) operation.Operand).Offset, currentStack, methodBody);
 					break;
 
-				case Code.Calli:
-					// TODO: currently not emitted by any mainstream compilers but we should implement
+				case Code.Calli: {
+						var signature = (CallSite) operation.Operand;
+						if (signature.HasThis && !signature.ExplicitThis) {
+							PopUnknown (currentStack, 1, methodBody, operation.Offset);
+						}
+
+						// Pop arguments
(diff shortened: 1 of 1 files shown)


marek-safar:
It'd be great to have this covered by test (it's new scenario and can break with other things). We could bump C# compiler dependency to make it easier

MichalStrehovsky:
> It'd be great to have this covered by test (it's new scenario and can break with other things). We could bump C# compiler dependency to make it easier

AFAIK this is still only available in the Roslyn myget feed (it wasn't part of the preview releases yet) - would we be okay with grabbing a compiler from there?

I really wanted to avoid writing IL tests for these because they're a pain to maintain (we have a bunch of default interface methods tests in CoreCLR repo that are in IL and all the reasons why they were in IL are now gone, but I'll have to deal with them for the IL Linker work now and it's just annoying...).

marek-safar:
> AFAIK this is still only available in the Roslyn myget feed

Are you sure? I think Roslyn comes from Arcade in our case so if this code which exposed this bug is about to land into dotnet/runtime you should be able to write the same code here as we sync on arcade dependency.

MichalStrehovsky:
The Roslyn we're using to build the tests is some old pre-C# 8 version. I'll see if I can bump it.

https://github.com/mono/linker/blob/cf697fca5162e2a87c77226798c67b42f6828178/test/Mono.Linker.Tests/Mono.Linker.Tests.csproj#L38

marek-safar:
I forgot we use CodeDom compiler, that might make it a bit tricky

MichalStrehovsky:
Hm, it's a bit more involved to get the right version of the package. I'll merge as-is so that people who're waiting in CoreLib are not blocked by this. If we get to adding tests for covariant returns, we can add testing for this too. I don't want to write either of them in IL because it's a maintenance pain.

Issue #1321, opened 2020-07-07
TypeReferenceExtensions.GetDisplayNameWithoutNamespace doesn't handle nested types inside generic types


eerhardt:
When linking the [CarChecker .NET 5 Blazor app](https://github.com/SteveSandersonMS/CarChecker/tree/net5) with `--feature System.Diagnostics.Tracing.EventSource.IsSupported false` I'm getting an exception with the latest mono/linker master code:

```
ILLinker: error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues
System.FormatException: Input string was not in a correct format.
   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
   at System.Number.ParseInt32(ReadOnlySpan`1 value, NumberStyles styles, NumberFormatInfo info)
   at System.Int32.Parse(String s)
   at Mono.Linker.TypeReferenceExtensions.GetDisplayNameWithoutNamespace(TypeReference type) in F:\git\linker\src\linker\Linker\TypeReferenceExtensions.cs:line 34
   at Mono.Linker.MethodReferenceExtensions.GetDisplayName(MethodReference method) in F:\git\linker\src\linker\Linker\MethodReferenceExtensions.cs:line 19
   at Mono.Linker.Steps.RemoveUnreachableBlocksStep.RewriteBody(MethodDefinition method) in F:\git\linker\src\linker\Linker.Steps\RemoveUnreachableBlocksStep.cs:line 143
   at Mono.Linker.Steps.RemoveUnreachableBlocksStep.RewriteBodies(Collection`1 types) in F:\git\linker\src\linker\Linker.Steps\RemoveUnreachableBlocksStep.cs:line 119
   at Mono.Linker.Steps.RemoveUnreachableBlocksStep.Process() in F:\git\linker\src\linker\Linker.Steps\RemoveUnreachableBlocksStep.cs:line 45
   at Mono.Linker.Steps.BaseStep.Process(LinkContext context) in F:\git\linker\src\linker\Linker.Steps\BaseStep.cs:line 60
   at Mono.Linker.Pipeline.ProcessStep(LinkContext context, IStep step) in F:\git\linker\src\linker\Linker\Pipeline.cs:line 135
   at Mono.Linker.Pipeline.Process(LinkContext context) in F:\git\linker\src\linker\Linker\Pipeline.cs:line 128
[...]

PR #1328, opened 2020-07-08
Fix problem with GetDisplayName on nested generic types

vitek-karas:
I changed the algorithm to not rely on teh backtick notation (parsing numbers from names) and instead use Cecil's generic parameter counts.

This also changes how instantiated generics are handled (GenericInstanceType) - now we just store the generic arguments and loop again on the open generic type (ElementType) which will apply the generic arguments if available.

Added test for the failing case as well as another one for partially instantiated generics.

Related https://github.com/mono/linker/issues/1321


vitek-karas:
I'll merge this to fix the failure - changing the formatting due to potentially unresolved type refs will be handled in a separate change.

PR #1330, opened 2020-07-08
Keep all fields on sequential/explicit layout classes if a field was kept

MichalStrehovsky:
Should fix the [issue](https://github.com/mono/linker/pull/1309) observed in CoreLib where `RawData` class is used with `Unsafe.As` to break type safety and linker stripped its fields because it's not seen as instantiated.


MichalStrehovsky:
> It'd be nice to keep only fields up to the field which is used (for sequential layout) but that can be done later

I would be wary of that because if the class is not sealed, we would be affecting the derived classes (that would then have the same problem as here).

It could also cause the class to have different alignment requirement (e.g. if the stripped fields happen to be something like `long` or `double`) - not sure whether that matters in classes, but something to think of.

This fix actually already has the same problem but from the other side - we need to walk the base hierarchy and mark all fields on the base types as well.

Let's not merge this yet.

MichalStrehovsky:
Merging now to unblock runtime integration.

PR #1335, opened 2020-07-09
Adds manual linker attributes removed for assembly and module level attributes

marek-safar:


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -540,14 +540,14 @@ void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo
 				if (ProcessLinkerSpecialAttribute (ca, provider, reason, sourceLocationMember))
 					continue;
 
+				if (_context.Annotations.HasLinkerAttribute<RemoveAttributeInstancesAttribute> (ca.AttributeType.Resolve ()))
```
tlakollo, line 543: Just double checking this will only print once per type instead of once per typeref right? Is this the desired behavior @vitek-karas?

marek-safar: I don't fully understand the comment but moved the logic back to unblock other bug fixes. With the current code, the attribute removal logic has 2 very different code paths which I don't think is desirable.

Issue #1341, opened 2020-07-09
Removing attributes breaks when an assembly is set to "copy"


eerhardt:
Using the `ILLink.LinkAttributes.xml` file from https://github.com/dotnet/runtime/pull/39000 and linking the [CarChecker app](https://github.com/SteveSandersonMS/CarChecker/tree/net5), I'm getting the following error:

```
Mono.Linker.LinkerFatalErrorException: ILLinker: error IL1011: Failed to write 'obj\Release\net5.0\browser-wasm\linked\CarChecker.Client.dll
 ---> Mono.Cecil.ResolutionException: Failed to resolve System.ComponentModel.EditorBrowsableState
   at Mono.Cecil.Mixin.CheckedResolve(TypeReference self)
   at Mono.Cecil.SignatureWriter.WriteCustomAttributeEnumValue(TypeReference enum_type, Object value)
   at Mono.Cecil.SignatureWriter.WriteCustomAttributeValue(TypeReference type, Object value)
   at Mono.Cecil.SignatureWriter.WriteCustomAttributeElement(TypeReference type, CustomAttributeArgument argument)
   at Mono.Cecil.SignatureWriter.WriteCustomAttributeConstructorArguments(CustomAttribute attribute)
   at Mono.Cecil.MetadataBuilder.GetCustomAttributeSignature(CustomAttribute attribute)
   at Mono.Cecil.MetadataBuilder.AddCustomAttributes(ICustomAttributeProvider owner)
   at Mono.Cecil.MetadataBuilder.AddProperty(PropertyDefinition property)
   at Mono.Cecil.MetadataBuilder.AddProperties(TypeDefinition type)
   at Mono.Cecil.MetadataBuilder.AddType(TypeDefinition type)
   at Mono.Cecil.MetadataBuilder.AddTypes()
   at Mono.Cecil.MetadataBuilder.BuildTypes()
   at Mono.Cecil.MetadataBuilder.BuildModule()
   at Mono.Cecil.MetadataBuilder.BuildMetadata()
   at Mono.Cecil.ModuleWriter.<>c.<BuildMetadata>b__2_0(MetadataBuilder builder, MetadataReader _)
   at Mono.Cecil.ModuleDefinition.Read[TItem,TRet](TItem item, Func`3 read)
   at Mono.Cecil.ModuleWriter.BuildMetadata(ModuleDefinition module, MetadataBuilder metadata)
[...]

PR #1344, opened 2020-07-10
Attribute removal should not kick in for assemblies which are not linked

vitek-karas:
MarkStep go over all assemblies (Regardless if they're being linked or just copied) and marks everything as usual. But in the end it will not "trim" anything from the copy assembly. So we must make sure that everything the assembly needs is marked.

Attribute removal needs to comply with this behavior as well.

This change fixes several other issues around link attribute XML files:
- If the XML refers to an attribute which will in turn bring in new assemblies to the closure it would not correctly register all of them - the step needs to behave just like any other step which can bring in new assembly to the closure
- Several places in MarkStep didn't use the CustomAttributeSource to detect presense of attributes - meaning that injecting attributes via the XML on a member which had no other (IL based) attributes didn't actually do anything.
- Also fixed the annotation's handling of the same issue - presence of custom attributes needs to be determined via the CustomAttributeSource in any place where attributes which are recognized by the linker as special are handled.

Added tests for all of the changes above

Fixes https://github.com/mono/linker/issues/1341.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -531,10 +531,11 @@ namespace Mono.Linker.Steps
 
 		void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo reason, IMemberDefinition sourceLocationMember)
 		{
-			if (!provider.HasCustomAttributes)
+			if (!_context.CustomAttributes.HasAttributes (provider))
 				return;
 
-			bool markOnUse = _context.KeepUsedAttributeTypesOnly && Annotations.GetAction (GetAssemblyFromCustomAttributeProvider (provider)) == AssemblyAction.Link;
+			bool providerInLinkedAssembly = Annotations.GetAction (GetAssemblyFromCustomAttributeProvider (provider)) == AssemblyAction.Link;
+			bool markOnUse = _context.KeepUsedAttributeTypesOnly && providerInLinkedAssembly;
 
 			foreach (CustomAttribute ca in provider.CustomAttributes) {
 				if (ProcessLinkerSpecialAttribute (ca, provider, reason, sourceLocationMember))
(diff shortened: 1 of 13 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -531,10 +531,11 @@ void MarkMarshalSpec (IMarshalInfoProvider spec, in DependencyInfo reason, IMemb
 
 		void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo reason, IMemberDefinition sourceLocationMember)
 		{
-			if (!provider.HasCustomAttributes)
+			if (!_context.CustomAttributes.HasAttributes (provider))
```
marek-safar, line 534: Wouldn't be better to have `if (provider.HasCustomAttributes) { ... } ` for the section which iterates `provider.CustomAttributes`. The logic after that will work as it's

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -808,11 +809,13 @@ bool MarkDependencyField (TypeDefinition type, string name, in DependencyInfo re
 
 		void LazyMarkCustomAttributes (ICustomAttributeProvider provider)
 		{
-			if (!provider.HasCustomAttributes)
+			if (!_context.CustomAttributes.HasAttributes (provider))
```
marek-safar, line 812: No need to change this

eerhardt:
I pulled this change and verified it fixes the error I am seeing when trying to use the changes in https://github.com/dotnet/runtime/pull/39000 to link a Blazor application.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: You're right - for now... I reverted the change.

PR #1346, opened 2020-07-10
Check that events are kept when preserving a generic type

mateoatr:
/cc @layomia

### DIFF
--- a/test/Mono.Linker.Tests.Cases/LinkXml/UnusedGenericTypeWithPreserveAllHasAllMembersPreserved.cs
+++ b/test/Mono.Linker.Tests.Cases/LinkXml/UnusedGenericTypeWithPreserveAllHasAllMembersPreserved.cs
@@ -1,4 +1,5 @@
-﻿using Mono.Linker.Tests.Cases.Expectations.Assertions;
+﻿using System;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
 using Mono.Linker.Tests.Cases.Expectations.Metadata;
 
 namespace Mono.Linker.Tests.Cases.LinkXml
@@ -85,6 +86,12 @@ namespace Mono.Linker.Tests.Cases.LinkXml
 			internal static void Method6 ()
 			{
 			}
+
+			[Kept]
+			[KeptBackingField]
+			[KeptEventAddMethod]
+			[KeptEventRemoveMethod]
+			event EventHandler<EventArgs> Event;
 		}
 	}
 }


test/Mono.Linker.Tests.Cases/LinkXml/UnusedTypeWithPreserveAllHasAllMembersPreserved.xml
```diff
@@ -1,5 +1,6 @@
 ﻿<linker>
   <assembly fullname="test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null">
     <type fullname="Mono.Linker.Tests.Cases.LinkXml.UnusedTypeWithPreserveAllHasAllMembersPreserved/Unused" preserve="all" />
+    <type fullname="Mono.Linker.Tests.Cases.LinkXml.UnusedTypeWithPreserveAllHasAllMembersPreserved/UnusedGenericType`1" preserve="all" />
```
tlakollo, line 4: @MichalStrehovsky  I know its probably too late because the descriptor files had been there for a while, but I remember there was a discussion about Generic Methods not having mangled names. Just curious because is something that going forward will differ from LinkAttributes xml file and the Descriptor one. 
https://github.com/mono/linker/pull/1176/files/4bf49e87876f206405df3489823f283ff4abd16d#r421475580

MichalStrehovsky: This seems to refer to a type, not a method - can you clarify what you're referring to?

The C# compiler doesn't include the grave accent in method names and there are no grave accents in the metadata.

tlakollo: Well, the thing I was divided into two things, knowing if this case also applied for not having support for mangled names. I guess by what you write type do support mangled names. The second is that ResolveFromXmlStep has a function GetMethodSignature that is looking for mangled names for methods
https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/ResolveFromXmlStep.cs#L239

MichalStrehovsky: Yes, the C# compiler mangles the names of types because you can have `class Foo { }`, `class Foo<T> { }` and `class Foo<T, U> { }` in the same assembly and types need to have unique names. So the C# compiler appends the grave accent so that it's possible to disambiguate.

Method names don't get mangled by the C# compiler because they're uniquely identified by name + signature. The number of generic parameters is part of the signature.

That particular formatting you're linking to is super weird (I think e.g. methods that take `in` parameters are going to look very interesting in this format). I guess we're stuck with it for where this is used.

mateoatr:
This was already being tested. This PR simply adds a check that we're keeping events.

PR #1357, opened 2020-07-14
Add support for default interface methods

### LINKED ISSUE #427: Support for default interface methods
.NET Core 3.0 (and possibly [.NET Standard 2.1](https://github.com/dotnet/standard/pull/1019)) is adding support for non-abstract instance methods on interfaces in support of the C# 8.0 language feature. [Spec here](https://github.com/dotnet/coreclr/pull/21564).

IL Linker will need to understand what this semantic means and:

* Be able to scan instance methods on interfaces
* If IL Linker has any sort of interface method resolution logic, this logic will need to be updated to understand the resolution to default interface methods
* Tests will have to be added

As an aside, C# 8.0 is also adding support for static methods and fields on interface types. The ECMA-335 standard has supported this for a while, but it's likely the scenario is untested in IL Linker (e.g. support for static constructors in interfaces, etc.).

MichalStrehovsky:
Spec for default interface methods is here: https://github.com/dotnet/runtime/blob/master/docs/design/features/default-interface-methods.md

But the short version is basically:

* Default interface methods only kick in if there's no implementation on the class
* The default implementation can either be provided by the method that declares the interface slot, or by any other interface that requires the interface in question. In that case, we look for an override record. These are not done by name and sig match.

Fixes #427.

### DIFF
--- a/src/linker/Linker.Steps/TypeMapStep.cs
+++ b/src/linker/Linker.Steps/TypeMapStep.cs
@@ -26,6 +26,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 //
 
+using System.Collections.Generic;
 using Mono.Cecil;
 
 namespace Mono.Linker.Steps
@@ -57,21 +58,41 @@ namespace Mono.Linker.Steps
 			if (!type.HasInterfaces)
 				return;
 
+			// Foreach interface and for each newslot virtual method on the interface, try
+			// to find the method implementation and record it.
 			foreach (var interfaceImpl in type.GetInflatedInterfaces ()) {
 				foreach (MethodReference interfaceMethod in interfaceImpl.InflatedInterface.GetMethods ()) {
 					MethodDefinition resolvedInterfaceMethod = interfaceMethod.Resolve ();
 					if (resolvedInterfaceMethod == null)
 						continue;
 
+					// TODO-NICE: if the interface method is implemented explicitly (with an override),
+					// we shouldn't need to run the below logic. This results in linker potentially
+					// keeping more methods than needed.
+
+					if (!resolvedInterfaceMethod.IsVirtual
+						|| resolvedInterfaceMethod.IsFinal
+						|| !resolvedInterfaceMethod.IsNewSlot)
+						continue;
+
+					// Try to find an implementation with a name/sig match on the current type
 					MethodDefinition exactMatchOnType = TryMatchMethod (type, interfaceMethod);
 					if (exactMatchOnType != null) {
 						AnnotateMethods (resolvedInterfaceMethod, exactMatchOnType);
 						continue;
 					}
 
+					// Next try to find an implementation with a name/sig match in the base hierarchy
 					var @base = GetBaseMethodInTypeHierarchy (type, interfaceMethod);
-					if (@base != null)
+					if (@base != null) {
 						AnnotateMethods (resolvedInterfaceMethod, @base, interfaceImpl.OriginalImpl);
+						continue;
+					}
+
+					// Look for a default implementation last.
+					foreach (var defaultImpl in GetDefaultInterfaceImplementations (type, resolvedInterfaceMethod)) {
+						Annotations.AddDefaultInterfaceImplementation (resolvedInterfaceMethod, type, defaultImpl);
+					}
 				}
 			}
 		}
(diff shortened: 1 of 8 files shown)


src/linker/Linker.Steps/TypeMapStep.cs
```diff
@@ -137,6 +158,53 @@ static MethodDefinition GetBaseMethodInTypeHierarchy (TypeDefinition type, Metho
...
+				if (potentialImplInterface == null)
+					continue;
+
+				bool foundImpl = false;
+
+				foreach (var potentialImplMethod in potentialImplInterface.Methods) {
+					if (potentialImplMethod == interfaceMethod &&
+						!potentialImplMethod.IsAbstract) {
+						yield return interfaceImpl;
+					}
```
vitek-karas, line 180: Should this also break out of the loop? If we've found the exact method match with a body... how could there be another method on the interface which overrides the method we're looking for.

src/linker/Linker.Steps/TypeMapStep.cs
```diff
@@ -137,6 +158,53 @@ static MethodDefinition GetBaseMethodInTypeHierarchy (TypeDefinition type, Metho
...
+
+					if (!potentialImplMethod.HasOverrides)
+						continue;
+
+					// This method is an override of something. Let's see if it's the method we are looking for.
+					foreach (var @override in potentialImplMethod.Overrides) {
+						if (@override.Resolve () == interfaceMethod) {
+							yield return interfaceImpl;
```
vitek-karas, line 188: Just curious: Is it guaranteed that the override is not an abstract method?

src/linker/Linker/Annotations.cs
```diff
@@ -58,6 +58,7 @@ public partial class AnnotationStore
 		protected readonly Dictionary<MethodDefinition, List<MethodDefinition>> base_methods = new Dictionary<MethodDefinition, List<MethodDefinition>> ();
 		protected readonly Dictionary<AssemblyDefinition, ISymbolReader> symbol_readers = new Dictionary<AssemblyDefinition, ISymbolReader> ();
 		readonly Dictionary<IMemberDefinition, LinkerAttributesInformation> linker_attributes = new Dictionary<IMemberDefinition, LinkerAttributesInformation> ();
+		protected readonly Dictionary<MethodDefinition, List<(TypeDefinition, InterfaceImplementation)>> default_interface_implementations = new Dictionary<MethodDefinition, List<(TypeDefinition, InterfaceImplementation)>> ();
```
vitek-karas, line 61: Nit: please add names to the tuple items - it would avoid the `Item1`/`Item2` code above which is harder to read.

src/linker/Linker.Steps/TypeMapStep.cs
MichalStrehovsky: In IL you actually can put an override record that overrides a method within the the same class. I don't know what it's good for, but it works and there's tests for that in the CoreCLR tree.

We don't allow this for interfaces but not as a conscious choice, but because it's a spec/implementation deficiency (the runtime will crash with `COR_E_EXECUTIONENGINE` if you do that on interfaces; I just tried).

I don't want to burn into linker more handling for this than the runtime has. If the runtime ever decides it wants to do the same thing for interfaces as it can do for classes, bailing here early would be incorrect.

src/linker/Linker.Steps/TypeMapStep.cs
MichalStrehovsky: It could be abstract. I think this is the C# syntax for it:

```csharp
interface IFoo
{
    void Frob() { Console.WriteLine("Hi"); }
}

interface IBar : IFoo
{
    abstract void IFoo.Frob();
}
```

MichalStrehovsky: If we filter them out, then we would "fix" following code:

```
interface IFoo
{
    void Frob() { Console.WriteLine("Hi"); }
}

interface IBar : IFoo
{
    abstract void IFoo.Frob();
}

class Program : IBar
{
    static void Main() => ((IFoo)new Program()).Frob();
}
```

Before linking, we would get an exception because the default implementation got reabstracted and it's no longer available in IBar. After linking, linker would strip IBar and IBar.IFoo.Frob out as "unused" and make the program work. I don't think that's desirable.

MichalStrehovsky: (Of course this is only testable from IL, or by compiling into multiple assemblies and switching them around before running. The C# compiler will not allow generation of an "incomplete" type.)

MichalStrehovsky:
@marek-safar is it intentional that the Mono/Linux builds make things look like .NET Core, but then fail to provide reference assemblies that are .NET Core?

The Linux leg is getting:

```
error CS8701: Target runtime doesn't support default interface implementation.
```

Roslyn will typically do this if the ref assemblies don't declare RuntimeFeature.DefaultImplementationsOfInterfaces.

I tried making this test ignored outside NETCOREAPP but the Mono builds still fail on this.

MichalStrehovsky:
> I resolved this with the last commit - core issue - we build tests twice, ignoring the test case only disables the second build, not the first.

Thank you!

PR #1358, opened 2020-07-14
Allow stripping Delegate.BeginInvoke/EndInvoke

MichalStrehovsky:
Per ECMA-335 I.8.9.3, the `BeginInvoke` and `EndInvoke` methods are not mandatory. On .NET Core, these methods only throw, so it's unlikely that apps would actually have a reference to them. This lets us strip the methods and the associated interfaces.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2379,7 +2379,7 @@ namespace Mono.Linker.Steps
 					methodPair = "EndInvoke";
 				else if (method.Name == "EndInvoke")
 					methodPair = "BeginInvoke";
-				
+
 				if (methodPair != null) {
 					TypeDefinition declaringType = method.DeclaringType;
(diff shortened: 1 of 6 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2373,6 +2376,10 @@ protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo
 			if (ShouldParseMethodBody (method))
 				MarkMethodBody (method.Body);
 
+			if (method.DeclaringType.IsMulticastDelegate () && (method.Name == "BeginInvoke" || method.Name == "EndInvoke")) {
+				MarkMulticastDelegate (method.DeclaringType, markAllMethods: true);
```
marek-safar, line 2380: I don't understand the rationale behind this logic. Why not mark only BeginInvoke/EndInvoke pair when one of them is marked

MichalStrehovsky: Just lest code to maintain - the net effect is the same. I can redo it to make it mark the other method instead.

marek-safar: I'd prefer clarity over less code to maintain

PR #1360, opened 2020-07-14
Show error on invalid command-line argument

### LINKED ISSUE #1354: Parameter parsing errors should use new error logging
Errors about missing or invalid arguments are printed to stdout, which shows up with low importance. It doesn't get printed at all with default MSBuild verbosity settings, and with `diagnostic` verbosity, shows up like this on my machine:

Screen Shot 2020-07-13 at 11 51 31 AM

Notice how hard it is to read the actual error, highlighted here:

Screen Shot 2020-07-13 at 11 52 05 AM

 They should ideally use the new error logging infrastructure and show up as errors in red. If not, they should at least be printed to stderr so that they get high importance.

mateoatr:
Should fix #1354

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -55,6 +55,10 @@ error and warning codes.
 
 - Element in 'XML document location' contains a 'featuredefault' attribute with an invalid value. This attribute only supports the true value, to indicate that this is the default behavior for a feature when a value is not given.
 
+#### `IL1015`: Unrecognized command-line option: 'option'
+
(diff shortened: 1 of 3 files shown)


mateoatr:
> Thanks! Could we do the same for other parameter parsing errors? Basically the places we call `WriteLine` (for example in `GetOptimizationName`, `GetBoolParam`, etc.) for an error should produce a proper error or at least log to stderr.

Sure thing! I'll put those in a follow-up PR.

PR #1361, opened 2020-07-15
Improve warnings on internal attribute use

tlakollo:
-Fix warning problems when invalid scenarios happen which end up printing IL2029
-Create warnings for invalid scenarios.
-Create test

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -246,10 +246,10 @@ error and warning codes.
 
 - All overrides of a virtual method including the base method must have the same DynamicallyAccessedMemberAttribute usage on all it's components (return value, parameters and generic parameters).
 
-#### `IL2048`: Internal attribute RemoveAttributeInstances is being used on member that is not a type
+#### `IL2048`: Internal attribute 'RemoveAttributeInstances' can only be used on a type, but is being used on '<member>'
 
-- Internal attribute RemoveAttributeInstances is a special attribute that should only be used on custom attribute types. If referenced in an assembly, module, method, property or event the internal attribute doesn't take any effect
+- Internal attribute 'RemoveAttributeInstances' is a special attribute that should only be used on custom attribute types and is being used on 'member'.
 
-#### `IL2049`: Not recognized internal attribute
+#### `IL2049`: Unrecognized internal attribute 'attribute'
 
-- The internal attribute name being used in the xml is not supported by the linker, check the spelling and the supported internal attributes
+- The internal attribute name 'attribute' being used in the xml is not supported by the linker, check the spelling and the supported internal attributes.
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -245,3 +245,11 @@ error and warning codes.
 #### `IL2047`: DynamicallyAccessedMemberTypes in DynamicallyAccessedMembersAttribute on <member> don't match overridden <base member>. All overriden members must have the same DynamicallyAccessedMembersAttribute usage.
 
 - All overrides of a virtual method including the base method must have the same DynamicallyAccessedMemberAttribute usage on all it's components (return value, parameters and generic parameters).
+
+#### `IL2048`: Internal attribute RemoveAttributeInstances is being used on member that is not a type
+
+- Internal attribute RemoveAttributeInstances is a special attribute that should only be used on custom attribute types. If referenced in an assembly, module, method, property or event the internal attribute doesn't take any effect
+
+#### `IL2049`: Not recognized internal attribute
+
+- The internal attribute name being used in the xml is not supported by the linker, check the spelling and the supported internal attributes
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -246,9 +246,9 @@ error and warning codes.
 
 - All overrides of a virtual method including the base method must have the same DynamicallyAccessedMemberAttribute usage on all it's components (return value, parameters and generic parameters).
 
-#### `IL2048`: Internal attribute 'RemoveAttributeInstances' can only be used on a type, but is being used on '<member>'
+#### `IL2048`: Internal attribute 'RemoveAttributeInstances' can only be used on a type, but is being used on 'member type' 'member'
 
-- Internal attribute 'RemoveAttributeInstances' is a special attribute that should only be used on custom attribute types and is being used on 'member'.
+- Internal attribute 'RemoveAttributeInstances' is a special attribute that should only be used on custom attribute types and is being used on'member type' 'member'.
 
 #### `IL2049`: Unrecognized internal attribute 'attribute'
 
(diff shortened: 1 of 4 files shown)


src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -36,9 +36,22 @@ IEnumerable<CustomAttribute> ProcessAttributes (XPathNavigator nav, ICustomAttri
 				TypeDefinition attributeType;
 
 				string internalAttribute = GetAttribute (iterator.Current, "internal");
-				if (internalAttribute == "RemoveAttributeInstances" && provider != null && !Annotations.IsMarked (provider)) {
-					IEnumerable<Attribute> removeAttributeInstance = new List<Attribute> { new RemoveAttributeInstancesAttribute () };
-					Context.CustomAttributes.AddInternalAttributes (provider, removeAttributeInstance);
+				if (internalAttribute != String.Empty) {
+					switch (internalAttribute) {
+					case "RemoveAttributeInstances":
```
mateoatr, line 41: Why use a switch-case when you only have one case, isn't it better to use a good old if-else here :)?

docs/error-codes.md
```diff
@@ -245,3 +245,11 @@ error and warning codes.
 #### `IL2047`: DynamicallyAccessedMemberTypes in DynamicallyAccessedMembersAttribute on <member> don't match overridden <base member>. All overriden members must have the same DynamicallyAccessedMembersAttribute usage.
 
 - All overrides of a virtual method including the base method must have the same DynamicallyAccessedMemberAttribute usage on all it's components (return value, parameters and generic parameters).
+
+#### `IL2048`: Internal attribute RemoveAttributeInstances is being used on member that is not a type
```
mateoatr, line 249:
```suggestion
#### `IL2048`: 'RemoveAttributeInstances' attribute can only be used on a type
```

vitek-karas, line 249:
```suggestion
#### `IL2048`: Internal attribute RemoveAttributeInstances is being used on '<member>' that is not a type
```

Write out what it's used on - much easier to find and resolve - and we have the information anyway.

docs/error-codes.md
```diff
@@ -245,3 +245,11 @@ error and warning codes.
...
+
+- Internal attribute RemoveAttributeInstances is a special attribute that should only be used on custom attribute types. If referenced in an assembly, module, method, property or event the internal attribute doesn't take any effect
+
+#### `IL2049`: Not recognized internal attribute
```
vitek-karas, line 253:
```suggestion
#### `IL2049`: 'RandomName' is not a recognized internal attribute
```

Similar to above, give information which makes it easier to find and fix the problem

src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -36,9 +36,22 @@ IEnumerable<CustomAttribute> ProcessAttributes (XPathNavigator nav, ICustomAttri
...
+						if (provider != null) {
```
mateoatr, line 42: I think I don't understand why `provider != null` implies that this is a type.

src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -36,9 +36,22 @@ IEnumerable<CustomAttribute> ProcessAttributes (XPathNavigator nav, ICustomAttri
...
+							if (!Annotations.IsMarked (provider)) {
+								IEnumerable<Attribute> removeAttributeInstance = new List<Attribute> { new RemoveAttributeInstancesAttribute () };
+								Context.CustomAttributes.AddInternalAttributes (provider, removeAttributeInstance);
+							}
+						} else {
+							Context.LogWarning ($"Internal attribute RemoveAttributeInstances is being used on member that is not a type", 2048, _xmlDocumentLocation);
+						}
+						break;
+					default:
+						Context.LogWarning ($"Not recognized internal attribute", 2049, _xmlDocumentLocation);
```
mateoatr, line 52:
```suggestion
						Context.LogWarning ($"Unrecognized internal attribute '{internalAttribute}'", 2049, _xmlDocumentLocation);
```

src/linker/Linker.Steps/LinkAttributesStep.cs
tlakollo: ProcessAttributes function is only called with a provider when it's called inside ProcessType method, all the other methods don't pass a provider and ProcessAttributes function has a provider by default in null. But I guess I will need to change that if I want to print which member type was being used.

src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -36,9 +36,17 @@ IEnumerable<CustomAttribute> ProcessAttributes (XPathNavigator nav, ICustomAttri
 				TypeDefinition attributeType;
 
 				string internalAttribute = GetAttribute (iterator.Current, "internal");
-				if (internalAttribute == "RemoveAttributeInstances" && provider != null && !Annotations.IsMarked (provider)) {
-					IEnumerable<Attribute> removeAttributeInstance = new List<Attribute> { new RemoveAttributeInstancesAttribute () };
-					Context.CustomAttributes.AddInternalAttributes (provider, removeAttributeInstance);
+				if (internalAttribute != String.Empty) {
```
marek-safar, line 39: Please rewrite this nested ifs into `if (condition) { .. continue; }` style

src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -36,9 +36,17 @@ IEnumerable<CustomAttribute> ProcessAttributes (XPathNavigator nav, ICustomAttri
...
+					if (internalAttribute == "RemoveAttributeInstances") {
+						if (provider != null) {
+							if (!Annotations.IsMarked (provider)) {
+								IEnumerable<Attribute> removeAttributeInstance = new List<Attribute> { new RemoveAttributeInstancesAttribute () };
+								Context.CustomAttributes.AddInternalAttributes (provider, removeAttributeInstance);
+							}
+						} else
+							Context.LogWarning ($"Internal attribute 'RemoveAttributeInstances' can only be used on a type, but is being used on '<{nav.Name}>'", 2048, _xmlDocumentLocation);
```
marek-safar, line 47: Why is this not error?

vitek-karas: I think we should keep this a warning - mainly because there's no way to suppress an error (and if this comes from a 3rd party component, there's no way to avoid the error if it can't be suppressed).

vitek-karas, line 47: I think we should pass the provider always and print out the provider in the warning message. That will point exactly at the spot where the problem is. Currently it will only say "method", but not which one. (Unlike roslyn, we can't rely on providing context via source/line info - although we could add that and then we could).

src/linker/Linker.Steps/LinkAttributesStep.cs
tlakollo: I changed the implementation but I'm not sure if this is what you were referring

src/linker/Linker.Steps/LinkAttributesStep.cs
tlakollo: I changed to use the providers for everything, in the specific case of method params I still print the method as the provider since I cannot pass an enumeration of method params to the function.

PR #1362, opened 2020-07-15
Make the --deterministic the default

### LINKED ISSUE #1342: Change deterministic output  to be the default linker mode

vitek-karas:
Make linker deterministic by default.

These are changes from #587 I just didn't want to force push to that PR (I had to rebase everything).

I tested that running linker on a single assembly it produces the exact same output many times over. I also tested that the produced assembly can be run through linker again (the produced assembly is not corrupted in any way).

I also tested with latest SDK that publishing a Blazor template app with this linker produces exactly the same output twice over (only validated the _framework directory content).

Fixes #1342

### DIFF
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -369,7 +369,7 @@ namespace Mono.Linker
 
 							continue;
 						}
-								
+
 					case "--feature": {
 							string featureName = null;
 							if (!GetStringParam (token, l => featureName = l))
@@ -579,8 +579,7 @@ namespace Mono.Linker
 			}
 
 			// Default to deterministic output
-			if (!new_mvid_used && !deterministic_used)
-			{
+			if (!new_mvid_used && !deterministic_used) {
 				context.DeterministicOutput = true;
 			}
 			if (dumpDependencies)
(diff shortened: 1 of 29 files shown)


vitek-karas:
The CI failure is caused by https://github.com/jbevain/cecil/issues/680.

vitek-karas:
I have PR to mono/cecil with the fix here: https://github.com/mono/cecil/pull/27

vitek-karas:
The cecil fix is in - CI passes now.

src/linker/Linker/Driver.cs
```diff
@@ -572,6 +573,10 @@ protected int SetupContext (ILogger customLogger = null)
 				return -1;
 			}
 
+			// Default to deterministic output
+			if (!new_mvid_used && !deterministic_used) {
```
marek-safar, line 577: Does it mean `--deterministic true` will not enable deterministic output

vitek-karas: No - `--deterministic true` will do exactly the same (it's effectively the default now). See https://github.com/mono/linker/blob/4936cd82cb5f1d438d83c7e06c69ba1ba32eab2e/src/linker/Linker/Driver.cs#L395

What this means is mostly that specifying for example `--new-mvid true` will still take effect even if deterministic is the default (`--new-mvid` effectively overrides the default).

marek-safar: Yep, missed that

PR #1364, opened 2020-07-16
Warn on invalid arguments

mateoatr:
Follow-up on #1360.


docs/error-codes.md
```diff
@@ -59,6 +59,22 @@ error and warning codes.
 
 - The linker was passed a string that was not a linker option.
 
+### `IL1016`: Missing argument
+
+### `IL1017`: Invalid argument
+
```
marek-safar, line 65: I don't think we should use the same error code for quite different problems (I understand the are in the same bucket)

vitek-karas: I agree - we should try to get to a place where a given code number has exactly one message (1:1 mapping between number and message string).

Just curious: @agocke how does Roslyn handle command line parsing errors like these?

sbomer: Just tried this:
```
> dotnet bin/recent_sdk/sdk/5.0.100-preview.8.20351.5/Roslyn/bincore/csc.dll -invalid -out
Microsoft (R) Visual C# Compiler version 3.7.0-3.20312.3 (ec484126)
Copyright (C) Microsoft Corporation. All rights reserved.

error CS2007: Unrecognized option: '-invalid'
error CS2005: Missing file specification for '-out' option
warning CS2008: No source files specified.
error CS1562: Outputs without source must have the /out option specified
```

agocke: Yeah, invalid argument and "argument missing a parameter" are different enough that they're considered two separate arguments.

The way I would decide that they are different is: I can't think of a way to write an error message that encompasses both that isn't painfully vague, e.g. "invalid argument: `blah`" is so vague that I wouldn't have any idea what's wrong with the argument.

agocke: Also, sharing codes for warnings is more important than errors, because warnings can be configured.

vitek-karas: > Also, sharing codes for warnings is more important than errors, because warnings can be configured.

I assume this means "Try hard not to share codes between different warnings"

agocke: > I assume this means "Try hard not to share codes between different warnings"

Yes, for warnings try to share only if you think the right behavior is to configure both things together. For instance, there's a particular type of nullable warning where you convert a nullable type into a non-nullable type through a cast in a method body. There are number of ways you can do this, but the result is the same, in that the type changed, but the flow state did not, all because it happened in a method body. We referred to these as "non-safety" warnings, because the flow analysis was still tracking whether the variable was null or not, the only purpose was to warn users about mixing up nullable and non-nullable types.

Even though there were a variety of ways you could encounter this issue, we grouped them all under a single warning code, specifically so users could turn off all those warnings at once. The theory was that they were all fundamentally the same problem, and if you didn't care about one instance, you didn't care about any of them.

Conversely, we made sure to never categorize a flow analysis warning and "non-safety" warning together because we wanted to emphasize that these were not the same problem at all. The non-safety warning was more of a hygiene concern, while the flow analysis warning was pointing at something which would likely null-ref at runtime.

PR #1369, opened 2020-07-20
Use class hierarchy in IsTypeInterestingForDataflow

### LINKED ISSUE #1365: IsTypeInterestingForDataflow logic is not sufficient
```csharp
		static bool IsTypeInterestingForDataflow (TypeReference typeReference)
		{
			// We will accept any System.Type* as interesting to enable testing
			// It's necessary to be able to implement a custom "Type" type in tests to validate
			// the correct propagation of the annotations on "this" parameter. And tests can't really
			// override System.Type - as it creates too many issues.
			// It also covers TypeBuilder and RuntimeType.
			return typeReference.MetadataType == MetadataType.String ||
				(typeReference.Name.Contains ("Type") && typeReference.Namespace.StartsWith ("System"));
		}
```

We will need to extend this to cover "the type derives from System.Type".

I'm hitting a problem that instance methods on EnumBuilder cannot be annotated because linker is telling me the special `this` parameter annotation cannot be used on EnumBuilders' methods, so linker ignores them, but then I get warnings that the EnumBuilder annotations don't match annotations on virtual methods inherited from System.Type.

I'm worried about perf impact of extending this to checking the base type hierarchy. We probably need a cache.

Cc @vitek-karas

MichalStrehovsky:
Fixes #1365.

Extracted the base type caching into a separate class. We might want to make this more global and use this where we're doing similar checks within linker (we have existing code that is checking whether a type is EventSource or deriving from some security attribute). It feels like a general purpose question, but those things are out of scope of this pull request.

### DIFF
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -377,7 +377,7 @@ namespace Mono.Linker.Dataflow
 		bool IsTypeInterestingForDataflow (TypeReference typeReference)
 		{
 			return typeReference.MetadataType == MetadataType.String ||
-				_hierarchyInfo.IsSystemType(typeReference) ||
+				_hierarchyInfo.IsSystemType (typeReference) ||
 				(typeReference.Name == "IReflect" && typeReference.Namespace == "System.Reflection");
 		}
 
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -378,7 +378,7 @@ namespace Mono.Linker.Dataflow
 		{
 			return typeReference.MetadataType == MetadataType.String ||
 				_hierarchyInfo.IsSystemType (typeReference) ||
-				(typeReference.Name == "IReflect" && typeReference.Namespace == "System.Reflection");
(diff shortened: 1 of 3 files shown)


src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -373,15 +374,11 @@ bool ScanMethodBodyForFieldAccess (MethodBody body, bool write, out FieldDefinit
...
 		{
-			// We will accept any System.Type* as interesting to enable testing
-			// It's necessary to be able to implement a custom "Type" type in tests to validate
-			// the correct propagation of the annotations on "this" parameter. And tests can't really
-			// override System.Type - as it creates too many issues.
-			// It also covers TypeBuilder and RuntimeType.
 			return typeReference.MetadataType == MetadataType.String ||
-				(typeReference.Name.Contains ("Type") && typeReference.Namespace.StartsWith ("System"));
+				_hierarchyInfo.IsSystemType(typeReference) ||
+				(typeReference.Name == "IReflect" && typeReference.Namespace == "System.Reflection");
```
vitek-karas, line 381: Won't we need to also allow annotations on types which implement `IReflect`? Otherwise it won't be possible to implement the interface without warnings from the linker.

MichalStrehovsky:
> so I'll leave that up to you.

Sigh. I added it because I know we'll have to come back to this for some obscure reason otherwise.

PR #1372, opened 2020-07-21
Warn by default

mateoatr:
Log warnings and errors regardless of the `--verbose` option.

Dependent upon https://github.com/dotnet/runtime/pull/39133.


sbomer:
@layomia added bulk suppressions for warnings in runtime (https://github.com/dotnet/runtime/pull/40106), so I think this is unblocked.

sbomer:
For testing, maybe it would be enough to remove `--verbose` from the default test options: https://github.com/mono/linker/blob/7c9b806037e88df7eb40a8151808730133676668/test/Mono.Linker.Tests/TestCasesRunner/TestCaseMetadaProvider.cs#L90 and add it explicitly for any tests that check informational messages?

mateoatr:
> For testing, maybe it would be enough to remove `--verbose` from the default test options:
> 
> https://github.com/mono/linker/blob/7c9b806037e88df7eb40a8151808730133676668/test/Mono.Linker.Tests/TestCasesRunner/TestCaseMetadaProvider.cs#L90
> 
> and add it explicitly for any tests that check informational messages?

Yes, I think this should be enough, although I think this could impact #1385. @vitek-karas should I wait until the warnings cleanup is merged?

src/linker/Linker.Steps/BlacklistStep.cs
```diff
@@ -91,10 +91,10 @@ protected override void Process ()
 									.Where (res => res.Name.Equals ("ILLink.LinkAttributes.xml", StringComparison.OrdinalIgnoreCase))
 									.Cast<EmbeddedResource> ()) {
 					try {
-						Context.LogMessage (MessageContainer.CreateInfoMessage ($"Processing embedded {rsc.Name} from {asm.Name}"));
+						Context.LogMessage ($"Processing embedded {rsc.Name} from {asm.Name}");
 						steps_to_add.Push (GetExternalLinkAttributesStep (rsc, asm));
 					} catch (XmlException ex) {
-						Context.LogMessage (MessageContainer.CreateErrorMessage ($"Error processing {rsc.Name} from {asm.Name}: {ex}", 1003));
+						Context.LogError ($"Error processing {rsc.Name} from {asm.Name}: {ex}", 1003);
```
agocke, line 97: Since we've segmented the diagnostics into default severity by error code, shouldn't we know what the default severity should be just by looking at the code? Why do we have to say what it is at the callsite (i.e. LogError vs LogMessage)

mateoatr: Good point. We could replace every `LogWarning`/`LogError` by a single `Log` and then just check the code... at the same time, I also think that it doesn't hurt to be explicit about what's the thing that we're logging (error or warning.)

mateoatr: Opened #1406 to track this.

Issue #1371, opened 2020-07-20
IL2045 for blazor app


sbomer:
This warning will show up by default in every blazor app unless we silence it as part of https://github.com/dotnet/sdk/pull/12388. @tlakollo fixed a similar issue in https://github.com/mono/linker/pull/1332, for conflicting root descriptors and attribute removal directives. Here the conflict is between an assembly action and a removal directive.

The blazor defaults have some `copy` assemblies, which means that attribute instances in that assembly are kept, causing the attribute type to be kept (since https://github.com/mono/linker/pull/1344 I think).

We also have linker attribute [files](https://github.com/mono/linker/pull/1332) which say to remove attributes like `NullableAttribute`. When an attribute type is kept despite such a directive, we emit `IL2045`.

@vitek-karas @tlakollo

vitek-karas:
I tried this and there are two cases where we generate 2045:
* The attributes themselves have self-references and those generate 2045 - this will be fixed by #1373 
* Code in stack trace processing relies on some of the removed attributes - this has been discussed and we want to leave it like that - so we will have to do something about this in the runtime (either suppress the warning or wrap the usage with the debugger feature switch)

PR #1373, opened 2020-07-21
Don't generate warning 2045 due to the attribute type members referencing the type itself

vitek-karas:
If an attribute is marked for removal of instances, but the linker keeps the attribute type for other reasons (for example because it's used from a copy assembly) then there will be several other references to the attribute type from the code within the attribute type itself. These warnings are useless and should not be generated.

Other changes:
* Moved link attribute tests into their own separate folder - they don't belong to DataFlow tests
* The log verification logic now removes messages which where matched by LogContains attributes, so it's possible to validate that there are not other messages of the same kind by adding LogDoesntContain at the end

Fixes most warnings from #1371


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1380,11 +1380,26 @@ internal protected virtual TypeDefinition MarkType (TypeReference reference, Dep
...
-				_context.LogWarning ($"Custom Attribute {type.GetDisplayName ()} is being referenced in code but the linker was " +
-					$"instructed to remove all instances of this attribute. If the attribute instances are necessary make sure to " +
-					$"either remove the linker attribute XML portion which removes the attribute instances, or to override this use " +
-					$"the linker XML descriptor to keep the attribute type (which in turn keeps all of its instances).", 2045, sourceLocationMember);
+			if (_context.Annotations.HasLinkerAttribute<RemoveAttributeInstancesAttribute> (type)) {
+				bool shouldWarn = true;
+
+				// Don't warn about references from code which is not linked
+				//if (Annotations.GetAction (sourceLocationMember.DeclaringType.Module.Assembly) != AssemblyAction.Link)
+				//	shouldWarn = false;
```
vitek-karas, line 1388: I'm not sure we should suppress the warning in this case. The scenario is basically:
* An attribute is marked for removal (let's say attribute `ToBeRemovedAttribute`)
* The attribute instantiated in some places in linked assemblies - and those instantiations will be removed
* There's an assembly which is not linked ("copy") which has a type reference to the attribute type - so for example `typeof(ToBeRemovedAttribute)`.

The question is, should this type reference generate a 2045 warning even if it's from an assembly which is not linked?

My take is: It should - the code is likely to break if it uses the attribute's type to look up attribute instances as those won't be there. The fact that it's in a copy assembly is not different than if it would be a linked assembly, but in code which is reachable.

I intentionally left the code commented out and a TODO - once we agree on the answer I'll update the PR.

sbomer: Will this case also be hit for a "copy" or "copyused" assembly that has an instance of `ToBeRemovedAttribute` (rather than referencing it via `typeof`)? I am wondering how often this is likely to show up in user scenarios.

vitek-karas: Having an instance doesn't trigger this warning - that's exactly what happened with the warnings in CoreLib you reported - copy assemblies had instances of some of the attributes  - the instance itself doesn't cause warnings even without this change. It does cause warnings within the attribute type - which is what this change fixes.

tlakollo:
It's missing moving the XmlAnnotations.cs and XmlAnnotations.xml file to LinkAttributes folder, and maybe while we move it we should change the name too. This was one of the first files to be created and now the name is kind of odd.

vitek-karas:
I actually left the XmlAnnotations behind intentionally - that test uses the link attribute basically solely to inject `DynamicallyAccessedMembers` attributes and the validation is solely based on the DataFlow logic. So it kind of belongs to both DataFlow and LinkAttributes. To me it feels a little bit better to leave it in DataFlow.

But if somebody feels strongly that it belongs to LinkAttributes I'll be happy to move it.

src/linker/Linker.Steps/MarkStep.cs
sbomer: In that case I agree we should emit the warning even for non-linked assemblies.

Issue #1356, opened 2020-07-14
Support warnaserror


sbomer:
We should have some way to treat all warnings or specific warnings as errors. What exists today:
- SDK properties without special behavior: (`TreatWarningsAsErrors` bool, `WarningsAsErrors` list)
  These are explicitly passed as task arguments to tasks that support them.
  They are the properties that are set by the VS UI.
- MSBuild properties that automatically promote warnings to errors (`MSBuildTreatWarningsAsErrors` bool, `MSBuildWarningsAsErrors` list)
  https://github.com/microsoft/msbuild/issues/4421 is about inferring the MSBuild properties from the above.
  The MSBuild flag `/warnaserror` behaves similarly to setting `MSBuildTreatWarningsAsErrors`.

It would be nice to rely on the MSBuild properties instead of re-implementing support for `--warnaserror` - however, we will probably need a workaround for https://github.com/microsoft/msbuild/issues/5511 if we do that. If there's no easy workaround, we probably need to support `--warnaserror`.

If we also want to support `WarningsNotAsErrors`, we will need to either support this via `--warnaserror-`, or wait for `MSBuildWarningsNotAsErrors` (https://github.com/microsoft/msbuild/issues/3062).

marek-safar:
Isn't the work mostly on SDK side for this task?

vitek-karas:
Based on some offline discussions we're leaning towards not relying on MSBuild for this functionality. That means we need the linker to promote warnings to errors (the task or the SDK can't really do that AFAIK). Which leads to adding the `--warnaserror` command line option.

@sbomer - could you please add a short "spec" of how the `--warnaserror` should work? (I assume we will copy the behavior from Roslyn for the most part).

sbomer:
Here's how it would work if we do what csc does (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/warnaserror-compiler-option), with minor syntax changes:

```
syntax:                          --warnaserror[+|-] [warning-list]
all warnings as errors:          --warnaserror
                                 --warnaserror+
all warnings not as errors::     --warnaserror-
specific warnings as errors:     --warnaserror IL2006,IL2007
                                 --warnaserror+ IL2006,IL2007
specific warnings not as errors  --warnaserror- IL2006,IL2007
```
Options are processed in order, last wins - so you can for example treat all warnings except IL2006 as errors with:
```
--warnaserror --warnaserror- IL2006
```
[...]

PR #1377, opened 2020-07-22
Add warnaserror option

mateoatr:
This PR implements the `warnaserror` option as discussed in #1356.

/cc @agocke

### DIFF
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -442,11 +442,11 @@ namespace Mono.Linker
 						var warningList = GetNextStringValue ();
 						if (!string.IsNullOrEmpty (warningList)) {
 							foreach (var warning in ParseWarnings (warningList))
-								context.WarnAsError.Add ((warning, true));
+								context.WarnAsError[warning] = true;
 
 						} else {
 							context.GeneralWarnAsError = true;
-							context.WarnAsError = new HashSet<(uint, bool)> ();
+							context.WarnAsError.Clear ();
 						}
 
 						continue;
@@ -455,11 +455,11 @@ namespace Mono.Linker
 						warningList = GetNextStringValue ();
 						if (!string.IsNullOrEmpty (warningList)) {
 							foreach (var warning in ParseWarnings (warningList))
-								context.WarnAsError.Add ((warning, false));
+								context.WarnAsError[warning] = false;
 
 						} else {
 							context.GeneralWarnAsError = false;
-							context.WarnAsError = new HashSet<(uint, bool)> ();
+							context.WarnAsError.Clear ();
 						}
 
 						continue;
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -437,6 +437,33 @@ namespace Mono.Linker
 						context.NoWarn.UnionWith (ParseWarnings (noWarnArgument));
 						continue;
 
+					case "--warnaserror":
+					case "--warnaserror+":
+						var warningList = GetNextStringValue ();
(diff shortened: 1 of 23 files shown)


src/linker/Linker/Driver.cs
```diff
@@ -437,6 +437,33 @@ protected int SetupContext (ILogger customLogger = null)
...
+					case "--warnaserror":
+					case "--warnaserror+":
+						var warningList = GetNextStringValue ();
+						if (!string.IsNullOrEmpty (warningList)) {
+							foreach (var warning in ParseWarnings (warningList))
+								context.WarnAsError.Add ((warning, true));
+
+						} else {
+							context.GeneralWarnAsError = true;
+							context.WarnAsError = new HashSet<(uint, bool)> ();
```
vitek-karas, line 449:
```suggestion
							context.WarnAsError.Clear ();
```

vitek-karas: Same below

src/linker/Linker/LinkContext.cs
```diff
@@ -531,6 +537,12 @@ public void LogWarning (string text, int code, MessageOrigin origin, string subc
 			if (!LogMessages)
 				return;
 
+			if ((GeneralWarnAsError && !WarnAsError.Contains (((uint) code, false))) ||
+				(!GeneralWarnAsError && WarnAsError.Contains (((uint) code, true)))) {
+				LogError (text, code, subcategory, origin, true);
```
vitek-karas, line 542:
```suggestion
				LogError (text, code, subcategory, origin, isWarnAsError: true);
```

For documentation purposes

test/Mono.Linker.Tests.Cases/Warnings/CanDisableWarnAsError.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Mono.Linker.Tests.Cases.Warnings
+{
+	[SkipKeptItemsValidation]
+	[SetupLinkerSubstitutionFile ("CanDisableWarnAsErrorSubstitutions.xml")]
+	[SetupLinkerArgument ("--verbose")]
+	[SetupLinkerArgument ("--warnaserror")]
+	[SetupLinkerArgument ("--warnaserror-", "IL2011,IL2012,IgnoreThis")]
+	[LogContains ("error IL20(07|08|09|10)", true)]
+	[LogContains ("warning IL20(11|12)", true)]
```
vitek-karas, line 13: Don't we need to list these one by one? Currently this will match all of the errors/warnings, but there's no validation that it got them all. For example if for some reason the `IL2012` won't be treated as warning, but `IL2011` still is, the test would still pass.

src/linker/Linker/LinkContext.cs
```diff
@@ -531,6 +537,12 @@ public void LogWarning (string text, int code, MessageOrigin origin, string subc
 			if (!LogMessages)
 				return;
 
+			if ((GeneralWarnAsError && !WarnAsError.Contains (((uint) code, false))) ||
+				(!GeneralWarnAsError && WarnAsError.Contains (((uint) code, true)))) {
```
vitek-karas, line 541: My understanding of the desired behavior is that command line like this:
```console
--warnaserror IL2020 --warnaserror- IL2020
```

Will effectively mean a no-op, that is `IL2020` will be reported as a warning. Currently the code will report this as an error. Basically this part of the "Spec":
> Options are processed in order, last wins

We should also have test for these cases.

I would implement this by changing the `HashSet` to a `Dictionary<uint, bool>`, but there are probably other ways to do this.

Issue #1359, opened 2020-07-14
Support --warn <version_int>


sbomer:
We would like to turn on all of the new warnings by default from the command-line (without requiring `--verbose`). The current [plan](https://github.com/mono/linker/pull/1303#issuecomment-653433498) is to then disable them from MSBuild for .NET5, by setting `NoWarn` to contain the trim analysis warnings. For backward compatibility, ideally none of the new warnings would be emitted when targeting `netcoreapp3.0`.

We can do this via `NoWarn`, or we can use versioned warnings, passing a `--warnversion` flag to the linker, like Roslyn is doing. Back-compat may not be as essential in this release (since in 3.0 the trimmer was experimental, and tended to silently break apps), but we will have similar problems as we introduce new warnings in future releases.

https://github.com/dotnet/roslyn/issues/45701 is a pretty thorough summary of the Roslyn plan, which we would likely follow. The biggest open question is:

> do we bump the number for .NET releases or language releases? Today they will be the same but what if that changes?

This is more relevant for the linker because we are introducing disabled "future" warnings even in .NET5 (since the trim analysis warnings will be available, but off). Should we have three "versions" (no warnings, all but trim analysis, and all warnings)? Or do we want to keep the "experimental" trim analysis warnings under a flag of its own? It feels a bit strange to give them a version number if we are not confident that they're ready to be on by default in a specific release.

agocke:
I said the same in the internal Roslyn design doc, but `decimal` is a mistake. Version numbers may have dots in them, but that doesn't make them numbers. Consider `5.0.1`. I think there are two correct representations: on the command line, the version is a string that gets parsed (obviously). The only data type there is string. The result, I think, should be an enum. This matches Roslyn's LangVersion, which accepts versions like 8.0, but also strings like `Latest` and `Preview`. I think it's the most flexible option that also helps to reduce confusion after parsing, since you get type safety in the value.

vitek-karas:
About the data type: How will this work in the future? When .NET 6 comes around we will add the value `6` to the set. But what happens if I pass `6` to .NET 5 linker? Should it fail? Should it effectively mean `5`?... 

Even though I agree that using double as the data type for this is wrong - I think we should follow whatever Roslyn does in .NET 5 in the end - consistency beats everything else here.

As for the proposal above:
[...]

PR #1381, opened 2020-07-22
Add support for --warn <version>

sbomer:
Adds support for warning versions, following the design described in https://github.com/dotnet/roslyn/issues/46148. Fixes https://github.com/mono/linker/issues/1359.

- This passes in a warning version at each callsite, like what we are doing with the warning strings. I plan to update the callsites after we agree on the API.
- The version is optional. Non-versioned warnings always show up (but I think we will make all of the warnings versioned anyway). Custom steps can optionally use versions for their warnings - they will be controlled by the same global version setting as the built-in warnings.
- Internally it uses an enum to represent the versions as @agocke suggested
- Our command-line default is `9999` - the "neutral" default that will produce all warnings we support. There is separate SDK work to enable this option with different defaults per TFM (`0` for netcoreapp3.0, `5` for net5.0).

filed https://github.com/dotnet/sdk/issues/12601 for the SDK work.

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -5,6 +5,12 @@ identification. The known codes are in the range 1000 to 6000. Custom
 steps should avoid using this range not to collide with existing or future
 error and warning codes.
 
+For versioned warnings, the warning version is indicated in parentheses following
+the error code. For example:
+
+#### `ILXXXX` (version): Message
+- Description of the error code including more details.
+
 ## Error Codes
 
 #### `IL1001`: Failed to process 'XML document location'. Feature 'feature' does not specify a "featurevalue" attribute
@@ -86,7 +92,7 @@ error and warning codes.
 
 - The member 'member' in `PreserveDependency` attribute could not be resolved.
 
-#### `IL2006`: Unrecognized reflection pattern
+#### `IL2006` (5): Unrecognized reflection pattern
 
 - The linker found an unrecognized reflection access pattern. The most likely reason for this is that the linker could not resolve a member that is being accessed dynamicallly. To fix this, use the `DynamicallyAccessedMemberAttribute` and specify the member kinds you're trying to access.
 
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -59,6 +59,10 @@ error and warning codes.
 
 - The linker was passed a string that was not a linker option.
 
+#### `IL1016`: Invalid warning version 'version'
+
+- The value given for the --warn argument was not a valid warning version. Valid versions include integers in the range 0-9999, though not all of these map to distinct warning waves.
+
 ----
 ## Warning Codes
 
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -464,6 +464,18 @@ namespace Mono.Linker
 
 						continue;
 
+					case "--warn":
+						string warnVersionArgument = null;
+						if (!GetStringParam (token, l => warnVersionArgument = l))
+							return -1;
+
+						if (!GetWarnVersion (warnVersionArgument, out WarnVersion version))
+							return -1;
+
+						context.WarnVersion = version;
+
+						continue;
+
 					case "--version":
 						Version ();
 						return 1;
@@ -937,6 +949,19 @@ namespace Mono.Linker
 			return assemblyAction;
 		}
 
+		bool GetWarnVersion (string text, out WarnVersion version)
+		{
+			if (int.TryParse(text, out int versionNum)) {
+				version = (WarnVersion)versionNum;
+				if (version >= WarnVersion.ILLink0 && version <= WarnVersion.Latest)
+					return true;
+			}
+
+			context.LogError ($"Invalid warning version '{text}'", 1016);
+			version = 0;
+			return false;
+		}
+
 		static bool GetOptimizationName (string text, out CodeOptimizations optimization)
 		{
 			switch (text.ToLowerInvariant ()) {
@@ -1059,6 +1084,8 @@ namespace Mono.Linker
 			Console.WriteLine ("  -out PATH                     Specify the output directory. Defaults to 'output'");
 			Console.WriteLine ("  --about                       About the {0}", _linker);
 			Console.WriteLine ("  --verbose                     Log messages indicating progress and warnings");
+			Console.WriteLine ("  --warn VERSION                Only print out warnings with version <= VERSION. Defaults to '9999'");
+			Console.WriteLine ("                                  VERSION is an integer in the range 0-9999.");
 			Console.WriteLine ("  --warnaserror[+|-]            Report all warnings as errors");
 			Console.WriteLine ("  --warnaserror[+|-] WARN-LIST  Report specific warnings as errors");
 			Console.WriteLine ("  --nowarn WARN-LIST            Disable specific warning messages");
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -951,8 +951,8 @@ namespace Mono.Linker
 
 		bool GetWarnVersion (string text, out WarnVersion version)
 		{
-			if (int.TryParse(text, out int versionNum)) {
-				version = (WarnVersion)versionNum;
+			if (int.TryParse (text, out int versionNum)) {
+				version = (WarnVersion) versionNum;
 				if (version >= WarnVersion.ILLink0 && version <= WarnVersion.Latest)
 					return true;
 			}
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -507,7 +507,8 @@ namespace Mono.Linker
 
 			// Note: message.Version is nullable. The comparison is false if it is null.
 			// Unversioned warnings are not controlled by WarnVersion.
-			if (message.Category == MessageCategory.Warning &&
+			// Error messages are guaranteed to only have a version if they were created for a warning due to warnaserror.
+			if ((message.Category == MessageCategory.Warning || message.Category == MessageCategory.Error) &&
 				message.Version > WarnVersion) {
 				// This warning was turned off by --warn.
 				return;
@@ -608,7 +609,7 @@ namespace Mono.Linker
 			if (!LogMessages)
 				return;
 
-			var error = MessageContainer.CreateErrorMessage (text, code, subcategory, origin, isWarnAsError);
+			var error = MessageContainer.CreateErrorMessage (text, code, subcategory, origin, isWarnAsError, version);
 			LogMessage (error);
 		}
 
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -543,6 +543,9 @@ namespace Mono.Linker
 		/// <param name="code">Unique warning ID. Please see https://github.com/mono/linker/blob/master/doc/error-codes.md for the list of warnings and possibly add a new one</param>
 		/// <param name="origin">Filename or member where the warning is coming from</param>
 		/// <param name="subcategory">Optionally, further categorize this warning</param>
+		/// <param name="version">Optional warning version number. Versioned warnings can be controlled with the
+		/// warning wave option --warn VERSION. Unversioned warnings are unaffected by this option. </param>
+		/// <returns>New MessageContainer of 'Warning' category</returns>
 		public void LogWarning (string text, int code, MessageOrigin origin, string subcategory = MessageSubCategory.None, WarnVersion? version = null)
 		{
 			if (!LogMessages)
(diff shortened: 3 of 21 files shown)


src/linker/Linker/LinkContext.cs
```diff
@@ -542,10 +551,11 @@ public void LogWarning (string text, int code, MessageOrigin origin, string subc
 		/// <param name="code">Unique warning ID. Please see https://github.com/mono/linker/blob/master/doc/error-codes.md for the list of warnings and possibly add a new one</param>
 		/// <param name="origin">Type or member where the warning is coming from</param>
 		/// <param name="subcategory">Optionally, further categorize this warning</param>
-		public void LogWarning (string text, int code, IMemberDefinition origin, int? ilOffset = null, string subcategory = MessageSubCategory.None)
+		// TODO
+		public void LogWarning (string text, int code, IMemberDefinition origin, int? ilOffset = null, string subcategory = MessageSubCategory.None, WarnVersion? version = null)
```
vitek-karas, line 555: I don't have a problem with the API as is - I would just like better "defaults".

For warnings from the linker itself:
* we should imply `ILLink5`
* in the future we should imply warning version based on the code (new warnings will get higher numbers, so there will be a clear line - if above this number -> higher version)

For warnings from custom steps - it should basically mean "report always" (ignore warn version) by default (unless they pass a value in this API). Would be interesting how Roslyn plans to handle this for analyzers... but not a big deal I think.

src/linker/Linker/LinkContext.cs
```diff
@@ -496,6 +499,12 @@ public void LogMessage (MessageContainer message)
 				return;
 			}
 
+			if (message.Category == MessageCategory.Warning &&
+				message.Version > WarnVersion) {
```
vitek-karas, line 503: How does this work if the `message.Version` is `null` ... it always evaluates to `false`?, right? (so unspecified version means report always). I think we should add a comment here (it's not obvious that `message.Version` is nullable, and even with that it's not obvious what `<` does on nullables)

src/linker/Linker/MessageContainer.cs
```diff
@@ -61,16 +65,20 @@ public static MessageContainer CreateErrorMessage (string text, int code, string
 		/// for the list of warnings and possibly add a new one</param>
 		/// /// <param name="origin">Filename or member where the warning is coming from</param>
 		/// <param name="subcategory">Optionally, further categorize this warning</param>
+		/// <param name="version">Optional warning version number</param>
 		/// <returns>New MessageContainer of 'Warning' category</returns>
-		public static MessageContainer CreateWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, string subcategory = MessageSubCategory.None)
+		public static MessageContainer CreateWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, string subcategory = MessageSubCategory.None, WarnVersion? version = null)
```
mateoatr, line 70: Why do we default to `null` and not to `Latest`, here and also in LinkContext?

src/linker/Linker/LoggingReflectionPatternRecorder.cs
```diff
@@ -45,7 +45,7 @@ public void RecognizedReflectionAccessPattern (IMemberDefinition source, Instruc
 		public void UnrecognizedReflectionAccessPattern (IMemberDefinition source, Instruction sourceInstruction, IMetadataTokenProvider accessedItem, string message)
 		{
 			var origin = new MessageOrigin (source, sourceInstruction?.Offset);
-			_context.LogWarning (message, 2006, origin, MessageSubCategory.TrimAnalysis);
+			_context.LogWarning (message, 2006, origin, MessageSubCategory.TrimAnalysis, WarnVersion.ILLink5);
```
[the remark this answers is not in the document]
sbomer: That's my plan - unless I notice anything that I think we really need to log in 3.x. Whether it's a regression depends on how you look at it I guess - whether you mind the introduction of warnings more, or the removal of some hard-to-find mesages.

src/linker/Linker/LinkContext.cs
sbomer: As I understand the plan from reading through github issues and PRs, Roslyn allows analyzers to ship with some MSBuild files that are imported to the build, which are expected to map the SDK `AnalysisLevel` to an editorconfig file also shipped with the analyzer. The editorconfig files control warning severity, so this lets analyzers pick warning behavior per wave. They are passed to Roslyn on the command-line and combined with user editorconfigs. I think analyzers are also be able to read the warning version passed directly to Roslyn and make decisions based on that - but the public warning API doesn't mention versions, so it would be up to each analyzer to filter its own warnings based on the version.

At least that's what I pieced together... I could be completely mistaken. :) @agocke would be able to answer better.

If we put a default version in the API, how would we change it for future versions? We need some place that remembers the version when a warning was introduced - currently it would be in the callsite that creates a warning. The current API also has the "report always" behavior when no version is given.

src/linker/Linker/LinkContext.cs
sbomer: Yes, it'll be `false` - I will add a comment.

src/linker/Linker/MessageContainer.cs
sbomer: I don't want any warnings to explicitly have version `Latest` - they should either be unversioned (always on - which is what the `null` does), or versioned with a specific number. I personally find that more clear than giving the unversioned warnings a version `Latest`, since that suggests they were recently introduced.

src/linker/Linker/LinkContext.cs
vitek-karas: I was not asking for changing the default in the API - as in defaulting the value to some constant. I was asking for: If the caller doesn't specify the value, then based on the code, calculate the version.

Reasoning:
* code -> version mapping must be fixed
  * every callsite using a given code must use the same version - hard to guarantee
  * code -> version mapping must not change in the future (easier to guarantee)
* would be easier to review/make sure it's correct if in a single location

Basically - the callsite should really only worry about which warning to use (code). In the future I want us to remove the message from the callsite as well (resources), so the callsite would only specify the code and parameters for the message.

sbomer: I see - that's what I was referring to in the first bullet in the PR. I would like to simplify the API to just take a code and map it to a message and version in one place, but given how contentious that idea was last time, I thought it better to be consistent with the existing API until we decided to factor it.

Maybe now is a good opportunity to do that factoring? I'm curious what @marek-safar thinks.

vitek-karas: I think we need sort of a double treatment here:
* For the warnings from the linker itself - I strongly feel that having the API only take code is the right approach
* For custom steps we need the current API which allows everything to be passed in - since we can't build the code->message/version map into the linker

I'm working on a separate PR which will make sure that we only use one message per code (still no API refactoring though). So once that is done we can do the refactoring and create the map.

src/linker/Linker/LinkContext.cs
```diff
@@ -532,53 +532,53 @@ public void LogDiagnostic (string message)
...
 		/// </summary>
 		/// <param name="text">Humanly readable message describing the warning</param>
 		/// <param name="code">Unique warning ID. Please see https://github.com/mono/linker/blob/master/doc/error-codes.md for the list of warnings and possibly add a new one</param>
 		/// <param name="origin">Filename or member where the warning is coming from</param>
 		/// <param name="subcategory">Optionally, further categorize this warning</param>
-		/// <param name="version">Optional warning version number. Versioned warnings can be controlled with the
-		/// warning wave option --warn VERSION. Unversioned warnings are unaffected by this option. </param>
 		/// <returns>New MessageContainer of 'Warning' category</returns>
-		public void LogWarning (string text, int code, MessageOrigin origin, string subcategory = MessageSubCategory.None, WarnVersion? version = null)
+		public void LogWarning (string text, int code, MessageOrigin origin, string subcategory = MessageSubCategory.None)
```
vitek-karas: I thought we would keep the optional parameter for custom steps - as for those we can't really infer the version.

src/linker/Linker/LinkContext.cs
```diff
@@ -605,6 +605,12 @@ public bool IsWarningSuppressed (int warningCode, MessageOrigin origin)
 
 			return Suppressions.IsSuppressed (warningCode, origin, out _);
 		}
+
+		public static WarnVersion GetWarningVersion (int code)
+		{
+			// TODO: Once we add warnings beyond .NET5, give them a new WarnVersion
+			return WarnVersion.ILLink5;
```
vitek-karas, line 612: If the code is in custom steps range we should return the null version - basically always show the warning (steps should specify warning version via the API)

src/linker/Linker/LinkContext.cs
sbomer: I put the mapping from code -> version into LinkContext.LogWarning, assuming that will be the API that we eventually factor for warnings from the linker itself. PTAL.

src/linker/Linker/LinkContext.cs
sbomer: My thinking is that we have CreateWarningMessage, exposed in the ref assembly, as the API for custom steps, taking an optional warning version. The LinkContext.LogWarning helpers are not exposed in the ref assembly and would be refactored to just take a code - does that seem reasonable?

vitek-karas: Didn't know we don't expose these in ref assembly - in that case I completely agree.

src/linker/Linker/LinkContext.cs
```diff
@@ -585,6 +605,12 @@ public bool IsWarningSuppressed (int warningCode, MessageOrigin origin)
 
 			return Suppressions.IsSuppressed (warningCode, origin, out _);
 		}
+
+		public static WarnVersion GetWarningVersion (int code)
+		{
+			// TODO: Once we add warnings beyond .NET5, give them a new WarnVersion
```
vitek-karas, line 611: Nit: this is no TODO - at least to me :wink:

src/linker/Linker/LinkContext.cs
sbomer: What is the custom step range? As far as I can tell we only expose CreateWarningMessage to custom steps, and we don't check for conflicts with built-in warnings - I thought this is what you meant to do in the PR you mentioned you were working on.

src/linker/Linker/LinkContext.cs
sbomer: Removed TODO and rephrased :)

src/linker/Linker/LinkContext.cs
sbomer: Great - I can see some of those helpers being useful for custom steps too - I figure we can agree on additional APIs later.

src/linker/Linker/LinkContext.cs
vitek-karas: https://github.com/mono/linker/blob/master/docs/error-codes.md#illink-errors-and-warnings specifically:
> The known codes are in the range 1000 to 6000. Custom steps should avoid using this range not to collide with existing or future error and warning codes.

sbomer: I see, then we should change the check in CreateWarningMessage. I'm hesitant to do that here since I don't know whether there are already steps relying on the current behavior - I'll file an issue about it.

sbomer: https://github.com/mono/linker/issues/1384

src/linker/Linker/Driver.cs
```diff
@@ -1059,6 +1084,8 @@ static void Usage ()
 			Console.WriteLine ("  -out PATH                     Specify the output directory. Defaults to 'output'");
 			Console.WriteLine ("  --about                       About the {0}", _linker);
 			Console.WriteLine ("  --verbose                     Log messages indicating progress and warnings");
+			Console.WriteLine ("  --warn VERSION                Only print out warnings with version <= VERSION. Defaults to '9999'");
```
marek-safar, line 1087: This description does not make much sense. We should be explicit about what is supported and that's right now the only version 5

docs/error-codes.md
```diff
@@ -5,6 +5,12 @@ identification. The known codes are in the range 1000 to 6000. Custom
 steps should avoid using this range not to collide with existing or future
 error and warning codes.
 
+For versioned warnings, the warning version is indicated in parentheses following
+the error code. For example:
+
+#### `ILXXXX` (version): Message
```
marek-safar, line 11: What is the added value in duplicating the version in every message? We could just group the warnings per warning with one line

src/linker/ref/Linker/MessageContainer.cs
```diff
@@ -6,8 +6,8 @@ namespace Mono.Linker
 {
 	public readonly struct MessageContainer
 	{
-		public static MessageContainer CreateErrorMessage (string text, int code, string subcategory = "", MessageOrigin? origin = null, bool isWarnAsError = false) { throw null; }
-		public static MessageContainer CreateWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, string subcategory = "") { throw null; }
+		public static MessageContainer CreateErrorMessage (string text, int code, string subcategory = "", MessageOrigin? origin = null, bool isWarnAsError = false, WarnVersion? version = null) { throw null; }
```
marek-safar, line 9: This does not make sense to me `bool isWarnAsError = false, WarnVersion? version = null` is an implementation detail and should not be exposed to public API.

PR #1382, opened 2020-07-23
Warn about some COM usage

MichalStrehovsky:
COM represents a problem for trimming because the usage of the interfaces are not visible to the linker.

The problem is twofold:
* The IDispatch interface that COM objects can be cast to is basically reflection.
* IUnknown.QueryInterface lets the native side get to any COM visible interface on the managed side. COM visible is pretty much everything because it's the default. We can't root those because it would make trimming ineffective.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2574,13 +2574,30 @@ namespace Mono.Linker.Steps
 
 			TypeDefinition returnTypeDefinition = method.ReturnType.Resolve ();
 
+			bool didWarnAboutCom = false;
+
 			const bool includeStaticFields = false;
-			if (returnTypeDefinition != null && !returnTypeDefinition.IsImport) {
-				MarkDefaultConstructor (returnTypeDefinition, new DependencyInfo (DependencyKind.InteropMethodDependency, method), method);
-				MarkFields (returnTypeDefinition, includeStaticFields, new DependencyInfo (DependencyKind.InteropMethodDependency, method));
+			if (returnTypeDefinition != null) {
+				if (!returnTypeDefinition.IsImport) {
+					// What we keep here is correct most of the time, but not every time. Fine for now.
+					MarkDefaultConstructor (returnTypeDefinition, new DependencyInfo (DependencyKind.InteropMethodDependency, method), method);
+					MarkFields (returnTypeDefinition, includeStaticFields, new DependencyInfo (DependencyKind.InteropMethodDependency, method));
+				}
+
+				// Best-effort attempt to find COM marshalling.
+				// It might seem reasonable to allow out parameters, but once we get a foreign object back (an RCW), it's going to look
+				// like a regular managed class/interface, but every method invocation that takes a class/interface implies COM
+				// marshalling. We can't detect that once we have an RCW.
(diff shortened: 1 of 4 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2590,13 +2607,68 @@ void ProcessInteropMethod (MethodDefinition method)
...
+				nativeType = marshalInfoProvider.MarshalInfo.NativeType;
+			}
+
+			if (nativeType == NativeType.IUnknown || nativeType == NativeType.IDispatch || nativeType == NativeType.IntF) {
+				return true;
+			}
+
+			if (nativeType == NativeType.None) {
+				if (parameterType.IsTypeOf ("System", "Array")) {
+					return true;
```
vitek-karas, line 2649: I think it's worth to add some comments as to why - the delicacies of COM marshalling is black magic to pretty everybody but a select few on the runtime team. So having at least some description of why we do these things would be good.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2590,13 +2607,68 @@ void ProcessInteropMethod (MethodDefinition method)
...
+				} else if (parameterType.IsTypeOf ("System", "String") ||
+					parameterType.IsTypeOf ("System", "StringBuilder")) {
+					return false;
+				}
+
+				var parameterTypeDef = parameterType.Resolve ();
+				if (parameterTypeDef != null) {
+					if (parameterTypeDef.IsInterface) {
+						return true;
+					} else if (parameterTypeDef.IsMulticastDelegate ()) {
```
vitek-karas, line 2659: Can we rule out value types before we do the potentially more expensive checks below?

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: Do you have some specific comments in mind?

"If marshalling is unspecified, System.Array marshals as a COM interface" is kind of a useless one-line summary of two lines of code and I can't think of anything better.

vitek-karas: That's exactly the information I don't have - looking at the code it's like "What have arrays to do with COM?". I know it looks obvious to you since you're writing the code now, but it's definitely not obvious to me (and I've read the explanation of these issues in the offline discussions).

MichalStrehovsky:
> What's the impact on framework? How many such warnings show up when we run this over CoreLib and friends?

For a trimmed Console app there's no warnings. No warnings for Blazor either. About a dozen warnings for a WinForms app.

There are some warnings within the framework, e.g. System.Drawing (`Image.Save` uses COM for example since it's just a thin wrapper over GDI+).

MichalStrehovsky:
We should wait merging this until your big warning overhaul goes through.

vitek-karas:
Actually I would rather do it the other way round - merge this when possible - I have some other reshuffling of warnings to do in the overhaul PR.

MichalStrehovsky:
> Actually I would rather do it the other way round - merge this when possible - I have some other reshuffling of warnings to do in the overhaul PR.

Works for me if it's not causing extra trouble for you, thanks! I'll merge once I see the CI green.

PR #1385, opened 2020-07-23
Warnings cleanup

vitek-karas:
The most important change is to make warning codes and messages unique - that is for each code there's exactly one message.
This means adding quite a few new codes as we were sharing them rather heavily.

As I was going over these I also changes some of the messages to make them cleaner and easier to understand.

This change also adds more tests for some of the warnings.

Infra changes:
* Add the warning code parameter to unrecognized reflection call on the interface
* Add a new ExpectedWarning test attribute which validates the code, the location and optionally parts of the message
* Improved the unrecognized reflection pattern attribute to accept only parts of the message (and to accept more than one)
* Add the ability to validat the warning code for unrecognised reflection patterns


vitek-karas:
Please focus on the codes and infra changes. The warning messages are a work in progress.
I went through all the codes and made sure we now don't have any duplicates and so on. We need to get the code changes in relatively soon as we will probably need to modify several suppression attributes in the dotnet/runtime repo to match the new codes.

I plan to work on the messages (probably in a separate PR) and get them into some kind of starting shape. Then I plan to have a round of warning message review (mostly working with just the doc) followed by a PR which will hopefully get them to the final state.

The long term goal for tests is to not have validation for the messages themselves, but rather only for codes and whatever procedural parameters the messages take. This change makes some improvements in that area, but there's still lot of work left to get there.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1332,24 +1327,24 @@ void ProcessCreateInstanceByName (ref ReflectionPatternContext reflectionContext
...
 							if (resolvedAssembly == null) {
-								reflectionContext.RecordUnrecognizedPattern ($"Activator call '{reflectionContext.MemberWithRequirements}' inside '{reflectionContext.Source}' references assembly '{assemblyNameStringValue.Contents}' which could not be found");
+								reflectionContext.RecordUnrecognizedPattern (2056, $"The assembly '{assemblyNameStringValue.Contents}' can not be found");
 								continue;
 							}
 
 							var resolvedType = resolvedAssembly.FindType (typeNameStringValue.Contents);
 							if (resolvedType == null) {
-								reflectionContext.RecordUnrecognizedPattern ($"Activator call '{reflectionContext.MemberWithRequirements}' inside '{reflectionContext.Source}' references type '{typeNameStringValue}' in assembly '{resolvedAssembly.FullName}' which could not be found");
+								reflectionContext.RecordUnrecognizedPattern (2057, $"The type '{typeNameStringValue.Contents}' can not be found in assembly '{resolvedAssembly.FullName}'");
```
sbomer, line 1336: Why do we treat unresolved type as a warning for `CreateInstance` but not for `GetType`?

docs/error-codes.md
```diff
@@ -256,4 +462,166 @@ error and warning codes.
...
+
+  ``` C#
+  void TestMethod(string typeName)
+  {
+      // IL2058 Trim analysis: Unrecognized value passed to the parameter 'typeName' of method 'CreateInstance'. It's not possible to guarantee the availability of the target type.
+      Activator.CreateInstance("MyAssembly", typeName);
+  }
+  ```
+
+#### `IL2059` Trim analysis: Unrecognized value passed to the parameter 'assemblyName' of method 'CreateInstance'. It's not possible to guarantee the availability of the target type.
```
sbomer, line 588: Just checking - we want to share 2056-2059 across various CreateInstance* methods right?

vitek-karas: They're already shared between AppDomain/Assembly/Activator CreateInstance methods

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: Short - no idea - but thinking about it, it does make sense:
* GetType returns null when the type can't be found - so it's reasonable to consider non-failure cases (ignoring the overload which can throw... at least for now)
* CreateInstance throws if the type can't be found - so it makes more sense to treat this as potentially dangerous

But I do agree that this is a tricky thing to get right - I could see us going either way.

MichalStrehovsky: > But I do agree that this is a tricky thing to get right - I could see us going either way.

We should only warn if there is risk trimming will change runtime behavior.

Trying to be helpful like this ("hey, I think you have a bug here") has resulted in support calls whenever we tried that. People have dead code (e.g. unreachable platform specific code) in their apps all the time. These warnings pretty much always find the dead code. If it was live code, people would have noticed. The code is "safe" for trimming purposes and that's all that should matter for linker.

I would vote to get rid of this warning.

I also filed #1387 because `FindType` is... inadequate.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -250,7 +250,8 @@ TypeAnnotations BuildTypeAnnotations (TypeDefinition type)
 						}
 
 						if (annotatedMethods.Any (a => a.Method == setMethod)) {
-							_context.LogWarning ($"Trying to propagate DynamicallyAccessedMemberAttribute from property '{property.FullName}' to its setter '{setMethod.GetDisplayName ()}', but it already has such attribute on the 'value' parameter.", 2043, setMethod);
+							_context.LogWarning ($"Trying to propagate DynamicallyAccessedMemberAttribute from property '{property.FullName}' to its setter '{setMethod.GetDisplayName ()}', but it already has such attribute on the 'value' parameter.",
```
mateoatr, line 253: Should this be DynamicallyAccessedMember**s**Attribute (notice the 's')? If that's the case, this happens in multiple places.

src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -65,6 +65,10 @@ IEnumerable<CustomAttribute> ProcessAttributes (XPathNavigator nav, ICustomAttri
 				else {
 					try {
 						assembly = GetAssembly (Context, AssemblyNameReference.Parse (assemblyName));
+						if (assembly == null) {
```
mateoatr, line 68: It looks like `GetAssembly` will throw whenever `assembly == null` (thus never hitting this code).

test/Mono.Linker.Tests.Cases.Expectations/Assertions/UnrecognizedReflectionAccessPatternAttribute.cs
```diff
@@ -16,35 +16,24 @@ namespace Mono.Linker.Tests.Cases.Expectations.Assertions
...
 	public class UnrecognizedReflectionAccessPatternAttribute : BaseExpectedLinkedBehaviorAttribute
 	{
-		public UnrecognizedReflectionAccessPatternAttribute (Type reflectionMethodType, string reflectionMethodName, Type[] reflectionMethodParameters,
-			string message = null,
+		public UnrecognizedReflectionAccessPatternAttribute (
+			Type reflectionMethodType,
+			string reflectionMethodName,
+			Type[] reflectionMethodParameters,
+			object message = null,
+			string messageCode = null,
```
mateoatr, line 24: Why isn't this `int`?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: I'll remove this warning

src/linker/Linker.Dataflow/FlowAnnotations.cs
vitek-karas: Good catch - fixed everwhere.

src/linker/Linker.Steps/LinkAttributesStep.cs
vitek-karas: It actually can return null (`--skip-unresolved true` will do that). There's a bug in this step but in another place - thanks for pointing this out - fixed.

test/Mono.Linker.Tests.Cases.Expectations/Assertions/UnrecognizedReflectionAccessPatternAttribute.cs
vitek-karas: I intentionally made all the warning codes in tests strings:
* It makes the tests seem operate on the visible output (the fact that sometimes we check against the internal number is just an implementation detail)
* It makes the tests closer to the user XP
* (low pri) I still hope that one day we will be able to use non-numerical values

docs/error-codes.md
```diff
@@ -266,4 +472,154 @@ the error code. For example:
...
+  void TestMethod()
+  {
+      // IL2050 Trim analysis: Making a generic type instantiation from 'Lazy<T>' which has 
+      // 'DynamicallyAccessedMembersAttribute' on some of its generic parameters. ILLink currently doesn't analyze 
+      // type values for generic parameters when making a generic type instantiation via 'System.Type.MakeGenericType'
+      typeof(Lazy<>).MakeGenericType(new Type[] { typeof(TestType) });
+  }
+  ```
+
+  #### `IL2051` Trim analysis: Unrecognized type value used to called the method 'MakeGenericType'. It's not possible to guarantee the availability of requirements of such type.
```
marek-safar, line 496: I don't understand the warning message :-(

docs/error-codes.md
```diff
@@ -266,4 +472,154 @@ the error code. For example:
...
+
+- If `System.Type.MakeGenericType` is called on an type which is not statically analyzable. If the actual type has generic parameters with `DynamicallyAccessedMembersAttribute` ILLink would be required to fulfill the requirements declared by those attributes, but since the ILLink doesn't know the type, it can't determine if such requirements exist.  
+
+  ``` C#
+  void TestMethod(Type type)
+  {
+      // IL2051 Trim analysis: Unrecognized value passed to the parameter 'typeArguments' of method 'System.Type.MakeGenericType'
```
marek-safar, line 503: Should we copy the message here (it can go out of sync) ?

docs/error-codes.md
```diff
@@ -75,63 +75,269 @@ the error code. For example:
...
-#### `IL2007`: Could not resolve assembly 'assembly' specified in the 'XML document location'
+  ```C#
+  // IL2005: Could not resolve dependency member 'NonExistentMethod' declared on type 'MyType' specified in a 'PreserveDependency' attribute
+  [PreserveDependency("NonExistentMethod", "MyType", "MyAssembly")]
+  void TestMethod()
+  {
+  }
+  ```
+
+#### `IL2006` Trim analysis: The requirements declared via the `DynamicallyAccessedMembersAttribute` on <value description> don't match those on <target description>. The source value must declare at least the same requirements as those declared on the target location it's assigned to
```
marek-safar, line 149: What is the benefit of `Trim analysis` in the warning message for the developers?

docs/error-codes.md
```diff
@@ -266,4 +472,154 @@ the error code. For example:
 
 #### `IL2049`: Unrecognized internal attribute 'attribute'
 
-- The internal attribute name 'attribute' being used in the xml is not supported by the linker, check the spelling and the supported internal attributes.
\ No newline at end of file
+- The internal attribute name 'attribute' being used in the xml is not supported by the linker, check the spelling and the supported internal attributes.
+
+#### `IL2050` Trim analysis: Making a generic type instantiation from 'type' which has 'DynamicallyAccessedMembersAttribute' on some of its generic parameters. ILLink currently doesn't analyze type values for generic parameters when making a generic type instantiation via 'System.Type.MakeGenericType'
```
marek-safar, line 477: Could you unify quotes around DynamicallyAccessedMembersAttribute and other names. It looks like there are for some reasons used with "`", "'" and with no quotes

marek-safar: It looks like the warning message is missing the reason.

```suggestion
#### `IL2050` Making a generic type instantiation from 'type' which has 'DynamicallyAccessedMembersAttribute' on some of its generic parameters is not supported.
```

docs/error-codes.md
```diff
@@ -266,4 +472,154 @@ the error code. For example:
...
+
+  ``` C#
+  void TestMethod(Type type)
+  {
+      // IL2051 Trim analysis: Unrecognized value passed to the parameter 'typeArguments' of method 'System.Type.MakeGenericType'
+      type.MakeGenericType(new Type[] { typeof(TestType) });
+  }
+  ```
+
+#### `IL2052` Trim analysis: Unrecognized value passed to the parameter 'typeName' of method 'System.Type.GetType(Type typeName)'. It's not possible to guarantee the availability of the target type.
```
marek-safar, line 508:
```suggestion
#### `IL2052` Argument type used for parameter 'typeName' of method 'System.Type.GetType(Type typeName)' cannot be determined.`
```

docs/error-codes.md
```diff
@@ -266,4 +472,154 @@ the error code. For example:
...
+  void TestMethod()
+  {
+      string typeName = ReadName();
+
+      // IL2052 Trim analysis: Unrecognized value passed to the parameter 'typeName' of method 'System.Type.GetType(Type typeName)'
+      Type.GetType(typeName);
+  }
+  ```
+
+#### `IL2053` Trim analysis: 'Assembly.CreateInstance' is not supported with trimming. Use 'Type.GetType' and `Activator.CreateInstance` instead
```
marek-safar, line 522:
```suggestion
#### `IL2053` Arguments used for 'System.Assembly.CreateInstance(...)' method cannot be analyzed. Consider using `System.Activator.CreateInstance` method instead.
```

docs/error-codes.md
```diff
@@ -266,4 +472,154 @@ the error code. For example:
...
+  {
+      // IL2053 Trim analysis: 'Assembly.CreateInstance' is not supported with trimming. Use 'Type.GetType' instead
+      AssemblyLoadContext.Default.Assemblies.First(a => a.Name == "MyAssembly").CreateInstance("MyType");
+
+      // This can be replaced by
+      Activator.CreateInstance(Type.GetType("MyType, MyAssembly"));
+  }
+  ```
+  
+#### `IL2054` Trim analysis: Unrecognized value passed to the parameter 'type' of method 'System.Runtime.CompilerServices.RuntimeHelpers.RunClassConstructor'. It's not possible to guarantee the availability of the target static constructor.
```
marek-safar, line 537: We should be consistent in the warning message. If we include method name we should include it in the same form in all warnings (e.g. with parameters and parameter names)

docs/error-codes.md
```diff
@@ -266,4 +472,154 @@ the error code. For example:
...
+
+  ``` C#
+  void TestMethod()
+  {
+      // IL2055 Trim analysis: Call to `System.Reflection.MethodInfo.MakeGenericMethod` can not be statically analyzed. It's not possible to guarantee the availability of requirements of the generic method.
+      typeof(MyType).GetMethod("MyMethod").MakeGenericMethod(new Type[] { typeof(MyType) });
+  }
+  ```
+
+#### `IL2056` Trim analysis: The assembly 'assembly name' can not be found
```
marek-safar, line 562:
```suggestion
#### `IL2056` The assembly name 'name' used in method call 'method' references assembly which is not available.
```

docs/error-codes.md
```diff
@@ -266,4 +472,154 @@ the error code. For example:
...
+
+  ``` C#
+  void TestMethod(string typeName)
+  {
+      // IL2058 Trim analysis: Unrecognized value passed to the parameter 'typeName' of method 'CreateInstance'. It's not possible to guarantee the availability of the target type.
+      Activator.CreateInstance("MyAssembly", typeName);
+  }
+  ```
+
+#### `IL2059` Trim analysis: Unrecognized value passed to the parameter 'assemblyName' of method 'CreateInstance'. It's not possible to guarantee the availability of the target type.
```
marek-safar, line 586: Why not use IL2058 for this?

docs/error-codes.md
```diff
@@ -266,4 +472,154 @@ the error code. For example:
...
+
+  ``` C#
+  void TestMethod(string assemblyName)
+  {
+      // IL2059 Trim analysis: Unrecognized value passed to the parameter 'assemblyName' of method 'CreateInstance'. It's not possible to guarantee the availability of the target type.
+      Activator.CreateInstance(assemblyName, "MyType");
+  }
+  ```
+
+#### `IL2060` Trim analysis: Unrecognized type value passed to <target description>. It's not possible to guarantee that the requirements declared by the `DynamicallyAccessedMembersAttribute` are met.
```
marek-safar, line 598: I don't understand this warning message.:-(

docs/error-codes.md
```diff
@@ -240,9 +446,9 @@ the error code. For example:
 
 - The property 'property' has DynamicallyAccessedMembersAttribute on it, but the linker could not determine the backing fields for the property to propagate the attribute to the field.
 
-#### `IL2043`: Trying to propagate DynamicallyAccessedMemberAttribute from property 'property' to its getter 'method', but it already has such attribute.
+#### `IL2043`: Trying to propagate DynamicallyAccessedMemberAttribute from property 'property' to its setter 'method', but it already has such attribute.
```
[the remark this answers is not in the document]
vitek-karas: yes - I guess we could check if they're different and warn about that, but in general there needs to be a warning in this place.

docs/error-codes.md
vitek-karas: I was trying to show how the message would look with specific references to code - since the message with parametrized pieces can be hard to understand.

docs/error-codes.md
vitek-karas: As per the code sample - calling `MakeGenericType` on a `Type` instance we can't statically analyze - if that type has annotations on its generic parameters, we would be required to validate those - but can't since we don't know the type.
If anybody has a better idea about the message, I'm more than happy to change the text.

docs/error-codes.md
vitek-karas: This is the subcategory - I'm adding it to this doc both as documentation and as a way to quickly review the subcategories. It will directly map to the `TrimAnalysis` property in MSBuild (which will suppress these messages).

docs/error-codes.md
vitek-karas: As per the example, it is used when a value which we can't statically analyze is assigned to something (target) which has data flow annotation - and thus linker can't guarantee that the annotation's requirement is fulfilled. I welcome any suggestions to improve the message.

docs/error-codes.md
vitek-karas: I tried to unify it like this:
* In messages it should be single-quoted : 'DynamicallyAccessedAttribute' - just like any other reference to names from the code (all method names, type names and so on should also be single-quoted)
* In the documentation it should be backtick-quoted so that the docs show it formatted as code.
* There should be no double quotes anywhere

vitek-karas: As for the message - I don't think we should say "not supported" - it is supported in the code, it's just that linker doesn't analyze it.

docs/error-codes.md
vitek-karas: I find the new message confusing - "Argument type" to me means the declared type of the argument, which in this case will pretty always be `System.Type` - and that one was determined (otherwise linker would not have a problem with it). It's the value passed to the argument which we could not figure out.
I guess there's also a question of terminology - argument or parameter?

docs/error-codes.md
vitek-karas: I tried to not use full formatting the message titles as these are all parametrized in code. The message in the sample tries to be better about the formatting but still not perfect (I didn't run linker on the code to get the exact string it would produce).

docs/error-codes.md
vitek-karas: Good point - fixed.

docs/error-codes.md
marek-safar: How is that different from IL2052? Do we need to expose the limitation over warning codes?

Does it make sense to tweak IL2052 to cover all such cases?

```
#### `IL2052` Method `MakeGenericType(...)` is called on instance which cannot be determined.
```

docs/error-codes.md
marek-safar: hmm, not sure this is best option. Maybe we could add Category bellow to not break the search

docs/error-codes.md
marek-safar: ok but I think the sentence needs tweaking as "Making a generic type instantiation from 'type' which has 'DynamicallyAccessedMembersAttribute' on some of its generic parameters" is weird

docs/error-codes.md
marek-safar: I think this is quite clear. From Wikipedia

> Unlike argument in usual mathematical usage, the argument in computer science is thus the actual input expression passed/supplied to a function, procedure, or routine in the invocation/call statement, whereas the parameter is the variable inside the implementation of the subroutine. 

I agree we need to settle on how we are going to identify the type and we are not consistent in other messages. In this case, it could be `Argument instance used for parameter ....`

docs/error-codes.md
marek-safar: I think this could be something like

```
#### `IL2059` Argument instance used for parameter 'x' in method 'm' doesn't need to meet DynamicallyAccessedMembersAttribute requirements.
```

docs/error-codes.md
vitek-karas: That message is also not correct - but what about this:
```
IL2051 Trim analysis: Call to `System.Reflection.MethodInfo.MakeGenericType` can not be statically analyzed. It's not possible to guarantee the availability of requirements of the generic type.
```

Basically the same as IL2056 but for types. And to avoid future question - I do want to keep the `MakeGenericType` and `MakeGenericMethod` warnings separate. If nothing else, to give us freedom of improving the warning message if the current one is not good enough (and sharing it between both cases would make it much harder to come up with a message)

docs/error-codes.md
vitek-karas: I don't understand the concern? Is it about the ability to search this document?

Note, the output of the linker looks like this:
```
File.cs(11,22): Trim analysis warning IL2006: The requirements ....
```

docs/error-codes.md
vitek-karas: What about:
```
Value passed to <target description> can not be statically determined and may not meet 'DynamicallyAccessedMembersAttribute' requirements.
```

So specifically for a method parameter:
```
Value passed to the parameter 'x' of method 'm' can not be statically determined and may not meet 'DynamicallyAccessedMembersAttribute' requirements.
```

docs/error-codes.md
marek-safar: Yep, this is better

docs/error-codes.md
marek-safar: Yep, I think if I copy `IL2006: The requirements` text it should find me the doc about the warning. I don't see a value in having `Trim analysis` in the warning output. It just makes the output verbose and harder to read as different warnings have different formats and the information about Trim analysis is or could be already included in the warning code

docs/error-codes.md
marek-safar: I think that's better but `Value passed to the parameter 'x' of method 'm'` should have its own warning code because if someone will be searching for `Value passed to the parameter` it probably won't find this reference easily

docs/error-codes.md
vitek-karas: Thanks for the explanation. I will take a note and do this sweeping change in a separate PR (this is already getting out of hand). I will have another PR focusing just on the messages anyway.

docs/error-codes.md
vitek-karas: There are currently 5 cases:
* `parameter 'name' of method 'name'`
* `return value of method 'name'`
* `field 'name'`
* `implicit 'this' parameter of method 'name'`
* `generic parameter 'name' from 'type/method'`

You think we should have separate code for each of them?

marek-safar: I'd say yes as they are about construction and would allow us to have better warning message for each case

vitek-karas: What about 2047 - it uses the same pattern. For example:
```
IL2047: 'DynamicallyAccessedMemberTypes' in 'DynamicallyAccessedMembersAttribute' on parameter 'type' on method 'Derived.TestMethod' don't match overridden parameter 'type' on method 'Base.TestMethod'. All overridden members must have the same 'DynamicallyAccessedMembersAttribute' usage.
```

marek-safar: Yeah, that will need the same treatment

vitek-karas:
This should be it - I'm not aware of any more warnings which would need attention. @sbomer can you please take a look - specifically please review if the warnings are "split" correctly.

sbomer:
LGTM!

Issue #1380, opened 2020-07-22
Add an option to generate the suppressions in the linker attribute XML format


vitek-karas:
Currently the `--generate-warning-suppressions` will produce the suppressions in C# - as a C# source code using the attribute syntax. This is useful if I want to make the suppressions part of my shipping code (as I include the source file with my project), but it's not useful if I only want to use the suppressions within the build system of my project, but not include the suppressions in the final bits.

One way to do the latter would be to produce the suppressions in the XML format, and only pass them as a command line for the linker - for that invocation of the linker those suppressions would take effect, but they would not get baked into the final build.

See https://github.com/dotnet/runtime/pull/39133 for detailed discussion why this is an issue in the dotnet/runtime repo

As a technical solution, we could change the `--generate-warning-suppressions` option to take a value (not just `true`/`false`) - so basically either `--generate-warning-suppressions cs` or `--generate-warning-suppressions xml`.

PR #1386, opened 2020-07-24
Generate warning suppressions in an XML format 

mateoatr:
See #1380. This changes the existing `--generate-warning-suppressions` option to receive a string argument (`cs`/`xml`) which tells the suppression writer in which format should the warning suppressions be emitted, either as a C# source file (this is the current behavior) or as an XML file that follows the linker's annotation format.

/cc @layomia


src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -2,19 +2,30 @@
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Xml;
+using System.Xml.Linq;
 using Mono.Cecil;
 
 namespace Mono.Linker
 {
+	public enum WarningSuppressionWriterFileOutputKind
```
marek-safar, line 11: Why is it not nested enum inside `WarningSuppressionWriter` ?

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -2,19 +2,30 @@
...
+	{
+		Cs,
```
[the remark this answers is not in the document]
marek-safar: I'd also call it CSharp

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -31,46 +42,87 @@ public void AddWarning (int code, IMemberDefinition memberDefinition)
...
+				if (lhs == rhs)
+					return a.Code.CompareTo (b.Code);
+
+				return string.CompareOrdinal (lhs, rhs);
+			});
+
+			return listOfWarnings;
+		}
+
+		string GetWarningSuppressionScopeString (IMemberDefinition member)
```
marek-safar, line 113: Should be static

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -31,46 +42,87 @@ public void AddWarning (int code, IMemberDefinition memberDefinition)
...
+
+				sw.Write (sb.ToString ());
 			}
 		}
+
+		List<(int Code, IMemberDefinition Member)> GetListOfWarnings (AssemblyNameDefinition assemblyName)
+		{
+			List<(int Code, IMemberDefinition Member)> listOfWarnings = _warnings[assemblyName].ToList ();
+			listOfWarnings.Sort ((a, b) => {
+				string lhs = a.Member is MethodReference lhsMethod ? lhsMethod.GetDisplayName () : a.Member.FullName;
```
marek-safar, line 102: Will this logic work for members in nested types? I guess Cecil FullName won't compare well with GetDisplayName

src/linker/Linker/Driver.cs
```diff
@@ -430,8 +430,10 @@ protected int SetupContext (ILogger customLogger = null)
 						if (!GetStringParam (token, l => generateWarningSuppressionsArgument = l))
 							return -1;
 
-						if (!GetWarningSuppressionWriterFileOutputKind (generateWarningSuppressionsArgument, out var fileOutputKind))
+						if (!GetWarningSuppressionWriterFileOutputKind (generateWarningSuppressionsArgument, out var fileOutputKind)) {
+							context.LogError ($"Invalid value '{generateWarningSuppressionsArgument}'specified for '--generate-warning-suppressions' option", 2061);
```
vitek-karas, line 434:
```suggestion
							context.LogError ($"Invalid value '{generateWarningSuppressionsArgument}' specified for '--generate-warning-suppressions' option", 2061);
```

src/linker/Linker/Driver.cs
```diff
@@ -938,6 +946,22 @@ static bool GetOptimizationName (string text, out CodeOptimizations optimization
...
+			switch (text.ToLowerInvariant ()) {
+			case "cs":
+				fileOutputKind = WarningSuppressionWriterFileOutputKind.Cs;
+				return true;
+			case "xml":
+				fileOutputKind = WarningSuppressionWriterFileOutputKind.Xml;
+				return true;
+			}
+
+			Console.WriteLine ($"Invalid warning suppressions file format '{text}'");
```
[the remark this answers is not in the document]
mateoatr: You're right, following https://github.com/mono/linker/pull/1364/ this should  be an error.

src/linker/Linker/WarningSuppressionWriter.cs
mateoatr: I think I don't understand why would there be a problem with members in nested types. We only use this strings for sorting the list.

marek-safar: I believe GetDisplayName uses `.` as separator whereas FullName uses `+` as separator

src/linker/Linker/WarningSuppressionWriter.cs
```diff
@@ -31,46 +36,93 @@ public void AddWarning (int code, IMemberDefinition memberDefinition)
...
-
-						DocumentationSignatureGenerator.Instance.VisitMember (warningOrigin, sb);
-						sb.AppendLine ("\")]");
-					}
-
-					sw.Write (sb.ToString ());
+				if (_fileOutputKind == FileOutputKind.Xml)
+					OutputSuppressionsXmlFormat (assemblyName);
+
+				OutputSuppressionsCsFormat (assemblyName);
```
layomia, line 42: Should this be in an `else` branch?

src/linker/Linker/WarningSuppressionWriter.cs
sbomer: I think it'd be good to use the signature to sort, since that's what we're outputing.

PR #1388, opened 2020-07-24
Add WarningsAsErrors and WarningsNotAsErrors

mateoatr:
Adds the ILLink Task options `WarningsAsErrors` and `WarningsNotAsErrors` which map to the command-line options `--warnaserror` and `--warnaserror-` respectively.

### DIFF
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -334,7 +334,7 @@ namespace ILLink.Tasks
 				args.Append ("--warn ").AppendLine (Quote (Warn));
 
 			if (TreatWarningsAsErrors)
-				args.AppendLine ("--warnaserror");
+				args.Append ("--warnaserror");
 			else
 				args.AppendLine ("--warnaserror-");
 
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -71,7 +71,8 @@ namespace ILLink.Tasks
 		/// Treat all warnings as errors.
 		/// Maps to '--warnaserror' if true, '--warnaserror-' if false.
 		/// </summary>
-		public bool TreatWarningsAsErrors { get; set; }
+		public bool TreatWarningsAsErrors { set => _treatWarningsAsErrors = value; }
+		bool? _treatWarningsAsErrors;
 
 		/// <summary>
 		/// The list of warnings to report as errors.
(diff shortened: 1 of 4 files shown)


src/ILLink.Tasks/LinkTask.cs
```diff
@@ -67,6 +67,18 @@ public class ILLink : ToolTask
...
+		/// Maps to '--warnaserror'.
+		/// </summary>
+		public string WarningsAsErrors { get; set; }
+
+		/// <summary>
+		/// The list of warnings to report as usual.
+		/// Maps to '--warnaserror-'.
+		/// </summary>
+		public string WarningsNotAsErrors { get; set; }
+
```
sbomer, line 81: Would you be able to add a boolean `TreatWarningsAsErrors` for the global setting? This should go first on the command-line, so that the individual settings take priority.

test/ILLink.Tasks.Tests/ILLink.Tasks.Tests.cs
```diff
@@ -306,6 +302,40 @@ public void TestWarn (string warnArg, WarnVersion expectedVersion)
 			}
 		}
 
+		[Theory]
+		[InlineData ("", "blah", true, new uint[] { }, new uint[] { })]
```
sbomer, line 306: Just FYI, when invoked from MSBuild targets, I think that the task will get a `null` (rather than an empty string) if the task parameter is not set or if it is set to an empty MSBuild property.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -67,6 +67,24 @@ public class ILLink : ToolTask
 		/// </summary>
 		public string Warn { get; set; }
 
+		/// <summary>
+		/// Treat all warnings as errors.
+		/// Maps to '--warnaserror' if true, '--warnaserror-' if false.
+		/// </summary>
+		public bool TreatWarningsAsErrors { get; set; }
```
sbomer, line 74: Could you use a nullable bool like we do for some of the other boolean options? That way if the task is used without this setting, it won't be passed on the command-line - just to make it a little cleaner.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -315,6 +333,17 @@ protected override string GenerateResponseFileCommands ()
 			if (Warn != null)
 				args.Append ("--warn ").AppendLine (Quote (Warn));
 
+			if (TreatWarningsAsErrors)
+				args.AppendLine ("--warnaserror");
```
vitek-karas, line 337:
```suggestion
				args.Append ("--warnaserror");
```

Also below.

Issue #1390, opened 2020-07-28
Trimming projects with satellite assemblies results in null ref


pranavkm:
```
Fatal error in IL Linker
  Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
     at Mono.Linker.Steps.MarkStep.GetMethodWithNoParameters(TypeDefinition type, String methodname)
     at Mono.Linker.Steps.MarkStep.MarkTypeWithDebuggerDisplayAttribute(TypeDefinition type, CustomAttribute attribute)
     at Mono.Linker.Steps.MarkStep.MarkTypeSpecialCustomAttributes(TypeDefinition type)
     at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, IMemberDefinition sourceLocationMember)
     at Mono.Linker.Steps.MarkStep.MarkField(FieldDefinition field, DependencyInfo& reason)
     at Mono.Linker.Steps.MarkStep.MarkEntireType(TypeDefinition type, Boolean includeBaseTypes, DependencyInfo& reason, IMemberDefinition sourceLocationMember)
     at Mono.Linker.Steps.MarkStep.MarkEntireAssembly(AssemblyDefinition assembly)
     at Mono.Linker.Steps.MarkStep.InitializeAssembly(AssemblyDefinition assembly)
     at Mono.Linker.Steps.MarkStep.Initialize()
     at Mono.Linker.Steps.MarkStep.Process(LinkContext context)
     at Mono.Linker.Pipeline.Process(LinkContext context)
     at Mono.Linker.Driver.Run(ILogger customLogger)
     at Mono.Linker.Driver.Main(String[] args)
```

SDK: v5.0.100-rc.1.20377.3

marek-safar:
Should be fixed with #1376

PR #1391, opened 2020-07-28
Prevent null-ref in GetMethodWithNoParameters

pranavkm:
Fixes https://github.com/mono/linker/issues/1390

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1038,7 +1038,7 @@ namespace Mono.Linker.Steps
 				if (method != null)
 					return method;
 
-				type = type.BaseType.Resolve ();
+				type = type.BaseType?.Resolve ();
 			}
 
 			return null;


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1038,7 +1038,7 @@ MethodDefinition GetMethodWithNoParameters (TypeDefinition type, string methodna
 				if (method != null)
 					return method;
 
-				type = type.BaseType.Resolve ();
+				type = type.BaseType?.Resolve ();
```
pranavkm, line 1041: I don't necessarily know if this fixes the issue, but this looks like an outlier compared to:

* https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/MarkStep.cs#L1021-L1028
* https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/MarkStep.cs#L985-L992

tlakollo:
This seems to be a dup for #1376

marek-safar:
Closing as dup of ##1376

PR #1395, opened 2020-07-31
Add suppressions before checking for ca removal instances

### LINKED ISSUE #1392: Suppressed messages are no longer suppressed when UnconditionalSuppressMessageAttribute is marked for removal
### Repro steps

1. Download the latest SDK (https://github.com/dotnet/installer#installers-and-binaries). I'm using `5.0.100-rc.1.20378.7`.
2. `dotnet new blazorwasm` using the latest SDK
3. Add `<_ExtraTrimmerArgs>$(_ExtraTrimmerArgs) --verbose</_ExtraTrimmerArgs>` to the new .csproj that was created so you see the linker warnings.
4. `dotnet.exe publish -c Release`
5. Inspect the warnings. You will see warnings that are suppressed in code still show up.

ex. 

```
ILLink : Trim analysis warning IL2006: System.Runtime.InteropServices.Marshal.GetCustomMarshalerInstance(Type,String): The parameter 'type' of method 'System.Runtime.InteropServices.Marshal.GetCustomMarshalerInstance(Type,String)' with dynamically accessed member kinds 'None' is passed into the implicit 'this' parameter of method 'System.Type.GetMethod(String,BindingFlags,Binder,Type[],ParameterModifier[])' which requires dynamically accessed member kinds 'NonPublicMethods | PublicMethods'. To fix this add DynamicallyAccessedMembersAttribute to it and specify at least these member kinds 'NonPublicMethods | PublicMethods'. [F:\DotNetTest\LinkerSuppressBugRepro\LinkerSuppressBugRepro.csproj]
```

However, this is suppressed in code:

https://github.com/dotnet/runtime/blob/6072e4d3a7a2a1493f514cdf4be75a3d56580e84/src/mono/netcore/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.Mono.cs#L376-L378

```C#
        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2006:UnrecognizedReflectionPattern",
            Justification = "Implementation detail of MarshalAs.CustomMarshaler")]
        internal static ICustomMarshaler? GetCustomMarshalerInstance(Type type, string cookie)
        {
```

### Analysis

It appears to me that since [WASM is removing the `UnconditionalSuppressMessageAttribute`](https://github.com/dotnet/runtime/blob/9dc8384cf5551947f8be3f770bc54b611dc6c477/src/mono/netcore/System.Private.CoreLib/src/ILLink/ILLink.LinkAttributes.wasm.xml#L30-L32), that this is telling the linker to no longer respect the suppressions. If I remove those lines in the `ILLink.LinkAttributes.wasm.xml` file, the warnings that are already suppressed no longer appear in the output.

cc @vitek-karas @mateoatr

mateoatr:
Fixes #1392. This will add suppressions before checking if the attribute (`UnconditionalSuppressMessage`) was marked for removal by a custom attribute annotation.

/cc @eerhardt

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -558,6 +558,9 @@ namespace Mono.Linker.Steps
 						continue;
 					}
 
+					if (UnconditionalSuppressMessageAttributeState.TypeRefHasUnconditionalSuppressions (ca.Constructor.DeclaringType))
+						_context.Suppressions.AddSuppression (ca, provider);
+
 					if (_context.Annotations.HasLinkerAttribute<RemoveAttributeInstancesAttribute> (ca.AttributeType.Resolve ()) && providerInLinkedAssembly)
 						continue;
 
@@ -1590,8 +1593,6 @@ namespace Mono.Linker.Steps
 					provider,
 					sourceLocationMember);
 				return true;
-			} else if (UnconditionalSuppressMessageAttributeState.TypeRefHasUnconditionalSuppressions (dt)) {
-				_context.Suppressions.AddSuppression (ca, provider);
 			}
 
 			return false;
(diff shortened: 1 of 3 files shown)


marek-safar:
Any chance to include test as well?

PR #1396, opened 2020-07-31
Do not call MarkInterfaceImplementation if already marked

### LINKED ISSUE #1394: Linker enters cycle when compiling ASP.NET benchmarks
Unfortunately I couldn't figure out a minimal repro quickly. To repro, clone https://github.com/aspnet/Benchmarks and build and link with `dotnet.exe publish -f net5.0 -c release -r win-x64 -p:BenchmarksTargetFramework=net5.0 -p:PublishTrimmed=true -p:PublishReadyToRun=true -p:PublishSingleFile=true src/Bechmarks/Benchmarks.csproj`

The linking step will continue forever. After debugging, I think this is a problem with the default interface implementations pass.

MichalStrehovsky:
Fixes #1394.

The problem was that marking the interface implementation marks the custom attributes on the InterfaceImpl record which enqueues the attribute constructor. The constructor is already processed, but enqueuing a new method causes the main processing loop to run again. This will rerun the virtual method logic and cause the attribute constructor to be re-added, forever.

Normally, linker protects itself from things like this by having a "processed" state along with "marked" state for various entities. InterfaceImpls don't have a "processed" state and calling Mark will cause the same logic to run. Alternative fix would be to add a "processed" state.

There's no regression test because I couldn't get Roslyn to generate the custom attribute on the MethodImpl record. I'm unclear why there's an attribute in the EF case to begin with. The problematic MethodImpl is this one: https://github.com/dotnet/efcore/blob/59734ea22f29d22f8d0c1673c59a99c54ec4e78d/src/EFCore.Relational/Scaffolding/ProviderCodeGenerator.cs#L15. There's a NullableAttribute on the impl, but I'm quite puzzled why.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -529,8 +529,7 @@ namespace Mono.Linker.Steps
 			if (!Annotations.IsInstantiated (typeWithDefaultImplementedInterfaceMethod))
 				return;
 
-			if (!Annotations.IsMarked (implementation))
-				MarkInterfaceImplementation (implementation, typeWithDefaultImplementedInterfaceMethod);
+			MarkInterfaceImplementation (implementation, typeWithDefaultImplementedInterfaceMethod);
 		}
 
 		void MarkMarshalSpec (IMarshalInfoProvider spec, in DependencyInfo reason, IMemberDefinition sourceLocationMember)
@@ -2898,6 +2897,9 @@ namespace Mono.Linker.Steps
 
 		protected virtual void MarkInterfaceImplementation (InterfaceImplementation iface, TypeDefinition type)
 		{
+			if (Annotations.IsMarked (iface))
+				return;
+
 			// Blame the type that has the interfaceimpl, expecting the type itself to get marked for other reasons.
 			MarkCustomAttributes (iface, new DependencyInfo (DependencyKind.CustomAttribute, iface), type);
 			// Blame the interface type on the interfaceimpl itself.
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -529,7 +529,8 @@ namespace Mono.Linker.Steps
 			if (!Annotations.IsInstantiated (typeWithDefaultImplementedInterfaceMethod))
 				return;
 
-			MarkInterfaceImplementation (implementation, typeWithDefaultImplementedInterfaceMethod);
+			if (!Annotations.IsMarked (implementation))
+				MarkInterfaceImplementation (implementation, typeWithDefaultImplementedInterfaceMethod);
 		}
 
 		void MarkMarshalSpec (IMarshalInfoProvider spec, in DependencyInfo reason, IMemberDefinition sourceLocationMember)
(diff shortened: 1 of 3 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -529,7 +529,8 @@ void ProcessDefaultImplementation (TypeDefinition typeWithDefaultImplementedInte
 			if (!Annotations.IsInstantiated (typeWithDefaultImplementedInterfaceMethod))
 				return;
 
-			MarkInterfaceImplementation (implementation, typeWithDefaultImplementedInterfaceMethod);
+			if (!Annotations.IsMarked (implementation))
```
vitek-karas, line 532: Should we move this check into the `MarkInterfaceImplementation`? There are multiple callsites for that method - I guess none has this problem... but it would not hurt and would make the code more robust.

agocke:
> There's no regression test because I couldn't get Roslyn to generate the custom attribute on the MethodImpl record

Roslyn tries to save space with attributes by figuring out what the most common case is in the type and then only putting attributes on the less common cases. Is it possible that the methodimpl in your test is not getting marked because the most common nullability in your file matches the nullability in your impl?

agocke:
Also, can you not create a linker test with IL?

MichalStrehovsky:
> Roslyn tries to save space with attributes by figuring out what the most common case is in the type and then only putting attributes on the less common cases

I still don't understand why we need the nullability annotation on the interface impl record though. In pseudo-C#, it's this kind of annotation:

```csharp
class Foo : [Nullable(0)] IFoo
{
}
```

What nullability information do we need for a non-generic interface? That `this` is not null?

This is still a linker bug, but the IL generated by Roslyn doesn't make sense to me (and might as well be a bug that was fixed, explaining why I can't get it to repro in C#).

> Also, can you not create a linker test with IL?

So I wasted some time making the test now in a shape that would be consumable by the test system here...

...and it doesn't work because the desktop ILASM (that is used to build IL tests in this repo) cannot generate default interface methods because it has a damn `if` check that we deleted as the first thing when standing up the default interface methods feature (check to ensure interfaces don't have non-abstract instance methods).

So no, it's not possible to write a regression test for this in IL without upgrading the repo to use the .NET Core ILASM first.

akoeplinger:
> So no, it's not possible to write a regression test for this in IL without upgrading the repo to use the .NET Core ILASM first.

This should already be the case, we have code to download the ilasm package and use it here:
- https://github.com/mono/linker/blob/master/eng/ilasm.ilproj
- https://github.com/mono/linker/blob/master/test/Mono.Linker.Tests/TestCasesRunner/ILCompiler.cs#L70-L81

agocke:
> What nullability information do we need for a non-generic interface? That this is not null?

Yeah, that's confusing. This might be a bug. I'll follow up separately.

I don't think having a regression test is blocking, but we should have a bug tracking it.

MichalStrehovsky:
> This should already be the case, we have code to download the ilasm package and use it here:

Whoops, my bad. The error message from the test runner was unhelpful so I just assumed it's that since I compiled it manually with ILASM before and knew it worked. The error was a different problem.

image

I'll push the regression test shortly.

PR #1397, opened 2020-07-31
Add support for property tag on CA annotations

mateoatr:
Needed for #1386. This PR adds the ability to set properties when using custom attributes via a linker annotations XML and refactors the `ProcessAttributes` method which contained much of the logic needed by the `LinkAttributes` step.

/cc @layomia


src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -32,103 +32,150 @@ IEnumerable<CustomAttribute> ProcessAttributes (XPathNavigator nav, ICustomAttri
...
+			MethodDefinition constructor = attributeType.Methods.Where (method => method.IsInstanceConstructor ()).FirstOrDefault (c => c.Parameters.Count == attributeArguments.Length);
+			if (constructor == null) {
+				Context.LogWarning ($"Could not find a constructor for type '{attributeType}' that receives '{attributeArguments.Length}' arguments as parameter", 2022, _xmlDocumentLocation);
+				return null;
+			}
+
+			CustomAttribute customAttribute = new CustomAttribute (constructor);
+			var arguments = ProcessAttributeArguments (constructor, attributeArguments);
+			if (arguments == null)
+				return null;
```
tlakollo, line 70: Here the custom attribute still exists just we don't want to execute the arguments code section, so you should return the customAttribute instead of null

src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -32,103 +32,150 @@ IEnumerable<CustomAttribute> ProcessAttributes (XPathNavigator nav, ICustomAttri
...
+
+		void ProcessInternalAttribute (ICustomAttributeProvider provider, string internalAttribute)
+		{
+			if (internalAttribute != "RemoveAttributeInstances") {
+				Context.LogWarning ($"Unrecognized internal attribute '{internalAttribute}'", 2049, _xmlDocumentLocation);
+				return;
+			}
+
+			if (provider.MetadataToken.TokenType != TokenType.TypeDef || Annotations.IsMarked (provider)) {
+				Context.LogWarning ($"Internal attribute 'RemoveAttributeInstances' can only be used on a type, but is being used on '{provider}'", 2048, _xmlDocumentLocation);
```
tlakollo, line 147: This will warn when Annotations.IsMarked() returns true. This is not the intended behavior, the code before it will just continue without printing a message. This basically will warn every time someone suppresses the attribute deletion by adding it in a descriptor file.

PR #1413, opened 2020-08-05
Warn on case insensitive overload usage

### LINKED ISSUE #1322: Action when unsupported IgnoreCase is called for Type.GetType
There are two overloads of the Type.GetType method that receives a bool specifying a case insensitive search. This case is special since we cannot fallback to mark a default set of values. So there are several options to choose: 
-Limit the number of handled overloads in the intrinsic, by not recognizing the overloads with 3 and 5 arguments without even knowing if the boolean flag was set to true or false.
-Check the value of the boolean and then take a decision based on its value, I guess the decision would have to be to record as an unrecognized pattern. Otherwise, I would leave the code as is since the current behavior is that it will try to look up the type using the FullyQualifiedTypeName in a case sensitive way if it doesn't find it the code will intentionally ignore it and mark it as a HandledPattern.
This is a special case of #1297

tlakollo:
This fixes #1322


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -783,6 +783,15 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 				case IntrinsicId.Type_GetType: {
 						reflectionContext.AnalyzingPattern ();
 
+						var parameters = calledMethod.Parameters;
+						if ((parameters.Count == 3 && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
```
MichalStrehovsky, line 787: There are two overloads of `Type.GetType` with 3 parameters so this needs to be narrowed down.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -783,6 +783,15 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+									(parameters.Count == 5 && (methodParams[4].AsConstInt () == null || methodParams[4].AsConstInt () != 0))) {
+							_context.LogWarning ($"Reflection call '{calledMethod.GetDisplayName ()}' inside " +
+								$"'{((reflectionContext.Source is MethodDefinition method) ? method.GetDisplayName () : reflectionContext.Source.ToString ())}' " +
```
MichalStrehovsky, line 790:
```suggestion
								$"'{callingMethodDefinition.GetDisplayName ()}' " +
```

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -783,6 +783,15 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 				case IntrinsicId.Type_GetType: {
 						reflectionContext.AnalyzingPattern ();
 
+						var parameters = calledMethod.Parameters;
+						if ((parameters.Count == 3 && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
+									(parameters.Count == 5 && (methodParams[4].AsConstInt () == null || methodParams[4].AsConstInt () != 0))) {
+							_context.LogWarning ($"Reflection call '{calledMethod.GetDisplayName ()}' inside " +
```
MichalStrehovsky, line 789: This should use `reflectionContext.RecordUnrecognizedPattern` which does extra things.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -783,6 +783,15 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+								$"'{((reflectionContext.Source is MethodDefinition method) ? method.GetDisplayName () : reflectionContext.Source.ToString ())}' " +
+								$"is trying to make a case insensitive lookup which is not supported by the linker", 2055, callingMethodDefinition);
+							reflectionContext.RecordHandledPattern ();
```
MichalStrehovsky, line 792: Shouldn't be needed once we switch to RecordUnrecognizedPattern above.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -783,6 +783,13 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 				case IntrinsicId.Type_GetType: {
 						reflectionContext.AnalyzingPattern ();
 
+						var parameters = calledMethod.Parameters;
+						if ((parameters.Count == 3 && methodParams[2].Kind != ValueNodeKind.MergePoint && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
+							(parameters.Count == 5 && (methodParams[4].AsConstInt () == null || methodParams[4].AsConstInt () != 0))) {
+							reflectionContext.RecordUnrecognizedPattern ($"Reflection call '{calledMethod.GetDisplayName ()}' inside " +
+								$"'{callingMethodDefinition.GetDisplayName ()}' is trying to make a case insensitive lookup which is not supported by the linker");
```
vitek-karas, line 790:
```suggestion
							reflectionContext.RecordUnrecognizedPattern ($"Reflection call '{calledMethod.GetDisplayName ()}' inside " +
								$"'{callingMethodDefinition.GetDisplayName ()}' is trying to make a case insensitive lookup which is not supported by ILLink");
```

test/Mono.Linker.Tests.Cases/Reflection/TypeUsedViaReflection.cs
```diff
@@ -25,6 +29,14 @@ public static void Main ()
 			TestDeeplyNested ();
 			TestTypeOf ();
 			TestTypeFromBranch (3);
+			TestTypeUsingCaseInsensitiveFlag ();
+			TestTypeUsingCaseUnknownByTheLinker ();
+			/*
+			There is a test hole that involves overload Type.GetType(String, Func<AssemblyName,Assembly>, Func<Assembly,String,Boolean,Type>)
+			This couldn't be tested since Func<> create functions that cannot be marked with [Kept] and the test will end up generating an 
+			error message stating: Type `Mono.Linker.Tests.Cases.Reflection.TypeUsedViaReflection/<>c' should have been removed
+			The test case was validated manually and is commented in this file as function TestTypeOverloadWith3Parameters
```
vitek-karas, line 38: Will it work if instead of lambdas you use either static or local functions? Those can be annotated. Basically you want to avoid creating closures (lambdas do that), since it generates classes and code in the compiler. But you can create a Func over a normal method or local function (local functions also have closures, but it should be possible to add attributes to them).

test/Mono.Linker.Tests.Cases/Reflection/TypeUsedViaReflection.cs
```diff
@@ -1,9 +1,13 @@
 ﻿using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 using Mono.Linker.Tests.Cases.Expectations.Metadata;
 
 namespace Mono.Linker.Tests.Cases.Reflection
 {
+	[LogContains ("IL2006: Mono.Linker.Tests.Cases.Reflection.TypeUsedViaReflection.TestTypeUsingCaseInsensitiveFlag(): Reflection call 'System.Type.GetType(String,Boolean,Boolean)' " +
+		"inside 'Mono.Linker.Tests.Cases.Reflection.TypeUsedViaReflection.TestTypeUsingCaseInsensitiveFlag()' is trying to make a case insensitive lookup which is not supported by the linker")]
```
vitek-karas, line 10: Can you please merge with master and use the new `ExpectedWarning` attribute - you'll have to put it on the method which produces the warning though.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas, line 790: After you merge with master please create a new warning code for this, add it to the doc and so on.

docs/error-codes.md
```diff
@@ -1427,3 +1427,15 @@ This is technically possible if a custom assembly defines `DynamicDependencyAttr
     public override void TestMethod<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] T>() {}
   }
   ```
+
+#### `IL2096`: Trim analysis: Reflection call 'reflection call' inside 'calling method' is trying to make a case insensitive lookup which is not supported by ILLink"
```
vitek-karas, line 1431:
```suggestion
#### `IL2096`: Trim analysis: Call to 'Type.GetType method' can perform case insensitive lookup of the type, currently ILLink can not guarantee presence of all the matching types"
```

There's no need to include the calling method, warnings always have location (origin) which is the calling method (or even more precisely the instruction).

test/Mono.Linker.Tests.Cases/Reflection/TypeUsedViaReflection.cs
```diff
@@ -254,16 +253,40 @@ static bool GetCase ()
...
+		{
+			const string reflectionTypeKeptString = "mono.linker.tests.cases.reflection.TypeUsedViaReflection+CaseUnknown2, test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null";
+			var typeKept = Type.GetType (reflectionTypeKeptString, false, fieldHideCase);
+		}
+
 		[Kept]
 		public class OverloadWith3Parameters { }
 
 		[Kept]
 		static void TestTypeOverloadWith3Parameters ()
```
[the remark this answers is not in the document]
tlakollo: Initially, I only create the TestTypeOverloadWith3Parameters method to test both overloads with 3 parameters work. But sure I can add a test for all the others.

docs/error-codes.md
```diff
@@ -1428,14 +1428,14 @@ This is technically possible if a custom assembly defines `DynamicDependencyAttr
...
-#### `IL2096`: Trim analysis: Reflection call 'reflection call' inside 'calling method' is trying to make a case insensitive lookup which is not supported by ILLink"
+#### `IL2096`: Trim analysis: Call to 'Type.GetType method' can perform case insensitive lookup of the type, currently ILLink can not guarantee presence of all the matching types"
 
 - Specifying a case-insensitive search on an overload of `System.Type.GetType` is not supported by ILLink. Specify false to perform a case-sensitive search or use an overload that does not use a ignoreCase bolean.
 
   ``` C#
   void TestMethod()
   {
-      // IL2096 Trim analysis: Reflection call 'System.Type.GetType(String,Boolean,Boolean)' inside 'TestMethod()' is trying to make a case insensitive lookup which is not supported by ILLink
+      // IL2096 Trim analysis: Reflection call 'System.Type.GetType(String,Boolean,Boolean)' is trying to make a case insensitive lookup which is not supported by ILLink
```
vitek-karas, line 1438: Nit - also change this to match the warning message

test/Mono.Linker.Tests.Cases/Reflection/TypeUsedViaReflection.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
+using System.Security.AccessControl;
```
vitek-karas, line 4: Nit: this is weird, probably not needed

Issue #1406, opened 2020-08-03
Infer diagnostic severity using the message's code


mateoatr:
See [comment](https://github.com/mono/linker/pull/1372#discussion_r463761346). Since we use specific code ranges for the different severity levels that we use when logging linker warnings/errors, we should be able to infer the severity of the message being logged by looking at its code (if any) instead of specifying it in the call site by using `LogError`/`LogWarning`.

marek-safar:
What would be the benefit of doing that? I think it's useful to see in the code if the code path leads to error or warning. On top of that, we'd need to pass redundant arguments as these code paths usually have different content requirements.

vitek-karas:
This is now actually enforced in the code. For example calling `MessageContainer.CreateErrorMessage("text", "2010")` will throw since only 1xxx is the error range.

Issue #1384, opened 2020-07-23
Clarify warning API for custom steps


sbomer:
From some discussion with @vitek-karas in https://github.com/mono/linker/pull/1381:

The docs https://github.com/mono/linker/blob/master/docs/error-codes.md#illink-errors-and-warnings say:

> The known codes are in the range 1000 to 6000. Custom steps should avoid using this range not to collide with existing or future error and warning codes.

The ref assembly exposes `CreateWarningMessage` which has the following check:
```csharp
			if (!(code > 2000 && code <= 6000))
				throw new ArgumentException ($"The provided code '{code}' does not fall into the warning category, which is in the range of 2001 to 6000 (inclusive).");
```

Following the docs, we should either invert this check and use a different API for warnings created by the linker internally, or expose a different API for custom steps to use. Some other considerations brought up by @vitek-karas:

> I think we need sort of a double treatment here:
> 
> - For the warnings from the linker itself - I strongly feel that having the API only take code is the right approach
> - For custom steps we need the current API which allows everything to be passed in - since we can't build the code->message/version map into the linker
> 
> I'm working on a separate PR which will make sure that we only use one message per code (still no API refactoring though). So once that is done we can do the refactoring and create the map.

A quick proposal about the purpose of these APIs:
- Factor `LinkContext.LogWarning` to take only a code and message parameters. Only use it for warnings from the linker itself, and don't expose it in the ref assembly. Use an internal helper similar to `CreateWarningMessage` to validate that the code is in the range [2001-6000].
- Use `CreateWarningMessage` for custom steps. It checks that the code is > 6000.

Do we need a way for custom steps to emit known linker-defined warnings?

/cc @mateoatr @marek-safar

marek-safar:
> Do we need a way for custom steps to emit known linker-defined warnings?

No.

The simplest solution could be to just make the check less strict. Essentially changing the logic to something like

```
	if (code < 2000)
		throw new ArgumentException ($"The warning code '{code}' does not fall into the warning range");
```

For error handling, we could have different entry points and move the arguments checking there

marek-safar:
This has been resolved some time ago with the introduction of CreateCustom<kind>Message methods

PR #1417, opened 2020-08-06
NoWarn wins over WarnAsError

### LINKED ISSUE #1415: Nowarn should win over warnaserror
Warnings that have been silenced with `nowarn` should not be displayed, regardless of the `warnaserror` settings - even if warnaserror comes last. Currently the `warnaserror` setting will always win regardless of order, which means `nowarn` can't be used to silence warnings if the build has `TreatWarningsAsErrors` set globally.

This is the behavior I see from csc:

```csharp
using System;

namespace con
{
    class Program
    {
        {
            throw new Exception();
            Console.WriteLine("Hello World!");
        }
    }
}
```

```sh
#!/usr/bin/env bash

dotnet "/home/sven/bin/sdk/5.0.100-preview.8.20363.2/Roslyn/bincore/csc.dll" \
/reference:/home/sven/bin/packs/Microsoft.NETCore.App.Ref/5.0.0-preview.8.20361.2/ref/net5.0/System.Console.dll \
/reference:/home/sven/bin/packs/Microsoft.NETCore.App.Ref/5.0.0-preview.8.20361.2/ref/net5.0/System.Runtime.dll \
/out:obj/Debug/net5.0/linux-x64/con.dll /target:exe /nologo \
Program.cs "$@"
```
```
~/con$ ./compile.sh
Program.cs(10,13): warning CS0162: Unreachable code detected
~/con$ ./compile.sh /warnaserror
Program.cs(10,13): error CS0162: Unreachable code detected
~/con$ ./compile.sh /nowarn:CS0162 /warnaserror
~/con$
```

@mateoatr

mateoatr:
Fixes #1415.


src/linker/Linker/LinkContext.cs
```diff
@@ -502,10 +502,14 @@ public void LogMessage (MessageContainer message)
...
-				// This warning was turned off by --nowarn.
-				return;
+			if (message.Category == MessageCategory.Warning) {
+				if (NoWarn.Contains ((uint) message.Code))
+					// This warning was turned off by --nowarn.
+					return;
+
+				if ((GeneralWarnAsError && (!WarnAsError.TryGetValue ((uint) message.Code, out var warnAsError) || warnAsError)) ||
+					(!GeneralWarnAsError && (WarnAsError.TryGetValue ((uint) message.Code, out warnAsError) && warnAsError)))
+					message = MessageContainer.CreateErrorMessage (message.Text, message.Code.Value, message.SubCategory, message.Origin, isWarnAsError: true, message.Version);
```
sbomer, line 512: It would be nice to factor this so that the check for the code range (in Create*Message) happens just once when the warning/error is created, and not have to track isWarnAsError/version on the error.

sbomer: I don't want to block the PR on it though. Also related to https://github.com/mono/linker/issues/1384.

mateoatr: I agree we should factor this. This is also related to #1406.

sbomer:
@marek-safar do you mind if we take this before https://github.com/mono/linker/pull/1419? We'd like the behavior fix to flow to the SDK so that the temporary suppressions will work correctly with `TreatWarningsAsErrors`.

mateoatr:
#1419 already contains this changes. Closing.

PR #1419, opened 2020-08-07
Make warnings logging work with custom steps API

### LINKED ISSUE #1415: Nowarn should win over warnaserror

marek-safar:
As part of the change I also cleaned up uint values usage with only
made the code clumsy.

Also fixes #1415 and possibly other issues


src/linker/Linker/LinkContext.cs
```diff
@@ -174,9 +174,9 @@ public class LinkContext : IDisposable
 
 		public WarningSuppressionWriter WarningSuppressionWriter { get; private set; }
 
-		public HashSet<uint> NoWarn { get; set; }
+		public HashSet<int> NoWarn { get; set; }
 
-		public Dictionary<uint, bool> WarnAsError { get; set; }
+		public HashSet<int> WarnAsError { get; set; }
```
sbomer, line 179: Note that this also needs to be able to treat individual warnings not as errors to override the global warnaserror setting (https://github.com/mono/linker/issues/1356#issuecomment-658303720).

docs/error-codes.md
```diff
@@ -73,6 +73,10 @@ the error code. For example:
 
 - Invalid value 'value' was used for command-line option '--generate-warning-suppressions'; must be 'cs' or 'xml'.
 
+#### `IL1018`: "Invalid value 'value' was used as warning code
+
+- All warning codes must start with `IL` prefix followed by numeric code.
```
mateoatr, line 78: Should we mention the valid code range for warnings?

src/linker/Linker/MessageContainer.cs
```diff
@@ -74,14 +64,24 @@ public static MessageContainer CreateErrorMessage (string text, int code, string
...
 		/// warning wave option --warn VERSION. Unversioned warnings are unaffected by this option. </param>
 		/// <returns>New MessageContainer of 'Warning' category</returns>
-		public static MessageContainer CreateWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, string subcategory = MessageSubCategory.None, WarnVersion? version = null)
+		public static MessageContainer CreateWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, WarnVersion version, string subcategory = MessageSubCategory.None)
 		{
-			ValidateWarning (code, version);
+			if (!(code > 2000 && code <= 6000))
+				throw new ArgumentException ($"The provided code '{code}' does not fall into the warning category, which is in the range of 2001 to 6000 (inclusive).");
+
+			if (version >= WarnVersion.ILLink0 && version <= WarnVersion.Latest)
```
mateoatr, line 72:
```suggestion
			if (!(version >= WarnVersion.ILLink0 && version <= WarnVersion.Latest))
```

src/linker/Linker/LinkContext.cs
```diff
@@ -502,21 +502,6 @@ public void LogMessage (MessageContainer message)
...
-			// Unversioned warnings are not controlled by WarnVersion.
-			// Error messages are guaranteed to only have a version if they were created for a warning due to warnaserror.
-			if ((message.Category == MessageCategory.Warning || message.Category == MessageCategory.Error) &&
-				message.Version > WarnVersion) {
-				// This warning was turned off by --warn.
-				return;
-			}
-
 			if (OutputWarningSuppressions && message.Category == MessageCategory.Warning && message.Origin?.MemberDefinition != null)
 				WarningSuppressionWriter.AddWarning (message.Code.Value, message.Origin?.MemberDefinition);
```
mateoatr, line 506: We can also move this to CreateWarningMessage.

docs/error-codes.md
marek-safar: Not sure it's worth this, the whole file is about error/warning codes

src/linker/Linker/LinkContext.cs
marek-safar: I think that code is correct here. The warning should go to output file only when it's printed out not if it's constructed.

src/linker/Linker/LinkContext.cs
marek-safar: Yep, missed that

src/linker/Linker/LinkContext.cs
```diff
@@ -502,21 +502,6 @@ public void LogMessage (MessageContainer message)
...
-			// Note: message.Version is nullable. The comparison is false if it is null.
-			// Unversioned warnings are not controlled by WarnVersion.
-			// Error messages are guaranteed to only have a version if they were created for a warning due to warnaserror.
-			if ((message.Category == MessageCategory.Warning || message.Category == MessageCategory.Error) &&
-				message.Version > WarnVersion) {
-				// This warning was turned off by --warn.
-				return;
-			}
-
 			if (OutputWarningSuppressions && message.Category == MessageCategory.Warning && message.Origin?.MemberDefinition != null)
```
sbomer, line 505: I think this should also output suppressions for for `MessageCategory.WarningAsError`

src/linker/Linker/Driver.cs
```diff
@@ -775,7 +775,7 @@ public int Run (ILogger customLogger = null)
 
 		partial void PreProcessPipeline (Pipeline pipeline);
 
-		private static HashSet<uint> ParseWarnings (string value)
+		private IEnumerable<int> ProcessWarningsCodes (string value)
```
sbomer, line 778: nit:
```suggestion
		private IEnumerable<int> ProcessWarningCodes (string value)
```

src/linker/Linker/MessageContainer.cs
```diff
@@ -74,14 +64,24 @@ public static MessageContainer CreateErrorMessage (string text, int code, string
 		/// <param name="version">Optional warning version number. Versioned warnings can be controlled with the
 		/// warning wave option --warn VERSION. Unversioned warnings are unaffected by this option. </param>
 		/// <returns>New MessageContainer of 'Warning' category</returns>
-		public static MessageContainer CreateWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, string subcategory = MessageSubCategory.None, WarnVersion? version = null)
+		public static MessageContainer CreateWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, WarnVersion version, string subcategory = MessageSubCategory.None)
 		{
-			ValidateWarning (code, version);
+			if (!(code > 2000 && code <= 6000))
```
sbomer, line 69: Not sure if you wanted to fix it here, but this doesn't address https://github.com/mono/linker/issues/1384.

agocke:
@marek-safar Can we get this in today? We'd really like this in before the snap EOD

src/linker/Linker/MessageContainer.cs
vitek-karas: I'm choosing to specifically not address #1384 with this PR - it's related to the intent of these changes, but not really. We need this PR to fix the --nowarn --warnaserror problems, so focusing on that part.

sbomer:
LGTM thanks!

Issue #1420, opened 2020-08-10
Linker should warn when `DynamicallyAccessedMembersAttribute` is used on an item which is not type-like


vitek-karas:
We already warn about using the attribute on a method directly (as that is meant for special cases when we need to annotate the "this" parameter).

But we don't warn when the attribute is placed on the wrong thing - for example a property of type `System.Object`. In that case linker silently ignores the attribute.

vitek-karas:
@MichalStrehovsky what do you think?
I personally don't see a problem with adding the warning instead of silently ignoring it.

MichalStrehovsky:
Yup, that might be good. I think the only reason why we chose not to do anything about it is perf - so that we don't have to check for the attribute. The perf impact might be just in the noise range though.

vitek-karas:
We already check for the attribute as far as I can tell pretty much everywhere. The `FlowAnnotations.BuildTypeAnnotations` does that. And it has an explicit check if it's on an "interesting type" - currently we just skip the member if it's not interesting - the proposal is to swap the checks - look for attribute first and then do the interesting check - if there's an attribute but it's not interesting - log a warning.

MichalStrehovsky:
Yeah, the "is interesting" check was supposed to be a quick filter, until it turned out to be... not that quick because it has to walk base hierarchy instead of doing a simple name check. It might even be a perf improvement to swap the checks, as most things don't have any attributes.

PR #1431, opened 2020-08-14
Warn when DynamicallyAccessedMembersAttribute is used on a member of unsupported type

vitek-karas:
`DynamicallyAccessedMembersAttribute` is only supported on items of type `System.Type` or `System.String` (and derived classes) otherwise linker will ignore it. When it happens it's very likely not intentional and so linker should warn about it. Since it only affects analysis results, these are marked as TrimAnalysis.

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -1427,3 +1427,41 @@ This is technically possible if a custom assembly defines `DynamicDependencyAttr
     public override void TestMethod<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] T>() {}
   }
   ```
+
+#### 'IL2096': Trim analysis: Field 'field' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to fields of type 'System.Type' or 'System.String'
+
+- `DynamicallyAccessedMembersAttribute` is only applicable to items of type `System.Type` or `System.String` (or derived), on all other types the attribute will be ignored. Using the attribute on any other type is likely incorrect and unintentional.
+
+  ```C#
+  // IL2096: Field '_valueField' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to fields of type 'System.Type' or 'System.String'
+  [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
+  object _valueField;
+
+  // IL2096: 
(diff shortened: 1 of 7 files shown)


vitek-karas:
Fixes #1420

docs/error-codes.md
```diff
@@ -1438,4 +1438,36 @@ This is technically possible if a custom assembly defines `DynamicDependencyAttr
...
+
+- `DynamicallyAccessedMembersAttribute` is only applicable to items of type `System.Type` or `System.String` (or derived), on all other types the attribute will be ignored. Using the attribute on any other type is likely incorrect and unintentional.
+
+  ```C#
+  // IL2097: Field '_valueField' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to fields of type 'System.Type' or 'System.String'
+  [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
+  object _valueField;
+  ```
+
+#### 'IL2098': Trim analysis: Parameter 'parameter' of method 'method' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to fields of type 'System.Type' or 'System.String'
```
sbomer, line 1453:
```suggestion
#### 'IL2098': Trim analysis: Parameter 'parameter' of method 'method' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to parameters of type 'System.Type' or 'System.String'
```

docs/error-codes.md
```diff
@@ -1438,4 +1438,36 @@ This is technically possible if a custom assembly defines `DynamicDependencyAttr
...
+
+  ```C#
+  // IL2098: Parameter 'param' of method 'TestMethod' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to fields of type 'System.Type' or 'System.String'
+  void TestMethod([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] object param)
+  {
+    // param.GetType()....
+  }
+  ```
+
+#### 'IL2099': Trim analysis: Property 'property' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to fields of type 'System.Type' or 'System.String'
```
sbomer, line 1465:
```suggestion
#### 'IL2099': Trim analysis: Property 'property' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to properties of type 'System.Type' or 'System.String'
```

PR #1437, opened 2020-08-18
Don't error on non-IL warning codes

sbomer:
We pass NoWarn which may contain warning codes for other tools. This is causing failures in https://github.com/dotnet/sdk/pull/12948.

Adds a check that testcases don't produce unexpected errors, which validates this behavior together with this existing testcase: https://github.com/mono/linker/blob/2b04c065c84aa8f30600e3b1989c537fc77de95b/test/Mono.Linker.Tests.Cases/Warnings/CanDisableWarnings.cs#L12

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -113,10 +113,6 @@ the error code. For example:
 
 #### `IL1030`: Invalid argument for '{token}' option
 
-#### `IL1031`: "Invalid value 'value' was used as warning code
-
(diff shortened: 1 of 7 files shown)


sbomer:
Yeah, I missed it too. Which testcases did you mean?

vitek-karas:
https://github.com/mono/linker/commit/f9dfc8a29e9ae0d2388e6fcad8b7f823f9c09e9e#diff-da4616dd8c05b4b19d4124a0fb10e500

PR #1471, opened 2020-09-04
Update to target .NET5 framework

marek-safar:


Directory.Build.props
```diff
@@ -7,6 +7,8 @@
     <IsReferenceAssembly Condition="'$(IsReferenceAssembly)' == '' and '$([System.IO.Path]::GetFileName($(MSBuildProjectDirectory)))' == 'ref'">true</IsReferenceAssembly>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(IsReferenceAssembly)' == 'true' ">
+    <!-- Since .NET file reference assemblies are always produced -->
```
akoeplinger, line 10: file -> 5 ?

src/linker/Mono.Linker.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(MonoBuild)' == ''">
+    <TargetFramework>net5.0</TargetFramework>
```
sbomer, line 8: https://github.com/mono/linker/blob/master/src/linker/Directory.Build.props#L13 sets the `TargetFramework` for this and the ref project - either update it there, or remove it if you prefer to set it in each project.

src/ILLink.Tasks/ILLink.Tasks.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <!-- Keep these in sync with _ILLinkTasksTFM in Sdk.props. -->
```
sbomer, line 3: Sdk.props also needs to be updated so that MSBuild picks up the right version of the task.

marek-safar: Done

test/Mono.Linker.Tests.Cases/LinkAttributes/OverrideAttributeRemoval.xml
```diff
@@ -3,4 +3,7 @@
   <assembly fullname="Mono.Linker.Tests.Cases.Expectations">
```
vitek-karas, line 3: The one from test assembly should not be needed anymore.

marek-safar: It'll be required for mono test run

src/ILLink.Tasks/ILLink.Tasks.csproj
```diff
@@ -2,7 +2,7 @@
...
     <!-- Keep the netcoreapp TFM in sync with the Mono.Linker.csproj condition below. -->
-    <TargetFrameworks>netcoreapp3.0;net472</TargetFrameworks>
+    <TargetFrameworks>net5.0;net472</TargetFrameworks>
```
ViktorHofer, line 5: I recommend to use an msbuild property for the used .NETCoreApp version to keep changes as small as possible in the future.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -235,7 +235,7 @@ public class ILLink : ToolTask
 
 				var taskDirectory = Path.GetDirectoryName (Assembly.GetExecutingAssembly ().Location);
 				// The linker always runs on .NET Core, even when using desktop MSBuild to host ILLink.Tasks.
-				_illinkPath = Path.Combine (Path.GetDirectoryName (taskDirectory), "netcoreapp3.0", "illink.dll");
+				_illinkPath = Path.Combine (Path.GetDirectoryName (taskDirectory), "net5.0", "illink.dll");
```
ViktorHofer, line 238: The proposed msbuild property for the used .NETCoreApp version can also flow into this file by supplying the constant in the project file.

src/ILLink.Tasks/ILLink.Tasks.csproj
vitek-karas: Good idea - we should probably do that as a follow-up - the plan is to try to backport this to 5.0 release, so I would like to keep the changes to minimum.

ViktorHofer:
@nohwnd I spent some time trying to figure out why VSTest doesn't work when invoked via the VSTestTask, but without success. Can you please take a look? The SDK being used is 5.0 RC1 final. Here's the binlog: 
[msbuild.zip](https://github.com/mono/linker/files/5202181/msbuild.zip)

image

marek-safar:
@ViktorHofer I think the msbuild issue is a fallback from vstest not to be able to find the runtime which works fine with `dotnet build` so it looks to me like there is another issue happening before MSB4181.

ViktorHofer:
@nohwnd any idea what the cause of the actual issue could be? The error in the pipeline suggests that a matching runtime can't be found.

nohwnd:
@ViktorHofer ah sure, you don't have your PATH / DOTNET_ROOT set to the .dotnet directory where dotnet-install.ps1 installs your dotnet SDK, so when dotnet test tries to run it only finds the runtimes installed on the system, and there is no 5.0 runtime. I don't see why exactly, but adding the env variable manually should fix the issue.

ViktorHofer:
@alexperovich seems like Arcade needs to have set DOTNET_ROOT or override the PATH to point to the local .dotnet folder when invoking the VSTest target.

ViktorHofer:
OK this makes sense now. I see that the test project overrides the test target and routes it to the VSTest target: https://github.com/mono/linker/blob/395e24a5ccaf6c976feb4891756ee1f73a4672c9/test/Mono.Linker.Tests/Mono.Linker.Tests.csproj#L77-L81.

nohwnd:
For the PATH. The dotnet-install script should do that automatically, but in old versions the script itself was doing `exit 0` on success, so there are examples of using it in separate process, from which the PATH / DOTNET_ROOT won't propagate. `Obtain script and install .NET Core CLI one-liner examples` has that issue.  https://docs.microsoft.com/cs-cz/dotnet/core/tools/dotnet-install-script 

This would probably have also the same issue: 
https://github.com/dotnet/arcade/blob/4873d157a8f34f8cc7e28b3f9938b32c642ef542/src/Microsoft.DotNet.Arcade.Sdk/tools/InstallDotNetCore.targets#L8

ViktorHofer:
> ah sure, you don't have your PATH / DOTNET_ROOT set to the .dotnet directory where dotnet-install.ps1 installs your dotnet SDK, so when dotnet test tries to run it only finds the runtimes installed on the system, and there is no 5.0 runtime. I don't see why exactly, but adding the env variable manually should fix the issue.

@nohwnd looking at the latest binlog, I see that PATH contains the local .dotnet SDK still the testhost fails to find the right SDK/runtime. Still I don't see DOTNET_PAT being set. Should this work with the correct %PATH% or do we need DOTNET_PATH as well?

ViktorHofer:
@nohwnd also I see that the right dotnet invokes the test project:

```
D:\a\1\s\.dotnet\dotnet.exe test D:\a\1\s\artifacts\bin\ILLink.Tasks.Tests\Release\net5.0\ILLink.Tasks.Tests.dll --logger:"console;verbosity=normal" --logger:"trx;LogFileName=ILLink.Tasks.Tests_net5.0_x64.xml" "--ResultsDirectory:D:\a\1\s\artifacts\TestResults\Release" "--Framework:.NETCoreApp,Version=v5.0" > "D:\a\1\s\artifacts\log\Release\ILLink.Tasks.Tests_net5.0_x64.log" 2>&1
```

is it expected that PATH or DOTNET_PATH need to be set in such a case?

nohwnd:
@ViktorHofer Yes that is expected. I saw it as well that it runs via the correct dotnet, but when testhost.exe then starts it will try to resolve the dotnet host again and fails to find it, because the one that was used for dotnet test is not part of PATH or is not in DOTNET_ROOT.

ViktorHofer:
What surprises me here is that the "right" dotnet is actually in the path, at least the binlog shows that.

vitek-karas:
Unless `dotent test` propagates its location into `DOTNET_ROOT` the `testhost.exe` will not find the runtime. Apphost (which `testhost.exe` is an instance of) does not look at `PATH` - instead it uses `DOTNET_ROOT`.

nohwnd:
Ah okay, when running via dotnet test it will set DOTNET_ROOT to the directory of dotnet.exe, that is why testhost is able to start even if PATH is empty. 

But this won't happen when dotnet vstest.console.dll is executed. 

```shell
C:\t\UsePath> C:\d\dotnet.exe test
  Determining projects to restore...
  All projects are up-to-date for restore.
  You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
  UsePath -> C:\t\UsePath\bin\Debug\net5.0\UsePath.dll
Test run for C:\t\UsePath\bin\Debug\net5.0\UsePath.dll (.NETCoreApp,Version=v5.0)
Microsoft (R) Test Execution Command Line Tool Version 16.8.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:     1, Skipped:     0, Total:     1, Duration: 2 ms - UsePath.dll (net5.0)



C:\t\UsePath> C:\d\dotnet.exe "C:\d\sdk\5.0.100-rc.2.20465.1\vstest.console.dll"  C:\t\UsePath\bin\Debug\net5.0\UsePath.dll
Microsoft (R) Test Execution Command Line Tool Version 16.8.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
Testhost process exited with error: A fatal error occurred, the required library hostfxr.dll could not be found.
If this is a self-contained application, that library should exist in [C:\t\UsePath\bin\Debug\net5.0\].
If this is a framework-dependent application, install the runtime in the default location [C:\Program Files\dotnet] or use the DOTNET_ROOT environment variable to specify the runtime location.
. Please check the diagnostic logs for more information.

Test Run Aborted.
```

So yes what Vitek says, `DOTNET_ROOT.` needs to be set.

ViktorHofer:
That makes sense now. Thanks to all who help root cause this. The fix likely involves setting DOTNET_PATH in the vstest.targets file in Arcade as it doesn't use dotnet test (IMO it should). Will submit as PR

nohwnd:
@ViktorHofer do you mean DOTNET_ROOT? You said DOTNET_PATH.

ViktorHofer:
Yeah sorry. DOTNET_ROOT.

ViktorHofer:
Wait, Arcade's VSTest implementation is using `dotnet test`: https://github.com/dotnet/arcade/blob/5a630b52300710b806f22b34c079b2e6031d57c6/src/Microsoft.DotNet.Arcade.Sdk/tools/VSTest.targets#L22.

ViktorHofer:
@nohwnd based on your commend and [this code path](https://github.com/dotnet/sdk/blob/85cdd92940c7797a33c51a1b5cb61644ea2b332e/src/Cli/dotnet/commands/dotnet-test/Program.cs#L77-L81) I assumed that DOTNET_ROOT would be set when using `dotnet test`. Now I'm confused...

ViktorHofer:
OK, found the reason. This is a bug in the SDK when invoking `dotnet test`: https://github.com/dotnet/sdk/issues/13611

vitek-karas:
Thanks a lot @ViktorHofer for figuring this out!

ViktorHofer:
I will submit a quick fix in Arcade to unblock this after I submitted the product fix. Hopefully I'll have enough time for both today.

ViktorHofer:
Here's the product fix https://github.com/dotnet/sdk/pull/13619 and here's the Arcade fix that should unblock this PR: https://github.com/dotnet/arcade/pull/6182.

ViktorHofer:
This should be unblocked as soon as a new Arcade.SDK package is produced which is currently blocked by a 16.8 dependency on Arcade. I expect bits to be flowing later today.

ViktorHofer:
Nearly there. One more Arcade update needed tomorrow :)

vitek-karas (approved):

ViktorHofer (approved):

ViktorHofer:
et voila - this is now unblocked. @marek-safar feel free to merge.

Issue #1469, opened 2020-09-01
DynamicallyAccessedMembers with a string TypeName of a generic type doesn't appear to work


eerhardt:
Annotating a `string` property that contains a generic TypeName doesn't appear to work with the ILLinker.

This appears to be the root of the issue reported in https://github.com/dotnet/runtime/issues/39709. In `dotnet/runtime` we have a [TypeDescriptionProviderAttribute](https://github.com/dotnet/runtime/blob/master/src/libraries/System.ObjectModel/src/System/ComponentModel/TypeDescriptionProviderAttribute.cs), which can contain a `string` pointing to a Type. We annotated this attribute correctly to preserve the `PublicParameterlessConstructor`, since we are going to [instantiate the Type using reflection](https://github.com/dotnet/runtime/blob/aa5fdab9654d74bc6274c0b5d820272c8e859621/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/TypeDescriptor.cs#L330-L333).

However, one of the usages of this attribute on `XElement` and `XAttribute` points to a [generic Type](https://github.com/dotnet/runtime/blob/aa5fdab9654d74bc6274c0b5d820272c8e859621/src/libraries/System.Private.Xml.Linq/src/System/Xml/Linq/XElement.cs#L34):

```C#
    [System.ComponentModel.TypeDescriptionProvider("MS.Internal.Xml.Linq.ComponentModel.XTypeDescriptionProvider`1[[System.Xml.Linq.XElement, System.Xml.Linq, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]],System.ComponentModel.TypeConverter")]
    public class XElement : XContainer, IXmlSerializable
    {
```

This issue can be observed with the following stand-alone application. Run the following in an untrimmed application, and it will print 

```
ConsoleApp1.TypeDescriptionProvider
ConsoleApp1.MyGenericTypeProvider`1[ConsoleApp1.MyTypeWithGeneric]
```

Then Publish with PublishTrimmed=true and TrimMode=link, and adding the application assembly as a `TrimMode=link` assembly. Run the published app and you will see:

```
ConsoleApp1.TypeDescriptionProvider
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'type')
   at System.Activator.CreateInstance(Type type, Boolean nonPublic, Boolean wrapExceptions)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
[...]

Issue #1387, opened 2020-07-24
FindType implementation is inadequate


MichalStrehovsky:
.NET Native analyzer had a type name parser extracted out of the runtime reflection stack so that we don't end up in a situation where static analysis is not capable to parse a string that runtime can parse.

E.g. we probably fail hard for generic instances (`Type.GetType("Foo[Bar]")`) should make sure both `Foo` and `Bar` are kept).

PR #1472, opened 2020-09-05
Port .NET Native type name parser

mateoatr:
See #1387.

This adds the type name parsing logic used by CoreRT in reflection scenarios. With this we can successfully parse generic types, fixing scenarios such as #1469.

For reviewing I'd suggest paying particular attention to `ReflectionTypeNameResolver` in TypeParsing.cs; all other new files were ported from CoreRT with insignificant modifications.


src/linker/Linker/TypeParsing.cs
```diff
@@ -0,0 +1,77 @@
...
+					var foundType = ResolveTypeName (assemblyDefinition, assemblyQualifiedTypeName.TypeName);
+					if (foundType == null)
+						continue;
+
+					return foundType;
+				}
+
+				return null;
+			} else if (typeName is NonQualifiedTypeName) {
+				return ResolveTypeName (null, (NonQualifiedTypeName) typeName);
```
vitek-karas, line 52: This explicitly calls ResolveTypeName with assembly null, but the method below will NullRef on null assembly.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1548,25 +1547,15 @@ protected virtual void DoAdditionalInstantiatedTypeProcessing (TypeDefinition ty
...
-					targetTypeReference = (TypeReference) property.Argument.Value;
-					break;
-				}
+				if (property.Name == "Target")
+					return ((TypeReference) property.Argument.Value)?.Resolve ();
 
-				if (property.Name == "TargetTypeName") {
-					if (TypeNameParser.TryParseTypeAssemblyQualifiedName ((string) property.Argument.Value, out string typeName, out string assemblyName)) {
-						if (string.IsNullOrEmpty (assemblyName))
-							targetTypeReference = asm.MainModule.GetType (typeName);
```
vitek-karas, line 1561: If the type name is not assembly qualified, this will resolve it against the current assembly.
But the new code will search all assemblies and return the first hit. For things like Nullable attributes which are added into each assembly - the behavior will suddenly be very different.

test/Mono.Linker.Tests/Tests/TypeNameParserTests.cs
```diff
@@ -1,125 +0,0 @@
-﻿using NUnit.Framework;
-
-namespace Mono.Linker.Tests
-{
-	[TestFixture]
-	public class TypeNameParserTests
```
vitek-karas, line 6: We should try to adapt these test cases - even if the code is very different the tests themselves seem to be valid.

vitek-karas (changes requested):
We should try hard to avoid changing existing behavior. I didn't dig deep enough to tell how much of a difference there is between the old and new algorithms, but at least the visible differences around assembly resolution should be removed.
Also - we should take this opportunity to add more tests - the fact that none of the assembly differences caused test failures means that we're missing coverage.
Also - given the bug which started all this - we should have tests for that bug and similar cases as well.

src/linker/Linker/AssemblyNameHelpers.cs
```diff
@@ -0,0 +1,160 @@
...
+				throw new FileLoadException ();
+
+			StringBuilder sb = new StringBuilder ();
+			if (a.Name != null) {
+				sb.AppendQuoted (a.Name);
+			}
+
+			if (a.Version != null) {
+				Version canonicalizedVersion = a.Version.CanonicalizeVersion ();
+				if (canonicalizedVersion.Major != ushort.MaxValue) {
```
MichalStrehovsky, line 29: Did you make any changes to these files that we should review?

@marek-safar The fact that this repo uses formatting convention not used by the .NET teams makes source sharing frustrating.

src/linker/Linker/TypeParsing.cs
```diff
@@ -0,0 +1,77 @@
...
+
+			// This is unreachable
+			throw new NotImplementedException ();
+		}
+
+		public TypeDefinition ResolveTypeName (AssemblyDefinition assembly, NonQualifiedTypeName typeName)
+		{
+			if (typeName is ConstructedGenericTypeName) {
+				ConstructedGenericTypeName genericTypeName = (ConstructedGenericTypeName) typeName;
+				return assembly.MainModule.GetType (genericTypeName.GenericType.ToString ());
```
MichalStrehovsky, line 63: Will this lose the generic arguments?

One of the things #1387 calls out is  `Type.GetType("Foo[Bar]"))` should make sure both Foo and Bar are kept. It doesn't look like that would happen here.

src/linker/Linker/AssemblyNameHelpers.cs
```diff
@@ -0,0 +1,160 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+// Copyright (c) Microsoft Corporation. All rights reserved.
+
+using System.IO;
+using System.Text;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace System.Reflection.Runtime.Assemblies
```
marek-safar, line 11: The namespace does not match the file path (and tbh it has quite odd name)

src/linker/Linker/TypeParsing.cs
```diff
@@ -0,0 +1,77 @@
...
+			return ResolveTypeName (parsedTypeName);
+		}
+
+		private TypeDefinition ResolveTypeName (TypeName typeName)
+		{
+			if (typeName is AssemblyQualifiedTypeName) {
+				AssemblyQualifiedTypeName assemblyQualifiedTypeName = (AssemblyQualifiedTypeName) typeName;
+				RuntimeAssemblyName assemblyName = assemblyQualifiedTypeName.AssemblyName;
+				foreach (var assemblyDefinition in _context.GetAssemblies ()) {
+					if (assemblyName != null && assemblyDefinition.Name.Name != assemblyName.Name)
```
marek-safar, line 40: Why is `if (assemblyName != null` inside the loop?

src/linker/Linker/TypeParsing.cs
```diff
@@ -0,0 +1,77 @@
...
+
+			return ResolveTypeName (parsedTypeName);
+		}
+
+		private TypeDefinition ResolveTypeName (TypeName typeName)
+		{
+			if (typeName is AssemblyQualifiedTypeName) {
+				AssemblyQualifiedTypeName assemblyQualifiedTypeName = (AssemblyQualifiedTypeName) typeName;
+				RuntimeAssemblyName assemblyName = assemblyQualifiedTypeName.AssemblyName;
+				foreach (var assemblyDefinition in _context.GetAssemblies ()) {
```
marek-safar, line 39: This should use `_context.GetLoadedAssembly (assemblyName.Name)`

src/linker/Linker/TypeParsing.cs
```diff
@@ -0,0 +1,77 @@
...
+					if (assemblyName != null && assemblyDefinition.Name.Name != assemblyName.Name)
+						continue;
+
+					var foundType = ResolveTypeName (assemblyDefinition, assemblyQualifiedTypeName.TypeName);
+					if (foundType == null)
+						continue;
```
marek-safar, line 45: I don't think `continue` will ever succeed

src/linker/Linker/TypeParsing.cs
```diff
@@ -0,0 +1,77 @@
...
+			} catch (System.IO.FileLoadException) {
+				return null;
+			}
+
+			return ResolveTypeName (parsedTypeName);
+		}
+
+		private TypeDefinition ResolveTypeName (TypeName typeName)
+		{
+			if (typeName is AssemblyQualifiedTypeName) {
```
marek-safar, line 36:
```suggestion
			if (typeName is AssemblyQualifiedTypeName typeName) {
```

marek-safar (changes requested):

src/linker/Linker/AssemblyNameHelpers.cs
marek-safar: I agree and this code should not land under `src/linker/` but under `external/<something>` where it can use any formatting convention and it's clear where it's coming from. I don't know where this code was copied from but it should also copy tests as I'm not comfortable landing hundreds of lines of a parser code without test coverage.

MichalStrehovsky: This is originally extracted from the runtime reflection stack. The tests for that are the Type.GetType tests in the runtime repo - converting them to something workable is probably a non-trivial amount of work. If we need to service this, the fix should go through the runtime.

Given this code shipped in several major versions of a .NET runtime and the .NET Native compiler, I have pretty high confidence in it even with light (integration-level) test coverage here.

src/linker/Linker/AssemblyNameHelpers.cs
mateoatr: Should I change all the places where this namespace is used to `Mono.Linker`?

src/linker/Linker/AssemblyNameHelpers.cs
mateoatr: @MichalStrehovsky 
 > Did you make any changes to these files that we should review?

Not really. Probably the most interesting change is this - how we parse the assembly version, which is a bit different from how we originally did it for .NET Native, however the code here is how we currently do it in CoreRT: https://github.com/dotnet/corert/blob/9d8bd29a71f68941aaa2c00e9c9a0eed86cafa10/src/System.Private.CoreLib/shared/System/Reflection/AssemblyNameFormatter.cs#L26-L52
That, and the fact that the `ToString` of `NestedTypeName` returns `DeclaringType + "/" Name` instead of `DeclaringType + "+" Name`, because of how Cecil encodes things.

test/Mono.Linker.Tests/Tests/TypeNameParserTests.cs
mateoatr: I adapted these.

src/linker/Linker.Steps/MarkStep.cs
mateoatr: @vitek-karas I made changes to this so that we honor the old (current) behavior.

@MichalStrehovsky I can change this back if it's too risky. I felt comfortable changing this since it seems that any type name that returns true when passed to `TryParseTypeAssemblyQualifiedName` should be correctly parsed by the new type name parser.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1335,7 +1335,8 @@ void ProcessCreateInstanceByName (ref ReflectionPatternContext reflectionContext
 								continue;
 							}
 
-							var resolvedType = resolvedAssembly.FindType (typeNameStringValue.Contents);
+							var typeName = System.Reflection.Runtime.TypeParsing.TypeParser.ParseAssemblyQualifiedTypeName (typeNameStringValue.Contents).TypeName;
```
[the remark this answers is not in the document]
mateoatr: I did not. I changed this to be able to remove `FindType`.

src/linker/Linker/AssemblyNameHelpers.cs
MichalStrehovsky: No - move the files that come from the runtime to the external directory like Marek suggests below (something like external\TypeParsing\System\Reflection\Runtime\Assemblies) and add a README.md to the root of TypeParsing that tells the source and commit ID that was used as the base. That way we'll have future reference.

Also undo the formatting changes on the files that come from outside so that this is easier to diff, should we have a need.

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -0,0 +1,71 @@
...
+			if (assembly == null)
+				return null;
+
+			if (typeName is AssemblyQualifiedTypeName assemblyQualifiedTypeName) {
+				return ResolveTypeName (assembly, assemblyQualifiedTypeName.TypeName);
+			} else if (typeName is ConstructedGenericTypeName constructedGenericTypeName) {
+				var genericTypeDef = ResolveTypeName (assembly, constructedGenericTypeName.GenericType)?.Resolve ();
+				var genericInstanceType = new GenericInstanceType (genericTypeDef);
+				foreach (var arg in constructedGenericTypeName.GenericArguments) {
+					genericInstanceType.GenericArguments.Add (ResolveTypeName (assembly, arg));
```
MichalStrehovsky, line 60: If ResolveTypeName returns null, this will create a generic instance over null. That will probably crash things.

(Also for HasElementTypeName below.)

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -797,13 +797,14 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
 							if (typeNameValue is KnownStringValue knownStringValue) {
-								TypeDefinition foundType = AssemblyUtilities.ResolveFullyQualifiedTypeName (_context, knownStringValue.Contents);
+								TypeReference foundType = _context.TypeNameResolver.ResolveTypeName (knownStringValue.Contents);
 								if (foundType == null) {
 									// Intentionally ignore - it's not wrong for code to call Type.GetType on non-existing name, the code might expect null/exception back.
 									reflectionContext.RecordHandledPattern ();
 								} else {
-									reflectionContext.RecordRecognizedPattern (foundType, () => _markStep.MarkTypeVisibleToReflection (foundType, new DependencyInfo (DependencyKind.AccessedViaReflection, callingMethodDefinition), callingMethodDefinition));
-									methodReturnValue = MergePointValue.MergeValues (methodReturnValue, new SystemTypeValue (foundType));
+									var typeDef = foundType?.Resolve ();
```
vitek-karas, line 805: This should be moved above the if and we should check that Resolve didn't return null. It's "easy" to create a type ref in Cecil - but there's no guarantee that it will actually resolve. If the Resolve fails here, we will likely nullref below.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1344,7 +1345,7 @@ void ProcessCreateInstanceByName (ref ReflectionPatternContext reflectionContext
 								continue;
 							}
 
-							MarkConstructorsOnType (ref reflectionContext, resolvedType,
+							MarkConstructorsOnType (ref reflectionContext, resolvedType?.Resolve (),
```
vitek-karas, line 1348: Same here - Resolve may return null - which needs to be handled.

external/TypeParsing/README.md
```diff
@@ -0,0 +1 @@
+The code in this folder was adapted from dotnet/corert commit c8bfca5f4554badfb89b80d2319769f83512bf62
```
marek-safar, line 1: The folder name should start with `external/corert/`

external/TypeParsing/System/Reflection/AssemblyNameFormatter.cs
```diff
@@ -0,0 +1,156 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
marek-safar, line 1: It might be better to match the file path as well (e.g. `src/System.Private.CoreLib/shared/System/Reflection/AssemblyNameFormatter.cs` in this case)

src/linker/Linker/ModuleDefinitionExtensions.cs
```diff
@@ -0,0 +1,14 @@
+using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	public static class ModuleDefinitionExtensions
+	{
+
+		public static bool IsCrossgened (this ModuleDefinition module)
```
marek-safar, line 8: Why is it needed?

mateoatr: This used to be in `AssemblyResolver`, I simply splitted that file to fit `<Type>Extensions.cs`. Looks like this is used for writing back pure IL for crossgen'd assemblies: https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/OutputStep.cs#L117-L118

vitek-karas (approved):

marek-safar (approved):

PR #1475, opened 2020-09-08
Speed up tests execution

marek-safar:
This changes reduced total tests execution time under netcore from about
15 minutes on local or CI machine to about 3 minutes.

### DIFF
--- a/src/linker/Linker.Steps/DynamicDependencyLookupStep.cs
+++ b/src/linker/Linker.Steps/DynamicDependencyLookupStep.cs
@@ -15,6 +15,9 @@ namespace Mono.Linker.Steps
 	{
 		protected override void ProcessAssembly (AssemblyDefinition assembly)
 		{
+			if (Annotations.GetAction (assembly) == AssemblyAction.Skip)
+				return;
+
 			var module = assembly.MainModule;
 
 			foreach (var type in module.Types) {
--- a/test/Mono.Linker.Tests/TestCasesRunner/LinkerArgumentBuilder.cs
+++ b/test/Mono.Linker.Tests/TestCasesRunner/LinkerArgumentBuilder.cs
@@ -184,6 +184,8 @@ namespace Mono.Linker.Tests.TestCasesRunner
 		{
 			if (options.CoreAssembliesAction != null)
 				AddCoreLink (options.CoreAssembliesAction);
+			else
+				AddCoreLink ("skip");
 
 			if (options.UserAssembliesAction != null)
 				AddUserLink (options.UserAssembliesAction);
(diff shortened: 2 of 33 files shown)


test/Mono.Linker.Tests/TestCasesRunner/LinkerArgumentBuilder.cs
```diff
@@ -184,6 +184,8 @@ public virtual void ProcessOptions (TestCaseLinkerOptions options)
 		{
 			if (options.CoreAssembliesAction != null)
 				AddCoreLink (options.CoreAssembliesAction);
+			else
+				AddCoreLink ("skip");
```
vitek-karas, line 188: This feels a bit off - how would we test the default behavior of the linker (That is omit the parameter)?
Not saying this it needs to be easy, but it feels wrong that we obviously don't test that behavior.

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
```diff
@@ -169,6 +169,19 @@ protected static NPath MakeSupportingAssemblyReferencePathAbsolute (NPath output
 			if (Path.IsPathRooted (referenceFileName))
 				return referenceFileName.ToNPath ();
 
+			if (referenceFileName.StartsWith ("System.", StringComparison.Ordinal) ||
+				referenceFileName.StartsWith ("Mono.", StringComparison.Ordinal) ||
+				referenceFileName.StartsWith ("Microsoft.", StringComparison.Ordinal) ||
+				referenceFileName == "netstandard.dll") {
+#if NETCOREAPP
```
vitek-karas, line 176: Put the entire if in the ifdef.

vitek-karas (approved):
Mostly minor comments - looks good!

test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompiler.cs
marek-safar: Changed

test/Mono.Linker.Tests/TestCasesRunner/LinkerArgumentBuilder.cs
marek-safar: I think such a test should be part of the integration like tests (basically different setup). This setup is primary about IL/metadata logic not about default behaviour.

src/linker/Linker.Steps/DynamicDependencyLookupStep.cs
```diff
@@ -15,6 +15,9 @@ public class DynamicDependencyLookupStep : LoadReferencesStep
 	{
 		protected override void ProcessAssembly (AssemblyDefinition assembly)
 		{
+			if (Annotations.GetAction (assembly) == AssemblyAction.Skip)
```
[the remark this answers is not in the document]
marek-safar: I see that more as a bug fix (we should not process assembly if it's explicitly skipped). We could backport that to 5 where it was introduced if you think it's significant.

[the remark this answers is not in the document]
vitek-karas: I don't think it's significant enough to meet the bar for 5.0 at this time.
I agree that the new behavior is better - my question was more about our thinking about intentional behavioral changes (like this one). In short - similar change in runtime would be discussed in a more detail to determine potential impact of breaking something.

PR #1476, opened 2020-09-08
Update to recent nunit test adapter and clean up unused dependencies

marek-safar:
Depends on #1475


vitek-karas (approved):
Looks good - we should validate that it works with VS - that is, it's still possible to run tests from within VS itself.

marek-safar:
I don't have VS setup available

vitek-karas:
OK - I'll try it

vitek-karas:
Verified that it works with VS.

PR #1477, opened 2020-09-08
[release/5.0] Build against Cecil submodule

sbomer:
Doing this will change a few things:

- Our package will no longer include Mono.Cecil.Mdb or Mono.Cecil.Rocks. (These all included with the official package, but we don't depend on them when building against the individual projects. We might want to filter them out even when we go back to building against the package).
- The ref package produced from `release/5.0` will now have a dependency on the local cecil version (it gets a `5.0.0-rc*` version from our Versions.props). To consume it, the consuming project needs to include a version of Mono.Cecil compatible with this version, which in practice probably means also building against the submodule and giving it a custom version. @marek-safar I assume that this is OK since we will go back to shipping with the official package when use the ref assembly for plugins in 6. The alternative is to build the ref and lib against different versions of Cecil.

### DIFF
--- a/external/Mono.Cecil.overrides
+++ b/external/Mono.Cecil.overrides
@@ -8,6 +8,11 @@
     <!-- Cecil sets PublicSign on windows, but we always want to
          PublicSign. -->
     <PublicSign>true</PublicSign>
+    <!-- When we build the cecil submodule, use the same version number
+         as the package dependency. This way the linker ref assembly package
+         will depend on the publicly available package. Note that we can't set
+         just set Version here because Arcade will override it. -->
+    <PackageVersion>$(MonoCecilVersion)</PackageVersion>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(MonoBuild)' == ''">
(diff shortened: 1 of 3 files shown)


vitek-karas:
We discussed the ref issue and for 5.0 it should not matter - we're not changing public APIs in Cecil in the fork, so consumers can still refer the official Cecil package and compile against that. At runtime the forked Cecil will be used, but since public APIs are identical everything should work just fine.

sbomer:
The ref package will now depend on a package version that matches the official cecil, as discussed with @vitek-karas.

vitek-karas (approved):

external/Mono.Cecil.overrides
```diff
@@ -8,6 +8,11 @@
     <!-- Cecil sets PublicSign on windows, but we always want to
          PublicSign. -->
     <PublicSign>true</PublicSign>
+    <!-- When we build the cecil submodule, use the same version number
+         as the package dependency. This way the linker ref assembly package
+         will depend on the publicly available package. Note that we can't set
+         just set Version here because Arcade will override it. -->
+    <PackageVersion>$(MonoCecilVersion)</PackageVersion>
```
marek-safar, line 15: It'd be better to stop the build when the versions don't match but it does not look like it's doable with current Cecil setting

marek-safar (approved):

PR #1480, opened 2020-09-09
Remove reference to System.Text.Json

vitek-karas:
Linker doesn't need it and it causes us to carry 3 more files in the SDK.

### DIFF
--- a/src/linker/Mono.Linker.csproj
+++ b/src/linker/Mono.Linker.csproj
@@ -35,7 +35,6 @@
 
   <ItemGroup>
     <PackageReference Condition="'$(UseCecilPackage)' == 'true'" Include="Mono.Cecil" Version="$(MonoCecilVersion)" />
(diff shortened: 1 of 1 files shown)


vitek-karas:
I think we should consider taking this for 5.0 - linker in 5.0 carries these files:
* System.Text.Json.dll (version 4.0.1.1)
* System.Text.Encodings.Web.dll (version 4.0.5.0)
* System.Runtime.CompilerServices.Unsafe.dll (4.0.6.0)

Not only they're not needed, but they're old versions - all of these are now part of the 5.0 framework, so even if they would be needed, linker would not use them anyway (it would pick up the newer versions from the framework).

sbomer (approved):
Thanks!

marek-safar (approved):

marek-safar:
Please create backport PR to target net5

PR #1482, opened 2020-09-10
[release/5.0] Remove reference to System.Text.Json (#1480)

vitek-karas:
Linker doesn't need it and it causes us to carry 3 more files in the SDK.

This is the 5.0 port of #1480


marek-safar:
@monojenkins build

sbomer (approved):

marek-safar (approved):

PR #1490, opened 2020-09-12
Prevent multiple allocations of ReaderParameters

mateoatr:
`ReaderParams` gets allocated multiple times inside `DirectoryAssemblyResolver`, accounting for about 4.6% of the memory allocated in the process when linking a hello world app. Preventing this many allocations can help us improve the performance of the linker just a little bit by easing up the GC work.

image

Comparison table (linking a hello world app)

|                      | Before   | After    |
|----------------------|----------|----------|
| Total Allocs            | 324MB | 305MB |
| Total CPU Time       | 20,642ms | 19,672ms |
| Total GC CPU Time    | 4,546ms  | 4,235ms  |
| Total GC Pause       | 4,795ms  | 4,468ms  |
| % Time paused for GC | 24%      | 23.2%    |

/cc @agocke

### DIFF
--- a/src/linker/Linker/DirectoryAssemblyResolver.cs
+++ b/src/linker/Linker/DirectoryAssemblyResolver.cs
@@ -20,23 +20,16 @@ namespace Mono.Linker {
 
 		readonly List<MemoryMappedViewStream> viewStreams = new List<MemoryMappedViewStream> ();
 
+		readonly ReaderParameters defaultReaderParameters;
+
 		public void AddSearchDirectory (string directory)
 		{
 			directories.Add (directory);
 		}
 
-		public void RemoveSearchDirectory (string directory)
-		{
-			directories.Remove (directory);
-		}
-
-		public string [] GetSearchDirectories ()
-		{
-			return this.directories.ToArray ();
-		}
-
 		protected DirectoryAssemblyResolver ()
 		{
+			defaultReaderParameters = new ReaderParameters ();
 			directories = new Collection<string> (2) { "." };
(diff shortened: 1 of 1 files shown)


src/linker/Linker/DirectoryAssemblyResolver.cs
```diff
@@ -71,7 +64,8 @@ protected AssemblyDefinition GetAssembly (string file, ReaderParameters paramete
 
 		public virtual AssemblyDefinition Resolve (AssemblyNameReference name)
 		{
-			return Resolve (name, new ReaderParameters ());
+			defaultReaderParameters.AssemblyResolver = this;
```
vitek-karas, line 67: Why not assign the resolver in the constructor?

vitek-karas (approved):

MichalStrehovsky (approved):
Looks good modulo Vitek's question.

If you're looking into GC performance, turning on server GC might work as well. .NET Native had a similar IL processing stage and we ran it with server GC turned on because it provided superior performance.

Adding a `<ServerGarbageCollection>true</ServerGarbageCollection>` property to the main EXE project should be [all that's needed](https://docs.microsoft.com/en-us/dotnet/core/run-time-config/garbage-collector) (check PerfView GC stats - it should tell what mode the GC runs in).

src/linker/Linker/DirectoryAssemblyResolver.cs
mateoatr: Right 😄.

PR #1505, opened 2020-09-23
Enable some code analyzers and make the build pass

marek-safar:


vitek-karas (approved):

external/corert/src/System.Private.CoreLib/shared/System/Reflection/AssemblyNameFormatter.cs
```diff
@@ -64,7 +64,7 @@ public static string ComputeDisplayName(RuntimeAssemblyName a)
 			if (pkt != null)
 			{
 				if (pkt.Length > PUBLIC_KEY_TOKEN_LEN)
-					throw new ArgumentException();
+					throw new ArgumentException("Invalid token length", nameof(a));
```
MichalStrehovsky, line 67: I thought the `external` directory is not subject to the style rules - it makes this code harder to share with a "death by a thousand insignificant changes" - even though I admit some of this code is dumb.

marek-safar: For formatting changes yes but there is no way to exclude single file or folder for roslyn analyzers so I went ahead and fixed the issues this way (it should be all one lines so not that painful to compare)

Directory.Build.props
```diff
@@ -19,7 +19,8 @@
 
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" Condition="'$(MonoBuild)' == ''" />
   <!-- Still use global versions even if not using arcade. -->
-  <Import Project="eng/Versions.props" Condition="'$(MonoBuild)' != ''" />
+  <Import Project="$(RepositoryEngineeringDir)Versions.props" Condition="'$(MonoBuild)' != ''" />
+  <Import Project="$(RepositoryEngineeringDir)Analyzers.props" />
```
sbomer, line 23: The mono build is failing because RepositoryEngineeringDir isn't defined (since it doesn't build with Arcade).

eng/CodeAnalysis.ruleset
```diff
@@ -0,0 +1,472 @@
+<RuleSet Name="Microsoft.Analyzers.ManagedCodeAnalysis" Description="Microsoft.Analyzers.ManagedCodeAnalysis" ToolsVersion="14.0">
```
sbomer, line 1: Could we use an editorconfig file instead? It [sounds like](https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#convert-an-existing-ruleset-file-to-editorconfig-file) rulesets are deprecated and editorconfig is preferred going forward.

sbomer (approved):

external/corert/src/System.Private.CoreLib/shared/System/Reflection/AssemblyNameFormatter.cs
akoeplinger: or upstream these changes to corert :)

Directory.Build.props
marek-safar: Yep, fixed

eng/CodeAnalysis.ruleset
marek-safar: Good suggestion! I moved the rules to editorconfig

Issue #1508, opened 2020-09-25
Linker fails during publishing of the project


Denterl:
Hello,

Executing the line: `dotnet publish -c Release --self-contained=true` for version `5.0.100-rc.1.20452.10`
With the following settings in proj file:
```
<PublishTrimmed>true</PublishTrimmed>
<RuntimeIdentifier>win-x64</RuntimeIdentifier>
```

Returns the error: 
```
ILLink : error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues 
  Fatal error in IL Linker
  Unhandled exception. System.Collections.Generic.KeyNotFoundException: The given key '2' was not present in the dictionary.
     at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
     at Mono.Linker.Steps.RemoveUnreachableBlocksStep.ConstantExpressionMethodAnalyzer.GetLocalsValue(Int32 index, MethodBody body)
     at Mono.Linker.Steps.RemoveUnreachableBlocksStep.ConstantExpressionMethodAnalyzer.Analyze()
     at Mono.Linker.Steps.RemoveUnreachableBlocksStep.FindConstantExpressionsMethods(Collection`1 types)
     at Mono.Linker.Steps.RemoveUnreachableBlocksStep.FindConstantExpressionsMethods(Collection`1 types)
     at Mono.Linker.Steps.RemoveUnreachableBlocksStep.Process()
     at Mono.Linker.Steps.BaseStep.Process(LinkContext context)
     at Mono.Linker.Pipeline.ProcessStep(LinkContext context, IStep step)
     at Mono.Linker.Pipeline.Process(LinkContext context)
     at Mono.Linker.Driver.Run(ILogger customLogger)
     at Mono.Linker.Driver.Main(String[] args)
```

vitek-karas:
@mateoatr are we confident that #1511 fixes this problem? If so - we should close this issue.

mateoatr:
Yes, I think so. Closing.

marek-safar:
I think this one might be worth backporting to 5.0

pintset:
@mateoatr , @vitek-karas . The issue was reproduced on RC2: 5.0.100-rc.2.20479.15
Thought that fix will be included in the next RC.

daz10000:
Yes, can confirm problem is there in `5.0.100-rc.2.20479.15` as well for my build.  Same error message.

vitek-karas:
[...]

PR #1511, opened 2020-09-25
Use TryGetValue when retrieving local variables from stack

mateoatr:
Should avoid #1508 from happening. I suspect that the only way in which the key wouldn't exist here is by linking unverifiable CIL. From the ECMA: 

> For verifiable code, this instruction [ldloc] shall guarantee that it is not loading an uninitialized value –
> whether that initialization is done explicitly by having set the localsinit bit for the method, or by
> previous instructions (where the CLI performs definite-assignment analysis).

/cc @agocke


marek-safar:
Please add test case for the change

marek-safar (approved):

PR #1513, opened 2020-09-26
[5.0] Process also exception filters when marking unreachable blocks

### LINKED ISSUE #1507: ILLinker corrupts exception handler ranges for method
Steps to reproduce:

1. Install VS 16.8 Preview 3 and .net 5.0 rc1
2. Build https://github.com/modulardev2/ThreadException (Client wasm project)
3. Go to binaries folder, disassemble `System.Net.Http.dll`, look for method `<GetStringAsyncCore>d__41 MoveNext ()`. The EH clause ranges for the method are correct
4. Publish the same project (folder publish). Go to binaries folder, disassemble `System.Net.Http.dll`, look for same method as before and notice the clause ranges have been corrupted.

As part for publishing, `_RunILLink` is ran. This goes over the assemblies from the sdk, which are correct, and outputs them in the `obj\Release\net5.0\linked\` folder. At this moment the method is corrupted. The linker seems to run as follows : https://gist.github.com/BrzVlad/e0f358c9d096c5acef745a1c49aae8d1

While I haven't tried building the project from linux with dotnet command, the behavior should be the same.

marek-safar:
Backport of #1512

### DIFF
--- a/src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
+++ b/src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
@@ -591,6 +591,8 @@ namespace Mono.Linker.Steps
 								condBranches = new Stack<int> ();
 
 							condBranches.Push (GetInstructionIndex (handler.HandlerStart));
+							if (handler.FilterStart != null)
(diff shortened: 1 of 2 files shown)


marek-safar:
Yep, will submit that for approval tomorrow

PR #1515, opened 2020-09-26
Bump cecil version

marek-safar:


eng/common/templates/job/job.yml
```diff
@@ -217,7 +217,7 @@ jobs:
       displayName: Publish TRX Test Results
       inputs:
         testResultsFormat: 'VSTest'
-        testResultsFiles: '*.trx' 
+        testResultsFiles: '*.xml'
```
vitek-karas, line 220: Is this a break in Arcade or do we need to somehow backport this to Arcade?

marek-safar: Yep, I'm just testing if this is the right fix

vitek-karas:
The Windows tests are broken by https://github.com/mono/cecil/commit/191f9fc0a79a138fdb634fbb51cf836a86b4e565
When we rewrite a method's body to a throw we remove all debug scope information and all sequence points. When the assembly is written out the new behavior will force-read symbols from the original PDB for the method, but those symbols don't match the method body anymore.
The failure is because the native PDB writer validates this and fails. (I guess managed PDB writer doesn't care a produces broken PDB).

vitek-karas:
Actually this is a known issue now: https://github.com/jbevain/cecil/issues/694

marek-safar:
@vitek-karas I reverted that change and still the same problem :-(

vitek-karas:
That's because linker master still uses package reference to Mono.Cecil - so the fixes you picked up into the submodule don't apply. We will either have to wait for the fix to make it into cecil's master and eventually package, or switch fully to rely on the submodule.

marek-safar:
@vitek-karas yeah, I completely forgot about that

vitek-karas:
There's a formatting bug in the master - I just added a fix into this PR as it was faster...
Yesterday I saw build errors when building with the submodule on Windows - but apparently something in this PR fixes that - no build errors for me anymore (apart from the formatting fix which shows up as an error).

vitek-karas (approved):

PR #1521, opened 2020-10-01
Cleanup predefined attributes values extraction

marek-safar:


src/linker/Linker/DynamicDependency.cs
```diff
@@ -86,40 +86,28 @@ public DynamicDependency (DynamicallyAccessedMemberTypes memberTypes, string typ
...
+			DynamicDependency? result = args[0].Value switch
+			{
+				string stringMemberSignature => args.Count switch
+				{
+					1 => new DynamicDependency (stringMemberSignature),
+					2 when args[1].Value is TypeReference type => new DynamicDependency (stringMemberSignature, type),
+					3 when args[1].Value is string typeName && args[2].Value is string assemblyName => new DynamicDependency (stringMemberSignature, typeName, assemblyName),
+					_ => null,
+				},
+				int memberTypes => args.Count switch
```
vitek-karas, line 98: This changes the semantics of the code - not sure if it matters though. Previous we would check that the type of the `args[0].Type` was `DynamicallyAccessedMemberTypes`, now we simply look at the value and use it if it's an int. So this can probably consume more cases without failing.

Additionally before we casted the value to the enum and if it failed we also "failed", now we don't catch that exception. So if the cast fails for some reason, this would throw. That said, I'm actually not sure if casting int to enum can every fail.

vitek-karas (approved):
Even though there are functional differences I think this is OK to take.

src/linker/Linker/DynamicDependency.cs
sbomer: I'd prefer to keep the stricter behavior with the `IsTypeOf` check, but it makes little difference in practice.

The cast check was there to prevent using an enum backed by the wrong type. It would fail if the user tried to redefine the attribute to take `enum DynamicallyAccessedMemberTypes : UInt32` for example. I think hard-coding the implementation type `int` in the pattern match should cover the same cases.

sbomer (approved):

PR #1533, opened 2020-10-05
AddDirectDependency adjustments

mrvoorhe:
I don't know what we are going to do with `DependencyInfo` yet.  For now I'm just having us pass `new DependencyInfo(DependencyKind.Custom, null)` in a bunch of places.

This causes a lot of pointless tracing being recorded, which leads to CanEnableReducedTracing failing for us.

By making AddDirectDependency virtual we can override it and then ignore our dependency info's that we don't want recorded.

### DIFF
--- a/src/linker/Linker/Tracer.cs
+++ b/src/linker/Linker/Tracer.cs
@@ -68,7 +68,7 @@ namespace Mono.Linker
 			return recorders != null;
 		}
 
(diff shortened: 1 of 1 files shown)


mrvoorhe:
@marek-safar suggested using `Unspecified`.  I'll open a PR to add that to the list of acceptable reasons and see how that goes.

PR #1534, opened 2020-10-05
Don't record unspecified dependencies

mrvoorhe:
This gives us a way to punt on giving reasons to all of our marking behaviors.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -55,6 +55,7 @@ namespace Mono.Linker.Steps
 
 #if DEBUG
 		static readonly DependencyKind[] _entireTypeReasons = new DependencyKind[] {
+			DependencyKind.Unspecified,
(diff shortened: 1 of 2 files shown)


src/linker/Linker/Tracer.cs
```diff
@@ -70,6 +70,9 @@ bool IsRecordingEnabled ()
 
 		public void AddDirectDependency (object target, in DependencyInfo reason, bool marked)
 		{
+			if (reason.Kind == DependencyKind.Unspecified)
```
sbomer, line 73: Could we instead let each dependency recorder decide what to do with `Unspecified` dependencies?

mrvoorhe: Sure.  Done.

mrvoorhe:
@marek-safar Does this look good to merge?

mrvoorhe:
@sbomer Are you OK with the changes I made based on your comment?

vitek-karas (approved):

PR #1535, opened 2020-10-05
Unresolved attribute fixes

### LINKED ISSUE #1525: Linker crashes when skipping unresolved references
In .NET 5 we run linker with `--skip-unresolved` because even Microsoft.NETCore.App has assembly references pointing outside of its closure. So it's perfectly expected that calls to `.Resolve()` will sometime return null (if they point to missing assemblies).
Linker should handle these gracefully (not sure what that exactly means, but either a very descriptive error, warning, or some other kind of explicit handling). Currently linker hard crashes in several places if this happens.
For example - using attribute from unresolved assembly will end up in `NullReferenceException` from callstack like this:
```C#
 	illink.dll!Mono.Linker.CustomAttributeSource.HasCustomAttributes(Mono.Cecil.ICustomAttributeProvider provider) Line 46	C#
 	illink.dll!Mono.Linker.CustomAttributeSource.HasAttributes(Mono.Cecil.ICustomAttributeProvider provider) Line 77	C#
 	illink.dll!Mono.Linker.AnnotationStore.HasLinkerAttribute<Mono.Linker.RemoveAttributeInstancesAttribute>(Mono.Cecil.IMemberDefinition member) Line 474	C#
	illink.dll!Mono.Linker.Steps.MarkStep.MarkCustomAttributes(Mono.Cecil.ICustomAttributeProvider provider, Mono.Linker.DependencyInfo reason, Mono.Cecil.IMemberDefinition sourceLocationMember) Line 584	C#
```

We need to determine what is the correct behavior of linker in these situations, and implement it. It might go as far as actually banning explicit calls to `.Resolve` anywhere in the linker - or to start annotating linker with nullable annotations which should expose these errors as well.

mrvoorhe:
When we fail to resolve a type we try and recreate it and keep the IL valid rather than fail.  We have a test that ensures an unresolved attribute on an AssemblyDefinition can be recreated.  It started failing with the sync up.   LazyMarkCustomAttributes should not skip any attributes.  This would give inconsistent behavior with references that only have attributes marked.  It was also problematic because LazyMarkCustomAttributes would call HandleUnresolvedType on the type where as ProcessLazyAttributes will call HandleUnresolvedMethod. A call to HandleUnresolvedMethod is necessary for us to recreate valid IL.

I went ahead and added null checks to other places that were missing them such as the location @marek-safar mentioned in https://github.com/mono/linker/issues/1525

Fixes #1525

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -581,7 +581,13 @@ namespace Mono.Linker.Steps
 					if (UnconditionalSuppressMessageAttributeState.TypeRefHasUnconditionalSuppressions (ca.Constructor.DeclaringType))
 						_context.Suppressions.AddSuppression (ca, provider);
 
-					if (_context.Annotations.HasLinkerAttribute<RemoveAttributeInstancesAttribute> (ca.AttributeType.Resolve ()) && providerInLinkedAssembly)
+					var resolvedAttributeType = ca.AttributeType.Resolve ();
+					if (resolvedAttributeType == null) {
+						HandleUnresolvedType (ca.AttributeType);
+						continue;
+					}
+
+					if (_context.Annotations.HasLinkerAttribute<RemoveAttributeInstancesAttribute> (resolvedAttributeType) && providerInLinkedAssembly)
 						continue;
 
 					MarkCustomAttribute (ca, reason, sourceLocationMember);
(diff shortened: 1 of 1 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1194,6 +1188,11 @@ bool ProcessLazyAttributes ()
 					continue;
 				}
 
+				if (_context.Annotations.HasLinkerAttribute<RemoveAttributeInstancesAttribute> (resolved.DeclaringType)) {
```
marek-safar, line 1191: Use `&&`

mrvoorhe: Done

PR #1538, opened 2020-10-06
Expose method on Driver

mrvoorhe:
We have our own driver class and would like to reuse this helper method to parse the value of the --generate-warning-suppressions


mrvoorhe:
build

marek-safar (approved):

Issue #1536, opened 2020-10-05
DynamicallyAccessedMembers with a string TypeName of a generic type with different assemblies doesn't appear to work 


eerhardt:
See https://github.com/dotnet/runtime/issues/39709 for the tests that found this issue.

Using the following code:

```C#
using System;
using System.ComponentModel;
using System.Xml.Linq;

class Program
{
    static void Main(string[] args)
    {
        var xel = new XElement("someElement");
        var props = TypeDescriptor.GetProperties(xel);
        foreach (PropertyDescriptor prop in props)
        {
            Console.WriteLine(prop.Name);
        }
    }
}
```

Run the above program without trimming, and you get:

```
Attribute
Descendants
Element
Elements
Value
Xml
FirstAttribute
HasAttributes
HasElements
IsEmpty
LastAttribute
Name
NodeType
Value
FirstNode
LastNode
NextNode
PreviousNode
BaseUri
Document
Parent
```

Run the above program after it has been trimmed (TrimMode=link) and you get:

```
Name
```

The `TypeDescriptionProvider` attribute is attributed with `DynamicallyAccessedMembers`. However, the value being passed in is causing problems for the linker.

```C#
    [System.ComponentModel.TypeDescriptionProvider("MS.Internal.Xml.Linq.ComponentModel.XTypeDescriptionProvider`1[[System.Xml.Linq.XElement, System.Xml.Linq, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]],System.ComponentModel.TypeConverter")]
    public class XElement : XContainer, IXmlSerializable
    {
```

It appears this code is the problem:

https://github.com/mono/linker/blob/90fd1feac709a910ba2aa849657b0c04777fc67d/src/linker/Linker/TypeNameResolver.cs#L63-L66

This code assumes that the generic type argument is in the same assembly as the outer generic type. But this is not the case above - the outer generic type is in `System.ComponentModel.TypeConverter`. While the generic argument is in `System.Xml.Linq`.

I tried fixing this myself, but then I ran into that `_context.GetLoadedAssembly("System.Xml.Linq")` was returning `null` for some reason. We should investigate this complete scenario and ensure it works properly.

cc @mateoatr @vitek-karas @marek-safar

vitek-karas:
@eerhardt What have you done 😄 - as I dig into this I find more and more "broken things".

[...]

PR #1546, opened 2020-10-07
Correctly handle assembly qualified names in generic parameters

vitek-karas:
We always assumed that the generic parameters would come from the same assembly as the generic type itself.
Adds a new test to validate the scenario.

Fixes the primary issue in #1536 
The other parts to make that scenario work fully are tracked by different issues.

### DIFF
--- a/test/Mono.Linker.Tests.Cases/DataFlow/ApplyTypeAnnotations.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/ApplyTypeAnnotations.cs
@@ -150,11 +150,11 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 		}
 
 		[Kept]
-		[KeptMember(".ctor()")]
+		[KeptMember (".ctor()")]
 		class FromStringConstantWithGenericAndAssemblyQualified<T>
 		{
 			[Kept]
-			public T GetValue() { return default (T);  }
+			public T GetValue () { return default (T); }
 		}
 
 		[Kept]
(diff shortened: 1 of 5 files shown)


test/Mono.Linker.Tests.Cases/DataFlow/ApplyTypeAnnotations.cs
```diff
@@ -123,5 +125,45 @@ class FromStringConstantWithAnnotationTestType
...
+		class FromStringConstantWithGeneric<T>
+		{
+			[Kept]
+			public T GetValue () { return default (T); }
+		}
+
+		[Kept]
+		static void TestFromStringConstantWithGeneric ()
+		{
+			RequireCombinationOnString ("Mono.Linker.Tests.Cases.DataFlow.ApplyTypeAnnotations+FromStringConstantWithGeneric`1[[Mono.Linker.Tests.Cases.DataFlow.ApplyTypeAnnotations+FromStringConstantWithGenericInner]]");
```
marek-safar, line 147: Do we have a test with assembly name which does not exist or is invalid?

mateoatr (approved):
LGTM. We should add the test for invalid assemblies.

test/Mono.Linker.Tests.Cases/DataFlow/ApplyTypeAnnotations.cs
vitek-karas: I added them - but they're not really testing much - we intentionally ignore unresolved type names, so it's really just testing that we don't crash on it.

marek-safar (approved):

PR #1548, opened 2020-10-07
Allow skipping the processing of a reflection call if its MarkStep's ProcessReflectionDependency() returns true

brian-taylor-unity:
Allow ReflectionMethodBodyScanner's HandleCall() to skip processing an instruction if its MarkStep's ProcessReflectionDependency() returns true.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -487,13 +487,17 @@ namespace Mono.Linker.Dataflow
 
 		public override bool HandleCall (MethodBody callingMethodBody, MethodReference calledMethod, Instruction operation, ValueNodeList methodParams, out ValueNode methodReturnValue)
 		{
+			methodReturnValue = null;
+
+			var reflectionProcessed = _markStep.ProcessReflectionDependency (callingMethodBody, operation);
+			if (reflectionProcessed)
+				return false;
+
 			var callingMethodDefinition = callingMethodBody.Method;
 			bool shouldEnableReflectionWarnings = ShouldEnableReflectionPatternReporting (callingMethodDefinition);
 			var reflectionContext = new ReflectionPatternContext (_context, shouldEnableReflectionWarnings, callingMethodDefinition, calledMethod.Resolve (), operation);
 
 			DynamicallyAccessedMemberTypes returnValueDynamicallyAccessedMemberTypes = 0;
-
-			methodReturnValue = null;
 
 			var calledMethodDefinition = calledMethod.Resolve ();
 			if (calledMethodDefinition == null)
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -487,14 +487,18 @@ static IntrinsicId GetIntrinsicIdForMethod (MethodDefinition calledMethod)
 
 		public override bool HandleCall (MethodBody callingMethodBody, MethodReference calledMethod, Instruction operation, ValueNodeList methodParams, out ValueNode methodReturnValue)
 		{
+			methodReturnValue = null;
+
+			var reflectionProcessed = _markStep.ProcessReflectionDependency (callingMethodBody, operation);
```
vitek-karas, line 492: It seems the method will be called twice for each call. Once here and second time in the existing `MarkStep.MarkReflectionLikeDependencies`.

I don't know exactly what Unity wants here - maybe it would be easier to add an extension point which disables the data flow analysis completely (my understanding is that Unity has its own version of that anyway).
There's no virtual method to do that yet, but it would be pretty simple to add one like that.

mrvoorhe: See https://discordapp.com/channels/732297728826277939/751137004007456849/763083086782529556

Open to ideas.  We were struggling to find a clean way to accomplish our goal.

As for being called twice, we should remove that.  This entire loop is pointless https://github.com/mono/linker/blob/9a4a5a06d54ed13f2215a7e0d0e39663ee10b51d/src/linker/Linker.Steps/MarkStep.cs#L2965 .  @brian-taylor-unity Can you update this PR to remove that entire loop over the instructions and calling `ProcessReflectionDependency`?

vitek-karas: Thanks - I read up on the Discord and it makes much more sense now. Once we remove the useless loop I'm fine with this change - if it works for you.

vitek-karas: Nit: Do we need the local variable?

vitek-karas (approved):

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
brian-taylor-unity: Not necessary, I added it mainly to encapsulate some of the meaning of why it would cause an early-return in the method, but I don't really have a strong opinion either way.

vitek-karas: OK - I don't have strong feeling about it either

Issue #1404, opened 2020-08-03
Linker warnings should be grouped by type/file


davidfowl:
When dumping linker warnings it would be ideal to tackle warnings file by file or type by type but the linker seems to dump warnings in the order in which types of visited. This is understandable by not as ordering the errors by type/file.

marek-safar:
linker does not know about source files and the order there were used for compilation in all the time so that's not a viable option.

However, grouping the marking step warnings on assembly level would make sense to me.

mateoatr:
We could improve the warning experience by caching the warning messages so that we can group them and sort them in a predictable way before printing.

davidfowl:
That would be great 😄

mateoatr:
We've added a `MessageContainer` cache that sorts the warnings and prints them just before the `SweepStep`. The ordering of the warnings is done by [comparing their origin](https://github.com/mateoatr/linker/blob/b7cee486dcc61a782034fcfffef895663e7ebb1d/src/linker/Linker/MessageContainer.cs#L173-L182), if none of the warnings being compared has an attached origin then we compare their warning code.

PR #1550, opened 2020-10-08
Add cache to the default logger

mateoatr:
Allow `ConsoleLogger` to cache logged messages based on their `MessageCategory`. With this we can delay the printing of warnings until the linker is finished and postprocess as needed. This also adds an ordering for `MessageContainer` so that we can sort warnings (#1404).


src/linker/Linker.Steps/ProcessWarnings.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class ProcessWarnings : BaseStep
```
vitek-karas, line 9: This should be called `ProcessWarningsStep` - for consistency.

src/linker/Linker/ConsoleLogger.cs
```diff
@@ -2,15 +2,59 @@
...
 	public class ConsoleLogger : ILogger
 	{
-		public void LogMessage (MessageContainer message)
+		public Dictionary<MemberReference, string> ComputedStrings;
+
+		readonly List<MessageContainer> _messageContainers;
+		readonly MessageCategory _categoriesToCache;
+		readonly StreamWriter _streamWriter;
+
+		public ConsoleLogger (StreamWriter streamWriter, MessageCategory categoriesToCache = MessageCategory.None)
```
vitek-karas, line 19: Do we use the fact that we can redirect to any stream? If not, I would omit the parameter and simply do what we did before - `Console.WriteLine`.

vitek-karas: Also do we have a use case for caching anything but warnings?

I don't see a need to build a parametrized logger if we're not using it that way.

src/linker/Linker.Steps/ProcessWarnings.cs
```diff
@@ -0,0 +1,28 @@
...
+			if (Context.Logger is not ConsoleLogger consoleLogger)
+				return;
+
+			foreach (var mc in consoleLogger.GetCachedMessages ()) {
+				if (mc.Origin?.MemberDefinition is not MemberReference memberReference ||
+					memberReference == null)
+					continue;
+
+				consoleLogger.ComputedStrings[memberReference] = mc.ToString ();
+			}
```
vitek-karas, line 25: This should be in the ConsoleLogger - no need to expose the internals of the logger just so that we can do it here.

src/linker/Linker/MessageCategory.cs
```diff
@@ -2,15 +2,19 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
+
 namespace Mono.Linker
 {
+	[Flags]
```
vitek-karas, line 9: I would not do this - it will be confusing for all the callsites where we specify message category for loggin - where only one value is every allowed.

src/linker/Linker/ConsoleLogger.cs
```diff
@@ -2,15 +2,59 @@
...
+using System.Collections.Generic;
+using System.IO;
+using Mono.Cecil;
 
 namespace Mono.Linker
 {
 	public class ConsoleLogger : ILogger
 	{
-		public void LogMessage (MessageContainer message)
+		public Dictionary<MemberReference, string> ComputedStrings;
```
vitek-karas, line 13: Storing this per member reference will not always work. What if there are two warnings for the same member reference?
Anyway - the only reason to do is so that we could potentially sort the messages at the very end - but that won't work since the MemberRefereces will be useless at that point (if removed by Clean).
I would actually just immediately output all of the messages in the ProcessWarningsStep.

Thinking about this more - this would work with any logger - simply cache messages until the ProcessWarningStep - then potentially sort them and immediately output.

If there's any warning after that step, we can either cache and flush at the end , or disable the cache and immediately output.

That way we would get the same behavior for all loggers - which is probably preferable.

src/linker/Linker/ConsoleLogger.cs
```diff
@@ -2,15 +2,59 @@
...
+
+			_streamWriter?.WriteLine (messageContainer.ToString ());
+		}
+
+		public void Flush ()
+		{
+			if (_streamWriter is null)
+				return;
+
+			_messageContainers.Sort ();
```
vitek-karas, line 42: I assume this will crash if one of the warnings is on a removed member reference - right?

src/linker/Linker/MessageOrigin.cs
```diff
@@ -75,5 +76,23 @@ public override string ToString ()
 		public override int GetHashCode () => (FileName, MemberDefinition, SourceLine, SourceColumn).GetHashCode ();
 		public static bool operator == (MessageOrigin lhs, MessageOrigin rhs) => lhs.Equals (rhs);
 		public static bool operator != (MessageOrigin lhs, MessageOrigin rhs) => !lhs.Equals (rhs);
+
+		public int CompareTo (MessageOrigin other)
+		{
+			if (MemberDefinition != null && other.MemberDefinition != null) {
+				return (MemberDefinition.DeclaringType.Module.Assembly.Name.Name, MemberDefinition.DeclaringType.Name, MemberDefinition.Name).CompareTo
+					((other.MemberDefinition.DeclaringType.Module.Assembly.Name.Name, other.MemberDefinition.DeclaringType.Name, other.MemberDefinition.Name));
```
vitek-karas, line 84: We should do `MemberDefinition?.DeclaringType?.Module?.Assembly?.Name?.Name` - this code should better work no matter what crazy input it gets - and with Cecil it's hard to rely on non-null anywhere.
Simply put - I would rather have wrong sorting than crash the linker while writing out warnings.

vitek-karas (changes requested):

src/linker/Linker/ConsoleLogger.cs
```diff
@@ -2,15 +2,59 @@
...
+				_messageContainers.Add (messageContainer);
+				return;
+			}
+
+			_streamWriter?.WriteLine (messageContainer.ToString ());
+		}
+
+		public void Flush ()
+		{
+			if (_streamWriter is null)
```
marek-safar, line 39: When will this happen and why don't check this earlier?

src/linker/Linker/ConsoleLogger.cs
```diff
@@ -2,15 +2,59 @@
...
+				return;
+
+			_messageContainers.Sort ();
+			foreach (var messageContainer in _messageContainers) {
+				_messageContainers.Remove (messageContainer);
```
marek-safar, line 44: I'm surprised this does not crash, you are removing item while enumerating the container

src/linker/Linker/MessageCategory.cs
```diff
@@ -2,15 +2,19 @@
...
 	public enum MessageCategory
 	{
-		Error = 0,
-		Warning,
-		Info,
-		Diagnostic,
-
-		WarningAsError = 0xFF
+		None = 0,
```
marek-safar, line 12: What is `MessageCategory.None` suppose to represent?

src/linker/Linker/MessageOrigin.cs
```diff
@@ -75,5 +76,23 @@ public override string ToString ()
 		public override int GetHashCode () => (FileName, MemberDefinition, SourceLine, SourceColumn).GetHashCode ();
 		public static bool operator == (MessageOrigin lhs, MessageOrigin rhs) => lhs.Equals (rhs);
 		public static bool operator != (MessageOrigin lhs, MessageOrigin rhs) => !lhs.Equals (rhs);
+
+		public int CompareTo (MessageOrigin other)
+		{
+			if (MemberDefinition != null && other.MemberDefinition != null) {
```
marek-safar, line 82: All this should use switch expression

src/linker/Linker/ConsoleLogger.cs
mateoatr: With the changes I just pushed there's no way this can happen. I originally put this so that we could swallow all the output in the test infra (I'm now redirecting `Console` stdout to a `StringWriter`).

src/linker/Linker/ConsoleLogger.cs
mateoatr: > Simply cache messages until the ProcessWarningStep - then potentially sort them and immediately output.

I ended up doing this. We keep caching after the `ProcessWarningsStep` and then we flush at the very end.

src/linker/Linker/ConsoleLogger.cs
mateoatr: > Also do we have a use case for caching anything but warnings?

I don't think so. I've changed this back so that we keep using `Console` and the only messages ever cached (and sorted) are the warnings.

src/linker/Linker/ConsoleLogger.cs
mateoatr: I think that by adding the null conditionals on `MessageOrigin.CompareTo` this shouldn't happen anymore.

src/linker/Linker/MessageCategory.cs
```diff
@@ -6,11 +6,11 @@ namespace Mono.Linker
 {
 	public enum MessageCategory
 	{
-		Error = 0,
-		Warning,
-		Info,
-		Diagnostic,
-
-		WarningAsError = 0xFF
+		Error = 1,
```
vitek-karas, line 9: The `All` value doesn't make sense anymore. I would just revert this entire file.

src/linker/ref/Linker/MessageCategory.cs
```diff
@@ -6,9 +6,9 @@ namespace Mono.Linker
 {
 	public enum MessageCategory
 	{
-		Error = 0,
-		Warning,
-		Info,
-		Diagnostic
+		Error = 1,
```
vitek-karas, line 9: Any reason to make these values a bit-field?
Note that technically we've shipped this, so this is a breaking change. I don't mind that much, but we should have a reason - I don't see one right now.

src/linker/Linker/ConsoleLogger.cs
```diff
@@ -3,14 +3,41 @@
...
 
 using System;
+using System.Collections.Generic;
 
 namespace Mono.Linker
 {
 	public class ConsoleLogger : ILogger
 	{
-		public void LogMessage (MessageContainer message)
+		readonly List<MessageContainer> _messageContainers;
```
vitek-karas, line 12: Any reason why this is tied to the ConsoleLogger? I would prefer if we have the same behavior across all loggers. Especially since there's no technical reason anymore. Basically build this into the context and just call Logger.LogMessage after sorting to output.

vitek-karas (changes requested):

src/linker/Linker/LinkContext.cs
```diff
@@ -171,7 +172,7 @@ public class LinkContext : IDisposable
 
 		public bool LogMessages { get; set; }
 
-		public ILogger Logger { private get; set; } = new ConsoleLogger ();
+		public ILogger Logger { internal get; set; }
```
vitek-karas, line 175: Do we need this to be internal?

vitek-karas (approved):
LGTM

marek-safar (approved):

PR #1556, opened 2020-10-08
Restore the ability to not mark System.Object.Finalize

mrvoorhe:
Previously we overrode `GetRequiredMethodsForInstantiatedType` and used that to disable marking of Finalize during a special use case we have.

That broke with https://github.com/mono/linker/commit/9c84659e8ff12a7af366a5f4d64ffc6c3623618c

This PR adds a simple bool property to allow us to switch the new behavior that change added off.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1448,7 +1448,7 @@ namespace Mono.Linker.Steps
 				MarkMulticastDelegate (type);
 			}
 
-			if (type.IsClass && type.BaseType == null && type.Name == "Object")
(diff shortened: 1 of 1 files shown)


marek-safar (approved):

MichalStrehovsky:
We also have such special environments, but we fix that by simply not putting the `Finalize` method on `object` in the base class library. The nice side effect is that one can't even declare a finalizer in the C# source code (instead of declaring it and then being surprised in never runs because it was stripped).

mrvoorhe:
Our use case is a little more complex than I led on.  We have the small profile without Object.Finalize as well.

What we do is use our linker (running in a special mode) to strip down the netstandard reference assembly to contain only what is in our small profile. That gives us a "tiny netstandard" that we then compile against.

PR #1563, opened 2020-10-13
Add first roslyn analyzer

agocke:
This analyzer approximates the warnings produced by the linker around
the RequiresUnreferencedCodeAttribute.

Tip for anyone not used to reviewing code w/ localization support: the XLF files are only for loc and can be hidden by going to the File Filter at the top of the review page and unchecking the "xlf" box for file types.


illink.sln
```diff
@@ -27,6 +27,10 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ref", "ref", "{03EB085F-3E2
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Mono.Linker", "src\linker\ref\Mono.Linker.csproj", "{57BE47DF-DCDF-44EE-B77F-F8E8AD069076}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "RoslynAnalyzer.Tests", "test\RoslynAnalyzer.Tests\RoslynAnalyzer.Tests.csproj", "{51BE55C4-FA11-4FD2-BC65-F27885A1BF0B}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "RoslynAnalyzer", "src\RoslynAnalyzer\RoslynAnalyzer.csproj", "{63562853-477E-4F98-82C9-BC0A45639868}"
```
marek-safar, line 32: What is the naming convention for analyzers? RoslynAnalyzer sounds a bit odd to me

src/RoslynAnalyzer/AnalyzerOptionsExtensions.cs
```diff
@@ -0,0 +1,23 @@
+
```
marek-safar, line 1: Missing licence header

test/RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1+Test.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Testing.Verifiers;
+using Microsoft.CodeAnalysis.Text;
+
+namespace Analyzer1.Test
+{
+    public static partial class CSharpAnalyzerVerifier<TAnalyzer>
+        where TAnalyzer : DiagnosticAnalyzer, new()
+    {
+        public class Test : CSharpAnalyzerTest<TAnalyzer, XUnitVerifier>
```
marek-safar, line 13: What is this class actually trying to do (perhaps rename it) ?

test/RoslynAnalyzer.Tests/RoslynAnalyzer.Tests.csproj
```diff
@@ -0,0 +1,24 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(MonoBuild)' != ''">
+    <TargetFramework>net472</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.XUnit" Version="1.0.1-beta1.*" />
+     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.7.0" />
+     <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Analyzer.Testing.XUnit" Version="1.0.1-beta1.*" />
```
marek-safar, line 20: What is this for?

test/RoslynAnalyzer.Tests/RoslynAnalyzer.Tests.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <DefineConstants>$(DefineConstants);ILLINK</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(MonoBuild)' != ''">
+    <TargetFramework>net472</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.XUnit" Version="1.0.1-beta1.*" />
+     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.7.0" />
```
marek-safar, line 19: The version should come from Arcade

test/RoslynAnalyzer.Tests/RoslynAnalyzer.Tests.csproj
agocke: This analyzer should be functional in VB as well, but I decided not to explicitly test it because it doesn't seem worth the effort.

test/RoslynAnalyzer.Tests/RoslynAnalyzer.Tests.csproj
agocke: This is a bit tricky. We probably want two versions here: Arcade for the tests, hardcoded for the project. This is because analyzers are plugins, so upgrading the required analyzer version effectively reduces your platform support to only newer .NET SDKs/Visual Studio installs.

test/RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1+Test.cs
agocke: This is a standard (generated) part of the Roslyn Analyzer SDK -- we can modify it, but I just left it as is for now.

src/RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -0,0 +1,104 @@
...
+
+                context.RegisterOperationAction(operationContext =>
+                {
+                    var invocation = (IInvocationOperation)operationContext.Operation;
+                    var method = invocation.TargetMethod;
+                    var attributes = method.GetAttributes();
+
+                    foreach (var attr in attributes)
+                    {
+                        if (attr.AttributeClass is { } attrClass &&
```
vitek-karas, line 65: Perf nit: this will full scan all attributes on all called methods - for a reasonably sized assembly this will perform a lot of comparisons. I wonder if there's some way to "cache" some information to prevent such full scans.
Linker has even bigger potential problem (as it processes lot more code at once) - it solves it by storing per method cache of "linker attributes" (basically attributes recognized by linker). Since almost all methods have this cache empty, it's very fast to check that there's nothing to do.

I don't know if there's some typical pattern how roslyn analyzers solve this problem (or if they actually solve it at all).

src/RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -0,0 +1,104 @@
...
+							IsNamedType(attrClass, "System.Diagnostics.CodeAnalysis.RequiresUnreferencedCodeAttribute") &&
+							attr.ConstructorArguments.Length == 1 &&
+							attr.ConstructorArguments[0] is { Type: { SpecialType: SpecialType.System_String } } ctorArg)
+                        {
+							operationContext.ReportDiagnostic (Diagnostic.Create (
```
vitek-karas, line 70: As the first prototype this is good. Eventually we will need to make sure that if the caller has the attribute, then it doesn't generate warnings in its body, for this attribute.

test/RoslynAnalyzer.Tests/RequiresUnreferencedCodeTest.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace ILTrimmingAnalyzer.Test
+{
+    public class RequiresUnreferencedCodeTest
+    {
+        [Fact]
+        public Task SimpleDiagnostic()
+        {
+            var src = @"
+using System.Diagnostics.CodeAnalysis;
```
vitek-karas, line 18: Looking forward - is there some typical way how these are written as .cs files so that writing them can use intellisense and so on?
Not sure how important that is for Roslyn, but in our case all of these snippets will be perfectly valid C#.

test/RoslynAnalyzer.Tests/RequiresUnreferencedCodeTest.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Threading.Tasks;
+using Xunit;
+using VerifyCS = Analyzer1.Test.CSharpAnalyzerVerifier<
```
vitek-karas, line 7: `Analyzer1` ? That's a really weird namespace...

test/RoslynAnalyzer.Tests/Verifiers/CSharpVerifierHelper.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        internal static ImmutableDictionary<string, ReportDiagnostic> NullableWarnings { get; } = GetNullableWarningsFromCompiler();
+
+        private static ImmutableDictionary<string, ReportDiagnostic> GetNullableWarningsFromCompiler()
+        {
+            string[] args = { "/warnaserror:nullable" };
+            var commandLineArguments = CSharpCommandLineParser.Default.Parse(args, baseDirectory: Environment.CurrentDirectory, sdkDirectory: Environment.CurrentDirectory);
+            var nullableWarnings = commandLineArguments.CompilationOptions.SpecificDiagnosticOptions;
+
+            // Workaround for https://github.com/dotnet/roslyn/issues/41610
```
vitek-karas, line 25: Not this code's problem - but the issue is closed... does it mean we're not using the latest bits?

src/RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
agocke: A cache here would probably be fine, but symbols are not guaranteed reference equality in Roslyn, so in cases where there is no reference-equal member in the cache, it may be faster to loop over the attributes instead of comparing the symbols.

test/RoslynAnalyzer.Tests/RequiresUnreferencedCodeTest.cs
agocke: Nope, no way.

test/RoslynAnalyzer.Tests/Verifiers/CSharpVerifierHelper.cs
agocke: Just an older template, hasn't been updated apparently.

agocke:
@vitek-karas I think this is ready for re-review.

src/ILLink.RoslynAnalyzer/OperationExtensions.cs
```diff
@@ -0,0 +1,251 @@
...
+            | sizeof(x)                |      |       |             |             |       ✔️        | ️
+            | typeof(x)                |      |       |             |             |       ✔️        | ️
+            | out var x                |      |  ✔️   |             |             |                 | ️
+            | case X x:                |      |  ✔️   |             |             |                 | ️
+            | obj is X x               |      |  ✔️   |             |             |                 |
+            | ref var x =              |      |       |     ✔️      |     ✔️      |                 |
+            | ref readonly var x =     |      |       |     ✔️      |             |                 |
+ 
+            */
+			if (operation is ILocalReferenceOperation localReference &&
```
marek-safar, line 43: Why not to use switch instead?

test/ILLink.RoslynAnalyzer.Tests/ILLink.RoslynAnalyzer.Tests.csproj
```diff
@@ -0,0 +1,23 @@
...
+  <PropertyGroup>
+    <Nullable>enable</Nullable>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(MonoBuild)' == ''">
+    <TargetFramework>net5.0</TargetFramework>
+    <DefineConstants>$(DefineConstants);ILLINK</DefineConstants>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(MonoBuild)' != ''">
```
marek-safar, line 12: The analyzer should be built for netcore setup only

test/ILLink.RoslynAnalyzer.Tests/ILLink.RoslynAnalyzer.Tests.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <TargetFramework>net471</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing.XUnit" Version="1.0.1-beta1.*" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.7.0" />
```
marek-safar, line 18: Please extract all the version to Versions.props

vitek-karas (approved):
The product side looks good.
I have some doubts about the test design, but it works, so we can use it until we figure out what we want.

src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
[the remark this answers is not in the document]
agocke: Analyzers have to be netstandard2.0 because Roslyn also runs on the desktop FX.

src/ILLink.RoslynAnalyzer/OperationExtensions.cs
agocke: Copy-paste from Roslyn, I didn't change the code. Hopefully this will become a public API at some point.

test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
```diff
@@ -0,0 +1,33 @@
...
+using Xunit;
+
+namespace ILLink.RoslynAnalyzer.Tests
+{
+	/// <summary>
+	/// Test cases stored in files
+	/// </summary>
+	public class LinkerTestCases : TestCaseUtils
+	{
+		[Theory]
```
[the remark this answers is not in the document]
agocke: In this case it's filtering out methods since the analyzer doesn't support them. However, it's still somewhat useful to be able to narrow a test down to a small scenario. Also, my goal was to not touch the linker test code at all.

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -0,0 +1,200 @@
...
+			}
+		}
+
+		internal static void RunTest (MethodDeclarationSyntax m, List<AttributeSyntax> attrs, ImmutableArray<Diagnostic> diags)
+		{
+			var filtered = diags.Where (d => d.Location.SourceSpan.IntersectsWith (m.Span))
+								.Select (d => d.GetMessage ());
+			foreach (var attr in attrs) {
+				switch (attr.Name.ToString ()) {
+				case "LogContains": {
```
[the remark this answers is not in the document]
agocke: I do check that the analyzer span overlaps with the method span, so there is some testing of location, although it's not precise.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+	[DiagnosticAnalyzer (LanguageNames.CSharp)]
+	public class RequiresUnreferencedCodeAnalyzer : DiagnosticAnalyzer
+	{
+		public const string DiagnosticId = "IL2026";
+
+		/// <summary>
+		/// Configuration only for Mono.Linker.Tests.Cases to specially format types
+		/// to match the Cecil representation.
+		/// </summary>
+		private const string s_cecilCompatPropertyName = "use_cecil_compat_format";
```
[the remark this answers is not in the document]
agocke: It's not I forgot to remove it.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+						}
+					}
+				}
+			});
+		}
+
+		/// <summary>
+		/// Returns true if <see paramref="type" /> has the same name as <see paramref="typename" />
+		/// </summary>
+		internal static bool IsNamedType (INamedTypeSymbol type, string typeName)
```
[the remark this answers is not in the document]
agocke: Yeah, seems reasonable.

src/ILLink.RoslynAnalyzer/OperationExtensions.cs
marek-safar: Could you please add a comment where the file was copied from

src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
```diff
@@ -0,0 +1,15 @@
...
+    <LangVersion>8</LangVersion>
+    <Nullable>enable</Nullable>
+    <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Include="Resources.resx" GenerateSource="true" />
+    <PackageReference Include="microsoft.codeanalysis.csharp" Version="3.7.0" />
```
marek-safar, line 12: Move version to versions.prop

marek-safar (approved):

PR #1565, opened 2020-10-13
Fix processing of DynamicDependency when specified via attribute XML

### LINKED ISSUE #1547: DynamicDependencyAttributes are not correctly processed if specified via linkattributes.xml
Specifying `DynamicDependencyAttribute` via the `linkattributes.xml` does fully work since the code incorrectly checks for their presence only if the annotated method has any real attributes (from IL).

vitek-karas:
The `DynamicDependencyLookupStep` would not process `DynamicDependencyAttribute` specified via XML on a member without real attributes (from IL).
Fix is to not rely on `member.HasCustomAttributes` as that doesn't see attributes specified via XML.

Fixes #1547


tlakollo (approved):
LGTM

marek-safar (approved):

Issue #1537, opened 2020-10-06
DynamicallyAccessedMembers does not mark the type if passed in as string


vitek-karas:
The below has to be built into a classlib which is then aggressively trimmed:
```C#
namespace lib
{
    public class Test
    {
        public static void TestIt()
        {
            object o = CreateInstanceOf("lib.GenOut`1[[lib.GenIn]]");
            Console.WriteLine(o);
        }

        static object CreateInstanceOf([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] string n)
        {
            return Activator.CreateInstance(Type.GetType(n));
        }
    }


    class GenOut<T> where T : new()
    {
        T GetValue() { return new T(); }
    }

    class GenIn
    {
    }
}
```

Per annotations, it is perfectly valid to call `Activator.CreateInstance` due to the annotation on the parameter of `CreateInstanceOf`. Linker doesn't produce any warnings from the code above.

But when executed, the application fails with:
```console
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'type')
   at System.Activator.CreateInstance(Type type, Boolean nonPublic, Boolean wrapExceptions)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic)
   at System.Activator.CreateInstance(Type type)
   at lib.Test.CreateInstanceOf(String n) in F:\ILLinker\repro\TypeNameRep\lib\Class1.cs:line 16
   at lib.Test.TestIt() in F:\ILLinker\repro\TypeNameRep\lib\Class1.cs:line 10
   at Program.Main(String[] args) in F:\ILLinker\repro\TypeNameRep\TypeNameRep\Program.cs:line 17
```

The failure is because `Type.GetType` returns null. In the trimmed assembly there's actually no type `GenOut`.
There are two weird things here:
1. `GenOut<>.ctor()` should be kept, but for some reason we leave it out
1. Even if I force include `GenOut<>` (by having a direct typeref to it), `GenIn` also needs to be kept, which won't happen either

vitek-karas:
Interestingly changing the code to call ``CreateInstanceOf("lib.GenOut`1")`` actually works, the `GenOut<>.ctor` is kept and the app fails because it's invalid to instantiate open generic type.

vitek-karas:
I also found https://github.com/dotnet/sdk/issues/13999 when working on this bug.

vitek-karas:
[...]

Issue #1559, opened 2020-10-09
Generic type array causes InvalidOperationException in ReflectionMethodBodyScanner.GetValueNodeFromGenericArgument


vitek-karas:
Repro copied from https://github.com/dotnet/runtime/issues/43222

### Steps To Reproduce

0. Install .NET 5 RC1

1. Create some project:

``` 
dotnet new blazorwasm
```

2. Add some new *.cs file with following C# code:

```c#
using System.Text.Json;

namespace SomeNamespace
{
    public static class StaticClass
    {
        public static void GenericMethod<T>()
        {
            T[] value = null;
            JsonSerializer.Serialize(value);
        }
    }
}
```

3. Try to publish

```
dotnet publish
```

### Exception

```
Class.cs(8,9): error IL1005: SomeNamespace.StaticClass.GenericMethod<T>(): Error processing method 'SomeNamespace.StaticClass.GenericMethod<T>
()' in assembly 'TrimmingIssue.dll' [C:\projectsBlazor\Net5Publish\TrimmingIssue\TrimmingIssue.csproj]
  Mono.Linker.LinkerFatalErrorException: C:\projectsBlazor\Net5Publish\TrimmingIssue\Class.cs(8,9): error IL1005: SomeNamespace.StaticClass.GenericMethod<T>(): Error processing method 'S
omeNamespace.StaticClass.GenericMethod<T>()' in assembly 'TrimmingIssue.dll'
   ---> System.InvalidOperationException: Operation is not valid due to the current state of the object.
     at Mono.Linker.Dataflow.ReflectionMethodBodyScanner.GetValueNodeFromGenericArgument(TypeReference genericArgument)
     at Mono.Linker.Dataflow.ReflectionMethodBodyScanner.ProcessGenericArgumentDataFlow(GenericParameter genericParameter, TypeReference genericArgument, IMemberDefinition source)
     at Mono.Linker.Steps.MarkStep.MarkGenericArgumentConstructors(IGenericInstance instance, IMemberDefinition sourceLocationMember)
     at Mono.Linker.Steps.MarkStep.MarkGenericArguments(IGenericInstance instance, IMemberDefinition sourceLocationMember)
     at Mono.Linker.Steps.MarkStep.GetOriginalMethod(MethodReference method, DependencyInfo reason, IMemberDefinition sourceLocationMember)
[...]

Issue #1087, opened 2020-04-09
[feature/dataflow] Support `DynamicallyAccessedMemberKinds.RecursiveProperties`


eerhardt:
For (de)serialization cases, like in `System.Text.Json`, we need a way to tell the linker to keep the properties of type `T`, and also the properties of any types referenced by `T`'s properties recursively.

Take the example in https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to:

```C#
public class WeatherForecast
{
    public DateTimeOffset Date { get; set; }
    public int TemperatureCelsius { get; set; }
    public string Summary { get; set; }
    public string SummaryField;
    public IList<DateTimeOffset> DatesAvailable { get; set; }
    public Dictionary<string, HighLowTemps> TemperatureRanges { get; set; }
    public string[] SummaryWords { get; set; }
}

public class HighLowTemps
{
    public int High { get; set; }
    public int Low { get; set; }
}

...

jsonString = File.ReadAllText(fileName);
weatherForecast = JsonSerializer.Deserialize<WeatherForecast>(jsonString);
```

Note that the properties of `HighLowTemps` need to be preserved even though I didn't directly pass `HighLowTemps` into `JsonSerializer.Deserialize`. Also note that when walking the properties recursively, we need to handle generic collections: `Dictionary<string, HighLowTemps>`.

MichalStrehovsky:
Can we have someone who understand System.Text.Json really well look at whether they can make things work with just a capability like this? We spent a lot of time trying to make reflection-based serialization in .NET Native "just work", to no avail - serializers pretty much never operate in these naive ways, and most then have a bunch of pluggability that makes any analysis impossible.

For this example:
* How is this going to instantiate types? I assume we need to new up the `Dictionary` somehow to deserialize, but nothing guarantees the constructor is kept.
* How does deserialization of an IList work? I assume System.Text.Json special cases/hardcodes something for this that can actually be instantiated instead of the interface.

eerhardt:
cc @steveharter @layomia @Jozkee - the owners of System.Text.Json.

> look at whether they can make things work with just a capability like this?

I'm sure it will take other capabilities as well, for example `PreserveDependency` that is used now:

https://github.com/dotnet/runtime/blob/d59f143221bfa500fe2b974827dcb47eb4ce327c/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Converters/Collection/IEnumerableConverterFactory.cs#L28-L46

[...]

PR #1566, opened 2020-10-14
Mark generic arguments of dynamically accessed types passed as string

mateoatr:
Currently, the linker won't mark the generic arguments of a dynamically accessed type that is passed as a string. For an example, see #1537.

### DIFF
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -709,9 +709,7 @@ namespace Mono.Linker.Dataflow
 			}
 
 			if (operation.Operand is TypeReference typeReference) {
-				var resolvedReference = typeReference is ArrayType ?
-					typeReference.Module.ImportReference (typeof (Array))?.Resolve () : typeReference.Resolve ();
-
+				var resolvedReference = typeReference.ResolveToMainTypeDefinition ();
 				if (resolvedReference != null) {
 					StackSlot slot = new StackSlot (new RuntimeTypeHandleValue (resolvedReference));
 					currentStack.Push (slot);
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -708,6 +708,12 @@ namespace Mono.Linker.Dataflow
 				return;
 			}
 
+			if (operation.Operand is ArrayType arrayType) {
+				StackSlot slot = new StackSlot (new ArrayValue (new ConstIntValue (arrayType.Rank)));
+				currentStack.Push (slot);
+				return;
+			}
+
 			if (operation.Operand is TypeReference typeReference) {
 				var resolvedReference = typeReference.Resolve ();
 				if (resolvedReference != null) {
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -708,14 +708,10 @@ namespace Mono.Linker.Dataflow
 				return;
 			}
 
-			if (operation.Operand is ArrayType arrayType) {
-				StackSlot slot = new StackSlot (new ArrayValue (new ConstIntValue (arrayType.Rank)));
-				currentStack.Push (slot);
-				return;
-			}
-
 			if (operation.Operand is TypeReference typeReference) {
-				var resolvedReference = typeReference.Resolve ();
+				var resolvedReference = typeReference is ArrayType ?
+					typeReference.Module.ImportReference (typeof (Array))?.Resolve () : typeReference.Resolve ();
+
 				if (resolvedReference != null) {
 					StackSlot slot = new StackSlot (new RuntimeTypeHandleValue (resolvedReference));
 					currentStack.Push (slot);
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1577,11 +1577,8 @@ namespace Mono.Linker.Dataflow
 						// Intentionally ignore - it's not wrong for code to call Type.GetType on non-existing name, the code might expect null/exception back.
 						reflectionContext.RecordHandledPattern ();
 					} else {
+						MarkType (ref reflectionContext, typeRef);
 						MarkTypeForDynamicallyAccessedMembers (ref reflectionContext, foundType, requiredMemberTypes);
-						if (typeRef is GenericInstanceType genericType) {
-							foreach (var arg in genericType.GenericArguments)
-								MarkType (ref reflectionContext, arg);
-						}
 					}
 				} else if (uniqueValue == NullValue.Instance) {
 					// Ignore - probably unreachable path as it would fail at runtime anyway.
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -146,7 +146,7 @@ namespace Mono.Linker.Dataflow
 				return new SystemTypeForGenericParameterValue (inputGenericParameter, _context.Annotations.FlowAnnotations.GetGenericParameterAnnotation (inputGenericParameter));
 			} else {
 				TypeDefinition genericArgumentTypeDef = genericArgument is ArrayType ?
-					genericArgument.Module.ImportReference (typeof(Array))?.Resolve () : genericArgument.Resolve ();
+					genericArgument.Module.ImportReference (typeof (Array))?.Resolve () : genericArgument.Resolve ();
 				if (genericArgumentTypeDef != null) {
 					return new SystemTypeValue (genericArgumentTypeDef);
 				} else {
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -113,7 +113,7 @@ namespace Mono.Linker.Dataflow
 		{
 			ValueNode valueNode;
 			if (argument.Type.Name == "Type") {
-				TypeDefinition referencedType = ((TypeReference) argument.Value).Resolve ();
+				TypeDefinition referencedType = ((TypeReference) argument.Value).ResolveToMainTypeDefinition ();
 				valueNode = referencedType == null ? null : new SystemTypeValue (referencedType);
 			} else if (argument.Type.MetadataType == MetadataType.String) {
 				valueNode = new KnownStringValue ((string) argument.Value);
@@ -145,8 +145,7 @@ namespace Mono.Linker.Dataflow
 				// That said we only use it to perform the dynamically accessed members checks and for that purpose treating it as System.Type is perfectly valid.
 				return new SystemTypeForGenericParameterValue (inputGenericParameter, _context.Annotations.FlowAnnotations.GetGenericParameterAnnotation (inputGenericParameter));
 			} else {
-				TypeDefinition genericArgumentTypeDef = genericArgument is ArrayType ?
-					genericArgument.Module.ImportReference (typeof (Array))?.Resolve () : genericArgument.Resolve ();
+				TypeDefinition genericArgumentTypeDef = genericArgument.ResolveToMainTypeDefinition ();
 				if (genericArgumentTypeDef != null) {
 					return new SystemTypeValue (genericArgumentTypeDef);
 				} else {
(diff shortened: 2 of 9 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1622,29 +1617,43 @@ static bool BindingFlagsAreSupported (BindingFlags bindingFlags)
...
-				case null:
-					var source = reflectionContext.Source;
-					reflectionContext.RecordRecognizedPattern (typeDefinition, () => _markStep.MarkEntireType (typeDefinition, includeBaseTypes: true, new DependencyInfo (DependencyKind.AccessedViaReflection, source), source));
-					break;
+			var foundType = typeReference?.Resolve ();
+			if (foundType == null) {
+				// Intentionally ignore - it's not wrong for code to call Type.GetType on non-existing name, the code might expect null/exception back.
+				reflectionContext.RecordHandledPattern ();
+				return;
+			} else {
```
marek-safar, line 1627: Remove else indentation

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1622,29 +1617,43 @@ static bool BindingFlagsAreSupported (BindingFlags bindingFlags)
...
+						reflectionContext.RecordRecognizedPattern (foundType, () => _markStep.MarkEntireType (foundType, includeBaseTypes: true, new DependencyInfo (DependencyKind.AccessedViaReflection, source), source));
+						break;
+					}
+				}
+			}
+
+			// If this is a generic type, mark all its arguments.
+			if (typeReference is GenericInstanceType genericType) {
+				foreach (var arg in genericType.GenericArguments) {
+					MarkTypeForDynamicallyAccessedMembers (ref reflectionContext, arg, requiredMemberTypes);
```
vitek-karas, line 1656: This is excessive and sort of wrong.

Take for example:
```C#
class MyType
{
    public void UnusedMethod() {}
}

void Test()
{
    Type listType = Type.GetType("System.Collections.Generic.List`1[MyType]");
    MethodInfo addMethod = listType.GetMethod("Add");
}
```

The call to `GetMethod` is annotated with `DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)`. So with your code it would mark `List`1.Add` but also `MyType.UnusedMethod`.

I think the easiest way to fix this would be to simply call `MarkType` on the `TypeReference` and then apply additional marking based on the annotations. `MarkType` on its own should not include more then necessary for the type (on simply cases it will not mark any methods and so on).

For non-generic cases this is already happening, since for example marking a method on the type (due to annotations) will also mark the type (MarkMethod does that). The problem is with generics since by calling `.Resolve` we lose the generic instantiation part.

vitek-karas: In a way this is a very similar problem to https://github.com/mono/linker/issues/1559.
Can you please validate if your change also affects the behavior of arrays as noted in #1559?
Either make sure to note the differences in the other issue, or maybe try to fix both at once - I would expect the fixes to be relatively similar.
Basically we need to pass around `TypeReference` instead of `TypeDefinition`.

vitek-karas (changes requested):

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
mateoatr: Makes sense. I wasn't sure how much to keep of a type that is passed as generic argument in a dynamically accessed member - I changed things so that we use `MarkTypeVisibleToReflection` to mark the type references of these.

Based on the discussion in #1559 I changed `ProcessGenericArgumentDataFlow` so that we mark the element type of the generic argument that is passed. This way, passing `T[]` will result in the linker marking `T`.

vitek-karas: The discussion in #1559 is at best inconclusive. I think we definitely need to mark the array type itself (based on the annotations). Right now we're also marking the element type based on the annotations, which is technically incorrect. I'm starting to agree with @MichalStrehovsky that we can stop doing that - the cases where this would matter should be very few.

Note that the original repro with JSON serializing `T[]` is sort of "wrong" anyway. We don't have an annotation which would work for serializers - it would have to be something like `RecursivePublicProperties`. Once we have that, it could also handle arrays (as in mark the element types recursively). Something like this has been proposed a while ago: https://github.com/mono/linker/issues/1087.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1571,12 +1571,17 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
...
+					TypeDefinition foundType = typeRef?.Resolve ();
 					if (foundType == null) {
 						// Intentionally ignore - it's not wrong for code to call Type.GetType on non-existing name, the code might expect null/exception back.
 						reflectionContext.RecordHandledPattern ();
 					} else {
 						MarkTypeForDynamicallyAccessedMembers (ref reflectionContext, foundType, requiredMemberTypes);
+						if (typeRef is GenericInstanceType genericType) {
+							foreach (var arg in genericType.GenericArguments)
+								MarkType (ref reflectionContext, arg);
+						}
```
vitek-karas, line 1584: This is not necessary and in also not complete. The issue at hand is about generics, but the same problem applies to `T[]` and/or `T*` and so on. All of this is already handled by `MarkType` - which calls `GetOriginalType` which does the recursive walk of the element types and/or generic arguments.

So all we need to do here is call `MarkType(typeRef)`. It should not include "more" as marking a type really only guarantees presence of the type, but none of its members.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1571,12 +1571,17 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
 				} else if (uniqueValue is SystemTypeValue systemTypeValue) {
 					MarkTypeForDynamicallyAccessedMembers (ref reflectionContext, systemTypeValue.TypeRepresented, requiredMemberTypes);
 				} else if (uniqueValue is KnownStringValue knownStringValue) {
-					TypeDefinition foundType = _context.TypeNameResolver.ResolveTypeName (knownStringValue.Contents)?.Resolve ();
+					TypeReference typeRef = _context.TypeNameResolver.ResolveTypeName (knownStringValue.Contents);
+					TypeDefinition foundType = typeRef?.Resolve ();
```
vitek-karas, line 1575: This will be a problem when arrays get here - since for some weird Cecil reason calling `Resolve` on `T[]` returns resolved `T` (it drops the "array" part). As per the discussion in the array issue, we definitely need to mark the array itself at least.

vitek-karas (changes requested):

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -708,6 +708,12 @@ private static void ScanLdtoken (Instruction operation, Stack<StackSlot> current
 				return;
 			}
 
+			if (operation.Operand is ArrayType arrayType) {
+				StackSlot slot = new StackSlot (new ArrayValue (new ConstIntValue (arrayType.Rank)));
```
MichalStrehovsky, line 712: This doesn't look right. It essentially makes `typeof(Foo[])` act like `new Foo[1]`.

We'll need to grab the TypeDefinition of System.Array.

Can you add a test that will fail for the code that is here now?

You can unit test that `typeof(int[]).GetProperty("LongLength")` is a recognized pattern. I don't think the recognition will work now because the expected result of ScanLdtoken is a `RuntimeTypeHandleValue`, not `ArrayValue`.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -144,6 +144,8 @@ ValueNode GetValueNodeFromGenericArgument (TypeReference genericArgument)
 				// Technically this should be a new value node type as it's not a System.Type instance representation, but just the generic parameter
 				// That said we only use it to perform the dynamically accessed members checks and for that purpose treating it as System.Type is perfectly valid.
 				return new SystemTypeForGenericParameterValue (inputGenericParameter, _context.Annotations.FlowAnnotations.GetGenericParameterAnnotation (inputGenericParameter));
+			} else if (genericArgument is ArrayType arrayType) {
+				return new ArrayValue (new ConstIntValue (arrayType.Rank));
```
vitek-karas, line 148: This is also not correct. This if will run if I do something like:
```
void TestMethod<T>() { return new List<T[]>(); }
```

It needs to return a node which represents the `T[]` here. Maybe we should rename the method to something like `GetTypeValueNodeFromGenericArgument` or `GetValueNodeForGenericArgumentType`. The returned value node should represent the `typeof(T[])` in this case.

So it should be a SystemTypeValue for `System.Array`.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1571,13 +1573,23 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
...
 						MarkTypeForDynamicallyAccessedMembers (ref reflectionContext, foundType, requiredMemberTypes);
 					}
+				} else if (uniqueValue is ArrayValue arrayValue) {
+					TypeReference typeReference = _context.TypeNameResolver.ResolveTypeName ("System.Array");
+					TypeDefinition systemArrayType = typeReference?.Resolve ();
+					if (systemArrayType == null) {
+						reflectionContext.RecordHandledPattern ();
+					} else {
+						MarkTypeForDynamicallyAccessedMembers (ref reflectionContext, systemArrayType, requiredMemberTypes);
+					}
```
vitek-karas, line 1592: This should never happen - it would mean that the application code is using array in place of Type. Or that annotations are completely wrong. This should be removed.

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -77,6 +77,9 @@ TypeReference ResolveTypeName (AssemblyDefinition assembly, TypeName typeName)
 				if (elementType == null)
 					return null;
 
+				if (typeName is ArrayTypeName)
+					return new ArrayType (elementType);
+
 				return elementType;
```
vitek-karas, line 83: If we were to really fix this it should work for all potential types - so anything which derives from `HasElementType`:
- array
- multi-dim array
- by-ref types
- pointer types

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
mateoatr: We're now passing the typref to `MarkType`, this shouldn't be a problem anymore.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -709,7 +709,9 @@ private static void ScanLdtoken (Instruction operation, Stack<StackSlot> current
 			}
 
 			if (operation.Operand is TypeReference typeReference) {
-				var resolvedReference = typeReference.Resolve ();
+				var resolvedReference = typeReference is ArrayType ?
+					typeReference.Module.ImportReference (typeof (Array))?.Resolve () : typeReference.Resolve ();
```
vitek-karas, line 713: Can you please introduce a helper method for this? Maybe an extension method on `TypeReference` - something like `ResolveToMainTypeDefinition` ... I can't come up with a better name - maybe @MichalStrehovsky  will know. It should:
* For normal types it does what `.Resolve` does today
* For generics it resolves to the open generic type (which is what `.Resolve` does today as well)
* For arrays it resolves to `System.Array` type def.
* The question is what should happen for other "weird" cases like - multi-dim arrays, pointers, by-refs, ... we should probably look into this some more if it can every happen and what does it mean when it does happen.

Eventually (We should either fix this right away or file an issue) this should handle not just arrays, but pointers, by-refs and so on.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1571,11 +1572,13 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
 				} else if (uniqueValue is SystemTypeValue systemTypeValue) {
 					MarkTypeForDynamicallyAccessedMembers (ref reflectionContext, systemTypeValue.TypeRepresented, requiredMemberTypes);
 				} else if (uniqueValue is KnownStringValue knownStringValue) {
-					TypeDefinition foundType = _context.TypeNameResolver.ResolveTypeName (knownStringValue.Contents)?.Resolve ();
+					TypeReference typeRef = _context.TypeNameResolver.ResolveTypeName (knownStringValue.Contents);
+					TypeDefinition foundType = typeRef?.Resolve ();
```
vitek-karas, line 1576: This should call the new helper as well - if typeRef is an array type, this code would still apply the annotations on the element type and not on `System.Array`.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
mateoatr: I added this as an extension method as you suggested. Note that for multidim arrays it will behave the same way as for unidimensional arrays, since Cecil does not differentiate between the two (both are `ArrayType`). Not sure if we want different "resolve semantics" for pointers or by-refs though.

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -375,5 +375,14 @@ public static bool IsSubclassOf (this TypeReference type, string ns, string name
 
 			return false;
 		}
+
+		public static TypeDefinition ResolveToMainTypeDefinition (this TypeReference type)
```
vitek-karas, line 379: This is probably worth a comment what the method is supposed to do (and when it's useful) - I still don't like the name, but I don't have a better alternative.

vitek-karas (changes requested):
There are more places where we call .Resolve and end up with the wrong thing. Simply search for .Resolve () in the ReflectionMethodBodyScanner.
I wrote tests for the cases which I don't think are covered by this change: https://github.com/mono/linker/pull/1580

vitek-karas (approved):

vitek-karas:
Nice :+1: :shipit:

PR #1578, opened 2020-10-20
Switch to correct test results format

marek-safar:


akoeplinger:
This doesn't seem to run tests anymore?

marek-safar:
Yeah, waiting on Arcade changed to be packaged

Issue #1579, opened 2020-10-21
Microsoft.NET.ILLink package has a reference Mono.Cecil.Pdb which doesn't exist


vitek-karas:
The way we build the linker package is not correct currently. We want:
* Build the product assembly and have a project ref to Mono.Cecil and Mono.Cecil.Pdb from the submodule - and we should ship these 3 binaries in the product (that is in the ILLink.Tasks package)
* Produce a package Microsoft.NET.ILLink which is the "reference" package to build custom steps against
  * This should have only the illink assembly in it (actually probably just the ref assembly, we don't need/want the implementation assembly)
  * It should not have the Mono.Cecil assemblies in it even as ref assemblies - we want to rely on the public Mono.Cecil nuget package for that instead.
  * So it should have a package reference to Mono.Cecil, and nothing else

Currently this is not the case, the package contains just one illink assembly (both implementation and ref) but it has package dependency to Mono.Cecil and Mono.Cecil.Pdb. The latter doesn't exist.
This is because the default behavior of nuget pack is to turn project references to package dependencies.

PR #1585, opened 2020-10-23
Exclude Mono.Cecil.Pdb from package dependencies

vitek-karas:
The goal is to build a package which has the ref assembly in it and a package dependency on Mono.Cecil package (which acts like ref for Mono.Cecil and Mono.Cecil.Pdb).

NuGet automatically converts all ProjectReferences into package dependencies - to avoid that PrivateAssets=all must be added.

Fixes https://github.com/mono/linker/issues/1579


vitek-karas:
The change is a bit hacky, but I could not figure out a better way. We want something which is pretty weird:
* Build against the submodule - and ship this with the ILLink.Task
* Publish ILLink against the official Mono.Cecil package

I verified that the NuGets produces by this change are identical as before with the only change being the removal of package dependency to `Mono.Cecil.Pdb`.

marek-safar (approved):

PR #1587, opened 2020-10-24
Try to resolve type from exported types

mateoatr:
The linker currently fails to resolve forwarded types that were passed as strings since these exist in a module's exported types, which we never look into.

### DIFF
--- a/src/linker/Linker/DocumentationSignatureParser.cs
+++ b/src/linker/Linker/DocumentationSignatureParser.cs
@@ -500,7 +500,7 @@ namespace Mono.Linker
 			Debug.Assert (module != null);
 
 			if (declaringType == null) {
-				var type = module.GetType (name);
+				var type = module.GetType (name) ?? module.ExportedTypes?.FirstOrDefault (et => et.FullName == name)?.Resolve ();
 				if (type != null) {
 					results.Add (type);
 				}
--- a/src/linker/Linker/DocumentationSignatureParser.cs
+++ b/src/linker/Linker/DocumentationSignatureParser.cs
@@ -500,7 +500,7 @@ namespace Mono.Linker
 			Debug.Assert (module != null);
 
 			if (declaringType == null) {
-				var type = module.GetType (name) ?? module.ExportedTypes?.FirstOrDefault (et => et.FullName == name)?.Resolve ();
+				var type = module.ResolveType (name);
 				if (type != null) {
 					results.Add (type);
 				}
--- a/src/linker/Linker/ModuleDefinitionExtensions.cs
+++ b/src/linker/Linker/ModuleDefinitionExtensions.cs
@@ -10,5 +10,25 @@ namespace Mono.Linker
 			return (module.Attributes & ModuleAttributes.ILOnly) == 0 &&
 				(module.Attributes & ModuleAttributes.ILLibrary) != 0;
 		}
+
+		public static TypeDefinition ResolveType (this ModuleDefinition module, string typeFullName)
+		{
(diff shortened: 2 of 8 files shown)


src/linker/Linker/DocumentationSignatureParser.cs
```diff
@@ -500,7 +500,7 @@ static void GetMatchingTypes (ModuleDefinition module, TypeDefinition? declaring
 			Debug.Assert (module != null);
 
 			if (declaringType == null) {
-				var type = module.GetType (name);
+				var type = module.GetType (name) ?? module.ExportedTypes?.FirstOrDefault (et => et.FullName == name)?.Resolve ();
```
vitek-karas, line 503: Did you try constructing a TypeReference and resolving it instead? That feels like a more robust approach.
Also if we do keep this code - we should create a helper share it - since the doc parser below has exactly the same logic.

Specific case - I wonder how this works for nested types...

test/Mono.Linker.Tests.Cases/DynamicDependencies/DynamicDependencyOnForwardedType.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+
+namespace Mono.Linker.Tests.Cases.DynamicDependencies
+{
+	[LogDoesNotContain ("IL2036")]
+	class DynamicDependencyOnForwardedType
```
marek-safar, line 16: Please write a self-contained test instead of relying on framework implementation which can change anytime

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -80,7 +81,8 @@ TypeReference ResolveTypeName (AssemblyDefinition assembly, TypeName typeName)
 				return elementType;
 			}
 
-			return assembly.MainModule.GetType (typeName.ToString ());
+			return assembly.MainModule.GetType (typeName.ToString ()) ??
```
marek-safar, line 84: All this can be avoided when assembly has no exported types

src/linker/Linker/ModuleDefinitionExtensions.cs
```diff
@@ -10,5 +10,25 @@ public static bool IsCrossgened (this ModuleDefinition module)
 			return (module.Attributes & ModuleAttributes.ILOnly) == 0 &&
 				(module.Attributes & ModuleAttributes.ILLibrary) != 0;
 		}
+
+		public static TypeDefinition ResolveType (this ModuleDefinition module, string typeFullName)
```
vitek-karas, line 14: Nit: Add a comment or maybe even an assert that the `typeFullName` must be a simple type name, no generics, arrays and so on.

vitek-karas (approved):

src/linker/Linker/ModuleDefinitionExtensions.cs
mateoatr: Shouldn't `Module.GetType` work fine with either of those?

vitek-karas: `Module.GetType` will only works for type definitions - so for open generics. Generic instantiations work work. Similarly for arrays, pointers and so on. That's OK, the places where we call the new method handle those cases and only calls this for "type defs", I just wanted to make sure that we don't make that mistake in the future.

PR #1590, opened 2020-10-26
[release/5.0] Unresolved attribute fixes

marek-safar:
Backport of #1535

> When we fail to resolve a type we try and recreate it and keep the IL valid rather than fail.  We have a test that ensures an unresolved attribute on an AssemblyDefinition can be recreated.  It started failing with the sync up.   LazyMarkCustomAttributes should not skip any attributes.  This would give inconsistent behavior with references that only have attributes marked.  It was also problematic because LazyMarkCustomAttributes would call HandleUnresolvedType on the type where as ProcessLazyAttributes will call HandleUnresolvedMethod. A call to HandleUnresolvedMethod is necessary for us to recreate valid IL.

> I went ahead and added null checks to other places that were missing them.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -580,7 +580,13 @@ namespace Mono.Linker.Steps
 					if (UnconditionalSuppressMessageAttributeState.TypeRefHasUnconditionalSuppressions (ca.Constructor.DeclaringType))
 						_context.Suppressions.AddSuppression (ca, provider);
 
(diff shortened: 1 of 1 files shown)


agocke:
My read of the SDK schedule here is that we're done -- no further modifications can be made to .NET 5 unless it's ship-blocking. Are you looking at something else?

marek-safar:
Only RTM is locked, servicing is open and I don't think we need a separate branch for that

vitek-karas:
Do we need to reconfigure DART to flow the release/5.0 from mono/linker to the right servicing branch in SDK then?

agocke:
@marek-safar Are we still planning on taking this?

marek-safar:
Approved by tactics

PR #1602, opened 2020-11-02
Exclude Mono.Cecil.Pdb from package dependencies (#1585)

radekdoulik:
The goal is to build a package which has the ref assembly in it and a package dependency on Mono.Cecil package (which acts like ref for Mono.Cecil and Mono.Cecil.Pdb).

NuGet automatically converts all ProjectReferences into package dependencies - to avoid that PrivateAssets=all must be added.


radekdoulik:
This is cherry-pick from master branch, which should help with this build error in projects using the `Microsoft.NET.ILLink/5.0.0-*`  package:
```
error NU1101: Unable to find package Mono.Cecil.Pdb. No packages exist with this id in source(s): ...
```

marek-safar:
5.0 is taking servicing fixed only. Please switch to 6.0 instead

radekdoulik:
It looks to me like servicing fix. Without it we are unable to use newer 5.0 `Microsoft.NET.ILLink` nugets in XA.

Last time I [tried](https://github.com/xamarin/xamarin-android/pull/5239) to use 6.0 it wasn't possible yet (10 days ago). I will try it again. Meanwhile we are stuck with older 5.0.

marek-safar:
We don't treat Microsoft.NET.ILLink as officially released package. If there is a problem with 6.0 let me know.

PR #1606, opened 2020-11-04
Add command line option to disable type checks inlining

marek-safar:
It's needed in quirky linker mode used for libraries build

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2928,6 +2928,9 @@ namespace Mono.Linker.Steps
 					if (type.IsInterface)
 						break;
 
+					if (!_context.IsOptimizationEnabled (CodeOptimizations.UnusedTypeChecks, method.DeclaringType))
+						break;
+
 					if (!Annotations.IsInstantiated (type)) {
 						_pending_isinst_instr.Add ((type, method.Body, instruction));
 						return;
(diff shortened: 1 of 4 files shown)


vitek-karas:
I'm wondering if we should instead introduce something like "assume all types can be instantiated" - or maybe "all public types can be instantiated". For "libraries mode" (which we will need to support officially anyway) this would be the correct behavior as the linker will not see the whole app and thus can't assume it sees all instantiations.

This should in theory cover both the "unused type checks" and "unused interfaces" and maybe some other optimizations as well.

Sk21renee37 (approved):

Sk21renee37:
> It's needed in quirky linker mode used for libraries build

Hey I need some help here please

marek-safar:
> For "libraries mode" (which we will need to support officially anyway) this would be the correct behavior as the linker will not see the whole app and thus can't assume it sees all instantiations.

We had a few different attempts to fix this in the past. I don't think we should block this PR on it as it will be a much bigger change and it has many alternatives. I'll create a new issue for that to discuss the options.

vitek-karas:
OK - I'm fine with that approach then.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2928,6 +2928,9 @@ protected virtual void MarkInstruction (Instruction instruction, MethodDefinitio
 					if (type.IsInterface)
 						break;
 
+					if (!_context.IsOptimizationEnabled (CodeOptimizations.UnusedTypeChecks, method.DeclaringType))
+						break;
+
 					if (!Annotations.IsInstantiated (type)) {
```
vitek-karas, line 2934: Nit: I would merge the two conditions - to make it obvious that it's about the same thing. If we were to add additional processing here, it would be easy to make the mistake of leaving the optimization check as is and hide another functionality behind it.
But it's a very minor thing.

vitek-karas (approved):

PR #1609, opened 2020-11-05
Allow external warnings to use codes >= 6001

### LINKED ISSUE #1596: Custom steps can't use error codes outside [1000, 2000]
Creating an error message for an error code outside of the range of [1000, 2000] results in:

    Unhandled exception. System.ArgumentException: The provided code '8000' does not fall into the error category, which is in the range of 1000 to 2000 (inclusive). (TaskId:134)
        at Mono.Linker.MessageContainer.CreateErrorMessage(String text, Int32 code, String subcategory, Nullable`1 origin) (TaskId:134)
        at Xamarin.Linker.LinkerConfiguration.Report(IList`1 exceptions) (TaskId:134)

Yet the documentation says to not use that range: "The known codes are in the range 1000 to 6000. Custom steps should avoid using this range not to collide with existing or future error and warning codes." (https://github.com/mono/linker/blob/7e27fb84e68f88d7d9082af16740117ff7985e61/docs/error-codes.md#illink-errors-and-warnings)

mateoatr:
Fixes #1596.

/cc @rolfbjarne

### DIFF
--- a/src/linker/Linker/MessageContainer.cs
+++ b/src/linker/Linker/MessageContainer.cs
@@ -67,10 +67,10 @@ namespace Mono.Linker
 		{
 			var callingAssembly = Assembly.GetCallingAssembly ();
 #if DEBUG
-			Debug.Assert (!string.Equals (callingAssembly.GetName ().Name, Constants.ILLink, StringComparison.OrdinalIgnoreCase),
+			Debug.Assert (!(Assembly.GetCallingAssembly () == typeof (MessageContainer).Assembly),
 				"Internal error: 'CreateCustomErrorMessage' is intended to be used by external assemblies only. Use 'CreateErrorMessage' instead.");
 #endif
-			if ((typeof(MessageContainer).Assembly != callingAssembly) && code <= 600)
+			if (code <= 6000)
 				throw new ArgumentException ($"The provided code '{code}' does not fall into the permitted range for external errors. To avoid possible collisions " +
 					"with existing and future {Constants.ILLink} errors, external messages should use codes starting from 6001.");
 
@@ -111,12 +111,11 @@ namespace Mono.Linker
 		/// <returns>Custom MessageContainer of 'Warning' category</returns>
 		public static MessageContainer CreateCustomWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, WarnVersion version, string subcategory = MessageSubCategory.None)
 		{
-			var callingAssembly = Assembly.GetCallingAssembly ();
 #if DEBUG
-			Debug.Assert (!string.Equals (callingAssembly.GetName ().Name, Constants.ILLink, StringComparison.OrdinalIgnoreCase),
+			Debug.Assert (!(Assembly.GetCallingAssembly () == typeof (MessageContainer).Assembly),
 				"Internal error: 'CreateCustomWarningMessage' is intended to be used by external assemblies only. Use 'CreateWarningMessage' instead.");
 #endif
-			if ((typeof (MessageContainer).Assembly != callingAssembly) && code <= 6000)
+			if (code <= 6000)
 				throw new ArgumentException ($"The provided code '{code}' does not fall into the permitted range for external warnings. To avoid possible collisions " +
 					$"with existing and future {Constants.ILLink} warnings, external messages should use codes starting from 6001.");
 
--- a/src/linker/Linker/MessageContainer.cs
+++ b/src/linker/Linker/MessageContainer.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
+using System.Diagnostics;
 using System.Reflection;
 using System.Text;
 using Mono.Cecil;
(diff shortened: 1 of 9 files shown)


src/linker/Linker/MessageContainer.cs
```diff
@@ -46,8 +47,12 @@ namespace Mono.Linker
 		/// <returns>New MessageContainer of 'Error' category</returns>
 		public static MessageContainer CreateErrorMessage (string text, int code, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null)
 		{
-			if (!(code >= 1000 && code <= 2000))
+			if ((Assembly.GetExecutingAssembly () == Assembly.GetCallingAssembly ()) &&
```
vitek-karas, line 50: This will work only as long we keep the LinkContext.LogWarning and similar methods NOT used by custom steps. They're not in the ref assembly, but they're public. Maybe we should make them internal and explicitly comment that we don't want them to be made available to custom steps.

That said those methods are very helpful, so I'm hesitant to do that. Not sure about this yet.

src/linker/Linker/MessageContainer.cs
```diff
@@ -66,8 +71,12 @@ public static MessageContainer CreateErrorMessage (string text, int code, string
 		/// <returns>New MessageContainer of 'Warning' category</returns>
 		public static MessageContainer CreateWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, WarnVersion version, string subcategory = MessageSubCategory.None)
 		{
-			if (!(code > 2000 && code <= 6000))
+			if ((Assembly.GetExecutingAssembly () == Assembly.GetCallingAssembly ()) &&
+				!(code > 2000 && code <= 6000))
 				throw new ArgumentException ($"The provided code '{code}' does not fall into the warning category, which is in the range of 2001 to 6000 (inclusive).");
+			else if ((Assembly.GetExecutingAssembly () != Assembly.GetCallingAssembly ()) && code <= 6000)
```
vitek-karas, line 77: I'm a bit worried about perf here. For start let's make sure we only call GetCallingAssembly once in this method. And GetExecutingAssembly is probably slower then typeof(MessageContainer).Assembly which should have the same effect.

src/linker/Linker/MessageContainer.cs
vitek-karas: I guess if we expose those methods we could also hook them up into this logic - so it's not a blocker.
In that case I would just make them internal and add comment for now - noting that if we do expose them, we would need to correctly wire up the logic so that this keeps working as expected.

vitek-karas (changes requested):

src/linker/Linker/MessageContainer.cs
marek-safar: I think this is over complicated. We should just expose the method differently by having 2 different methods.

```c#
public static MessageContainer CreateErrorMessage (string text, int code, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null); // 1000-2000 range

public static MessageContainer CreateCustomErrorMessage (string text, int code, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null) // exposed in ref with check for 6000+ range

```

vitek-karas: I like that - we could then also make the `CreateErrorMessage` internal (and exclude it from the ref assembly) and make the other one public and include it in ref. We could also add an assert with the code here to validate that the external one is not called from within the linker.

mateoatr: Adding a new method for creating custom messages makes sense (this saves us from any perf impact), although then there's no point on keeping `CreateErrorMessage` in the ref assembly, we could just remove it.

src/linker/Linker/MessageContainer.cs
```diff
@@ -44,14 +46,37 @@ namespace Mono.Linker
...
+		/// <param name="origin">Filename or member where the error is coming from</param>
+		/// <returns>Custom MessageContainer of 'Error' category</returns>
+		public static MessageContainer CreateCustomErrorMessage (string text, int code, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null)
+		{
+			var callingAssembly = Assembly.GetCallingAssembly ();
+#if DEBUG
+			Debug.Assert (!string.Equals (callingAssembly.GetName ().Name, Constants.ILLink, StringComparison.OrdinalIgnoreCase),
+				"Internal error: 'CreateCustomErrorMessage' is intended to be used by external assemblies only. Use 'CreateErrorMessage' instead.");
+#endif
+			if ((typeof(MessageContainer).Assembly != callingAssembly) && code <= 600)
```
rolfbjarne, line 73:
```suggestion
			if ((typeof(MessageContainer).Assembly != callingAssembly) && code <= 6000)
```

vitek-karas, line 73: Why do we need the calling assembly check here. The assert above should make sure that linker itself never calls this, so we can get here only from custom steps. So simply check the code is above 6000 and that should do it.

vitek-karas (approved):

src/linker/Linker/MessageContainer.cs
```diff
@@ -44,14 +46,37 @@ namespace Mono.Linker
...
+		/// </summary>
+		/// <param name="text">Humanly readable message describing the error</param>
+		/// <param name="code">A custom error ID. This code should be greater than or equal to 6001
+		/// to avoid any collisions with existing and future linker errors</param>
+		/// <param name="subcategory">Optionally, further categorize this error</param>
+		/// <param name="origin">Filename or member where the error is coming from</param>
+		/// <returns>Custom MessageContainer of 'Error' category</returns>
+		public static MessageContainer CreateCustomErrorMessage (string text, int code, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null)
+		{
+			var callingAssembly = Assembly.GetCallingAssembly ();
```
marek-safar, line 68: This line is redundant

src/linker/Linker/MessageContainer.cs
```diff
@@ -44,14 +46,37 @@ namespace Mono.Linker
...
+#if DEBUG
+			Debug.Assert (!(Assembly.GetCallingAssembly () == typeof (MessageContainer).Assembly),
+				"Internal error: 'CreateCustomErrorMessage' is intended to be used by external assemblies only. Use 'CreateErrorMessage' instead.");
```
marek-safar, line 71: No need to prefix Assert with `Internal error:`

src/linker/Linker/MessageContainer.cs
```diff
@@ -44,14 +46,37 @@ namespace Mono.Linker
...
+		/// <param name="code">A custom error ID. This code should be greater than or equal to 6001
+		/// to avoid any collisions with existing and future linker errors</param>
+		/// <param name="subcategory">Optionally, further categorize this error</param>
+		/// <param name="origin">Filename or member where the error is coming from</param>
+		/// <returns>Custom MessageContainer of 'Error' category</returns>
+		public static MessageContainer CreateCustomErrorMessage (string text, int code, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null)
+		{
+			var callingAssembly = Assembly.GetCallingAssembly ();
+#if DEBUG
+			Debug.Assert (!(Assembly.GetCallingAssembly () == typeof (MessageContainer).Assembly),
```
marek-safar, line 70: You could just use `!=` instead

src/linker/Linker/MessageContainer.cs
```diff
@@ -64,11 +89,41 @@ public static MessageContainer CreateErrorMessage (string text, int code, string
...
+		/// <param name="subcategory"></param>
+		/// <returns>Custom MessageContainer of 'Warning' category</returns>
+		public static MessageContainer CreateCustomWarningMessage (LinkContext context, string text, int code, MessageOrigin origin, WarnVersion version, string subcategory = MessageSubCategory.None)
+		{
+#if DEBUG
+			Debug.Assert (!(Assembly.GetCallingAssembly () == typeof (MessageContainer).Assembly),
+				"Internal error: 'CreateCustomWarningMessage' is intended to be used by external assemblies only. Use 'CreateWarningMessage' instead.");
+#endif
+			if (code <= 6000)
+				throw new ArgumentException ($"The provided code '{code}' does not fall into the permitted range for external warnings. To avoid possible collisions " +
```
marek-safar, line 119: Should be ArgumentOutOfRangeException

Issue #1164, opened 2020-05-04
Change referenced assemblies processing to be on demand in MarkStep


marek-safar:
For historical reasons ILLinker assembly resolution logic was very explicit and done as a single step in the pipeline `LoadReferencesStep` this was easy to implement but it does not work well with the more advanced features which are have been and will keep be adding to the linker.

There are several drawbacks which would be addressed if the linker was able to resolve the assembly reference on demand during MarkStep (support for on-demand assembly resolution in any step is not required at this point) as well as potential speed improvement for loading and analyzing fewer assemblies. It'd also get ILLinker closer to csc behaviour and require only necessary references for processing to be specified and not all possible references as it's required today.

Related issues which would be addressed by this
* #943
* #843
* #1079

sbomer:
I plan to break the fix up into the following phases:

- [ ]  Phase 1
  - Lazily TypeMap on assembly load. Allows linking lazy assemblies
  - Lazy load assemblies from reflection, DynamicDependency, DynamicallyAccessedMembers
    Will fix recursive DynamicDependency
    Fixes https://github.com/mono/linker/issues/943 (lazy load reflection assemblies)
    Fixes https://github.com/mono/linker/issues/1079 (lazy load DynamicallyAccessedMembers assemblies)

  XML still not processed lazily, only from static reference closure
  - Doesn't solve problem where XML is not processed from DynamicDependency attribute
    (https://github.com/mono/linker/issues/843)
  - Inlining still happens entirely before markstep, by getting inline candidates from static reference closure

- [ ] Phase 2
  - [ ] 2.1
    - Eliminate global pre-pass for DynamicDependency
      Will break inlining into DynamicDependency assemblies
    - Support lazily processing XML for additional dependencies
      (But not substitutions - as those could potentially modify already-marked IL)
    Fixes (https://github.com/mono/linker/issues/843)

  - [ ] 2.2
    - Support inlining into lazily loaded assemblies
      Gives back inlining behavior for DynamicDependency assemblies,
      also for DynamicallyAccessedMembers and reflection

  Check in 2.1 and 2.2 together to prevent breaking change

- [ ] Phase 3 (optional)
  - Support substitutions XML from lazy assemblies
    Limit to substitutions within the assembly (not its dependencies, which may have already been marked)
    Will make substitutions XML behavior more consistent with dependency XML

vitek-karas:
Looks good!

> (But not substitutions - as those could potentially modify already-marked IL)

[...]

PR #1611, opened 2020-11-06
Enable processing of dynamically referenced assemblies

### LINKED ISSUE #943: Reflection patterns can not resolve assemblies outside of the reference closure
When a supported reflection pattern references an assembly that is part of the linker input set, but not in the reference closure of any roots (and thus not loaded into the LinkContext assemblies by LoadReferencesStep or any other step that comes before MarkStep), the pattern will fail to resolve. Since simple reflection patterns are often used for activation without a build-time dependency, this seems like something worth supporting. See https://github.com/dotnet/sdk/issues/4017 for the original issue with a repro.

https://github.com/mono/linker/issues/843 is a similar issue where `PreserveDependency` doesn't cause xml roots from the referenced assembly to be processed.

If we fix this, we should be able to undo https://github.com/dotnet/runtime/pull/32023.

### LINKED ISSUE #1079: [feature/dataflow] Add a pre-pass to scan type name strings in the app
Due to structure of the step architecture in linker, steps that might optionally pull in new assembly dependencies into the closure currently need to run a global pre-pass over the entire app before we reach `MarkStep`. This architectural deficiency will have to be fixed at some point, but before we do that, dataflow will also need to run such inefficient global prepass to locate all strings stored into locations marked with `DynamicallyAccessedMemebers` so that we can potentially add new assemblies into the closure.

We should do whatever is the fastest thing. We could potentially do something as dumb as going over all user strings in the assembly, looking for ones that look like a fully qualified type name (have a comma in it and maybe some other heuristics to trim the list of looked at strings down) and try to open the associated assembly, so that we don't need to actually parse IL.

Plus we need to scan custom attributes if their constructor/setter/field is marked DynamicallyAccessed.

sbomer:
This introduces the ability to add assemblies to LinkContext lazily and have them
processed in MarkStep. For now, the only additional processing is to
load references and run the TypeMap logic. Later, embedded XML processing and
constant propagation will also happen for lazy assemblies.

This is Phase 1 outlined in https://github.com/mono/linker/issues/1164#issuecomment-714732178.

Fixes #943
Fixes #1079

This still needs a bit of work to handle a case where a dynamic assembly is resolved but not marked (for example when using `GetType` for an assembly-qualified type where the assembly exists, but the type doesn't). In this case, SweepStep can make some false assumptions.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -807,6 +807,7 @@ namespace Mono.Linker.Dataflow
 									// Intentionally ignore - it's not wrong for code to call Type.GetType on non-existing name, the code might expect null/exception back.
 									reflectionContext.RecordHandledPattern ();
 								} else {
+									_context.ProcessReferenceClosure (foundType.Module.Assembly);
 									reflectionContext.RecordRecognizedPattern (foundType, () => _markStep.MarkTypeVisibleToReflection (foundTypeRef, new DependencyInfo (DependencyKind.AccessedViaReflection, callingMethodDefinition), callingMethodDefinition));
 									methodReturnValue = MergePointValue.MergeValues (methodReturnValue, new SystemTypeValue (foundType));
 								}
@@ -1339,6 +1340,7 @@ namespace Mono.Linker.Dataflow
 								reflectionContext.RecordUnrecognizedPattern (2061, $"The assembly name '{assemblyNameStringValue.Contents}' passed to method '{calledMethod.GetDisplayName ()}' references assembly which is not available.");
 								continue;
 							}
+							_context.ProcessReferenceClosure (resolvedAssembly);
 
 							var typeRef = _context.TypeNameResolver.ResolveTypeName (resolvedAssembly, typeNameStringValue.Contents);
 							var resolvedType = typeRef?.Resolve ();
@@ -1579,6 +1581,7 @@ namespace Mono.Linker.Dataflow
 						// Intentionally ignore - it's not wrong for code to call Type.GetType on non-existing name, the code might expect null/exception back.
 						reflectionContext.RecordHandledPattern ();
 					} else {
+						_context.ProcessReferenceClosure (foundType.Module.Assembly);
 						MarkType (ref reflectionContext, typeRef);
 						MarkTypeForDynamicallyAccessedMembers (ref reflectionContext, foundType, requiredMemberTypes);
 					}
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1335,14 +1335,17 @@ namespace Mono.Linker.Dataflow
 				if (assemblyNameValue is KnownStringValue assemblyNameStringValue) {
 					foreach (var typeNameValue in methodParams[methodParamsOffset + 1].UniqueValues ()) {
 						if (typeNameValue is KnownStringValue typeNameStringValue) {
-							var resolvedAssembly = _context.GetLoadedAssembly (assemblyNameStringValue.Contents);
+							AssemblyDefinition resolvedAssembly = null;
+							try {
+								resolvedAssembly = _context.Resolve (assemblyNameStringValue.Contents);
+							} catch (AssemblyResolutionException) { }
 							if (resolvedAssembly == null) {
 								reflectionContext.RecordUnrecognizedPattern (2061, $"The assembly name '{assemblyNameStringValue.Contents}' passed to method '{calledMethod.GetDisplayName ()}' references assembly which is not available.");
 								continue;
 							}
 							_context.ProcessReferenceClosure (resolvedAssembly);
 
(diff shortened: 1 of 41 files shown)


src/linker/Linker.Steps/IAssemblyStep.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Mono.Cecil;
+using System.Collections.Generic;
+
+namespace Mono.Linker.Steps
+{
+	public interface IAssemblyStep
+	{
+		void Initialize (LinkContext context);
+		void ProcessAssemblies (HashSet<AssemblyDefinition> assembly);
```
mateoatr, line 13: Nit:
```suggestion
		void ProcessAssemblies (HashSet<AssemblyDefinition> assemblies);
```

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -13,51 +13,54 @@ public TypeNameResolver (LinkContext context)
 			_context = context;
 		}
 
-		public TypeReference ResolveTypeName (string typeNameString)
+		static TypeName TryParseTypeName (string typeNameString)
```
mateoatr, line 16: Should we change this to:

```suggestion
		static bool TryParseTypeName (string typeNameString, out TypeName typeName)
```

src/linker/Linker.Steps/LoadReferencesStep.cs
```diff
@@ -32,20 +32,42 @@
...
 
-		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		readonly HashSet<AssemblyDefinition> newReferences = new HashSet<AssemblyDefinition> ();
+
+		public void Initialize (LinkContext context)
+		{
+			_context = context;
+		}
+
+		public virtual void ProcessAssemblies (HashSet<AssemblyDefinition> assemblies)
```
marek-safar, line 50: Why does it take HashSet? It seems like huge complication if any step could mutate that set

src/linker/Linker.Steps/LoadReferencesStep.cs
```diff
@@ -32,20 +32,42 @@
 
 namespace Mono.Linker.Steps
 {
-	public class LoadReferencesStep : BaseStep
+	public class LoadReferencesStep : IAssemblyStep
 	{
+		LinkContext _context;
+
+		LinkContext Context => _context;
```
marek-safar, line 39: Why have both?

src/linker/Linker.Steps/IAssemblyStep.cs
```diff
@@ -0,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Mono.Cecil;
+using System.Collections.Generic;
+
+namespace Mono.Linker.Steps
+{
+	public interface IAssemblyStep
```
marek-safar, line 10: Not sure if it should be interface

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -807,6 +807,7 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 									// Intentionally ignore - it's not wrong for code to call Type.GetType on non-existing name, the code might expect null/exception back.
 									reflectionContext.RecordHandledPattern ();
 								} else {
+									_context.ProcessReferenceClosure (foundType.Module.Assembly);
```
marek-safar, line 810: I don't think this is the right approach. It might work in this special case it's not the general solution. You need to hook up into IAssemblyResolver otherwise we would have to hunt every single Resolve call including ones inside Cecil.

test/Mono.Linker.Tests.Cases/Attributes/CoreLibraryAssemblyAttributesAreKept.cs
```diff
@@ -7,6 +7,9 @@ namespace Mono.Linker.Tests.Cases.Attributes
 {
 	[Reference ("System.dll")]
 	[SetupLinkerCoreAction ("link")]
+	// IComponent has TypeConverterAttribute referencing System.dll, which has
+	// unresolved references to Win32 assemblies.
+	[SetupLinkerArgument ("--skip-unresolved", "true")]
```
marek-safar, line 12: I don't understand this. Is ths PR is making the situation worse instead of better?

src/linker/Linker/Pipeline.cs
```diff
@@ -140,6 +150,11 @@ public IStep[] GetSteps ()
 			return _steps.ToArray ();
 		}
 
+		public IAssemblyStep[] GetAssemblySteps ()
```
marek-safar, line 153: It's not great to allocate array on every call for data which are static. You could just have the loop here.

src/linker/Linker.Steps/IAssemblyStep.cs
marek-safar, line 10: Please use a better name, it's not Step as it can run repeatedly and I don't think we need to leak one argument type to concept name

marek-safar (changes requested):

src/linker/Linker.Steps/LoadReferencesStep.cs
sbomer: This is going away with the new approach.

src/linker/Linker.Steps/IAssemblyStep.cs
sbomer: Going away with the new approach.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
sbomer: Going away in the new approach.

sbomer:
Closing in favor of https://github.com/mono/linker/pull/1666

PR #1621, opened 2020-11-10
Fix ArgumentOutOfRange in GetConstructor scanning

eerhardt:
With https://github.com/dotnet/runtime/issues/42753, we are adding a new overload to Type.GetConstructor that only takes BindingFlags and Type[].

However, this new overload exposes a bug in the ILLinker where it is using the wrong parameter count to find how many constructor parameters are being used.

Fixing the issue by using the correct parameter count in the switch statement.

@vitek-karas @MichalStrehovsky @marek-safar

### DIFF
--- a/test/Mono.Linker.Tests.Cases/DataFlow/EmptyArrayIntrinsicsDataFlow.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/EmptyArrayIntrinsicsDataFlow.cs
@@ -17,6 +17,7 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 			TestGetPublicParameterlessConstructorWithEmptyTypes ();
 			TestGetPublicParameterlessConstructorWithArrayEmpty ();
 			TestGetPublicParameterlessConstructorWithUnknownArray ();
+			TestGetConstructorOverloads ();
 		}
 
 		[UnrecognizedReflectionAccessPattern (typeof (Type), nameof (Type.GetMethod), new Type[] { typeof (string) }, messageCode: "IL2080")]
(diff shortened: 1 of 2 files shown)


test/Mono.Linker.Tests.Cases/DataFlow/EmptyArrayIntrinsicsDataFlow.cs
```diff
@@ -39,6 +40,14 @@ static void TestGetPublicParameterlessConstructorWithUnknownArray ()
 			s_typeWithKeptPublicParameterlessConstructor.GetConstructor (s_localEmptyArrayInvisibleToAnalysis);
 		}
 
+		[UnrecognizedReflectionAccessPattern (typeof (Type), nameof (Type.GetMethod), new Type[] { typeof (string) }, messageCode: "IL2080")]
+		static void TestGetConstructorOverloads ()
+		{
+			s_typeWithKeptPublicParameterlessConstructor.GetConstructor (BindingFlags.Public, null, Type.EmptyTypes, null);
+			s_typeWithKeptPublicParameterlessConstructor.GetConstructor (BindingFlags.Public, null, CallingConventions.Any, Type.EmptyTypes, null);
```
vitek-karas, line 47: If these are new overloads the test is likely to fail on the monolinker.
These should be under `#if NETCOREAPP` - similar to https://github.com/mono/linker/blob/1ce464d7c9478fda885ec52c05ac6a7602a6274b/test/Mono.Linker.Tests.Cases/Reflection/MethodUsedViaReflection.cs#L90-L117

eerhardt: This test is for the existing APIs. It is testing the switch statement for overloads with `4` and `5` parameters, which were in .NET Framework 1.1.

I can't add the new overload in `dotnet/runtime` because of this ILLinker issue. It is causing my build to fail when I introduce a new `GetConstructor` overload that only takes `2` parameters because the code is throwing ArgumentOutOfRangeException.

(The reason the new API is being used is because we currently have some internal extension methods on `Type` that only take the 2 arguments. So the existing code is binding to the new API and not to the extension methods.)

That being said, once I do add the new APIs to `net6.0`, we can make another PR into `mono/linker` to support and test for the new overload.

vitek-karas: You're right - sorry I missed that.

vitek-karas (approved):

PR #1622, opened 2020-11-10
Roslyn analyzer improvements

tlakollo:
Some Roslyn analyzer improvements
-Support for Expected Warning
-Fix name mismatch on Properties
-Fix name mismatch on Generic arguments
-Support for URL named argument


src/ILLink.RoslynAnalyzer/xlf/Resources.tr.xlf
```diff
@@ -3,8 +3,8 @@
   <file datatype="xml" source-language="en" target-language="tr" original="../Resources.resx">
     <body>
       <trans-unit id="RequiresUnreferencedCodeAnalyzerMessage">
-        <source>Calling '{0}' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. {1}.</source>
-        <target state="new">Calling '{0}' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. {1}.</target>
```
Sk21renee37, line 7: New

Sk21renee37 (approved):

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -79,6 +79,12 @@ public override void Initialize (AnalysisContext context)
 					Location location)
 				{
 					var attributes = method.GetAttributes ();
+					string methodName;
+					if (method.MethodKind == MethodKind.PropertyGet || method.MethodKind == MethodKind.PropertySet)
```
agocke, line 83: This is a really hot path -- this code is going to get run for basically every single method call in the program.

We should follow some guidelines:

1. Only do work if it's a method we're interested in (has the right attributes)
2. Avoid allocation as much as possible. That generally means no LINQ.

I'd move this entire block into the `foreach` loop, and let'se see if we can minimize the string concatenation, maybe by using a `StringBuilder`

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -60,6 +62,20 @@ internal static void RunTest (MethodDeclarationSyntax m, List<AttributeSyntax> a
...
 			foreach (var attr in attrs) {
 				switch (attr.Name.ToString ()) {
+				case "ExpectedWarning":
+					Assert.True (
+						filtered.Any (mc => {
+							foreach (var attributeArgument in attr.ArgumentList!.Arguments) {
+								if (GetStringFromExpr (attributeArgument.Expression).StartsWith ("IL"))
+									continue;
+								if (!mc.Contains (GetStringFromExpr (attributeArgument.Expression)))
+									return false;
```
vitek-karas, line 72: I know this is tests, but this kind of thing is actually pretty slow. `GetStringFromExpr(...)` can be moved outside of all the loops, that should help a lot.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -85,11 +85,33 @@ public override void Initialize (AnalysisContext context)
 							IsNamedType (attrClass, "System.Diagnostics.CodeAnalysis.RequiresUnreferencedCodeAttribute") &&
 							attr.ConstructorArguments.Length == 1 &&
 							attr.ConstructorArguments[0] is { Type: { SpecialType: SpecialType.System_String } } ctorArg) {
+
+							StringBuilder methodName = new StringBuilder ();
+							if (method.MethodKind == MethodKind.PropertyGet || method.MethodKind == MethodKind.PropertySet) {
+								methodName.Append (method.ContainingType + "." + method.Name + "(");
+								bool shouldPrependComma = false;
```
vitek-karas, line 92: I assume this is to match the reporting done by the linker?
If that is the case we might want to consider slightly different solution - either change linker to report property getters/setters similarly to Roslyn, or adapt in the test verification.
Roslyn in general has better warning messages and better syntax to refer to code - "better" means, users I familiar with how Roslyn does it. We've already made changes in the linker to adapt to C#/Roslyn like type formatting and such, if there are gaps in that, we should change linker some more.

agocke: Seems reasonable -- I didn't suggest this because I didn't know how comfortable we were with changing tests.

In that case the standard C# format is different in two ways here:

1. C# formats property accessors by saying the name of the property and then the name of the accessor, so instead of `get_MyProperty()` you'd have `MyProperty.set`.
2. C# uses type aliases by default instead of system type names, so `System.Int32` is `int`.

This is the default behavior of the C# analyzer framework -- VB has VB-specific formatting. It's up to you if you want to standardize on C# formatting, or try to stick to IL formatting. I would agree that C# is probably the most common format users see.

vitek-karas: @mateoatr worked on making the linker formatting "better" so he will know the details of what is already done. I don't have a problem with either approach honestly, I just don't think that the Roslyn analyzer should adapt to linker. We should either keep them different (and compensate in tests) or adapt the linker to Roslyn.

There were discussions what is better for linker especially around generics, since technically linker can't reliably provide the same output as Roslyn (unresolved generic type references don't have enough info in IL to print out the same string). But in the end I think we still keep the C#-like formatting since that's what most people are used to. The alternative was to use the Reflection syntax, but that is sometimes rather weird.

As for the two examples above, I personally don't like the Roslyn way of describing properties (it doesn't match the actual code syntax in any way), but either format would work I think. Just as mentioned above, Roslyn analyzer should be consistent with other Roslyn analyzers, linker can be somewhat different. And for using the type aliases, I think @mateoatr did add a couple of them, so we can either extend that list, or again, keep linker different.

mateoatr: IIRC we don't do any type aliasing, but I think it should be fairly simple to create such a list and use it in `GetDisplayName`. Before having the `GetDisplayName` extension for member references, we used to report the source location of a warning by simply using the `FullName` property of Cecil, which was pretty verbose (the tests [here](https://github.com/mono/linker/blob/master/test/Mono.Linker.Tests/Tests/GetDisplayNameTests.cs) give a good idea of how the linker formats different member signatures.)

src/ILLink.RoslynAnalyzer/Resources.resx
```diff
@@ -121,6 +121,6 @@
     <value>RequiresUnreferencedCodeAnalyzer</value>
   </data>
   <data name="RequiresUnreferencedCodeAnalyzerMessage" xml:space="preserve">
-    <value>Calling '{0}' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. {1}.</value>
+    <value>Calling '{0}' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. {1}. {2}.</value>
```
mateoatr, line 124: The last dot shouldn't be there https://github.com/mono/linker/blob/master/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs#L1289-L1297

test/ILLink.RoslynAnalyzer.Tests/AnalyzerTests.cs
```diff
@@ -23,9 +23,80 @@ class C
...
-			return VerifyCS.VerifyAnalyzerAsync (src,
+			return VerifyCS.VerifyAnalyzerAsync (TestRequiresWithMessageOnlyOnMethod,
 				// (8,17): warning IL2026: Calling 'System.Int32 C::M1()' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. message.
-				VerifyCS.Diagnostic ().WithSpan (8, 17, 8, 21).WithArguments ("C.M1()", "message")
+				VerifyCS.Diagnostic ().WithSpan (8, 17, 8, 21).WithArguments ("C.M1()", "message", "")
+				);
+		}
+
+		[Fact]
+		public Task TestRequiresWithMessageAndUrlOnMethod ()
```
mateoatr, line 33: Why do we need to rewrite these tests here?

vitek-karas (approved):

test/ILLink.RoslynAnalyzer.Tests/AnalyzerTests.cs
tlakollo: We don't, I just believe that it was important to have an individual test on named arguments since it exercises a specific case in the resource file, and something that test setters and getters separated from the other.

Issue #1616, opened 2020-11-09
Incorrect error message IL2070


stephentoub:
I have the following method:
```C#
private ConstructorInfo? GetDefaultConstructor(Type type) =>
    type.IsValueType ? null :
    type.GetConstructor(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly, null, Type.EmptyTypes, null);
```
This is resulting in the following linker error:
```
Trim analysis error IL2070: System.Xml.Serialization.ReflectionXmlSerializationReader.GetDefaultConstructor(Type): The requirements declared via the 'DynamicallyAccessedMembersAttribute' on the parameter 'type' of method 'System.Xml.Serialization.ReflectionXmlSerializationReader.GetDefaultConstructor(Type)' don't match those on the implicit 'this' parameter of method 'System.Type.GetConstructor(BindingFlags,Binder,Type[],ParameterModifier[])'. The source value must declare at least the same requirements as those declared on the target location it's assigned to
```
It's talking about an attribute on the `type` parameter of this method, but... there is no attribute on the type parameter of that method.  If it's intending to say there needs to be one, it should say so, as otherwise it's misleading and confusing referring to an attribute that doesn't exist.

mateoatr:
> it's misleading and confusing referring to an attribute that doesn't exist

Agreed. Would `The parameter 'type' of method 'System.Xml.Serialization.ReflectionXmlSerializationReader.GetDefaultConstructor(Type)' must be annotated with the 'DynamicallyAccessedMembersAttribute' matching the access requirements of the implicit 'this' parameter of method 'System.Type.GetConstructor(BindingFlags,Binder,Type[],ParameterModifier[])'. The source value must declare at least the same requirements as those declared on the target location it's assigned to.`
 make sense here? Asking before updating all places in ReflectionMethodBodyScanner where we have similar warning messages.

/cc @sbomer

sbomer:
I like that a lot better, but I think we should state it in terms of what's wrong, not how to fix it (this seems more consistent with the C# error messages). How about "is not annotated with" instead of "must be annotated with"? I think the second sentence "must declare at least..." is OK because it's clarifying the first part, and I can't come up with a clearer way to say it.

[...]

PR #1624, opened 2020-11-11
Update IL2067-91 warning messages

mateoatr:
See #1616. This should help us output a more accurate warning in cases where the linker thinks that something needs to be annotated with `DynamicallyAccessedMembers`.


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1621,6 +1547,45 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
 			reflectionContext.RecordHandledPattern ();
 		}
 
+		static string GetDidNotMatchRequirementsWarningMessage (IMetadataTokenProvider source, IMetadataTokenProvider target)
```
vitek-karas, line 1550: We actually had this approach before - but intentionally switched to the current one where the message is not calculated. The reason is localization and resources. Calculated messages are VERY hard to localize. The medium term (hopefully 6.0) plan is to move all the warning/error messages to resources - in which case the "calculation" approach will not work very well.
I know it is tedious to fix 40-something message by hand, but it's what we want long-term.

vitek-karas (changes requested):

vitek-karas:
Ideally @samsp-msft  would take a look at the new messages (but we haven't done official review for 5.0 unfortunately).

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
mateoatr: That makes sense - I was curious why we had that many hard-written messages with an easy to predict structure. I'll change this back.

vitek-karas: Since you tripped on this - can you please add a comment with the reasoning?

agocke: Would this be a good time to start moving the message format to a resource? Since we're changing each of the locations anyway?

vitek-karas: That would be nice - but I don't know how complex it will be to setup resources for linker (it needs to build for mono as well). If it's reasonably easy, we could start migrating to resources for sure.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1621,6 +1522,31 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
 			reflectionContext.RecordHandledPattern ();
 		}
 
+		static string[] GetDidNotMatchRequirementsWarningMessageArgs (IMetadataTokenProvider source, IMetadataTokenProvider target)
+		{
+			List<string> args = new List<string> ();
```
vitek-karas, line 1527: I would prefer to keep the code spread around above. It will make it easier to customize the message if we wanted to. This way the above switch could really just determine the warning code and the warning and message construction could be unified. I know it goes a bit against code deduplication, but the point here is to make it easy to improve UX going forward.

test/Mono.Linker.Tests/TestCases/Dependencies/SortedWarnings.txt
```diff
@@ -1,10 +1,10 @@
...
-ILLink: Trim analysis warning IL2075: Mono.Linker.Tests.Cases.Warnings.Individual.WarningsAreSorted.B.Z(): The requirements declared via the 'DynamicallyAccessedMembersAttribute' on the return value of method 'System.Type.GetType(String)' don't match those on the implicit 'this' parameter of method 'System.Type.GetMethod(String)'. The source value must declare at least the same requirements as those declared on the target location it's assigned to
\ No newline at end of file
+﻿ILLink: Trim analysis warning IL2072: Mono.Linker.Tests.Cases.Warnings.Dependencies.TriggerWarnings_Lib.NestedType.Warning3(): '#0' parameter does not satisfy 'DynamicallyAccessedMembersAttribute' in call to 'System.Linq.Expressions.MethodCallExpression System.Linq.Expressions.Expression::Call(System.Type,System.String,System.Type[],System.Linq.Expressions.Expression[])'. The return value of method 'Mono.Linker.Tests.Cases.Warnings.Dependencies.TriggerWarnings_Lib.TriggerUnrecognizedPattern()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
```
vitek-karas, line 1: I think the wording is "Reversed". The parameter is NOT the one to blame, the parameter declares requirements. The return value is the thing which needs to be "Fixed" (typically).
So either the wording should be "argument" as that denotes the value passed to the parameter, or we should change it around somehow.

vitek-karas (approved):

src/linker/Resources/Strings.resx
```diff
@@ -0,0 +1,195 @@
...
+  </resheader>
+  <data name="IL2067" xml:space="preserve">
+    <value>'{0}' argument does not satisfy 'DynamicallyAccessedMembersAttribute' in call to '{1}'. The parameter '{2}' of method '{3}' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.</value>
+  </data>
+  <data name="IL2068" xml:space="preserve">
+    <value>'{0}' method return value does not satisfy 'DynamicallyAccessedMembersAttribute' requirements. The parameter '{1}' of method '{2}' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.</value>
+  </data>
+  <data name="IL2069" xml:space="preserve">
+    <value>'{0}' field does not satisfy 'DynamicallyAccessedMembersAttribute' requirements. The parameter '{1}' of method '{2}' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.</value>
```
sbomer, line 127:
```suggestion
    <value>value stored in field '{0}' does not satisfy 'DynamicallyAccessedMembersAttribute' requirements. The parameter '{1}' of method '{2}' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.</value>
```
(same for IL2074/79/84/89 below). It should be clear that this is about the value, not annotations on the field itself. Same goes for

sbomer (commented):
Thanks!

It would also be nice for the warnings to include the specific `DynamicallyAccessedMemberTypes` that were not satisfied. For example:

`'{0}' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicFields' in call to '{1}'.`

If there are multiple unsatisfied MemberTypes, I think it should be fine to list them:

`...does not satisfy 'DynamicallyAccessedMemberTypes.PublicFields', 'DynamicallyAccessedMemberTypes.NonPublicFields'...`

We should special-case `DynamicallyAccessedMemberTypes.All`.

What do you think? @mateoatr @vitek-karas

mateoatr:
I don't see any reason not to do that; it would definitely make things easier for the user.

Sk21renee37 (approved):

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1392,175 +1394,179 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
 					reflectionContext.RecordHandledPattern ();
 				} else if (uniqueValue is LeafValueWithDynamicallyAccessedMemberNode valueWithDynamicallyAccessedMember) {
 					if (!valueWithDynamicallyAccessedMember.DynamicallyAccessedMemberTypes.HasFlag (requiredMemberTypes)) {
+						string missingMemberTypes = "''DynamicallyAccessedMembersAttribute''";
+						if (requiredMemberTypes != DynamicallyAccessedMemberTypes.All) {
+							var missingMemberTypesList = Enum.GetValues (typeof (DynamicallyAccessedMemberTypes))
+								.Cast<DynamicallyAccessedMemberTypes> ()
+								.Where (damt => ((requiredMemberTypes ^ valueWithDynamicallyAccessedMember.DynamicallyAccessedMemberTypes) & damt) == damt && damt != DynamicallyAccessedMemberTypes.None)
```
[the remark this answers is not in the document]
mateoatr: I think we are safe from this because of the `& damt`.

sbomer (approved):
This is great, thank you!

Issue #1269, opened 2020-06-12
Add compatibility attribute when linker is used as analyzer tool


marek-safar:
This is a proposal to add global assembly attribute for assemblies which linker processes as an analyzer. The attribute could be something like `ILLinkCompatibility ("{mode}", generated: true)` which would indicate that during the assembly processing illinker didn't find any problems and are safe to be linked. The rules what it means exactly would need to be defined but it should include things like no warnings from flow-analysis, no usage of linker incompatible APIs, etc. The linker would not add this attribute if added manually by the user.

The mode value would reflect the highest linker compatibility level and would make to linker actions, specifically

- link
- copy
- copyused

The usage scenario would then be for any NuGet packages to be automatically linker enabled without doing any manual work on packages or references level.

eerhardt:
I considered opening a new issue for the write-up below, but I felt this proposal is close enough to this discussion, that I decided to add it to this issue:

### Problem

With Blazor WASM in .NET 5 we shipped an untrimmed `System.IO.Pipelines.dll` assembly in the default Blazor app. See https://github.com/dotnet/aspnetcore/issues/27782. In 3.2, Blazor had the rule any assembly whose name starts with `System.` will get trimmed. Moving to the official .NET SDK in 5.0 changed this logic. By default, the .NET SDK will only trim assemblies that are shipped in the `Microsoft.NETCore.App` "runtimepack" (or think "Shared Framework"). There are a number of `System.` assemblies that aren't packaged into the `Microsoft.NETCore.App` "runtimepack", but instead dotnet/runtime only ships them as a NuGet packages. For example:

* System.CodeDom (in WindowsDesktop shared fx)
* System.Composition.*
* System.Configuration.ConfigurationManager (in WindowsDesktop shared fx)
* System.Diagnostics.EventLog (in ASP.NET shared fx)
* System.Diagnostics.PerformanceCounter
* System.DirectoryServices (in WindowsDesktop shared fx)
* System.Drawing.Common (in WindowsDesktop shared fx)
* System.Formats.Cbor
* System.IO.Packaging (in WindowsDesktop shared fx)
* System.IO.Pipelines (in ASP.NET shared fx)
* System.IO.Ports
* System.Management
* System.Net.Connections
* System.Net.Http.WinHttpHandler
* System.Net.WebSockets.WebSocketProtocol
* System.Reflection.MetadataLoadContext
* System.Resources.Extensions (in WindowsDesktop shared fx)
* System.Runtime.Caching
[...]

PR #1626, opened 2020-11-12
Add design for lazy loading

sbomer:
This is a proposal for lazy loading to solve https://github.com/mono/linker/issues/1164 and related issues. This replaces the plan outlined in https://github.com/mono/linker/issues/1164#issuecomment-714732178, but there is some overlap because we still need to process references up-front for the `copy` action, if we want to preserve the current behavior.

I created a new branch for the feature that we can use to avoid breaking the constant propagation behavior in master.

### DIFF
--- a/docs/design/assembly-loading.md
+++ b/docs/design/assembly-loading.md
@@ -1,6 +1,8 @@
 # ILLink Assembly Loading
 
-We would like to support loading and processing assemblies later in the pipeline, so that dynamically referenced assemblies can be correctly processed. A secondary goal is to avoid unnecessary processing of unused assemblies.
+Current versions of the linker do not support loading and processing assemblies that are only referenced dynamically (via reflection, strings in special attributes, or embedded XML), which leads to incorrect outputs that are missing code required by these dependencies. We need a way to load and process assemblies during later steps of the pipeline, to solve these correctness issues.
+
+The linker also does a lot of unnecessary processing because it loads static references of root assemblies up-front. Pipeline steps process all assemblies in the static reference closure, reading embedded XML, propagating constants, and doing branch elimination, whether or not they contain any code that is used at runtime. A secondary goal is to avoid unnecessary processing of such unused assemblies to improve the linker's performance.
 
 ## Assembly processing
 
@@ -12,7 +14,7 @@ Assemblies loaded into the context usually need to have the following processing
 - process the assembly IL for constant methods
 - remove unused branches after propagating constants (possibly from another assembly)
 
-## .NET5 behavior
+## Current behavior
 
 ### Assembly references loaded up-front
 
@@ -26,12 +28,11 @@ XML descriptors from the static reference closure are processed whether or not t
 
 Later steps do not load additional assemblies, with rare exceptions. Dynamic references (from reflection, or from special attributes) are only looked up in already-loaded assemblies, so the linker is unable to find assemblies only referenced dynamically, producing incorrect outputs which are missing the code required by such references. This problem is partially mitigated for `DynamicDependencyAttribute` by a step that pre-scans for assemblies referenced from instances of this attribute, but it does not recursively handle `DynamicDependencyAttribute` instances in these referenced assemblies.
 
-
 ## .NET6 behavior
 
 ### Assembly references loaded lazily
 
-Assemblies will be resolved and loaded lazily, whether by name or from a metadata reference. Unused references will typically not be loaded at all, removing the need to process potentially large amounts of unused code. The result of calling `Resolve` will stay predictable (it will either find a cached result, or look for and possibly load a new assembly). Just resolving an assembly will not trigger any additional processing or recursive assembly resolution.
+Assemblies will be resolved and loaded lazily, whether by name or from a metadata reference. We will avoid loading referenced assemblies up-front, instead relying on various APIs that all ultimately call `IAssemblyResolver.Resolve` to load assemblies as needed. Trying to `Resolve` an assembly name (using `IAssemblyResolver` or a `LinkContext` helper) or a Cecil member reference (`MethodReference`, `ExportedType`, etc...) will continue to have predictable assembly loading behavior - it will either find a cached result, or look for and possibly load a new assembly. Unused references will typically not be loaded at all, removing the need to process potentially large amounts of unused code. Just resolving an assembly will not trigger any additional processing or recursive assembly resolution.
 
 ### Additional processing done lazily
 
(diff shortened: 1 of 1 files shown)


docs/design/assembly-loading.md
```diff
@@ -0,0 +1,83 @@
+# ILLink Assembly Loading
+
+We would like to support loading and processing assemblies later in the pipeline, so that dynamically referenced assemblies can be correctly processed. A secondary goal is to avoid unnecessary processing of unused assemblies.
```
marek-safar, line 3: It'd be nice to improve the wording here. "We would like to" I think it's more than that, we need solution for specific problems. "A secondary goal is to avoid unnecessary processing of unused assemblies" needs reasons, etc.

docs/design/assembly-loading.md
```diff
@@ -0,0 +1,83 @@
...
+
+Assemblies loaded into the context usually need to have the following processing done before or during marking:
+- build the type hierarchy info
+- read descriptor XML and mark its new dependencies
+- read attribute XML and process it for additional/removed attributes
+- read substitutions XML and process it for constant methods
+- process the assembly IL for constant methods
+- remove unused branches after propagating constants (possibly from another assembly)
+
+## .NET5 behavior
```
marek-safar, line 15: This looks like it's .NET5 behaviour only, perhaps change this to "Current behavior"

docs/design/assembly-loading.md
```diff
@@ -0,0 +1,83 @@
...
+### Dynamic assembly references not discovered
+
+Later steps do not load additional assemblies, with rare exceptions. Dynamic references (from reflection, or from special attributes) are only looked up in already-loaded assemblies, so the linker is unable to find assemblies only referenced dynamically, producing incorrect outputs which are missing the code required by such references. This problem is partially mitigated for `DynamicDependencyAttribute` by a step that pre-scans for assemblies referenced from instances of this attribute, but it does not recursively handle `DynamicDependencyAttribute` instances in these referenced assemblies.
+
+
+## .NET6 behavior
+
+### Assembly references loaded lazily
+
+Assemblies will be resolved and loaded lazily, whether by name or from a metadata reference. Unused references will typically not be loaded at all, removing the need to process potentially large amounts of unused code. The result of calling `Resolve` will stay predictable (it will either find a cached result, or look for and possibly load a new assembly). Just resolving an assembly will not trigger any additional processing or recursive assembly resolution.
```
marek-safar, line 34: Which `Resolve` are you referring to?

docs/design/assembly-loading.md
```diff
@@ -0,0 +1,83 @@
...
+XML descriptors from the static reference closure are processed whether or not the embedding assembly is marked.
+
+### Dynamic assembly references not discovered
+
+Later steps do not load additional assemblies, with rare exceptions. Dynamic references (from reflection, or from special attributes) are only looked up in already-loaded assemblies, so the linker is unable to find assemblies only referenced dynamically, producing incorrect outputs which are missing the code required by such references. This problem is partially mitigated for `DynamicDependencyAttribute` by a step that pre-scans for assemblies referenced from instances of this attribute, but it does not recursively handle `DynamicDependencyAttribute` instances in these referenced assemblies.
+
+
+## .NET6 behavior
+
+### Assembly references loaded lazily
```
marek-safar, line 32: I think it would be good to cover what will trigger the assembly reference loading

docs/design/assembly-loading.md
```diff
@@ -0,0 +1,83 @@
...
+
+Additional processing will be done lazily (not triggered by `Resolve`), running on a single assembly at a time.
+
+- Type info base/override methods will be built per assembly as requested
+
+  The overrides tracked for a method may be incomplete, and will be updated as new assemblies are loaded. This is compatible with the current design of `MarkStep` which re-processes virtual methods, but this could be optimized in the future.
+
+- XML from an assembly will be processed only if the assembly is marked
+
+  Embedded XML from statically referenced assemblies may no longer be processed if the embedding assembly is not marked.
```
marek-safar, line 46: This is a breaking change, right? We should outline that

docs/design/assembly-loading.md
```diff
@@ -0,0 +1,83 @@
...
+
+- Constant propagation and branch elimination will will run per marked assembly
```
marek-safar, line 48: double `will will`

docs/design/assembly-loading.md
```diff
@@ -0,0 +1,83 @@
...
+
+To support on-demand processing one assembly at a time, we will temporarily remove support for constant propagation across assembly boundaries. Note that to remove unused branches based on constant callees, we need to know whether callees - possibly in other assemblies - are constant. This would require loading and processing direct assembly references for constants before doing branch elimination. In general case, this could require recursive processing of assembly references for new constants introduced by branch elimination. There are plans to move constant propagation to `MarkStep`, which will add back cross-assembly constant propagation. 
+
+### Exceptions to lazy loading
+
+When the linker default action is `copy`, we will preserve the .NET5 behavior that keeps all statically referenced assemblies (that don't override the default action). For consistency with this behavior, static references of dynamically loaded assemblies will be kept as well. This will be done by pre-loading the reference closure of assemblies if the linker's default action is `copy`.
```
marek-safar, line 62: > When the linker default action is `copy`

What does it mean? linker rarely runs in a single action mode

sbomer: I meant `-c`/`-u`. I'm struggling a bit with how this should work. Maybe it would be ok to remove support for `-c copy`/`-u copy` instead, though we probably need to figure out https://github.com/mono/linker/issues/1269 before making that decision (there was a suggestion to use `copy` as the default action).

docs/design/assembly-loading.md
sbomer: Reorganized this a bit

docs/design/assembly-loading.md
sbomer: Added some more description

docs/design/assembly-loading.md
sbomer: Add more details here

docs/design/assembly-loading.md
sbomer: I think we will always need to load all referenced assemblies, to support opting in to trimming via attributes. Even if we don't do it by passing `-c copy -u copy`, we want everything without an attribute to be kept - so we need to look at references. Maybe we can do it without reading the entire assembly, so that we still get some perf savings from lazy loading.

docs/design/assembly-loading.md
```diff
@@ -0,0 +1,98 @@
...
+
+  Embedded XML descriptors will logically be considered dependencies of the assembly, not global dependencies.
+
+- Constant propagation and branch elimination will run per marked assembly
+
+  This leaves room to optimize constant propagation to run at a more granular level inline with the marking logic.
+
+### Breaking changes
+
+We will restrict the embedded XML so that it may only modify the containing assembly. This is to prevent cases where a (possibly lazily loaded) assembly modifies code in another assembly that we have already processed, breaking assumptions and creating logical inconsistencies. XML passed on the command-line will continue to work as they do today.
```
vitek-karas, line 55: This is correct, but I would still point out that descriptors will be allowed to bring in/mark parts of other assemblies, including referencing new assemblies.
This also adds a restriction on descriptors in that they can only add code, they cannot remove or modify code (now and in the future).

vitek-karas (approved):

PR #1635, opened 2020-11-17
[release/5.0] Mark generic arguments of dynamically accessed types passed as string

mateoatr:
Backport of https://github.com/mono/linker/pull/1566. Since 1566 makes use of `TypeNameResolver`, I had to cherry pick other commits as well:

Allow ReflectionMethodBodyScanner's HandleCall() to skip processing an instruction if its MarkStep's ProcessReflectionDependency() returns true. (#1548)
Correctly handle assembly qualified names in generic parameters (#1546)
Port .NET Native type name parser (#1472)


vitek-karas:
I'm a bit nervous to include the new type name parser in a servicing fix. Would you have some estimate how much of the fix would "not work" if we stick to the old ways of handling "type name" -> "type def"?

mateoatr:
I think we'd be good for the most part, I did some testing with the example provided in #1559 and it seems to work fine. I have a branch with the fix using the old type name parser: https://github.com/mono/linker/compare/release/5.0...mateoatr:markGenArgs

vitek-karas:
I actually like that fix way better for servicing. One comment: https://github.com/mono/linker/compare/release/5.0...mateoatr:markGenArgs#diff-565c4d3a1fda40fe49e36a001479e217a6a484722f6e97ee132c3ddcefdbd585R1572-R1573 the `ResolveFullyQualifiedTypeName` returns `TypeDefinition` so calling `Resolve` on it won't do anything. This also probably means that technically it will not mark arrays correctly as is. Not sure if it's such a big problem for 5 servicing. We might be OK with this. It will "over mark" since it will mark the element type for arrays (as before the fix in other places), but that should not matter too much.

mateoatr:
Agreed. Closing this PR in favor of #1637.

PR #1639, opened 2020-11-19
Warn on every method annotated with RequiresUnreferencedCode

### LINKED ISSUE #1607: RequiresUnreferencedCode does not produce warning if the method is not seen in a call instruction
Linker can mark methods through various means, only one of which is seeing a direct call instruction. Some of other cases:
* Using reflection like `GetMethod` via intrinsic handling
* Through `DynamicallyAccessedMembers` annotation which ends up marking methods sometimes
* Marking of virtual methods on instantiated types
* Marking of interface methods for instantiated types
* several other cases

Currently the linker only recognizes the `RequiresUnreferencedCode` attribute when analyzing method body instructions, then it reports a warning for calls to method with such attribute. If the method with such attribute is marked for any other reason linker will not produce the warning.

But if linker decides to keep the method (that is, the method is marked) and it is also annotated with `RequiresUnreferencedCode` attribute, linker should always report a warning since such method is potentially reachable.

mateoatr:
Fix #1607. This moves the processing of `RequiresUnreferencedCode` to `MarkMethod`, so that we produce a warning on every method which is annotated with this attribute. There are only two cases in which we'll not warn:

1. Annotated methods whose caller is annotated as well:
```C#
static void Foo() { Bar(); } // Calling this method will warn
static void Bar() { } // Calling this from Foo() will not warn
```
2. Annotated override methods: the warning should be produced when marking the inherited method instead.


test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -163,14 +168,61 @@ static void RequiresUnreferencedCodeOnlyThroughReflection ()
...
+		{
+			[RequiresUnreferencedCode ("Message for --VirtualMethodRequiresUnreferencedCode--")]
+			public virtual void VirtualMethodRequiresUnreferencedCode ()
+			{
+			}
+		}
+
+		class TypeWhichOverridesMethod : BaseType
+		{
+			public override void VirtualMethodRequiresUnreferencedCode ()
```
vitek-karas, line 189: This should be currently producing a warning - all overrides must have the same annotations.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -24,6 +26,9 @@ public static void Main ()
 			TestRequiresSuppressesWarningsFromReflectionAnalysis ();
 			TestDuplicateRequiresAttribute ();
 			TestRequiresUnreferencedCodeOnlyThroughReflection ();
+			TestVirtualMethodRequiresUnreferencedCode ();
+			TestStaticCctorRequiresUnreferencedCode ();
+			TestDynamicallyAccessedMembersWithRequiresUnreferencedCode (typeof (BaseType));
```
vitek-karas, line 31: This should use the `DynamicallyAccessedTypeWithRequiresUnreferencedCode` type as the parameter and it should produce a warning on the .ctor right?

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -15,6 +15,8 @@ namespace Mono.Linker.Tests.Cases.RequiresCapability
 	[SkipKeptItemsValidation]
 	public class RequiresUnreferencedCodeCapability
 	{
+		[ExpectedWarning ("IL2026", "Calling 'Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.BaseType.VirtualMethodRequiresUnreferencedCode()' " +
+			"which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. Message for --VirtualMethodRequiresUnreferencedCode--.")]
```
vitek-karas, line 19: This is really weird - the place where the warning is reported....

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2337,11 +2332,36 @@ protected virtual MethodDefinition MarkMethod (MethodReference reference, Depend
...
 
+		protected internal void ProcessRequiresUnreferencedCode (MethodDefinition method, MessageOrigin sourceLocationMember)
+		{
+			// If the caller of a method is already marked with `RequiresUnreferencedCodeAttribute` a new warning should not
+			// be produced for the callee.
+			// Overrides of methods annotated with `RequiresUnreferencedCodeAttribute` should not produce a warning here,
+			// the warning should be put in the virtual method instead.
+			if (sourceLocationMember.MemberDefinition != null &&
+				Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (sourceLocationMember.MemberDefinition) ||
+				Annotations.GetBaseMethods (method) != null)
```
vitek-karas, line 2349: I don't think I understand the last check about the base methods.
If the `method` is in a direct call (can be non-virtual in fact) then the warning should be reported.
I think the interesting question is what happens if I have:
```C#
interface ITest
{
    [RequiresUnreferencedCode]
    void DoDanger();
}

class Test : ITest
{
    [RequiredUnreferencedCode]
    void DoDanger() { }
}

void TestIt()
{
    ITest it = new Test();
    it.DoDanger(); // This should warn, but that is the easy part
}

// There should not be another warning about DoDanger even though there will be a MarkMethod(Test.DoDanger) probably coming from the interface as it's detected to be instantiated.
```

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2294,21 +2292,21 @@ protected virtual bool MarkMethods (TypeDefinition type, in DependencyInfo reaso
...
 
 		protected internal void MarkIndirectlyCalledMethod (MethodDefinition method, in DependencyInfo reason, IMemberDefinition sourceLocationMember)
 		{
-			MarkMethod (method, reason, sourceLocationMember);
+			MarkMethod (method, reason, new MessageOrigin (sourceLocationMember));
 			Annotations.MarkIndirectlyCalledMethod (method);
 		}
 
-		protected virtual MethodDefinition MarkMethod (MethodReference reference, DependencyInfo reason, IMemberDefinition sourceLocationMember)
+		protected virtual MethodDefinition MarkMethod (MethodReference reference, DependencyInfo reason, MessageOrigin sourceLocationMember)
```
vitek-karas, line 2304: I don't expect this to have a big impact, but can you please roughly measure that this doesn't slow linker down?

I would also rename the parameter to something like `origin` or `markOrigin` or `sourceLocation` or... I don't know... it's definitely not `sourceLocationMember`.

src/linker/Linker.Steps/MarkStep.cs
mateoatr: > I don't think I understand the last check about the base methods.

It should prevent us from producing two warnings in cases such as:
```C#
Bar(typeof(DerivedClass));

void Bar([[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type type) {}

class BaseClass
{
	[RequiresUnreferencedCode ("")]
	public virtual void Foo() {}
}

class DerivedClass : BaseClass
{
	[RequiresUnreferencedCode ("")]
	public override void Foo() {}
}
```

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -212,6 +210,8 @@ static void TestStaticCctorRequiresUnreferencedCode ()
 			_ = new StaticCtor ();
 		}
 
+		[ExpectedWarning ("IL2026", "'Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.DynamicallyAccessedTypeWithRequiresUnreferencedCode.RequiresUnreferencedCode()' method " +
+			"has 'RequiresUnreferencedCodeAttribute' which can break functionality when trimming application code. Message for --RequiresUnreferencedCode--.")]
 		public class DynamicallyAccessedTypeWithRequiresUnreferencedCode
```
vitek-karas: I don't understand why the warning should be reported on the type. The type itself does not have a reference - it's not the "usage". The usage in this case is in the Main method. So ideally the warning should be reported there.

Good way to think about this:
* Let's say the type is DangerousFeature and has public .ctor which is marked with RequiresUnreferencedCode - from assembly Dangerous.dll
* My app has a package reference to Dangerous
* In my app I use `Lazy<DangerousFeature>` in my Main (Lazy<T> has annotation on T to preserve public ctor)
* Where should the warning go?
  * With the current code it would be reported on DangerousFeature - but that makes little sense, that thing is annotated as dangerous, so why add a warning to it
  * The place which causes the Dangerous to be included is in the Main, so the warning should be reported there.
  * It also doesn't belong to Lazy - that in itself didn't do anything which would make the app problematic for trimming.

mateoatr: Got it. I was a little confused by a previous comment, I've changed this back.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
mateoatr: I agree - although I'm not sure what to do in cases like this, where a method is kept because of a `DynamicallyAccessedMember` attribute. The method is not being called, we are just marking it in case the user later calls it, so I think maybe this should not warn.

vitek-karas: You're right that we preserve it because it may be needed later - but that's kind of the point. Linker can't figure out when exactly it's going to be used, so instead we have data flow annotations and will "detect" the usage somewhere else. Since we can't tell where it's actually being used, there's no point trying to report it there. Instead we should report it where we know about it - it's not 100% but it's pretty close.

Not warning is not an option - if the method is going to be used for real (linker can't tell), that would break the promise and potentially the app. So we have to play it safe and warn.

I think the linker should warn at the place where it detects the need to preserve the method - in cases of DynamicallyAccessedMember, it's the same place where we would issue warnings about problems with the DynamicallyAccessedMember attribute itself (for example if the type being used is not statically known).

src/linker/Linker.Steps/MarkStep.cs
mateoatr: Overall, the linker takes relatively the same time for trimming a template hello world app (the linker that I build locally is actually 1 sec faster than the one shipped with the SDK, but this is unrelated to this change - 4.5s vs 5.5s). I can see `ProcessMethod` taking 2% more time (12.5% of total linking time, compared to the current 10.5%.) Currently, MarkMethod gets inlined - with this change this is no longer the case.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -163,14 +167,84 @@ static void RequiresUnreferencedCodeOnlyThroughReflection ()
...
+		{
+			[RequiresUnreferencedCode ("Message for --VirtualMethodRequiresUnreferencedCode--")]
+			public virtual void VirtualMethodRequiresUnreferencedCode ()
+			{
+			}
+		}
+
+		class TypeWhichOverridesMethod : BaseType
+		{
+			[ExpectedWarning ("IL2046", "All overridden methods must have 'RequiresUnreferencedCodeAttribute'.")]
```
vitek-karas, line 188: I don't think testing the behavior with only partially annotated hierarchies is necessary. I think it's more important to test the case where the hierarchy is fully annotated (otherwise the user will get warnings anyway).

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -163,14 +167,84 @@ static void RequiresUnreferencedCodeOnlyThroughReflection ()
...
+			public override void VirtualMethodRequiresUnreferencedCode ()
+			{
+			}
+		}
+
+		[ExpectedWarning ("IL2026", "--VirtualMethodRequiresUnreferencedCode--")]
+		static void TestVirtualMethodRequiresUnreferencedCode ()
+		{
+			var tmp = new TypeWhichOverridesMethod ();
+			tmp.VirtualMethodRequiresUnreferencedCode ();
```
vitek-karas, line 198: I would test both cases - calling it on the derived class as well as on the base class.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -163,14 +167,84 @@ static void RequiresUnreferencedCodeOnlyThroughReflection ()
...
+		class TypeWhichOverridesMethod : BaseType
+		{
+			[ExpectedWarning ("IL2046", "All overridden methods must have 'RequiresUnreferencedCodeAttribute'.")]
+			public override void VirtualMethodRequiresUnreferencedCode ()
+			{
+			}
+		}
+
+		[ExpectedWarning ("IL2026", "--VirtualMethodRequiresUnreferencedCode--")]
+		static void TestVirtualMethodRequiresUnreferencedCode ()
```
vitek-karas, line 195: And just like in the interface case, I would validate that we don't get a warning due to the virtual override marking.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: I finally figured out the counter example:
```C#
		abstract class CovariantReturnBase
		{
			[RequiresUnreferencedCode ("Message for --CovariantReturnBase.GetRequiresUnreferencedCode--")]
			public abstract BaseReturnType GetRequiresUnreferencedCode ();
		}

		class CovariantReturnDerived : CovariantReturnBase
		{
			[RequiresUnreferencedCode ("Message for --CovariantReturnBase.GetRequiresUnreferencedCode--")]
			public override DerivedReturnType GetRequiresUnreferencedCode ()
			{
				return null;
			}
		}

		[LogDoesNotContain("--CovariantReturnBase.GetRequiresUnreferencedCode--")]
		[ExpectedWarning ("IL2026", "--CovariantReturnBase.GetRequiresUnreferencedCode--")]
		static void TestCovariantReturnCallOnDerived ()
		{
			var tmp = new CovariantReturnDerived ();
			tmp.GetRequiresUnreferencedCode ();
		}
```

The problem is that the callvirt in this case is not to the base method. The code as written in the MarkStep basically assumes that if an override is marked, it is never a target of a call - it's only because of virtual hierarchy marking. Which is mostly true for C#, since the compiler makes sure it generates callvirt to the base method regardless of which type the method was called on. This might not be true for non-C#, but it's definitely not true with covariant returns as shown above, in this case C# intentionally generates call to the override method.

The above test sample fails because there's no warning generated. Which is weird for multiple reasons, I'm still looking into it.

vitek-karas: Here is a change with the new test and modification to the markStep which makes all of the tests pass: https://github.com/mateoatr/linker/compare/requiresUnreferencedCode...vitek-karas:requiresUnreferencedCode

marek-safar:
@mateoatr what are the next steps here?

mateoatr:
> @mateoatr what are the next steps here?

I'll make a new PR with this changes -- more convenient than trying to rebase. This PR already had the fix for places where we don't warn on direct calls to methods annotated with the attribute. I also need to make sure that we warn on annotated methods that are referenced via reflection and that we don't for methods that are kept because of an assembly being copied but are otherwise not used.

PR #1640, opened 2020-11-19
Fix handling of newlines in warning arguments

sbomer:
Fixes https://github.com/mono/linker/issues/1638

The generated response file has newlines and whitespace like this:
```
--warnaserror "
                    CS8600;
                    CS8601;
                "
```
but the response file parser assumed that each argument was on a single line:
https://github.com/mono/linker/blob/bfab847356063d21eb15e79f2b6c03df5bd6ef3d/src/linker/Linker/Driver.cs#L88

This fixes the parser to handle newlines correctly, and allows more whitespace between the warning codes.

### DIFF
--- a/src/linker/Linker/Driver.cs
+++ b/src/linker/Linker/Driver.cs
@@ -85,8 +85,8 @@ namespace Mono.Linker
 				if (arg.StartsWith ("@")) {
 					try {
 						string responseFileName = arg.Substring (1);
-						IEnumerable<string> responseFileLines = File.ReadLines (responseFileName);
-						ParseResponseFileLines (responseFileLines, result);
+						using (var responseFileText = new StreamReader (responseFileName))
+							ParseResponseFile (responseFileText, result);
 					} catch (Exception e) {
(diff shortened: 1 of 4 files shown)


src/linker/Linker/Driver.cs
```diff
@@ -99,49 +99,54 @@ public static bool ProcessResponseFile (string[] args, out Queue<string> result)
...
 						}
-						if (copyChar) {
-							argBuilder.Append (responseFileText[idx]);
-						}
-						idx++;
+						// treat backslashes before " as escapes
+						numBackslash /= 2;
 					}
-					result.Enqueue (argBuilder.ToString ());
+					argBuilder.Append (new String ('\\', numBackslash));
```
vitek-karas, line 142: Nit: The most common case will be with no backslashes and this allocates empty string every time that happens. We should only do this if `numBackslash > 0`

In fact if I read it correctly, this allocates empty string for every single character on the input.

vitek-karas (approved):

PR #1647, opened 2020-11-25
Add design doc for trimming opt-in/opt-out

sbomer:
This includes some background on what we added for .NET5, and a proposal for more options in .NET6 :
- An assembly-level attribute to opt into trimming
- Simplified trimming opt-in from the SDK
- ~~Simplified "trim all assemblies" flag~~

This is a summary of the discussion in https://github.com/mono/linker/issues/1269 and https://github.com/dotnet/sdk/issues/14642, and the .NET 5 discussion in https://github.com/dotnet/sdk/issues/12035.


sbomer:
@eerhardt PTAL (I can't seem to request your review)

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Other options
+
+.NET 5 introduced a host of additional SDK options that map directly to the underlying illink options. The full list is documented at https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options.
+
+## .NET 6
+
+### `IsTrimmable` `AssemblyMetadataAttribute`
+
+An assembly-level `AssemblyMetadataAttribute` may be placed on an assembly to indicate that it should be trimmed:
```
marek-safar, line 75:
```suggestion
An assembly-level `AssemblyMetadataAttribute` may be placed on an assembly to indicate that it can be trimmed:`

File: docs/design/trimmed-assemblies.md [hunk 2]
```diff
@@ -0,0 +1,183 @@
...
+
+```csharp
+[AssemblyMetadataAttribute("IsTrimmable", "true")]
+```
+
+The behavior is the same as the `IsTrimmable` MSBuild metadata, so that:
+- Assemblies with this attribute are trimmed with the global `TrimMode`
+- Assemblies without this attribute are rooted, and given the `copy` action
+
+The only understood value is `true`. Adding `[AssemblyMetadataAttribute("IsTrimmable", "false")]` will have no effect on the linker's behavior, because unattributed assemblies are assumed not to be trimmable by default. We will issue a warning in this case, to discourage misleading use of the attribute.
```
marek-safar, line 85: > We will issue a warning in this case, to discourage misleading use of the attribute.

Should we allow that to make that symmetric with MSBuild metadata behaviour?

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,183 @@
...
+
+If `IsTrimmable` MSBuild metadata is set for an assembly, this overrides the `IsTrimmable` attribute. This allows a developer to opt an assembly into trimming even if it does not have the attribute, or to disable trimming of an assembly that has the attribute.
+
+Instead of using `IsTrimmable` metadata in the SDK to trim netcoreapp by default, we will ship netcoreapp assemblies and OOB assemblies that are conceptually part of the BCL with `IsTrimmable` `AssemblyMetadataAttribute`, and not set `IsTrimmable` MSBuild metadata in the SDK. The MSBuild metadata will now be used only as a developer opt-in or opt-out.
```
marek-safar, line 89:
```suggestion
Instead of using `IsTrimmable` metadata in the SDK to control trimmable assemblies, we will move to a model where all trimmable SDKs assemblies are built with `[AssemblyMetadataAttribute("IsTrimmable", "true")]`

File: docs/design/trimmed-assemblies.md [hunk 4]
```diff
@@ -0,0 +1,183 @@
...
+.NET 5 introduced a host of additional SDK options that map directly to the underlying illink options. The full list is documented at https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options.
+
+## .NET 6
+
+### `IsTrimmable` `AssemblyMetadataAttribute`
+
+An assembly-level `AssemblyMetadataAttribute` may be placed on an assembly to indicate that it should be trimmed:
+
+```csharp
+[AssemblyMetadataAttribute("IsTrimmable", "true")]
```
marek-safar, line 78: Should we add a note that this attribute always survives trimming?

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,183 @@
...
+
+Instead of using `IsTrimmable` metadata in the SDK to trim netcoreapp by default, we will ship netcoreapp assemblies and OOB assemblies that are conceptually part of the BCL with `IsTrimmable` `AssemblyMetadataAttribute`, and not set `IsTrimmable` MSBuild metadata in the SDK. The MSBuild metadata will now be used only as a developer opt-in or opt-out.
+
+### `TrimmableAssembly`
+
+This ItemGroup contains assembly names that will be opted into trimming via `IsTrimmable` metadata. For simple cases, this provides an easier way to enable trimming of additional assemblies, without requiring a custom MSBuild target.
+
+```xml
+<ItemGroup>
+  <TrimmableAssembly Include="MyAssembly" />
```
marek-safar, line 97: I think we need to allow also mode where for explicit opt-out. Perhaps something like

```xml
	<TrimmableAssembly Exclude="MyAssembly" />
```

mapping to `-p`

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,183 @@
...
+</Target>
+```
+
+### `TrimAllAssemblies`
+
+This property simplifies the process of enabling trimming for all assemblies. It is equivalent to setting `IsTrimmable` to `true` on every assembly that is input to the linker. For example:
+
+```xml
+<PropertyGroup>
+  <TrimAllAssemblies>true</TrimAllAssemblies>
```
marek-safar, line 120: What is MSBuild recommendation? Do we need special property or would

```
	<TrimmableAssembly Include="*" />
```

work?

marek-safar (commented):
Nice write up!

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,183 @@
...
+        BeforeTargets="PrepareForILLink">
+  <ItemGroup>
+    <ManagedAssemblyToLink>
+      <IsTrimmable>true</IsTrimmable>
+    </ManagedAssemblyToLink>
+  </ItemGroup>
+</Target>
+```
+
+## Alternatives considered
```
vitek-karas, line 137: I think we should add a section considering future evolution.
* Very long term we want SDKs to enable full trimming on everything.
  * This would require opt-out mechanisms - we would likely have to support `[AssemblyMetadataAttribute("IsTrimmable", "false")] - or maybe even support specifying link action in the attribute.
* Medium term we may want to "upgrade" the console app trimming default from `copyused` to `link`.
  * This would require per-assembly opt-out - so some ability (in the SDK) to specify link action per-assembly (`TrimMode`)

I think we're fine on all of these, but we should mention these and how we plan to go about them.

docs/design/trimmed-assemblies.md
sbomer: This might make more sense if the defaults change - I mentioned it in the Future section.

docs/design/trimmed-assemblies.md
sbomer: I mentioned this in the Future section. I agree that it would be OK not to have `NonTrimmableAssembly` for now.

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,183 @@
...
+        BeforeTargets="PrepareForILLink">
+  <ItemGroup>
+    <ManagedAssemblyToLink Condition="'%(Filename)' == 'MyAssembly'">
+      <IsTrimmable>true</IsTrimmable>
+    </ManagedAssemblyToLink>
+  </ItemGroup>
+</Target>
+```
+
+### `TrimAllAssemblies`
```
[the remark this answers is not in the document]
sbomer: Moved this to a possible future addition.

docs/design/trimmed-assemblies.md
sbomer: Thanks - I added a Future section. PTAL.

vitek-karas (approved):

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,212 @@
...
+## Background
+
+The SDK publish targets run `ILLink`, which has subtargets that process `ResolvedFileToPublish`. This list gets filtered down to the managed assemblies with `PostProcessAssemblies == true`, which are passed to the linker. Those with `IsTrimmable != true` by default are rooted and linked with the `copy` action, and the rest have the action determined by `TrimMode`.
+
+It is worth reiterating that there are three conditions that influence the behavior:
+1. `PostProcessAssemblies` controls whether the linker will see the assembly at all
+2. `IsTrimmable` controls whether the linker will tree-shake the assembly (if not, it gets rooted, and gets action `copy`)
+3. the action (per-assembly `TrimMode` metadata, or global `TrimMode`) controls the level of tree-shaking
+
+Different SDKs have different defaults for these options. The .NET Core sets `TrimMode` to `copyused`, which does assembly-level trimming, by default, but the Blazor SDK sets it to `link` for more aggressive trimming. Blazor also uses extension points to control which assemblies are trimmed by filtering on the assembly name, to generate custom "type-granularity" roots for some assemblies.
```
eerhardt, line 12:
```suggestion
Different SDKs have different defaults for these options. The .NET Core sets `TrimMode` to `copyused`, which does assembly-level trimming, by default, but the Blazor SDK sets it to `link` for more aggressive trimming. Blazor also uses extension points to control which assemblies are trimmed by filtering on the assembly name, and to generate custom "type-granularity" roots for some assemblies.
```

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,212 @@
...
+
+### `PrepareForILLink`
+There is a new public target `PrepareForILLink` that runs before the `ILLink` target, and provides a convenient place to hook into the pipeline to modify metadata for trimming. SDK components can use this as an extension point via `BeforeTargets` and `AfterTargets`.
+
+The global `TrimMode` may be set any time before `PrepareForILLink` runs, which will set it to a default value if not set previously. 
+
+### `ManagedAssemblyToLink`
+The `PrepareForILLink` has a dependency that creates the ItemGroup `ManagedAssemblyToLink`, which represents the set of assemblies that will be passed to the linker. Custom targets may modify `IsTrimmable` and `TrimMode` metadata on these assemblies before `PrepareForILLink`, which will set the assembly action based on this metadata, or they may modify the metadata after `PrepareForILLink` has run.
+
+It will be illegal to change the items in `ManagedAssemblyToLink`, since this represents the set that needs to be filtered and replaced in the publish output. To change which assemblies are passed to the linker, a different extension point should be used to set `PostProcessAssemblies` metadata.
```
eerhardt, line 33:
```suggestion
It is not possible to change the items in `ManagedAssemblyToLink`, since this represents the set that needs to be filtered and replaced in the publish output. To change which assemblies are passed to the linker, a different extension point should be used to set `PostProcessAssemblies` metadata.
```

Should use the past tense since .NET 5 has shipped.

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,212 @@
...
+
+The only understood value is `True` (case-insensitive). Adding `[assembly: AssemblyMetadata("IsTrimmable", "False")]` will have no effect on the linker's behavior, because unattributed assemblies are assumed not to be trimmable by default. We will issue a warning in this case, to discourage misleading use of the attribute.
+
+The attribute will survive trimming like other assembly-level attributes do.
+
+If `IsTrimmable` MSBuild metadata is set for an assembly, this overrides the `IsTrimmable` attribute. This allows a developer to opt an assembly into trimming even if it does not have the attribute, or to disable trimming of an assembly that has the attribute.
+
+Instead of using `IsTrimmable` metadata in the SDK to control trimmable assemblies, we will move to a model where all trimmable SDK assemblies are built with `[assembly: AssemblyMetadata("IsTrimmable", "True")]`.
+
+### `TrimmableAssembly`
```
eerhardt, line 93: Who do we expect to use this option? Is it necessary to build it now given that we have the `PrepareForILLink` target? It is simply a convenience.

I wonder if this should be moved into `Future evolution` and not proposed for .NET 6.

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,212 @@
...
+could be used instead of
+
+```xml
+<Target Name="ConfigureTrimming"
+        BeforeTargets="PrepareForILLink">
+  <ItemGroup>
+    <ManagedAssemblyToLink>
+      <IsTrimmable>true</IsTrimmable>
+    </ManagedAssemblyToLink>
+  </ItemGroup>
```
eerhardt, line 141: This actually doesn't work as it is written. You need to add 

```xml
      <!-- Pass the app assembly as a root -->
      <TrimmerRootAssembly Include="@(IntermediateAssembly)" />
```

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,212 @@
...
+</Target>
+```
+
+This could be set by default in future SDKs, or it could be set by the developer in SDKs where it is not the default. We prefer not to introduce such a property at the moment, because it makes it too easy to enable this more "dangerous" behavior. We may consider adding this in the future when more of the .NET ecosystem has been made compatible with trimming.
+
+### `AssemblyMetadata("IsTrimmable", "False")`
+
+With more aggressive defaults, it could make sense to support an attribute opt-out via `[assembly: AssemblyMetadata("IsTrimmable", "False")]`. This would provide a way for developers to indicate that their assemblies should not be trimmed.
+
+We should also consider whether such an opt-out should prevent the linker from rewriting the attributed assembly. This could be useful as a way to preserve assemblies that have invariants which would be broken by rewriting, or which contain data that would be removed by the linker even with the `copy` action.
```
eerhardt, line 151: > should prevent the linker from rewriting the attributed assembly

At a minimum, it needs to re-write the references, right? For example, removing `System.Runtime` and adding `System.Private.CoreLib`.

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,212 @@
...
+
+We would also need to decide the precedence betwen `TrimmableAssembly` and `NonTrimmableAssembly`, or issue a warning if an assembly is in both ItemGroups.
+
+## Alternatives considered
+
+### `IsTrimmable` MSBuild metadata vs ItemGroup vs Property
+
+It is not always obvious from the project file which assemblies will be included in the published application. Some assemblies are shipped as part of a framework reference, and others as OOB packages. Additionally, the SDK has its own configuration knobs that control whether files are eligible for post-processing. The final list is only "known" until the publish targets run, which is why the most flexible way to control the set of trimmed assemblies is during `PrepareForILLink`. Per-item metadata is a natural way to do this in MSBuild, an can be used for example to filter on the filenames in `ManagedAssemblyToLink`.
```
eerhardt, line 163:
```suggestion
It is not always obvious from the project file which assemblies will be included in the published application. Some assemblies are shipped as part of a framework reference, and others as OOB packages. Additionally, the SDK has its own configuration knobs that control whether files are eligible for post-processing. The final list is only "known" until the publish targets run, which is why the most flexible way to control the set of trimmed assemblies is during `PrepareForILLink`. Per-item metadata is a natural way to do this in MSBuild, and can be used for example to filter on the filenames in `ManagedAssemblyToLink`.
```

eerhardt (approved):
Looks good, thanks for writing this up.

I only have one comment of real substance - whether we need to implement `@(TrimmableAssembly)` in .NET 6 or if we can push it to future.

docs/design/trimmed-assemblies.md
sbomer: That's true - my concern is that a developer might expect such an attribute to prevent modification. I rewrote this section to clarify what I meant.

docs/design/trimmed-assemblies.md
sbomer: Thanks, I forgot about this...

docs/design/trimmed-assemblies.md
sbomer: It's purely a convenience on top of the MSBuild `IsTrimmable` metadata. I added this in response to https://github.com/mono/linker/issues/1269#issuecomment-731449888:
> I think it's crucial during the transition period to allow developers to easily experiment with enabling trimming for any assembly. It should be as simple as passing/adding msbuild property.

I'll let @marek-safar comment on whether it belongs in .NET6 or future.

marek-safar: We already got reports from .net5 developers asking for an easy way to trim more assemblies. Similarly, we will most likely need it for our own usage (trimming testing frameworks) in dotnet/runtime and messing up with msbuild metadata for this is well, messy. I think we should make this operation convenient in .net6.

docs/design/trimmed-assemblies.md
```diff
@@ -0,0 +1,215 @@
...
+    </ManagedAssemblyToLink>
+  </ItemGroup>
+</Target>
+```
+
+This could be set by default in future SDKs, or it could be set by the developer in SDKs where it is not the default. We prefer not to introduce such a property at the moment, because it makes it too easy to enable this more "dangerous" behavior. We may consider adding this in the future when more of the .NET ecosystem has been made compatible with trimming.
+
+### `AssemblyMetadata("IsTrimmable", "False")`
+
+With more aggressive defaults, it could make sense to support an attribute opt-out via `[assembly: AssemblyMetadata("IsTrimmable", "False")]`. This would provide a way for developers to indicate that their assemblies should not be trimmed.
```
rolfbjarne, line 150: For Xamarin.iOS it's fairly common thing to make the linker as aggressive as possible (to get the app size down), and then figure out how to unbreak everything that broke - by disabling the trimming for specific assemblies (or even give the linker xml definitions of what to preserve).

Scenario:

Customer has a class library project, and multiple app head projects (say one for Xamarin.iOS and Xamarin.TVOS) that reference the class library project. The customer tries to trim all assemblies, but realizes that the linker removes too much from his class library project, and wants to tell the linker to not trim his class library project. From this proposal, it seems that the customer has to modify the project file for each app head project (and any future projects that may end up referencing his class library project) to add the IsTrimmable metadata. Is that correct?

marek-safar: I think this scenario should be sufficiently covered with `TrimmableAssembly`, so the developer would end up with something like this in the csproj file (assuming none of the assemblies have IsTrimmable set).

```xml
<ItemGroup>
  <TrimmableAssembly Include="assembly1" />
  <TrimmableAssembly Include="assembly2" />
  <TrimmableAssembly Include="assembly4" />
<!--  <TrimmableAssembly Include="assembly5" /> cannot be linked -->
  <TrimmableAssembly Include="assembly6" />
</ItemGroup>
```

rolfbjarne: @marek-safar afaict that won't work if the developer sets `<TrimAllAssemblies>true</TrimAllAssemblies>` first.

marek-safar: TrimAllAssemblies is not planned for .NET6, it's under Possible future work section https://github.com/mono/linker/pull/1647/files#diff-9abbbcd880aff3bdefcbefe9d49f810ae6eb372bf40816da53f3e470a8c483bfR116

eerhardt: In the future if we had TrimAllAssemblies, we could add metadata to the TrimmableAssembly item to set Trimmable=false which would opt that assembly out of trimming.

docs/design/trimmed-assemblies.md
eerhardt: Ok, as long as we have user scenarios for it, that’s fine with me.

docs/design/trimmed-assemblies.md
vitek-karas: The UX for this scenario is still not ideal. See https://github.com/dotnet/runtime/issues/45353 which just came in and is basically about a similar problem.

If all of these are NuGet packages it will be possible to inject custom properties/targets from the NuGet into the app's project to mark assemblies as non-trimmable. Ideally it should be possible for the author of the library which has trouble being trimmed to mark itself as non-trimmable. With the current design there are basically only two ways to do this:
* Using NuGet's MSBuild integration to set the IsTrimmable to false for the assembly (using custom targets)
* Using embedded descriptor to force root the entire assembly.

This proposal doesn't help with that scenario in any way really.

I honestly don't think it's in such a bad shape - the possible solution would be to enable `[AssemblyMetadataAttribute("IsTrimmable", "False")]` as described above. But there's basically no real difference between that and rooting the entire assembly via embedded descriptor. It's just not as convenient.

rolfbjarne: @marek-safar there will be a way to enable linking for all assemblies (currently implemented with the `MtouchLink` MSBuild variable), so the point still stands: if linking is enabled for all assemblies, then your solution won't work.

MichalStrehovsky: > For Xamarin.iOS it's fairly common thing to make the linker as aggressive as possible

Do we have actual telemetry for this is this more of a gut feel?

In .NET Native, we played with settings like this a lot and imagined how useful they are (offering various opt-in/opt-out granularities at assembly levels). We wrote blog posts and tools on how to tune the size.

Turned out we lived in a bubble where all of this was clear, easy to use, and useful, but the real world disagreed. Inspecting actual real world apps in the Windows Store we found out only around 3% of all apps in the Store ever touched any of the defaults.

Out of those 3%, pretty much everyone just asked to root more things because we were breaking their app. Only 0.03% of apps opted out of the defaults and asked for more aggressive trimming, even though our testing indicated 70% of apps could do that without seeing any trimming related bugs.

This was when Windows Phone was still a thing and size mattered more.

sbomer: > From this proposal, it seems that the customer has to modify the project file for each app head project (and any future projects that may end up referencing his class library project) to add the IsTrimmable metadata. Is that correct?

@rolfbjarne Just to confirm - that's correct. If aggressive linking is already common, I can see the usefulness of adding `AssemblyMetadataAttribute("IsTrimmable", "False")` in .NET 5, but I have some concerns:

In the scenario you outlined, is the desire to disable trimming only locally for the projects that reference it, or also for other developers? The attribute has the downside that it could be too permanent. If the attributed library ships as a nuget package, another developer who only uses a small portion of it would end up with a larger app. I added some notes along these lines to the doc.

> we could add metadata to the TrimmableAssembly item to set Trimmable=false

@eerhardt I added this suggestion as an alternative under `NonTrimmableAssembly` - but I don't think this would address @rolfbjarne's concern since it still needs to be set in each app's project file.

> Using embedded descriptor to force root the entire assembly.

@vitek-karas with lazy loading, this will only keep the assembly if some part of it is used in a way that's understood by the linker (`IsTrimmable=false` metadata/attribute would keep it regardless). This may or may not be the right behavior, depending on @rolfbjarne's scenario.

rolfbjarne: @MichalStrehovsky 

> > For Xamarin.iOS it's fairly common thing to make the linker as aggressive as possible
> 
> Do we have actual telemetry for this is this more of a gut feel?

I'll try to get some telemetry for this.

@sbomer 

> In the scenario you outlined, is the desire to disable trimming only locally for the projects that reference it, or also for other developers?

Also for other developers.

> The attribute has the downside that it could be too permanent. If the attributed library ships as a nuget package, another developer who only uses a small portion of it would end up with a larger app.

From this proposal it sounded like it would be possible to override the attribute with an MSBuild property, if the consumer of the library wanted to do so.

> > Using embedded descriptor to force root the entire assembly.
> 
> @vitek-karas with lazy loading, this will only keep the assembly if some part of it is used in a way that's understood by the linker (`IsTrimmable=false` metadata/attribute would keep it regardless). This may or may not be the right behavior, depending on @rolfbjarne's scenario.

Yes, this sounds like something that could work from a technical standpoint.

But from a developer perspective it just looks unpolished if I try to use `AssemblyMetadataAttribute("IsTrimmable", "False")` and then get a warning that either tells me to read some documentation that tells me to add a resource file with a specific name and content to accomplish more or less the same thing as what `AssemblyMetadataAttribute("IsTrimmable", "False")` would do.

sbomer: >> The attribute has the downside that it could be too permanent.[...]
>
> From this proposal it sounded like it would be possible to override the attribute with an MSBuild property

True, the MSBuild overrides make any combination possible - I'm just worried about the defaults you get from the attributes, especially as more people start to use aggressive trimming.

I added a description of your scenario to the doc under the future section @rolfbjarne. I'll go ahead and merge this since I think there's agreement about the opt-in, and if there's more data about the opt-out we can revisit.

Issue #1634, opened 2020-11-17
Unresolved generic argument causes System.InvalidOperationException in GetValueNodeFromGenericArgument


Deadpikle:
## Environment

* .NET 5.0.100 AND 5.0.200-preview.20566.9 (tested with both; same error)
* macOS Catalina (10.15.7)

## Command

`dotnet publish "path/project.csproj" --configuration "Release" --framework "net5" --runtime "osx-x64" --output "dotnet-build/" --self-contained True /p:PublishTrimmed=true`

## Error

ILLink : error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues [/Users/myname/Documents/Projects/path/here/project.csproj]
  Fatal error in IL Linker
```
  Unhandled exception. System.InvalidOperationException: Operation is not valid due to the current state of the object.
     at Mono.Linker.Dataflow.ReflectionMethodBodyScanner.GetValueNodeFromGenericArgument(TypeReference genericArgument)
     at Mono.Linker.Dataflow.ReflectionMethodBodyScanner.ProcessGenericArgumentDataFlow(GenericParameter genericParameter, TypeReference genericArgument, IMemberDefinition source)
     at Mono.Linker.Steps.MarkStep.MarkGenericArgumentConstructors(IGenericInstance instance, IMemberDefinition sourceLocationMember)
     at Mono.Linker.Steps.MarkStep.MarkGenericArguments(IGenericInstance instance, IMemberDefinition sourceLocationMember)
     at Mono.Linker.Steps.MarkStep.GetOriginalType(TypeReference type, DependencyInfo reason, IMemberDefinition sourceLocationMember)
     at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, IMemberDefinition sourceLocationMember)
     at Mono.Linker.Steps.MarkStep.MarkField(FieldDefinition field, DependencyInfo& reason)
     at Mono.Linker.Steps.MarkStep.MarkEntireTypeInternal(TypeDefinition type, Boolean includeBaseTypes, DependencyInfo& reason, IMemberDefinition sourceLocationMember, HashSet`1 typesAlreadyVisited)
     at Mono.Linker.Steps.MarkStep.MarkEntireAssembly(AssemblyDefinition assembly)
     at Mono.Linker.Steps.MarkStep.InitializeAssembly(AssemblyDefinition assembly)
     at Mono.Linker.Steps.MarkStep.Initialize()
[...]

PR #1650, opened 2020-11-26
[release/5.0] Fix linker crash on unresolved types in generic and attribute data flow handling

vitek-karas:
Fixes https://github.com/mono/linker/issues/1634
This is a servicing version of https://github.com/mono/linker/pull/1645
Only the necessary changes to make the fix - no cleanup, no better warnings, no new test features.

### Issue
If linker finds an unresolved type used in either attribute parameter, or generic parameter and the parameter has data flow annotations on it, linker will fail and cause the entire dotnet publish to fail as well.
Unresolved types are relatively common for linker and it should be able to handle these without failing.

### Impact
Any application which contains references to types which are not in the application closure may fail to link (depending on where and how the type is used). Missing types or assemblies are relatively common, for example even the core framework has references to assemblies which are not part of it in mscorlib and similar, but it's also relatively common in NuGets, especially if using NuGets which are targeted for .NET Framework on .NET Core.

### Changes
Unresolved types should be handled as "unknown" value nodes. This is already the case when they appear in a method body. But when they're used either as generic arguments or in attribute instantiations linker doesn't handle this case correctly.

For generics linker would fail with an exception. For attributes linker would ignore the type.

Both cases are wrong since linker needs to make sure to issue a warning if it can't guarantee to meet the requirements posed by the data flow annotations. So ignoring unresolved types is also wrong.

The change fixes these behaviors to treat the value as "unknown" which leads into warnings when used somewhere with data flow requirements.

### Risk
Low - the change will avoid throwing exception, so cases which always failed before, should now run through the linker.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -114,7 +114,10 @@ namespace Mono.Linker.Dataflow
 			ValueNode valueNode;
 			if (argument.Type.Name == "Type") {
 				TypeDefinition referencedType = ((TypeReference) argument.Value).ResolveToMainTypeDefinition ();
-				valueNode = referencedType == null ? null : new SystemTypeValue (referencedType);
+				if (referencedType == null)
+					valueNode = UnknownValue.Instance;
+				else
+					valueNode = new SystemTypeValue (referencedType);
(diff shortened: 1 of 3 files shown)


mateoatr (approved):

marek-safar (approved):

marek-safar:
Approved by tactics for 5.0.2

PR #1656, opened 2020-11-30
Add tlens tool

marek-safar:
Trimming Lens is an advanced tool for developers to help them explore ways to reduce their trimmed app or library size. The tool does not replace ILLinker but it's intended to be used in combination with it to produce smaller outputs.

/cc @eerhardt @vitek-karas


vitek-karas:
Definitely sounds interesting.

Would it be better to base the implementation on `System.Reflection.Metadata` instead?

marek-safar:
I tried to convert it to SRM and it was not as easy as I thought. Merging to make it easier to run and will try to do it again sometime later.

PR #1661, opened 2020-12-01
Temporary undo Arcade update to 5.0 rtm dependency to unblock linker 

marek-safar:
updates in dotnet/runtime repo


vitek-karas (approved):
As a change I don't see the problem... but it would be nice to describe what problem are we trying to avoid.

marek-safar:
It's the minimal required SDK mixture (https://github.com/dotnet/runtime/pull/45108)

PR #1665, opened 2020-12-03
Port single-file analyzer

mateoatr:


test/ILLink.RoslynAnalyzer.Tests/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -0,0 +1,123 @@
...
+			const string src = @"
+using System.Reflection;
+class C
+{
+    public string M() => Assembly.GetExecutingAssembly().Location;
+}";
+
+			return VerifyCS.VerifyAnalyzerAsync (src,
+				new (string, string)[] { ($"build_property.{MSBuildPropertyOptionNames.PublishSingleFile}", "true") },
+				VerifyCS.Diagnostic (AvoidAssemblyLocationInSingleFile.IL3000).WithSpan (5, 26, 5, 66).WithArguments ("System.Reflection.Assembly.Location"));
```
tlakollo, line 26: You are missing the comment here:
// (5,26): warning IL3000: 'System.Reflection.Assembly.Location' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.

test/ILLink.RoslynAnalyzer.Tests/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -0,0 +1,123 @@
...
+        var a = Assembly.GetExecutingAssembly();
+        _ = a.Location;
+        // below methods are marked as obsolete in 5.0
+        // _ = a.CodeBase;
+        // _ = a.EscapedCodeBase;
+    }
+}";
+			return VerifyCS.VerifyAnalyzerAsync (src,
+				new (string, string)[] { ($"build_property.{MSBuildPropertyOptionNames.PublishSingleFile}", "true") },
+				// /0/Test0.cs(8,13): warning IL3000: 'System.Reflection.Assembly.Location' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.
```
tlakollo, line 47: What's the /0/Tests0.cs?

src/ILLink.RoslynAnalyzer/DiagnosticCategory.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace ILLink.RoslynAnalyzer
+{
+	internal static class DiagnosticCategory
+	{
+		public const string Publish = nameof (Publish);
```
vitek-karas, line 9: Why is this "publish" - a very general term - and we use it for single-file
And then we have "trimming" - which is specific just for trimming.

Shouldn't it be "single-file" and "trimming" instead?

tlakollo (approved):

test/ILLink.RoslynAnalyzer.Tests/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -0,0 +1,124 @@
...
+    public void M()
+    {
+        var a = Assembly.GetExecutingAssembly();
+        _ = a.GetFile(""/some/file/path"");
+        _ = a.GetFiles();
+    }
+}";
+			return VerifyCS.VerifyAnalyzerAsync (src,
+				new (string, string)[] { ($"build_property.{MSBuildPropertyOptionNames.PublishSingleFile}", "true") },
+				// /0/Test0.cs(8,13): warning IL3001: Assemblies embedded in a single-file app cannot have additional files in the manifest.
```
tlakollo, line 69: There is still "/0/Tests.cs" in other warnings

test/ILLink.RoslynAnalyzer.Tests/AvoidAssemblyLocationInSingleFileTests.cs
mateoatr: Not sure, I'd guess @agocke copy-pasted the warning from a testing file -- Removed.

src/ILLink.RoslynAnalyzer/DiagnosticCategory.cs
mateoatr: True. I've changed the category to SingleFile.

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
```diff
@@ -71,9 +71,9 @@ public static DiagnosticResult Diagnostic (DiagnosticDescriptor descriptor)
 		}
 
 		/// <inheritdoc cref="AnalyzerVerifier{TAnalyzer, TTest, TVerifier}.VerifyAnalyzerAsync(string, DiagnosticResult[])"/>
-		public static async Task VerifyAnalyzerAsync (string src, params DiagnosticResult[] expected)
+		public static async Task VerifyAnalyzerAsync (string src, (string, string)[]? analyzerOptions = null, params DiagnosticResult[] expected)
```
agocke, line 74: To avoid having to add `expected:` everywhere, you could add an overload instead of an optional parameter

test/ILLink.RoslynAnalyzer.Tests/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -0,0 +1,124 @@
...
+		{
+			const string src = @"
+using System.Reflection;
+class C
+{
+    public string M() => Assembly.GetExecutingAssembly().Location;
+}";
+
+			return VerifyCS.VerifyAnalyzerAsync (src,
+				new (string, string)[] { ($"build_property.{MSBuildPropertyOptionNames.PublishSingleFile}", "true") },
```
agocke, line 25: I'd add a helper method in this class to pass `PublishSingleFile` as true (since almost every test passes it)

src/ILLink.RoslynAnalyzer/MSBuildPropertyOptionNames.cs
```diff
@@ -0,0 +1,25 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+using System.Linq;
+
+namespace ILLink.RoslynAnalyzer
+{
+	public static class MSBuildPropertyOptionNames
+	{
+		public const string PublishSingleFile = nameof (PublishSingleFile);
```
agocke, line 12: We probably need to add `PublishTrimmed` now since otherwise all the trim warnings will automatically be turned on during single-file

marek-safar: Wouldn't be better to think about the more general area than to hardcode the single narrow scenario? Are we going to add a property for each of remaining scenario which would benefit from the same logic?

agocke: Perhaps there's some broader feature work we could do, but we certainly need to recognize whether trimming or single-file are enabled, so we at least need to recognize the two MSBuild properties that already exist.

src/ILLink.RoslynAnalyzer/MSBuildPropertyOptionNames.cs
```diff
@@ -0,0 +1,26 @@
...
+		public const string IncludeAllContentForSelfExtract = nameof (IncludeAllContentForSelfExtract);
+		public const string PublishTrimmed = nameof (PublishTrimmed);
+	}
+
+	internal static class MSBuildPropertyOptionNamesHelpers
+	{
+		public static void VerifySupportedPropertyOptionName (string propertyOptionName)
```
agocke, line 19: This method appears to be unused

test/ILLink.RoslynAnalyzer.Tests/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        _ = a.GetFiles();
+    }
+}";
+			return VerifySingleFileAnalyzer (src,
+				// (8,13): warning IL3000: 'System.Reflection.Assembly.Location' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.
+				VerifyCS.Diagnostic (AvoidAssemblyLocationInSingleFile.IL3000).WithSpan (8, 13, 8, 23).WithArguments ("System.Reflection.Assembly.Location"),
+				// (9,13): warning IL3001: Assemblies embedded in a single-file app cannot have additional files in the manifest.
+				VerifyCS.Diagnostic (AvoidAssemblyLocationInSingleFile.IL3001).WithSpan (9, 13, 9, 25).WithArguments ("System.Reflection.Assembly.GetFiles()")
+				);
+		}
```
agocke, line 125: Can you add a test that we don't get any diagnostics when `PublishSingleFile` isn't set?

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
```diff
@@ -23,9 +31,9 @@ class C
     int M1() => 0;
     int M2() => M1();
 }";
-			return VerifyCS.VerifyAnalyzerAsync (TestRequiresWithMessageOnlyOnMethod,
+			return VerifyRequiresUnreferencedCodeAnalyzer (TestRequiresWithMessageOnlyOnMethod,
 				// (8,17): warning IL2026: Calling 'System.Int32 C::M1()' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. message.
-				VerifyCS.Diagnostic ().WithSpan (8, 17, 8, 21).WithArguments ("C.M1()", "message", "")
+				expected: VerifyCS.Diagnostic ().WithSpan (8, 17, 8, 21).WithArguments ("C.M1()", "message", "")
```
agocke, line 36: annotating with `expected` is now unnecessary, I think

PR #1666, opened 2020-12-04
Enable processing of dynamically referenced assemblies

sbomer:
Replaces https://github.com/mono/linker/pull/1611, now using the design from https://github.com/mono/linker/pull/1626.


src/linker/Linker.Steps/DynamicDependencyLookupStep.cs
```diff
@@ -4,14 +4,13 @@
...
 using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
 using Mono.Cecil;
 
 #nullable enable
 
 namespace Mono.Linker.Steps
 {
-	public class DynamicDependencyLookupStep : LoadReferencesStep
+	public class DynamicDependencyLookupStep : BaseStep
```
marek-safar, line 13: Where is this type used?

docs/error-codes.md
```diff
@@ -1511,3 +1511,29 @@ This is technically possible if a custom assembly defines `DynamicDependencyAttr
   [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
   object TestProperty { get; set; }
   ```
+
+### 'IL2100': XML may not contain wildcard for assembly fullname
```
marek-safar, line 1515: I'm wondering if we should land the new restrictions in separate PR

src/linker/Linker/LinkContext.cs
```diff
@@ -466,6 +469,26 @@ public AssemblyDefinition GetLoadedAssembly (string name)
...
+
+			while (toProcess.Count > 0) {
+				var assembly = toProcess.Dequeue ();
+				foreach (var reference in ResolveReferences (assembly)) {
+					if (!loaded.Add (reference))
+						continue;
+					yield return reference;
+					toProcess.Enqueue (reference);
+				}
+			}
```
marek-safar, line 489: I'm not sure about this, which code does need all expanded references

src/linker/Linker/EarlyMarker.cs
```diff
@@ -0,0 +1,43 @@
+using Mono.Cecil;
+using System.Diagnostics;
+
+namespace Mono.Linker
+{
+	class EarlyMarker : IMarker
```
marek-safar, line 6: This looks like unnecessary abstraction, why not to simply have 4 methods inside ResolveFromXmlStep

src/linker/Linker.Steps/DynamicDependencyLookupStep.cs
sbomer: Accidentally left over

src/linker/Linker/LinkContext.cs
sbomer: I'm still cleaning this up, but currently:
- SweepStep does this to handle some edge cases (hoping to get rid of this in the common case)
- it's used to match the current behavior of type name resolution which can find unqualified type names in any assembly
- XML that uses '*' to select all assemblies

src/linker/Linker/EarlyMarker.cs
sbomer: This is so that ResolveFromXml logic can run either early (to mark things before MarkStep calls Initialize*) or during MarkStep (calling the full Mark* logic). I might be able to get rid of the Initialize* logic completely - I am looking into it.

sbomer: Thinking more about this, I think it makes sense to keep this abstraction. If we were to call Mark* logic directly from ResolveFromXmlStep, it would "violate" the pipeline's order of execution, resulting in MarkStep logic called before MarkStep is executing. We could fix that by running the ResolveFromXmlSteps during MarkStep - but then we just end up with a new mini-pipeline of XML steps that get processed during MarkStep.

Another alternative is to never call MarkStep logic from ResolveFromXmlStep (instead only marking things using Annotations) - but this would mean separately keeping track of partially marked things that need to be fully marked later during MarkStep. We would need another HashSet or similar in Annotations to track everything marked from XML but not yet fully marked, which seems a bit wasteful.

src/linker/Linker.Steps/BaseAssemblyStep.cs
```diff
@@ -0,0 +1,40 @@
...
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+
+	public abstract class BaseAssemblyStep : IStep
+	{
+		private LinkContext context;
+
+		protected readonly AssemblyDefinition assembly;
```
vitek-karas, line 14: Code style comment: I personally would prefer the `_assembly` style for fields. We do use it in places in the linker (for example LinkContext is done this way).

I don't know if there's a guide one way or another... so I can't argue either way.

My personal preference is visible difference between fields and local variables.

vitek-karas: In this specific case I may actually change this to a readonly property... but again, just personal preference.

src/linker/Linker.Steps/EmbeddedXmlStep.cs
```diff
@@ -86,8 +88,8 @@ protected override void Process ()
 									.Where (res => res.Name.Equals ("ILLink.Substitutions.xml", StringComparison.OrdinalIgnoreCase))
 									.Cast<EmbeddedResource> ()) {
 					try {
-						Context.LogMessage ($"Processing embedded substitution descriptor {rsc.Name} from {asm.Name}");
-						steps_to_add.Push (GetExternalSubstitutionStep (rsc, asm));
+						Context.LogMessage ($"Processing embedded substitution descriptor {rsc.Name} from {assembly.Name}");
+						steps.Add (GetExternalSubstitutionStep (rsc, assembly));
```
vitek-karas, line 92: Do we want to build in some mechanism to be able to run the substitutions step early? We will need this for the constant propagation work - there we need to resolve assembly and run substitutions, but avoid running descriptors (and attributes don't matter one way or another right now).

src/linker/Linker/LinkContext.cs
vitek-karas: > XML that uses '*' to select all assemblies

We will need a different solution for this unfortunately. Especially if we rely on this mechanism to remove nullable annotations. We would need something like: register a callback which is called on all assemblies - already loaded and those in the future. And in it apply the necessary stuff from the XML.

We can definitely solve this in a followup change - but we should keep track of these things. Can you please create a tracking issue (or reuse existing one) and add the check mark list of things which need to happen before we are confident that we got everything we needed done.

I added a test for this case here: https://github.com/mono/linker/pull/1705

sbomer: Thanks, I forgot that we will always hit this case because of the corelib descriptors. I'll track it in https://github.com/mono/linker/issues/1708.

src/linker/Linker.Steps/EmbeddedXmlStep.cs
sbomer: I thought about this too - it might be nice for the required processing to be triggered directly when we ask questions that require it. For example`HasLinkerAttribute` could trigger scanning for embedded attribute XML and traversing the method body could ensure substitutions from that assembly have run. This way wouldn't require a call to `ProcessAssemblySteps` at key points in MarkStep.

The downside is that it would be a little more complicated if we want to avoid scanning resources multiple times (once each for attribute/substitutions/descriptor XML), and it would take separate caches for each assembly step that can be run separately - but it's not really that hard. I just went with this approach for now because it seemed simpler.

Could you clarify this?
> there we need to resolve assembly and run substitutions, but avoid running descriptors

I thought that the descriptors would need to be processed anyway - for example even if an assembly is only used via a call to a constant-returning method, that call would still be kept in the final output, so I'd expect the descriptors to be processed. If they need to be processed anyway, it reduces the benefit of letting these steps run separately.

src/linker/Linker.Steps/BaseAssemblyStep.cs
sbomer: I changed `context` to be a property, and `assembly` to `_assembly` (I didn't make it a property because it looks too much like `System.Reflection.Assembly` which is used by some derived classes). I'm never sure which of the inconsistent conventions to follow - I'm happy to take suggestions.

src/linker/Linker/LinkContext.cs
sbomer: Oh, and the expanded references are also needed for they `copy/save` action: https://github.com/mono/linker/pull/1666/commits/b3fef48e10896c2506abfe9d258ad1d9afe910f8#diff-f3ab7d627296ba105613b9cc039ca6f4ddc7a9aa66c6060ca82f6456ae0ede4fR416.

src/linker/Linker.Steps/EmbeddedXmlStep.cs
vitek-karas: About not running descriptors:
I think the rule was, don't do anything on an assembly until it's actually marked. To correctly implement constant propagation we need to break that a little bit and say "It's OK to process substitutions on it". But having to load an assembly from the constant propagation doesn't mean it will be marked. Maybe it's in a branch which will be removed. I think it's gonna take a lot of changes to make the constant propagation perform branch removal in one pass (I looked into it, the code currently is in no way going to work even remotely like that, it would probably require complete rewrite and it's pretty tricky to do so).

I think it would also be a good idea to separate the concerns - constant propagation and marking. If we do process descriptors before the assembly is marked, the descriptor processing may end up marking more stuff than we need.

MichalStrehovsky: How much do we actually rely on the fact that constant propagation can propagate things recursively?

I.e. that we can tell from `int Foo => IntPtr.Size; int Bar => Foo` that the `Bar` is constant. I understand that in theory it will help, but is it really helping enough to justify the engineering complexities?

vitek-karas: Not much - I just tried and the deepest is 3 - so basically the sample you have. For example `System.Environment.Is64BitOperatingSystem` is determined as constant, but to do that we need to recurse 2 more levels.

That said the problem discussed here is not about recursion really. Single level propagation is enough to trigger it, the problem is about cross-assembly. I don't have the numbers for that - but basically any `EventSource` in non-CoreLib assembly will run into it (`EventSource.IsEnabled` is substituted and is called from many other assemblies).

Some fun statistics (simple hello world, but we analyze big part of the framework under the current design):
* 3 iterations - so we run over all methods 3 times
* 72K methods analyzed (x 3) - this is the perf killer right now - we analyze 217K methods in total - LOT of them without any result and lot of them will be trimmed in the app
* 11 methods substituted (as in there 11 methods which were encountered and are substituted via the XML)
* 44 methods analyzed as constant after propagation (so at least one recursion)
* 1577 methods analyzed as constant in total
* 3183 callsites of constant methods found (probably exaggerated by 3 iterations, so it's more like 1K+ in reality)

MichalStrehovsky: Thank you! Really eye-opening to see the statistics!

sbomer: I decoupled the steps from each other so that it's easy to run just substitutions or other steps (and to avoid unnecessary descriptor processing).
- Right now substitutions are triggered when we ask for a method action (see [GetMethodAction](https://github.com/sbomer/linker/blob/f61d38cb9f766dc4ffa2e27bbf69383633abde14/src/linker/Linker.Steps/MarkStep.cs#L2488))
- Constant propagation is requested directly from MarkMethodBody. When we know how to solve the constant prop problem we might be able to put this inside of a helper.

src/linker/Linker.Steps/LinkAttributesStep.cs
```diff
@@ -156,7 +156,7 @@ bool GetAttributeType (XPathNodeIterator iterator, string attributeFullName, out
 		{
 			string assemblyName = GetAttribute (iterator.Current, "assembly");
 			if (string.IsNullOrEmpty (assemblyName)) {
-				attributeType = Context.GetType (attributeFullName);
+				attributeType = Context.TypeNameResolver.ResolveTypeName (attributeFullName)?.Resolve ();
```
sbomer, line 159: Separate this change.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -366,13 +376,58 @@ void Process ()
...
+						scanReferences = true;
+						break;
+					}
+
+					// The action should not change from the explicit command-line action
+					Debug.Assert (_context.Annotations.GetAction (assembly) == action);
+				}
+			}
+
+			var assembliesToCheck = scanReferences ? _context.ReferencedAssemblies ().ToArray () : _context.GetAssemblies ();
```
sbomer, line 422: Comment: Beware this works on loaded assemblies, not marked assemblies, so should not be tied to marking. We could do it when an assembly is loaded (since a load event is the only way to change the value of scanReferences), but we do it here to avoid doing it so frequently. We could further optimize to avoid doing this at all if the last mark iteration didn't load anything new.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1188,9 +1254,29 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
...
 
+		void ProcessDescriptorXml (AssemblyDefinition assembly)
+		{
+			if (Annotations.ProcessedDescriptorXml (assembly))
+				return;
+
+			new EmbeddedXmlStep (assembly).ProcessDescriptors (_context, this);
+		}
+
+		void ProcessAssemblySteps (AssemblyDefinition assembly)
```
sbomer, line 1265:
```suggestion
		void ProcessPerAssemblySteps (AssemblyDefinition assembly)
```
To match the change above. Or ProcessStepsForAssembly?

src/linker/Linker/CustomAttributeSource.cs
```diff
@@ -35,6 +66,8 @@ public IEnumerable<CustomAttribute> GetCustomAttributes (ICustomAttributeProvide
 					yield return customAttribute;
 			}
 
+			EnsureProcessedAttributeXml (provider);
```
sbomer, line 69: Do we really need to process attribute XML for CustomAttributeSource, or is it enough to ensure XML is processed for LinkerAttributesInformation?

sbomer, line 69: Avoid an extra lookup in case the assembly has no attribute XML.

sbomer (commented):
Adding some feedback from @vitek-karas. Thank you!

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -201,16 +214,12 @@ protected virtual void InitializeAssembly (AssemblyDefinition assembly)
...
 			case AssemblyAction.Copy:
+			case AssemblyAction.CopyUsed:
 			case AssemblyAction.Save:
-				Tracer.AddDirectDependency (assembly, new DependencyInfo (DependencyKind.AssemblyAction, action), marked: false);
-				MarkEntireAssembly (assembly);
-				break;
-			case AssemblyAction.Link:
 			case AssemblyAction.AddBypassNGen:
 			case AssemblyAction.AddBypassNGenUsed:
-				MarkAssembly (assembly);
```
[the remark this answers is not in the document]
sbomer: @vitek-karas the answer is - typically nothing actually marks the assembly. Instead, we mark the module if used (in MarkScope). SweepStep actually checks the module, despite a misleading method name:
https://github.com/mono/linker/blob/c45a25d77dfe278bd8fa675cb998cb0573dcb1c0/src/linker/Linker.Steps/SweepStep.cs#L256-L259

I think we should rename `MarkAssembly`.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -193,6 +200,12 @@ public virtual void Process (LinkContext context)
 
 		void Initialize ()
 		{
+			// Pre-load corelib so that its XML gets processed first. This is necessary because the
+			// corelib atttribute XML can contain modifications to other assemblies.
```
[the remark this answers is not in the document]
sbomer: @vitek-karas since marking is separated from attribute XML processing, marking corelib here would only "happen to" process the attribute XML (when we mark some attribute, to check whether it should be removed). In a world where corelib didn't have any attributes, marking it wouldn't process the attribute XML.

So I think it's better to explicitly run just the attribute processing here, and let the normal mark logic take care of marking. I'll add a comment so that it's more obvious what this is for.

src/linker/Linker/CustomAttributeSource.cs
sbomer: We do need it for CustomAttributeSource, because we don't always go through LinkerAttributesInformation - for example:
https://github.com/mono/linker/blob/c45a25d77dfe278bd8fa675cb998cb0573dcb1c0/src/linker/Linker.Dataflow/FlowAnnotations.cs#L110

src/linker/Linker.Steps/MarkStep.cs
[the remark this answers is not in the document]
sbomer: I left it as MarkAssembly, but added a clarifying comment.

src/linker/Linker.Steps/EmbeddedXmlStep.cs
```diff
@@ -37,76 +37,104 @@
 
 namespace Mono.Linker.Steps
 {
-
-	public class BlacklistStep : BaseStep
+	public class EmbeddedXmlStep : BasePerAssemblyStep
```
marek-safar, line 40: Why does it need to be based on BasePerAssemblyStep. Couldn't be something simpler like `ProcessAssemblyEmbeddedXmls` with static ProcessXYZ methods?

src/linker/Linker.Steps/EmbeddedXmlStep.cs
```diff
@@ -37,76 +37,104 @@
...
+
+		public void ProcessSubstitutions ()
+		{
+			if (Annotations.GetAction (_assembly) == AssemblyAction.Skip)
+				return;
+
+			var steps = new List<BodySubstituterStep> ();
+			var embeddedXml = GetEmbeddedXml (_assembly);
+
+			foreach (var rsc in embeddedXml
```
marek-safar, line 100: Why do you need foreach when you are looking for single entry name? Wouldn't be better to change `GetEmbeddedXml` to return single entry?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -193,29 +198,26 @@ public virtual void Process (LinkContext context)
...
-				MarkEntireAssembly (assembly);
-				break;
-			case AssemblyAction.Link:
-			case AssemblyAction.AddBypassNGen:
-			case AssemblyAction.AddBypassNGenUsed:
-				MarkAssembly (assembly);
-
-				foreach (TypeDefinition type in assembly.MainModule.Types)
-					InitializeType (type);
+			if (action == AssemblyAction.Skip || action == AssemblyAction.Delete)
```
marek-safar, line 216: .Delete check is redundant here

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1474,12 +1579,15 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 						2045, sourceLocationMember, subcategory: MessageSubCategory.TrimAnalysis);
 			}
 
+			ApplyPreserveInfo (type);
```
marek-safar, line 1582: Why is this needed. PreserveInfo data should be static during MarkStep

src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -47,15 +47,18 @@ public class ResolveFromXmlStep : ProcessLinkerXmlStepBase
 
 		static readonly string[] _accessorsAll = new string[] { "all" };
 		static readonly char[] _accessorsSep = new char[] { ';' };
+		IMarker Marker { get; }
```
marek-safar, line 50: I still don't understand why this abstraction is needed. It should be fine to keep the code as-is and it'd work as annotations are always populated before assembly marking

marek-safar (changes requested):
Did another pass (but there is still a lot to be improved)

src/linker/Linker.Steps/ResolveFromXmlStep.cs
sbomer: It's so that `ResolveFromXmlStep` can run on XML descriptors from delay loaded assemblies. In this case it needs to fully mark XML dependencies (IMarker will be the MarkStep itself), instead of just marking annotations (the Initialize logic won't run again, nor should it).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1180,11 +1200,22 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
 
 			foreach (TypeDefinition type in assembly.MainModule.Types)
 				MarkEntireType (type, includeBaseTypes: false, new DependencyInfo (DependencyKind.TypeInAssembly, assembly), null);
+
+			ProcessEmbeddedXml (assembly);
+		}
+
+		void ProcessEmbeddedXml (AssemblyDefinition assembly)
```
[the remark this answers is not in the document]
sbomer: This is outdated and has been removed.

Based on this [discussion](https://github.com/mono/linker/pull/1666#discussion_r544475617) it doesn't make sense to process all XML in the same place. The descriptors should only be processed if the assembly is marked, whereas substitutions/attributes might be needed even if not marked. For example we might ask whether an attribute has `RemoveAttributeInstances`, when the attribute is defined in an assembly that's never marked.

If we need a per-assembly pipeline concept, the best place to do it might be here: https://github.com/sbomer/linker/blob/aa4728ce65768b4067c275694cad630c5efa42e8/src/linker/Linker.Steps/MarkStep.cs#L1276-L1286 - this is called when an assembly is marked - but it doesn't include all of the XML steps.

src/linker/Linker.Steps/MarkStep.cs
sbomer: Delay loaded embedded descriptor XML can change the typepreserve.

src/linker/Linker.Steps/MarkStep.cs
sbomer: How come? I don't think we want to initialize Delete assemblies.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -49,19 +50,31 @@ public SweepStep (bool sweepSymbols = true)
 
 		protected override void Process ()
 		{
-			assemblies = Context.Annotations.GetAssemblies ().ToArray ();
+			// To keep facades, scan all references so that even unused facades are kept
```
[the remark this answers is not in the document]
sbomer: This seems to be what https://github.com/mono/linker/blob/master/test/Mono.Linker.Tests.Cases/TypeForwarding/UnusedForwarderWithAssemblyLinkedAndFacadesKept.cs is checking. It is unusual though - the kept facade is empty (its forwarders are removed, it only has a module type). Assuming that is a bug, we can avoid loading references here.

src/linker/Linker/CustomAttributeSource.cs
vitek-karas: In theory everything could go through LinkerAttributeInformation - for example to change the FlowAnnotations to that is pretty simple:https://github.com/mono/linker/compare/master...vitek-karas:UseLinkerAttributeInformationForDataFlow (and that contains a change to parameter type which is the messiest part of that).

Unfortunately I don't think we should do this as it comes with degradation:
* The LinkerAttributeInformation will cache the result - but FlowAnnotations will never ask for this attribute on any member more than once (since it has its own cache) - so we would be paying for cache which is never hit again
* It requires us to create the actual attribute instance - which in this case is pure waste as we want just the one value from it and we will store it in a separate data structure right away.

So let's keep it as is for now - maybe later on we could introduce a new API on Annotations which would not cache and call the custom attributes directly - but that's just layering stuff - not very interested in doing that.

The LinkerAttributeInformation is for cases where we will ask on a member potentially multiple times and we don't hold separate cache for it elsewhere.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -48,7 +48,7 @@ public SweepStep (bool sweepSymbols = true)
 
 		protected override void Process ()
 		{
-			assemblies = Context.Annotations.GetAssemblies ().ToArray ();
+			assemblies = Context.ReferencedAssemblies ().ToArray ();
```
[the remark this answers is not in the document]
sbomer: Outdated - see https://github.com/mono/linker/pull/1666#discussion_r552937262

src/linker/Linker.Steps/EmbeddedXmlStep.cs
sbomer: It could, but then we would need to pass the LinkContext and AssemblyDefinition through the static methods and all their helpers instead (same is true for the other per-assembly steps).

src/linker/Linker/CustomAttributeSource.cs
sbomer: I would like to optimize this in the follow-up when I address the special case for '*' in corelib attributes. It interacts with this change because both need some way to track whether corelib's XML has been applied to an assembly.

src/linker/Linker.Steps/EmbeddedXmlStep.cs
sbomer: This was existing code that I moved around - I cleaned it up a bit now.

src/linker/Linker.Steps/SweepStep.cs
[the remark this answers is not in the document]
sbomer: Get rid of this

src/linker/Linker/EarlyMarker.cs
sbomer: Add some pipeline concept?

sbomer:
RemoveSecurityStep looks wrong - it should not require a separate step before MarkStep.

src/linker/Linker.Steps/MarkStep.cs
sbomer: Can be removed in anticipation of Marek's change

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -366,13 +376,58 @@ void Process ()
 						if (!Annotations.IsMarked (type))
 							continue;
 						_context.MarkingHelpers.MarkExportedType (exported, assembly.MainModule, new DependencyInfo (DependencyKind.ExportedType, type));
+
+						markedExport = true;
 					}
+
+					if (markedExport)
+						ProcessDescriptorXml (assembly);
```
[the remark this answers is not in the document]
sbomer: If this can't be handled uniformly with removal of --keep-facades, maybe just issue a warning. No need to support.

src/linker/Linker/EarlyMarker.cs
sbomer: This has changed to use the mechanism in https://github.com/mono/linker/pull/1768.

src/linker/Linker.Steps/MarkStep.cs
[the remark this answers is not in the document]
sbomer: As discussed, the steps which run lazily on marked assemblies will be hard-coded for now. A separate pipeline may be introduced as part of https://github.com/mono/linker/issues/1735.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
sbomer: This has been removed

src/linker/Linker.Steps/MarkStep.cs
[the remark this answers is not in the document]
sbomer: Removed the check for descriptors in pure facades. The --keep-facades removal is tracked in https://github.com/mono/linker/issues/1740.

src/linker/Linker.Steps/SweepStep.cs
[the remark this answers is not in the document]
sbomer: I'd like to scope this change to preserving the current facade behavior, while also enabling loading of dynamically referenced assemblies. Loading assemblies here is only necessary for `--keep-facades`, and it will go away when we fix https://github.com/mono/linker/issues/1740.

sbomer:
@marek-safar @vitek-karas I addressed all the feedback above - PTAnother look. I cleaned up the commit history to (hopefully) make it a bit easier to review. To summarize the state of things:
- This now works on top of the "mark pending" logic instead of a separate `IMarker` interface
- There's no per-assembly pipeline, as discussed. Instead, the per-assembly steps are hard-coded in `MarkStep.ProcessPerAssemblySteps`.
- Works with cross-assembly constant prop thanks to @vitek-karas's changes :)
- This still needs to load all references to handle the following cases:
  - `copy/save` action (we could re-examine this when we switch to using an attribute opt-in)
  - `--keep-facades` (will go away with https://github.com/mono/linker/issues/1740)
  - `GetType` with an unqualified type name (tracked in https://github.com/mono/linker/issues/1708)
  - `*` in attribute XML (tracked in https://github.com/mono/linker/issues/1708)
  - (new) scanning for `DynamicInterfaceCastableImplementation` attributes  (see https://github.com/mono/linker/issues/1788)
- This doesn't try to change how `RemoveSecurityStep` works, other than running it per marked assembly. (https://github.com/mono/linker/pull/1732 has the beginnings of an attempt to handle it in SweepStep, but it needs more work and doesn't need to block this change).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -333,11 +350,54 @@ void Process ()
...
+
+				MarkFullyPreservedAssemblies ();
 			}
 
 			ProcessPendingTypeChecks ();
 		}
 
+		static bool IsFullyPreservedAction (AssemblyAction action) => action == AssemblyAction.Copy || action == AssemblyAction.Save;
+
+		void MarkFullyPreservedAssemblies ()
```
marek-safar, line 362: I find the logic here odd. Why is this method called in the marking loop? What does it need to precompute all the references?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1267,9 +1311,22 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
 				MarkEntireType (type, includeBaseTypes: false, new DependencyInfo (DependencyKind.TypeInAssembly, assembly), null);
 		}
 
+		void ProcessPerAssemblySteps (AssemblyDefinition assembly)
```
marek-safar, line 1314: I'd not call it Steps it's just old artifact name at this point for the logic

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -200,7 +211,7 @@ void Initialize ()
 		protected virtual void InitializeAssembly (AssemblyDefinition assembly)
 		{
 			var action = _context.Annotations.GetAction (assembly);
-			if (action == AssemblyAction.Copy || action == AssemblyAction.Save)
+			if (IsFullyPreservedAction (_context.Annotations.GetAction (assembly)))
```
vitek-karas, line 214:
```suggestion
			if (IsFullyPreservedAction (action))
```

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -307,9 +318,15 @@ private void MarkEntireTypeInternal (TypeDefinition type, bool includeBaseTypes,
...
 		void Process ()
 		{
-			while (ProcessPrimaryQueue () || ProcessLazyAttributes () || ProcessLateMarkedAttributes ()) {
+			while (ProcessPrimaryQueue () || ProcessMarkedPending () || ProcessLazyAttributes () || ProcessLateMarkedAttributes ()) {
 
 				// deal with [TypeForwardedTo] pseudo-attributes
+
+				// This marks all exported types that resolve to a marked type. Note that it
+				// may mark unused exported types that happen to resolve to a type which was
+				// marked from a different reference. This seems incorrect.
```
vitek-karas, line 327: We should create an issue for this - otherwise it's just another todo in the code which nobody will ever look at.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -333,11 +350,54 @@ void Process ()
 						_context.MarkingHelpers.MarkExportedType (exported, assembly.MainModule, di);
 					}
 				}
+
+				MarkFullyPreservedAssemblies ();
```
vitek-karas, line 354: I wonder if it would be better to make this return `bool`, just like `ProcessPrimaryQueue` and similar and call it from the `while` condition. The advantage would be that it would be done only once everything else was processed. Currently it runs every time we finish one of the queues. So very likely more often.

We could probably do something similar for the type forwards stuff above as well. Both of these operations are not order sensitive, we just need to run them eventually and then process all queues ones they're done. But it would be more optimal to run them late so that we don't have to repeat them that often.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -466,15 +526,15 @@ void ProcessMarkedTypesWithInterfaces ()
 
 		void DiscoverDynamicCastableImplementationInterfaces ()
 		{
-			Debug.Assert (!_dynamicInterfaceCastableImplementationTypesDiscovered);
-
-			foreach (var assembly in _context.GetAssemblies ()) {
+			foreach (var assembly in _context.GetReferencedAssemblies ().ToArray ()) {
```
vitek-karas, line 529: Just to make sure I understand this - the effect of this is that we will always load full closure still. We need https://github.com/mono/linker/issues/1788 resolved to "fix" that, right?

src/linker/Linker.Steps/BasePerAssemblyStep.cs
```diff
@@ -0,0 +1,28 @@
...
+
+	public abstract class BasePerAssemblyStep
+	{
+		protected readonly AssemblyDefinition _assembly;
+
+		protected LinkContext Context { get; private set; }
+
+		public AnnotationStore Annotations => Context?.Annotations;
+
+		public BasePerAssemblyStep (AssemblyDefinition assembly, LinkContext context)
```
vitek-karas, line 18: I wonder if we should take this a step further and simply remove the notion of "Steps" - rename - and remove the base class. There's no real need for it.

Even if keep the "Steps" notation, I don't like that we create it for each assembly (that's not how the new extensibility will work either). I would expect `Process(assembly)` style... or just make it static and get rid of "steps".

src/linker/Linker/MemberActionStore.cs
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		public void SetAction (MethodDefinition method, MethodAction action)
+		{
+			_methodActions[method] = action;
+		}
+
+		public MethodAction GetAction (MethodDefinition method)
+		{
+			EnsureProcessedSubstitutionXml (method.Module.Assembly);
```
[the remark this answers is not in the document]
vitek-karas: The process-per-assembly only runs on marked assemblies, but substitutions are needed on potentially all assemblies. The constant propagation operates on any assembly it needs to - and currently it can be a loaded but unmarked assembly. It also needs this information before the assembly will be marked typically.

In theory it would be possible to integrate constant propagation into marking itself (method body processing), in which case it might be possible to only rely on marked assemblies in this case and remove this requirement - but that's LOT of refactoring and I'm still not sure it's possible to do it that way 100%.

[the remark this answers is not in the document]
vitek-karas: Alternatively we could do this for all loaded assemblies at the point where they are loaded, but that feels unnecessary.
Note that linker attributes has similar behavior - it makes sure the XML is parsed when anything asks for an attribute on a given assembly.

test/Mono.Linker.Tests.Cases/Reflection/AssemblyImportedViaReflectionWithSweptReferences.cs
```diff
@@ -0,0 +1,36 @@
...
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+using Mono.Linker.Tests.Cases.Reflection.Dependencies;
+
+namespace Mono.Linker.Tests.Cases.Reflection
+{
+	[SetupCSharpCompilerToUse ("csc")]
+	[SetupCompileBefore ("unusedreference.dll", new[] { "Dependencies/UnusedAssemblyDependency.cs" })]
+	[SetupCompileBefore ("reference.dll", new[] { "Dependencies/AssemblyDependency.cs" }, addAsReference: false)]
+	[SetupCompileBefore ("library.dll", new[] { "Dependencies/AssemblyDependencyWithMultipleReferences.cs" }, new[] { "reference.dll", "unusedreference.dll" }, addAsReference: false)]
+	// TODO: keep library even if type is not found in it
```
vitek-karas, line 12: Do we need to file an issue for this?

src/linker/Linker.Steps/MarkStep.cs
sbomer: See https://github.com/mono/linker/blob/feature/lazyload/docs/design/assembly-loading.md#exceptions-to-lazy-loading. It's called in the marking loop because marking can load new assemblies which may introduce references that would get the copy/save action.

src/linker/Linker/LinkContext.cs
```diff
@@ -150,6 +150,12 @@ public class LinkContext : IDisposable
 
 		public string PInvokesListFile;
 
+		public bool StripSecurity { get; set; }
+
+		public System.Collections.IDictionary Actions {
```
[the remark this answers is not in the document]
sbomer: This was existing code left over after I rebased - changed to generic dictionary.

src/linker/Linker.Steps/BasePerAssemblyStep.cs
```diff
@@ -0,0 +1,28 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+
+	public abstract class BasePerAssemblyStep
```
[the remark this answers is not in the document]
sbomer: It was used - but I've removed it now and made the derived classes static.

src/linker/Linker.Steps/MarkStep.cs
sbomer: Added a link to https://github.com/mono/linker/issues/1740

src/linker/Linker.Steps/MarkStep.cs
sbomer: Inlined this logic

src/linker/Linker.Steps/MarkStep.cs
sbomer: Changed to return a bool

src/linker/Linker.Steps/MarkStep.cs
sbomer: That's right - I added a comment here.

test/Mono.Linker.Tests.Cases/Reflection/AssemblyImportedViaReflectionWithSweptReferences.cs
sbomer: Filed https://github.com/mono/linker/issues/1795 and mentioned it in the comment

src/linker/Linker.Steps/BasePerAssemblyStep.cs
sbomer: I made them static classes instead, and got rid of the "Step" terminology.

vitek-karas (approved):
Looks good. One last thing (can be done as a follow-up probably):
In order to avoid breaking Xamarin in a bad way, we could add a simple logic where if any custom step is added, insert a step which preloads the reference closure (just like we did before this change). The whole thing should still work just fine, it would just make it easier for the Xamarin custom steps to migrate over time.

src/linker/Linker/MemberActionStore.cs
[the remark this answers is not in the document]
marek-safar: > Alternatively we could do this for all loaded assemblies

That's what I meant. It seems to be much more logical than calling it within all kinds of GetXXX methods and pay extra cache hit checks everywhere.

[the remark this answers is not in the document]
vitek-karas: For substitutions it would probably work, for linker attributes it's a bit more complex as that can cause loading of additional assemblies. Doable... I guess it's a matter of preference. The current solution is as lazy as we can get. We could switch to a more eager solution.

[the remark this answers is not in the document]
sbomer: One principle I tried to follow is that assembly load shouldn't implicitly change global state (like the method actions here) - instead the extra processing should be done lazily by the logic that depends on it. I find this easier to reason about, but doing it lazily does require an extra check. On the other hand it can save us from doing unnecessary work for all loaded assemblies.

The different XMLs are needed at different points: it would be incorrect to process descriptor XML on assembly load, so we have to do at least that part lazily (on mark). While we *could* process attribute/substitution XML on load without introducing correctness issues, I find it more consistent to process all of the XMLs as-needed. Doing it on load would leave extra unused global state when SweepStep loads assemblies.

Are you more concerned about the lazy architecture, or about the perf impact of the check?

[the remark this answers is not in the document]
sbomer: To summarize our discussion - the main concern was that as written, getting attributes/actions could implicitly trigger processing which would add more attributes to the global store, and the calls to `EnsureProcessed` are easy to get wrong. I've changed it so that:
- There is an explicit difference between the global store and the per-assembly embedded XML for both substitutions and attributes
- The getters won't implicitly modify the global store, and will do XML processing as part of the cache initialization for the per-assembly stores.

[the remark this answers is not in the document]
sbomer: BTW @vitek-karas, I ended up avoiding structs for the `AttributeInfo` and `SubstitutionInfo` because there's no way (AFAIK) to ensure the struct members are initialized, but we assume they are non-null in a lot of places. Other options I considered were:
- Put null checks everywhere we access the struct members. This is easier to get wrong and arguably makes the code less readable. We could use `#nullable enable`, but this requires patterns like:
```csharp
if (embeddedXml.CustomAttributes?.TryGetValue (provider, out annotations) == true)
```
(the `== true` is necessary), which wasn't received well last time I introduced it.
- Struct with lazy getters replacing the fields. It can't be a readonly struct, there's just a bit more boilerplate, and the `Add*` methods either need to access the struct members through its own getters or have null checks everywhere, but it works otherwise.

src/linker/Linker/AttributeInfo.cs
```diff
@@ -0,0 +1,40 @@
...
+
+		public AttributeInfo ()
+		{
+			CustomAttributes = new Dictionary<ICustomAttributeProvider, IEnumerable<CustomAttribute>> ();
+			InternalAttributes = new Dictionary<ICustomAttributeProvider, IEnumerable<Attribute>> ();
+		}
+
+		public void AddCustomAttributes (ICustomAttributeProvider provider, IEnumerable<CustomAttribute> customAttributes)
+		{
+			if (!CustomAttributes.ContainsKey (provider))
```
marek-safar, line 26: It should use TryGetValue

src/linker/Linker/AttributeInfo.cs
```diff
@@ -0,0 +1,40 @@
...
+				CustomAttributes[provider] = customAttributes;
+			else
+				CustomAttributes[provider] = CustomAttributes[provider].Concat (customAttributes);
+		}
+
+		public void AddInternalAttributes (ICustomAttributeProvider provider, IEnumerable<Attribute> attributes)
+		{
+			if (!InternalAttributes.ContainsKey (provider))
```
marek-safar, line 34: It should use TryGetValue

src/linker/Linker/MemberActionStore.cs
```diff
@@ -9,79 +9,72 @@ namespace Mono.Linker
...
-			_methodActions[method] = action;
+			return xmlInfo != null;
 		}
 
 		public MethodAction GetAction (MethodDefinition method)
 		{
-			EnsureProcessedSubstitutionXml (method.Module.Assembly);
-
-			if (_methodActions.TryGetValue (method, out MethodAction action))
+			if (GlobalSubstitutionInfo.MethodActions.TryGetValue (method, out MethodAction action))
```
marek-safar, line 36: It's odd that global substitution wins over assembly local substitution but it's probably just a naming issue.

src/linker/Linker/SubstitutionInfo.cs
```diff
@@ -0,0 +1,45 @@
...
+using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	public class SubstitutionInfo
+	{
+		public Dictionary<MethodDefinition, MethodAction> MethodActions;
+		public Dictionary<MethodDefinition, object> MethodStubValues;
+		public Dictionary<FieldDefinition, object> FieldValues;
+		public HashSet<FieldDefinition> FieldInit;
```
marek-safar, line 15: Does it need to be public when you already have several Set* methods?

marek-safar (approved):
Few more comments but otherwise LGTM

src/linker/Linker/SubstitutionInfo.cs
sbomer: Thanks, these should have been properties (they need to be public)

src/linker/Linker/MemberActionStore.cs
sbomer: It's to keep the behavior where `SetAction` from a custom step wins over the action from a substitution XML. I changed it to "PrimarySubstitutionInfo" - not sure if that's much better, but at least it doesn't mislead about the precedence.

src/linker/Linker/AttributeInfo.cs
sbomer: More existing code that moved around... fixed it!

src/linker/Linker/SubstitutionInfo.cs
```diff
@@ -0,0 +1,45 @@
...
+		{
+			MethodStubValues[method] = value;
+		}
+
+		public void SetFieldValue (FieldDefinition field, object value)
+		{
+			FieldValues[field] = value;
+		}
+
+		public void SetSubstitutedInit (FieldDefinition field)
```
vitek-karas, line 40: This should read `SetFieldInit` right?

src/linker/Linker/SubstitutionInfo.cs
```diff
@@ -0,0 +1,45 @@
...
+
+		public SubstitutionInfo ()
+		{
+			MethodActions = new Dictionary<MethodDefinition, MethodAction> ();
+			MethodStubValues = new Dictionary<MethodDefinition, object> ();
+			FieldValues = new Dictionary<FieldDefinition, object> ();
+			FieldInit = new HashSet<FieldDefinition> ();
+		}
+
+		public void SetAction (MethodDefinition method, MethodAction action)
```
vitek-karas, line 25: Nit: I would call this `SetMethodAction`.

src/linker/Linker.Steps/BodySubstitutionParser.cs
```diff
@@ -0,0 +1,201 @@
...
+			: base (document, xmlDocumentLocation)
+		{
+		}
+
+		public BodySubstitutionParser (XPathDocument document, EmbeddedResource resource, AssemblyDefinition resourceAssembly, string xmlDocumentLocation = "")
+			: base (document, resource, resourceAssembly, xmlDocumentLocation)
+		{
+		}
+
+		public SubstitutionInfo Parse (LinkContext context)
```
vitek-karas, line 23: Not important, but I personally would prefer to get the context in the .ctor.

src/linker/Linker.Steps/BodySubstitutionParser.cs
```diff
@@ -0,0 +1,201 @@
...
+		{
+			_substitutionInfo = new SubstitutionInfo ();
+			Process (context);
+			return _substitutionInfo;
+		}
+
+		public void Parse (LinkContext context, SubstitutionInfo xmlInfo)
```
vitek-karas, line 30: If we need the "in -> modify -> out " overload (this one), then I would only have that one. The caller can trivially create new info as input. It's better than to have to reason about which overload do I need to call.

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -0,0 +1,371 @@
...
+using System.Linq;
+using System.Text;
+using System.Xml.XPath;
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class LinkAttributesParser : ProcessLinkerXmlStepBase
+	{
+		AttributeInfo _xmlInfo;
```
vitek-karas: `_attributeInfo` instead ?

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -0,0 +1,371 @@
...
+			_xmlInfo = new AttributeInfo ();
+			Process (context);
+			return _xmlInfo;
+		}
+
+		public void Parse (LinkContext context, AttributeInfo xmlInfo)
+		{
+			_xmlInfo = xmlInfo;
+			Process (context);
+		}
```
vitek-karas: Same comment as for the substitution - I would prefer just one overload and context in .ctor

src/linker/Linker.Steps/BodySubstitutionParser.cs
```diff
@@ -0,0 +1,201 @@
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Xml.XPath;
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class BodySubstitutionParser : ProcessLinkerXmlStepBase
```
vitek-karas, line 9: Nit: rename `ProcessLinkerXmlStepBase` - it's not a "Step base" anymore.

vitek-karas (approved):

src/linker/Linker.Steps/MarkStep.cs
sbomer: I filed https://github.com/mono/linker/issues/1800 to track the ideas we discussed which might help us get rid of this.

src/linker/Linker/SubstitutionInfo.cs
sbomer: This was the existing name, but I like your suggestion better - changed

src/linker/Linker/SubstitutionInfo.cs
sbomer: Same as above - changed (note that the Annotations helper is still called SetAction)

src/linker/Linker.Steps/BodySubstitutionParser.cs
sbomer: Good idea, fixed!

src/linker/Linker.Steps/LinkAttributesParser.cs
sbomer: Already changed

src/linker/Linker.Steps/BodySubstitutionParser.cs
sbomer: It actually was still a base step - in the case of descriptor xml, it was called from the pipeline. And even for the embedded XML, parsing relied on the Process (LinkContext context) method from `IStep`.

I factored this out some more so that the shared XML processing logic is no longer an `IStep`. There's a new derived class for the descriptor XML, and ResolveFromXmlStep becomes a wrapper for it like the other XML steps.

src/linker/Linker.Steps/BodySubstitutionParser.cs
sbomer: Changed

vitek-karas (approved):

marek-safar:
The previous linker update in runtime is in, this is good to go.

sbomer:
Thanks for all the feedback @vitek-karas and @marek-safar!

PR #1667, opened 2020-12-04
Don't aggressively mark serialization constructors for illink

### LINKED ISSUE #1603: Serializable dependencies are always marked
The [code](https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/MarkStep.cs#L1478) inside linker which checks for a need to mark serializable dependencies fires every time it sees serializable type. This is wasteful and brings a lot of unnecessary dependencies. There should be a better way for the linker to keep the constructors only when they are needed.

@vitek-karas @MichalStrehovsky for their thoughts

marek-safar:
Fixes #1603

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1371,16 +1371,30 @@ namespace Mono.Linker.Steps
 			Annotations.Mark (scope, new DependencyInfo (DependencyKind.ScopeOfType, type));
 		}
 
+#if !FEATURE_ILLINK
 		protected virtual void MarkSerializable (TypeDefinition type)
 		{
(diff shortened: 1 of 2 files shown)


vitek-karas (approved):
Maybe add a test?
It's weird we already done have any for the code which was there... so maybe we should add it and "disable"/"change outcome" for ILLink.

marek-safar:
I'm not sure I follow, you mean to add test which will be disabled for illink?

vitek-karas:
The code which you "ifdefed" out does something - we probably don't have tests for it, otherwise it should have failed with your change.

For ILLink - have a test which marks something as serializable and validate that linker will not keep anything on it (unless normally referenced).

vitek-karas (approved):

PR #1668, opened 2020-12-04
LinkerDumpDependencies -> _TrimmerDumpDependencies?

layomia:
`LinkerDumpDependencies` didn't work for me, but `_TrimmerDumpDependencies` did. Not sure what needs to be updated.

cc @vitek-karas

### DIFF
--- a/src/analyzer/README.md
+++ b/src/analyzer/README.md
@@ -16,12 +16,12 @@ can be retrieved by enabling dependencies dumping during linking of a
 Xamarin.Android or a Xamarin.iOS project.
 
 That can be done on the command line by setting
-`LinkerDumpDependencies` property to `true` and building the
+`_TrimmerDumpDependencies` property to `true` and building the
 project. (make sure the LinkAssemblies task is called, it might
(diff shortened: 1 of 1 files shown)


vitek-karas:
@sbomer - the linker/SDK expert.

sbomer:
It looks like `LinkerDumpDependencies` was used when the analyzer lived in mono. `_TrimmerDumpDependencies` is the right option to use with the .NET SDK. However the rest of these instructions still mention Xamarin, and `_TrimmerDumpDependencies` will only work with the .NET Core SDK. Maybe @marek-safar has more context.

marek-safar:
LinkerDumpDependencies is actually correct property name when used with pre-netcore Xamarin SDKs. We need to clarify that in the doc. Though I'm not sure we want to document `_TrimmerDumpDependencies` private property.

layomia:
Based on the feedback here, closing this PR in favor of an issue https://github.com/mono/linker/issues/1676 to consider making the docs more friendly when using the .NET Core SDK

PR #1674, opened 2020-12-07
Add description how to embed linker XML files

vitek-karas:
Some additional description of the various formats and when/how they can be used.

### DIFF
--- a/docs/data-formats.md
+++ b/docs/data-formats.md
@@ -4,9 +4,9 @@ IL linker uses several data formats to control or influence the linking process.
 
 ## Descriptor Format
 
-Descriptors are used to direct linker to always keep some items in the assembly, regardless if linker can find any references to them.
+Descriptors are used to direct the linker to always keep some items in the assembly, regardless of if the linker can find any references to them.
 
-Descriptor XML can be embedded in an assembly. In that case it must be stored as an embedded resource with logical name `ILLink.Descriptors.xml`. To achieve this when build an assembly use this in the project file to include the XML:
+Descriptor XML can be embedded in an assembly. In that case it must be stored as an embedded resource with logical name `ILLink.Descriptors.xml`. To achieve this when building an assembly use this in the project file to include the XML:
 
 ```xml
   <ItemGroup>
--- a/docs/data-formats.md
+++ b/docs/data-formats.md
@@ -18,7 +18,7 @@ Descriptor XML can be embedded in an assembly. In that case it must be stored as
 
 Embedded descriptors only take effect if the containing assembly is included in the linker output, so if something from that assembly is marked to be kept.
 
-Descriptor XML can also be passed to the linker on the command via the [`-x` parameter](https://github.com/mono/linker/blob/master/docs/illink-options.md#linking-from-an-xml-descriptor).
+Descriptor XML can also be passed to the linker on the command via the [`-x` parameter](illink-options.md#linking-from-an-xml-descriptor).
 
 The `fullname` attribute specifies the full name of the type in the format specified by ECMA-335. This is in certain cases not the same as the one reported by Type.FullName for example for nested types.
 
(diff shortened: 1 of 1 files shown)


docs/data-formats.md
```diff
@@ -4,7 +4,23 @@ IL linker uses several data formats to control or influence the linking process.
 
 ## Descriptor Format
 
-The `fullname` attribute specifies the fullname of the type in the format specified by ECMA-335. This is in certain cases not the same as the one reported by Type.FullName for example for nested types.
+Descriptors are used to direct linker to always keep some items in the assembly, regardless if linker can find any references to them.
+
+Descriptor XML can be embedded in an assembly. In that case it must be stored as an embedded resource with logical name `ILLink.Descriptors.xml`. To achieve this when build an assembly use this in the project file to include the XML:
```
sbomer, line 9:
```suggestion
Descriptor XML can be embedded in an assembly. In that case it must be stored as an embedded resource with logical name `ILLink.Descriptors.xml`. To achieve this when building an assembly use this in the project file to include the XML:
```

docs/data-formats.md
```diff
@@ -166,6 +182,24 @@ The `required` attribute specifies that if the type is not marked, during the ma
 
 ## Substitution Format
 
+Substitutions direct the linker to replace specific method's body with either a throw or return constant statements.
+
+Substitutions only work for methods from assemblies which are linked with assembly action `link`, in all other cases substitutions are ignored (no warning or message is issued).
```
sbomer, line 187: We don't output the substituted IL, but couldn't the substitutions still affect constant propagation?

docs/data-formats.md
```diff
@@ -4,7 +4,23 @@ IL linker uses several data formats to control or influence the linking process.
 
 ## Descriptor Format
 
-The `fullname` attribute specifies the fullname of the type in the format specified by ECMA-335. This is in certain cases not the same as the one reported by Type.FullName for example for nested types.
+Descriptors are used to direct linker to always keep some items in the assembly, regardless if linker can find any references to them.
```
sbomer, line 7:
```suggestion
Descriptors are used to direct the linker to always keep some items in the assembly, regardless of if the linker can find any references to them.
```

sbomer (approved):

docs/data-formats.md
```diff
@@ -4,7 +4,23 @@ IL linker uses several data formats to control or influence the linking process.
...
+      <LogicalName>ILLink.Descriptors.xml</LogicalName>
+    </EmbeddedResource>
+  </ItemGroup>
+```
+
+Embedded descriptors only take effect if the containing assembly is included in the linker output, so if something from that assembly is marked to be kept.
+
+Descriptor XML can also be passed to the linker on the command via the [`-x` parameter](https://github.com/mono/linker/blob/master/docs/illink-options.md#linking-from-an-xml-descriptor).
+
+The `fullname` attribute specifies the full name of the type in the format specified by ECMA-335. This is in certain cases not the same as the one reported by Type.FullName for example for nested types.
```
MichalStrehovsky, line 23: While you're touching this, we should delete this description of fullname. The ECMA-335 format this refers to is https://github.com/dotnet/runtime/issues/4416, but what linker uses is not that format.

Fullname in linker is the full name as understood by Cecil, period. (Cecil differs from ECMA in generics, it differs by the fact that Cecil's fullname cannot distinguish `szArray` and `singleDimensionalArray` in the ECMA doc linked above, it differs in escaping rules because Cecil has none, etc.).

docs/data-formats.md
marek-safar, line 9: `logical name` sounds a bit offensive to me ;-)

docs/data-formats.md
```diff
@@ -4,7 +4,23 @@ IL linker uses several data formats to control or influence the linking process.
...
+  <ItemGroup>
+    <EmbeddedResource Include="ILLink.Descriptors.xml">
+      <LogicalName>ILLink.Descriptors.xml</LogicalName>
+    </EmbeddedResource>
+  </ItemGroup>
+```
+
+Embedded descriptors only take effect if the containing assembly is included in the linker output, so if something from that assembly is marked to be kept.
+
+Descriptor XML can also be passed to the linker on the command via the [`-x` parameter](https://github.com/mono/linker/blob/master/docs/illink-options.md#linking-from-an-xml-descriptor).
```
marek-safar, line 21: use shorter link version

marek-safar (approved):

docs/data-formats.md
vitek-karas: I simply removed that line - it's basically useless anyway - if we want better, we should add a list of examples instead.

Sk21renee37 (approved):

docs/data-formats.md
vitek-karas: You're right - I didn't even realize that. Changed the comment. It's relatively unlikely, would basically require a trimmed assembly to call non-trimmed on, but it's is a possible case.

docs/data-formats.md
vitek-karas: That's an MSBuild terminology - I didn't come up with that 😞

PR #1683, opened 2020-12-10
Fixes removal of unused assembly references for linked assemblies

### LINKED ISSUE #1296: When linker removes all references to an assembly the assembly ref is not removed
I noticed this when looking at the trimmed assemblies in the .NETCore shared framework.

The assembly System.Diagnostics.TraceSource had a reference to System.Diagnostics.Process in one place which was unreachable and trimmed by the linker, however the assembly ref to System.Diagnostics.Process remains.

Some tools will see this and still try to load Process, the linker should remove this if possible.

### LINKED ISSUE #1576: SweepStep doesn't process all type references for forwarded types
Test case: 
[linkertestcase-aedcbaa.zip](https://github.com/mono/linker/files/5407615/linkertestcase-aedcbaa.zip)

Extract and execute `testcase.sh` (not test.sh):

```
$ ./testcase.sh
++ dirname ./testcase.sh
+ cd .
+ LINKER=/usr/local/share/dotnet/sdk/5.0.100-rc.2.20480.7/Sdks/Microsoft.NET.ILLink.Tasks/tools/netcoreapp3.0/illink.dll
+ /usr/local/share/dotnet/dotnet /usr/local/share/dotnet/sdk/5.0.100-rc.2.20480.7/Sdks/Microsoft.NET.ILLink.Tasks/tools/netcoreapp3.0/illink.dll @testcase.rsp --custom-step -LoadReferencesStep:Xamarin.SetupStep,/Users/rolf/test/linkertestcase/dotnet-linker.dll --custom-data LinkerOptionsFile=/Users/rolf/test/linkertestcase/custom-linker-options.txt
[...]
+ monodis './obj/iPhoneSimulator/Debug/net5.0-ios/ios-x64/linked/link all.dll'
+ grep BROKEN
	IL_0020:  call instance void [Xamarin.iOS]ObjCRuntime.BlockLiteral::SetupBlockImpl(<BROKEN CLASS token_ 1000024 due to Could not resolve type with token 01000024 from typeref (expected class 'System.Delegate' in assembly 'System.Collections, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a') assembly:System.Collections, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a type:System.Delegate member:(null)>, <BROKEN CLASS token_ 1000024 due to Could not resolve type with token 01000024 from typeref (expected class 'System.Delegate' in assembly 'System.Collections, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a') assembly:System.Collections, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a type:System.Delegate member:(null)>, bool, string)
	IL_0020:  call instance void [Xamarin.iOS]ObjCRuntime.BlockLiteral::SetupBlockImpl(<BROKEN CLASS token_ 1000024 due to Could not resolve type with token 01000024 from typeref (expected class 'System.Delegate' in assembly 'System.Collections, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a') assembly:System.Collections, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a type:System.Delegate member:(null)>, <BROKEN CLASS token_ 1000024 due to Could not resolve type with token 01000024 from typeref (expected class 'System.Delegate' in assembly 'System.Collections, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a') assembly:System.Collections, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a type:System.Delegate member:(null)>, bool, string)
[...]
+ echo '❌ FAILED'
❌ FAILED
+ exit 1
```

The problem occurs when one of our custom linker step rewrites IL to call a different method: https://github.com/xamarin/xamarin-macios/blob/699c1f98fbfb9ac1c5d790a99812aa2f2b1a001b/tools/linker/CoreOptimizeGeneratedCode.cs#L911-L918

The different method is imported into the module in question here: https://github.com/xamarin/xamarin-macios/blob/699c1f98fbfb9ac1c5d790a99812aa2f2b1a001b/tools/linker/CoreOptimizeGeneratedCode.cs#L1058-L1073

The problem occurs because the imported method has a parameter whose type is System.Delegate, from System.Runtime.dll, and when System.Runtime.dll is linked away, the corresponding TypeReference in this method's parameter's ParameterType is not updated in SweepStep to reference System.Private.CoreLib.dll instead.

The end result is an invalid assembly (as can be seen from the monodis output) where the imported method's parameter's type is apparently System.Delegate from System.Collections.dll.

Here is an ugly hack that fixes the problem: https://gist.github.com/rolfbjarne/45e8e6d001f5fe9c9c1cf4036ad212f6

marek-safar:
The previous logic only removed assembly reference for linked assembly
if the reference didn't exist in the closure of input assemblies. This
was wrong because it kept unused references in one assembly if the
another assembly used same reference.

Fixes #1296
Fixes #1576


src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -728,16 +553,345 @@ protected virtual void ElementRemoved (IMetadataTokenProvider element)
...
+				case GenericParameter _:
+#endif
+					// Nothing to update
+					return;
+				}
+
+				//
+				// Resolve to type definition to remove any type forwarding imports
+				//
+				TypeDefinition td = type.Resolve ();
```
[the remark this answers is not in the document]
vitek-karas: Hmm - it was inconsistent before - for TypeRefs we did, but not for type refs in attributes...

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -728,16 +553,345 @@ protected virtual void ElementRemoved (IMetadataTokenProvider element)
...
+					}
+				}
+
+				if (typeDefinition.HasEvents) {
+					foreach (var e in typeDefinition.Events) {
+						UpdateCustomAttributesTypesScopes (e);
+						// e.EventType is not saved
+					}
+				}
+
```
vitek-karas, line 621: What about generic parameters on types? Specifically constrains and custom attributes?

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -728,16 +553,345 @@ protected virtual void ElementRemoved (IMetadataTokenProvider element)
...
+					UpdateCustomAttributesTypesScopes (p);
+					UpdateTypeScope (p.ParameterType);
+					UpdateMarshalInfoTypeScope (p);
+				}
+			}
+
+			void UpdateTypeScope (Collection<ExportedType> forwarders)
+			{
+				foreach (var f in forwarders) {
+					TypeDefinition td = f.Resolve ();
```
vitek-karas, line 695: Same comment about potential throw

test/Mono.Linker.Tests.Cases/TypeForwarding/TypeForwardedIsUpdatedForMissingType.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+	[SkipUnresolved (true)]
+	[KeepTypeForwarderOnlyAssemblies ("false")]
+	[SetupCompileBefore ("Lib.dll", new[] { "Dependencies/TypeForwardedIsUpdatedForMissingTypeLib.cs" })]
+	[SetupCompileBefore ("AnotherLibrary.dll", new[] { "Dependencies/TypeForwardedIsUpdatedForMissingTypeLib2.cs" })]
+
+	[SetupCompileAfter ("AnotherLibrary.dll", new[] { "Dependencies/TypeForwardedIsUpdatedForMissingTypeLib2.cs" })]
+	[SetupCompileAfter ("Implementation.dll", new[] { "Dependencies/TypeForwardedIsUpdatedForMissingTypeLib.cs" })]
+	[SetupCompileAfter ("Lib.dll", new[] { "Dependencies/TypeForwardedIsUpdatedForMissingTypeFwd.cs" }, references: new[] { "Implementation.dll" })]
+	[RemoveFileAfter ("Implementation.dll")]
```
vitek-karas, line 15: I added `removeFromLinkerInput` to `SetupCompileBefore` for basically exactly this purpose. If we need it on `SetupCompileAfter` it should be easy to add it there as well.
As is - we have two solutions to basically the same problem...

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -154,12 +190,26 @@ protected virtual void SweepAssembly (AssemblyDefinition assembly)
...
+			//
+			// We used to run over list returned by GetTypeReferences but
+			// that returns typeref(s) of original assembly and we don't track
+			// which types are needed for which assembly which left us
+			// with dangling assembly references
+			//
+			assembly.MainModule.AssemblyReferences.Clear ();
+
+			var ars = new AssemblyReferencesCorrector (assembly);
+			ars.Process ();
```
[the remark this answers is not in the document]
marek-safar: Sweep is pretty cheap. On hello world like app the stats is

sweep time before: ~35 ms
sweep time after: ~74 ms
total time: ~4600 ms

that's about 1% increase which I think is fine

src/linker/Linker.Steps/SweepStep.cs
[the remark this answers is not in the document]
marek-safar: I don't think it can throw right now. It can return null when you have `--skip-unresolved` which is handled here and covered by tests.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -728,16 +553,345 @@ protected virtual void ElementRemoved (IMetadataTokenProvider element)
...
+					UpdateForwardedTypesScope (caa);
+					break;
+				case CustomAttributeArgument[] array:
+					foreach (var item in array)
+						UpdateForwardedTypesScope (item);
+					break;
+				}
+			}
+
+			void UpdateTypeScope (TypeReference type)
```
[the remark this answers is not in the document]
marek-safar: Changed to UpdateScopeOfTypeReference

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -72,6 +80,33 @@ void RemoveUnusedAssembly (AssemblyDefinition assembly)
...
+					continue;
+
+				case AssemblyAction.Copy:
+				case AssemblyAction.CopyUsed:
+				case AssemblyAction.AddBypassNGen:
+				case AssemblyAction.AddBypassNGenUsed:
+					// Assembly was removed in the output but it's referenced by
+					// other assembly with action which does not update references
+					if (HasReferenceTo (a, assembly))
+						UpdateAssemblyCopyLikeActions (a, action);
```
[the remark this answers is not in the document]
marek-safar: Done

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -728,16 +553,345 @@ protected virtual void ElementRemoved (IMetadataTokenProvider element)
 		{
 		}
 
-		protected virtual void ReferenceRemoved (AssemblyDefinition assembly, AssemblyNameReference reference)
```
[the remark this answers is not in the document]
marek-safar: If they still needed it they need to move the logic to `SweepAssembly` override and compare the list before and after.

src/linker/Linker.Steps/SweepStep.cs
marek-safar: Yep, missed that one.

src/linker/Linker.Steps/SweepStep.cs
marek-safar: Same as above.

test/Mono.Linker.Tests.Cases/TypeForwarding/TypeForwardedIsUpdatedForMissingType.cs
marek-safar: ok, extended SetupCompileAfter instead

vitek-karas (approved):

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -72,6 +80,71 @@ void RemoveUnusedAssembly (AssemblyDefinition assembly)
...
+				case AssemblyAction.AddBypassNGen:
+				case AssemblyAction.AddBypassNGenUsed:
+					// Assembly was removed in the output but it's referenced by
+					// other assembly with action which does not update references
+					if (!HasReferenceTo (a, assembly))
+						continue;
+
+					switch (action) {
+					case AssemblyAction.CopyUsed:
+						if (IsUsedAssembly (a)) {
```
sbomer, line 105: I think this check is unnecessary because an unused `CopyUsed` assembly would already have been marked `Delete`.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -72,6 +80,71 @@ void RemoveUnusedAssembly (AssemblyDefinition assembly)
...
+						// forwarding assembly (facade).
+						//
+						//		foo.dll -> facade.dll    -> lib.dll
+						//		copy    |  copy (delete) |  link
+						//
+						Annotations.SetAction (a, AssemblyAction.Save);
+						continue;
+
+					case AssemblyAction.AddBypassNGenUsed:
+						if (IsUsedAssembly (a)) {
```
sbomer, line 130: Same here

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -728,16 +550,348 @@ protected virtual void ElementRemoved (IMetadataTokenProvider element)
...
+					// This can happen when not all assembly refences were provided and we
+					// run in `--skip-unresolved` mode. We cannot fully sweep and keep the
+					// original assembly reference
+					//
+					var anr = (AssemblyNameReference) type.Scope;
+					type.Scope = importer.ImportReference (anr);
+					return;
+				}
+
+				var tr = assembly.MainModule.ImportReference (td);
```
sbomer, line 891: For a typeref that resolves to an unused assembly, I think this will keep the assembly reference. We [used to](https://github.com/mono/linker/blob/master/src/linker/Linker.Steps/SweepStep.cs#L295) remove such unused assemblyrefs (at the expense of leaving inconsistent typerefs in unused code). Is this expected?

sbomer (approved):

src/linker/Linker.Steps/SweepStep.cs
marek-safar: The logic here is different, the processing is running only on types/members which are marked (anything unused was removed from the collection before this logic was entered)

src/linker/Linker.Steps/SweepStep.cs
marek-safar: This is for the case where you have an assembly that is `copyused` and it's used (required in the output) and references another assembly that was removed. We cannot do a simple copy in such cases.

sbomer: Right - I am saying that if we get here for the `CopyUsed` action, we already know `IsUsedAssembly` is true - otherwise `RemoveUnmarkedAssembly` would have set the action to `Delete` earlier. You could get rid of the check and just fall through to the `Copy` case.

src/linker/Linker.Steps/SweepStep.cs
sbomer: Couldn't we get here for unmarked members in a `copyused` assembly through `SweepAssemblyReferences`?

src/linker/Linker.Steps/SweepStep.cs
marek-safar: Fixed

src/linker/Linker.Steps/SweepStep.cs
marek-safar: Fixed as well

src/linker/Linker.Steps/SweepStep.cs
marek-safar: We could and in that case we don't care about marked/unmarked because for copy/copyused we want to keep all members, essentially this code processed assembly in its final form.

PR #1689, opened 2020-12-11
Add some more tests for constant propagation

vitek-karas:


sbomer (approved):
LGTM. Just note that delay loading will break the cross-assembly cases per https://github.com/mono/linker/pull/1626/files#diff-ff3b28814148befe0afa0edfaaaf1ac508be245fe00d98ed0fca0674916d3cd3R61.

vitek-karas:
I wanted to capture the cases which already exist in the libraries - if we don't make all these tests work it won't really fully fix the problem. But it's OK if it takes a while.

vitek-karas:
I guess I need help with this one. On my machine it builds just fine for both Windows and Linux - but in CI it consistently fails with:
```console
D:\a\1\s\test\Mono.Linker.Tests.Cases\UnreachableBlock\BodiesWithSubstitutions.cs(15,3): error CS0182: An attribute argument must be a constant expression, typeof expression or array creation expression of an attribute parameter type 
```

I don't see what's wrong with that code ... anybody has an idea.
Or can anybody try this locally?

test/Mono.Linker.Tests.Cases/UnreachableBlock/BodiesWithSubstitutions.cs
```diff
@@ -1,13 +1,21 @@
...
 	[SetupCSharpCompilerToUse ("csc")]
 	[SetupCompileArgument ("/optimize+")]
 	[SetupLinkerArgument ("--enable-opt", "ipconstprop")]
+	[SetupCompileBefore (
+		"LibReturningConstant.dll", 
+		new string[] { "Dependencies/LibReturningConstant.cs" })]
+	[SetupCompileBefore (
+		"LibWithConstantSubstitution.dll", 
+		new string[] { "Dependencies/LibWithConstantSubstitution.cs" }, 
+		resources: new string[] { "Dependencies/LibWithConstantSubstitution.xml" })]
```
tlakollo, line 18: I'm not sure why it doesn't want the string [] in the resources, changing it to object[] like is being used in other parts of the code solve it.

vitek-karas: Thanks for the suggestion - unfortunately it didn't help :-(.

MichalStrehovsky:
Since you're adding tests - do we track missing test coverage for this feature somewhere? I only skimmed the code but the feature tries to do a lot of things (that I would probably just scope out) with only very light test coverage. The things that jumped right out were:

Treating signed and unsigned comparisons like it's the same thing - I think there will be interesting behavior for values over `Int32.MaxValue`:

https://github.com/mono/linker/blob/5232d23ddf96b32e1269a03f561235e280cba140/src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs#L718-L748

Hoping that the last `ret` in the cctor method body will be executed and it's not part of error handling or something like that (for the `initialize` option of a field constant):

https://github.com/mono/linker/blob/5232d23ddf96b32e1269a03f561235e280cba140/src/linker/Linker.Steps/CodeRewriterStep.cs#L58-L71

(If it was up to me, I would fix these by deleting a bunch of code from the feature instead of fixing the bugs and adding test coverage...)

marek-safar (approved):

vitek-karas:
I don't think we have a test coverage for this tracked anywhere.
I fixed the unsigned/signed stuff - cause it was simple, and it's something which actually may exist in framework or elsewhere.

I filed https://github.com/mono/linker/issues/1704 for the rest.

Issue #1498, opened 2020-09-18
Undocumented linker warning IL3000


mateoatr:
`D:\tests\foo\Program.cs(1,26): warning IL3000: 'System.Reflection.Assembly.Location' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'. [D:\tests\foo\foo.csproj]`

Not sure where this warning is coming from, it doesn't seem to be coming out of the linker.
Repro:
```C#
System.Console.WriteLine(System.Reflection.Assembly.GetExecutingAssembly().Location);
```
Publish as single-file: `dotnet publish /p:PublishSingleFile=true -r win-x64`

agocke:
It's a new analyzer I added to the SDK. Looks like it's working! Documented here: https://docs.microsoft.com/en-us/visualstudio/code-quality/il3000?view=vs-2019

mateoatr:
Ah, yes, roslyn-analyzers. Should we keep track of these warning codes here as well?

agocke:
We could mark them reserved, and maybe have a link to the authoritative location, but I don't think we should try to duplicate any info.

marek-safar:
This is pretty bad! Didn't we spend time investigating that there will be no prefix collision to avoid exactly this situation? Using the same error code prefix for two very different tools in the same output will be confusing.

What about changing illinker messages prefix to `TR` while we still can?

MichalStrehovsky:
I think the IL prefix was chosen on purpose: https://github.com/dotnet/roslyn-analyzers/pull/3921 and I would expect we'll want to make sure the warning means the same thing in the linker. For the user it doesn't matter what tool this comes from if the fix is the same.

MichalStrehovsky:
One thing that occurred to me that even though we would share error codes across analyzers and linker, the suppression mechanism is still different.

We might need to ask Roslyn team to do https://github.com/dotnet/roslyn/issues/43639 sooner than "backlog". @agocke can you help with that?

marek-safar:
That still does not make much sense to me. Why would we want to use `IL` prefix for warnings triggered by a different tool? I think it matters for users and developers to know which tool produced the warnings, your argument that it does not matter if the fix is same is like saying C#/VB/F# compilers should not use their own prefixes because if the error is missing assembly reference the fix is always same in VS.

[...]

PR #1691, opened 2020-12-12
Add support for single file analysis

tlakollo:
This change allows the linker to produce warnings using the ReflectionMethodBodyScanner when the flag --single-file-analysis is used. On this first PR is only allowed to print single file warnings only if trimming is also enabled. Future PR should pursue the end goal of enabling linker analysis without trimming, so we could print single file warnings even when the app is not going to be trimmed

-Add flag for single file analysis in the linker
-Add single file dangerous patterns to the ReflectionMethodBodyScanner list
-Add test for single file dangerous patterns

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -246,6 +246,11 @@ namespace Mono.Linker.Dataflow
 			AppDomain_CreateInstanceFrom,
 			AppDomain_CreateInstanceFromAndUnwrap,
 			Assembly_CreateInstance,
+			Assembly_Location,
+			Assembly_CodeBase,
+			Assembly_EscapedCodeBase,
+			Assembly_GetFile,
+			Assembly_GetFiles,
 			RuntimeReflectionExtensions_GetRuntimeEvent,
 			RuntimeReflectionExtensions_GetRuntimeField,
 			RuntimeReflectionExtensions_GetRuntimeMethod,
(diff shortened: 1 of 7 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1227,6 +1257,52 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+					_context.LogWarning ($"'{calledMethodDefinition.GetDisplayName ()}' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.", 3000, callingMethodDefinition);
+					break;
+
+				//
+				// System.Reflection.Assembly
+				//
+				// CodeBase ()
+				//
+				case IntrinsicId.Assembly_CodeBase when _context.SingleFileAnalysis == true:
+					_context.LogWarning ($"'{calledMethodDefinition.GetDisplayName ()}' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.", 3000, callingMethodDefinition);
```
mateoatr, line 1275: This will actually throw, not return a empty string. I think we should add new warnings for this and other APIs (other than Assembly.Location and Assembly.GetFiles).

src/linker/Linker/LinkContext.cs
```diff
@@ -106,6 +107,11 @@ public class LinkContext : IDisposable
 			set { _keepTypeForwarderOnlyAssemblies = value; }
 		}
 
+		public bool SingleFileAnalysis {
```
mateoatr, line 110: nit: similar methods in LinkContext tend to be named Is<Something>, maybe name this `IsSingleFile`?

mateoatr: Also, why expand the getter/settter. 
```C#
public bool SingleFileAnalysis { get; set; }
```

test/Mono.Linker.Tests.Cases/SingleFileAnalysis/WarnOnSingleFileDangerousPatterns.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.SingleFileAnalysis
+{
+	[SetupLinkerArgument ("--single-file-analysis", "true")]
+	public class WarnOnSingleFileDangerousPatterns
```
mateoatr, line 18: If we move this tests here, can we remove them from ILLink.RoslynAnalyzer.Tests and call them in the analyzer with RunTest? Similar to what we already have for RequiresCapability tests.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -246,6 +246,11 @@ enum IntrinsicId
 			AppDomain_CreateInstanceFrom,
 			AppDomain_CreateInstanceFromAndUnwrap,
 			Assembly_CreateInstance,
+			Assembly_Location,
```
mateoatr, line 249: Should we also add AssemblyName_CodeBase, AssemblyName_EscapedCodeBase, Module_Name, and Module_FullyQualifiedName?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
mateoatr: And now that we have these warnings being produced by the linker, we should probably add them to the docs and close https://github.com/mono/linker/issues/1498.

MichalStrehovsky:
What's the value prop for introducing intrinsics for this vs just marking these as SingleFileUnsupported in CoreLib (dotnet/runtime#45705) and have linker understand that attribute?

I think the latter will leave us with less code to maintain and I tend to prefer that.

marek-safar (changes requested):
I think this is a wrong approach for several reasons
- it forces the linker to be always used for single-file analysis
- it hardcodes implementation which could change or evolve
- it ignores runtime pack variations
- it sets the wrong precedent for all other framework APIs that have a similar problem

tlakollo:
Closing in favor of https://github.com/mono/linker/pull/1722

PR #1694, opened 2020-12-14
Report error exit code when any error is reported

marek-safar:
also don't produce any output binaries when error occurs

### DIFF
--- a/src/linker/Linker.Steps/OutputStep.cs
+++ b/src/linker/Linker.Steps/OutputStep.cs
@@ -76,11 +76,15 @@ namespace Mono.Linker.Steps
 			throw new BadImageFormatException ("unrecognized module attributes");
 		}
 
+		protected override bool ConditionToProcess ()
+		{
+			return Context.ErrorsCount == 0;
+		}
+
 		protected override void Process ()
 		{
 			CheckOutputDirectory ();
 			OutputPInvokes ();
-			OutputSuppressions ();
 			Tracer.Finish ();
 		}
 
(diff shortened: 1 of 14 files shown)


src/linker/Linker.Steps/OutputStep.cs
```diff
@@ -76,11 +76,15 @@ static TargetArchitecture CalculateArchitecture (TargetArchitecture readyToRunAr
 			throw new BadImageFormatException ("unrecognized module attributes");
 		}
 
+		protected override bool ConditionToProcess ()
```
mateoatr, line 79: I think after ProcessWarningsStep, there's no step that could produce any warning/error. Should these steps have this condition as well, or maybe short-circuit steps after this point if errors were found?

src/linker/Linker/Driver.cs
```diff
@@ -725,6 +721,7 @@ protected int SetupContext (ILogger customLogger = null)
 			// MarkStep
 			// ReflectionBlockedStep [optional]
 			// ProcessWarningsStep
+			// OutputWarningSuppressions
```
mateoatr, line 724: nit: I think this doesn't match its actual place in the pipeline.

marek-safar: yeah, forgot to update the code. The comment is correct.

src/linker/Linker.Steps/OutputStep.cs
marek-safar: I don't think it's worth it (optimize for failure), they should not produce errors but they could in the future.

vitek-karas (approved):

Issue #1623, opened 2020-11-10
Harden Intrinsic Reflection API Matching


eerhardt:
When `dotnet/runtime` adds new Reflection APIs (or adds overloads to existing APIs), it is possible that the new API / overload may break the ILLinker. An example of this is #1621.

Today, we the ILLinker does a basic test for intrinsic Reflection APIs:

https://github.com/mono/linker/blob/d8cbe373195ed3fc727d12d12eb5923b188a7990/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs#L326-L346

As long as the method is on `System.Type` and named `GetConstructor` (or named `GetMethod` and has a string as a first parameter), the ILLinker will treat the method as "intrinsic".

We should make sure this intrinsic handling works as `dotnet/runtime` adds new APIs and overloads to `System.Type`.

cc @vitek-karas

marek-safar:
ILLinker could also run with a custom SPC which is rare but we have seen it before

vitek-karas:
The goal should be that linker does full signature match for each intrinsic it recognizes. Since only then it can know the expected behavior of the API. The current matching which is very broad leaves holes and it means sometimes linker assumes different behavior that what runtime does.

This will be mostly syntax challenge (so that the code is easy to read), perf is not a big problem since reflection calls are rare and we would do namespace/type/method checks first before checking the signature.

PR #1701, opened 2020-12-15
Mark everything if BindingFlags are not recognized by the linker

### LINKED ISSUE #1617: Non constant BindingFlags causes ILLinker to trim required methods
When a non-public method is called through Reflection, but the `BindingFlags` parameter is not a constant, the ILLinker won't preserve the called method, and it won't log a warning. Thus the user thinks their app will work successfully, but the app doesn't since the called method is trimmed.

This scenario was found in System.Private.Xml and is one of the underlying causes of issue https://github.com/dotnet/runtime/issues/41389.

https://github.com/dotnet/runtime/blob/819a3cc69e2bf606787a44f9f107ebbbed0fcb11/src/libraries/System.Private.Xml/src/System/Xml/Serialization/CodeGenerator.cs#L24-L28

### Example

Fully trim the following application:

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>

  <Target Name="EnsureAllAssembliesAreLinked"
        BeforeTargets="PrepareForILLink">
    <ItemGroup>
      <ManagedAssemblyToLink>
        <TrimMode>link</TrimMode>
      </ManagedAssemblyToLink>

      <!-- Pass the app assembly as a root -->
      <TrimmerRootAssembly Include="@(IntermediateAssembly)" />
    </ItemGroup>
  </Target>
</Project>
```

```C#
    internal class CodeGenerator
    {
        internal static BindingFlags InstanceBindingFlags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic;
    }

    public abstract class XmlSerializationWriter
    {
        protected void WriteStartDocument()
        {
        }
    }

    internal class Program
    {
        public static void Main()
        {
            MethodInfo XmlSerializationWriter_WriteStartDocument = typeof(XmlSerializationWriter).GetMethod(
                "WriteStartDocument",
                CodeGenerator.InstanceBindingFlags,
                null,
                Array.Empty<Type>(),
                null
                )!;
            Console.WriteLine(XmlSerializationWriter_WriteStartDocument.Name);
        }
    }
```

When trimmed, the application will throw:

```
   at ConsoleApp111.Program.Main() in C:\Users\eerhardt\source\repos\ConsoleApp111\ConsoleApp111\Program.cs:line 29
```

cc @vitek-karas

tlakollo:
-Add a flag to mark when the overload has a binding flag that the linker could not figure out
-Mark everything when an unknown binding flag is present
-Add tests

Fixes #1617

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -847,9 +847,12 @@ namespace Mono.Linker.Dataflow
 						reflectionContext.AnalyzingPattern ();
 
 						var parameters = calledMethod.Parameters;
-						BindingFlags? bindingFlags = BindingFlags.Public | BindingFlags.Instance;
+						BindingFlags? bindingFlags;
 						if (parameters.Count > 1 && calledMethod.Parameters[0].ParameterType.Name == "BindingFlags")
-							bindingFlags = GetBindingFlags (methodParams[1]);
+							bindingFlags = GetBindingFlagsFromValue (methodParams[1]);
+						else
+							// Assume a default value for BindingFlags for methods that don't use BindingFlags as a parameter
+							bindingFlags = BindingFlags.Public | BindingFlags.Instance;
 
 						int? ctorParameterCount = parameters.Count switch
 						{
@@ -895,11 +898,14 @@ namespace Mono.Linker.Dataflow
 				case IntrinsicId.Type_GetMethod: {
 						reflectionContext.AnalyzingPattern ();
 
-						BindingFlags? bindingFlags = BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public;
+						BindingFlags? bindingFlags;
 						if (calledMethod.Parameters.Count > 1 && calledMethod.Parameters[1].ParameterType.Name == "BindingFlags")
-							bindingFlags = GetBindingFlags (methodParams[2]);
+							bindingFlags = GetBindingFlagsFromValue (methodParams[2]);
 						else if (calledMethod.Parameters.Count > 2 && calledMethod.Parameters[2].ParameterType.Name == "BindingFlags")
-							bindingFlags = GetBindingFlags (methodParams[3]);
+							bindingFlags = GetBindingFlagsFromValue (methodParams[3]);
+						else
+							// Assume a default value for BindingFlags for methods that don't use BindingFlags as a parameter
+							bindingFlags = BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public;
 						var requiredMemberTypes = GetDynamicallyAccessedMemberTypesFromBindingFlagsForMethods (bindingFlags);
 						foreach (var value in methodParams[0].UniqueValues ()) {
 							if (value is SystemTypeValue systemTypeValue) {
@@ -930,9 +936,12 @@ namespace Mono.Linker.Dataflow
 				case IntrinsicId.Type_GetNestedType: {
 						reflectionContext.AnalyzingPattern ();
 
-						BindingFlags? bindingFlags = BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public;
+						BindingFlags? bindingFlags;
 						if (calledMethod.Parameters.Count > 1 && calledMethod.Parameters[1].ParameterType.Name == "BindingFlags")
-							bindingFlags = GetBindingFlags (methodParams[2]);
+							bindingFlags = GetBindingFlagsFromValue (methodParams[2]);
+						else
+							// Assume a default value for BindingFlags for methods that don't use BindingFlags as a parameter
+							bindingFlags = BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public;
 
 						var requiredMemberTypes = GetDynamicallyAccessedMemberTypesFromBindingFlagsForNestedTypes (bindingFlags);
 						foreach (var value in methodParams[0].UniqueValues ()) {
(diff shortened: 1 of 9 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -863,7 +865,10 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 						// Go over all types we've seen
 						foreach (var value in methodParams[0].UniqueValues ()) {
 							if (value is SystemTypeValue systemTypeValue) {
-								MarkConstructorsOnType (ref reflectionContext, systemTypeValue.TypeRepresented, null, bindingFlags);
+								if (BindingFlagsAreSupported (bindingFlags) || unknownBindingFlags)
```
vitek-karas, line 868: I somehow missed this in the original PR - the method should be called `BindingFlagsAreUnsupported`

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -848,10 +848,12 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
-						if (parameters.Count > 1) {
-							if (methodParams[1].AsConstInt () != null)
-								bindingFlags |= (BindingFlags) methodParams[1].AsConstInt ();
-						}
+						bool unknownBindingFlags = false;
+						if (parameters.Count > 1 && calledMethod.Parameters[0].ParameterType.Name == "BindingFlags" && methodParams[1].AsConstInt () != null)
+							bindingFlags |= (BindingFlags) methodParams[1].AsConstInt ();
+						else if (parameters.Count > 1 && calledMethod.Parameters[0].ParameterType.Name == "BindingFlags" && methodParams[1].AsConstInt () == null)
+							unknownBindingFlags = true;
+
```
vitek-karas, line 856: I think I would probably prefer some kind of if/switch over all the overloads - and specifically set binding flags in those cases where they are specified (And maybe have a helper which checks the constant value of that parameter). And then specifically set them to the right defaults for overloads which don't specify them.

We could also use nullable to avoid the extra boolean flag.

tlakollo: I think I got everything solved except for the part of an if/switch that goes over all the overloads, I think that asking for `ParameterType.Name == "BindingFlags"` is a good optimized way to solve if we are dealing with BindingFlags and doesn't require to go over all overloads.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1672,10 +1654,11 @@ void RequireDynamicallyAccessedMembers (ref ReflectionPatternContext reflectionC
 			reflectionContext.RecordHandledPattern ();
 		}
 
-		static bool BindingFlagsAreSupported (BindingFlags bindingFlags)
-		{
-			return (bindingFlags & BindingFlags.IgnoreCase) == BindingFlags.IgnoreCase || (int) bindingFlags > 255;
-		}
+		static BindingFlags? GetBindingFlags (ValueNode parameter) => (BindingFlags?) parameter.AsConstInt ();
```
vitek-karas, line 1657: Nit: I would rename this to `GetBindingFlagsFromValue` - just to make a bit more explicit.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: The overloads which do have `BindingFlags` parameter are the easy ones (Since they explicitly state binding flags). I'm more worried about the overloads which don't take the parameter. I guess we rely on the fact that those all have the same behavior - which might be OK. In that case I think we should at the very least add a comment about that. To make it clear it was a conscious decision. This is very much related to https://github.com/mono/linker/issues/1623 - that's why I'm very picky about this topic :-)

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -847,11 +847,10 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
 
 						var parameters = calledMethod.Parameters;
-						BindingFlags bindingFlags = BindingFlags.Public | BindingFlags.Instance;
-						if (parameters.Count > 1) {
-							if (methodParams[1].AsConstInt () != null)
-								bindingFlags |= (BindingFlags) methodParams[1].AsConstInt ();
-						}
+						BindingFlags? bindingFlags = BindingFlags.Public | BindingFlags.Instance;
+						if (parameters.Count > 1 && calledMethod.Parameters[0].ParameterType.Name == "BindingFlags")
+							bindingFlags = GetBindingFlags (methodParams[1]);
```
marek-safar, line 852: I think it'd be nicer to write such initialisations as if/else. Following is easier for me to read than the double initialization

```c#
	BindingFlags? bindingFlags;
	if (parameters.Count > 1 && calledMethod.Parameters[0].ParameterType.Name == "BindingFlags")
		bindingFlags = GetBindingFlags (methodParams[1]);
	else
		bindingFlags = BindingFlags.Public | BindingFlags.Instance;

```

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
tlakollo: I added a comment to reflect that we are using a default value for methods that don't use BindingFlags

vitek-karas (approved):

PR #1702, opened 2020-12-16
Package ILLink.RoslynAnalyzer for SDK

mateoatr:
This packs the analyzer build so that we can integrate it with the SDK and remove the single-file analyzer from dotnet/roslyn-analyzers.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
+++ b/src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
@@ -6,14 +6,27 @@
     <OutputType>Library</OutputType>
     <Nullable>enable</Nullable>
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
+    <AutoGenerateAssemblyVersion>true</AutoGenerateAssemblyVersion>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+  </PropertyGroup>
+
+  <!-- Analyzer package properties -->
+  <PropertyGroup>
     <IsPackable>true</IsPackable>
     <PackageId>Microsoft.NET.ILLink.Analyzers</PackageId>
-    <AutoGenerateAssemblyVersion>true</AutoGenerateAssemblyVersion>
+    <Authors>Microsoft</Authors>
+    <RepositoryUrl>https://github.com/mono/linker</RepositoryUrl>
+    <Description>Description of the package</Description>
+    <NoPackageAnalysis>true</NoPackageAnalysis>
   </PropertyGroup>
 
   <ItemGroup>
     <EmbeddedResource Include="Resources.resx" GenerateSource="true" />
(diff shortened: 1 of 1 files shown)


sbomer:
I think we need to ensure that the analyzer dll gets placed in `analyzers/dotnet/cs` in the package to match the [convention](https://docs.microsoft.com/en-us/nuget/guides/analyzers-conventions), something like this:
 https://github.com/vitek-karas/linker/pull/7/files#diff-cb5218b72194e60ffe7501d384a7dd5cb3f6191d76901cf3717f54cc5f560e15R38.

This [sample](https://github.com/dotnet/roslyn-sdk/blob/master/samples/CSharp/Analyzers/Analyzers.Implementation/Analyzers.CSharp.csproj) and these [docs](https://roslyn-analyzers.readthedocs.io/en/latest/create-nuget-package.html#using-support-from-the-csproj) are probably good examples to follow.

src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
```diff
@@ -2,14 +2,33 @@
...
   </PropertyGroup>
 
   <ItemGroup>
     <EmbeddedResource Include="Resources.resx" GenerateSource="true" />
-    <PackageReference Include="microsoft.codeanalysis.csharp" Version="3.7.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.7.0" PrivateAssets="all" />
   </ItemGroup>
 
-</Project>
+  <Target Name="IncludeAnalyzerAssembliesInPkg" AfterTargets="AfterBuild">
```
agocke, line 28: Might want to do `DependsOn=Build` and `BeforeTargets=Pack` here, since this needs to run before `pack`

mateoatr: I've added a `BeforeTargets`. The Build target already depends on `AfterBuild`, which would cause a circular dependency if added too.

src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
```diff
@@ -2,14 +2,35 @@
...
   <ItemGroup>
     <EmbeddedResource Include="Resources.resx" GenerateSource="true" />
-    <PackageReference Include="microsoft.codeanalysis.csharp" Version="3.7.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.7.0" PrivateAssets="all" />
   </ItemGroup>
 
-</Project>
+  <Target Name="IncludeAnalyzerAssembliesInPkg"
+          AfterTargets="AfterBuild"
+          BeforeTargets="_GetPackageFiles">
```
agocke, line 30: Hmm I don't think we can do this. We don't own the `_GetPackageFiles` target and it's `_` prefixed, so we can't depend on it

agocke (approved):
LGTM assuming the SDK actually includes the analyzer :)

PR #1705, opened 2020-12-16
Add a test for attribute removal using wildcard and lazy loading

vitek-karas:
Currently this doesn't work as it should, the test is to cover the scenario - once we fully implement lazy loading this should be fixed.

### DIFF
--- a/test/Mono.Linker.Tests.Cases/LinkAttributes/Dependencies/LinkerAttributeRemovalEmbeddedAndLazyLoad.cs
+++ b/test/Mono.Linker.Tests.Cases/LinkAttributes/Dependencies/LinkerAttributeRemovalEmbeddedAndLazyLoad.cs
@@ -1,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Mono.Linker.Tests.Cases.LinkAttributes.Dependencies
+{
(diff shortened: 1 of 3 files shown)


marek-safar (approved):

sbomer:
Didn't we decide that the wildcard was only supported for corelib? https://github.com/mono/linker/pull/1675 intentionally removed support for `*` in embedded XML except corelib. I would expect that the attribute isn't removed in this case, with or without lazy loading.

vitek-karas:
Good point - that basically means there's no way to write a test for it, right? (I can't fake corelib)
But the functionality is still broken...
Should we maybe add some test hook to enable it on other assemblies so that it can be tested?

Also - this pretty much means that the LinkerAttributeRemoval test should not work, right? So how come it's passing?

sbomer:
It's a bit of a hack, but I managed to write a test that mocks corelib: https://github.com/mono/linker/pull/1675/files#diff-6bc3fb243979b37206fe4be289fcc1c77b1b6ec8173e7e63a090603639dd7c75R20. It fails PeVerify because system types can't be resolved, but it does check the attribute removal in other assemblies.

As to why the test is passing - I just noticed it's actually not using embedded xml. :) SetupLinkAttributesFile will use the command-line xml.

vitek-karas:
Good point about the file not being embedded - in that case the test is perfectly valid, right?

sbomer:
Right, good point. And it's a good test to have, since I have been focusing mostly on the embedded XML.

sbomer (approved):

PR #1706, opened 2020-12-17
Support for RequiresAndCallsOtherRequiresMethods

tlakollo:
-Check for the parent to see if it has any UnreferencedCodeAttribute
-Since AllowMultiple is disabled by default, only check the attributes until first UnreferencedCodeAttribute is found

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
@@ -76,26 +76,22 @@ namespace ILLink.RoslynAnalyzer
 					IMethodSymbol method,
 					Location location)
 				{
-					var attributes = method.GetAttributes ();
-
-					foreach (var attr in attributes) {
-						if (attr.AttributeClass is { } attrClass &&
-							IsNamedType (attrClass, "System.Diagnostics.CodeAnalysis.RequiresUnreferencedCodeAttribute") &&
-							attr.ConstructorArguments.Length == 1 &&
-							attr.ConstructorArguments[0] is { Type: { SpecialType: SpecialType.System_String } } ctorArg) {
-
-							string urlArgument = "";
-							foreach (var namedArgument in attr.NamedArguments) {
-								if (namedArgument.Key == "Url")
-									urlArgument = namedArgument.Value.Value?.ToString () ?? "";
-							}
-							operationContext.ReportDiagnostic (Diagnostic.Create (
-								s_rule,
-								location,
-								method.OriginalDefinition.ToString (),
-								(string) ctorArg.Value!,
-								urlArgument));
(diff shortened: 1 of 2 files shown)


src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -76,26 +76,22 @@ public override void Initialize (AnalysisContext context)
...
-									urlArgument = namedArgument.Value.Value?.ToString () ?? "";
-							}
-							operationContext.ReportDiagnostic (Diagnostic.Create (
-								s_rule,
-								location,
-								method.OriginalDefinition.ToString (),
-								(string) ctorArg.Value!,
-								urlArgument));
+					AttributeData? attribute = GetRequiresUnreferencedCodeAttribute (operationContext.ContainingSymbol.GetAttributes ());
+					if (attribute != null)
```
vitek-karas, line 80: Please add a comment why we want to "skip" in this case.

vitek-karas (approved):

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -76,26 +76,22 @@ public override void Initialize (AnalysisContext context)
...
+						return;
+					attribute = GetRequiresUnreferencedCodeAttribute (method.GetAttributes ());
+					if (attribute != null) {
+						string urlArgument = "";
+						foreach (var namedArgument in attribute.NamedArguments) {
+							if (namedArgument.Key == "Url")
+								urlArgument = namedArgument.Value.Value?.ToString () ?? "";
```
mateoatr, line 87: nit: why not use this in the declaration of `urlArgument`? Something like `attribute.NamedArguments.FirstOrDefault (na => na.Key == "Url").Value.Value?.ToString ()` then we don't need the `foreach`.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -121,5 +117,21 @@ internal static bool IsNamedType (INamedTypeSymbol type, string typeName)
 
 			return true;
 		}
+
+		/// <summary>
+		/// Returns a RequiresUnreferencedCodeAttribute if found
+		/// </summary>
+		static AttributeData? GetRequiresUnreferencedCodeAttribute (ImmutableArray<AttributeData> attributes)
```
mateoatr, line 124: nit: `RequiresUnreferencedCode` can only be placed in methods, so why not limit this to `IMethodSymbol`? We are already checking this only for methods since we only call this from within `CheckMethodOrCtorCall`. We could also change this to a `TryGet...` and return the attribute in an `out` and then use this above as `if (TryGet...(method, out requiresUnreferncedCode) { ... }`.

PR #1712, opened 2020-12-22
Implement marking of custom marshaler required members

### LINKED ISSUE #1636: Expand DynamicDependencyAttribute handling with interfaces
We have the following cases today which have to use linker descriptors and dependencies are always kept even though they are rarely used and should be trimmed to reduce the size.

All cases are a mix of managed and native code which need to cooperate but linker has no visibility to what native code requires/depends on.

## ICustomMarshaler case

We need to keep `ICustomMarshaler` methods and their implementations for any type which exists in the final output (we could go beyond that but that's probably stretch). Then if any of the interface methods are kept we need to also keep managed logic in `Marshal.GetCustomMarshalerInstance` method.

Proposed extensions

```c#
[DynamicDependencyAttribute("MarshalNativeToManaged")]
[DynamicDependencyAttribute("MarshalManagedToNative")]
[DynamicDependencyAttribute("CleanUpNativeData")]
[DynamicDependencyAttribute("CleanUpManagedData")]
[DynamicDependencyAttribute("GetNativeDataSize")]
public interface ICustomMarshaler
{
	// Keep the dependency if it's never used in managed code
	[DynamicDependencyAttribute("Marshal.GetCustomMarshalerInstance")]
    object MarshalNativeToManaged(IntPtr pNativeData);
    IntPtr MarshalManagedToNative(object ManagedObj);
    void CleanUpNativeData(IntPtr pNativeData);
    void CleanUpManagedData(object ManagedObj);
    int GetNativeDataSize();
}
```

## Implicit Array interfaces

For MonoVM we have implementations of implicit array method done via magic mapping between vtable and managed method implementation. This goes a bit against the case above because ideally, we would like to keep the methods only if they are used.

Proposed extension

```c#
public interface IReadOnlyCollection<out T> : IEnumerable<T>
{
    [DynamicDependency("InternalArray__IReadOnlyCollection_get_Count", typeof(Array))]
    int Count { get; }
}
```

This one is trickier as ideally the dependency should be kept only when Array is kept but in this particular case that's very unlikely.

@vitek-karas

marek-safar:
Fixes #1636

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -588,7 +588,13 @@ namespace Mono.Linker.Steps
 
 			if (spec.MarshalInfo is CustomMarshalInfo marshaler) {
 				MarkType (marshaler.ManagedType, reason, sourceLocationMember);
-				MarkGetInstanceMethod (marshaler.ManagedType.Resolve (), in reason, sourceLocationMember);
+				TypeDefinition type = marshaler.ManagedType.Resolve ();
+				if (type != null) {
+					MarkICustomMarshalerMethods (type, in reason, sourceLocationMember);
+					MarkCustomMarshalerGetInstance (type, in reason, sourceLocationMember);
+				} else {
+					HandleUnresolvedType (marshaler.ManagedType);
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2007,16 +2013,62 @@ protected bool MarkDefaultConstructor (TypeDefinition type, in DependencyInfo re
...
+								"MarshalNativeToManaged" when ptype == MetadataType.IntPtr => true,
+								"MarshalManagedToNative" when ptype == MetadataType.Object => true,
+								"CleanUpNativeData" when ptype == MetadataType.IntPtr => true,
+								"CleanUpManagedData" when ptype == MetadataType.Object => true,
+								_ => false
+							};
+						}
+
+						return m.Name == "GetNativeDataSize" && !m.HasParameters;
+					}, reason, sourceLocationMember);
```
MichalStrehovsky, line 2064: Is there a scenario where this extra code would be useful? I usually go by "less is better" philosophy.

```suggestion
					MarkMethodsIf (tdef.Methods, m => !m.IsStatic, reason, sourceLocationMember);
```

MichalStrehovsky (approved):

PR #1714, opened 2020-12-27
Update scope of modreq/modopt types

marek-safar:


marek-safar:
Merging to unblock runtime bump

PR #1719, opened 2021-01-01
Better intrinsic handling of GetNestedType if input has All

vitek-karas:
If the parent type value is annotated with `DynamicallyAccessedMemberTypes.All` then the GetNestedType can return a value also annotated with `DynamicallyAccessedMemberTypes.All` since the `All` is recursively applied to nested types (`MarkStep.MarkEntireType` basically recursively calls `MarkEntireType` on nested types).

This helps with patterns like:
```C#
void Test([DynamicallyAccessedMembers(All)] parentType)
{
    parentType.GetNestedType("KnownNestedType").GetFields();
}
```

Without this there would be no way to do this without warning suppressions, even though it's a perfectly valid code as per the rules around `All`.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -7,7 +7,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
-using System.Text;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using Mono.Linker.Steps;
@@ -959,6 +958,7 @@ namespace Mono.Linker.Dataflow
 							bindingFlags = BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public;
 
 						var requiredMemberTypes = GetDynamicallyAccessedMemberTypesFromBindingFlagsForNestedTypes (bindingFlags);
+						bool everyParentTypeHasAll = true;
 						foreach (var value in methodParams[0].UniqueValues ()) {
 							if (value is SystemTypeValue systemTypeValue) {
 								foreach (var stringParam in methodParams[1].UniqueValues ()) {
@@ -984,7 +984,18 @@ namespace Mono.Linker.Dataflow
 								// Otherwise fall back to the bitfield requirements
 								RequireDynamicallyAccessedMembers (ref reflectionContext, requiredMemberTypes, value, calledMethodDefinition);
 							}
+
+							if (value is not LeafValueWithDynamicallyAccessedMemberNode { DynamicallyAccessedMemberTypes: DynamicallyAccessedMemberTypes.All })
+								everyParentTypeHasAll = false;
 						}
+
+						// If the parent type (all the possible values) has DynamicallyAccessedMemberTypes.All it means its nested types are also fully marked
+						// (see MarkStep.MarkEntireType - it will recursively mark entire type on nested types). In that case we can annotated 
+						// the returned type (the nested type) with DynamicallyAccessedMemberTypes.All as well.
+						// Note it's OK to blindly overwrite any potential annotation on the return value from the method definition
+						// since DynamicallyAccessedMemberTypes.All is a superset of any other annotation.
+						if (everyParentTypeHasAll && methodReturnValue == null)
+							methodReturnValue = new MethodReturnValue (calledMethodDefinition.MethodReturnType, DynamicallyAccessedMemberTypes.All);
 					}
 					break;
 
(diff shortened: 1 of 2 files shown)


vitek-karas:
Needs more tests.

@LakshanF - a WIP PR with the improvement to intrinsic handling of `GetNestedType`. Can you please add some description (and possibly links) to the event source case where this is needed?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -984,7 +984,18 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 								// Otherwise fall back to the bitfield requirements
 								RequireDynamicallyAccessedMembers (ref reflectionContext, requiredMemberTypes, value, calledMethodDefinition);
 							}
+
+							if (value is not LeafValueWithDynamicallyAccessedMemberNode { DynamicallyAccessedMemberTypes: DynamicallyAccessedMemberTypes.All })
+								everyParentTypeHasAll = false;
 						}
+
+						// If the parent type (all the possible values) has DynamicallyAccessedMemberTypes.All it means its nested types are also fully marked
+						// (see MarkStep.MarkEntireType - it will recursively mark entire type on nested types). In that case we can annotated
```
tlakollo, line 993: I think there is a typo and should be 'annotate' instead of 'annotated'

test/Mono.Linker.Tests.Cases/DataFlow/GetNestedTypeOnAllAnnotatedType.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+
+namespace Mono.Linker.Tests.Cases.DataFlow
+{
+	[RecognizedReflectionAccessPattern]
+	[SkipKeptItemsValidation]
```
tlakollo, line 16: I didn't found any test that verifies the scenario when we want to keep something inside a nested type. Maybe we can have two nested classes and use the Kept item validation to verify that things are kept in one but not in the other

tlakollo:
Here is the snippet of the event source issue
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L2949-L2961

test/Mono.Linker.Tests.Cases/DataFlow/GetNestedTypeOnAllAnnotatedType.cs
tlakollo: Nevermind, I noticed that the GetNestedType does keep all the information inside the nested class

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -984,7 +984,18 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 								// Otherwise fall back to the bitfield requirements
 								RequireDynamicallyAccessedMembers (ref reflectionContext, requiredMemberTypes, value, calledMethodDefinition);
 							}
+
+							if (value is not LeafValueWithDynamicallyAccessedMemberNode { DynamicallyAccessedMemberTypes: DynamicallyAccessedMemberTypes.All })
```
tlakollo, line 988: Mono run uses an analyzer that complains about the way this statement is written, so I will change it a bit

MichalStrehovsky:
I think we discussed nested types in the past but I can't find the conversation - the current behavior where `DynamicallyAccessedMembers.NestedTypes` only ensures that the nested type itself is kept, but without any members is kind of useless. I can't come up with a scenario where that behavior would be useful.

I think we would want to make `DynamicallyAccessedMembers.NestedTypes` mean "keep the nested types in their entirety". That way `GetNestedType` can always return things annotated as `.All` (it's not conditioned on having `.All` on the owning type). The outside behavior of `.All` would be the same as this PR is going for (because All includes NestedTypes), but the implementation will be different.

LakshanF:
@vitek-karas, a WIP PR for the EventSource.cs warnings with nested type is [here](https://github.com/dotnet/runtime/pull/46510)

vitek-karas:
I think this is good to go - I don't want to block this on the decision around nested type annotations (as per @MichalStrehovsky comment). If we decide to do that, the product change in this would be different, but all the tests should continue to work, so I think it's worth putting this into the product now anyway.

marek-safar (approved):

tlakollo:
While reviewing the Xamarin warnings I found a similar one to the event source, and with the current solution would need annotations. As of this moment produces an IL2072 because after GetNestedType there is a call to GetRuntimeMethods
https://github.com/xamarin/java.interop/blob/2244407d04f9e344dfbe76efc8d735cf2a82cef6/src/Java.Interop/Java.Interop/JniRuntime.JniTypeManager.cs#L268

PR #1725, opened 2021-01-06
Clean up of trimming roots handling

marek-safar:
to make running linker with multiple root assemblies easier. This changes
included several improvements.

- Add error handling for invalid inputs
- Make configurations with multiple different roots/inputs assemblies
    behavior predictable
- Allow/Fix -p option to alter inputs/roots trimming behaviour by respecting
    assembly action
- Avoid building huge marking queue when rooting a lot of inputs
- Expose rooting via entry point and visible members as separate task settings


marek-safar:
@vitek-karas for some reason I cannot set you as a reviewer

docs/error-codes.md
```diff
@@ -113,6 +113,18 @@ the error code. For example:
 
 #### `IL1030`: Invalid argument for '{token}' option
 
+#### `IL1031`: Invalid assembly action '{action}'
+
+#### `IL1032`: Trimming assembly '{assemblyFile}' could not be found
```
vitek-karas, line 118: Nit: I don't really understand the meaning of "trimming assembly" - does it mean the same as "assembly which is being trimmed"?

I guess I don't see why we need to prefix these with "Trimming" - we don't do that for the other error messages either...

src/linker/Linker.Steps/TrimUsingEntryPoint.cs
```diff
@@ -0,0 +1,64 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.IO;
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class TrimUsingEntryPoint : BaseStep
```
vitek-karas, line 10: I think it would be better to call this `RootEntryPointStep` or something like that... "trim using entry point" feels weird - it acts on a single assembly, but this step doesn't actually do the trimming - so it's confusing.

src/linker/Linker.Steps/TrimUsingVisibleMembers.cs
```diff
@@ -0,0 +1,76 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.IO;
+using System.Runtime.CompilerServices;
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class TrimUsingVisibleMembers : BaseStep
```
vitek-karas, line 12: Same comment as for the entry point one - maybe call this `RootVisibleMembersStep`?

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/InterfaceTypeInOtherUsedOnlyBySavedAssembly.cs
```diff
@@ -4,6 +4,8 @@
 
 namespace Mono.Linker.Tests.Cases.Inheritance.Interfaces.OnReferenceType
 {
+	[IgnoreTestCase ("save action is not allowed")]
```
vitek-karas, line 7: Why not just delete the test case (or test that it fails)?

test/Mono.Linker.Tests.Cases/Libraries/CopyAssemblyWithMainEntryRoot.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Mono.Linker.Tests.Cases.Libraries
+{
+#if !NETCOREAPP
+	[IgnoreTestCase ("Correctly handled by illink only")]
+#endif
+	[Kept]
+	[KeptMember (".ctor()")]
+	[SetupLinkerAction ("copyused", "test")]
+	public class CopyAssemblyWithMainEntryRoot
```
vitek-karas, line 12: Should the name be `CopyUsedASsemblyWithMainEntryRoot` ? since the test runs it with "copyused" action.

test/Mono.Linker.Tests.Cases/Libraries/CopyAssemblyWithPublicRoots.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Mono.Linker.Tests.Cases.Libraries
+{
+#if !NETCOREAPP
+	[IgnoreTestCase ("Correctly handled by illink only")]
+#endif
+	[Kept]
+	[KeptMember (".ctor()")]
+	[SetupLinkerAction ("copyused", "test")]
+	[SetupLinkerLinkPublicAndFamily]
+	public class CopyAssemblyWithPublicRoots
```
vitek-karas, line 13: "CopyUsed" instead?

test/Mono.Linker.Tests.Cases/Libraries/DefaultLibraryLinkBehavior.cs
```diff
@@ -4,6 +4,9 @@
 namespace Mono.Linker.Tests.Cases.Libraries
 {
 	[SetupCompileAsLibrary]
+#if NETCOREAPP
+	[IgnoreTestCase("Requires better testing framework support when no files are produced")]
```
vitek-karas, line 8: Should at least create an issue for this....

src/linker/Linker.Steps/TrimUsingVisibleMembers.cs
```diff
@@ -0,0 +1,76 @@
...
+				return;
+			case AssemblyAction.Link:
+				break;
+			default:
+				Context.LogError ($"Trimming assembly '{assembly.Name}' cannot use action '{action}'", 1035);
+				return;
+			}
+
+			TypePreserve preserve = TypePreserve.All |
+				(Annotations.HasLinkerAttribute<InternalsVisibleToAttribute> (assembly) ? TypePreserve.AccessibilityVisibleOrInternal : TypePreserve.AccessibilityVisible);
```
vitek-karas, line 41: This will not work unless we teach the linker attributes about this attribute - in the `LinkerAttributesInformation` .ctor - there's an if/else over all recognized attributes - we would have to add it there.

This also probably means that we're missing a test case for this.

src/linker/Linker/Driver.cs
```diff
@@ -516,27 +519,86 @@ protected int SetupContext (ILogger customLogger = null)
...
+							}
+
+							context.RegisterAssemblyAction (assemblyName, action.Value);
+							continue;
+						}
 					case "t":
 						context.KeepTypeForwarderOnlyAssemblies = true;
 						continue;
+#if FEATURE_ILLINK
+					case "x": {
```
sbomer, line 562: Can we reuse the "x" case from monolinker? Could we just remove the call to `GetFiles`, or does monolinker really need the '@descriptor.xml' support?

src/linker/Linker/Driver.cs
```diff
@@ -700,9 +772,7 @@ protected int SetupContext (ILogger customLogger = null)
 			//
 			// Pipeline setup with all steps enabled
 			//
-			// ResolveFromAssemblyStep [optional, possibly many]
-			// ResolveFromXmlStep [optional, possibly many]
-			// [mono only] ResolveFromXApiStep [optional, possibly many]
```
sbomer, line 705: nit: instead of removing these, could you indicate that they're mono only?

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/InterfaceTypeInOtherUsedOnlyBySavedAssembly.cs
sbomer, line 7: I am glad to see this going away :)

src/linker/Linker/Annotations.cs
```diff
@@ -468,7 +468,7 @@ public bool SetPreservedStaticCtor (TypeDefinition type)
 			return marked_types_with_cctor.Add (type);
 		}
 
-		public bool HasLinkerAttribute<T> (IMemberDefinition member) where T : Attribute
+		public bool HasLinkerAttribute<T> (ICustomAttributeProvider member) where T : Attribute
```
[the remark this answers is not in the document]
marek-safar: Reverted the change

src/linker/Linker/Driver.cs
marek-safar: tweaked the text differently

src/linker/Linker/Driver.cs
marek-safar: Trying not to touch monolinker code in this PR

src/linker/Linker/Driver.cs
```diff
@@ -516,27 +519,86 @@ protected int SetupContext (ILogger customLogger = null)
...
+
+							if (!File.Exists (assemblyFile)) {
+								context.LogError ($"Trimming assembly '{assemblyFile}' could not be found'", 1032);
+								return -1;
+							}
+
+							inputs.Add (new TrimUsingEntryPoint (assemblyFile));
+							continue;
+						}
+					case "l": {
```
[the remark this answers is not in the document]
marek-safar: I don't think it's worth at this point (will go away soon)

src/linker/Linker.Steps/TrimUsingVisibleMembers.cs
```diff
@@ -0,0 +1,76 @@
...
+		{
+			switch (preserve & TypePreserve.AccessibilityMask) {
+			case TypePreserve.AccessibilityVisible when !IsTypeVisible (type):
+				return;
+			case TypePreserve.AccessibilityVisibleOrInternal when !IsTypeVisibleOrInternal (type):
+				return;
+			}
+
+			Annotations.Mark (type, new DependencyInfo (DependencyKind.RootAssembly, type.Module.Assembly));
+			Annotations.SetPreserve (type, preserve);
```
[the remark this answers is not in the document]
marek-safar: It's not needed (see the test). It'd be if we mixed it with methods/fields which is not needed/supported scenario right now

src/linker/Linker.Steps/TrimUsingEntryPoint.cs
```diff
@@ -0,0 +1,64 @@
...
+				// Mark Step wil take care of marking whole assembly
+				return;
+			case AssemblyAction.Link:
+				break;
+			default:
+				Context.LogError ($"Trimming assembly '{assembly.Name}' cannot use action '{action}'", 1035);
+				return;
+			}
+
+			Annotations.Mark (ep.DeclaringType.Module, di);
```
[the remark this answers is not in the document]
marek-safar: I don't think so (it's one of the bugs this change fixes)

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -36,17 +36,18 @@ public class ILLink : ToolTask
...
-		///   illink with "-a", which roots the entry point for
-		///   executables, and everything for libraries. To control
-		///   the linker more explicitly, either pass descriptor
-		///   files, or pass extra arguments for illink.
+		///   The names of the assemblies to root using their entry point.
+		///   This value should be full file path.
+		///   Maps to '-a' command line option
 		/// </summary>
-		[Required]
-		public ITaskItem[] RootAssemblyNames { get; set; }
```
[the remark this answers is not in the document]
marek-safar: I went with a different approach in follow up PR

src/linker/Linker.Steps/TrimUsingEntryPoint.cs
```diff
@@ -0,0 +1,64 @@
...
+			Annotations.Mark (ep.DeclaringType, di);
+			Annotations.AddPreservedMethod (ep.DeclaringType, ep);
+		}
+
+		public static AssemblyDefinition LoadAssemblyFile (string fileName, LinkContext context)
+		{
+			AssemblyDefinition assembly = context.Resolver.GetAssembly (fileName, context.ReaderParameters);
```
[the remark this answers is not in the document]
marek-safar: Because we don't know what's the assembly name until we load the file

src/linker/Linker.Steps/TrimUsingVisibleMembers.cs
marek-safar: Didn't go with the cache after all

test/Mono.Linker.Tests.Cases/Libraries/DefaultLibraryLinkBehavior.cs
marek-safar: Made different change which makes this go away

test/Mono.Linker.Tests.Cases/Libraries/CopyAssemblyWithPublicRoots.cs
marek-safar: Fixed

test/Mono.Linker.Tests.Cases/Libraries/CopyAssemblyWithMainEntryRoot.cs
marek-safar: Fixed

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/InterfaceTypeInOtherUsedOnlyBySavedAssembly.cs
marek-safar: Yep, it's gone now

src/linker/Linker.Steps/TrimUsingEntryPoint.cs
marek-safar: Changed the approach in follow up PR

src/linker/Linker.Steps/TrimUsingVisibleMembers.cs
marek-safar: Changed the approach in follow up PR

docs/error-codes.md
marek-safar: Changed

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -36,17 +36,18 @@ public class ILLink : ToolTask
...
+		public ITaskItem[] RootAssemblyEntryPoint { get; set; }
```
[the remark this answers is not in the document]
marek-safar: Changed the approach in follow up PR

src/linker/Linker.Steps/TrimUsingEntryPoint.cs
```diff
@@ -0,0 +1,64 @@
...
+		{
+			this.fileName = fileName;
+		}
+
+		protected override void Process ()
+		{
+			AssemblyDefinition assembly = LoadAssemblyFile (fileName, Context);
+			var ep = assembly.MainModule.EntryPoint;
+			if (ep == null) {
+				Context.LogError ($"Trimming assembly '{assembly.Name}' does not have entry point", 1034);
```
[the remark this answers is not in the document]
marek-safar: Unless necessary I'd leave it for later. There is no good API to call it with LogError and don't want to do it as part of this PR

src/linker/Linker.Steps/TrimUsingVisibleMembers.cs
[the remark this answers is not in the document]
sbomer: If the extra bits are stored via SetPreserve, and nothing else calls SetPreserve, ChoosePreserveActionWhichPreservesTheMost won't ever run - then won't the extra bits be returned by TryGetPreserve?

test/Mono.Linker.Tests/Tests/PreserveActionComparisonTests.cs
```diff
@@ -20,7 +20,11 @@ public class PreserveActionComparisonTests
 		[TestCase (TypePreserve.Nothing, TypePreserve.All, TypePreserve.All)]
 		[TestCase (TypePreserve.Nothing, TypePreserve.Methods, TypePreserve.Methods)]
 		[TestCase (TypePreserve.Nothing, TypePreserve.Fields, TypePreserve.Fields)]
-		[TestCase (TypePreserve.Nothing, TypePreserve.Nothing, TypePreserve.Nothing)]
+		[TestCase (TypePreserve.All, TypePreserve.All | TypePreserve.AccessibilityVisible, TypePreserve.All)]
+		[TestCase (TypePreserve.AccessibilityVisibleOrInternal, TypePreserve.All, TypePreserve.All)]
+		[TestCase (TypePreserve.All | TypePreserve.AccessibilityVisible, TypePreserve.All | TypePreserve.AccessibilityVisible, TypePreserve.All | TypePreserve.AccessibilityVisible)]
+		[TestCase (TypePreserve.Nothing, TypePreserve.All | TypePreserve.AccessibilityVisible, TypePreserve.All | TypePreserve.AccessibilityVisible)]
+		[TestCase (TypePreserve.All | TypePreserve.AccessibilityVisibleOrInternal, TypePreserve.Nothing, TypePreserve.All | TypePreserve.AccessibilityVisibleOrInternal)]
 		public void VerifyBehaviorOfChoosePreserveActionWhichPreservesTheMost (TypePreserve left, TypePreserve right, TypePreserve expected)
```
sbomer, line 28: Does this need to handle a case like (TypePreserve.All | TypePreserve.AccessibilityVisible, TypePreserve.Fields)?

src/linker/Linker.Steps/TrimUsingEntryPoint.cs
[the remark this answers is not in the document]
sbomer: The Initialize logic in MarkStep should take care of it I think.

src/linker/Linker.Steps/TrimUsingEntryPoint.cs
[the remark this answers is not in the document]
sbomer: Could we just assume the filename without extension is the assembly name? We already do so when resolving an assemblyname ref to a file.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -42,8 +42,7 @@ public class ILLink : ToolTask
 		///   is subject to change. Currently these get passed to
 		///   illink with "-a", which roots the entry point for
 		///   executables, and everything for libraries. To control
-		///   the linker more explicitly, either pass descriptor
-		///   files, or pass extra arguments for illink.
+		///   the behaviour explicitly, set RootsMode metadata.
```
sbomer, line 45: May I suggest the name "RootMode", singular? Also the description doesn't match the GetMetadata call.

test/ILLink.Tasks.Tests/ILLink.Tasks.Tests.cs
```diff
@@ -106,20 +113,21 @@ public void TestReferenceAssemblyPaths (string[] referenceAssemblyPaths)
...
+					AssemblyAction actualAction = driver.Context.CalculateAssemblyAction (ad);
 					Assert.Equal (AssemblyAction.Skip, actualAction);
 				}
 			}
 		}
 
 		[Theory]
-		[InlineData (new object[] { new string[] { "AssemblyName" } })]
-		public void TestRootAssemblyNames (string[] rootAssemblyNames)
+		[InlineData (new object[] { new string[] { "illink.dll" } })]
```
sbomer, line 124: Could you keep the test without extension? That's really the expected use for RootAssemblyNames.

docs/illink-options.md
```diff
@@ -6,14 +6,29 @@ The `illink` is IL linker version shipping with .NET Core or .NET 5 platforms. I
...
+
+- `all` - Keep all members in root assembly
+- `default` - Use entry point for applications and all members for libraries
+- `entrypoint` - Use assembly entry point as only root in the assembly
+- `visible` - Keep all members and types visible outside of root assembly
+
+You can retain all public members of `Program.exe` application by calling
+linker like
+
+`illink -a Program.exe --root visible Program`
```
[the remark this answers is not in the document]
marek-safar: The reason is that I want to keep -a simple and --root is advanced rarely used option. -a, -p and --root have different intend.

> "--root all Program" is giving us another way to say "-p copy Program"

Yes, but why is this problematic?

> I think "--root" should be combined with either "-a" or "-p".

What would be the benefit of pushing rarely used option to common options?

test/Mono.Linker.Tests/Tests/PreserveActionComparisonTests.cs
marek-safar: No, this is not needed/possible scenario

test/ILLink.Tasks.Tests/ILLink.Tasks.Tests.cs
marek-safar: It's not how it's used today. This test change actually reflects how it's used today.

src/linker/Linker.Steps/TrimUsingEntryPoint.cs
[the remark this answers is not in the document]
marek-safar: I don't think we should and it's not needed here.

src/ILLink.Tasks/LinkTask.cs
marek-safar: Yep, happy to take it (went back and forth on this)

src/linker/Linker.Steps/TrimUsingVisibleMembers.cs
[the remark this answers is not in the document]
marek-safar: Changed the API to make it more obvious.

src/linker/Linker.Steps/TrimUsingEntryPoint.cs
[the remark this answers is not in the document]
marek-safar: Yep, it works now.

vitek-karas (approved):
Other than the comments it looks good.

If I understand this correctly - this will likely break runtime once brought there, right? (the special library mode it uses will now look different I think).

test/ILLink.Tasks.Tests/ILLink.Tasks.Tests.cs
sbomer: The SDK uses this for [TrimmerRootAssembly](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options#root-assemblies) which expects no extension.

src/linker/Linker/Annotations.cs
```diff
@@ -293,7 +293,22 @@ public TypePreserve GetPreserve (TypeDefinition type)
```
sbomer, line 293: `GetPreserve` can still return the accessibility bits - should it call `TryGetPreserve`? Same with `SetPreserve`, otherwise it could hit a case where the root assembly sets `TypePreserve.All | TypePreserve.AccessibilityVisible`, and an XML sets `TypePreserve.Methods` for example, and it would end up with just `TypePreserve.Methods`.

src/linker/Linker/Driver.cs
```diff
@@ -414,6 +417,25 @@ protected int SetupContext (ILogger customLogger = null)
 
 						continue;
 
+					case "--roots": {
```
sbomer, line 420: This should match the docs which say "--root" (I prefer singular FWIW).

docs/illink-options.md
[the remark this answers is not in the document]
sbomer: Keeping "-a" simple makes sense, I just don't like the redundancy if you are using "--root". Maybe we could have "--root" imply "-a"? Or have syntax like `-a <assembly> [optional root mode]`. I personally would vote for `--root <assembly> [optional root mode]` to replace "-a".

> why is this problematic?

Just because it's another concept to keep track of. The assembly actions are already a source of confusion, and it would be best not to further muddy the concepts.

In my mind, 'copy' used to mean something different - it used to affect the output but not the rooting behavior - it would keep the whole assembly possibly with dangling references, because it wasn't necessarily marked. But now that 'copy' has changed to mean "root everything" I think it makes sense to get rid of '-p copy' and use '--root all' instead.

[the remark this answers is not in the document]
marek-safar: Originally I didn't want to add it to `-a` because that will make it harder to remove `-a` as it's a redundant option (you want `illink myfile.dll` just work) but thinking about it again it should be possible to make it work with that for most cases.

test/ILLink.Tasks.Tests/ILLink.Tasks.Tests.cs
marek-safar: Oh boy, made this quirk to work as well

docs/illink-options.md
```diff
@@ -6,14 +6,29 @@ The `illink` is IL linker version shipping with .NET Core or .NET 5 platforms. I
...
 
-will use the assembly `Program.exe` as a source. That means that the linker will
-start with the main entry point method of `Program.exe` (typically the `Main` method) and process all its dependencies to determine what is necessary for this assembly to run.
+will use the assembly `Program.exe` as root linker input. That means that the linker will
+start with the main entry point method of `Program.exe` (typically the `Main` method) and
+process all its dependencies to determine what is necessary for this assembly to run.
+
+It's possible to use multiple input files and linker will treat them as multiple sources of
+input. When a library is used instead of executable linker will mark all members instead. This
+behaviour can be further customized using `--root` option which allows the following options.
```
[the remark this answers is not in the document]
marek-safar: Changed the wording completely.

src/linker/Linker.Steps/RootAssemblyInputStep.cs
```diff
@@ -0,0 +1,122 @@
...
+		{
+			AssemblyDefinition assembly = Context.Resolver.GetAssembly (fileName, Context.ReaderParameters);
+			AssemblyDefinition loaded = Context.GetLoadedAssembly (assembly.Name.Name);
+
+			// The same assembly could be already loaded if there are multiple inputs pointing to same file
+			if (loaded != null)
+				return loaded;
+
+			Context.Resolver.CacheAssembly (assembly);
+			Context.RegisterAssembly (assembly);
```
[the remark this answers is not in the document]
marek-safar: To me, this looks better than having in Context. The is ever called/used only once, if we move it to context anyone can call root assembly API and I don't think we can handle it anywhere except at the beginning of the pipeline.

src/linker/Linker/Driver.cs
```diff
@@ -182,8 +183,10 @@ protected int SetupContext (ILogger customLogger = null)
 			bool removeCAS = true;
 			bool new_mvid_used = false;
 			bool deterministic_used = false;
+			AssemblyAction? action = null;
```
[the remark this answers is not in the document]
marek-safar: changed

src/linker/Linker/Annotations.cs
marek-safar: I didn't notice as the method is never used. Remove it intead.

src/linker/Linker/Driver.cs
marek-safar: Not relevant anymore

src/linker/Linker.Steps/RootAssemblyInputStep.cs
```diff
@@ -0,0 +1,122 @@
...
+				Annotations.Mark (ep.DeclaringType, di);
+				Annotations.AddPreservedMethod (ep.DeclaringType, ep);
+				break;
+			case AssemblyRootsMode.VisibleMembers:
+				TypePreserve preserve = TypePreserve.All |
+					(HasInternalsVisibleTo (assembly) ? TypePreserve.AccessibilityVisibleOrInternal : TypePreserve.AccessibilityVisible);
+
+				foreach (var type in assembly.MainModule.Types)
+					MarkAndPreserveVisible (type, preserve);
+				break;
```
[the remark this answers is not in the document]
marek-safar: Added extra test for that. They are kept if visible mode is used

vitek-karas (approved):

test/ILLink.Tasks.Tests/ILLink.Tasks.Tests.cs
rolfbjarne: It's not a quirk, it's documented behavior: "Additional assemblies may be "rooted" by name (without the .dll extension):"

Ref: https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options#root-assemblies

PR #1728, opened 2021-01-07
Run TypeMap logic on demand

sbomer:
Separating this change out from https://github.com/mono/linker/pull/1666. This implements the behavior described in the first point under https://github.com/mono/linker/blob/feature/lazyload/docs/design/assembly-loading.md#additional-processing-done-lazily.

> The overrides tracked for a method may be incomplete, and will be updated as new assemblies are loaded. This is compatible with the current design of MarkStep which re-processes virtual methods, but this could be optimized in the future.

### DIFF
--- a/src/linker/Linker/TypeMapInfo.cs
+++ b/src/linker/Linker/TypeMapInfo.cs
@@ -1,5 +1,5 @@
 ﻿//
-// TypeMapStep.cs
+// TypeMapInfo.cs
 //
 // Author:
 //   Jb Evain (jbevain@novell.com)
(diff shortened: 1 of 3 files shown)


src/linker/Linker/TypeMapInfo.cs
```diff
@@ -29,18 +29,79 @@
...
+
+		public List<MethodDefinition> GetBaseMethods (MethodDefinition method)
+		{
+			EnsureProcessed (method.Module.Assembly);
+			return GetBaseMethodsHelper (method);
+		}
+
+		List<MethodDefinition> GetBaseMethodsHelper (MethodDefinition method)
+		{
+			if (base_methods.TryGetValue (method, out List<MethodDefinition> bases))
```
marek-safar, line 72: not: the if is redundat (you could also remove the one line method)

src/linker/Linker/TypeMapInfo.cs
```diff
@@ -29,18 +29,79 @@
...
 
-namespace Mono.Linker.Steps
+namespace Mono.Linker
 {
 
-	public class TypeMapStep : BaseStep
+	public class TypeMapInfo
 	{
+		readonly LinkContext _context;
+		AnnotationStore Annotations => _context.Annotations;
```
marek-safar, line 38: Not a fan of this circular dependency. Could you move default_interface_implementations cache to this type to avoid that?

src/linker/Linker/TypeMapInfo.cs
```diff
@@ -29,18 +29,79 @@
...
+				base_methods[method] = methods;
+			}
+
+			methods.Add (@base);
+		}
+
+		public void AddOverride (MethodDefinition @base, MethodDefinition @override, InterfaceImplementation matchingInterfaceImplementation = null)
+		{
+			if (!override_methods.TryGetValue (@base, out List<OverrideInformation> methods)) {
+				methods = new List<OverrideInformation> ();
```
vitek-karas, line 98: Unlike the AddBaseMethod you don't use the GetOverridesHelper here...
I would probably remove the helpers fully - they're not very helpful - the amount of code duplication is small.

vitek-karas (approved):

marek-safar (approved):

PR #1732, opened 2021-01-08
Clean up security information sweeping

sbomer:
- Unify the logic from RemoveSecurityStep and SweepStep
- Make the handling of security attributes more consistent

This way RemoveSecurityStep won't require a special case in the lazy per-assembly logic.


src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -357,30 +360,60 @@ protected void SweepGenericParameters (Collection<GenericParameter> genericParam
...
 			}
+		}
+
+		bool ShouldRemoveCustomAttribute (CustomAttribute attribute)
+		{
+			if (!Annotations.IsMarked (attribute))
+				return true;
+
+			// StripSecurity removes security custom attributes, even if marked.
+			if (Context.StripSecurity && IsSecurityAttributeType (attribute.AttributeType.Resolve ()))
```
marek-safar, line 415: nit: replace with `return Context.StripSecurity && IsSecurityAttributeType (attribute.AttributeType.Resolve ());`

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -357,30 +360,60 @@ protected void SweepGenericParameters (Collection<GenericParameter> genericParam
...
+			if (provider.HasSecurityDeclarations)
+				return true;
 
-			if (removed && method.HasSecurity && ShouldSetHasSecurityToFalse (method, method))
-				method.HasSecurity = false;
+			foreach (var attr in (provider as ICustomAttributeProvider).CustomAttributes) {
+				var attributeType = attr.AttributeType.Resolve ();
+				if (attributeType == null)
+					continue;
+				if (attributeType.Namespace == "System.Security" && attributeType.FullName == "System.Security.SuppressUnmanagedCodeSecurityAttribute")
```
marek-safar, line 392: Don't use `FullName`

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -357,30 +360,60 @@ protected void SweepGenericParameters (Collection<GenericParameter> genericParam
...
+			case MethodDefinition method:
+				if (method.HasSecurity)
+					method.HasSecurity = RequiresHasSecurity (provider);
+				break;
+			}
 		}
 
-		protected void SweepCustomAttributes (MethodDefinition method)
+		static bool RequiresHasSecurity (ISecurityDeclarationProvider provider)
 		{
```
marek-safar, line 381: Isn't this missing

```
if (Context.StripSecurity)
	return false;
```

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -390,37 +423,16 @@ static bool IsSecurityAttributeType (TypeDefinition definition)
 			if (definition == null)
 				return false;
 
-			if (definition.Namespace == "System.Security") {
-				return definition.FullName switch
-				{
-					// This seems to be one attribute in the System.Security namespace that doesn't count
-					// as an attribute that requires HasSecurity to be true
-					"System.Security.SecurityCriticalAttribute" => false,
```
vitek-karas, line 398: What this part removed intentionally?

sbomer: Yeah - this logic was requiring HasSecurity for all System.Security attributes except SecurityCriticalAttribute. Really, it is only required for SuppressUnmanagedCodeSecurityAttribute, or ones derived from System.SecurityAttribute - and the latter are represented as SecurityAttributes instead of CustomAttributes.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -377,19 +377,22 @@ protected void SweepSecurityInfo (ISecurityDeclarationProvider provider)
...
+
 			if (provider.HasSecurityDeclarations)
 				return true;
 
 			foreach (var attr in (provider as ICustomAttributeProvider).CustomAttributes) {
 				var attributeType = attr.AttributeType.Resolve ();
 				if (attributeType == null)
 					continue;
-				if (attributeType.Namespace == "System.Security" && attributeType.FullName == "System.Security.SuppressUnmanagedCodeSecurityAttribute")
+				if (attributeType.FullName == "System.Security.SuppressUnmanagedCodeSecurityAttribute")
```
vitek-karas, line 395: Why not use `IsOfType` ?

sbomer:
This needs more work in MarkStep to ensure that we don't keep the security attribute types if they're only referenced from link assemblies with '--strip-security' - and that is looking like it will take some cleanup of attribute marking in general.

Issue #1733, opened 2021-01-08
Make constant propagation and branch removal per-method, on-demand and called from MarkStep


vitek-karas:
The changes made to solve https://github.com/mono/linker/issues/1164 will mean that we can't run the constant propagation and branch removal before MarkStep as not all assemblies will be loaded at that time. It also means the algorithm can't rely on having visibility into all assemblies at the same time.

So we need to change it to be:
* per-method - it could be per-assembly, but it doesn't help much (not a big difference between per-method and per-assembly) and would be wasteful
* on-demand - only compute the constant propagation and branch removal when needed - this is basically a fallback of making assembly loading on-demand
* called from MarkStep right before ProcessMethod - this is basically an optimization to only perform the computations for methods which are actually going to be marked (no need to compute this for methods which will be trimmed from the app)

To do this the plan is to:
* [x] - refactor the `RemoveUnreachableBlocksStep` so that it doesn't precompute constant returns for all methods - do this on-demand from `TryInlineBodyDependencies`. But keep the iterative algorithm - meaning we will go over all methods multiple times to perform the recursive propagation. https://github.com/mono/linker/pull/1734
* [x] - refactor the `RemoveUnreachableBlocksStep` to perform constant propagation and branch removal truly on-demand - so basically introduce a method which will do it for a single method - remove the iterative part of the algorithm. #1756
  This will probably mean to:
  * Introduce a queue of methods which must be processed
  * When reaching a method call to a method which hasn't been processed, add it to the queue and re-try the current method later on (this is basically avoiding recursion - and replacing it with re-try and a queue)
  * Ideally reduce the number of methods this needs to happen on even more
* [x] - remove the `RemoveUnreachableBlocksStep` completely and instead call the functionality from `MarkStep` right before `ProcessMethod`. So that it only happens on marked methods (but before we actually go and mark the method's body). #1771 

/cc @sbomer @marek-safar

vitek-karas:
With the merging of #1771 this should be done - closing the issue.

PR #1734, opened 2021-01-08
Only compute if method returns a constant when needed

vitek-karas:
This changes how the `RemoveUnreachableBlocksStep` computes if a method returns a constant or not. Previously we would go over all methods and detect those which return constants (and store only those in a dictionary). Then we would proceed with branch removal relying on this dictionary.

For a hello world console app this means we ran the detection of constants on almost 50K methods. On the other hand the dictionary stored only about 1.5K records (actually constant methods).

With this change the constant detection only runs on methods we need the result for later on. This means the dictionary stores records for all methods asked about (`null` value represents "We checked, and it's not constant").

For the same hello world console app we now run the detection only on ~5K methods (10x less). On the other hand we have records for all (5K -> 3x increase). So this trades CPU for memory.

The purpose of this change is not the CPU/memory tradeoff, it's to prepare the constant propagation to make it possible to call it on-demand per-method from `MarkStep` which will be needed once we switch over to the current lazy-loading of assemblies.

This is a just a first step in this process described by https://github.com/mono/linker/issues/1733.


src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -17,81 +17,68 @@ namespace Mono.Linker.Steps
 	public class RemoveUnreachableBlocksStep : BaseStep
 	{
 		Dictionary<MethodDefinition, Instruction> constExprMethods;
+		int constExprMethodsAdded;
```
marek-safar, line 20: Cannot this be `bool` value?

marek-safar (approved):

sbomer (approved):
LGTM modulo the case we discussed where a constant-returning method has conflicting substitutions

vitek-karas:
This change also "fixes" a bug - if the IPConstantPropagation feature was disabled, substitutions would still work for the first level of propagation. After this change, even the first level propagation requires the IPContantPropagation to be turned on.

Note that this has nothing to do with the replacement of the method bodies in substitutions, those work regardless of the options.

PR #1736, opened 2021-01-10
Add help link for IL3000 and IL3001

Youssef1313:
These rules had a help link when they were in roslyn-analyzers repository (see https://github.com/dotnet/roslyn-analyzers/pull/4680/), but the help link was lost after porting it here.

This PR adds the help link back.

Tagging @mavasani in case he wants the documentation to be ported from dotnet/docs to somewhere else.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/SingleFileAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/SingleFileAnalyzer.cs
@@ -28,7 +28,8 @@ namespace ILLink.RoslynAnalyzer
 				Resources.ResourceManager, typeof (Resources)),
 			DiagnosticCategory.SingleFile,
 			DiagnosticSeverity.Warning,
-			isEnabledByDefault: true);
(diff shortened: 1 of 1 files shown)


mateoatr (approved):
LGTM - Thanks! I'd wait for @agocke / @mavasani on where should the docs be.

mavasani:
Tagging @gewarren who is the doc owner for analyzer rules shipping in the .NET SDK

gewarren:
So even though the analyzers are arriving via the linker repo, they still ship with the SDK? If so, I think we should just leave the docs as is?

"Moving forward, the analyzer will be packed and flown to the SDK via the linker repo (see dotnet/sdk#15239)."

mateoatr:
> So even though the analyzers are arriving via the linker repo, they still ship with the SDK? If so, I think we should just leave the docs as is?

This is correct.

agocke:
Yup, I'd leave the docs where they are. Change LGTM.

PR #1737, opened 2021-01-11
Fix the build after removal of dotnet.myget.org

vitek-karas:
That site is no longer available (returns 404) - and we don't need it to build anymore.


vitek-karas:
Merging this to unblock CI.

akoeplinger:
@vitek-karas would you mind backporting this to release/5.0? the build there fails for the same reason.

PR #1746, opened 2021-01-12
Updating the Linker Analyzer readme to use _TrimmerDumpDependencies

tannergooding:
As per the actual property checked in https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L122

### DIFF
--- a/src/analyzer/README.md
+++ b/src/analyzer/README.md
@@ -16,12 +16,12 @@ can be retrieved by enabling dependencies dumping during linking of a
 Xamarin.Android or a Xamarin.iOS project.
 
 That can be done on the command line by setting
-`LinkerDumpDependencies` property to `true` and building the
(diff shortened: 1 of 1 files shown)


marek-safar:
The analyzes can be actually used in Xamarin context where this name is correct. It'd be better to split the documentation instead.

radekdoulik (commented):
Indeed, we still support `$(LinkerDumpDependencies)` property in Xamarin.Android. So we might want to mention both or split documentation for legacy and net6.

tannergooding:
Updated to maintain the existing text for `Xamarin.Android` and `Xamarin.iOS`. There is a new section just below calling out the different variable to use for .NET SDK style projects.

PR #1752, opened 2021-01-13
Fix nullref caused by unknown value to the ignoreCase parameter of Type.GetType

### LINKED ISSUE #1741: Null-reference exception in Mono.Linker.Dataflow.ReflectionMethodBodyScanner.HandleCall
I use the [Z.EntityFramework.Plus.EFCore](https://www.nuget.org/packages/Z.EntityFramework.Plus.EFCore/5.1.11) library and it works great with .NET Core 5, but when I use it in my ready-to-run ARM executable, then the following error is shown:
```
ILLink : error IL1005: Z.Expressions.CodeCompiler.CSharp.ExpressionScope.(String,Type&): Error processing method 'Z.Expressions.CodeCompiler.CSharp.ExpressionScope.(String,Type&)' in assembly 'Z.Expressions.Eval.dll' [/builds/xxx.csproj]
  Mono.Linker.LinkerFatalErrorException: ILLink: error IL1005: Z.Expressions.CodeCompiler.CSharp.ExpressionScope.(String,Type&): Error processing method 'Z.Expressions.CodeCompiler.CSharp.ExpressionScope.(String,Type&)' in assembly 'Z.Expressions.Eval.dll'
   ---> System.NullReferenceException: Object reference not set to an instance of an object.
     at Mono.Linker.Dataflow.ReflectionMethodBodyScanner.HandleCall(MethodBody callingMethodBody, MethodReference calledMethod, Instruction operation, ValueNodeList methodParams, ValueNode& methodReturnValue)
     at Mono.Linker.Dataflow.MethodBodyScanner.HandleCall(MethodBody callingMethodBody, Instruction operation, Stack`1 currentStack)
     at Mono.Linker.Dataflow.MethodBodyScanner.Scan(MethodBody methodBody)
     at Mono.Linker.Dataflow.ReflectionMethodBodyScanner.ScanAndProcessReturnValue(MethodBody methodBody)
     at Mono.Linker.Steps.MarkStep.MarkReflectionLikeDependencies(MethodBody body, Boolean requiresReflectionMethodBodyScanner)
     at Mono.Linker.Steps.MarkStep.MarkMethodBody(MethodBody body)
     at Mono.Linker.Steps.MarkStep.ProcessMethod(MethodDefinition method, DependencyInfo& reason)
     at Mono.Linker.Steps.MarkStep.ProcessQueue()
     --- End of inner exception stack trace ---
     at Mono.Linker.Steps.MarkStep.ProcessQueue()
     at Mono.Linker.Steps.MarkStep.ProcessPrimaryQueue()
     at Mono.Linker.Steps.MarkStep.Process()
     at Mono.Linker.Steps.MarkStep.Process(LinkContext context)
```

vitek-karas:
We have special handling for cases where ignoreCase is not constant false (since linker won't be able to correctly mark the required types). This handling was written such that it assumed the value for the parameter was "known" that is we know at least something about it. If it's completely unknown, that's represented as null and the code crashes on it.

Added a test case for this and made the fix.

Fixes #1741

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -820,8 +820,7 @@ namespace Mono.Linker.Dataflow
 						reflectionContext.AnalyzingPattern ();
 
 						var parameters = calledMethod.Parameters;
-						if ((parameters.Count == 3 && (methodParams[2].Kind == ValueNodeKind.MethodReturn || methodParams[2].Kind == ValueNodeKind.ConstInt || methodParams[2].Kind == ValueNodeKind.LoadField)
-							&& (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
+						if ((parameters.Count == 3 && parameters[2].ParameterType.IsTypeOf ("System", "Boolean") && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
 							(parameters.Count == 5 && (methodParams[4].AsConstInt () == null || methodParams[4].AsConstInt () != 0))) {
 							reflectionContext.RecordUnrecognizedPattern (2096, $"Call to '{calledMethod.GetDisplayName ()}' can perform case insensitive lookup of the type, currently ILLink can not guarantee presence of all the matching types");
 							break;
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -820,7 +820,7 @@ namespace Mono.Linker.Dataflow
 						reflectionContext.AnalyzingPattern ();
 
 						var parameters = calledMethod.Parameters;
-						if ((parameters.Count == 3 && parameters[2].ParameterType.IsTypeOf ("System", "Boolean") && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
+						if ((parameters.Count == 3 && parameters[2].ParameterType.MetadataType == MetadataType.Boolean && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
 							(parameters.Count == 5 && (methodParams[4].AsConstInt () == null || methodParams[4].AsConstInt () != 0))) {
 							reflectionContext.RecordUnrecognizedPattern (2096, $"Call to '{calledMethod.GetDisplayName ()}' can perform case insensitive lookup of the type, currently ILLink can not guarantee presence of all the matching types");
 							break;
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -820,8 +820,8 @@ namespace Mono.Linker.Dataflow
 						reflectionContext.AnalyzingPattern ();
 
 						var parameters = calledMethod.Parameters;
-						if ((parameters.Count == 3 && parameters[2].ParameterType.MetadataType == MetadataType.Boolean && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
-							(parameters.Count == 5 && (methodParams[4].AsConstInt () == null || methodParams[4].AsConstInt () != 0))) {
+						if ((parameters.Count == 3 && parameters[2].ParameterType.MetadataType == MetadataType.Boolean && methodParams[2].AsConstInt () != 0) ||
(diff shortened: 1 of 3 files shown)


tlakollo (approved):
I got stuck thinking how to use the Kind to solve the boolean, looks like checking for parameter boolean works better and is more efficient

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -820,8 +820,7 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 						reflectionContext.AnalyzingPattern ();
 
 						var parameters = calledMethod.Parameters;
-						if ((parameters.Count == 3 && (methodParams[2].Kind == ValueNodeKind.MethodReturn || methodParams[2].Kind == ValueNodeKind.ConstInt || methodParams[2].Kind == ValueNodeKind.LoadField)
-							&& (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
+						if ((parameters.Count == 3 && parameters[2].ParameterType.IsTypeOf ("System", "Boolean") && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
```
MichalStrehovsky, line 823: Nit: check for primitive types can be done more efficiently

```suggestion
						if ((parameters.Count == 3 && parameters[2].ParameterType.MetadataType == MetadataType.Boolean && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
```

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -820,8 +820,7 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 						reflectionContext.AnalyzingPattern ();
 
 						var parameters = calledMethod.Parameters;
-						if ((parameters.Count == 3 && (methodParams[2].Kind == ValueNodeKind.MethodReturn || methodParams[2].Kind == ValueNodeKind.ConstInt || methodParams[2].Kind == ValueNodeKind.LoadField)
-							&& (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
+						if ((parameters.Count == 3 && parameters[2].ParameterType.MetadataType == MetadataType.Boolean && (methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0)) ||
```
marek-safar, line 823: nit: instead of `methodParams[2].AsConstInt () == null || methodParams[2].AsConstInt () != 0` you could add something like `methodParams[2].IsDefault ()`

test/Mono.Linker.Tests.Cases/Reflection/TypeUsedViaReflection.cs
```diff
@@ -359,5 +361,23 @@ static Type GetTypeFromAssembly (Assembly assembly, string name, bool caseSensit
 		{
 			return assembly == null ? Type.GetType (name, caseSensitive) : assembly.GetType (name, caseSensitive);
 		}
+
+		[Kept]
+		[ExpectedWarning ("IL2096", "'System.Type.GetType(String,Boolean,Boolean)'")]
+		static void TestUnkownIgnoreCase3Params (int num)
```
marek-safar, line 367: Do we have coverage for similar cases for other intrinsics?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: While technically true I'm not sure it's worth it. The case here is specific - the parameter is a boolean and we're looking for `true` or `unknown`. Neither is "default" semantically.

vitek-karas: Actually it can be simplified as simple `!=0` - or the boolean equivalent `!=false`.

test/Mono.Linker.Tests.Cases/Reflection/TypeUsedViaReflection.cs
vitek-karas: I checked all other cases where we consume the "method parameter" value in the intrinsic handling and none has this problem. We probably don't have coverage though. There are lot of places where we consume binding flags which could all be "null" - and the code has handling for it.

I'm adding a test for the null case for binding flags in one place - but all the other places use the same helper to handle this, so in terms of code coverage it's good. Writing one for each case would be lot of new test cases.

marek-safar (approved):

marek-safar:
We should consider backporting the fix to 5.0

vitek-karas:
I ported this to 5.0 as https://github.com/mono/linker/pull/1759
@marek-safar will you take it to triage?

marek-safar:
> will you take it to triage?

Yep

PR #1753, opened 2021-01-13
Keep types marked with DynamicInterfaceCastableImplementationAttribute when any interface they implement is kept

### LINKED ISSUE #1244: Add support for IDynamicInterfaceCastable
https://github.com/dotnet/runtime/issues/36654

On a high level:

* We need to consider `IDynamicInterfaceCastable.GetInterfaceImplementation` implicitly used (same as e.g. the `Object.Finalize` virtual method that is only called by the runtime)
* If an interface with `DynamicInterfaceCastableImplementationAttribute` is marked, do the same thing as if it was implemented by an instantiated type.

jkoritzinsky:
This PR adds support to the linker for IDynamicInterfaceCastable implementation types. These types might be referenced statically in an IDynamicInterfaceCastable implementation (custom targeted COM wrappers scenario), but they also may be dynamically discovered via reflection (C#/WinRT scenario).

This PR collects a list of all types with the `DynamicInterfaceCastableImplementationAttribute` applied that implement at least one interface and only marks them if any interface they implement is marked.

By collecting the list of types to mark in this way at initialization time, we handle both the dynamic and static cases at once.

Fixes #1244

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -51,6 +51,7 @@ namespace Mono.Linker.Steps
 		protected Queue<AttributeProviderPair> _assemblyLevelAttributes;
 		protected Queue<(AttributeProviderPair, DependencyInfo, IMemberDefinition)> _lateMarkedAttributes;
 		protected List<TypeDefinition> _typesWithInterfaces;
+		protected List<TypeDefinition> _dynamicInterfaceCastableImplementationTypes;
 		protected List<MethodBody> _unreachableBodies;
 
 		readonly List<(TypeDefinition Type, MethodBody Body, Instruction Instr)> _pending_isinst_instr;
@@ -176,6 +177,7 @@ namespace Mono.Linker.Steps
 			_assemblyLevelAttributes = new Queue<AttributeProviderPair> ();
 			_lateMarkedAttributes = new Queue<(AttributeProviderPair, DependencyInfo, IMemberDefinition)> ();
 			_typesWithInterfaces = new List<TypeDefinition> ();
+			_dynamicInterfaceCastableImplementationTypes = new List<TypeDefinition> ();
 			_unreachableBodies = new List<MethodBody> ();
 			_pending_isinst_instr = new List<(TypeDefinition, MethodBody, Instruction)> ();
 		}
@@ -226,6 +228,16 @@ namespace Mono.Linker.Steps
 			}
 		}
 
+		static bool TypeHasDynamicInterfaceCastableImplementationAttribute(TypeDefinition type)
+		{
+			foreach (var ca in type.CustomAttributes) {
+				TypeDefinition caType = ca.AttributeType.Resolve ();
+				if (caType.Name == "DynamicInterfaceCastableImplementationAttribute" && caType.Namespace == "System.Runtime.InteropServices")
+					return true;
+			}
+			return false;
+		}
+
 		void InitializeType (TypeDefinition type)
 		{
 			if (type.HasNestedTypes) {
@@ -233,10 +245,11 @@ namespace Mono.Linker.Steps
 					InitializeType (nested);
 			}
 
-			if (type.CustomAttributes.Any(ca => ca.AttributeType.Resolve().Name == "DynamicInterfaceCastableImplementationAttribute")) {
-				MarkType (type, DependencyInfo.DynamicInterfaceCastableImplementation, type);
+			if (TypeHasDynamicInterfaceCastableImplementationAttribute (type)) {
+				_dynamicInterfaceCastableImplementationTypes.Add (type);
 			}
-			else if (!Annotations.IsMarked (type))
+
+			if (!Annotations.IsMarked (type))
 				return;
 
 			// We may get here for a type marked by an earlier step, or by a type
@@ -386,6 +399,7 @@ namespace Mono.Linker.Steps
 				ProcessQueue ();
 				ProcessVirtualMethods ();
 				ProcessMarkedTypesWithInterfaces ();
+				ProcessDynamicCastableImplementationInterfaces ();
 				ProcessPendingBodies ();
 				DoAdditionalProcessing ();
 			}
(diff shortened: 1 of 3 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -456,6 +474,29 @@ void ProcessMarkedTypesWithInterfaces ()
...
+
+				Debug.Assert (TypeHasDynamicInterfaceCastableImplementationAttribute (type));
+
+				if (Annotations.IsMarked (type))
+					continue;
+
+				foreach (var iface in type.Interfaces) {
+					if (Annotations.IsMarked (iface.InterfaceType)) {
+						MarkType (type, new DependencyInfo (DependencyKind.DynamicInterfaceCastableImplementation, iface.InterfaceType), type);
+						_dynamicInterfaceCastableImplementationTypes.RemoveAt (i--);
```
AaronRobinsonMSFT, line 496: Do we really want to play the post-decrement side effect game? I don't think this costs much to pass in `i` and then `i--` in the next statement with a comment. Alternative reworking this logic to avoid manipulating a collection under inspection seems like the right thing to do _or_ even collecting the indices to remove and removing them post loop would be preferable in my opinion.

AaronRobinsonMSFT: Oh... This is a common pattern in here. Hmmm I guess never mind.

jkoritzinsky: Yeah it's not my favorite pattern, but it works and the rest of the code base uses it.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -225,13 +228,27 @@ void Complete ()
...
+
 		void InitializeType (TypeDefinition type)
 		{
 			if (type.HasNestedTypes) {
 				foreach (var nested in type.NestedTypes)
 					InitializeType (nested);
 			}
 
+			if (TypeHasDynamicInterfaceCastableImplementationAttribute (type) && type.Interfaces.Count > 0) {
+				_dynamicInterfaceCastableImplementationTypes.Add (type);
```
marek-safar, line 249: Can you build the queue inside ProcessDynamicCastableImplementationInterfaces on the first run instead? It could add fewer elements.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -225,13 +228,27 @@ void Complete ()
 			}
 		}
 
+		static bool TypeHasDynamicInterfaceCastableImplementationAttribute (TypeDefinition type)
+		{
+			foreach (var ca in type.CustomAttributes) {
```
marek-safar, line 233: This should add early exists for
- is type interface
- has any attribute

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -456,6 +474,32 @@ void ProcessMarkedTypesWithInterfaces ()
...
+
+				// If the type has already been marked, we can remove it from this list.
+				if (Annotations.IsMarked (type)) {
+					_dynamicInterfaceCastableImplementationTypes.RemoveAt (i--);
+					continue;
+				}
+
+				foreach (var iface in type.Interfaces) {
+					if (Annotations.IsMarked (iface.InterfaceType)) {
+						MarkType (type, new DependencyInfo (DependencyKind.DynamicInterfaceCastableImplementation, iface.InterfaceType), type);
```
marek-safar, line 495: What is the intention here, is marking type definition only enough? It would be nice to add a comment about the reason.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -225,13 +228,27 @@ void Complete ()
 			}
 		}
 
+		static bool TypeHasDynamicInterfaceCastableImplementationAttribute (TypeDefinition type)
+		{
+			foreach (var ca in type.CustomAttributes) {
+				TypeDefinition caType = ca.AttributeType.Resolve ();
+				if (caType.Name == "DynamicInterfaceCastableImplementationAttribute" && caType.Namespace == "System.Runtime.InteropServices")
```
marek-safar, line 235: I think this is a linker attribute only right? Should we remove it during trimming automatically?

src/linker/Linker.Steps/MarkStep.cs
jkoritzinsky: I don't know where I'd start building the queue from if I didn't do it here. We need to check all types that have the `[DynamicInterfaceCastableImplementation]` attribute and inherit at least one interface including ones that aren't marked anywhere else for the dynamic IDynamicInterfaceCastable scenario.

src/linker/Linker.Steps/MarkStep.cs
jkoritzinsky: This attribute actually isn't linker only. The runtime enforces that you use this attribute on the implementation to ensure that your implementation will be linker-friendly.

src/linker/Linker.Steps/MarkStep.cs
jkoritzinsky: Yes, marking the type definition only is enough. Because the linker will automatically mark interface implementations of marked interfaces on a marked type and mark implementations of methods from those interfaces, the linker already does the rest of the linking we need to do once we mark the type definition.

I'll add a comment.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -225,13 +228,30 @@ void Complete ()
 			}
 		}
 
+		static bool TypeIsDynamicInterfaceCastableImplementation (TypeDefinition type)
+		{
+			if (!type.IsInterface || type.Interfaces.Count == 0 || type.CustomAttributes.Count == 0)
```
marek-safar, line 233: Use `type.HasInterfaces` and `type.HasCustomAttributes` instead

src/linker/Linker.Steps/MarkStep.cs
marek-safar: That sounds odd and presumably, this is only supported by RyuJIT, right? Why is there no analyzer for this instead?

src/linker/Linker.Steps/MarkStep.cs
marek-safar: I think you need the cache only just before you enter `ProcessDynamicCastableImplementationInterfaces` and not anywhere else. You would iterate all assembly types which are not marked there instead eagerly here as the cached content here will be larger than after marking.

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: > That sounds odd and presumably, this is only supported by RyuJIT, right? Why is there no analyzer for this instead?

If the runtime didn't enforce this attribute, people wouldn't put the attribute here, illink would strip stuff, and people get weird failure modes at runtime with trimming enabled without a warning (unless we warn for all `IDynamicInterfaceCastable` uses and basically declare the whole feature not trimming safe).

Runtime enforces this attribute so that we give illink a chance to produce a consistent behavior (if the attribute is not there, illink is free to remove stuff since runtime is going to reject this either way).

src/linker/Linker.Steps/MarkStep.cs
jkoritzinsky: I've tried moving the cache creation. Let me know if you have any feedback on how I did it!

marek-safar (approved):

PR #1756, opened 2021-01-14
Constant propagation without iterations and stack based

vitek-karas:
Changes how the constant propagation works. Instead of iterating over all methods multiple times, use a stack to emulate recursive processing of methods as they are needed.

The doc in this change contains lot of details about the motivation and the algorithm used so please refer to that.

This essentially makes constant propagation and branch removal per-method and not global step. It's still left as a global step which walks all methods and processes them, but that part is now not required by the algorithm.

Interesting statistics (on a console hello world app):
* Before this change we scanned over all methods 3 times - in total more than 217K method body scans). After the change we only scan over a total of 73K method bodies (total number of methods is 72K, so we scan most methods only once)
* Since there are no loops in framework, the maximum number of times we scan any method is 2
* The maximum depth of the processing stack is 16 (but it's a simple app)

The great savings in CPU are accompanied with a larger memory usage. The step now stores a dictionary of all methods (and one reference as a value, the values are almost all non-unique so the storage necessary for values is tiny). This will be necessary anyway for once we move this into `MarkStep` as it has to keep a cache of already processed methods.

Part of #1733


vitek-karas:
Should be ready for final review - PTAL.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -102,53 +87,274 @@ void RewriteBodies (Collection<TypeDefinition> types)
...
+					// If any node was found which was not of current version (and moved to the top of the stack), move on to processing
+					// the stack - this will give a chance for these methods to be processed. It doesn't break the loop and we should come back here
+					// again due to the same loop as before, but now with more nodes processed (hopefully all of the dependencies of the nodes in the loop).
+					// In the worst case all of those nodes will become part of the loop - in which case we will move on to break the loop anyway.
+					if (foundNodesWithNonCurrentVersion) {
+						continue;
+					}
+
+					// No such node was found -> we only have nodes in the loop now, so we have to break the loop.
+					// We do this by processing it with special flag which will make it ignore any unprocessed dependencies
```
marek-safar, line 202: The comment suggests a better name for the field, which part of the name the bool value `treatUnprocessedAsNonConst` does change?

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -102,53 +87,274 @@ void RewriteBodies (Collection<TypeDefinition> types)
...
+			switch (processedState) {
+			case LinkedListNode<ProcessingNode> queueNode:
+				// Method is already in the stack - not yet processed
+				// Move it to the top of the stack
+				processingStack.Remove (queueNode);
+				processingStack.AddFirst (queueNode);
+
+				// Note that stack version is not changing - we're just postponing work, not resolving anything.
+
+				constantResultInstruction = null;
```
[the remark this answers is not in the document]
vitek-karas: I don't understand the question. Method can only have one final result. If it's still in the queue (as is this case) it doesn't have a result yet. That's why it returns `false` here - telling the caller that no result is available.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -16,72 +16,57 @@ namespace Mono.Linker.Steps
...
+		int processingStackVersion;
+
+		// Stores results of method processing. This state is kept forever to avoid reprocessing of methods.
+		// If method is not in the dictionary it has not yet been processed and it has not been added processing either.
+		// The value in this dictionary can be
+		//   - Reference to the linked list node in the processingStack if the method is on the stack and not yet processed.
+		//   - ProcessedUnchangedSentinel - which means the method has been fully processed and nothing was changed on it - its value is unknown
+		//   - NonConstSentinel - which means the method has been processed and the return value is not a const
+		//   - Instruction instance - method has been processed and it has a constant return value (the value of the instruction)
+		Dictionary<MethodDefinition, object> processedMethods;
```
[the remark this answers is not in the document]
vitek-karas: The only value which we could in theory store in a different place is the reference to the queue. All the others are valid "final" results. The instruction means it's a const method and has a value (just like before). The `NonConstSentinel` means it doesn't have a const value - this is new but necessary. Before this change we relied on multiple iterations to react to changes. The cache only stored const values - if a method was not in the cache it could mean either "nonconst" or "don't know/yet". In this change we need to do things per-method, and so we need to store the negative "nonconst" results in the cache as well.
The only optimization here is the `ProcessedUnchangedSentitel` which means "processed, but didn't compute result". This is pretty important optimization. In the hello world app, there's a total off 75K methods, 67K end up in this `ProcessedUnchanged` state - meaning we did process them (good), but nothing needed to their value. So we avoided running the const analyzer on ~80% of the methods. I could remove this, but it feels like a good optimization.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -102,53 +87,274 @@ void RewriteBodies (Collection<TypeDefinition> types)
...
+					// again due to the same loop as before, but now with more nodes processed (hopefully all of the dependencies of the nodes in the loop).
+					// In the worst case all of those nodes will become part of the loop - in which case we will move on to break the loop anyway.
+					if (foundNodesWithNonCurrentVersion) {
+						continue;
+					}
+
+					// No such node was found -> we only have nodes in the loop now, so we have to break the loop.
+					// We do this by processing it with special flag which will make it ignore any unprocessed dependencies
+					// treating them as non-const. These should only be nodes in the loop.
+					treatUnprocessedAsNonConst = true;
```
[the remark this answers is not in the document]
vitek-karas: I think only the inner most loop will be "caught" by the detection here - all nodes having the same version. This will "break" one of the nodes in the inner most loop. Continuing processing should then completely process the inner most loop, at which point the outer loop will be detected (since it will become the inner most at that point) and so on...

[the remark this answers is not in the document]
vitek-karas: I added a new tested with nested loops to validate that it doesn't break the algorithm.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -194,8 +400,22 @@ bool TryInlineBodyDependencies (ref BodyReducer reducer)
...
-					if (!TryGetConstantResultInstructionForMethod (md, out targetResult))
+					if (md == reducer.Body.Method) {
+						// Special case for direct recursion - simply assume non-const value
+						// since we can't tell.
+						break;
+					}
+
+					if (!TryGetConstantResultForMethod (md, out targetResult)) {
+						if (!treatUnprocessedAsNonConst)
+							hasUnprocessedDependencies = true;
```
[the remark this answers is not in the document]
vitek-karas: You mean that the `TryGetConstantResultForMethod` will return `false` for another reason than "unprocessed yet"? It won't - it returns `false` only if the method has not been in the queue at all (adds it) or if the method is already in the queue (but not yet processed).

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
vitek-karas: Sorry - I don't think I follow...
I guess the name could mention that it's for breaking loops - but it didn't felt right to me. We are using it to break loops, sure, but what it does is in its name - it will treat all unprocessed dependencies as non-const.

I'll change it to `treatUnprocessedDependenciesAsNonConst`.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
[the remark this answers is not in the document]
marek-safar: What I meant is that the code uses the cache for two different purpose

1. as the final method -> Instructions? cache
2. as temporary processing cache

and I'm just not sure it's worth dropping to object to facilitate that

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
marek-safar: Let me ask different, what does `treatUnprocessedDependenciesAsNonConst = false` do ?

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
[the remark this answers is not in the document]
marek-safar: true, why do you need the remove/addfirst then?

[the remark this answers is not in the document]
vitek-karas: To move it to the top of the stack. This is so that callees are processed before callers next time around. For example if both A->C,B and B->C... after processing A the stack would be B,C,A (first is top). After processing B without the shuffle it would remain B,C,A - and it would get stuck. So the remove/add moves C to the begining so that the stack is C,B,A - C is processed and then B can get processed and so on.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
vitek-karas: If there's at least one unprocessed dependency, the processing of the current method will be stopped and we go back to the stack - in the code/comments I call this "backing off". The method will be retried at some point.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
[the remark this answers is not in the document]
vitek-karas: Makes sense - split it into two dictionaries.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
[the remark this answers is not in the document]
marek-safar: Yeah, it looks like this is enough if we don't handle recursion. I couldn't come with a case where this would break ;-)

marek-safar (approved):

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -16,43 +16,268 @@ namespace Mono.Linker.Steps
...
+					//   meaning removing any of them should break the loop and allow to make progress.
+					//   There might be other methods in between these which don't have the current version but are dependencies of some of the method
+					//   in the loop.
+					//   If we don't process these, then we might miss constants and branches to remove. See the doc
+					//   `constant-propagation-and-branch-removal.md` in this repo for more details and a sample.
+
+					// To fix this go over the stack and find the "oldest" node with the current version - the "oldest" node which
+					// is part of the loop:
+					var lastNodeWithCurrentVersion = stackNode;
+					for (var currentNode = stackNode; currentNode != null; currentNode = currentNode.Next) {
```
sbomer, line 183: This will walk the whole stack . I'm not sure it matters since we expect loop cases are rare, but would lead to bad worst-case performance. A small optimization would be to walk the list backwards.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -16,43 +16,268 @@ namespace Mono.Linker.Steps
...
+							lastNodeWithCurrentVersion = currentNode;
+					}
+
+					// There should be at least 2 nodes with the latest version to form a loop
+					Debug.Assert (lastNodeWithCurrentVersion != stackNode);
+
+					// Now go back over all nodes from the "oldest" one back to the top and find any nodes which are not of current version.
+					// For all of them, move them to the top of the stack.
+					var candidateNodeToMoveToTop = lastNodeWithCurrentVersion;
+					bool foundNodesWithNonCurrentVersion = false;
```
sbomer, line 194: You could also compute foundNodesWithNonCurrentVersion in the loop above to avoid walking the loop nodes twice in case there are none.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -194,8 +406,22 @@ bool TryInlineBodyDependencies (ref BodyReducer reducer)
...
+						// since we can't tell.
 						break;
+					}
+
+					if (!TryGetConstantResultForMethod (md, out targetResult)) {
+						if (!treatUnprocessedDependenciesAsNonConst)
+							hasUnprocessedDependencies = true;
+						break;
+					} else if (targetResult == null || hasUnprocessedDependencies) {
+						// Even is const is detected, there's no point in rewriting anything
```
sbomer, line 420: I don't understand the comment - I thought that if `TryGetConstantResultForMethod` returns `true` the method is analyzed, so `targetResult == null` in this case means we've analyzed the method and know it is not constant. Why should we break in that case?

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
```diff
@@ -16,43 +16,268 @@ namespace Mono.Linker.Steps
...
+				//
+				if (!TryInlineBodyDependencies (ref reducer, treatUnprocessedDependenciesAsNonConst, out bool changed)) {
+					// Method has unprocessed dependencies - so back off and try again later
+					// Leave it in the stack on its current position (it should not be on the first position anymore)
+					Debug.Assert (processingStack.First != stackNode);
+					continue;
+				}
+
+				if (!changed) {
+					// All dependencies are processed and there where no const values found. There's nothing to optimize.
```
sbomer, line 242: nit
```suggestion
					// All dependencies are processed and there were no const values found. There's nothing to optimize.
```

sbomer (commented):
Just wanted to note - I think this has worst-case performance of O(n^2) in the number of loop methods since we can iterate over the loop methods again whenever we encounter a new call to something already on the stack. But it shouldn't matter as long as the loop cases are rare and small.

I left a couple other comments about optimizing the loop cases, but feel free to disregard them (I think readability matters more as long as we are already assuming loops are uncommon).

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
vitek-karas: That's a good idea - changed.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
vitek-karas: Actually it can all be done in one backward full walk of the stack - changed the implementation.

src/linker/Linker.Steps/RemoveUnreachableBlocksStep.cs
vitek-karas: It's non-const - so nothing to do - break in this case means continue with next instruction (break from the switch over instruction opcode).

sbomer (approved):

PR #1759, opened 2021-01-15
[release/5.0] Fix nullref caused by unknown value to the ignoreCase parameter of Type.GetType

vitek-karas:
5.0 port of #1752 

Fix for #1741


marek-safar:
Approved by tactics

Issue #1610, opened 2020-11-05
Better handling of a library trimming


marek-safar:
ILLinker does not have a good concept of a library trimming. The initial approach was made with `-r assembly` options but it's incomplete.

There is a conceptual problem because options `-a` and -`r` are not exclusive, one can use the same assembly or assemblies with both modes at the same time. We should probably break this expectation and allow only one mode at the same time. This would allow us to set different defaults in library mode which is used mostly for IL trimming only.

The list of settings which would be set differently in this mode (unless set manually)
- --strip-link-attributes false
- --strip-substitutions false
- --strip-descriptors false
- --disable-opt unusedinterfaces
- --disable-opt overrideremoval
- --disable-opt unusedtypechecks

@vitek-karas @sbomer

sbomer:
Probably also:
- --keep-dep-attributes true

Related: https://github.com/mono/linker/issues/1541.

PR #1763, opened 2021-01-18
Refactor type members preservation coming from assembly roots.

marek-safar:
- It'd be too messy to track inside single enum
- Delay resolving exported types to the point where all references are known

Contributes to #1610

### DIFF
--- a/src/linker/Linker.Steps/MarkExportedTypesTargetStep.cs
+++ b/src/linker/Linker.Steps/MarkExportedTypesTargetStep.cs
@@ -1,0 +1,40 @@
+// Licensed to the.NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class MarkExportedTypesTargetStep : BaseStep
+	{
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			if (!assembly.MainModule.HasExportedTypes)
+				return;
+
+			foreach (var type in assembly.MainModule.ExportedTypes)
+				InitializeExportedType (type);
+		}
+
+		void InitializeExportedType (ExportedType exportedType)
+		{
+			if (!Annotations.IsMarked (exportedType))
+				return;
+
+			if (!Annotations.TryGetPreservedMembers (exportedType, out TypePreserveMembers members))
+				return;
+
+			TypeDefinition type = exportedType.Resolve ();
+			if (type == null) {
+				if (!Context.IgnoreUnresolved)
+					Context.LogError ($"Exported type '{type.Name}' cannot be resolved", 1038);
+
+				return;
+			}
+
+			Context.Annotations.Mark (type, new DependencyInfo (DependencyKind.ExportedType, exportedType));
+			Annotations.SetMembersPreserve (type, members);
+		}
+	}
+}
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -210,7 +210,6 @@ namespace Mono.Linker.Steps
 
 				foreach (TypeDefinition type in assembly.MainModule.Types)
 					InitializeType (type);
-
 				break;
 			}
 		}
@@ -2287,7 +2286,7 @@ namespace Mono.Linker.Steps
 				case TypePreserve.All:
 					MarkFields (type, true, di);
 					MarkMethods (type, di, type);
-					break;
+					return;
 
 				case TypePreserve.Fields:
 					if (!MarkFields (type, true, di, true))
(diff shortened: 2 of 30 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2244,46 +2243,47 @@ void ApplyPreserveInfo (TypeDefinition type)
...
+				var di = new DependencyInfo (DependencyKind.TypePreserve, type);
 
-			switch (preserve) {
-			case TypePreserve.All:
-				Annotations.TryGetPreserveAccessibility (type, out preserve);
 				switch (preserve) {
-				case TypePreserve.AccessibilityVisible:
+				case TypePreserve.All:
+					MarkFields (type, true, di);
+					MarkMethods (type, di, type);
```
vitek-karas, line 2252: This is a change in behavior - before if it was TypePreserve.All and AllVisible we would only call `MarkMethodsIf (..., IsMethodVisible, ...)` and so on.
Now we call `MarkMethods` and then also `MarkMethodsIf`. 

Also the `MarkMethodsIf` later on will not have any effect really (since all methods are already marked).

marek-safar: > if it was TypePreserve.All and AllVisible we would only call MarkMethodsIf (..., IsMethodVisible, ...) and so on.

yeah, that's one of the bugs this is fixing

> Also the MarkMethodsIf later on will not have any effect really (

Fixed that

src/linker/Linker.Steps/MarkExportedTypesTargetStep.cs
```diff
@@ -0,0 +1,40 @@
...
+
+			foreach (var type in assembly.MainModule.ExportedTypes)
+				InitializeExportedType (type);
+		}
+
+		void InitializeExportedType (ExportedType exportedType)
+		{
+			if (!Annotations.IsMarked (exportedType))
+				return;
+
```
vitek-karas, line 24: Since this runs before `MarkStep` - what will happen if I specify descriptor like this?
```xml
<type fullname="MyExportedType" required="false" preserve="methods"/>
```

The exported type won't get marked by the descriptor (since `required=false`), and thus this step won't do anything. But when `MarkStep` actually marks it, it probably won't apply the `preserve=methods` to the target of the exported type, right? Should this be done in two parts - first handle the marked types (and mark the target) and then handle type preserve (and propagate it to the target regardless of marking)?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2281,46 +2279,47 @@ void ApplyPreserveInfo (TypeDefinition type)
...
+					break;
+				case TypePreserve.Methods:
+					if (!MarkMethods (type, di, type))
+						_context.LogWarning ($"Type {type.GetDisplayName ()} has no methods to preserve", 2002, type);
+					break;
+				}
+			}
+
+			if (Annotations.TryGetPreservedMembers (type, out TypePreserveMembers members)) {
+				var di = new DependencyInfo (DependencyKind.TypePreserve, type);
```
vitek-karas, line 2303: I guess I don't understand how the type preserve and type preserve members should interact. It seems that:
* `All` -> all regardless of members visibility (and regardless of `AllVisible` or `AllVisibleOrInternal`).
* `Methods` -> all methods regardless of visibility - but then we also apply `AllVisible` or `AllVisibleOrInternal` on top of that.
* `Fields` -> similar to `Methods` but with fields.
* `None` -> still apply `AllVisible` or `AllVisibleOrInternal`.

Is that the desired behavior?

vitek-karas: So to simplify the two settings are orthogonal - we apply the union of both. I think it would make sense to somehow change the naming - they seem related since the names are so similar. Or maybe it would make sense to unify them into one thing - something like:
* `AllMethods`, `VisibleMethods`, `VisibleOrInternalMethods`
* `AllFields`, `VisibleFields`, `VisibleOrInternalFields`

And then `preserve=all` would be set as `AllMethods | AllFields`, `preserve=methods` would mean `AllMethods` and marking an assembly as `visible` would add `VisibleMethods | VisibleFields` to all types in the assembly.

Then the implementation here would be straightforward...

vitek-karas: Or at least add a comment about the intended behavior somewhere :-)

marek-safar: > Then the implementation here would be straightforward

I tried to unify them because I thought it'd be straight forward but had to undo that because it's too complex to do the merge and not worth the code complexity. For example when end up preserving All Methods and Visible Fields. I'd have to change existing enum to have only flags that could break custom steps.

vitek-karas: Right - I didn't realize that the enum is publicly visible...
OK - let's keep it separate.

src/linker/Linker.Steps/MarkExportedTypesTargetStep.cs
vitek-karas: It's OK to load the target assembly in this step to add the preserve onto the target type - so I would think that should work just fine. But test would be nice :-)

marek-safar: Made extra tests to cover this and they pass

vitek-karas (approved):

Issue #1735, opened 2021-01-08
Rethink how SubStepsDispatcher is used to mark types and assemblies


marek-safar:
As part of #1164 work we need to rethink how we are going to allow custom steps to mark additional elements either conditionally or unconditionally. Currently, many custom steps are run before mark as BaseSubStep called via SubStepsDispatcher which means they assume that `context.GetAssemblies ()` returns full assemblies set. This needs to change if we want to really delay load assemblies.

The logic inside these custom sub-steps is usually something that marks type or type members when the type is found. Here are a few examples

- https://github.com/xamarin/xamarin-android/blob/master/src/Microsoft.Android.Sdk.ILLink/PreserveRegistrations.cs
- https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/PreserveApplications.cs
- https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/PreserveJavaTypeRegistrations.cs
- https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/PreserveJavaExceptions.cs

It seems like if we had a solution on how to run additional marking logic for types and members conditionally we could convert these SubSteps into something that does not need `GetAssemblies` call.

@vitek-karas @sbomer

vitek-karas:
@sbomer - you're the expert on delay loading (well, let me rephrase, you probably remember more than me :wink:). Is this still a problem? I though we redesigned the Xamarin custom steps to avoid the problem.

PR #1768, opened 2021-01-20
Track pending marked members

sbomer:
Separated out from https://github.com/mono/linker/pull/1666. 

This enables tracking of pending marked items which need to be fully marked by MarkStep. Contributes to https://github.com/mono/linker/issues/1735 - this change is in preparation for running custom steps during MarkStep, where they could change the Annotations state in ways that interact with other marking logic. The idea is that custom steps can call Annotations methods without triggering a lot of other processing, because the full logic is deferred. This way custom steps don't need to be re-entrant.

This change causes `Annotations.Mark` to place members into a set of "pending" items which will get fully marked ("processed") later. "pending" items are already considered marked.

`AddPreservedMethod` conceptually adds a conditional dependency from source -> destination - so if source gets marked, destination should get marked. This change will immediately call Annotations.Mark on the destination if the source is already marked, and otherwise track the condition to be applied if the source gets marked later.

`SetPreserve` can change the TypePreserve of a type after it has been marked. This change will track any changes to TypePreserve and apply them later even for types which have already been marked. This works a little differently from `AddPreservedMethod`, to avoid traversing type members in Annotations.

Note that once we allow custom steps to run during during MarkStep, `Process` will call `ProcessMarkedPending` - but this isn't required in this change. The intention here is just to add adding extra tracking (that will actually be used by https://github.com/mono/linker/pull/1666), while mostly preserving existing behavior.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2289,7 +2289,7 @@ namespace Mono.Linker.Steps
 		{
 			if (Annotations.TryGetPreserve (type, out TypePreserve preserve)) {
 				if (!Annotations.SetAppliedPreserve (type, preserve))
-					throw new InvalidOperationException ($"Type {type} already has applied {preserve}.");
+					throw new InternalErrorException ($"Type {type} already has applied {preserve}.");
 
 				var di = new DependencyInfo (DependencyKind.TypePreserve, type);
 
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -358,11 +358,6 @@ namespace Mono.Linker.Steps
 		{
 			bool marked = false;
 			foreach (var pending in Annotations.MarkedPending ()) {
-				var assembly = GetAssemblyFromMetadataTokenProvider (pending);
-				var assemblyAction = Annotations.GetAction (assembly);
-				if (assemblyAction == AssemblyAction.Skip)
-					continue;
-
 				marked = true;
 
 				// Some pending items might be processed by the time we get to them.
@@ -372,7 +367,6 @@ namespace Mono.Linker.Steps
 				switch (pending) {
 				case TypeDefinition type:
 					MarkType (type, DependencyInfo.AlreadyMarked, type);
-					Debug.Assert (!Annotations.IsMarkedPending (type));
 					break;
 				case MethodDefinition method:
 					MarkMethod (method, DependencyInfo.AlreadyMarked, null);
@@ -380,11 +374,9 @@ namespace Mono.Linker.Steps
 					break;
 				case FieldDefinition field:
 					MarkField (field, DependencyInfo.AlreadyMarked);
-					Debug.Assert (!Annotations.IsMarkedPending (field));
 					break;
 				case ModuleDefinition module:
 					MarkModule (module, DependencyInfo.AlreadyMarked);
-					Debug.Assert (!Annotations.IsMarkedPending (module));
 					break;
 				default:
 					throw new NotImplementedException (pending.GetType ().ToString ());
@@ -677,7 +669,7 @@ namespace Mono.Linker.Steps
 		void MarkCustomAttributes (ICustomAttributeProvider provider, in DependencyInfo reason, IMemberDefinition sourceLocationMember)
 		{
 			if (provider.HasCustomAttributes) {
-				bool providerInLinkedAssembly = Annotations.GetAction (GetAssemblyFromMetadataTokenProvider (provider)) == AssemblyAction.Link;
+				bool providerInLinkedAssembly = Annotations.GetAction (GetAssemblyFromCustomAttributeProvider (provider)) == AssemblyAction.Link;
 				bool markOnUse = _context.KeepUsedAttributeTypesOnly && providerInLinkedAssembly;
 
 				foreach (CustomAttribute ca in provider.CustomAttributes) {
@@ -819,7 +811,7 @@ namespace Mono.Linker.Steps
 			}
 		}
 
-		protected static AssemblyDefinition GetAssemblyFromMetadataTokenProvider (IMetadataTokenProvider provider)
+		protected static AssemblyDefinition GetAssemblyFromCustomAttributeProvider (ICustomAttributeProvider provider)
 		{
 			return provider switch
 			{
@@ -1301,7 +1293,7 @@ namespace Mono.Linker.Steps
 					continue;
 				}
 
-				if (_context.Annotations.HasLinkerAttribute<RemoveAttributeInstancesAttribute> (resolved.DeclaringType) && Annotations.GetAction (GetAssemblyFromMetadataTokenProvider (assemblyLevelAttribute.Provider)) == AssemblyAction.Link)
+				if (_context.Annotations.HasLinkerAttribute<RemoveAttributeInstancesAttribute> (resolved.DeclaringType) && Annotations.GetAction (GetAssemblyFromCustomAttributeProvider (assemblyLevelAttribute.Provider)) == AssemblyAction.Link)
 					continue;
 
 				if (!ShouldMarkTopLevelCustomAttribute (assemblyLevelAttribute, resolved)) {
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -352,7 +352,7 @@ namespace Mono.Linker.Steps
 
 		void Process ()
 		{
-			while (ProcessPrimaryQueue () || ProcessLazyAttributes () || ProcessLateMarkedAttributes ()) {
+			while (ProcessPrimaryQueue () || ProcessMarkedPending() || ProcessLazyAttributes () || ProcessLateMarkedAttributes ()) {
 
 				// deal with [TypeForwardedTo] pseudo-attributes
 				foreach (AssemblyDefinition assembly in _context.GetAssemblies ()) {
@@ -374,7 +374,8 @@ namespace Mono.Linker.Steps
 							continue;
 						if (!Annotations.IsMarked (type))
 							continue;
-						_context.MarkingHelpers.MarkExportedType (exported, assembly.MainModule, new DependencyInfo (DependencyKind.ExportedType, type));
+						var di = new DependencyInfo (DependencyKind.ExportedType, type);
+						_context.MarkingHelpers.MarkExportedType (exported, assembly.MainModule, di);
 					}
 				}
 			}
@@ -397,6 +398,47 @@ namespace Mono.Linker.Steps
 			}
 
 			return true;
+		}
+
+		bool ProcessMarkedPending ()
+		{
+			var allPending = Annotations.MarkedPending ();
+			bool marked = false;
+			foreach (var pending in Annotations.MarkedPending ()) {
+				marked = true;
+
+				// Some pending items might be processed by the time we get to them.
+				if (Annotations.IsProcessed (pending))
+					continue;
+
+				switch (pending) {
+				case TypeDefinition type:
+					MarkType (type, DependencyInfo.AlreadyMarked, type);
+					Debug.Assert (!Annotations.IsMarkedPending (type));
+					break;
+				case MethodDefinition method:
+					EnqueueMethod (method, DependencyInfo.AlreadyMarked);
+					break;
+				case FieldDefinition field:
+					MarkField (field, DependencyInfo.AlreadyMarked);
+					Debug.Assert (!Annotations.IsMarkedPending (field));
+					break;
+				case ModuleDefinition module:
+					MarkModule (module, DependencyInfo.AlreadyMarked);
+					Debug.Assert (!Annotations.IsMarkedPending (module));
+					break;
+				default:
+					throw new NotImplementedException (pending.GetType ().ToString ());
+				}
+			}
+
+			foreach (var type in Annotations.PendingPreserve ()) {
+				marked = true;
+				Debug.Assert (Annotations.IsProcessed (type));
+				ApplyPreserveInfo (type);
+			}
+
+			return marked;
 		}
 
 		void ProcessPendingTypeChecks ()
@@ -1231,6 +1273,7 @@ namespace Mono.Linker.Steps
 			return !Annotations.SetProcessed (provider);
 		}
 
+
 		protected void MarkAssembly (AssemblyDefinition assembly, DependencyInfo reason)
 		{
 			// This conceptually marks an assembly (marking the module type and)
@@ -1392,7 +1435,7 @@ namespace Mono.Linker.Steps
 #endif
 
 			if (reason.Kind == DependencyKind.AlreadyMarked) {
-				Debug.Assert (Annotations.IsMarkedPending (field));
+				Debug.Assert (Annotations.IsMarked (field));
 			} else {
 				Annotations.Mark (field, reason);
 			}
@@ -1447,12 +1490,16 @@ namespace Mono.Linker.Steps
 			return Annotations.GetAction (assembly) != AssemblyAction.Link;
 		}
 
-		void MarkScope (IMetadataScope scope, TypeDefinition type)
+		void MarkModule (ModuleDefinition module, DependencyInfo reason)
 		{
-			Annotations.Mark (scope, new DependencyInfo (DependencyKind.ScopeOfType, type));
-			var di = new DependencyInfo (DependencyKind.ScopeOfType, type);
-			Annotations.Mark (scope, di);
-			MarkAssembly (scope.Assembly, di);
+			if (reason.Kind == DependencyKind.AlreadyMarked) {
+				Debug.Assert (Annotations.IsMarked (module));
+			} else {
+				Annotations.Mark (module, reason);
+			}
+			if (CheckProcessed (module))
+				return;
+			MarkAssembly (module.Assembly, new DependencyInfo (DependencyKind.AssemblyOfModule, module));
 		}
 
 		protected virtual void MarkSerializable (TypeDefinition type)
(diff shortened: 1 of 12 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1392,6 +1398,12 @@ void MarkField (FieldDefinition field, in DependencyInfo reason)
 				throw new ArgumentOutOfRangeException ($"Internal error: unsupported field dependency {reason.Kind}");
 #endif
 
+			if (reason.Kind == DependencyKind.AlreadyMarked) {
+				Debug.Assert (Annotations.IsMarked (field));
+			} else {
+				Annotations.Mark (field, reason);
+			}
+
```
marek-safar, line 1406: Could this be moved after CheckProcessed?

src/linker/Linker/Annotations.cs
```diff
@@ -190,9 +194,20 @@ public void Mark (CustomAttribute attribute, in DependencyInfo reason)
 			Tracer.AddDirectDependency (attribute, reason, marked: true);
 		}
 
+		public IEnumerable<IMetadataTokenProvider> MarkedPending ()
+		{
+			foreach (var pending in marked_pending.ToArray ())
```
marek-safar, line 199: Why does this need extra buffer?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -401,6 +355,52 @@ bool ProcessPrimaryQueue ()
...
+				marked = true;
+
+				// Some pending items might be processed by the time we get to them.
+				if (Annotations.IsProcessed (pending))
+					continue;
+
+				switch (pending) {
+				case TypeDefinition type:
+					MarkType (type, DependencyInfo.AlreadyMarked, type);
+					Debug.Assert (!Annotations.IsMarkedPending (type));
```
marek-safar, line 376: I think these asserts are not adding value

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -401,6 +355,52 @@ bool ProcessPrimaryQueue ()
...
 		}
 
+		bool ProcessMarkedPending ()
+		{
+			bool marked = false;
+			foreach (var pending in Annotations.MarkedPending ()) {
+				var assembly = GetAssemblyFromMetadataTokenProvider (pending);
+				var assemblyAction = Annotations.GetAction (assembly);
+				if (assemblyAction == AssemblyAction.Skip)
+					continue;
```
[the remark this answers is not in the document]
sbomer: The Initialize logic used to mark things except from Skip/Delete/CopyUsed assemblies. I think CopyUsed was a missing case, and Delete can't be passed on the command-line any more, so we only need to ignore Skip assemblies.

We aren't consistent about what happens when marking things in a Skip assembly. I'm not trying to change that here, just preserving the existing behavior - which doesn't Initialize skip assemblies, but still allows MarkStep to mark members in them - so we need a way to track marked items in Skip assemblies.

Alternatively, I could just remove the check so that Skip assemblies get initialized the same way, but I don't want to make any larger changes to Skip marking.

src/linker/Linker/Annotations.cs
```diff
@@ -239,9 +254,14 @@ public bool IsRelevantToVariantCasting (TypeDefinition type)
 			return types_relevant_to_variant_casting.Contains (type);
 		}
 
-		public void Processed (IMetadataTokenProvider provider)
+		public bool SetProcessed (IMetadataTokenProvider provider)
 		{
-			processed.Add (provider);
+			if (processed.Add (provider)) {
```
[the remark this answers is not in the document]
sbomer: I thought of adding something like MarkProcessed, but it would only be used in one place because MarkStep often does a small amount of extra work for items which are already processed and get marked again.

src/linker/Linker.Steps/MarkStep.cs
sbomer: It needs to go before CheckProcessed because the tracing can happen for already-processed items.

src/linker/Linker/Annotations.cs
sbomer: This is because marked_pending could change as we iterate over it, in general. I could make it a linked list + hashset instead, but then every Mark will need to be added to both. The common case is for MarkStep to call Mark, and CheckProcessed shortly after, so marked_pending tends to be small by the time we need it (usually empty except for the Initialize case).

src/linker/Linker.Steps/MarkStep.cs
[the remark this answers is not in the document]
sbomer: I decided to get rid of the Skip check - this seems more consistent.

src/linker/Linker.Steps/MarkStep.cs
sbomer: Removed

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1532,7 +1541,7 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 				// will call MarkType on the attribute type itself). 
 				// If for some reason we do keep the attribute type (could be because of previous reference which would cause IL2045
 				// or because of a copy assembly with a reference and so on) then we should not spam the warnings due to the type itself.
-				if (sourceLocationMember.DeclaringType != type)
+				if (sourceLocationMember != type && sourceLocationMember.DeclaringType != type)
```
vitek-karas, line 1544: Not really questioning this - but when is the case that type A marks "itself"? (which is the case where source == type).

src/linker/Linker.Steps/ResolveFromXmlStep.cs
```diff
@@ -167,12 +167,14 @@ protected override void ProcessMethod (TypeDefinition type, MethodDefinition met
...
-			Annotations.Mark (method, new DependencyInfo (DependencyKind.XmlDescriptor, _xmlDocumentLocation));
 			Annotations.MarkIndirectlyCalledMethod (method);
 			Annotations.SetAction (method, MethodAction.Parse);
 
-			if (!(bool) customData)
+			if (!(bool) customData) {
 				Annotations.AddPreservedMethod (type, method);
+			} else {
+				Annotations.Mark (method, new DependencyInfo (DependencyKind.XmlDescriptor, _xmlDocumentLocation));
+			}
```
vitek-karas, line 177: This is potentially a change of behavior - unless MarkIndirectlyCalledMethod will end up marking the method anyway (which it might). If it is a change of behavior we should add tests for this.

src/linker/Linker/Annotations.cs
```diff
@@ -49,9 +49,11 @@ public partial class AnnotationStore
 		protected readonly Dictionary<FieldDefinition, object> field_values = new Dictionary<FieldDefinition, object> ();
 		protected readonly HashSet<FieldDefinition> field_init = new HashSet<FieldDefinition> ();
 		protected readonly HashSet<TypeDefinition> fieldType_init = new HashSet<TypeDefinition> ();
-		protected readonly HashSet<IMetadataTokenProvider> marked = new HashSet<IMetadataTokenProvider> ();
+
+		protected readonly HashSet<IMetadataTokenProvider> marked_pending = new HashSet<IMetadataTokenProvider> ();
```
vitek-karas, line 53: I think it would be worth adding a short comment describing what this field stores and what it's for... as there's additional complexity to these, the code is not self-descriptive enough I think

src/linker/Linker/Annotations.cs
```diff
@@ -192,9 +196,20 @@ public void Mark (CustomAttribute attribute, in DependencyInfo reason)
 			Tracer.AddDirectDependency (attribute, reason, marked: true);
 		}
 
+		public IEnumerable<IMetadataTokenProvider> MarkedPending ()
```
vitek-karas, line 199: This sounds like it actually performs the marking. I would rename to at least `GetMarkedPending`...

src/linker/Linker/Annotations.cs
```diff
@@ -241,22 +256,72 @@ public bool IsRelevantToVariantCasting (TypeDefinition type)
 			return types_relevant_to_variant_casting.Contains (type);
 		}
 
-		public void Processed (IMetadataTokenProvider provider)
+		public bool SetProcessed (IMetadataTokenProvider provider)
 		{
-			processed.Add (provider);
+			if (processed.Add (provider)) {
+				Debug.Assert (marked_pending.Remove (provider));
```
vitek-karas, line 262: I would think the removal needs to happen in release mode as well?

src/linker/Linker/Annotations.cs
```diff
@@ -241,22 +256,72 @@ public bool IsRelevantToVariantCasting (TypeDefinition type)
...
+
+			return false;
 		}
 
 		public bool IsProcessed (IMetadataTokenProvider provider)
 		{
 			return processed.Contains (provider);
 		}
 
+		public IEnumerable<TypeDefinition> PendingPreserve ()
```
vitek-karas, line 274: `GetPendingPreserve` ?

src/linker/Linker/Annotations.cs
```diff
@@ -241,22 +256,72 @@ public bool IsRelevantToVariantCasting (TypeDefinition type)
...
+		{
+			foreach (var type in pending_preserve.ToArray ())
+				yield return type;
+		}
+
+		public bool SetAppliedPreserve (TypeDefinition type, TypePreserve preserve)
+		{
+			if (!preserved_types.TryGetValue (type, out (TypePreserve preserve, bool applied) existing)) {
+				throw new InvalidOperationException ();
```
vitek-karas, line 283: This should throw `InternalErrorException` and include description of what went wrong so that we get at least some idea what to look for. (similarly below)

src/linker/Linker/Annotations.cs
marek-safar: Then just return the buffer array

src/linker/Linker.Steps/MarkStep.cs
sbomer: Mark should happen before CheckProcessed, because multiple sources can mark the same field, and all of those dependencies should be traced (which happens whenever we Mark). It was incorrect before, and now the order is enforced by SetProcess which asserts that a processed item should have been marked (pending) first - that's how I caught this.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -399,6 +354,48 @@ bool ProcessPrimaryQueue ()
...
+				if (Annotations.IsProcessed (pending))
+					continue;
+
+				switch (pending) {
+				case TypeDefinition type:
+					MarkType (type, DependencyInfo.AlreadyMarked, type);
+					break;
+				case MethodDefinition method:
+					MarkMethod (method, DependencyInfo.AlreadyMarked, null);
+					// Methods will not actually be processed until we drain the method queue.
```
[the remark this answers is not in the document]
sbomer: Yup - I actually noticed this because MarkMethod sets the MethodAction to Parse, and that wasn't happening correctly in some cases when using EnqueueMethod.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -399,6 +354,48 @@ bool ProcessPrimaryQueue ()
...
+					break;
+				case FieldDefinition field:
```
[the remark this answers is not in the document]
sbomer: Currently they're not handled by Initialize, so I kept it that way but made it an exception instead. A step that marks a property/event would also need to mark the methods like we do in ResolveFromXml, but marking the methods will already mark their properties/events, so it would be a little redundant. There are some other options:
- Allow marking properties/events, and mark all of their methods when processing pending items
  - Custom steps which only want to mark some property methods (like we do for XML) would want to avoid it.
- Allow marking properties/events, but don't do any additional processing for them
  - I am not sure if it would ever make sense to mark properties/events without any of their methods.

src/linker/Linker.Steps/ResolveFromXmlStep.cs
sbomer: The behavior change is that Initialize used to only fully mark a method if it belonged to a marked type, so the method wouldn't actually be marked unless the type was - I will add a test for this. I think the call to Mark here has always been redundant.

The "pending" logic will fully mark any marked methods even if the declaring type wasn't marked, so to preserve the XML behavior we shouldn't mark the method. AddPreservedMethod already expresses the type -> method behavior that we had originally.

src/linker/Linker.Steps/MarkStep.cs
sbomer: Maybe this doesn't make sense - I had made the sourceLocationMember the type itself when fully marking a "pending" marked type - but it should probably just be null.

src/linker/Linker/Annotations.cs
sbomer: Happy to change it to `InternalErrorException` - but I don't really understand why we would use that instead of the more descriptive existing .NET exception types - what's the intended use of one vs the other?

src/linker/Linker/Annotations.cs
sbomer: Added a comment

src/linker/Linker.Steps/MarkStep.cs
sbomer: Replaced it with a null check

src/linker/Linker/Annotations.cs
sbomer: Fixed to return array

src/linker/Linker.Steps/MarkStep.cs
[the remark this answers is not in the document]
vitek-karas: Well - technically for example pure serializer would need to only mark getters, not setters, but it's probably unnecessary optimization.

I think that making it throw is a good idea - it's basically unsupported (as it won't do anything).

src/linker/Linker/Annotations.cs
vitek-karas: It's basically similar to ICE in compilers. I think it's a good differentiation between the linker itself throwing an exception and getting an exception from some underlying framework functionality (or Cecil). There's not much difference, I agree with that... but it feels better to be consistent (the fact that we don't do this everywhere is another matter).

src/linker/Linker.Steps/MarkStep.cs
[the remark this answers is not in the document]
sbomer: With the current behavior they could just mark the getter method, and MarkStep would take care of the property. They just need to avoid marking the Property directly.

src/linker/Linker.Steps/MarkStep.cs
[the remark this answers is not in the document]
sbomer: Actually - since this only allows early marking of MethodDefinition, not MethodReference, I don't think it will change the behavior for generics, but it will improve things if we add a MethodReference case.

src/linker/Linker/Annotations.cs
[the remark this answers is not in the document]
marek-safar: I think I counted more places than 1 and there is also the second combination doing a similar thing with Matk+CheckProcessed. I think it's worth looking if this can be written more effectively

src/linker/Linker.Steps/MarkStep.cs
marek-safar: So could we instead of double marking and adding AlreadyMarked special case not to mark it twice? It seems like this logic is needed only for preserved elements. Could we Mark it from preserve only?

sbomer: Not sure I understand the question - the AlreadyMarked check is there to avoid calling Annotations.Mark twice when a different step marks a member through Annotations (not just for typepreserve or preserved methods). But we still need to call Annotations.Mark if the MarkStep.Mark* method is called directly (for example when we mark declaring types).

src/linker/Linker/Annotations.cs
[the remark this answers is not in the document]
sbomer: You're right - I was looking at CheckProcessed, but there are more for SetProcessed. I will add a helper for these cases.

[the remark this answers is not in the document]
sbomer: I added a helper for the Mark+SetProcessed case, but I left CheckProcessed alone since derived classes might override it.

sbomer:
I've addressed all the feedback so far - PTAL

vitek-karas (approved):

src/linker/Linker.Steps/MarkStep.cs
marek-safar: What I meant is that you can simplify this handling by not doing the double-checking with AlreadyMarked and marked_pending at one line and removing it at the next line.

I think replacing the introduced redundant logic should be fixed. 

- For AlreadyMarked logic you could split the methods and call them with no DependencyInfo from `ProcessMarkedPending` and skip all the checks because the flow went through them already.

- For Mark+CheckProcessed logic

You could replace 
```
			Annotations.Mark (assembly, reason);
			if (CheckProcessed (assembly))
				return;
```

code flow with simple insertion to `processed` and not to touch `marked_pending` at all.

sbomer: Thanks for clarifying - I made this optimization for the `Mark`+`SetProcessed` cases, but like I mentioned in https://github.com/mono/linker/pull/1768#discussion_r562110659, I'd rather not change the behavior of existing callsites that use `CheckProcessed` here, same goes for the `AlreadyMarked` pattern. I opened https://github.com/mono/linker/issues/1776 to track this.

vitek-karas: I understand the perf impact, but it feel fragile to have a "already marked" version and "non marked" version and try to make sure that all callsites are correct. The overall impact should be pretty small (assuming we're not marking 10000s of methods from custom steps).

sbomer: I agree about the "already marked"/"non marked" split - there might be other ways to address https://github.com/mono/linker/issues/1776 without a split. Also FWIW I did a rough perf measurement on helloworld with this change, and it's probably within noise, but if anything it was a slight improvement (maybe because we avoid walking types in Initialize, though I didn't dig into it).

PR #1770, opened 2021-01-20
Add an option to remove exception filters.

vargaz:
Its controlled by the --remove-filters command line option.

This works by rewriting:

```csharp
catch (Exception ex) when (<cond>) {
  ..
}
```

into:

```csharp
catch (Exception ex) {
  if (!<cond>)
    throw ex;
  ..
}
```

### DIFF
--- a/src/linker/Linker.Steps/RewriteExceptionFiltersStep.cs
+++ b/src/linker/Linker.Steps/RewriteExceptionFiltersStep.cs
@@ -1,0 +1,75 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Mono.Collections.Generic;
+
+namespace Mono.Linker.Steps
+{
+	//
+	// This step removes filter clauses by doing the following transformation:
+	// catch (Exception ex) when (<cond>) {
+	//   ..
+	// }
+	// ->
+	// catch (Exception ex) {
+	//   if (!<cond>)
+	//	 throw ex;
+	//	 ..
+	//
+	// }
+	public class RewriteExceptionFiltersStep : BaseStep
+	{
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			RewriteBodies (assembly.MainModule.Types);
+		}
+
+		void RewriteBodies (Collection<TypeDefinition> types)
+		{
+			foreach (var type in types) {
+				if (type.HasMethods) {
+					foreach (var method in type.Methods) {
+						if (method.HasBody && method.Body.HasExceptionHandlers)
+							RewriteBody (method);
+					}
+				}
+
+				if (type.HasNestedTypes)
+					RewriteBodies (type.NestedTypes);
+			}
+		}
+
+		void RewriteBody (MethodDefinition method)
+		{
+			if (!(method.HasBody && method.Body.ExceptionHandlers.Any (clause => clause.HandlerType == ExceptionHandlerType.Filter)))
+				return;
+			var body = method.Body;
+			var processor = method.Body.GetILProcessor ();
+			foreach (var clause in method.Body.ExceptionHandlers.Where (clause => clause.HandlerType == ExceptionHandlerType.Filter)) {
+				var oldStart = clause.FilterStart;
+				// Add a dup at the beginning since the catch clause expects the exception object on the stack
+				var newStart = Instruction.Create (OpCodes.Dup);
+				processor.InsertBefore (clause.FilterStart, newStart);
+
+				// endfilter -> brtrue <catch> + throw
+				var endfilter = clause.HandlerStart.Previous;
+				if (endfilter.OpCode.Code != Code.Endfilter)
+					throw new NotImplementedException (endfilter.ToString ());
+				endfilter.OpCode = OpCodes.Brtrue;
+				endfilter.Operand = clause.HandlerStart;
+				// The exception is on the stack
+				var rethrow = Instruction.Create (OpCodes.Throw);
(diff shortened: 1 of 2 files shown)


MichalStrehovsky:
This test might be useful in this effort:

```csharp
using System;

class P
{
    interface Q
    {
        static int s_counter;

        static void Main()
        {
            try
            {
                throw new Exception();
            }
            catch (ArgumentException) when (True())
            {
                Console.WriteLine("Unreached");
            }
            catch (Exception) when (Throw())
            {
                Console.WriteLine("Unreached");
                s_counter = -1;
            }
            catch (Exception) when (False())
            {
                Console.WriteLine("Unreached");
                s_counter = -1;
            }
            catch (Exception)
            {
                Console.WriteLine("Works as expected");
                s_counter |= 4;
            }

            if (s_counter == 7)
            {
                Console.WriteLine("Success");
            }
            else
            {
                Console.WriteLine("Fail");
            }
        }

        static bool Throw()
        {
            s_counter |= 1;
            throw new Exception();
        }

        static bool False()
        {
            s_counter |= 2;
            return false;
        }

        static bool True()
        {
            s_counter = -1;
            return true;
        }
    }
}
```

src/linker/Linker.Steps/RemoveFiltersStep.cs
```diff
@@ -0,0 +1,76 @@
...
+		{
+			foreach (var type in types) {
+				if (type.IsInterface)
+					continue;
+
+				if (!type.HasMethods)
+					continue;
+
+				foreach (var method in type.Methods) {
+                    if (method.HasBody && method.Body.ExceptionHandlers.Any (clause => clause.HandlerType == ExceptionHandlerType.Filter))
```
marek-safar, line 43: I think should just check for `method.Body.HasExceptionHandlers` and leave the rest to RewriteBody

src/linker/Linker.Steps/RemoveFiltersStep.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            foreach (var assembly in Context.Annotations.GetAssemblies ()) {
+                RewriteBodies (assembly.MainModule.Types);
+            }
+        }
+
+		void RewriteBodies (Collection<TypeDefinition> types)
+		{
+			foreach (var type in types) {
+				if (type.IsInterface)
```
marek-safar, line 36: What about default interface methods?

src/linker/Linker.Steps/RemoveFiltersStep.cs
```diff
@@ -0,0 +1,76 @@
...
+		protected override void Process ()
+        {
+            foreach (var assembly in Context.Annotations.GetAssemblies ()) {
+                RewriteBodies (assembly.MainModule.Types);
+            }
+        }
+
+		void RewriteBodies (Collection<TypeDefinition> types)
+		{
+			foreach (var type in types) {
```
marek-safar, line 35: This does not handle nested types

src/linker/Linker.Steps/RemoveFiltersStep.cs
```diff
@@ -0,0 +1,76 @@
...
+    //   ..
+    // }
+    // ->
+    // catch (Exception ex) {
+    //   if (!<cond>)
+    //     throw ex;
+    //     ..
+    //     
+    // }
+	public class RemoveFiltersStep : BaseStep
```
marek-safar, line 24: Suggestion: Rename it to `RewriteExceptionFiltersStep`

src/linker/Linker.Steps/RemoveFiltersStep.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+		protected override void Process ()
```
marek-safar, line 26: override `ProcessAssembly (AssemblyDefinition assembly)` instead

src/linker/Linker.Steps/RemoveFiltersStep.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            foreach (var assembly in Context.Annotations.GetAssemblies ()) {
+                RewriteBodies (assembly.MainModule.Types);
```
marek-safar, line 29: This needs to check if we are linking assembly otherwise do something else (e.g. issue warning).

src/linker/Linker.Steps/RemoveFiltersStep.cs
```diff
@@ -0,0 +1,76 @@
...
+                endfilter.OpCode = OpCodes.Brtrue;
+                endfilter.Operand = clause.HandlerStart;
+                // The exception is on the stack
+                var rethrow = Instruction.Create (OpCodes.Throw);
+                processor.InsertAfter (endfilter, rethrow);
+
+                clause.HandlerType = ExceptionHandlerType.Catch;
+                clause.TryEnd = newStart;
+                clause.HandlerStart = newStart;
+                clause.CatchType = method.Module.ImportReference (BCL.FindPredefinedType ("System", "Exception", Context));
```
marek-safar, line 72: Please add tests for the logic

marek-safar:
I suspect that we'll need to run this code during the inner dev loop with an interpreter for consistency where hooking up linker would be a pain.

I'm wondering if we should take this as an opportunity to introduce a new task inside dotnet/runtime instead of injecting this functionality into linker. We have other similar needs for AOT pipeline which could be developed there instead of inside linker. It'd also allow us to run these steps heavily in parallel if we go with SRM instead of Cecil.

We would have run all of the following "steps" at the same time inside that task

- exception filters rewriter
- IL bodies remover
- pinvoke generator
- icalls generator

/cc @lewing

vargaz:
This currently can't handle the case when there are more than 1 filter clause for the same try range.

Closing for now in favor of: https://github.com/mono/mono/pull/20796

PR #1771, opened 2021-01-20
Constant propagation directly from MarkStep

vitek-karas:
Call constant propagation and branch removal processing directly from `MarkStep`:
* Renames `RemoveUnreachableBlocksStep` to `UnreachableBlocksOptimizer` - it's not a step anymore
* Rename members to use the `_name` naming convention (which is also used by `MarkStep`)
* Removes the "go over all methods in all assemblies and process them" logic
* Some small changes to the `ProcessMethod` API to be more suitable for the `MarkStep`
* The class is now instantiated by `MarkStep` and only lives as long as `MarkStep` does
* `MarkStep.ProcessMethod` calls the optimizer's `ProcessMethod` before doing anything else on the method (this is because the optimizer can modify methods body, so mark step should only look at the modified version of the method)

Some interesting statistics (on console helloworld):
* Before this change the optimizer processed 73K methods (every method at least once - total number of methods is 72K). After the change the optimizer only processes 12K methods (so a huge reduction). This also means that the large dictionary maintained by the optimizer on all processed methods is now 20% of before this change.
* Max stack size increased to 62 - this is because of different order of processing methods. The 62 comes from `ThrowHelpers` which initializes lot of resource strings (each a separate call to a different method), so at one point all of these dependencies are on the stack. This is not a problem, just slightly higher memory usage. It would become problematic if the stack depth reached several 1000s which is VERY unlikely to happen.
* Number of rewritten methods is down to 20% (from 746 to 117) - all of those not rewritten are actually not used by the app
* As far as I can tell, it produces the exact same output.

Very rough perf measurement on Blazor template app:
* Before all the constant prop changes (SDK from early December 2020): 6.1 seconds (on average)
* After this change: 3.4 seconds (on average)

Part of #1733


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -57,6 +56,7 @@ public partial class MarkStep : IStep
 		protected List<MethodBody> _unreachableBodies;
 
 		readonly List<(TypeDefinition Type, MethodBody Body, Instruction Instr)> _pending_isinst_instr;
+		UnreachableBlocksOptimizer _removeUnreachableBlocksStep;
```
sbomer, line 59: Maybe rename the variable too (to not be a Step)

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
```diff
@@ -10,12 +10,13 @@
 namespace Mono.Linker.Steps
```
sbomer, line 10: Should it move to the Mono.Linker namespace and directory?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2505,6 +2506,8 @@ protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo
 				throw new ArgumentOutOfRangeException ($"Internal error: unsupported method dependency {reason.Kind}");
 #endif
 
+			_removeUnreachableBlocksStep.ProcessMethod (method);
```
sbomer, line 2509: Could this be moved to after CheckProcessed?

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
```diff
@@ -64,54 +65,17 @@ public ProcessingNode (in ProcessingNode other, int newLastAttempStackVersion)
...
-					continue;
-
-				ProcessMethods (assembly.MainModule.Types);
-			}
-		}
-
-		void ProcessMethods (Collection<TypeDefinition> types)
-		{
-			foreach (var type in types) {
-				if (type.IsInterface)
```
sbomer, line 90: Looks like we no longer check for interfaces - does it just mean that default interface methods can have branches removed?

sbomer (approved):

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
marek-safar: I'd keep it inside Steps

marek-safar (approved):

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
vitek-karas: I thought about it - it's basically now part of MarkStep (in a way)... so I would like to keep it in the Steps.

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
vitek-karas: Good catch - and basically yes. I fixed it (unintentionally 😉 ). So I now added at least a basic test that branch removal works within default interface methods.

PR #1773, opened 2021-01-21
Ensure that CopyUsed entry point gets marked

sbomer:
I noticed this when some of my lazy loading tests started failing after a rebase: an explicit copyused root assembly doesn't get the entry point marked because of the missing case in Initialize. It looks like the bug has been there for a while, but it didn't show up because our tests used `[SetupLinkerUserAction ("copyused")]` - in this case, `ResolveFromAssemblyStep` would change the action of a root assembly to `link`. But when copyused is set explicitly for an assembly via "-p", we forget to root it.

This change fixes it by treating `copyused` the same as `link` in the root step - this will also have the side effect of changing the action to `copy` if the root mode was `AllMembers`.

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Libraries/CopyUsedAssemblyWithMainEntryRoot.cs
+++ b/test/Mono.Linker.Tests.Cases/Libraries/CopyUsedAssemblyWithMainEntryRoot.cs
@@ -1,5 +1,6 @@
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 using Mono.Linker.Tests.Cases.Expectations.Metadata;
+using Mono.Linker.Tests.Cases.Libraries.Dependencies;
 
 namespace Mono.Linker.Tests.Cases.Libraries
 {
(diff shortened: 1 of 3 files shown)


vitek-karas (approved):

test/Mono.Linker.Tests.Cases/Libraries/CopyUsedAssemblyWithMainEntryRoot.cs
```diff
@@ -9,11 +10,16 @@ namespace Mono.Linker.Tests.Cases.Libraries
 	[Kept]
 	[KeptMember (".ctor()")]
 	[SetupLinkerAction ("copyused", "test")]
+	[SetupCompileBefore ("lib.dll", new[] { "Dependencies/CopyUsedAssemblyWithMainEntryRoot_Lib.cs" })]
+	[KeptMemberInAssembly ("lib.dll", typeof (CopyUsedAssemblyWithMainEntryRoot_Lib), "Used()")]
+	// Marked CopyUsed assemblies are not fully marked like Copy assemblies, so the Unused dependency is not kept.
+	[RemovedMemberInAssembly ("lib.dll", typeof (CopyUsedAssemblyWithMainEntryRoot_Lib), "Unused()")]
```
marek-safar, line 16: This looks wrong, copyused should copy whole assembly. the only difference is that it will copy when anything from the assembly is used

sbomer: Agreed - I was surprised by this too - but I'm not trying to fix that here. I filed https://github.com/mono/linker/issues/1777 for it.

PR #1778, opened 2021-01-23
Warn on called methods annotated with RequiresUnreferencedCode

mateoatr:
See #1607. Most of the changes in this PR were already in #1639. We only run through `ProcessRequiresUnreferencedCode` whenever the annotated method was either called or found in a reflection scenario. A test case was added for making sure that methods marked with RequiresUnreferencedCodeAttribute in copied assemblies do not trigger any warnings (unless called or referenced through reflection).


src/ILLink.RoslynAnalyzer/Resources.resx
```diff
@@ -121,7 +121,7 @@
     <value>RequiresUnreferencedCodeAnalyzer</value>
   </data>
   <data name="RequiresUnreferencedCodeAnalyzerMessage" xml:space="preserve">
-    <value>Calling '{0}' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. {1}. {2}</value>
+    <value>'{0}' method has 'RequiresUnreferencedCodeAttribute' which can break functionality when trimming application code. {1}. {2}</value>
```
vitek-karas, line 124: Maybe we could just say "Using method which has RequiresUnreferencedCode ..." - it reads better to me (and it's a bit more precise - the linker really thinks the app is going to use the method"

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -24,11 +28,20 @@ public static void Main ()
...
+			TestCovariantReturnCallOnDerived ();
+			TestRequiresInMethodFromCopiedAssembly ();
+			TestRequiresThroughReflectionInMethodFromCopiedAssembly ();
 		}
 
 		[ExpectedWarning ("IL2026",
-			"Calling 'Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.RequiresWithMessageOnly()' " +
-			"which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. " +
+			"'Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.RequiresWithMessageOnly()' method " +
+			"has 'RequiresUnreferencedCodeAttribute' which can break functionality when trimming application code. " +
```
vitek-karas, line 44: Since you're here - can you just remove the full text validation - we should validate only the warning code and than the presence of the value parameters for the message - not the message itself.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -24,11 +28,20 @@ public static void Main ()
 			TestRequiresSuppressesWarningsFromReflectionAnalysis ();
 			TestDuplicateRequiresAttribute ();
 			TestRequiresUnreferencedCodeOnlyThroughReflection ();
+			TestBaseTypeVirtualMethodRequiresUnreferencedCode ();
+			TestTypeWhichOverridesMethodVirtualMethodRequiresUnreferencedCode ();
+			TestTypeWhichOverridesMethodVirtualMethodRequiresUnreferencedCodeOnBase ();
+			TestStaticCctorRequiresUnreferencedCode ();
+			TestDynamicallyAccessedMembersWithRequiresUnreferencedCode (typeof (DynamicallyAccessedTypeWithRequiresUnreferencedCode));
```
vitek-karas, line 35: This call should warn - we have no validation for that.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2441,21 +2445,21 @@ protected virtual bool MarkMethods (TypeDefinition type, in DependencyInfo reaso
...
 
 		protected internal void MarkIndirectlyCalledMethod (MethodDefinition method, in DependencyInfo reason, IMemberDefinition sourceLocationMember)
 		{
-			MarkMethod (method, reason, sourceLocationMember);
+			MarkMethod (method, reason, new MessageOrigin (sourceLocationMember));
 			Annotations.MarkIndirectlyCalledMethod (method);
 		}
 
-		protected virtual MethodDefinition MarkMethod (MethodReference reference, DependencyInfo reason, IMemberDefinition sourceLocationMember)
+		protected virtual MethodDefinition MarkMethod (MethodReference reference, DependencyInfo reason, MessageOrigin? origin)
```
vitek-karas, line 2457: Do we need this change to take `MessageOrigin`? It seems we only ever use the `MemberDefinition` from it, so the previous state which passed the member directly seem to work the same way.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2486,9 +2487,50 @@ protected virtual MethodDefinition MarkMethod (MethodReference reference, Depend
 
 			EnqueueMethod (method, reason);
 
+			// All override methods should have the same annotations as their base methods (else we will produce warning IL2046.)
+			// When marking override methods with RequiresUnreferencedCode on a type annotated with DynamicallyAccessedMembers,
+			// we should only issue a warning for the base method.
+			if (origin != null && (reason.Kind != DependencyKind.DynamicallyAccessedMember || !method.IsVirtual || Annotations.GetBaseMethods (method) == null))
+				ProcessRequiresUnreferencedCode (method, (MessageOrigin) origin, reason.Kind);
```
vitek-karas, line 2494: I don't understand why we don't want to warn for `DynamicallyAccessedMember`. The method is kept because linker thinks that it will be used via reflection - so it should warn.

vitek-karas (changes requested):

src/linker/Linker.Steps/MarkStep.cs
mateoatr: We should still warn on these. The second part of the if condition is there to make sure we warn on covariant returns (https://github.com/mono/linker/pull/1639#discussion_r539428886).

src/linker/Linker.Steps/MarkStep.cs
mateoatr: This was changed so that we don't lose the line information of the warnings that originate from `MarkInstruction`: https://github.com/mono/linker/blob/e7af1b8756a5845ce46909980d64f2b4c2702100/src/linker/Linker.Steps/MarkStep.cs#L3078
 We can change this back to IMemberDefinition if willing to loose that.

src/ILLink.RoslynAnalyzer/Resources.resx
mateoatr: I'll change this -- I was originally following the shape that we already have for warnings related to Dynamically Accessed Members: https://github.com/mono/linker/pull/1624.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Makes sense

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -163,14 +157,141 @@ static void RequiresUnreferencedCodeOnlyThroughReflection ()
...
+
+		[ExpectedWarning ("IL2026", "--Method--")]
+		static void TestRequiresInMethodFromCopiedAssembly ()
+		{
+			var tmp = new RequiresUnreferencedCodeInCopyAssembly ();
+			tmp.Method ();
+		}
+
+		[ExpectedWarning ("IL2026", "--MethodCalledThroughReflection--")]
+		static void TestRequiresThroughReflectionInMethodFromCopiedAssembly ()
```
vitek-karas, line 290: Given the use case in DispatchProxy (that one is about DynamicallyAccessedMember, but it's similar) - can you please add another test case where we apply the `All` annotation to an interface and we validate that it warns if one of the methods on the interface is annotated with `RequiresUnreferencedCode` - maybe a more complex one if the method is on another interface, for example:

```C#
interface A { [RequiresUnreferencedCode] void DoA (); }
interface B : A { void DoB (); }

RequiresAll(typeof(B));
```

vitek-karas (approved):

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -250,59 +253,85 @@ private void MarkEntireTypeInternal (TypeDefinition type, bool includeBaseTypes,
...
 			}
-			MarkCustomAttributes (type, new DependencyInfo (DependencyKind.CustomAttribute, type), type);
+			MarkCustomAttributes (type, new DependencyInfo (reason.Kind == DependencyKind.AccessedViaReflection ? reason.Kind : DependencyKind.CustomAttribute, type), type);
 			MarkTypeSpecialCustomAttributes (type);
 
 			if (type.HasInterfaces) {
 				foreach (InterfaceImplementation iface in type.Interfaces) {
-					MarkInterfaceImplementation (iface, type);
+					if (reason.Kind == DependencyKind.AccessedViaReflection)
+						MarkEntireTypeInternal (iface.InterfaceType.Resolve (), includeBaseTypes: true, reason, type, typesAlreadyVisited);
```
vitek-karas, line 278: I think we need to call `MarkInterfaceImplementation` anyway. If the interface is a generic instantiation we would lose that information and for example would not correctly apply `DynamicallyAccessedMembers`.
I think we'll need to look into this a bit more - the same applies to base types, and potentially even methods. I filed https://github.com/mono/linker/issues/1787 to look into this some more.

vitek-karas: Also - it's technically possible for Resole to return null, the should not crash on it. See handling of base types.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2574,15 +2641,18 @@ protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo
...
 				}
 			}
 
 			MarkMethodSpecialCustomAttributes (method);
 
-			if (method.IsVirtual)
+			if (method.IsVirtual) {
 				_virtual_methods.Add (method);
+				if (reason.Kind == DependencyKind.AccessedViaReflection)
+					ProcessRequiresUnreferencedCode (method, new MessageOrigin (reason.Source as IMemberDefinition), reason.Kind);
```
vitek-karas, line 2654: This feels wrong - I understand it's because MarkEntireType doesn't call MarkMethod - maybe it should...
Can you also check if there are other places which side-step MarkMethod and end up here instead? Those could all be potential issues as well.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -250,59 +253,85 @@ private void MarkEntireTypeInternal (TypeDefinition type, bool includeBaseTypes,
...
+				MarkEntireTypeInternal (baseTypeDefinition, includeBaseTypes: true, GetDependencyInfo (baseTypeDefinition, reason), type, typesAlreadyVisited);
 			}
-			MarkCustomAttributes (type, new DependencyInfo (DependencyKind.CustomAttribute, type), type);
+			MarkCustomAttributes (type, new DependencyInfo (reason.Kind == DependencyKind.AccessedViaReflection ? reason.Kind : DependencyKind.CustomAttribute, type), type);
 			MarkTypeSpecialCustomAttributes (type);
 
 			if (type.HasInterfaces) {
 				foreach (InterfaceImplementation iface in type.Interfaces) {
-					MarkInterfaceImplementation (iface, type);
+					if (reason.Kind == DependencyKind.AccessedViaReflection)
```
vitek-karas, line 277: I would prefer to pass this difference in as a separate parameter. It would make it cleaner in my mind. Just like we have a parameter which dictates if base types should be included or not, we should add one which dictates if interfaces should be fully included or not.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -261,77 +261,52 @@ private void MarkEntireTypeInternal (TypeDefinition type, bool includeBaseTypes,
...
 				}
 			}
 
 			if (type.HasMethods) {
 				foreach (MethodDefinition method in type.Methods) {
-					// Probably redundant since we EnqueueMethod below anyway.
-					Annotations.Mark (method, GetDependencyInfo (method, reason));
-					Annotations.SetAction (method, MethodAction.ForceParse);
-					EnqueueMethod (method, GetDependencyInfo (method, reason));
+					MarkMethod (method, new DependencyInfo (reason.Kind == DependencyKind.AccessedViaReflection ? reason.Kind : DependencyKind.MemberOfType, type), new MessageOrigin (reason.Source as IMemberDefinition));
```
vitek-karas: We should keep the SetAction - the difference in behavior will be pretty small, but still - no reason to NOT do that.

vitek-karas (approved):
Other than the .Resolve returning null concern in https://github.com/mono/linker/pull/1778#discussion_r566165151 this looks good.

mateoatr:
>Other than the .Resolve returning null concern in #1778 (comment) this looks good.

Doesn't

https://github.com/mono/linker/blob/bcc7dec192953e36ef9274c54f7a636d25b02f09/src/linker/Linker.Steps/MarkStep.cs#L277-L278

Solve this?

vitek-karas:
Of course it does... sorry - my bad.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -285,10 +292,8 @@ private void MarkEntireTypeInternal (TypeDefinition type, bool includeBaseTypes,
 
 			if (type.HasMethods) {
 				foreach (MethodDefinition method in type.Methods) {
-					// Probably redundant since we EnqueueMethod below anyway.
-					Annotations.Mark (method, new DependencyInfo (DependencyKind.MemberOfType, type));
 					Annotations.SetAction (method, MethodAction.ForceParse);
-					EnqueueMethod (method, new DependencyInfo (DependencyKind.MemberOfType, type));
+					MarkMethod (method, new DependencyInfo (reason.Kind == DependencyKind.AccessedViaReflection ? reason.Kind : DependencyKind.MemberOfType, type), new MessageOrigin (reason.Source as IMemberDefinition));
```
sbomer, line 296: Why the special case for `AccessedViaReflection` here - is it just to ensure the reason is one of the expected ones in ProcessRequiresUnreferencedCode?
- Can we do this without changing the reason for the marked method?
- If we go with the special case, should we do the same for `DependencyKind.DynamicDependency` - since we can get here from `MarkDynamicDependency`?

mateoatr: > Why the special case for AccessedViaReflection here - is it just to ensure the reason is one of the expected ones in ProcessRequiresUnreferencedCode?

Yes.

> Can we do this without changing the reason for the marked method?

We can do this if we add `DynamicDependency` to the list of reasons for marking a method... I think it makes sense and should make code cleaner/easier to read.

sbomer: I'm not sure you'd need to add `DynamicDependency` if we continue passing `MemberOfType` - I think you'd need to add `MemberOfType` to the list of cases in `ProcessRequiresUnreferencedCode`.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -584,11 +589,15 @@ void ProcessOverride (OverrideInformation overrideInformation)
 			// Only track instantiations if override removal is enabled and the type is instantiated.
 			// If it's disabled, all overrides are kept, so there's no instantiation site to blame.
 			if (_context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, method) && isInstantiated) {
-				MarkMethod (method, new DependencyInfo (DependencyKind.OverrideOnInstantiatedType, method.DeclaringType), method.DeclaringType);
+				// MarkMethod uses the MessageOrigin passed in here to determine the member which causes the method
+				// to be marked. This is important to avoid spurious warnings as virtual overrides can end up being
+				// marked without being referenced directly from user code. It is important to pass the base method
+				// here so that MarkMethod can see that it was marked because of an override.
+				MarkMethod (method, new DependencyInfo (DependencyKind.OverrideOnInstantiatedType, method.DeclaringType), new MessageOrigin (@base));
```
sbomer, line 596: Is this logic still needed to avoid spurious warnings? It looks to me like we don't, since ProcessRequiresUnreferencedCode doesn't have a case for OverrideOnInstantiatedType (or maybe it's missing a case?)

More generally (not directly related to your change) - I'm not sure what the function of sourceLocationMember is here - I think we should just pass null. If I'm not missing something, it looks like:
- At this point, method and @base are MethodDefinitions
- MarkMethod will only use sourceLocationMember for instantiated generic types or methods (the latter in GetOriginalMethod)

I think the sourceLocationMember is only used for instantiated generics, but can't have instantiated generics here, so we shouldn't use it. Does that sound right to you? @vitek-karas I think you introduced these - what do you think?

src/linker/Linker.Steps/MarkStep.cs
mateoatr: Just tried not special casing. If we simple reuse `reason.Kind` then we can end up passing `BaseType` or `NestedType`, which then throws, since these are not valid reasons for marking a method.
The reason to special case here is so that we warn on annotated methods on a type that is kept because of `DynamicallyAccessedMembers.All`. We don't have `MembersOfType` on the list of reasons to process `ProcessRequiresUnreferencedCode` since we don't want to emit warnings on methods that are marked because of a copied assembly (we only want to warn if there's a possibility of the method being referenced through reflection or if it's directly called).
I think it's easier to just keep this special case and of course, special case `DynamicDependency`  too.

sbomer: Ah, that makes sense, to avoid warnings for copy assemblies. Thanks for the explanation!

Just another thought - now I am wondering if we could use sourceLocationMember only, instead of DependencyInfo, for this purpose - since DependencyInfo was designed to be more granular than source locations, and what we need here are source locations. But that might require more changes to the sourceLocationMember logic - I'm not sure. I'll leave it as a suggestion in case you think it's worth looking into, but I think special-casing `AccessedViaReflection`/`DynamicDependency` is fine too.

sbomer (approved):
LGTM otherwise!

src/linker/Linker.Steps/MarkStep.cs
mateoatr: You are right it's no longer needed; when calling MarkMethod from here we will never emit a warning.
> More generally (not directly related to your change) - I'm not sure what the function of sourceLocationMember is here - I think we should just pass null.

Yes, since there's no way we will emit a warning in MarkMethod at this point, we can pass null here. Other than helping us report the warning, sourceLocationMember becomes relevant in the context of MarkMethod when dealing with generics, as you noted (we could end here:

https://github.com/mono/linker/blob/ba990ce8336dd278b31007aac30ac85f120d3683/src/linker/Linker.Steps/MarkStep.cs#L2267

 I'll change the origin here to null, unless @vitek-karas disagrees / there a reason to pass @base/method.DeclaringType here.

PR #1781, opened 2021-01-25
[WIP] Remove keep-facades option

### LINKED ISSUE #1740: Type forwarder handling needs cleanup
The type forwarder logic needs cleanup. A few problem areas:
- MarkStep marks exported types that point to resolved types, regardless of whether that type was actually referenced through the forwarder: https://github.com/mono/linker/blob/7a095faf3925689f2a3450da1a76fdd79a070de9/src/linker/Linker.Steps/MarkStep.cs#L363-L368
- `--keep-facades` preserves type forwarder assemblies from the entire reference closure, even if they weren't reachable through marked code. This can result in empty preserved assemblies (since we still remove the actual forwarder). We even have a testcase for this behavior: https://github.com/mono/linker/blob/7a095faf3925689f2a3450da1a76fdd79a070de9/test/Mono.Linker.Tests.Cases/TypeForwarding/UnusedForwarderWithAssemblyLinkedAndFacadesKept.cs#L19-L31
- We resolve type forwarders referenced from `DynamicDependency` or reflection, but we don't actually ensure that the forwarder assembly is kept in such cases. For example, adding `[KeptAssembly ("FacadeAssembly.dll")]` causes this testcase to fail: https://github.com/mono/linker/blob/7a095faf3925689f2a3450da1a76fdd79a070de9/test/Mono.Linker.Tests.Cases/DynamicDependencies/DynamicDependencyOnForwardedType.cs#L21-L28

See also some TODOs in the code:
https://github.com/mono/linker/blob/7a095faf3925689f2a3450da1a76fdd79a070de9/src/linker/Linker.Steps/MarkStep.cs#L1183-L1185

@marek-safar suggested getting rid of `--keep-facades`. I think this is OK since we don't set this or have a public option for it in the SDK, but we should still keep type forwards when they are referenced dynamically. We also need to consider how a future "copyoriginal" action would handle references to type forwarders (see https://github.com/mono/linker/issues/1669#issuecomment-738940392). My suggestion is to emit an error in this case, and add support for it in the future if it proves necessary. One challenge is the fact that resolving a typeref through cecil will transparently dereference through the type forwarder without giving the calling code an opportunity to make a decision about the exported type.

I propose we:
- Get rid of `--keep-facades`
- Don't mark exported type forwards that just happen to point to a marked type
- Mark type forwards that are resolved dynamically
- Keep all marked type forwards

marek-safar:
It was introduced when .net core added framework initial trimming for
libraries build. This is a third way to achieve same thing but less
logical than using -p or -a option and it's no longer used by libraries
build.

Fixes #1740


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1260,7 +1236,19 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
 			MarkCustomAttributes (assembly.MainModule, new DependencyInfo (DependencyKind.AssemblyOrModuleAttribute, assembly.MainModule), null);
 
 			if (assembly.MainModule.HasExportedTypes) {
-				// TODO: This needs more work accross all steps
+				foreach (var exportedType in assembly.MainModule.ExportedTypes) {
+					_context.Annotations.Mark (exportedType, new DependencyInfo (DependencyKind.ExportedType, exportedType));
```
sbomer, line 1240: This looks like a significant behavior change:
- We didn't used to keep type forwards in copy assemblies - this could result in some extra facade dlls being kept that weren't needed previously. It's probably uncommon, since facades are usually going to be copyused as part of the framework, but it's unclear how copyused is supposed to behave now. Should a marked copyused assembly also have exported types kept? (This would be a bigger issue since copyused are more likely to pull in facades).
- I *think* this will prevent SweepStep from sweeping assembly refs from copy assemblies, so it will never update scopes for Copy assemblies. Same question - how should this work for copyused assemblies?

If we go this direction and mark all copy type forwarders, I think we can simplify a few cases in SweepStep:
- get rid of the copy case from `UpdateAssembliesReferencingRemovedAssembly`
- get rid of the `SweepTypeForwarders` handling for copy assemblies

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1260,7 +1236,19 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
...
+
+					TypeDefinition type = exportedType.Resolve ();
+					if (type == null) {
+						if (!_context.IgnoreUnresolved)
+							_context.LogError ($"Exported type '{type.Name}' cannot be resolved", 1038);
```
sbomer, line 1245: Some of the facades have platform-specific type forwards, so I'm worried this could introduce spurious errors. It may be better to ignore these resolution failures. We should at least check the warnings for a console app with "-p copy <forwarder>" for common forwarders like System.Runtime.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -147,11 +147,7 @@ protected void ProcessAssemblyAction (AssemblyDefinition assembly)
 			case AssemblyAction.AddBypassNGen:
 				// FIXME: AddBypassNGen is just wrong, it should not be action as we need to
 				// turn it to Action.Save here to e.g. correctly update debug symbols
```
vitek-karas, line 149: Should we also delete the comment now?

marek-safar: I think that still wrong. The whole AddBypassNGen thing is weird :-) and has no tests.

src/linker/Linker.Steps/MarkStep.cs
marek-safar: I don't think ignoring them is correct behavior because it could generate invalid output. The error is ignored if you run with `--skip-unresolved`

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2315,6 +2303,11 @@ void ApplyPreserveInfo (TypeDefinition type)
 				var di = new DependencyInfo (DependencyKind.TypePreserve, type);
 
 				switch (members) {
+				case TypePreserveMembers.All:
```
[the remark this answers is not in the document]
marek-safar: they are enum values stored into a different annotation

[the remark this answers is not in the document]
sbomer: I can see that they're different enums. What I mean is, we have this case above:
```csharp
case TypePreserve.All:
	MarkFields (type, true, di);
	MarkMethods (type, di, type);
	return;
```

Why do we need `TypePreserveMembers.All` if it is only duplicating this logic? Could we do `Annotations.SetPreserve (type, TypePreserve.All)` below, instead of `Annotations.SetMembersPreserve (type, TypePreserveMembers.All)`? That seems to be the only place we use it.

Issue #1786, opened 2021-01-28
RequiresAssemblyFilesAttribute error code and messages


MichalStrehovsky:
I had a look at #1722 because I wanted to steal stuff from it for NativeAOT, but it's missing some important things.

* The error message is missing:

https://github.com/mono/linker/blob/92b9b51bb8548320e40fc432efcb395156ad1445/src/ILLink.RoslynAnalyzer/Resources.resx#L138-L142

* It claimed IL3002 error code but there's no documentation for it in the [master error code registry](https://github.com/mono/linker/blob/master/docs/error-codes.md).

agocke:
@tlakollo Can you pick this up? I think writing a resource message is the highest priority. Right now the analyzer doesn't really print anything.

tlakollo:
So far I created a PR that includes a default message and the arguments.
Seems like other error codes have a help link uri like in https://github.com/mono/linker/pull/1736, not sure if we need one for IL3002
Also not sure if we are not including the IL3XXX codes in the master error code registry in order to not confuse with other linker codes and patterns.

mateoatr:
> Also not sure if we are not including the IL3XXX codes in the master error code registry in order to not confuse with other linker codes and patterns

We should at least link to [IL3000](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/il3000) and IL3001 (https://github.com/mono/linker/issues/1498#issuecomment-695054111). Also, I think we should create a similar docs page for IL3002 once we decide on a specific message.

MichalStrehovsky:
If we expect that illink will eventually also have analysis for single-file friendliness, this is a illink error code like any other. The user doesn't care what tool produced the warning if the fix is the same. We want the codes to be same so that suppressions work the same way (if I suppress the warning in the analyzer, I really don't want yet another warning for the same problem from illink).

If we don't expect this functionality to ever converge with illink, we should have chosen a different prefix.

agocke:
> If we don't expect this functionality to ever converge with illink, we should have chosen a different prefix.

I'm not sure exactly what tool will perform the analysis, but I do assume that the linker or something else will produce single-file analysis as well.

tlakollo:
Closing via #1790

PR #1790, opened 2021-01-29
RequiresAssemblyFilesAttribute resource manager changes

tlakollo:
PR for #1786
-Change resource file to have a default message and accept arguments
-Add test for RequiresAssemblyFilesAttribute with Url named argument only

### DIFF
--- a/src/ILLink.RoslynAnalyzer/Resources.resx
+++ b/src/ILLink.RoslynAnalyzer/Resources.resx
@@ -117,10 +117,10 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="RequiresUnreferencedCodeAnalyzerTitle" xml:space="preserve">
-    <value>RequiresUnreferencedCodeAnalyzer</value>
+  <data name="RequiresUnreferencedCodeTitle" xml:space="preserve">
+    <value>Avoid calling members marked with `RequiresUnreferencedCodeAttribute` when trimming application code</value>
   </data>
-  <data name="RequiresUnreferencedCodeAnalyzerMessage" xml:space="preserve">
+  <data name="RequiresUnreferencedCodeMessage" xml:space="preserve">
     <value>Calling '{0}' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. {1}. {2}</value>
   </data>
   <data name="AvoidAssemblyLocationInSingleFileTitle" xml:space="preserve">
@@ -136,9 +136,9 @@
     <value>'{0}' will throw for assemblies embedded in a single-file app</value>
   </data>
   <data name="RequiresAssemblyFilesTitle" xml:space="preserve">
-    <value>RequiresAssemblyFilesAnalyzer</value>
+    <value>Avoid calling members marked with `RequiresAssemblyFilesAttribute` when publishing as a single-file</value>
   </data>
   <data name="RequiresAssemblyFilesMessage" xml:space="preserve">
-    <value>Calling '{0}' which has `RequiresAssemblyFilesAttribute` can break functionality when bundling the application code. {1}. {2}</value>
+    <value>Calling '{0}' which has `RequiresAssemblyFilesAttribute` can break functionality when embedded in a single-file app. {1}. {2}</value>
   </data>
 </root>
--- a/src/ILLink.RoslynAnalyzer/Resources.resx
+++ b/src/ILLink.RoslynAnalyzer/Resources.resx
@@ -135,10 +135,10 @@
   <data name="AvoidAssemblyGetFilesInSingleFileMessage" xml:space="preserve">
     <value>'{0}' will throw for assemblies embedded in a single-file app</value>
   </data>
-  <data name="RequiresAssemblyFilesMessage" xml:space="preserve">
-    <value />
-  </data>
   <data name="RequiresAssemblyFilesTitle" xml:space="preserve">
-    <value />
+    <value>RequiresAssemblyFilesAnalyzer</value>
+  </data>
+  <data name="RequiresAssemblyFilesMessage" xml:space="preserve">
+    <value>Calling '{0}' which has `RequiresAssemblyFilesAttribute` can break functionality when bundling the application code. {1}. {2}</value>
   </data>
 </root>
--- a/src/ILLink.RoslynAnalyzer/Resources.resx
+++ b/src/ILLink.RoslynAnalyzer/Resources.resx
@@ -118,9 +118,9 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="RequiresUnreferencedCodeTitle" xml:space="preserve">
-    <value>Methods marked with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code</value>
+    <value>Methods annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code</value>
   </data>
-  <data name="RequiresUnreferencedCodeAnalyzerMessage" xml:space="preserve">
+  <data name="RequiresUnreferencedCodeMessage" xml:space="preserve">
     <value>Using method '{0}' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. {1}. {2}</value>
   </data>
   <data name="AvoidAssemblyLocationInSingleFileTitle" xml:space="preserve">
--- a/src/ILLink.RoslynAnalyzer/Resources.resx
+++ b/src/ILLink.RoslynAnalyzer/Resources.resx
@@ -118,10 +118,10 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="RequiresUnreferencedCodeTitle" xml:space="preserve">
-    <value>Avoid calling members marked with `RequiresUnreferencedCodeAttribute` when trimming application code</value>
+    <value>Methods marked with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code</value>
   </data>
   <data name="RequiresUnreferencedCodeMessage" xml:space="preserve">
-    <value>Calling '{0}' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. {1}. {2}</value>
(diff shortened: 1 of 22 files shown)


src/ILLink.RoslynAnalyzer/Resources.resx
```diff
@@ -135,10 +135,10 @@
...
   </data>
-  <data name="RequiresAssemblyFilesMessage" xml:space="preserve">
-    <value />
-  </data>
   <data name="RequiresAssemblyFilesTitle" xml:space="preserve">
-    <value />
+    <value>RequiresAssemblyFilesAnalyzer</value>
+  </data>
+  <data name="RequiresAssemblyFilesMessage" xml:space="preserve">
+    <value>Calling '{0}' which has `RequiresAssemblyFilesAttribute` can break functionality when bundling the application code. {1}. {2}</value>
```
agocke, line 142: I'm not sure `bundling the application code` is a well-understood term. The above terminology uses "embedded in a single-file app," which seems understandable for things like Xamarin or Blazor.

src/ILLink.RoslynAnalyzer/Resources.resx
```diff
@@ -135,10 +135,10 @@
   <data name="AvoidAssemblyGetFilesInSingleFileMessage" xml:space="preserve">
     <value>'{0}' will throw for assemblies embedded in a single-file app</value>
   </data>
-  <data name="RequiresAssemblyFilesMessage" xml:space="preserve">
-    <value />
-  </data>
   <data name="RequiresAssemblyFilesTitle" xml:space="preserve">
-    <value />
+    <value>RequiresAssemblyFilesAnalyzer</value>
```
agocke, line 139: I don't think this is supposed to be the title of the analyzer, it's the title of the warning, which is like a short description of what the warning is supposed to prevent.

Something like, "Avoid calling members marked with `RequiresAssemblyFilesAttribute` when publishing as a single file"?

test/ILLink.RoslynAnalyzer.Tests/RequiresAssemblyFilesAnalyzerTests.cs
```diff
@@ -48,7 +48,7 @@ void M()
 	}
 }";
 			return VerifyRequiresAssemblyFilesAnalyzer (TestRequiresAssemblyFieldsOnEvent,
```
agocke, line 50: These tests should print out a comment with the message when the test fails. If you include the comment, you can see if it looks right in the message and make the code easier to review.

tlakollo: Are we sure we want the comments? they got deleted from the RequiresUnreferencedCodeAnalyzerTests too. I'm fine with or without them just want to homologate things between the analyzers code.

src/ILLink.RoslynAnalyzer/Resources.resx
```diff
@@ -135,10 +135,10 @@
   <data name="AvoidAssemblyGetFilesInSingleFileMessage" xml:space="preserve">
     <value>'{0}' will throw for assemblies embedded in a single-file app</value>
   </data>
-  <data name="RequiresAssemblyFilesMessage" xml:space="preserve">
-    <value />
-  </data>
   <data name="RequiresAssemblyFilesTitle" xml:space="preserve">
-    <value />
+    <value>Avoid calling members marked with `RequiresAssemblyFilesAttribute` when publishing as a single-file</value>
```
mateoatr, line 139: Nit: can we change (here and above) the backticks for common apostrophe (') similar to

https://github.com/mono/linker/blob/3c3771c5f0818f9b88387bd2bc1d9e58712e6b31/src/ILLink.RoslynAnalyzer/Resources.resx#L124

Or https://github.com/mono/linker/blob/master/src/linker/Resources/Strings.resx ?

src/ILLink.RoslynAnalyzer/Resources.resx
```diff
@@ -117,10 +117,10 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="RequiresUnreferencedCodeAnalyzerTitle" xml:space="preserve">
-    <value>RequiresUnreferencedCodeAnalyzer</value>
+  <data name="RequiresUnreferencedCodeTitle" xml:space="preserve">
+    <value>Avoid calling members marked with `RequiresUnreferencedCodeAttribute` when trimming application code</value>
```
mateoatr, line 121:
```suggestion
    <value>Avoid calling methods marked with 'RequiresUnreferencedCodeAttribute' when trimming application code</value>
```

Also I'm not sure that something marked with this attribute is something that the user should avoid calling, it's just a method that requires dynamic access, which is dangerous when linking the application (but I really don't have a good suggestion for framing this title).

src/ILLink.RoslynAnalyzer/Resources.resx
```diff
@@ -117,10 +117,10 @@
...
   </data>
-  <data name="RequiresUnreferencedCodeAnalyzerMessage" xml:space="preserve">
+  <data name="RequiresUnreferencedCodeMessage" xml:space="preserve">
     <value>Calling '{0}' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. {1}. {2}</value>
```
mateoatr, line 124: Since you're editing things here, I think we should update this message to match the one we produce from within the linker: https://github.com/mono/linker/pull/1778#discussion_r564429652

src/ILLink.RoslynAnalyzer/Resources.resx
marek-safar: Is calling the right terminology here?. Are we going to add a new message for each bundle like format?

tlakollo: > Is calling the right terminology here?

I'm changing the message to match something similar to https://github.com/mono/linker/pull/1778#discussion_r564429652
> Are we going to add a new message for each bundle like format?

Since they all will use the same custom attribute I would expect to print the same message independent of the kind of bundle

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -62,19 +62,19 @@ internal static void RunTest (MethodDeclarationSyntax m, List<AttributeSyntax> a
...
 				switch (attr.Name.ToString ()) {
 				case "ExpectedWarning":
-					var expectedMessageContains = attr.ArgumentList!.Arguments.Select (m => GetStringFromExpr (m.Expression)).ToList ();
-					if (!expectedMessageContains[0].StartsWith ("IL"))
+					var expectedMessagesContains = attr.ArgumentList!.Arguments.Select (m => GetStringsFromExpr (m.Expression)).ToList ();
+					var expectedWarningCode = expectedMessagesContains[0];
+					if (!expectedWarningCode[0].Contains ("IL"))
 						break;
-					expectedMessageContains.RemoveAt (0);
+					var expectedMessages = expectedMessagesContains[1];
```
agocke, line 69: It looks like this no longer handles `params` arrays properly, namely it only looks at the second argument, instead of all the arguments.

src/ILLink.RoslynAnalyzer/Resources.resx
agocke: We could rephrase by saying something like "Members annotated with `RequiresUnreferencedCodeAttribute` may behave differently or fail if the unreferenced code is removed by trimming"

I think that's a decent description of the problem, but it doesn't really give a call to action, like "don't call this when trimming." I don't have a strong opinion either way, but I think we should decide on what kind of title we want to present.

The context for this resource is that when you're looking at a list of the warnings, this will be the description of what the warning is about. The message is what's actually printed when the warning fires.

tlakollo: In a previous commit, I changed the message to:
Methods marked with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -56,7 +56,7 @@ public static void Main ()
 			TestRequiresInDynamicDependency ();
 		}
 
-		[ExpectedWarning ("IL2026", "--RequiresWithMessageOnly--")]
+		[ExpectedWarning ("IL2026", new string[] { "'Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.RequiresWithMessageOnly()'", "Message for --RequiresWithMessageOnly--." })]
```
agocke, line 59: Why add `new string[]`?

agocke (approved):
LGTM, although I'd prefer if we settle on one style for params array invocations in `ExpectedWarning`: either use `new string[]` everywhere, or don't use it anywhere.

mateoatr (changes requested):
Can we remove the `new string[]` usage in the tests? Why do we need it? Looks to me that it is enough to just look for the warning code and look for the message that was specified in the `Message` parameter of the method that we think is the origin of the warning.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
tlakollo: Ok, I changed the code to not support array creation, the code will throw in GetStringFromExpr if it ever tries to be used

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -108,7 +108,7 @@ static void TestRequiresOnPropertyGetterAndSetter ()
 			set { }
 		}
 
-		[ExpectedWarning ("IL2026", new string[] { "'Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.RequiresAndCallsOtherRequiresMethods<TPublicMethods>()'", "Message for --RequiresAndCallsOtherRequiresMethods--." })]
+		[ExpectedWarning ("IL2026", "'Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.RequiresAndCallsOtherRequiresMethods<TPublicMethods>()'", "Message for --RequiresAndCallsOtherRequiresMethods--.")]
```
mateoatr, line 111: It's enough to only validate the message that's coming from the annotated method. We explicitly removed full text validation previously: https://github.com/mono/linker/pull/1778#discussion_r564432966

tlakollo: My line of thinking was that this is not full-text validation, I only validate the method name and the message which are things that should not change in the message. The linker has code in the result checker to check for the method with the ExpectedWarning attribute. The analyzer on the other hand will match any messages that are within the span, which means that there is a chance for the analyzer to check for messages that were not produced by the method marked with ExpectedWarning. Although in this file, no message is repeated, so using the message is enough. I will change the implementation.

PR #1798, opened 2021-02-03
Fix overreporting of warnings due to RequiresUnreferencedCode

vitek-karas:
Linker has intrinsic handling for some methods which are annotated with `RequiresUnreferencedCodeAttribute`. The intrinsic handling will sometimes recognize the pattern and avoid generating a warning, at other times it will generate a better more specific warning. In cases where this happens the generic handler for `RequiresUnreferencedCodeAttribute` should not kick in, as it would still produce a warning.

The change:
* For direct calls, virtual calls and `.ctor` calls rely on `ReflectionMethodBodyScanner` to handle all warnings caused by `RequiresUnreferencedCodeAttribute`. It has knowledge about intrinsic handling and can handle all non-intrisic calls as well.
* Limit the generic handler in `MarkMethod` to all other cases (typically reflection or other indirect references to the method)

Test changes:
* We already have test cases which trigger these conditions, but we don't have a way to validate that extra warnings were not produced
* Added `ExpectedNoWarnings` test attribute which validates no warnings other than those explicitly stated by the test are produced.
* Used it on two test cases which trigger the interesting condition - and fixed the tests to correctly expect all warnings


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2497,33 +2497,41 @@ protected virtual MethodDefinition MarkMethod (MethodReference reference, Depend
...
 			case DependencyKind.TriggersCctorForCalledMethod:
-			case DependencyKind.VirtualCall:
 				break;
 
+			// DirectCall, VirtualCall and NewObj are handled by ReflectionMethodBodyScanner
+			// This is necessary since the ReflectionMethodBodyScanner has intrinsic handling for some
+			// of the methods annotated with the attribute (for example Type.GetType)
+			// and it know when it's OK and when it needs a warning. In this place we don't know
+			// and would have to warn every time.
+
```
marek-safar, line 2517: Is it worth adding assert for that?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2497,33 +2497,41 @@ protected virtual MethodDefinition MarkMethod (MethodReference reference, Depend
...
+		{
 			// If the caller of a method is already marked with `RequiresUnreferencedCodeAttribute` a new warning should not
 			// be produced for the callee.
 			if (origin.MemberDefinition != null &&
 				Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (origin.MemberDefinition))
 				return;
 
 			if (Annotations.TryGetLinkerAttribute (method, out RequiresUnreferencedCodeAttribute requiresUnreferencedCode)) {
-				string message = $"'{method.GetDisplayName ()}' method has 'RequiresUnreferencedCodeAttribute' which can break functionality when trimming application code.";
+				string message = $"Using method '{method.GetDisplayName ()}' which has 'RequiresUnreferencedCodeAttribute' which can break functionality when trimming application code.";
```
marek-safar, line 2534: Could you update https://github.com/mono/linker/blob/master/docs/error-codes.md

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2449,9 +2449,9 @@ void MarkMethodCollection (IList<MethodDefinition> methods, in DependencyInfo re
 				MarkMethod (method, reason, new MessageOrigin (sourceLocationMember));
 		}
 
-		protected internal void MarkIndirectlyCalledMethod (MethodDefinition method, in DependencyInfo reason, IMemberDefinition sourceLocationMember)
+		protected internal void MarkIndirectlyCalledMethod (MethodDefinition method, in DependencyInfo reason, MessageOrigin origin)
```
marek-safar, line 2452: Add `in` to `MessageOrigin origin`

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2497,33 +2497,41 @@ protected virtual MethodDefinition MarkMethod (MethodReference reference, Depend
 			return method;
 		}
 
-		protected internal void ProcessRequiresUnreferencedCode (MethodDefinition method, MessageOrigin origin, DependencyKind dependencyKind)
+		void ProcessRequiresUnreferencedCode (MethodDefinition method, MessageOrigin origin, DependencyKind dependencyKind)
```
marek-safar, line 2500: Add `in` to `MessageOrigin origin`

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2497,33 +2497,41 @@ protected virtual MethodDefinition MarkMethod (MethodReference reference, Depend
...
+			// and would have to warn every time.
+
 			default:
 				return;
 			}
 
+			CheckAndReportRequiresUnreferencedCode (method, origin);
+		}
+
+		internal void CheckAndReportRequiresUnreferencedCode (MethodDefinition method, MessageOrigin origin)
```
marek-safar, line 2525: Add `in` to `MessageOrigin origin`

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: It's very hard to "verify" this - the code will get here for DirectCall (and the others), actually before it gets to `ReflectionMethodBodyScanner`. The only sensible check would be in `ProcessMethodBody` that if we see a method call to annotated method, we enable `ReflectionMethodBodyScanner` - but that is already there, since that's the place which enables the data flow processing.

This code here runs during `MarkMethod`, the data flow analysis runs during `ProcessMethod`, so much later on.

I do agree that this relationship is a bit fragile, but I can't think of a better way to handle it right now.

marek-safar (approved):

mateoatr (approved):
LGTM

PR #1799, opened 2021-02-03
Reenable default UnreachableBodies optimization

marek-safar:
Whole runtime repo now passes with it and I could not repro xamarin issue
locally. If this will turn out problem again it can now temporary disabled
on per assembly basis.

### DIFF
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -235,13 +235,13 @@ namespace Mono.Linker
 			WarnAsError = new Dictionary<int, bool> ();
 			WarnVersion = WarnVersion.Latest;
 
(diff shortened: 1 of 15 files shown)


vitek-karas:
Just curious - do we have any data on how many methods are affected when running linker in dotnet/runtime (on libraries)?
Once we build the app, I'm not worried, but we would run this on assemblies shipped as official framework for everybody, that's where I'm a bit concerned.

vitek-karas (approved):

marek-safar:
I ran it on one of the runtime pack build and only 3 assemblies were changed

- System.Data.Common
- System.Security.Cryptography.X509Certificates.dll
- System.Transactions.Local.dll

PR #1811, opened 2021-02-08
Extend root mode options with library mode

### LINKED ISSUE #1610: Better handling of a library trimming

marek-safar:
Fixes #1610

### DIFF
--- a/docs/illink-options.md
+++ b/docs/illink-options.md
@@ -24,7 +24,8 @@ which can use one of following values.
 - `all` - Keep all members in root assembly
 - `default` - Use entry point for applications and all members for libraries
 - `entrypoint` - Use assembly entry point as only root in the assembly
-- `visible` - Keep all members and types visible outside of root assembly
+- `library` - All assembly members and data needed for secondary trimming are retained
+- `visible` - Keep all members and types visible outside of the assembly
 
 You can retain all public members of `Program.exe` application even if they are not
 referenced by any dependency by calling linker like
--- a/docs/illink-options.md
+++ b/docs/illink-options.md
@@ -21,11 +21,14 @@ When a library is used instead of executable linker will root and mark all membe
 assembly entry point. This rooting behaviour can be customized by passing additional option
 which can use one of following values.
 
-- `all` - Keep all members in root assembly
+- `all` - Keep all members in root assembly. This is equivalent to using `copy` link action
+for the assembly.
 - `default` - Use entry point for applications and all members for libraries
-- `entrypoint` - Use assembly entry point as only root in the assembly
-- `library` - All assembly members and data needed for secondary trimming are retained
-- `visible` - Keep all members and types visible outside of the assembly
+- `entrypoint` - Use assembly entry point as the only root in the assembly. This option is useful
+for multi entry-point libraries bundles.
+- `library` - All visible members and metadata are retained. This useful mode for trimming a library before publishing.
+- `visible` - Keep all members and types visible outside of the assembly. All internals members
+are also rooted when assembly contains InternalsVisibleToAttribute.
 
 You can retain all public members of `Program.exe` application even if they are not
 referenced by any dependency by calling linker like
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -802,23 +802,14 @@ namespace Mono.Linker.Steps
 			if (isPreserveDependency)
 				MarkUserDependency (member, ca, sourceLocationMember);
 
-			if (RetainDynamicDependencyAttribute (member)) {
-				MarkCustomAttribute (ca, reason, sourceLocationMember);
-			} else {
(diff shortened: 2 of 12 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3171,10 +3188,13 @@ protected virtual void MarkInstruction (Instruction instruction, MethodDefinitio
...
 					break;
 				case Code.Isinst:
+					if (operand is TypeSpecification || operand is GenericParameter)
+						break;
+
 					if (Annotations.GetAction (method.DeclaringType.Module.Assembly) != AssemblyAction.Link)
 						break;
 
-					if (operand is TypeSpecification || operand is GenericParameter)
+					if (!_context.IsOptimizationEnabled (CodeOptimizations.UnusedTypeChecks, method.DeclaringType))
```
vitek-karas, line 3197: These two statements both need the assembly of the method - extract to local variable and reuse.

src/linker/Linker.Steps/RootAssemblyInputStep.cs
```diff
@@ -55,15 +55,27 @@ protected override void Process ()
...
+				var preserve_library = TypePreserveMembers.Visible | TypePreserveMembers.Library;
+				if (HasInternalsVisibleTo (assembly))
+					preserve_library |= TypePreserveMembers.Internal;
+
+				MarkAndPreserve (assembly, preserve_library);
 
-				foreach (var type in module.Types)
-					MarkAndPreserveVisible (type, preserve);
+				// Assembly root mode wins over any enabled optimization which
+				// could conflict with library rooting behaviour
```
vitek-karas, line 73: Future refactoring idea: If assembly action is not Link we also effectively disable lot of optimizations, but it's done in each usage site explicitly (check for action and then check for the optimization). Maybe it would be cleaner to disable optimization based on actions as well (not just root mode).

src/linker/Linker.Steps/RootAssemblyInputStep.cs
```diff
@@ -85,14 +97,24 @@ AssemblyDefinition LoadAssemblyFile ()
...
+				MarkAndPreserve (type, preserve);
+		}
+
+		void MarkAndPreserve (TypeDefinition type, TypePreserveMembers preserve)
+		{
+			if ((preserve & TypePreserveMembers.Visible) != 0 && !IsTypeVisible (type))
+				preserve &= ~TypePreserveMembers.Visible;
+
+			if ((preserve & TypePreserveMembers.Internal) != 0 && !IsTypeVisibleOrInternal (type))
+				preserve &= ~TypePreserveMembers.Internal;
```
vitek-karas, line 117: `IsTypeVisibleOrInternal` doesn't feel right here. The current check is basically anything but private, but to match the new semantics of the enum, it should be "internal".

vitek-karas (approved):
Except for the "internal type" thing, looks good.

src/linker/Linker.Steps/RootAssemblyInputStep.cs
sbomer: Agreed - we are increasingly blurring the line between actions and root modes and it seems worthwhile to consider unifying them or separating out the overlapping behaviors.

docs/illink-options.md
```diff
@@ -24,7 +24,8 @@ which can use one of following values.
 - `all` - Keep all members in root assembly
 - `default` - Use entry point for applications and all members for libraries
 - `entrypoint` - Use assembly entry point as only root in the assembly
-- `visible` - Keep all members and types visible outside of root assembly
+- `library` - All assembly members and data needed for secondary trimming are retained
```
sbomer, line 27: Before reading the code it was pretty unclear to me what the library mode would do. I think we should add more detailed documentation for it.

sbomer (approved):

src/linker/Linker.Steps/MarkStep.cs
marek-safar: Fixed

src/linker/Linker.Steps/RootAssemblyInputStep.cs
marek-safar: Resolved

docs/illink-options.md
marek-safar: Updated the wording to be more descriptive.

Issue #1703, opened 2020-12-16
Add IL verifier to illink tests


marek-safar:
We are running with no IL and metadata verification for linker tests on netcore builds. We have only [support for perify](https://github.com/mono/linker/blob/master/test/Mono.Linker.Tests/TestCasesRunner/PeVerifier.cs) which is .NET framework tool and we need to port the support to use .NET Core verifier.

This is increasingly important as we run more tests on netcore configuration only and our tests don't have another way to verify the output metatada and IL are correct.

The tool we could use is https://github.com/dotnet/runtime/tree/10381a2cdc33860f0dc649b24cbb703d23b9ea33/src/coreclr/tools/ILVerify

vitek-karas:
This was implemented in https://github.com/dotnet/linker/pull/2960

PR #1812, opened 2021-02-08
Fix scopes not updated on SecurityAttributes

mrvoorhe:
This fixes a bug in the linker where it would not update the scope of TypeReferences on ISecurityAttributeProviders

This would result in references to `netstandard` surviving linking in situations where netstandard was linked away and SecurityAttribute's survived.  For us this would happen when facades were linked away and an assembly had the `AssemblyAction.Copy`.  There may have been other scenarios where it was possible to hit this.

I did not write a test for this.  It would be hard to write one.  I don't think it's worth the effort.

### DIFF
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -574,6 +574,7 @@ namespace Mono.Linker.Steps
 				updated = new HashSet<TypeReference> ();
 
 				UpdateCustomAttributesTypesScopes (assembly);
+				UpdateSecurityAttributesTypesScopes (assembly);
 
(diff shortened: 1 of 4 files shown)


marek-safar:
@mrvoorhe could you try to write a test? It could be with .il only

marek-safar:
I think such a test would still be valuable. The verifier will return in the form of #1703

mrvoorhe:
@marek-safar I added a test.  Note that it does assume https://github.com/mono/linker/issues/1703 will be addressed so that the test actually means something in the future.  Currently the test will pass with or without my fix.

marek-safar (approved):

PR #1813, opened 2021-02-08
Added method name to Linked Away

amieres:
Small change to identify the method that got linked away.

### DIFF
--- a/src/linker/Linker.Steps/CodeRewriterStep.cs
+++ b/src/linker/Linker.Steps/CodeRewriterStep.cs
@@ -137,7 +137,7 @@ namespace Mono.Linker.Steps
 			ctor = Context.MarkedKnownMembers.NotSupportedExceptionCtorString;
 			ctor = assembly.MainModule.ImportReference (ctor);
 
-			il.Emit (OpCodes.Ldstr, "Linked away");
+			il.Emit (OpCodes.Ldstr, "Linked away: " + method.FullName);
 			il.Emit (OpCodes.Newobj, ctor);
 			il.Emit (OpCodes.Throw);
 


MichalStrehovsky:
Out of curiosity - in what scenario is this necessary? The method in question is on the stack when the exception is thrown so if you have this under a debugger, read `exception.StackTrace`, etc. you can extract the method with the missing body pretty easily.

The extra strings in the message are going to have negative size impact.

amieres:
In my experience AOT & AOTInterp do not show a stack trace. Also, sometimes the code catches the exception only to throw it later which loses the stacktrace and makes it that much harder to know where it happened.
I suppose an alternative is to show it only in debug mode.

marek-safar:
I agree that this change compromises the size saving effect and could even inflate the assembly size when the body is smaller than the string.

AOT is "publishing" format so I'd suggest you run the same code in "build/debug" mode to get a  correct exception or enable linker analyzer to tell you which code will be trimmed unexpectedly.

amieres:
My usage scenario is in WASM, where there is no way to run debugger with AOT or AOTInterp at the moment.
So all you get is an exception that says: "Linked Away", no stack trace.
Without a stack trace it is not easy to know what method was the cause of the problem, specially if it happens in a third party library that I cannot control or modify for testing.

What is the linker analyzer? is that an option in monolinker.exe?

marek-safar:
Linker Away is/should not be codegen specific. You should be able to get the same exception without AOT codegen when trimming the output the same way.

MichalStrehovsky:
I think having _some sort_ of stack traces in AOT will be a must. If it's not there yet, there will be pressure from users to enable it.

It's possible to make them in WASM - the CoreRT project just calls into emscripten to make them [like this](https://github.com/dotnet/corert/pull/8201/files#diff-aabcfcdd0b77cec28b32ed3d104ba3005f6418394af72a9d068fe749f8c1ada3R22-R23). The Mono AOT pipeline might be stripping the names away to get size down and that makes perfect sense, but the AOT compiler should be able to produce a lookup table from `wasm-function[12314]` to IL names - and there can be an offline tool to convert the function IDs to readable names using a sidecar lookup table (i.e. you won't get nice names in `Exception.ToString()`, but using an extra tool that reads the lookup table you can get the names later).

amieres:
@marek-safar 
That is why I think that the method description could be turned on with a flag or option. Unfortunately I do not know how to make that change in the linker.

@MichalStrehovsky 
I agree, any kind of stack trace would be helpful, right now we get nothing at all.

Let me ask again: is the analyzer an option in the linker or another tool?, how do I use it? Is there documentation somewhere?

Thanks

lewing:
> I think having _some sort_ of stack traces in AOT will be a must. If it's not there yet, there will be pressure from users to enable it.
> 
> It's possible to make them in WASM - the CoreRT project just calls into emscripten to make them [like this](https://github.com/dotnet/corert/pull/8201/files#diff-aabcfcdd0b77cec28b32ed3d104ba3005f6418394af72a9d068fe749f8c1ada3R22-R23). The Mono AOT pipeline might be stripping the names away to get size down and that makes perfect sense, but the AOT compiler should be able to produce a lookup table from `wasm-function[12314]` to IL names - and there can be an offline tool to convert the function IDs to readable names using a sidecar lookup table (i.e. you won't get nice names in `Exception.ToString()`, but using an extra tool that reads the lookup table you can get the names later).

For the moment you can set
<WasmNativeStrip>False</WasmNativeStrip>
when building the app and that should skip stripping the function names.

amieres:
Here is the output I get. This is taken from the Chrome console:

    Error: System.NotSupportedException: Linked away 10:56:21.192 testing.js

As you can see there is no stack trace whatsoever. This is the code that produces the output (F#):

    async {
            try
                let! wsResp = Remoting.wsServer.Force().HandleRequest wsReq  // <--- somewhere inside this fails
                Remoting.returnValue(header, wsResp.Content)
            with e -> 
                Remoting.returnExn  (header, sprintf "%A" e)   // <--- here the exception is printed
    }

It is a simple `try catch` (`try with` in F#.)

With the trace on we a get a little more information, like this:

```
L: AOT: FOUND method System.Exception:ToString () [0x1a48 - 0x1a49 0x7de28]
L: AOT: FOUND method System.Exception:get_Message () [0x1a3a - 0x1a3b 0x7de28]
L: AOT: FOUND method System.RuntimeType:ToString () [0x1b0e - 0x1b0f 0x7de28]
L: AOT: FOUND method Microsoft.FSharp.Core.PrintfImpl/SmallStringPrintfEnv2:Write (string) [0x8aee - 0x8aef 0x175f48]
L: AOT: FOUND method Microsoft.FSharp.Core.PrintfImpl/SmallStringPrintfEnv2:Finish () [0x8aed - 0x8aee 0x175f48]
L: AOT: FOUND method (wrapper delegate-invoke) System.Func`1<WebAssembly.JSObject>:invoke_TResult () [0x3ac - 0x3ad 0x4b90]
L: AOT: FOUND method FsRoot.WsTranslator/Remoting/remoting@382:Invoke (Microsoft.FSharp.Core.Unit) [0x350 - 0x351 0x4b90]
L: AOT: FOUND method System.Reflection.Assembly:GetType (string) [0x40de - 0x40df 0x7de28]
L: AOT: FOUND method System.Reflection.RuntimeAssembly:GetType (string,bool,bool) [0x4187 - 0x4188 0x7de28]
L: AOT: FOUND method (wrapper other) object:gsharedvt_out_sig (intptr,object&,intptr) [0x67f5 - 0x67f6 0x7de28]
L: AOT: FOUND method System.Runtime.InteropServices.GCHandle:op_Explicit (intptr) [0x3a5e - 0x3a5f 0x7de28]
L: AOT: FOUND method (wrapper other) object:gsharedvt_out_sig (intptr,int&,intptr) [0x664f - 0x6650 0x7de28]
L: AOT: FOUND method System.Console:WriteLine (string) [0x1d5d - 0x1d5e 0x7de28]
L: AOT: FOUND method (wrapper synchronized) System.IO.TextWriter/SyncTextWriter:WriteLine (string) [0x617a - 0x617b 0x7de28]
L: AOT: FOUND method System.IO.StreamWriter:WriteLine (string) [0x4915 - 0x4916 0x7de28]
L: AOT: FOUND method Microsoft.FSharp.Control.AsyncPrimitives/unitAsync@577:Invoke (Microsoft.FSharp.Control.AsyncActivation`1<Microsoft.FSharp.Core.Unit>) [0x8522 - 0x8523 0x175f48]
L: AOT: FOUND method Microsoft.FSharp.Control.AsyncPrimitives/Start@894:Invoke (Microsoft.FSharp.Core.Unit) [0x8580 - 0x8581 0x175f48]
L: AOT: FOUND method System.IO.FileStream:Finalize () [0x4cb1 - 0x4cb2 0x7de28]
L: AOT: FOUND method System.Gen2GcCallback:Finalize () [0xf45 - 0xf46 0x7de28]
L: AOT: FOUND method (wrapper delegate-invoke) System.Func`2<object, bool>:invoke_TResult_T (object) [0x2ef9b - 0x2ef9c 0xbfe640]
L: AOT: FOUND method System.Buffers.ArrayPoolEventSource:BufferTrimPoll (int,int) [0x5bbc - 0x5bbd 0x7de28]
L: AOT: FOUND method System.Buffers.ArrayPoolEventSource:BufferTrimmed (int,int,int) [0x5bbb - 0x5bbc 0x7de28]
L: AOT: FOUND method System.GC:GetMemoryInfo (uint&,ulong&,uint&,uintptr&,uintptr&) [0x1a64 - 0x1a65 0x7de28]
L: AOT: FOUND method (wrapper other) object:gsharedvt_out_sig (intptr&,intptr&,intptr&,intptr&,intptr&,intptr) [0x687d - 0x687e 0x7de28]
L: AOT: FOUND method System.ThrowHelper:ThrowInvalidOperationException_InvalidOperation_EnumOpCantHappen () [0x1832 - 0x1833 0x7de28]
L: AOT: FOUND method (wrapper other) object:gsharedvt_out_sig (object&,int&,int&,intptr) [0x61c5 - 0x61c6 0x7de28]
L: AOT: FOUND method (wrapper other) object:gsharedvt_out_sig (object&,intptr) [0x61aa - 0x61ab 0x7de28]
L: [0x1] worker running in domain 0x41939b8 (outstanding requests 3)
L: AOT: FOUND method System.Random:Next (int) [0x12b6 - 0x12b7 0x7de28]
L: AOT: FOUND method System.Random:Sample () [0x12af - 0x12b0 0x7de28]
L: [0x1] worker running in domain 0x41939b8 (outstanding requests 2)
L: [0x1] worker running in domain 0x41939b8 (outstanding requests 1)
L: [0x1] worker running in domain 0x41939b8 (outstanding requests 0)
Error: System.NotSupportedException: Linked away
```
Not as useful as a trace, but better than nothing, although hard to read.

The method that got linked away and threw the exception was this: `System.Void Log(FSharp.Compiler.LogCompilerFunctionId)`

marek-safar:
I understand your perspective but I'm still not sure this is the right fix. We should focus on improving wasm stack-trace handling instead or your experience will be painful anyway.

ILLink warnings won't help in this case right now. Maybe a useful fix would be to print a list of unreachable methods that are rewritten to NotSupported in the verbose mode as that list is always a bit suspicious.

amieres:
What about the switch option? Something like

    monolinker.exe --identify-linked-away-methods

Something shorter of course. It should be easy to do. 

I agree a proper stacktrace would be the ideal solution but I assume that there must be good reasons why is not implemented yet. It probably will take a while to solve. In the mean time giving the developer the option would make it that much easier to pinpoint the ''Linked Away particular problem.

Thanks for your help.

src/linker/Linker.Steps/CodeRewriterStep.cs
```diff
@@ -137,7 +137,7 @@ MethodBody CreateThrowLinkedAwayBody (MethodDefinition method)
 			ctor = Context.MarkedKnownMembers.NotSupportedExceptionCtorString;
 			ctor = assembly.MainModule.ImportReference (ctor);
 
-			il.Emit (OpCodes.Ldstr, "Linked away");
+			il.Emit (OpCodes.Ldstr, "Linked away: " + method.FullName);
```
vargaz, line 140: This should be controlled by an option, since it's going to add a lot of extra strings to the assembly.

amieres: How do I add an option and what should it be called: `indetify-linked-away`?

PR #1822, opened 2021-02-10
Overwrite ILLinkTasksAssembly

ViktorHofer:
If the Microsoft.NET.ILLink.Tasks package is package referenced directly, overwriting `ILLinkTasksAssembly` guarantees that the SDK uses the right ILLink assembly. Github org searches didn't show any consumer of the private properties, hence removing them for simplification.

https://github.com/dotnet/arcade/pull/6936 benefits from this change.

### DIFF
--- a/src/ILLink.Tasks/Sdk/Sdk.props
+++ b/src/ILLink.Tasks/Sdk/Sdk.props
@@ -9,14 +9,11 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
 Copyright (c) .NET Foundation. All rights reserved.
 ***********************************************************************************************
 -->
-<Project ToolsVersion="14.0">
+<Project>
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


marek-safar (approved):

ViktorHofer:
For local builds, `ILLinkTasksAssembly` can still be passed in globally which always takes precedence and overwrites any values set here.

sbomer:
I typically like to set it in the project file rather than the command-line.

ViktorHofer:
That still works as props are imported before the project file which means that the value in the project file overwrites what is in the props file.

sbomer:
The Sdk.props are imported late in https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L14. Should that be changed?

ViktorHofer:
oh forget that. yes, we should definitely fix that as props should be imported before the project is evaluated.

ViktorHofer:
either via a ILLink.props file in the SDK or directly in the NET.sdk.props file

ViktorHofer:
@sbomer should I take care of that or do you want to?

sbomer:
I will take care of it, thanks!

Issue #1539, opened 2020-10-06
Consider disabling automatic probing for assemblies when running linker from SDK


vitek-karas:
Currently linker implements automatic probing for assemblies in specified paths. 
The place where we found it hurting recently is when the application contains C++/CLI assemblies. The ILLink task has code to detect these and remove them from the list passed to the linker - since linker won't be able to handle them correctly. But if there's an assembly reference from a pure-managed assembly to C++/CLI assembly and linker can find it in one of the probing paths, linker will load the C++/CLI assembly and treat it as a normal managed assembly. This leads to various issues:
* C++/CLI assemblies are not always 100% per ECMA spec - for example see https://github.com/mono/linker/issues/1527
* Depending the default action linker may try to roundtrip the assembly through Cecil, which will completely break it (as Cecil doesn't understand the native parts of it)

SDK and linker may also not agree on order in which paths are used for lookups and thus end up with different versions of the same assembly.

When running in SDK, it's the job of SDK to determine the full closure of the application before running linker. So any probing linker does is basically second guessing the decisions already made by SDK. It would be much cleaner if we disabled probing in linker in this case and solely relied on the set of assemblies passed to the linker.

vitek-karas:
See https://github.com/mono/linker/issues/1540 for related discussion on handling of C++/CLI in the ILLink task.

PR #1828, opened 2021-02-12
Disable probing for assemblies by default

sbomer:
We used to probe for assemblies in:
- the working directory
- next to any root assembly
- in directories passed on the command-line

Now we will only probe for directories passed on the command-line, which
aren't used by the SDK.

Some of the testcases have references to mscorlib (either from .il, or from
checked-in dlls), which references System.Security.Permissions. This isn't
included with netcoreapp, but the tests used to resolve it from the
working directory of Mono.Linker.Tests.dll. We pass --skip-unresolved true
to keep these testcases from throwing resolution errors with the change.

Fixes https://github.com/mono/linker/issues/1539

### DIFF
--- a/src/linker/Linker.Steps/RootAssemblyInputStep.cs
+++ b/src/linker/Linker.Steps/RootAssemblyInputStep.cs
@@ -92,8 +92,7 @@ namespace Mono.Linker.Steps
 			if (loaded != null)
 				return loaded;
 
-			Context.Resolver.CacheAssemblyWithPath (assembly);
-			Context.RegisterAssembly (assembly);
+			Context.Resolver.CacheAssembly (assembly);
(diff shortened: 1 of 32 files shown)


marek-safar:
https://github.com/mono/linker/blob/main/test/Mono.Linker.Tests.Cases/Mono.Linker.Tests.Cases.csproj#L65-L67 should not be needed then

vitek-karas (approved):
I like it :wink:

sbomer:
The PackageReference to `System.Security.Permissions` is still needed to compile some of the testcases which use security attributes (and those tests pass without needing to resolve the security types at link time).

marek-safar (approved):

Issue #1724, opened 2021-01-06
Type.GetProperties(Public | Static) should only need DynamicallyAccessedMemberTypes.PublicProperties


eerhardt:
Run the following code through the ILLinker:

```C#
        private static Dictionary<string, Color> GetColors()
        {
            var colors = new Dictionary<string, Color>(StringComparer.OrdinalIgnoreCase);
            FillWithProperties(colors, typeof(Color));
            FillWithProperties(colors, typeof(SystemColors));
            return colors;
        }

        private static void FillWithProperties(
            Dictionary<string, Color> dictionary,
            [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)] Type typeWithColors)
        {
            foreach (PropertyInfo prop in typeWithColors.GetProperties(BindingFlags.Public | BindingFlags.Static))
            {
                if (prop.PropertyType == typeof(Color))
                    dictionary[prop.Name] = (Color)prop.GetValue(null, null)!;
            }
        }
```

### Expected

Shouldn't have any warnings since `typeWithColors` is attributed as `DynamicallyAccessedMemberTypes.PublicProperties`.

### Actual

```
Trim analysis warning IL2070: ConsoleApp2.Program.FillWithProperties(Dictionary<String,Color>,Type): The requirements declared via the 'DynamicallyAccessedMembersAttribute' on the parameter 'typeWithColors' of method 'ConsoleApp2.Program.FillWithProperties(Dictionary<String,Color>,Type)' don't match those on the implicit 'this' parameter of method 'System.Type.GetProperties(BindingFlags)'. The source value must declare at least the same requirements as those declared on the target location it's assigned to
```

cc @vitek-karas @MichalStrehovsky

eerhardt:
Similar scenario in `System.Type.GetMember()` and `System.Type.GetMembers()`:

https://github.com/dotnet/runtime/blob/5282582d79c9cdc1a57c3a58e95485cb329efec7/src/libraries/System.Private.CoreLib/src/System/Type.cs#L211-L220

PR #1829, opened 2021-02-12
New linker intrinsics

tlakollo:
Add new intrinsics for:
- GetConstructors
- GetEvents
- GetFields
- GetProperties
- GetMethods
- GetNestedTypes
- GetMember
- GetMembers
Add tests

This should fix the problem in #1724

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1252,7 +1252,6 @@ namespace Mono.Linker.Dataflow
 
 						var parameters = calledMethod.Parameters;
 						BindingFlags? bindingFlags;
-						//var memberTypes = MemberTypes.All;
 						if (parameters.Count == 1) {
 							// Assume a default value for BindingFlags for methods that don't use BindingFlags as a parameter
 							bindingFlags = BindingFlags.Public | BindingFlags.Instance;
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1244,7 +1244,9 @@ namespace Mono.Linker.Dataflow
 				// GetMember (String, BindingFlags)
 				// GetMember (String, MemberTypes, BindingFlags)
 				//
-				case IntrinsicId.Type_GetMember: {
+				case IntrinsicId.Type_GetMember when calledMethod.DeclaringType.Namespace == "System"
+					&& calledMethod.DeclaringType.Name == "Type"
+					&& calledMethod.HasParameterOfType (0, "System", "String"): {
 						reflectionContext.AnalyzingPattern ();
 
 						var parameters = calledMethod.Parameters;
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -231,11 +231,19 @@ namespace Mono.Linker.Dataflow
 			Type_MakeGenericType,
 			Type_GetType,
 			Type_GetConstructor,
+			Type_GetConstructors,
 			Type_GetMethod,
+			Type_GetMethods,
 			Type_GetField,
+			Type_GetFields,
 			Type_GetProperty,
+			Type_GetProperties,
 			Type_GetEvent,
+			Type_GetEvents,
 			Type_GetNestedType,
+			Type_GetNestedTypes,
+			Type_GetMember,
+			Type_GetMembers,
 			Type_get_AssemblyQualifiedName,
 			Type_get_UnderlyingSystemType,
 			Expression_Call,
@@ -343,6 +351,13 @@ namespace Mono.Linker.Dataflow
 					&& calledMethod.HasThis
 					=> IntrinsicId.Type_GetConstructor,
 
+				// System.Type.GetConstructors (BindingFlags)
+				"GetConstructors" when calledMethod.IsDeclaredOnType ("System", "Type")
+					&& calledMethod.HasParameterOfType (0, "System.Reflection", "BindingFlags")
+					&& calledMethod.Parameters.Count == 1
+					&& calledMethod.HasThis
+					=> IntrinsicId.Type_GetConstructors,
+
 				// System.Type.GetMethod (string)
 				// System.Type.GetMethod (string, BindingFlags)
 				// System.Type.GetMethod (string, Type[])
@@ -358,12 +373,26 @@ namespace Mono.Linker.Dataflow
 					&& calledMethod.HasThis
 					=> IntrinsicId.Type_GetMethod,
 
+				// System.Type.GetMethods (BindingFlags)
+				"GetMethods" when calledMethod.IsDeclaredOnType ("System", "Type")
+					&& calledMethod.HasParameterOfType (0, "System.Reflection", "BindingFlags")
+					&& calledMethod.Parameters.Count == 1
+					&& calledMethod.HasThis
+					=> IntrinsicId.Type_GetMethods,
+
 				// System.Type.GetField (string)
 				// System.Type.GetField (string, BindingFlags)
 				"GetField" when calledMethod.IsDeclaredOnType ("System", "Type")
 					&& calledMethod.HasParameterOfType (0, "System", "String")
 					&& calledMethod.HasThis
 					=> IntrinsicId.Type_GetField,
+
+				// System.Type.GetFields (BindingFlags)
+				"GetFields" when calledMethod.IsDeclaredOnType ("System", "Type")
+					&& calledMethod.HasParameterOfType (0, "System.Reflection", "BindingFlags")
+					&& calledMethod.Parameters.Count == 1
+					&& calledMethod.HasThis
+					=> IntrinsicId.Type_GetFields,
 
 				// System.Type.GetEvent (string)
 				// System.Type.GetEvent (string, BindingFlags)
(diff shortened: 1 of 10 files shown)


test/Mono.Linker.Tests.Cases/Reflection/ConstructorsUsedViaReflection.cs
```diff
@@ -0,0 +1,214 @@
...
+		[RecognizedReflectionAccessPattern]
+		[Kept]
+		static void TestWithUnknownBindingFlags (BindingFlags bindingFlags)
+		{
+			// Since the binding flags are not known linker should mark all constructors on the type
+			var constructors = typeof (UnknownBindingFlags).GetConstructors (bindingFlags);
+		}
+
+		[Kept]
+		static void TestNullType ()
```
vitek-karas, line 47: RecognizedReflectionAccessPattern

test/Mono.Linker.Tests.Cases/Reflection/EventUsedViaReflection.cs
```diff
@@ -144,8 +144,8 @@ static void TestIfElse (int i)
 			typeof (BaseClass), nameof (BaseClass.PublicEventOnBase), (Type[]) null)]
 		static void TestEventInBaseType ()
 		{
-			typeof (DerivedClass).GetEvent ("ProtectedEventOnBase"); // Will not mark anything as it only works on public events
-			typeof (DerivedClass).GetEvent ("PublicEventOnBase");
+			var eventInfo = typeof (DerivedClass).GetEvent ("ProtectedEventOnBase"); // Will not mark anything as it only works on public events
+			eventInfo = typeof (DerivedClass).GetEvent ("PublicEventOnBase");
```
vitek-karas, line 148: What's the purpose of this change (also below) - I would remove it.

test/Mono.Linker.Tests.Cases/Reflection/EventsUsedViaReflection.cs
```diff
@@ -0,0 +1,254 @@
...
+		[Kept]
+		[RecognizedReflectionAccessPattern]
+		static void TestGetEvents ()
+		{
+			var events = typeof (Foo).GetEvents ();
+		}
+
+		[Kept]
+		// The event will not be kept as it's internal and the behavior of Type.GetEvent(string) is to only return public events
+		// But we don't mark it as unrecognized access pattern - we did recognize it fully, just didn't find the event being asked for
```
vitek-karas, line 39:
```suggestion
		// The event will not be kept as it's internal and the behavior of Type.GetEvents() is to only return public events
		// But we don't mark it as unrecognized access pattern - we did recognize it fully, just didn't find the event being asked for
```

test/Mono.Linker.Tests.Cases/Reflection/EventsUsedViaReflection.cs
```diff
@@ -0,0 +1,254 @@
...
+		// The behavior of the code will not change by linking it:
+		//   - Without linking the GetEvent will return null
+		//   - After linking the GetEvent will still return null
+		// We also don't mark it as recognized pattern since we didn't mark anything
```
vitek-karas, line 43:
```suggestion
		//   - Without linking the GetEvents will return null
		//   - After linking the GetEvents will still return null
		// We also don't mark it as recognized pattern since we didn't mark anything
```

test/Mono.Linker.Tests.Cases/Reflection/EventsUsedViaReflection.cs
```diff
@@ -0,0 +1,254 @@
...
+				myType = typeof (IfClass);
+			} else {
+				myType = typeof (ElseClass);
+			}
+			String myString;
+			if (i == 1) {
+				myString = "IfEvent";
+			} else {
+				myString = "ElseEvent";
+			}
```
vitek-karas, line 106: This is not used for anything

test/Mono.Linker.Tests.Cases/Reflection/FieldsUsedViaReflection.cs
```diff
@@ -0,0 +1,187 @@
...
+				myType = typeof (IfClass);
+			} else {
+				myType = typeof (ElseClass);
+			}
+			String myString;
+			if (i == 1) {
+				myString = "ifField";
+			} else {
+				myString = "elseField";
+			}
```
vitek-karas, line 93: Not used for anything

test/Mono.Linker.Tests.Cases/Reflection/MemberUsedViaReflection.cs
```diff
@@ -0,0 +1,487 @@
...
+
+namespace Mono.Linker.Tests.Cases.Reflection
+{
+	[SetupCSharpCompilerToUse ("csc")]
+	public class MemberUsedViaReflection
+	{
+		public static void Main ()
+		{
+			// Normally calls to GetMember use regex to find values, we took a conservative approach
+			// and preserve not based on the string passed but on the binding flags requirements
```
vitek-karas, line 15: I don't get this - regex?
I'm fine if we don't implement the name based lookups (this is not very common), but the comment here is weird.

vitek-karas: I see - well I would not call "C*" a regex (it's actually not a valid regex) - more like "wildcard".
Can you please rename anything with "Regex" in this class - it doesn't matter to what really, but regex is confusing.

test/Mono.Linker.Tests.Cases/Reflection/MemberUsedViaReflection.cs
```diff
@@ -0,0 +1,487 @@
...
+			var members = typeof (UnknownBindingFlags).GetMember ("*FoundViaRegex", bindingFlags);
+		}
+
+		[RecognizedReflectionAccessPattern]
+		[Kept]
+		static void TestWithMemberTypes ()
+		{
+			// Here we took the same conservative approach, instead of understanding MemberTypes we only use
+			// the information in the binding flags requirements and keep all the MemberTypes
+			var members = typeof (TestMemberTypes).GetMember ("*FoundViaRegex", MemberTypes.Method, BindingFlags.Public);
```
vitek-karas, line 63: BTW: "*C" will not work if I read the code in CoreLib correctly - it can only handle prefixes, not suffixes.

test/Mono.Linker.Tests.Cases/Reflection/MembersUsedViaReflection.cs
```diff
@@ -0,0 +1,448 @@
...
+		static void TestWithBindingFlags ()
+		{
+			var members = typeof (MembersBindingFlags).GetMembers (BindingFlags.Public);
+		}
+
+		[RecognizedReflectionAccessPattern]
+		[Kept]
+		static void TestWithUnknownBindingFlags (BindingFlags bindingFlags)
+		{
+			// Since the binding flags are not known linker should mark all constructors on the type
```
vitek-karas, line 42:
```suggestion
			// Since the binding flags are not known linker should mark all members on the type
```

vitek-karas:
Do we have a tracking work item to make corresponding changes to the runtime repo?
* Change `Type.GetMembers` annotation to not use `All` and instead list all of the other bits (the same way linker treats it as an intrinsic)
* Remove suppressions in reflection for the warning this resolves - for example `Type.GetProperties()` has a suppression which should not be necessary after this change. Might be worth going over all potential candidates (the less suppressions the better).
* Remove suppressions in the rest of libraries - @eerhardt is probably going to have an idea where this was used so far.

test/Mono.Linker.Tests.Cases/Reflection/MemberUsedViaReflection.cs
tlakollo: Yup, thanks for the heads up 
https://github.com/dotnet/runtime/blob/0026a202ab4087d237f6be8824dff322d3a7d6be/src/coreclr/System.Private.CoreLib/src/System/RuntimeType.CoreCLR.cs#L2059

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1116,6 +1181,106 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+
+
+				//
+				// GetMember (String)
+				// GetMember (String, BindingFlags)
+				// GetMember (String, MemberTypes, BindingFlags)
+				//
+				case IntrinsicId.Type_GetMember when calledMethod.DeclaringType.Namespace == "System"
+					&& calledMethod.DeclaringType.Name == "Type"
+					&& calledMethod.HasParameterOfType (0, "System", "String"): {
```
vitek-karas, line 1249: This doesn't make sense to me - we check exactly this in the first switch - so no reason to check the namespace, typename and first parameter here again. My comment was about the fact that the code below checks that BindingFlags parameter is in a certain position based on the number of parameters to the method. And if those checks don't succeed we fail. But the switch above which detects the `IntrinsicId` does not perform those checks.

So for example, if there was a method Type.GetMember(String, MemberTypes)  (no binding flags), the first switch would accept it as recognized and detected it as `IntrinsicId.Type_GetMember`, it would get here, and the code below would fail because the binding flags checks would fail - it would end up effectively crashing the linker. That's not good. The check which detects the `IntrinsicId` needs to be very precise, so that the code in here (HandleCall) can't ever fail (well at least for reasons of not recognizing the method).

vitek-karas (approved):
Other than the nit in https://github.com/mono/linker/pull/1829#discussion_r576207947 this looks good.

tlakollo:
The lint command updated the dotnet-format version from version '4.1.131201' to 'version '5.0.211103' which seems to be causing this error, I don't see any problems with the file, and the lint in my workstation is still 4.1.131201 so it won't fix the whitespace errors

marek-safar:
@tlakollo you need to merge with master locale to repro that

Issue #1800, opened 2021-02-03
Simplify copy action behavior


sbomer:
The linker accepts copy ~~and save~~ as actions on the command line, which has the behavior that input assemblies which are part of the reference closure of the roots get fully preserved. This requires loading all references, including unused ones. We should:
- Get rid of copy/~~save~~ *default* actions
  - I am not aware of scenarios where they are needed (it would effectively prevent linking if they were used)
- Unify the "copy"/"root" options
  - "copy" used to affect the output only, but it has changed to mean something similar to "-a Assembly.dll all"
  - The SDK should then change to pass a single option instead of marking `IsTrimmable="false"` assemblies as copy + root
  - ~~Open question: can we also get rid of "save"?~~ 
Edit: I forgot that https://github.com/mono/linker/pull/1725 already removed `save` as a command-line option. 
- Ensure that the unified behavior behaves the same as an attribute opt-in
  - It should preserve assemblies even if they were not part of the reference closure

mrvoorhe:
@sbomer I don't entire follow what this issue is about. Is `AssemblyAction.Copy` going to be removed or have it's behavior changed?  Unity depends on `AssemblyAction.Copy` today.  That's not to say we can't change, `AssemblyAction.Copy` does lead to some quirks, I'd just like to understand if this issue means there are plans to change/remove `AssemblyAction.Copy`.

sbomer:
@mrvoorhe  I think most importantly, we want to change the behavior so that:
- `copy` will automatically root an assembly (today, it only roots assemblies that are part of the reference closure)

Otherwise the same quirks (for example copy assemblies being rewritten due to removed type forwarders) would remain. We might then consider renaming it or merging it with the `-a` command-line option, but the important piece is the behavior change.

Would that be reasonable behavior for unity? For the .NET SDK it doesn't make a big difference because we only pass "copy" in conjunction with "-a" by default, though there could technically be edge cases where users specify "copy" for an unreferenced assembly (which would cause it to be kept after the change).

Next, I am also suggesting we:
- Stop supporting `-u copy` and `-c copy`

These are more options that are unused by the .NET SDK and unnecessarily complicate some cases in the linker.

It would be good to know how Unity uses `copy`.

mrvoorhe:
> copy will automatically root an assembly (today, it only roots assemblies that are part of the reference closure)

This seems OK.

> Stop supporting -u copy and -c copy

This seems OK if the internal support for `AssemblyAction.Copy` is still in place.

[...]

PR #1839, opened 2021-02-17
Support attribute trimming opt-in

sbomer:
This implements the linker side of the attribute opt-in described at https://github.com/mono/linker/blob/main/docs/design/trimmed-assemblies.md#assemblymetadataistrimmable-true.


docs/error-codes.md
```diff
@@ -1510,6 +1514,25 @@ This is technically possible if a custom assembly defines `DynamicDependencyAttr
   </linker>
   ```
 
+#### `IL2102`: Invalid AssemblyMetadata("IsTrimmable", ...) attribute in assembly 'assembly'. Value must be "True"
+
+- AssemblyMetadataAttribute may be used at the assembly level to turn on trimming for the assembly. The only supported value is "True", but the attribute contained an unsupported value.
+
+  ``` C#
+  // IL2102: Invalid AssemblyMetadata("IsTrimmable", "False") attribute in assembl 'fullname'. Value must be "True"
```
marek-safar, line 1522: typo: assembl

docs/illink-options.md
```diff
@@ -58,22 +58,25 @@ The linker can do the following things on all or individual assemblies
...
 
-`illink -p skip System.Windows.Forms`
+`illink --action skip System.Windows.Forms`
 
-Or you can specify what to do for the core assemblies.
+Or you can specify what to do for the trimmed assemblies.
 
-Core assemblies are the assemblies that belong to the base class library,
-like `System.Private.CoreLib.dll`, `System.dll` or `System.Windows.Forms.dll`.
+Trimmed assemblies are the assemblies that have `AssemblyMetadata("IsTrimmable", "True")`.
```
marek-safar, line 71:
```suggestion
Trimmable assembly is any assembly that includes `System.Reflection.AssemblyMetadata("IsTrimmable", "True")` attribute.

File: src/linker/Linker/LinkContext.cs [hunk 3]
```diff
@@ -394,37 +387,50 @@ public void SetAction (AssemblyDefinition assembly, AssemblyAction defaultAction
...
+				if (args[0].Value is not string key || !key.Equals ("IsTrimmable", StringComparison.OrdinalIgnoreCase))
+					continue;
+
+				if (args[1].Value is not string value || !value.Equals ("True", StringComparison.OrdinalIgnoreCase)) {
+					LogWarning ($"Invalid AssemblyMetadata(\"IsTrimmable\", \"{args[1].Value}\") attribute in assembly '{assembly.Name.Name}'. Value must be \"True\"", 2102, assemblyFileName);
+					continue;
+				}
+
+				if (isTrimmable) {
+					LogWarning ($"Duplicate AssemblyMetadata(\"IsTrimmable\", \"True\") attributes in assembly '{assembly.Name.Name}'", 2103, assemblyFileName);
```
marek-safar, line 426: Not sure we should be that pedantic

src/linker/Linker/Driver.cs
```diff
@@ -1086,15 +1086,18 @@ static void Usage ()
...
+			Console.WriteLine ("                              link: Remove any unused IL or metadata and optimizes the assembly");
+			Console.WriteLine ("                              skip: Do not process the assembly");
+			Console.WriteLine ("                              addbypassngen: Add BypassNGenAttribute to unused methods");
+			Console.WriteLine ("                              addbypassngenused: Same as addbypassngen but unused assemblies are removed");
+			// https://github.com/mono/linker/blob/main/docs/design/trimmed-assemblies.md#assemblymetadataistrimmable-false
+			// If we add support for opting out by setting "IsTrimmable" to "False", we should add an option to control behavior for such assemblies.
+			// Console.WriteLine ("  --no-trim-action ACTION   Sets action for assemblies annotated as non-trimmable. Defaults to 'copy'");
+			Console.WriteLine ("  --default-action ACTION   Sets action for assemblies that have no trimmability annotation. Defaults to 'link'");
+			Console.WriteLine ("  --action ACTION ASM       Overrides the default action for specific assembly name");
```
marek-safar, line 1100: Do we need two options? What about simpler

```
-action ACTION       Sets action for all assemblies that have no IsTrimmable attribute. Defaults to 'link'
-action ACTION ASM   Overrides the default action for the specific assembly name
```

src/linker/Linker/Driver.cs
```diff
@@ -1086,15 +1086,18 @@ static void Usage ()
...
-			Console.WriteLine ("  -p ACTION ASM       Overrides the default action for specific assembly name");
+			Console.WriteLine ("  --trim-action ACTION      Sets action for assemblies annotated as trimmable. Defaults to 'link'");
+			Console.WriteLine ("                              copy: Analyze whole assembly and save it to the output");
+			Console.WriteLine ("                              copyused: Same as copy but only for assemblies which are needed");
+			Console.WriteLine ("                              link: Remove any unused IL or metadata and optimizes the assembly");
+			Console.WriteLine ("                              skip: Do not process the assembly");
+			Console.WriteLine ("                              addbypassngen: Add BypassNGenAttribute to unused methods");
+			Console.WriteLine ("                              addbypassngenused: Same as addbypassngen but unused assemblies are removed");
+			// https://github.com/mono/linker/blob/main/docs/design/trimmed-assemblies.md#assemblymetadataistrimmable-false
+			// If we add support for opting out by setting "IsTrimmable" to "False", we should add an option to control behavior for such assemblies.
```
marek-safar, line 1097: I don't think this comment is useful

src/linker/Linker/Driver.cs
```diff
@@ -1086,15 +1086,18 @@ static void Usage ()
...
-			Console.WriteLine ("  -c ACTION           Sets action for all framework assemblies. Defaults to 'link'");
-			Console.WriteLine ("                        copy: Analyze whole assembly and save it to the output");
-			Console.WriteLine ("                        copyused: Same as copy but only for assemblies which are needed");
-			Console.WriteLine ("                        link: Remove any unused IL or metadata and optimizes the assembly");
-			Console.WriteLine ("                        skip: Do not process the assembly");
-			Console.WriteLine ("                        addbypassngen: Add BypassNGenAttribute to unused methods");
-			Console.WriteLine ("                        addbypassngenused: Same as addbypassngen but unused assemblies are removed");
-			Console.WriteLine ("  -u ACTION           Sets action for any user assembly. Defaults to 'link'");
-			Console.WriteLine ("  -p ACTION ASM       Overrides the default action for specific assembly name");
+			Console.WriteLine ("  --trim-action ACTION      Sets action for assemblies annotated as trimmable. Defaults to 'link'");
```
marek-safar, line 1089: I think keeping the name aligned with sdk would be better. What about

```
-trim-mode ACTION      Sets action for assemblies annotated with IsTrimmable attribute. Defaults to 'link'
```

src/linker/Linker.Steps/RootNonTrimmableAssemblies.cs
```diff
@@ -0,0 +1,55 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.IO;
+
+namespace Mono.Linker.Steps
+{
+	public class RootNonTrimmableAssemblies : BaseStep
```
marek-safar, line 10: I don't understand why this step is needed?

src/linker/Linker/AssemblyResolver.cs
```diff
@@ -80,16 +80,19 @@ public string GetAssemblyFileName (AssemblyDefinition assembly)
 			return assembly.MainModule.FileName;
 		}
 
-		AssemblyDefinition ResolveFromReferences (AssemblyNameReference name, Collection<string> references, ReaderParameters parameters)
+		AssemblyDefinition ResolveFromReferences (AssemblyNameReference name, HashSet<string> references, ReaderParameters parameters)
 		{
-			var fileName = name.Name + ".dll";
+			var extensions = new[] { ".dll", ".exe" };
```
marek-safar, line 85: should be static

src/linker/Linker/LinkContext.cs
```diff
@@ -394,37 +387,50 @@ public void SetAction (AssemblyDefinition assembly, AssemblyAction defaultAction
...
-			case "netstandard":
-				return true;
-			default:
-				return name.Name.StartsWith ("System")
-					|| name.Name.StartsWith ("Microsoft");
+			if (!assembly.HasCustomAttributes)
+				return false;
+
+			bool isTrimmable = false;
+			var assemblyFileName = Resolver.GetAssemblyFileName (assembly);
```
marek-safar, line 407: This is needed for warnings only

src/linker/Linker/LinkContext.cs
```diff
@@ -394,37 +387,50 @@ public void SetAction (AssemblyDefinition assembly, AssemblyAction defaultAction
...
-				return CoreAction;
+			if (CheckIsTrimmable (assembly))
+				return TrimAction;
 
-			return UserAction;
+			return DefaultAction;
 		}
 
-		public static bool IsCore (AssemblyNameReference name)
+		bool CheckIsTrimmable (AssemblyDefinition assembly)
```
marek-safar, line 401: nit: Check prefix is presumably redundant

marek-safar:
@sbomer it'd be useful to add the attribute to the runtime build before this change goes in to test it

src/linker/Linker/LinkContext.cs
sbomer: I removed the warning about duplicate attributes

src/linker/Linker/Driver.cs
sbomer: Removed

src/linker/Linker.Steps/RootNonTrimmableAssemblies.cs
sbomer: It's needed for example if someone passes `-reference UnreferencedAssembly --action copy UnreferencedAssembly` - then we need to make sure the unreferenced assembly is kept. This could happen if someone has an unused nuget package dependency (or one that's used via non-understood reflection).

src/linker/Linker/Driver.cs
sbomer: I took your naming suggestions. It's a little odd because `--action ACTION ASM` will override not just the `--action ACTION` for unattributed assemblies, but also the `--trim-mode ACTION` for ones with IsTrimmable. In other words, this might give the impression that there's only one "default" action, when really there are two - the `--trim-mode` and the `--action`. That's why I initially picked the names `--trim-action` and `--default-action`.

However I do like the consistency of your suggestion. Either way we will need to think carefully about the names on the SDK side.

sbomer:
Currently I'm running the tests by passing `--action skip` (or whatever action is specified by `SetupLinkerTrimMode` in the test) for the framework assemblies, which simulates the attribute. (Skip to avoid copying and analyzing assemblies that are irrelevant to most testcases).

src/linker/Linker.Steps/RootNonTrimmableAssemblies.cs
marek-safar: Why would we want to do that? It would keep unused assembly which someone redundantly passed as an argument. We have `-a` option which does the same thing and should be used if it's intentional.

sbomer: Maybe a better example is `-reference UnreferencedAssembly --action copy` (setting the *default* action to copy instead). In this case we would need to root the assembly if it isn't attributed as `IsTrimmable`.
- If someone publishes an app with a (statically unreferenced) nuget dependency, *without* trimming, they would expect to see it in the output. Any non-`IsTrimmable` assemblies should similarly be kept with `PublishTrimmed`.
- We don't want to pass `-a` for such cases because it would prevent trimming for dependencies that *are* `IsTrimmable`.

Then I would do the same with the per-assembly `copy` action for consistency (this is the first suggestion I made in https://github.com/mono/linker/issues/1800#issuecomment-773497902).

src/linker/Linker.Steps/RootNonTrimmableAssemblies.cs
```diff
@@ -0,0 +1,55 @@
...
+			// Walk over all -reference inputs and resolve any that may need to be rooted
+			foreach (var assemblyPath in GetInputAssemblyPaths ()) {
+				var assemblyName = Path.GetFileNameWithoutExtension (assemblyPath);
+
+				if (!MaybeIsFullyPreservedAssembly (assemblyName))
+					continue;
+
+				var assembly = Context.TryResolve (assemblyName);
+				if (assembly == null) {
+					Context.LogError ($"Reference assembly '{assemblyPath}' could not be loaded", 1039);
```
sbomer, line 23: I am wondering if this should be a warning instead. We shouldn't be passing in bad references in the first place, but the resolution logic here is stricter than what we used to do.

src/linker/Linker.Steps/ProcessReferencesStep.cs
```diff
@@ -0,0 +1,75 @@
...
+				}
+
+				// If the assigned action (now taking into account the IsTrimmable attribute) requires us
+				// to root the assembly, do so.
+				if (IsFullyPreservedAction (Annotations.GetAction (assembly)))
+					Annotations.Mark (assembly.MainModule, new DependencyInfo (DependencyKind.AssemblyAction, assembly));
+			}
+		}
+
+		public IEnumerable<string> GetInputAssemblyPaths ()
```
marek-safar, line 52: nit: should be private

src/linker/Linker.Steps/ProcessReferencesStep.cs
```diff
@@ -0,0 +1,75 @@
...
+		{
+			var assemblies = new HashSet<string> ();
+			foreach (var referencePath in Context.Resolver.GetReferencePaths ()) {
+				var assemblyName = Path.GetFileNameWithoutExtension (referencePath);
+				if (assemblies.Add (assemblyName))
+					yield return referencePath;
+			}
+		}
+
+		public static bool IsFullyPreservedAction (AssemblyAction action)
```
marek-safar, line 62: nit: should be private

marek-safar (approved):
Please hold the merge after .NET6 P2 SDK is snapped

src/linker/Linker.Steps/ProcessReferencesStep.cs
sbomer: This one is used also from MarkStep

src/linker/Linker.Steps/RootNonTrimmableAssemblies.cs
sbomer: Leaving it as an error for consistency with other command-line inputs. If it hurts people we can change to a warning.

PR #1848, opened 2021-02-24
Don't publish test results in official build

akoeplinger:
We don't run tests there so we get a warning about missing test result files in the official build.


marek-safar (approved):

ViktorHofer (commented):
LGTM

PR #1850, opened 2021-02-26
Don't override the ILLinkTaskAssembly in the SDK

ViktorHofer:
NuGet generated props and targets are imported before the SDK's props and targets. To avoid the SDK overriding the nuget path, add an additional property which is being checked on.


marek-safar (approved):

marek-safar:
Merging to test this for real

PR #1854, opened 2021-02-26
Mark facades containing forwarders referenced via type name strings

### LINKED ISSUE #1542: Linker doesn't resolve type references to facades in data flow cases
For example `[DynamicDependency(".ctor","System.Xml.Linq.XElement","System.Xml.Linq")]` will generate warning (IL2036) because it won't be able to find the type `XElement` in `System.Xml.Linq` - the type is actually not declared there, `System.Xml.Linq` only has a type forward for it.

Similarly when passing in a constant string to a value annotated with `DynamicallyAccessedMember` - the type resolution also doesn't see through type-forwards.
This also applies to intrinsic handling for things like `Type.GetType`.

mateoatr:
Reachable facade assemblies are currently dropped by the linker even if its forwarders are explicitly referenced in user annotations (note however that the user can work around this by rooting the facades).

This PR changes linker behavior so that type name strings pointing to a forwarded type cause the linker to mark the facade containing the forwarder as well as all other forwarders in the transitive chain pointing to the resolved type.

Fixes #1542

@eerhardt

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -807,8 +807,10 @@ namespace Mono.Linker.Steps
 			ModuleDefinition module = assembly.MainModule;
 			if (module.HasExportedTypes) {
 				foreach (var exportedType in module.ExportedTypes)
-					if (exportedType.FullName == type.FullName)
+					if (exportedType.Resolve() == type) {
 						_context.MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.DynamicDependency, type));
+						break;
+					}
 			}
 
 			IEnumerable<IMemberDefinition> members;
@@ -1804,7 +1806,7 @@ namespace Mono.Linker.Steps
 			TypeDefinition tdef = null;
 			switch (attribute.ConstructorArguments[0].Value) {
 			case string s:
-				tdef = _context.TypeNameResolver.ResolveTypeName (s).Type?.Resolve ();
+				tdef = _context.TypeNameResolver.ResolveTypeName (s, out _)?.Resolve ();
 				break;
 			case TypeReference type:
 				tdef = type.Resolve ();
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -807,7 +807,7 @@ namespace Mono.Linker.Steps
 			ModuleDefinition module = assembly.MainModule;
 			if (module.HasExportedTypes) {
 				foreach (var exportedType in module.ExportedTypes)
-					if (exportedType.Resolve() == type) {
+					if (exportedType.Resolve () == type) {
 						_context.MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.DynamicDependency, type));
 						break;
 					}
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -824,6 +824,11 @@ namespace Mono.Linker.Steps
 					_context.LogWarning ($"Unresolved type '{typeName}' in DynamicDependencyAttribute", 2036, context);
 					return;
 				}
+
+				var module = assembly.MainModule;
+				if (module.GetMatchingExportedType (type, out var exportedType)) {
+					_context.MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.ModuleOfExportedType, module));
+				}
 			} else if (dynamicDependency.Type is TypeReference typeReference) {
 				type = typeReference.Resolve ();
 				if (type == null) {
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -331,41 +331,7 @@ namespace Mono.Linker.Steps
 
 		void Process ()
 		{
-			while (ProcessPrimaryQueue () || ProcessMarkedPending () || ProcessLazyAttributes () || ProcessLateMarkedAttributes () || MarkFullyPreservedAssemblies ()) {
-
-				// deal with [TypeForwardedTo] pseudo-attributes
-
-				// This marks all exported types that resolve to a marked type. Note that it
-				// may mark unused exported types that happen to resolve to a type which was
-				// marked from a different reference. This seems incorrect.
-				// Note also that we may still remove type forwarder assemblies with marked exports in SweepStep,
-				// if they don't contain other used code.
-				// https://github.com/mono/linker/issues/1740
-				foreach (AssemblyDefinition assembly in _context.GetAssemblies ()) {
-					if (!assembly.MainModule.HasExportedTypes)
-						continue;
-
-					foreach (var exported in assembly.MainModule.ExportedTypes) {
-						bool isForwarder = exported.IsForwarder;
-						var declaringType = exported.DeclaringType;
-						while (!isForwarder && (declaringType != null)) {
-							isForwarder = declaringType.IsForwarder;
-							declaringType = declaringType.DeclaringType;
-						}
-
-						if (!isForwarder)
-							continue;
-						TypeDefinition type = exported.Resolve ();
-						if (type == null)
-							continue;
-						if (!Annotations.IsMarked (type))
-							continue;
-						var di = new DependencyInfo (DependencyKind.ExportedType, type);
-						_context.MarkingHelpers.MarkExportedType (exported, assembly.MainModule, di);
-					}
-				}
-			}
-
+			while (ProcessPrimaryQueue () || ProcessMarkedPending () || ProcessLazyAttributes () || ProcessLateMarkedAttributes () || MarkFullyPreservedAssemblies ()) ;
 			ProcessPendingTypeChecks ();
 		}
 
@@ -824,11 +790,6 @@ namespace Mono.Linker.Steps
 					_context.LogWarning ($"Unresolved type '{typeName}' in DynamicDependencyAttribute", 2036, context);
 					return;
 				}
-
-				var module = assembly.MainModule;
-				if (module.GetMatchingExportedType (type, out var exportedType)) {
-					_context.MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.ModuleOfExportedType, module));
-				}
 			} else if (dynamicDependency.Type is TypeReference typeReference) {
 				type = typeReference.Resolve ();
 				if (type == null) {
@@ -841,6 +802,13 @@ namespace Mono.Linker.Steps
 					_context.LogWarning ($"Unresolved type '{context.DeclaringType}' in DynamicDependencyAttribute", 2036, context);
 					return;
 				}
+			}
+
+			ModuleDefinition module = assembly.MainModule;
+			if (module.HasExportedTypes) {
+				foreach (var exportedType in module.ExportedTypes)
+					if (exportedType.FullName == type.FullName)
+						_context.MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.DynamicDependency, type));
 			}
 
 			IEnumerable<IMemberDefinition> members;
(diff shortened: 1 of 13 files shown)


src/linker/Linker/MarkingHelpers.cs
```diff
@@ -12,12 +11,30 @@ public MarkingHelpers (LinkContext context)
...
+		public void MarkExportedType (ExportedType exportedType, ModuleDefinition module, in DependencyInfo reason)
 		{
-			if (!_context.Annotations.MarkProcessed (type, reason))
+			if (!_context.Annotations.MarkProcessed (exportedType, reason))
 				return;
+
+			if (reason.Kind == DependencyKind.ModuleOfExportedType) {
+				var facadeAssembly = (reason.Source as ModuleDefinition).Assembly;
+				TypeDefinition resolvedType = exportedType.Resolve ();
+				while (facadeAssembly.FullName != resolvedType.Module.Assembly.FullName) {
```
MichalStrehovsky, line 22: Wondering if instead of trying to fix the forwarder chain we could get away with marking just the forwarder and then when we're writing out the output, we would "fix" it to point to a thing that exists (so if we're saving a forwarder, we would check whether the thing its pointing to is marked and follow that until we find the marked thing).

That way e.g. referencing System.Object from netstandard won't bring System.Runtime.dll.

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -86,7 +86,13 @@ TypeReference ResolveTypeName (AssemblyDefinition assembly, TypeName typeName)
 				};
 			}
 
-			return assembly.MainModule.ResolveType (typeName.ToString ());
+			ModuleDefinition module = assembly.MainModule;
+			TypeDefinition typeDefinition = module.ResolveType (typeName.ToString ());
+			if (module.GetMatchingExportedType (typeDefinition, out var exportedType)) {
+				_context.MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.ModuleOfExportedType, module));
```
marek-safar, line 92: This is very ugly. Why should Resolve (discovery) call mark a type?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -824,6 +824,11 @@ void MarkDynamicDependency (DynamicDependency dynamicDependency, IMemberDefiniti
 					_context.LogWarning ($"Unresolved type '{typeName}' in DynamicDependencyAttribute", 2036, context);
 					return;
 				}
+
+				var module = assembly.MainModule;
+				if (module.GetMatchingExportedType (type, out var exportedType)) {
```
marek-safar, line 829: Why do you need to do this when you have correct data in dynamicDependency.AssemblyName ?

src/linker/Linker/MarkingHelpers.cs
sbomer: I think we do something almost like this at https://github.com/mono/linker/blob/main/src/linker/Linker.Steps/SweepStep.cs#L683 - it will fix up forwarders so that they point directly at a resolved type - but it won't handle cases where the type can't be resolved. So if you have a -> b -> c -> (unresolved) where a and c are marked forwarders, it won't rewrite a to point to c. But I don't think it will necessarily keep b either - it would be great to add a testcase for this.

src/linker/Linker/MarkingHelpers.cs
```diff
@@ -12,12 +11,30 @@ public MarkingHelpers (LinkContext context)
...
 		}
 
-		public void MarkExportedType (ExportedType type, ModuleDefinition module, in DependencyInfo reason)
+		public void MarkExportedType (ExportedType exportedType, ModuleDefinition module, in DependencyInfo reason)
 		{
-			if (!_context.Annotations.MarkProcessed (type, reason))
+			if (!_context.Annotations.MarkProcessed (exportedType, reason))
 				return;
+
+			if (reason.Kind == DependencyKind.ModuleOfExportedType) {
```
sbomer, line 19: I wouldn't condition it on a particular Kind - I think we should treat all marked exported types similarly. For example if an XML references a type forwarder I think we should keep it. And this logic in MarkStep should just go away:
https://github.com/mono/linker/blob/59eef1f943e263e876bc1194dffcdd7900906be4/src/linker/Linker.Steps/MarkStep.cs#L364

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -531,6 +531,14 @@ void SweepDebugInfo (Collection<MethodDefinition> methods)
 
 		protected virtual bool ShouldRemove<T> (T element) where T : IMetadataTokenProvider
 		{
+			if (typeof (T) == typeof (ExportedType)) {
+				var typeDef = (element as ExportedType).Resolve ();
+				if (Context.KeepTypeForwarderOnlyAssemblies)
+					return !Annotations.IsMarked (typeDef);
+
+				return !Annotations.IsMarked (typeDef) || !Annotations.IsMarked (element);
```
marek-safar, line 539: What is the rationale behind keeping forwarders that are never used?

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -13,31 +13,35 @@ public TypeNameResolver (LinkContext context)
 			_context = context;
 		}
 
-		public TypeReference ResolveTypeName (string typeNameString)
+		public (AssemblyDefinition Assembly, TypeReference Type) ResolveTypeName (string typeNameString)
```
marek-safar, line 16: I'd use `public TypeReference ResolveTypeName (string typeNameString, out AssemblyDefinition typeAssembly)` so the callers can use discard to indicate that the value is not relevant

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -531,6 +531,14 @@ void SweepDebugInfo (Collection<MethodDefinition> methods)
 
 		protected virtual bool ShouldRemove<T> (T element) where T : IMetadataTokenProvider
 		{
+			if (typeof (T) == typeof (ExportedType)) {
+				var typeDef = (element as ExportedType).Resolve ();
```
marek-safar, line 535: If you need ExportedType special it'd be nicer to introduce SweepCollectionExportedTypes method

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -843,6 +804,13 @@ void MarkDynamicDependency (DynamicDependency dynamicDependency, IMemberDefiniti
 				}
 			}
 
+			ModuleDefinition module = assembly.MainModule;
+			if (module.HasExportedTypes) {
+				foreach (var exportedType in module.ExportedTypes)
+					if (exportedType.FullName == type.FullName)
+						_context.MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.DynamicDependency, type));
```
marek-safar, line 811: missing break?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -843,6 +804,13 @@ void MarkDynamicDependency (DynamicDependency dynamicDependency, IMemberDefiniti
 				}
 			}
 
+			ModuleDefinition module = assembly.MainModule;
+			if (module.HasExportedTypes) {
+				foreach (var exportedType in module.ExportedTypes)
+					if (exportedType.FullName == type.FullName)
```
marek-safar, line 810: It's best to avoid falling FullName it creates a new string on every call

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1330,7 +1298,8 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
 			MarkCustomAttributes (assembly.MainModule, new DependencyInfo (DependencyKind.AssemblyOrModuleAttribute, assembly.MainModule), null);
 
 			if (assembly.MainModule.HasExportedTypes) {
-				// TODO: This needs more work accross all steps
+				foreach (var exportedType in assembly.MainModule.ExportedTypes)
```
marek-safar, line 1301: Could you add a test for this?

src/linker/Linker.Steps/SweepStep.cs
mateoatr: Changed it -- we just keep forwarders when both the forwarder and the type it points to are marked (unless we use `--keep-forwarders`, in which case we keep the forwarder iff the type it points to is marked).

marek-safar (approved):
Please add the test and fix the FullName comparison and LGTM

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1325,7 +1300,8 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
...
+					_context.MarkingHelpers.MarkExportedType (exportedType, assembly.MainModule, new DependencyInfo (DependencyKind.ModuleOfExportedType, assembly.MainModule));
```
sbomer, line 1304: Same feedback I left in https://github.com/mono/linker/pull/1781#discussion_r563930820:

> This looks like a significant behavior change:
> 
>- We didn't used to keep type forwards in copy assemblies - this could result in some extra facade dlls being kept that weren't needed previously. It's probably uncommon, since facades are usually going to be copyused as part of the framework, but it's unclear how copyused is supposed to behave now. Should a marked copyused assembly also have exported types kept? (This would be a bigger issue since copyused are more likely to pull in facades).
>- I _think_ this will prevent SweepStep from sweeping assembly refs from copy assemblies, so it will never update scopes for Copy assemblies. Same question - how should this work for copyused assemblies?
> 
> If we go this direction and mark all copy type forwarders, I think we can simplify a few cases in SweepStep:
> 
>- get rid of the copy case from UpdateAssembliesReferencingRemovedAssembly
>- get rid of the SweepTypeForwarders handling for copy assemblies

test/Mono.Linker.Tests.Cases/LinkXml/CanPreserveAnExportedType.cs
```diff
@@ -10,7 +10,6 @@ namespace Mono.Linker.Tests.Cases.LinkXml
 	// Add another assembly in that uses the forwarder just to make things a little more complex
 	[SetupCompileBefore ("Forwarder.dll", new[] { "Dependencies/CanPreserveAnExportedType_Forwarder.cs" }, references: new[] { "Library.dll" })]
 
-	[RemovedAssembly ("Forwarder.dll")]
```
sbomer, line 13: Could you validate that the exported type is kept in Forwarder.dll?

test/Mono.Linker.Tests.Cases/LinkXml/CanPreserveExportedTypesUsingRegex.cs
```diff
@@ -10,7 +10,6 @@ namespace Mono.Linker.Tests.Cases.LinkXml
 	// Add another assembly in that uses the forwarder just to make things a little more complex
 	[SetupCompileBefore ("Forwarder.dll", new[] { "Dependencies/CanPreserveAnExportedType_Forwarder.cs" }, references: new[] { "Library.dll" })]
 
-	[RemovedAssembly ("Forwarder.dll")]
```
sbomer, line 13: Same here

test/Mono.Linker.Tests.Cases/LinkXml/UsedNonRequiredExportedTypeIsKept.cs
```diff
@@ -13,7 +13,6 @@ namespace Mono.Linker.Tests.Cases.LinkXml
 	[KeptMemberInAssembly ("libfwd.dll", typeof (UsedNonRequiredExportedTypeIsKept_Used1), "field", ExpectationAssemblyName = "lib.dll")]
 	[KeptMemberInAssembly ("libfwd.dll", typeof (UsedNonRequiredExportedTypeIsKept_Used2), "Method()", ExpectationAssemblyName = "lib.dll")]
 	[KeptMemberInAssembly ("libfwd.dll", typeof (UsedNonRequiredExportedTypeIsKept_Used3), "Method()", ExpectationAssemblyName = "lib.dll")]
-	[RemovedAssembly ("lib.dll")]
```
sbomer, line 16: Same here - and would you mind swapping the names of lib/libfwd? It looks like libfwd is the implementation and lib is the forwarder.

src/linker/Linker.Steps/MarkStep.cs
sbomer: To clarify, I see two options:
1. Ensure that copy *means* copy. The file is not modified, we keep exported types in it, we don't rewrite typerefs - so we need to keep referenced exported types in other assemsblies to ensure the typerefs remain valid.
2. We keep the behavior that copy has today - it may modify the file to rewrite type refs, sweep unmarked exported types in it - and later introduce an additional option (copyoriginal?) that has the behavior of 1.

mateoatr: This is trying to accomplish the second option you mention. The `copy` action should behave the way it does today (mostly) -- the relevant change is we now keep the facade of marked exported types. I'm adding the relevant tests for this.

mateoatr:
Closed in favor of #1869.

PR #1865, opened 2021-03-03
Add codefix for RequiresUnreferencedCode

agocke:
Simple codefix that just adds RequiresUnreferencedCode to the containing
method if a RequiresUnreferencedCode analyzer warning fires.


src/ILLink.CodeFix/ILLink.CodeFixProvider.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
marek-safar, line 1: Is this the common setup? I'd expect the analyzer and fix to live in the same assembly so they don't get accidentally out of sync

illink.sln
```diff
@@ -27,9 +27,11 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ILLink.RoslynAnalyzer", "sr
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ILLink.RoslynAnalyzer.Tests", "test\ILLink.RoslynAnalyzer.Tests\ILLink.RoslynAnalyzer.Tests.csproj", "{90D64CE4-C891-4B98-AF59-EE9B04BA1CBE}"
 EndProject
-Project("{9344BDBB-3E7F-41FC-A0DD-8665D75EE146}") = "Mono.Linker", "src\linker\ref\Mono.Linker.csproj", "{8DA71B3B-5809-44E5-A018-5DE5C6FF6C2A}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Mono.Linker", "src\linker\ref\Mono.Linker.csproj", "{8DA71B3B-5809-44E5-A018-5DE5C6FF6C2A}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "tlens", "src\tlens\tlens.csproj", "{B218CBE9-ADCE-4086-9DBF-E6C782136E6D}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "tlens", "src\tlens\tlens.csproj", "{B218CBE9-ADCE-4086-9DBF-E6C782136E6D}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ILLink.CodeFixProvider", "src\ILLink.CodeFix\ILLink.CodeFixProvider.csproj", "{6D20F334-B7E4-4585-854B-8A0E2B29B4AA}"
```
marek-safar, line 34: Please update also `eng/Build.props`

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Simplification;
+
+namespace ILLink.RoslynAnalyzer
```
marek-safar, line 16: The namespace name can be confusing for codefixers

src/ILLink.CodeFix/ILLink.CodeFixProvider.csproj
agocke: They can't -- the fixer is an IDE tool, while the analyzer is a compiler plugin. The environment for the two is different, and the workspaces layer (that the fixer depends on) doesn't exist in the compiler.

illink.sln
agocke: Looked -- didn't see any `Build.props` file

marek-safar: It's this file https://github.com/mono/linker/blob/main/eng/Build.props

agocke: Ah I just hadn't integrated from main.

agocke:
@sbomer @vitek-karas @LakshanF @tlakollo @mateoatr for review

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+		{
+			var editor = new SyntaxEditor (root, document.Project.Solution.Workspace);
+			var generator = editor.Generator;
+
+			var semanticModel = await document.GetSemanticModelAsync (cancellationToken).ConfigureAwait (false);
+			if (semanticModel is null) {
+				return document;
+			}
+			var methodSymbol = (IMethodSymbol) semanticModel.GetDeclaredSymbol (methodDecl);
+			var name = semanticModel.GetSymbolInfo (targetNode).Symbol?.Name;
```
sbomer, line 82: Maybe rename this to something like 'calledMethodName'?

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+			SyntaxNode[] attrArgs;
+			if (string.IsNullOrEmpty (name) || HasPublicAccessibility (methodSymbol)) {
+				attrArgs = Array.Empty<SyntaxNode> ();
+			} else {
+				attrArgs = new[] { generator.LiteralExpression ($"calls {name}") };
```
sbomer, line 87: I wonder if it should include the type name, particularly if it's a method in another type.

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+					Simplifier.AddImportsAnnotation);
+
+			editor.AddAttribute (methodDecl, newAttribute);
+
+			return document.WithSyntaxRoot (editor.GetChangedRoot ());
+		}
+
+		private static bool HasPublicAccessibility (IMethodSymbol m)
+		{
+			if (m is not { DeclaredAccessibility: Accessibility.Public or Accessibility.Protected }) {
```
sbomer, line 103: Neat!

src/ILLink.CodeFix/ILLink.CodeFixProvider.csproj
sbomer: Do we plan to ship this or is it just for our use? I'm wondering how we would package it to avoid including two copies of the analyzer dll in the sdk. The SDK looks like it uses a single input for both analyzers and code fixes: https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets#L111

mateoatr:
I'm not familiar with codefix specifics, but LGTM overall. Could you add a test where a method with RUC is called from within a lambda?
```C#
void Foo() {
   Action callMethodWithRUC = () => Bar();
   callMethodWithRUC();
}

void Bar() { }
```

mateoatr:
I'm not familiar with codefix specifics, but LGTM overall. Could you add a test where a method with RUC is called from within a lambda?
```C#
void Foo() {
   Action callMethodWithRUC = () => Bar();
   callMethodWithRUC();
}

void Bar() { }
```

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
```diff
@@ -36,6 +71,85 @@ class C
...
+    public class F
+    {
+        [RequiresUnreferencedCode()]
+        public int M5(C c) => c.M1();
+    }
+}
+";
+
+			await VerifyRequiresUnreferencedCodeCodeFix (test, fixtest, new[] {
+	// /0/Test0.cs(9,17): warning IL2026: Using method 'C.M1()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. message.
```
tlakollo, line 138: Nit: We should delete the name of the file /0/Tests0.cs to be consistent with the other tests

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpCodeFixVerifier`2+Test.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp.Testing;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Testing.Verifiers;
+using Microsoft.CodeAnalysis.Text;
+
+namespace ILLink.RoslynAnalyzer.Tests
+{
+	public partial class CSharpCodeFixVerifier<TAnalyzer, TCodeFix>
```
tlakollo, line 11: Why do we need this on a separate file and is not included in the  test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpCodeFixVerifier`2.cs file?

agocke: No real reason, it just happened to be the way that the code fix templates laid things out.

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
agocke: This seems to be the format written by the verifier utils -- I'd rather find some way to change that at the root. Otherwise we'll just be constantly fighting the tooling.

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
agocke: I struggled with the particular message to use here.

For adding the type name, shouldn't you be able to see that in the method body? And if you get rid of the codefix you'll see red underlines at the exact source.

I'm not sure if the verbosity helps or hurts here.

src/ILLink.CodeFix/ILLink.CodeFixProvider.csproj
agocke: I think we should probably ship this. I'll have a follow-up PR for the NuGet authoring for that.

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
sbomer: That's fair - I'd guess in most cases it'll be unambiguous anyway.

Nit: I would capitalize it so the full message for calls to these is:
`Using method 'Method' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Calls "CalleeName".`

tlakollo (approved):

marek-safar:
Please be careful when merging changes not to land accidental changes (submodule revert in this case).

PR #1866, opened 2021-03-03
Remove IVT attribute referring to removed or missing assemblies

marek-safar:


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -233,6 +235,25 @@ void Complete ()
...
+
+			foreach (var attr in _ivt_attributes) {
+				if (IsInternalsVisibleAttributeAssemblyMarked (attr.Attribute))
+					MarkCustomAttribute (attr.Attribute, new DependencyInfo (DependencyKind.AssemblyOrModuleAttribute, attr.Provider), null);
+			}
+
+			bool IsInternalsVisibleAttributeAssemblyMarked (CustomAttribute ca)
+			{
+				var assemblyName = (string) ca.ConstructorArguments[0].Value;
+				int comma = assemblyName.IndexOf (',');
```
MichalStrehovsky, line 247: Would it be better to roundtrip this through `AssemblyName`? This string has various rules that need to be followed (whitespace is allowed and gets trimmed, the name can be enclosed in quotation marks, etc.).

marek-safar: Good idea.

vitek-karas (approved):

Issue #1669, opened 2020-12-04
Disabling trimming per assembly with trimming does change the assembly


vitek-karas:
---
**Issue moved from dotnet/runtime#45598**
- Please respond to @NN---.

---

_From @NN--- on Friday, December 4, 2020 4:50:39 PM_

### Description

Specifying `TrimmerRootAssembly` and even `IsTrimmable false` does change the package file.

### Configuration

.NET Core 3.1 or .NET 5.0

project.csproj:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <TrimmerRootAssembly Include="SMBLibrary" />

    <PackageReference Include="SMBLibrary" Version="1.4.3">
      <IsTrimmable>false</IsTrimmable>
    </PackageReference>
  </ItemGroup>
</Project>
```

program.cs
```cs
class Program { public static void Main() {} }
```

Run:
`dotnet publish project.csproj -c Release -r win-x64 -o publish  /p:PublishSingleFile=true /p:PublishTrimmed=true`

Check file inside: `obj\Release\netcoreapp3.1\win-x64\linked\SMBLibrary.dll`

The file size is not the same as original file in 
%userprofile%\.nuget\packages\smblibrary\1.4.3\lib\netstandard2.0\SMBLibrary.dll

If you don't specify `PublishTrimmed=true` then the file is preserved but the resulting output is much bigger.

The expectation is that TrimmerRootAssembly preserves original file as described in [documentation](https://docs.microsoft.com/en-us/dotnet/core/deploying/trim-self-contained)

### Regression?

Not a regression.

vitek-karas:
---
**Issue moved from dotnet/runtime#45598**
- Please respond to @msftbot[bot].

---

_From @msftbot[bot] on Friday, December 4, 2020 4:50:43 PM_

Tagging subscribers to this area: @vitek-karas, @agocke, @coffeeflux
See info in area-owners.md if you want to be subscribed.
<details>
<summary>Issue Details</summary>
<hr />



### Description



Specifying `TrimmerRootAssembly` and even `IsTrimmable false` does change the package file.

### Configuration

.NET Core 3.1 or .NET 5.0

project.csproj:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
[...]

PR #1869, opened 2021-03-04
Change behavior of copy action to not rewrite assemblies

mateoatr:
This PR changes the behavior of `copy` and `copyused` assembly actions. The assemblies will be analyzed and marked as usual but contents will not be rewritten under any circumpstance -- we will simply copy the assembly as is to the output directory. This will help us move forward with enabling the analysis of C++/CLI assemblies and issues such as #1669 (and, of no lesser importance, have action names that behave as one would expect).

This also takes over from the changes in #1854 : type forwarders which are referenced via type name strings are now kept, even when these exist inside linked assemblies.


marek-safar:
What is the benefit of adding yet another copy mode (we have 3 already and they are a source of great confusion) ?

sbomer:
This is trying to solve the problem where copy assemblies can be rewritten. We need to either guarantee that copy doesn't modify an assembly as part of https://github.com/mono/linker/pull/1854, or introduce new functionality that has such a guarantee. This will be useful to address:
- [Disabling trimming per assembly with trimming does change the assembly](https://github.com/mono/linker/issues/1669)
- [Linker removes data from "copy" assemblies](https://github.com/mono/linker/issues/639)
- and more broadly this is important for [C++/CLI](https://github.com/mono/linker/issues/676).

mateoatr:
Yes. Main motivation of this is to keep the semantics of the `copy` action untouched and have a way to tell the linker "don't modify anything, don't even rewrite removed references".

test/Mono.Linker.Tests.Cases/TypeForwarding/UsedTransitiveForwarderIsDynamicallyAccessed.cs
```diff
@@ -0,0 +1,40 @@
...
+
+	[KeptAssembly ("FirstForwarder.dll")]
+	[KeptMemberInAssembly ("Implementation.dll", typeof (ImplementationLibrary), "GetSomeValue()")]
+	[RemovedAssembly ("SecondForwarder.dll")]
+	[RemovedForwarder ("FirstForwarder.dll", nameof (ImplementationStruct))]
+	class UsedTransitiveForwarderIsDynamicallyAccessed
+	{
+		static void Main ()
+		{
+			// FirstForwarder -> SecondForwarder -> Implementation
```
sbomer, line 29: Could you add a case like this where `FirstForwarder` is in a copy assembly?

sbomer (commented):
I think the following cases still aren't handled:
- `copy` assembly with a static reference to a forwarded type
  - This will show up whenever a non-trimmable assembly was built against a facade like netstandard.
- a reference that we can't rewrite (`copy` or string reference) -> forwarder in `copy` assembly (which we also can't rewrite) -> forwarder (currently won't get marked) -> type
  - This might be rare, but to be correct, whenever we mark a forwarder, we'd need to walk the path to the resolved type, marking the target of forwarders until we see one not in a `copy` assembly - after that, any remaining forwarders in the chain can be rewritten. I'm not sure we need to handle this, but it would at least be good to add a testcase for it with a comment explaining the current behavior.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -97,14 +97,14 @@ void UpdateAssemblyReferencesToRemovedAssemblies (AssemblyDefinition assembly)
 		{
 			var action = Annotations.GetAction (assembly);
 			switch (action) {
-			case AssemblyAction.Skip:
+			case AssemblyAction.Copy:
+			case AssemblyAction.CopyUsed:
```
sbomer, line 101: I know we discussed the idea of changing `CopyUsed` to be like `Copy` when it's marked, but I don't remember if we decided to do it. As-is I think it has the same problem as `Copy` where static typerefs won't get updated, but if we fix that it could bring in more facades than we want. I think it would be safest to keep the current behavior for back-compat, and just recommend not using `CopyUsed`.

test/Mono.Linker.Tests.Cases/LinkXml/UsedNonRequiredExportedTypeIsKept.cs
```diff
@@ -5,15 +5,18 @@ namespace Mono.Linker.Tests.Cases.LinkXml
 {
 	[SetupLinkerDescriptorFile ("UsedNonRequiredExportedTypeIsKept.xml")]
 
-	[SetupCompileBefore ("lib.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_lib.cs" })]
-	[SetupCompileAfter ("libfwd.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_lib.cs" })]
-	[SetupCompileAfter ("lib.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_fwd.cs" }, references: new[] { "libfwd.dll" })]
+	[SetupCompileBefore ("libfwd.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_lib.cs" })]
+	[SetupCompileAfter ("lib.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_lib.cs" })]
+	[SetupCompileAfter ("libfwd.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_fwd.cs" }, references: new[] { "lib.dll" })]
+	[SetupLinkerAction ("copy", "libfwd")]
```
sbomer, line 11: The "copy" action here is what's keeping the forwards, right? I think it'd be better to leave this one as "link" and add a new testcase where the forwarder is "copy". (Maybe leave a comment about the "link" behavior for forwarders referenced in descriptors - we might want to fix that in a separate change if not here)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -864,6 +830,15 @@ void MarkDynamicDependency (DynamicDependency dynamicDependency, IMemberDefiniti
 				}
 			}
 
+			ModuleDefinition module = assembly.MainModule;
+			if (module.HasExportedTypes) {
```
sbomer, line 834: Can this use `GetMatchingExportedType`?

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -172,22 +171,15 @@ protected void ProcessAssemblyAction (AssemblyDefinition assembly)
 				break;
 
 			case AssemblyAction.CopyUsed:
+				if (!Context.KeepTypeForwarderOnlyAssemblies && SweepTypeForwarders (assembly)) {
```
sbomer, line 174: Is the `KeepTypeForwarderOnlyAssemblies` condition here is to somewhat preserve the behavior that we had here in MarkStep?
https://github.com/mono/linker/blob/252c6652221834051f07b92fc9733e42a755ee36/src/linker/Linker.Steps/MarkStep.cs#L366

What goes wrong if we just remove this part of the condition? I thought that `KeepTypeForwarderOnlyAssemblies` would still allow sweeping forwarders, just not removing the assembly.

test/Mono.Linker.Tests.Cases/LinkXml/UsedNonRequiredExportedTypeIsKept.cs
```diff
@@ -5,15 +5,16 @@ namespace Mono.Linker.Tests.Cases.LinkXml
...
-	[SetupCompileAfter ("libfwd.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_lib.cs" })]
-	[SetupCompileAfter ("lib.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_fwd.cs" }, references: new[] { "libfwd.dll" })]
+	[SetupCompileBefore ("libfwd.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_lib.cs" })]
+	[SetupCompileAfter ("lib.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_lib.cs" })]
+	[SetupCompileAfter ("libfwd.dll", new[] { "Dependencies/UsedNonRequiredExportedTypeIsKept_fwd.cs" }, references: new[] { "lib.dll" })]
+
+	[KeptMemberInAssembly ("lib.dll", typeof (UsedNonRequiredExportedTypeIsKept_Used1), "field", ExpectationAssemblyName = "libfwd.dll")]
+	[KeptMemberInAssembly ("lib.dll", typeof (UsedNonRequiredExportedTypeIsKept_Used2), "Method()", ExpectationAssemblyName = "libfwd.dll")]
+	[KeptMemberInAssembly ("lib.dll", typeof (UsedNonRequiredExportedTypeIsKept_Used3), "Method()", ExpectationAssemblyName = "libfwd.dll")]
+	// Type forwarders are only kept when referenced through a type name string or the containing assembly has `copy` action.
```
sbomer, line 15: I'd add a note about the XML, something like:
Type forwarders referenced from descriptor XML are also not kept despite being referenced through a string.

test/Mono.Linker.Tests.Cases/References/AssemblyOnlyUsedByUsingWithCsc.cs
```diff
@@ -6,22 +6,21 @@ namespace Mono.Linker.Tests.Cases.References
...
-	// Here to assert that the test is setup correctly to copy the copied assembly.  This is an important aspect of the bug
 	[KeptMemberInAssembly ("copied.dll", typeof (AssemblyOnlyUsedByUsing_Copied), "Unused()")]
+	[KeptReferencesInAssembly ("copied.dll", new[] { "System.Private.CoreLib", "library" })]
 
-	// We library should be gone.  The `using` statement leaves no traces in the IL so nothing in `library` will be marked
 	[RemovedAssembly ("library.dll")]
 #if NETCOREAPP
-	[KeptReferencesInAssembly ("copied.dll", new[] { "System.Private.CoreLib" })]
+	[KeptReferencesInAssembly ("copied.dll", new[] { "System.Private.CoreLib", "library" })]
 #else
```
sbomer, line 24: While you're here I think we can get rid of the #if and the mscorlib case - we only ever build tests against netcoreapp now.

test/Mono.Linker.Tests.Cases/TypeForwarding/UnusedForwarderWithAssemblyCopyIsKept.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+	[SetupLinkerDefaultAction ("link")]
+	[SetupCompileBefore ("Forwarder.dll", new[] { "Dependencies/ReferenceImplementationLibrary.cs" }, defines: new[] { "INCLUDE_REFERENCE_IMPL" })]
+
+	[SetupCompileAfter ("Implementation.dll", new[] { "Dependencies/ImplementationLibrary.cs" })]
+	[SetupCompileAfter ("Forwarder.dll", new[] { "Dependencies/ForwarderLibrary.cs" }, references: new[] { "Implementation.dll" })]
+
+	[SetupLinkerAction ("copy", "Forwarder")]
+
+	[KeptAssembly ("Forwarder.dll")]
```
sbomer, line 19: Maybe check for a kept forwarded type specifically

test/Mono.Linker.Tests.Cases/TypeForwarding/UsedForwarderIsDynamicallyAccessedWithAssemblyCopyUsed.cs
```diff
@@ -0,0 +1,38 @@
...
+	[SetupLinkerAction ("copyused", "Forwarder")]
+	[KeepTypeForwarderOnlyAssemblies ("false")]
+
+	[SetupCompileBefore ("Forwarder.dll", new[] { "Dependencies/ReferenceImplementationLibrary.cs" }, defines: new[] { "INCLUDE_REFERENCE_IMPL" })]
+
+	// After compiling the test case we then replace the reference impl with implementation + type forwarder
+	[SetupCompileAfter ("Implementation.dll", new[] { "Dependencies/ImplementationLibrary.cs" })]
+	[SetupCompileAfter ("Forwarder.dll", new[] { "Dependencies/ForwarderLibrary.cs" }, references: new[] { "Implementation.dll" })]
+
+	[KeptAssembly ("Forwarder.dll")]
```
sbomer, line 22: Check for the kept forward

test/Mono.Linker.Tests.Cases/TypeForwarding/UsedTransitiveForwarderIsDynamicallyAccessed.cs
```diff
@@ -0,0 +1,40 @@
...
+	[KeepTypeForwarderOnlyAssemblies ("false")]
+
+	[SetupCompileBefore ("SecondForwarder.dll", new[] { "Dependencies/ReferenceImplementationLibrary.cs" }, defines: new[] { "INCLUDE_REFERENCE_IMPL" })]
+	[SetupCompileBefore ("FirstForwarder.dll", new[] { "Dependencies/ForwarderLibrary.cs" }, references: new[] { "SecondForwarder.dll" })]
+
+	// After compiling the test case we then replace the reference impl with implementation + type forwarder
+	[SetupCompileAfter ("Implementation.dll", new[] { "Dependencies/ImplementationLibrary.cs" })]
+	[SetupCompileAfter ("SecondForwarder.dll", new[] { "Dependencies/ForwarderLibrary.cs" }, references: new[] { "Implementation.dll" })]
+
+	[KeptAssembly ("FirstForwarder.dll")]
```
sbomer, line 21: Check that the forwarder is kept

test/Mono.Linker.Tests.Cases/TypeForwarding/UsedTransitiveForwarderInCopyUsedAssemblyIsDynamicallyAccessed.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+	[KeepTypeForwarderOnlyAssemblies ("false")]
+
+	[SetupCompileBefore ("SecondForwarder.dll", new[] { "Dependencies/ReferenceImplementationLibrary.cs" }, defines: new[] { "INCLUDE_REFERENCE_IMPL" })]
+	[SetupCompileBefore ("FirstForwarder.dll", new[] { "Dependencies/ForwarderLibrary.cs" }, references: new[] { "SecondForwarder.dll" })]
+
+	// After compiling the test case we then replace the reference impl with implementation + type forwarder
+	[SetupCompileAfter ("Implementation.dll", new[] { "Dependencies/ImplementationLibrary.cs" })]
+	[SetupCompileAfter ("SecondForwarder.dll", new[] { "Dependencies/ForwarderLibrary.cs" }, references: new[] { "Implementation.dll" })]
+	[SetupLinkerAction ("copyused", "FirstForwarder")]
```
sbomer, line 20: "copyused" still lets us rewrite FirstForwarder to point directly to Implementation - could you add a case where it's "copy"? (I believe this would remove SecondForwarder despite it being referenced by the copy assembly)

test/Mono.Linker.Tests.Cases/TypeForwarding/UsedTransitiveForwarderIsResolvedAndFacadeRemoved.cs
```diff
@@ -0,0 +1,33 @@
...
+
+	[KeptMemberInAssembly ("Implementation.dll", typeof (ImplementationLibrary), "GetSomeValue()")]
+	[RemovedMemberInAssembly ("Implementation.dll", nameof (ImplementationStruct))]
+	[RemovedAssembly ("FirstForwarder.dll")]
+	[RemovedAssembly ("SecondForwarder.dll")]
+	class UsedTransitiveForwarderIsResolvedAndFacadeRemoved
+	{
+		static void Main ()
+		{
+			var instance = new ImplementationLibrary ();
```
sbomer, line 29: Could you add a case like this where the test assembly is "copy"? I think that'll demonstrate the first case I mentioned in https://github.com/mono/linker/pull/1869#pullrequestreview-609959787 (even without the SecondForwarder)

test/Mono.Linker.Tests.Cases/LinkXml/UsedNonRequiredExportedTypeIsKept.cs
mateoatr: I forgot to update the XML here. Exported types referenced from a descriptor XML should be kept -- just as we keep any exported type which is referenced through a type name string.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1837,25 +1832,26 @@ protected virtual void MarkTypeConverterLikeDependency (CustomAttribute attribut
...
 
 			Tracer.AddDirectDependency (attribute, new DependencyInfo (DependencyKind.CustomAttribute, provider), marked: false);
-			if (MarkMethodsIf (tdef.Methods, predicate, new DependencyInfo (DependencyKind.ReferencedBySpecialAttribute, attribute), sourceLocationMember) &&
-				tdef.Module.HasExportedTypes) {
-				foreach (var exportedType in tdef.Module.ExportedTypes) {
-					if (exportedType.Resolve () == tdef)
-						_context.MarkingHelpers.MarkExportedType (exportedType, tdef.Module, new DependencyInfo (DependencyKind.ReferencedBySpecialAttribute, attribute));
+			if (MarkMethodsIf (typeDefinition.Methods, predicate, new DependencyInfo (DependencyKind.ReferencedBySpecialAttribute, attribute), sourceLocationMember) &&
+				assemblyDefinition != null && assemblyDefinition.MainModule.HasExportedTypes) {
+				foreach (var exportedType in assemblyDefinition.MainModule.ExportedTypes) {
```
sbomer, line 1852: Can this use GetMatchingExportedType?

src/linker/Linker.Steps/OutputStep.cs
```diff
@@ -285,18 +285,26 @@ protected virtual void CopyAssembly (AssemblyDefinition assembly, string directo
...
 
 			if (!Context.LinkSymbols)
 				return;
 
 			var mdb = source + ".mdb";
-			if (File.Exists (mdb))
-				CopyFileAndRemoveReadOnly (mdb, target + ".mdb");
+			if (File.Exists (mdb)) {
+				if (removeReadOnly)
+					CopyFileAndRemoveReadOnly (mdb, target + ".mdb");
```
marek-safar, line 296: Why about changing 

`static void CopyFileAndRemoveReadOnly (string src, string dest)`

to

`static void CopyFile (string src, string dest, bool removeReadOnly)`

test/Mono.Linker.Tests.Cases/TypeForwarding/UsedTransitiveForwarderInCopyUsedAssemblyIsDynamicallyAccessed.cs
mateoatr: Added: https://github.com/mono/linker/blob/1fd354197498e5273f0d3904b1bb49b280de8f7d/test/Mono.Linker.Tests.Cases/TypeForwarding/UsedTransitiveForwarderInCopyAssemblyIsDynamicallyAccessed.cs#L18

src/linker/Linker/MarkingHelpers.cs
```diff
@@ -11,12 +11,34 @@ public MarkingHelpers (LinkContext context)
...
 				return;
-			if (_context.KeepTypeForwarderOnlyAssemblies)
-				_context.Annotations.Mark (module, new DependencyInfo (DependencyKind.ModuleOfExportedType, type));
+
+			AssemblyDefinition facadeAssembly = module.Assembly;
+			TypeDefinition resolvedType = exportedType.Resolve ();
+			AssemblyAction facadeAction = _context.Annotations.GetAction (facadeAssembly);
+			if (facadeAssembly != null && resolvedType != null) {
+				while ((facadeAssembly != resolvedType.Module.Assembly) &&
+						facadeAction == AssemblyAction.Copy || facadeAction == AssemblyAction.CopyUsed) {
```
sbomer, line 24: I don't think we need the CopyUsed case here since we can still rewrite type forwarders in CopyUsed assemblies

src/linker/Linker/MarkingHelpers.cs
```diff
@@ -11,12 +11,34 @@ public MarkingHelpers (LinkContext context)
...
+					_context.Annotations.Mark (module, new DependencyInfo (DependencyKind.ModuleOfExportedType, exportedType));
+					foreach (var assemblyReference in module.AssemblyReferences) {
```
sbomer, line 26: I think this needs to resolve the pointed-to thing and mark it if it's a forwarder (not just mark the module). I would do it by extracting this into a helper that resolves a forwarder through only one level of indirection, to the next forwarder.

test/Mono.Linker.Tests.Cases/References/AssemblyOnlyUsedByUsingWithCsc.cs
```diff
@@ -6,25 +6,20 @@ namespace Mono.Linker.Tests.Cases.References
...
-	// Here to assert that the test is setup correctly to copy the copied assembly.  This is an important aspect of the bug
 	[KeptMemberInAssembly ("copied.dll", typeof (AssemblyOnlyUsedByUsing_Copied), "Unused()")]
+	[KeptReferencesInAssembly ("copied.dll", new[] { "System.Private.CoreLib", "library" })]
 
-	// We library should be gone.  The `using` statement leaves no traces in the IL so nothing in `library` will be marked
 	[RemovedAssembly ("library.dll")]
-#if NETCOREAPP
-	[KeptReferencesInAssembly ("copied.dll", new[] { "System.Private.CoreLib" })]
-#else
-	[KeptReferencesInAssembly ("copied.dll", new[] { "mscorlib" })]
```
mrvoorhe, line 26: It would be appreciated if you left the #if's for asserting mscorlib.  Unity will still be linking against an mscorlib for years to come.  Our linker is a super set of illink, which means we run and pass all of the same tests.  Dropping the #if's will introduce additional work for us to when we sync up as we'll have to add them back.

I believe someone added a constant at some point so that this code wouldn't have to use a #if?

vitek-karas: Use this instead: https://github.com/mono/linker/blob/main/test/Mono.Linker.Tests.Cases.Expectations/Helpers/PlatformAssemblies.cs

mateoatr: > Unity will still be linking against an mscorlib for years to come. 

I wasn't aware, thanks! Keeping the mscorlib check.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2588,6 +2574,30 @@ protected virtual MethodDefinition MarkMethod (MethodReference reference, Depend
...
+			// mark the transitive chain of forwarders.
+			TypeReference typeRef = memberReference is TypeReference ? memberReference as TypeReference : memberReference.DeclaringType;
+			TypeDefinition resolvedTypeDef = resolvedMember is TypeDefinition ? resolvedMember as TypeDefinition : resolvedMember.DeclaringType;
+			foreach (var assembly in memberReference.Module.AssemblyReferences) {
+				if (assembly.MetadataToken == typeRef.Scope.MetadataToken) {
+					ModuleDefinition refModule = typeRef.Module.AssemblyResolver.Resolve (assembly).MainModule;
+					if (refModule.GetMatchingExportedType (resolvedTypeDef, out var exportedType)) {
+						MarkingHelpers.MarkExportedType (exportedType, refModule, reason);
+
+						if (resolvedTypeDef.IsNested)
```
sbomer, line 2592: Regarding nested types: I believe the only typeref we need to check for forwarders is the outermost DeclaringType, since this is the one that could point to a type forward in another assembly. Instead of the recursive call for the declaring type (which will do unnecessary walks over the exported types), I would rather first get the outermost DeclaringType typeref, then resolve it and call MarkExportedType.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1359,15 +1334,20 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
...
-			if (assembly.MainModule.HasExportedTypes) {
-				// TODO: This needs more work accross all steps
-			}
-
-			foreach (TypeDefinition type in assembly.MainModule.Types)
+			foreach (TypeDefinition type in module.Types)
 				MarkEntireType (type, includeBaseTypes: false, includeInterfaceTypes: false, new DependencyInfo (DependencyKind.TypeInAssembly, assembly), null);
+
+			foreach (ExportedType exportedType in module.ExportedTypes)
+				MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.ExportedType, assembly));
```
sbomer, line 1345: Should also mark forwarded scope as discussed

src/linker/Linker.Steps/OutputStep.cs
```diff
@@ -271,42 +271,27 @@ FileInfo GetOriginalAssemblyFileInfo (AssemblyDefinition assembly)
...
-				WriteAssembly (assembly, directory, new WriterParameters ());
-				return;
-			}
-
 			FileInfo fi = GetOriginalAssemblyFileInfo (assembly);
 			string target = Path.GetFullPath (Path.Combine (directory, fi.Name));
 			string source = fi.FullName;
+
+			if (assembly.MainModule.HasSymbols && !Context.LinkSymbols && assembly.MainModule.SymbolReader is EmbeddedPortablePdbReader)
+				Context.LogWarning ("Debug symbols cannot be modified in a copied assembly.", 2026, new MessageOrigin (source));
```
sbomer, line 279: Use a new warning code

src/linker/Linker/MarkingHelpers.cs
```diff
@@ -11,12 +11,39 @@ public MarkingHelpers (LinkContext context)
...
-			if (!_context.Annotations.MarkProcessed (type, reason))
+			if (typeToMatch == null || assembly == null)
 				return;
-			if (_context.KeepTypeForwarderOnlyAssemblies)
-				_context.Annotations.Mark (module, new DependencyInfo (DependencyKind.ModuleOfExportedType, type));
+
+			ModuleDefinition module = assembly.MainModule;
+			if (module.GetMatchingExportedType (typeToMatch, out var exportedType)) {
+				MarkExportedType (exportedType, module, reason);
+				if (_context.Annotations.GetAction (assembly) == AssemblyAction.Copy) {
```
sbomer, line 22: Now that we're marking all exports and their scopes from copy assemblies, I don't think we need to check for copy here.

test/Mono.Linker.Tests.Cases/Symbols/ReferenceWithEmbeddedPdbCopyAction.cs
```diff
@@ -8,7 +8,8 @@ namespace Mono.Linker.Tests.Cases.Symbols
 	[SetupLinkerLinkSymbols ("false")]
 	[SetupLinkerAction ("copy", "LibraryWithEmbeddedPdbSymbols")]
 
-	[RemovedSymbols ("LibraryWithEmbeddedPdbSymbols.dll")]
+	// Copy assemblies cannot be modified.
+	[KeptSymbols ("LibraryWithEmbeddedPdbSymbols.dll")]
```
sbomer, line 12: Maybe check for the warning too

test/Mono.Linker.Tests.Cases/TypeForwarding/UnusedForwarderWithAssemblyCopyIsKept.cs
```diff
@@ -0,0 +1,28 @@
...
+	[SetupLinkerDefaultAction ("link")]
+	[SetupCompileBefore ("Forwarder.dll", new[] { "Dependencies/ReferenceImplementationLibrary.cs" }, defines: new[] { "INCLUDE_REFERENCE_IMPL" })]
+
+	[SetupCompileAfter ("Implementation.dll", new[] { "Dependencies/ImplementationLibrary.cs" })]
+	[SetupCompileAfter ("Forwarder.dll", new[] { "Dependencies/ForwarderLibrary.cs" }, references: new[] { "Implementation.dll" })]
+
+	[SetupLinkerAction ("copy", "Forwarder")]
+
+	[KeptMemberInAssembly ("Forwarder.dll", typeof (ImplementationLibrary))]
+	[RemovedAssembly ("Implementation.dll")]
```
sbomer, line 20: How is this test working? I thought that the copy forwarder would result in Implementation being kept due to the reference from the forwarder

test/Mono.Linker.Tests.Cases/TypeForwarding/UsedAndUnusedForwarderWithAssemblyCopy.cs
```diff
@@ -19,12 +19,14 @@ namespace Mono.Linker.Tests.Cases.TypeForwarding
...
 
-	[KeptAssembly ("Forwarder.dll")]
 	[KeptMemberInAssembly ("Implementation.dll", typeof (ImplementationLibrary), "GetSomeValue()")]
+	// The whole assembly is kept as is, since it is marked with the `copy` action.
 	[KeptMemberInAssembly ("Forwarder.dll", typeof (ImplementationLibrary))]
+	[KeptMemberInAssembly ("Forwarder.dll", "Mono.Linker.Tests.Cases.TypeForwarding.Dependencies.AnotherLibrary`1")]
+	[KeptReferencesInAssembly ("Forwarder.dll", new[] { "System.Private.CoreLib", "Implementation", "Unused" })]
+	// Even though `Forwarder` references this assembly, none of its members are marked (none is used) and, since `Unused`
+	// has `link` action, it is removed.
 	[RemovedAssembly ("Unused.dll")]
```
sbomer, line 29: Same question as above

test/Mono.Linker.Tests.Cases/TypeForwarding/UsedTransitiveForwarderInCopyAssemblyIsDynamicallyAccessed.cs
```diff
@@ -0,0 +1,43 @@
...
+	[SetupCompileBefore ("SecondForwarder.dll", new[] { "Dependencies/ReferenceImplementationLibrary.cs" }, defines: new[] { "INCLUDE_REFERENCE_IMPL" })]
+	[SetupCompileBefore ("FirstForwarder.dll", new[] { "Dependencies/ForwarderLibrary.cs" }, references: new[] { "SecondForwarder.dll" })]
+
+	// After compiling the test case we then replace the reference impl with implementation + type forwarder
+	[SetupCompileAfter ("Implementation.dll", new[] { "Dependencies/ImplementationLibrary.cs" })]
+	[SetupCompileAfter ("SecondForwarder.dll", new[] { "Dependencies/ForwarderLibrary.cs" }, references: new[] { "Implementation.dll" })]
+	[SetupLinkerAction ("copy", "FirstForwarder")]
+
+	[KeptMemberInAssembly ("FirstForwarder.dll", typeof (ImplementationLibrary))]
+	// Dynamically accessing a type forwarder T in a copy assembly will cause the linker
```
sbomer, line 21: Outdated comment. I think it might be useful to change this test to not access the forwarder at all, just to show that the forwarder target is kept for any copy forwarder.

src/linker/Linker/MarkingHelpers.cs
mateoatr: I think this still matters in the case where you dynamically access a type forwarder in a link assembly, then the forwarder will be kept, but we could change the resolution scope. There's an example for this scenario in UsedTransitiveForwarderIsDynamicallyAccessed.cs.

test/Mono.Linker.Tests.Cases/TypeForwarding/UnusedForwarderWithAssemblyCopyIsKept.cs
mateoatr: As discussed; we'll keep the whole copy assembly as is, but otherwise we will not mark anything that isn't either statically referenced or dynamically accessed.

sbomer: Thanks for clarifying offline @mateoatr. For anyone following along - we mark the scope of forwarders in a copy assembly, but don't necessarily mark the resolved type if nothing references it.

sbomer: Oh you beat me to it :)

src/linker/Linker/MarkingHelpers.cs
sbomer: Clarified that I meant to remove the whole if statement. Thanks!

mateoatr:
## Overview of changes

### Copy action

The current behavior of the copy action allows the linker to rewrite the contents of assemblies marked with this action whenever there's a reference to a removed assembly. The changes in this PR change the behavior of the copy action so that the assembly is still analyzed but none of its contents are touched -- the assembly will always be copied to the output folder as is. This also means that any embedded symbols will be kept, and hence having `-b true` will not update any of the symbols inside copy assemblies, which was allowed previously -- in this scenario we'll output warning `IL2014: Debug symbols cannot be modified in a copied assembly.`.

### Marking

- All type forwarders accessed through a type name string will be marked.
- In a copy assembly, all exported types will be marked.
- In a copy assembly, type references to forwarders will cause the marking of the scope of the forwarder.

Pertinent tests were added for the above scenarios.

## Size 

Size comparison of a blazorwasm template app linked in release mode before and after the changes in this PR:

Text in green indicate bigger size in the output.
Text in orange indicate smaller size in the output.
Text in red indicate a file that wasn't previously in the output.

Total size changed from 2.4M to 2.5M.

```diff
- 5.5K Microsoft.AspNetCore.Authorization.dll
 19K Microsoft.AspNetCore.Components.Web.dll
 55K Microsoft.AspNetCore.Components.WebAssembly.dll
137K Microsoft.AspNetCore.Components.dll
7.0K Microsoft.Extensions.Configuration.Abstractions.dll
8.5K Microsoft.Extensions.Configuration.Json.dll
 11K Microsoft.Extensions.Configuration.dll
 13K Microsoft.Extensions.DependencyInjection.Abstractions.dll
 38K Microsoft.Extensions.DependencyInjection.dll
 26K Microsoft.Extensions.Logging.Abstractions.dll
 17K Microsoft.Extensions.Logging.dll
 14K Microsoft.Extensions.Options.dll
+ 8.5K Microsoft.Extensions.Primitives.dll (before: 8.0K)
+ 20K Microsoft.JSInterop.WebAssembly.dll (before: 11K)
+ 51K Microsoft.JSInterop.dll (before: 42K)
- 4.5K System.Buffers.dll
 19K System.Collections.Concurrent.dll
 71K System.Collections.Immutable.dll
! 27K System.Collections.dll (before: 28K)
4.5K System.ComponentModel.dll
 14K System.Console.dll
- 4.5K System.Diagnostics.Debug.dll
+ 68K System.IO.Pipelines.dll (before: 58K)
 26K System.Linq.dll
! 20K System.Memory.dll (before: 21K)
 11K System.Net.Http.Json.dll
135K System.Net.Http.dll
7.0K System.Net.Primitives.dll
 11K System.ObjectModel.dll
1.1M System.Private.CoreLib.dll
 21K System.Private.Runtime.InteropServices.JavaScript.dll
 64K System.Private.Uri.dll
- 4.5K System.Resources.ResourceManager.dll
3.0K System.Runtime.CompilerServices.Unsafe.dll
- 4.5K System.Runtime.Extensions.dll
- 8.5K System.Runtime.dll
 23K System.Text.Encodings.Web.dll
235K System.Text.Json.dll
- 4.5K System.Threading.Tasks.dll
- 4.5K System.Threading.ThreadPool.dll
- 12K System.Threading.dll
 17K foo.dll
+ 22K foo.pdb (before: 13K)
```

src/linker/Linker/MarkingHelpers.cs
mateoatr: Per offline conversation; whole block removed ;)

sbomer:
- The size decreases can only be caused by us no longer marking all exports that point to marked types, right?
- I am wondering what is pulling in the extra files given that blazor trims most of the aspnet assemblies:
https://github.com/dotnet/sdk/blob/598ace99ef742e89f60b2fd32572162d4ca63bf7/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.Current.targets#L443-L445. Maybe these facades could be removed by tweaking some linker settings like setting `IsTrimmable` for some more assemblies. Would you mind sharing the dependency trace with your changes?
- We should check that the size increases look expected, for example for Microsoft.JSInterop.dll. Some of them may again be cases where we ought to consider adding `IsTrimmable` if we don't already.

sbomer (approved):
LGTM. Thanks for the great work! Please hold off on merging until we figure out what to do with the size diff.

mateoatr:
> The size decreases can only be caused by us no longer marking all exports that point to marked types, right?

Yes. I'd expect the decrease in size to be due to the marking we used to do in the while-loop in `MarkStep.Process()`.

> We should check that the size increases look expected, for example for Microsoft.JSInterop.dll. Some of them may again be cases where we ought to consider adding IsTrimmable if we don't already.

Agreed. I'll take a look at this.


[dependencies-dump.zip](https://github.com/mono/linker/files/6207929/dependencies-dump.zip)

mateoatr:
#### Microsoft.AspNetCore.Authorization:

A descriptor file is created by the SDK: https://github.com/dotnet/sdk/blob/446589256a6fc128d1adffe83551a727ebf7f234/src/BlazorWasmSdk/Targets/Microsoft.NET.Sdk.BlazorWebAssembly.Current.targets#L430
This keeps all types in the above assembly, thus marking the exported types and keeping the whole assembly:

```XML
<assembly fullname="Microsoft.AspNetCore.Authorization">
    <type fullname="*" required="false" preserve="all" />
  </assembly>
```

#### Other new assemblies:

There are three assemblies which are passed to the linker with the copy action whenever publishing a blazorwasm app, these are: `Microsoft.JSInterop`, `Microsoft.JSInterop.WebAssembly`, and `System.IO.Pipelines`; they all have various static type references to exported types declared in the newly seen assemblies (System.Buffers, System.Diagnostics.Debug, System.Resources.ResourceManager, System.Runtime.Extensions, System.Runtime, System.Threading.Tasks, System.Threading, System.Threading.ThreadPool).

In order to avoid linker from adding these new assemblies to the output, we should investigate whether it's possible to mark the the three copied assemblies with `IsTrimmable`.

marek-safar:
> This keeps all types in the above assembly, thus marking the exported types and keeping the whole assembly:

You should be able to tweak the descriptor to something like

```xml
<assembly fullname="Microsoft.AspNetCore.Authorization">
    <type fullname="Microsoft.AspNetCore.*" required="false" preserve="all" />
  </assembly>
```

> There are three assemblies which are passed to the linker with the copy action whenever publishing a blazorwasm app, these are: Microsoft.JSInterop, Microsoft.JSInterop.WebAssembly, and System.IO.Pipelines; 

That's odd. All these assemblies should have explicit metadata IsTrimmable set to true. Are you running a recent version?

src/linker/Linker.Steps/OutputStep.cs
```diff
@@ -271,42 +271,27 @@ FileInfo GetOriginalAssemblyFileInfo (AssemblyDefinition assembly)
...
-				WriteAssembly (assembly, directory, new WriterParameters ());
-				return;
-			}
-
 			FileInfo fi = GetOriginalAssemblyFileInfo (assembly);
 			string target = Path.GetFullPath (Path.Combine (directory, fi.Name));
 			string source = fi.FullName;
+
+			if (assembly.MainModule.HasSymbols && !Context.LinkSymbols && assembly.MainModule.SymbolReader is EmbeddedPortablePdbReader)
+				Context.LogWarning ("Debug symbols cannot be modified in a copied assembly.", 2104, new MessageOrigin (source));
```
marek-safar, line 279: Why is it a warning?

src/linker/Linker.Steps/OutputStep.cs
```diff
@@ -271,42 +271,23 @@ FileInfo GetOriginalAssemblyFileInfo (AssemblyDefinition assembly)
 
 		protected virtual void CopyAssembly (AssemblyDefinition assembly, string directory)
 		{
-			// Special case.  When an assembly has embedded pdbs, link symbols is not enabled, and the assembly's action is copy,
```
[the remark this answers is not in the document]
marek-safar: I'm not sure about the warning. What "ask us to remove symbols for a "copy" assembly," should do? The assembly is copied so the symbols are same. What am I missing?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -367,41 +368,7 @@ private void MarkEntireTypeInternal (TypeDefinition type, bool includeBaseTypes,
...
-							continue;
-						if (!Annotations.IsMarked (type))
-							continue;
-						var di = new DependencyInfo (DependencyKind.ExportedType, type);
-						_context.MarkingHelpers.MarkExportedType (exported, assembly.MainModule, di);
-					}
-				}
-			}
-
+			while (ProcessPrimaryQueue () || ProcessMarkedPending () || ProcessLazyAttributes () || ProcessLateMarkedAttributes () || MarkFullyPreservedAssemblies () || ProcessInternalsVisibleAttributes ()) ;
```
marek-safar, line 371: nit: split it to one call per line

marek-safar (changes requested):
external/cecil changes need to be reverted
You are adding redundant formatDiagnosticLog.binlog

src/linker/Linker/ExportedTypeExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Mono.Linker
+{
+	public static class ExportedTypeExtensions
+	{
+		public static TypeReference AsTypeReference (this ExportedType exportedType, ModuleDefinition module)
+		{
+			string exportedTypeFullName = exportedType.FullName;
+			int idx = exportedTypeFullName.LastIndexOf ('.');
+			(string typeNamespace, string typeName) = idx > 0 ? (exportedTypeFullName.Substring (0, idx), exportedTypeFullName[(idx + 1)..]) :
+				(string.Empty, exportedTypeFullName);
```
marek-safar, line 16: Why do you need that when Exported type has Namespace and Name properties?

src/linker/Linker/ExportedTypeExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Mono.Cecil;
+
+namespace Mono.Linker
+{
+	public static class ExportedTypeExtensions
+	{
+		public static TypeReference AsTypeReference (this ExportedType exportedType, ModuleDefinition module)
```
marek-safar, line 11: nit: As* to me indicates that it's "free" operation but this one isn't.

src/linker/Linker/MarkingHelpers.cs
```diff
@@ -11,12 +11,33 @@ public MarkingHelpers (LinkContext context)
 			_context = context;
 		}
 
-		public void MarkExportedType (ExportedType type, ModuleDefinition module, in DependencyInfo reason)
+		public void MarkMatchingExportedType (TypeDefinition typeToMatch, AssemblyDefinition assembly, in DependencyInfo reason)
```
marek-safar, line 14: How is this working when there are multiple typeforwarders in the chain?

src/linker/Linker.Steps/OutputStep.cs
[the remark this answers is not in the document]
sbomer: The option `TrimmerRemoveSymbols` (which maps to `-b`) will drop all symbols from the inputs, including for `copy` assemblies when the symbols are in a separate pdb file (and before this change, also for embedded symbols).

I agree that it only makes sense to keep the embedded symbols for "copy" - the warning just signals that a dev who wants to drop embedded PDBs needs to make sure they "link" the assembly.

mateoatr:
Re measured size impacts using latest SDK preview. I can see that the descriptor file that caused linker to keep Microsoft.AspNetCore.Authorization because of some exported types declared in it is no longer included (see [here]()). Impact now is much less noticeable. I'm adding the size differences bellow using the same color-codes as before, size are in bytes:

Total size: 2252 KB (used to be 2228 KB).

```diff
   17408 Microsoft.AspNetCore.Components.Web.dll
   48128 Microsoft.AspNetCore.Components.WebAssembly.dll
  121856 Microsoft.AspNetCore.Components.dll
    6144 Microsoft.Extensions.Configuration.Abstractions.dll
    8704 Microsoft.Extensions.Configuration.Json.dll
    7680 Microsoft.Extensions.Configuration.dll
   12800 Microsoft.Extensions.DependencyInjection.Abstractions.dll
   38400 Microsoft.Extensions.DependencyInjection.dll
   24064 Microsoft.Extensions.Logging.Abstractions.dll
   17408 Microsoft.Extensions.Logging.dll
   14848 Microsoft.Extensions.Options.dll
    8704 Microsoft.Extensions.Primitives.dll
    8704 Microsoft.JSInterop.WebAssembly.dll
   30720 Microsoft.JSInterop.dll
   17408 System.Collections.Concurrent.dll
   66560 System.Collections.Immutable.dll
!  22016 System.Collections.dll (before: 22528)
    4608 System.ComponentModel.dll
   13824 System.Console.dll
   25088 System.Linq.dll
!  17920 System.Memory.dll (before: 18432)
   10752 System.Net.Http.Json.dll
  138752 System.Net.Http.dll
    7168 System.Net.Primitives.dll
    7680 System.ObjectModel.dll
+1121280 System.Private.CoreLib.dll (before: 1117184)
   21504 System.Private.Runtime.InteropServices.JavaScript.dll
+  66048 System.Private.Uri.dll (before: 65536)
    7168 System.Runtime.CompilerServices.Unsafe.dll
-   5632 System.Runtime.dll
   29696 System.Text.Encodings.Web.dll
  239104 System.Text.Json.dll
   17408 foo.dll
+  21860 foo.pdb (before: 13032)
```

src/linker/Linker.Steps/OutputStep.cs
mateoatr: We used to allow symbols to be removed from copy assemblies; this would inform users that were doing this that the scenario is no longer supported. Although I think it's a fair point that it doesn't make sense to ask for removal in something tagged with "copy" -- I'll remove warning here.

src/linker/Linker/MarkingHelpers.cs
mateoatr: All copy assemblies will go through the `MarkEntireAssembly` method, which marks all exported types and the forwarded scope of every type reference (in particular, this will mark the scope of exported types -- which is the assembly to which the exported type ``points to''); this is enough for keeping all necessary forwarders in a chain.

e.g., if we have the following chain:
```
[copy] A -> [copy] B -> [link] C -> [link] D -> [implementation] E
```

and a static typeref to `A`, the logic would mark `A` and `B` (since these are exported types in a copy assembly), `C` would be marked by `MarkForwardedScope(B)`, `D` would be trimmed and `E` would be marked since it's the type that `C` resolves to:

https://github.com/mono/linker/blob/ea3525008ab1496e304d90476e543565f2cc522e/src/linker/Linker.Steps/MarkStep.cs#L1344-L1350

eerhardt:
@mateoatr - what's the size difference in the `.br` compressed files? If it is ~30 KB uncompressed, I'd assume this change makes it ~10 KB .br compressed. That's still a pretty large size regression for Blazor WASM...

vitek-karas:
I cloned this branch and tried it on 6.0 Preview 4 Blazor temp app.
The difference compressed is only +906 bytes. Note that with these small numbers the uncompressed size is deceiving. The PE file sections are basically 512 aligned. So some changes are not visible and other changes are exaggerated. For example the app.dll uncompressed size is exactly the same before/after, but compressed it's +114 bytes - something changed in the file which is harder to compress. On the other hand corelib is +4096 uncompressed, but only +309 bytes compressed.

eerhardt:
> The difference compressed is only +906 bytes.

That's good enough for me. Less than 1KB for more correctness seems like a fair trade.

vitek-karas:
Sorry - my bad - I missed that we've added a new file :-(
The difference is +2811 bytes which is borderline to me - it's not completely insignificant.

marek-safar:
As this is also a correctness fix I'm fine with +2k increase for Blazor and we'll save it elsewhere if needed.

PR #1879, opened 2021-03-10
Fix lint to avoid generating binlog file

vitek-karas:
Latest lint writes formatDiagnosticLog.binlong next to the solution file when running with --verbosity diagnostic. There's no way to redirect or disable this - https://github.com/dotnet/format/issues/1041.

So this change disables --verbosity diagnostic for local runs of lint and adds it only to the CI (where the machine will clean the enlistment anyway).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,9 +1,7 @@
 # top-most EditorConfig file
 root = true
-end_of_line = lf
 
 [*.cs]
-end_of_line = lf
 indent_style = tab
 indent_size = 4
 csharp_new_line_before_open_brace = types,methods
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,9 @@
 # top-most EditorConfig file
 root = true
+end_of_line = lf
 
 [*.cs]
+end_of_line = lf
 indent_style = tab
 indent_size = 4
 csharp_new_line_before_open_brace = types,methods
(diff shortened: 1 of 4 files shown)


MichalStrehovsky:
> Change #1867 breaks running lint on Windows since it will reformat pretty much all the source files in the repo (change of newline). So I reverted that change 74fb9a9.

I think this is because all files you have on disk are still checked out with LF line endings, despite the config change. There is probably some git command to fix that, but deleting your entire local clone and cloning again will fix that for sure.

I was hitting problems with the LF line endings too and I would have done Andy's change myself, but he pre-empted me (I was already asking about this in the pull request that introduced it https://github.com/mono/linker/pull/20#issuecomment-780509770). I was happy to see it changed. Going back to LF would be a step back.

vitek-karas:
I'm changing this PR to only fix the binlog - the newlines is still broken for me.

Tried a new clone of the repo - lint still edits almost 600 files. I don't understand how it's supposed to work - to me there's a mismatch between GIT and lint/editorconfig.
* GIT is still set to LF: https://github.com/mono/linker/blob/main/.gitattributes
* .editorconfig is now set to CRLF on Windows - and so they fight each other I think

MichalStrehovsky:
Yes, I got confused - the pull request I'm referencing in https://github.com/mono/linker/pull/20#issuecomment-780509770 is a change to gitattributes. I didn't realized we also have this setting in editorconfig and those two formats look too similar. I thought Andy changed gitattributes. We should change that one too to match.

marek-safar (approved):

Issue #1814, opened 2021-02-09
Intrinsically handle Expression.Property(Expression,MethodInfo)


eerhardt:
With the annotations applied in https://github.com/dotnet/runtime/pull/47938, `Expression.Property(Expression,MethodInfo)` is being marked as `RequiresUnreferencedCode`. This is because the linker doesn't see the Property being used - only the getter (or setter). See discussion in https://github.com/dotnet/runtime/pull/47938#discussion_r571314108. Thus, the linker may trim the property metadata, or the other side of the property accessor.

Because of this decision, users will now get an ILLink warning for the following code:

```C#
IQueryable<Customer> customers = GetCustomers();
foreach (var cust in customers.Where(c => c.Name == "Bob"))
{
}
```

```
C:\DotNetTest\Net6Hello\Program.cs(10,34): Trim analysis warning IL2026: Net6Hello.Program.Main(String[]): Calling 'System.Linq.Expressions.Expression.Property(Expression,MethodInfo)' which has `RequiresUnreferencedCodeAttribute` can break functionality when trimming application code. The Property metadata or other accessor may be trimmed.. [C:\DotNetTest\Net6Hello\Net6Hello.csproj]
```

The IL generated by the C# compiler to handle the above code looks like the following:

```
	𝐈𝐋_𝟎𝟎𝟑𝟑: 𝐥𝐝𝐭𝐨𝐤𝐞𝐧 𝐦𝐞𝐭𝐡𝐨𝐝 𝐢𝐧𝐬𝐭𝐚𝐧𝐜𝐞 𝐬𝐭𝐫𝐢𝐧𝐠 𝐂𝐮𝐬𝐭𝐨𝐦𝐞𝐫::𝐠𝐞𝐭_𝐍𝐚𝐦𝐞()
[...]

Issue #1727, opened 2021-01-06
MakeGenericMethod shouldn't cause IL2060 warning if the specified method doesn't have ILLinker annotations


eerhardt:
The following code in [`System.Text.RegularExpressions.RegexCompiler`](https://github.com/dotnet/runtime/blob/4486164506a56200e66cf4cd750fde6b54ab5abd/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexCompiler.cs#L59-L62) is causing IL2060 warnings:

```C#
private static readonly MethodInfo s_memoryMarshalGetReference = typeof(MemoryMarshal).GetMethod("GetReference", new Type[] { typeof(ReadOnlySpan<>).MakeGenericType(Type.MakeGenericMethodParameter(0)) })!.MakeGenericMethod(typeof(char));
private static readonly MethodInfo s_spanIndexOf = typeof(MemoryExtensions).GetMethod("IndexOf", new Type[] { typeof(ReadOnlySpan<>).MakeGenericType(Type.MakeGenericMethodParameter(0)), Type.MakeGenericMethodParameter(0) })!.MakeGenericMethod(typeof(char));
private static readonly MethodInfo s_spanIndexOfAnyCharChar = typeof(MemoryExtensions).GetMethod("IndexOfAny", new Type[] { typeof(ReadOnlySpan<>).MakeGenericType(Type.MakeGenericMethodParameter(0)), Type.MakeGenericMethodParameter(0), Type.MakeGenericMethodParameter(0) })!.MakeGenericMethod(typeof(char));
private static readonly MethodInfo s_spanIndexOfAnyCharCharChar = typeof(MemoryExtensions).GetMethod("IndexOfAny", new Type[] { typeof(ReadOnlySpan<>).MakeGenericType(Type.MakeGenericMethodParameter(0)), Type.MakeGenericMethodParameter(0), Type.MakeGenericMethodParameter(0), Type.MakeGenericMethodParameter(0) })!.MakeGenericMethod(typeof(char));
```

```
RegexCompiler.cs(59,9): Trim analysis warning IL2060: System.Text.RegularExpressions.RegexCompiler..cctor(): Call to `System.Reflection.MethodInfo.MakeGenericMethod(Type[])` can not be statically analyzed. It's not possible to guarantee the availability of requirements of the generic method. [C:\git\runtime\src\libraries\src.proj]
[...]

PR #1880, opened 2021-03-10
Basic tracking of MethodInfo and use it to improve Expression.Property and MakeGenericMethod

vitek-karas:
Adds the ability to track some basic patterns around `MethodBase`/`MethodInfo`. Linker now recognizes `ldtoken` for a method as well as tracks return value of `Type.GetMethod`.

With this, the change implements two improvements:

#1814 - Mark property as accessed via reflection when using Expression.Property(Expression,MethodInfo)

Added a new warning to report cases where the intrinsic handling of the Property call doesn't work.

#1727 - Ability to handle simple MakeGenericMethod calls without generating a warning

If the generic method has no annotations on the generic arguments, linker will no longer warn.

Added tests for all the new behaviors.
Improved tests for MakeGenericMethod and MakeGenericType.
Implement checking for new constraint in MakeGenericType - it now behaves the same as if the generic argument is annotated.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -67,9 +67,6 @@ namespace Mono.Linker.Dataflow
 
 		public void ScanAndProcessReturnValue (MethodBody methodBody)
 		{
-			if (methodBody.Method.Name == "GetCSharpThunk")
-				Debug.WriteLine ("");
-
 			Scan (methodBody);
 
 			if (methodBody.Method.ReturnType.MetadataType != MetadataType.Void) {
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1665,7 +1665,7 @@ namespace Mono.Linker.Dataflow
 			}
 		}
 
-		void ProcessGetMethodByName(
+		void ProcessGetMethodByName (
 			ref ReflectionPatternContext reflectionContext,
 			TypeDefinition typeDefinition,
 			string methodName,
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -67,6 +67,9 @@ namespace Mono.Linker.Dataflow
 
 		public void ScanAndProcessReturnValue (MethodBody methodBody)
 		{
+			if (methodBody.Method.Name == "GetCSharpThunk")
+				Debug.WriteLine ("");
+
 			Scan (methodBody);
 
 			if (methodBody.Method.ReturnType.MetadataType != MetadataType.Void) {
@@ -224,6 +227,7 @@ namespace Mono.Linker.Dataflow
 			TypeDelegator_Ctor,
 			Array_Empty,
 			TypeInfo_AsType,
+			MethodBase_GetMethodFromHandle,
 
 			// Anything above this marker will require the method to be run through
 			// the reflection body scanner.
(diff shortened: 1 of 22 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -67,6 +67,9 @@ public ReflectionMethodBodyScanner (LinkContext context, MarkStep parent)
 
 		public void ScanAndProcessReturnValue (MethodBody methodBody)
 		{
+			if (methodBody.Method.Name == "GetCSharpThunk")
+				Debug.WriteLine ("");
```
mateoatr, line 71: I guess this wasn't left here intentionally?

vitek-karas: Removed

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -779,6 +803,39 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+				// System.Linq.Expressions.Expression
+				// 
+				// static Property (Expression, MethodInfo)
+				//
+				case IntrinsicId.Expression_Property when calledMethod.HasParameterOfType (1, "System.Reflection", "MethodInfo"): {
+						reflectionContext.AnalyzingPattern ();
+
+						foreach (var value in methodParams[1].UniqueValues ()) {
+							if (value is SystemReflectionMethodBaseValue methodBaseValue) {
+								// We have one of the accessor for the property. The Expression.Property will in this case search
```
MichalStrehovsky, line 816:
```suggestion
								// We have one of the accessors for the property. The Expression.Property will in this case search
```

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -30,6 +30,9 @@ public enum ValueNodeKind
 		MethodParameter,                // symbolic placeholder
 		MethodReturn,                   // symbolic placeholder
 
+		RuntimeMethodHandle,            // known value - MethodRepresented
```
MichalStrehovsky, line 33: Do we need to add these to the various `switch` blocks below?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -993,10 +1048,24 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
-										if (BindingFlagsAreUnsupported (bindingFlags))
+										if (BindingFlagsAreUnsupported (bindingFlags)) {
 											RequireDynamicallyAccessedMembers (ref reflectionContext, DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods, value, calledMethodDefinition);
-										else
-											MarkMethodsOnTypeHierarchy (ref reflectionContext, systemTypeValue.TypeRepresented, m => m.Name == stringValue.Contents, bindingFlags);
+										} else {
+											bool foundAny = false;
+											foreach (var method in systemTypeValue.TypeRepresented.GetMethodsOnTypeHierarchy (m => m.Name == stringValue.Contents, bindingFlags)) {
+												MarkMethod (ref reflectionContext, method);
+												methodReturnValue = MergePointValue.MergeValues (methodReturnValue, new SystemReflectionMethodBaseValue (method));
```
MichalStrehovsky, line 1057: Would it make sense to add this handling to RuntimeReflectionExtensions_GetRuntimeMethod as well?

MichalStrehovsky (approved):

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: Fixed

src/linker/Linker.Dataflow/ValueNode.cs
vitek-karas: I only found one such block - added it there.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: Good point - I completely missed that one. Unified the handling and added some tests.

marek-safar (approved):

PR #1885, opened 2021-03-11
Change line endings to auto

vitek-karas:
This fixes the problem on Windows where files are checked out as LF, but formatting/lint will enforce CRLF.

To apply to an existing clone of the repo run from the repo root. There should be no un-commited local changes, it should work find with locally commited changes.

```console
git rm -r :/
git checkout HEAD -- :/
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,4 @@
 # Autodetect text files
 * text=auto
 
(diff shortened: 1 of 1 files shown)


MichalStrehovsky (approved):
Thank you!

PR #1889, opened 2021-03-12
Warn on cctors annotated with RUC if a field triggers its marking

mateoatr:
Whenever a static member is referenced, static constructors on the declared type are called. If the cctor is annotated with `RequiresUnreferencedCode`, this should warn.

### DIFF
--- a/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
@@ -22,6 +22,7 @@ namespace ILLink.RoslynAnalyzer.Tests
 			case "TestCovariantReturnCallOnDerived":
 			case "TestRequiresUnreferencedCodeOnlyThroughReflection":
 			case "TestStaticCctorRequiresUnreferencedCode":
+			case "TestStaticCtorMarkingIsTriggeredByFieldAccess":
 			case "TestTypeWhichOverridesMethodVirtualMethodRequiresUnreferencedCode":
 			case "TestRequiresInMethodFromCopiedAssembly":
 			case "TestRequiresThroughReflectionInMethodFromCopiedAssembly":
--- a/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
@@ -26,7 +26,7 @@ namespace ILLink.RoslynAnalyzer.Tests
 			case "TestTypeWhichOverridesMethodVirtualMethodRequiresUnreferencedCode":
 			case "TestRequiresInMethodFromCopiedAssembly":
 			case "TestRequiresThroughReflectionInMethodFromCopiedAssembly":
-			case "TestStaticCctorTriggeredByMethodCall":
+			case "TestStaticCtorTriggeredByMethodCall":
 			case "TestTypeIsBeforeFieldInit":
 				return;
 			}
--- a/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
@@ -27,6 +27,7 @@ namespace ILLink.RoslynAnalyzer.Tests
 			case "TestRequiresInMethodFromCopiedAssembly":
 			case "TestRequiresThroughReflectionInMethodFromCopiedAssembly":
 			case "TestStaticCctorTriggeredByMethodCall":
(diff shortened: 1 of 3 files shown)


test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -234,6 +235,22 @@ static void TestStaticCctorRequiresUnreferencedCode ()
...
 		}
 
+		class StaticCtorTriggeredByFieldAccess
+		{
+			[RequiresUnreferencedCode ("Message for --StaticCtorTriggeredByFieldAccess.Cctor--")]
+			static StaticCtorTriggeredByFieldAccess ()
+			{
+			}
+
+			public static int field = 0;
```
marek-safar, line 245: I think it'd be better not to use static fields to check that types without `beforefieldinit` modifiers work as expected.

vitek-karas: I think moving the field initialization into the static `.cctor` should be enough, right?

marek-safar: I think the test coverage should be better. Does the code print the warning in this case?

```c#
new C ().Foo ();

class C
{
	[RequiresUnreferencedCode ("Message")]
    static C ()
    {
    }
    
    public void Foo ()
    {
    }
}
```

sbomer: Also, AFAIK Roslyn won't make the type beforefieldinit in any of the above cases - the explicit static ctor in C# doesn't have the relaxed semantics allowed by beforefieldinit. To test with beforefieldinit you need to avoid writing a static ctor and let Roslyn generate the `.cctor` from the static field initializers:
```csharp
class C {
    static int field = RUC();
    [RequiresUnreferencedCode]
    static int RUC() => 0;
}
```

test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
```diff
@@ -22,9 +22,12 @@ public void RequiresCapability (MethodDeclarationSyntax m, List<AttributeSyntax>
 			case "TestCovariantReturnCallOnDerived":
 			case "TestRequiresUnreferencedCodeOnlyThroughReflection":
 			case "TestStaticCctorRequiresUnreferencedCode":
+			case "TestStaticCtorMarkingIsTriggeredByFieldAccess":
 			case "TestTypeWhichOverridesMethodVirtualMethodRequiresUnreferencedCode":
 			case "TestRequiresInMethodFromCopiedAssembly":
 			case "TestRequiresThroughReflectionInMethodFromCopiedAssembly":
+			case "TestStaticCctorTriggeredByMethodCall":
```
tlakollo, line 29: Extra c in the name "TestStaticCctorTriggeredByMethodCall" vs "TestStaticCtorTriggeredByMethodCall"

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -234,6 +237,60 @@ static void TestStaticCctorRequiresUnreferencedCode ()
...
+
+		class TypeIsBeforeFieldInit
+		{
+			public static int field = AnnotatedMethod ();
+
+			[RequiresUnreferencedCode ("Message from --TypeIsBeforeFieldInit.AnnotatedMethod--")]
+			public static int AnnotatedMethod () => 42;
+		}
+
+		[LogContains ("IL2026: Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.TypeIsBeforeFieldInit..cctor():" +
```
tlakollo, line 265: Why do we use LogContains here?

mateoatr: As @sbomer puts above, Roslyn generates a .cctor and annotates the type with `beforefieldinit`, this static constructor initializes the field and thus calls `AnnotatedMethod`, triggering the warning from within the .cctor. Because of this, there's no place where we could put the `ExpectedWarning` attribute (the call is made from generated code).

marek-safar (approved):

vitek-karas (approved):

PR #1892, opened 2021-03-15
Suppress warning if UnconditionalSuppressMessageAttribute is used

agocke:
If/when Roslyn ships a universal suppress behavior here we can remove this code.


src/ILLink.RoslynAnalyzer/AnalyzerUtils.cs
```diff
@@ -0,0 +1,81 @@
...
+			}
+
+			return true;
+		}
+
+		const string SuppressMessageFqn = "System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessageAttribute";
+
+		internal static bool IsDiagnosticSuppressed(OperationAnalysisContext operationContext, string diagnosticId)
+		{
+			var attributes = operationContext.ContainingSymbol.GetAttributes();
```
mateoatr, line 39: Will this work when the suppression is up in the tree?
e.g.
```C#
[UnconditionalSuppressMessage ("", "IL1234:MyWarning")]
class Foo {
   void Bar () {
     // Triggers warning IL1234
  }
}
```

mateoatr (commented):
What about unconditional suppressions that use the target property?

src/ILLink.RoslynAnalyzer/AnalyzerUtils.cs
agocke: Right, this only handles local suppression. Global suppression support will have to be added in another PR.

agocke:
As far as I can tell, the Target property is only relevant on attributes higher than the member level.

agocke:
I'm gonna close this for now, since it looks like the Roslyn change is gonna go through.

tlakollo:
Closing in favor of https://github.com/dotnet/roslyn/pull/51792

PR #1897, opened 2021-03-15
Analyzer can warn on more RequiresUnreferencedCode cases

tlakollo:
Add support in analyzer to detect RUC in overridden methods/properties
Add support in analyzer to detect RUC in static constructors
Add test for overridden property

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
@@ -76,7 +76,7 @@ namespace ILLink.RoslynAnalyzer
 					ImmutableArray<IMethodSymbol> constructors)
 				{
 					foreach (var constructor in constructors) {
-						if (constructor.Parameters.Length == 0 && constructor.HasAttribute (RequiresUnreferencedCodeAttribute)) {
+						if (constructor.Parameters.Length == 0 && constructor.HasAttribute (RequiresUnreferencedCodeAttribute) && constructor.MethodKind == MethodKind.StaticConstructor) {
 							if (constructor.TryGetAttributeWithMessageOnCtor (FullyQualifiedRequiresUnreferencedCodeAttribute, out AttributeData? requiresUnreferencedCode)) {
 								operationContext.ReportDiagnostic (Diagnostic.Create (
 									s_requiresUnreferencedCodeRule,
@@ -98,17 +98,16 @@ namespace ILLink.RoslynAnalyzer
 						operationContext.ContainingSymbol.HasAttribute (RequiresUnreferencedCodeAttribute))
 						return;
 
+					// If calling an instance constructor, check first for any static constructor since it will be called implicitly
 					if (method.ContainingType is { } containingType && operationContext.Operation is IObjectCreationOperation)
 						CheckStaticConstructors (operationContext, containingType.StaticConstructors);
 
 					if (!method.HasAttribute (RequiresUnreferencedCodeAttribute))
 						return;
 
-					if (method.OverriddenMethod != null) {
-						for (var t = method.OverriddenMethod; t is not null && SymbolEqualityComparer.Default.Equals (method.ReturnType, t.ReturnType); t = t.OverriddenMethod) {
-							method = t;
-						}
-					}
+					// Warn on the most derived base method taking into account covariant returns
+					while (method.OverriddenMethod != null && SymbolEqualityComparer.Default.Equals (method.ReturnType, method.OverriddenMethod.ReturnType))
+						method = method.OverriddenMethod;
 
 					if (method.TryGetAttributeWithMessageOnCtor (FullyQualifiedRequiresUnreferencedCodeAttribute, out AttributeData? requiresUnreferencedCode)) {
 						operationContext.ReportDiagnostic (Diagnostic.Create (
--- a/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
@@ -45,9 +45,6 @@ namespace ILLink.RoslynAnalyzer
 
 				context.RegisterOperationAction (operationContext => {
 					var call = (IInvocationOperation) operationContext.Operation;
-					if (call.IsVirtual && call.TargetMethod.OverriddenMethod != null)
-						return;
-
 					CheckMethodOrCtorCall (operationContext, call.TargetMethod);
 				}, OperationKind.Invocation);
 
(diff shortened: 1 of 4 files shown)


src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -67,6 +72,23 @@ public override void Initialize (AnalysisContext context)
 						CheckMethodOrCtorCall (operationContext, prop.SetMethod);
 				}, OperationKind.PropertyReference);
 
+				static void CheckStaticConstructors (OperationAnalysisContext operationContext,
+					ImmutableArray<IMethodSymbol> constructors)
+				{
+					foreach (var constructor in constructors) {
+						if (constructor.Parameters.Length == 0 && constructor.HasAttribute (RequiresUnreferencedCodeAttribute)) {
```
mateoatr, line 79: Why is this limited to parameterless cctors?

test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
```diff
@@ -19,14 +19,10 @@ public void RequiresCapability (MethodDeclarationSyntax m, List<AttributeSyntax>
...
 			case "MethodWithDuplicateRequiresAttribute":
-			case "TestCovariantReturnCallOnDerived":
 			case "TestRequiresUnreferencedCodeOnlyThroughReflection":
-			case "TestStaticCctorRequiresUnreferencedCode":
-			case "TestStaticCtorMarkingIsTriggeredByFieldAccess":
-			case "TestTypeWhichOverridesMethodVirtualMethodRequiresUnreferencedCode":
 			case "TestRequiresInMethodFromCopiedAssembly":
 			case "TestRequiresThroughReflectionInMethodFromCopiedAssembly":
-			case "TestStaticCtorTriggeredByMethodCall":
+			// The way roslyn and linker name the source location is different for BeforeFieldInit
```
mateoatr, line 25: I think I don't understand this comment

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -220,6 +221,27 @@ static void TestTypeWhichOverridesMethodVirtualMethodRequiresUnreferencedCodeOnB
...
+		{
+			public override int VirtualPropertyRequiresUnreferencedCode {
+				[RequiresUnreferencedCode ("Message for --TypeWhichOverridesProperty.VirtualPropertyRequiresUnreferencedCode--")]
+				get { return 1; }
+			}
+		}
+
+		[LogDoesNotContain ("TypeWhichOverridesProperty.VirtualPropertyRequiresUnreferencedCode")]
+		[ExpectedWarning ("IL2026", "--PropertyBaseType.VirtualPropertyRequiresUnreferencedCode--")]
+		static void TestTypeWhichOverridesVirtualPropertyRequiresUnreferencedCode ()
```
mateoatr, line 239: Is this any different from the `TestTypeWhichOverridesMethodVirtualMethodRequiresUnreferencedCode` test that we already have?

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -76,9 +98,18 @@ public override void Initialize (AnalysisContext context)
...
 						return;
 
+					if (method.ContainingType is { } containingType && operationContext.Operation is IObjectCreationOperation)
+						CheckStaticConstructors (operationContext, containingType.StaticConstructors);
+
 					if (!method.HasAttribute (RequiresUnreferencedCodeAttribute))
 						return;
 
+					if (method.OverriddenMethod != null) {
+						for (var t = method.OverriddenMethod; t is not null && SymbolEqualityComparer.Default.Equals (method.ReturnType, t.ReturnType); t = t.OverriddenMethod) {
```
mateoatr, line 108: nit: Just a matter of style but wouldn't this be clearer if you used a while-loop instead? I suppose this is to support covariant returns, maybe it'd be good to have a comment.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
tlakollo: This new test does not add a lot of value to the linker since it treats the properties as methods. The analyzer, on the other hand, registers different operation actions depending on if it's a method invocation or a property reference. Therefore the analyzer could handle different properties and methods. For example, the code before this change will warn incorrectly on derived properties, but wouldn't warn on a method invocation.
I think it's important to test it on the side of the analyzer and it doesn't hurt to have it in the linker.

test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
tlakollo: There is a discrepancy between the way the linker and roslyn represent the source location for this warning. Roslyn will use (8,29) and the linker will print Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.TypeIsBeforeFieldInit..cctor()

**Roslyn warning**
(8,29): warning IL2026: Using method 'C.TypeIsBeforeFieldInit.AnnotatedMethod()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Message from --TypeIsBeforeFieldInit.AnnotatedMethod--.

**Linker warning**
IL2026: Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.TypeIsBeforeFieldInit..cctor(): Using method 'Mono.Linker.Tests.Cases.RequiresCapability.RequiresUnreferencedCodeCapability.TypeIsBeforeFieldInit.AnnotatedMethod()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Message from --TypeIsBeforeFieldInit.AnnotatedMethod--.

Since linker is using LogContains the Roslyn test will try to match the `TypeIsBeforeFieldInit..cctor()` value against his diagnostic value and fails. The scenario is supported, therefore the addition of the test in the RequiresUnreferencedCodeAnalyzerTests.cs file. But you are right because I try to keep the comment short might not be understandable. I will update it.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
tlakollo: I guess I only reuse one pattern I saw somewhere else in the analyzer. But you are right using the while loop might be more understandable and actually would simplify the code

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
tlakollo: Static constructors in C# cannot have access modifiers or have parameters
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors

mateoatr: Wouldn't a normal parameterless constructor satisfy this condition as well? Seems like it's working because we always pass `INamedTypeSymbol.StaticConstructors` (which is fine), shouldn't as well verify `MethodKind` or maybe that the md name is `cctor`?

PR #1900, opened 2021-03-16
Intrinsics for new overloads of GetConstructor and GetMethod

### LINKED ISSUE #1783: Add Intrinsic handling support for new GetConstructor and GetMethod overloads
https://github.com/dotnet/runtime/pull/44529 added two new overloads to `System.Type` to allow for easily referencing a specific method/ctor:

```diff
namespace System
{
    public class Type
    {
        public System.Reflection.MethodInfo? GetMethod(string name) { throw null; }
        public System.Reflection.MethodInfo? GetMethod(string name, int genericParameterCount, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder? binder, System.Reflection.CallingConventions callConvention, System.Type[] types, System.Reflection.ParameterModifier[]? modifiers) { throw null; }
        public System.Reflection.MethodInfo? GetMethod(string name, int genericParameterCount, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder? binder, System.Type[] types, System.Reflection.ParameterModifier[]? modifiers) { throw null; }
        public System.Reflection.MethodInfo? GetMethod(string name, int genericParameterCount, System.Type[] types) { throw null; }
        public System.Reflection.MethodInfo? GetMethod(string name, int genericParameterCount, System.Type[] types, System.Reflection.ParameterModifier[]? modifiers) { throw null; }
        public System.Reflection.MethodInfo? GetMethod(string name, System.Reflection.BindingFlags bindingAttr) { throw null; }
        public System.Reflection.MethodInfo? GetMethod(string name, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder? binder, System.Reflection.CallingConventions callConvention, System.Type[] types, System.Reflection.ParameterModifier[]? modifiers) { throw null; }
        public System.Reflection.MethodInfo? GetMethod(string name, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder? binder, System.Type[] types, System.Reflection.ParameterModifier[]? modifiers) { throw null; }
        public System.Reflection.MethodInfo? GetMethod(string name, System.Type[] types) { throw null; }
        public System.Reflection.MethodInfo? GetMethod(string name, System.Type[] types, System.Reflection.ParameterModifier[]? modifiers) { throw null; }

+       public System.Reflection.MethodInfo? GetMethod(string name, System.Reflection.BindingFlags bindingAttr, System.Type[] types) { throw null; }

        public System.Reflection.ConstructorInfo? GetConstructor(System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder? binder, System.Reflection.CallingConventions callConvention, System.Type[] types, System.Reflection.ParameterModifier[]? modifiers) { throw null; }
        public System.Reflection.ConstructorInfo? GetConstructor(System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder? binder, System.Type[] types, System.Reflection.ParameterModifier[]? modifiers) { throw null; }
        public System.Reflection.ConstructorInfo? GetConstructor(System.Type[] types) { throw null; }

+        public System.Reflection.ConstructorInfo? GetConstructor(System.Reflection.BindingFlags bindingAttr, System.Type[] types) { throw null; }
     }
```

We should update the linker's intrinsic reflection handling to identify these new overloads, like it does for the rest of the overloads:

https://github.com/mono/linker/blob/92b9b51bb8548320e40fc432efcb395156ad1445/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs#L855-L897

and

https://github.com/mono/linker/blob/92b9b51bb8548320e40fc432efcb395156ad1445/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs#L900-L943

cc @vitek-karas

vitek-karas:
In .NET 6 we've added new overloads for `Type.GetConstructor` and `Type.GetMethod`, linker has intrinsics for these so this change adds support for the new overloads.

I made another change as well. Previously calling `type.GetConstructor(Types.Empty)` would be recognized as special (empty array) and it would only require `PublicParameterlessConstructor` annotation on `type`.

But calling `typeof(TestType).GetConstructor(Types.Empty)` would still mark all public `.ctor`s on `TestType`. This is at least inconsistent.

This change also implements the same special logic for empty types array for known types - so in the above case only the public parameterless `.ctor` of `TestType` will be preserved.

This change rewrites the tests for `GetConstructor` and `GetMethod` - previously these randomly shared some of the test types and methods leading to potentially wrong validation of `[Kept]` attributes. The new version separates each test case into its own sub-class to avoid the case where one test case accidentally marks things another test case also uses. The diff is big, but in general I kept the exact same test functionality and only added the overloads.

Fixes #1783

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -353,6 +353,7 @@ namespace Mono.Linker.Dataflow
 					=> IntrinsicId.Type_GetType,
 
 				// System.Type.GetConstructor (Type[])
+				// System.Type.GetConstructor (BindingFlags, Type[])
 				// System.Type.GetConstructor (BindingFlags, Binder, Type[], ParameterModifier [])
 				// System.Type.GetConstructor (BindingFlags, Binder, CallingConventions, Type[], ParameterModifier [])
 				"GetConstructor" when calledMethod.IsDeclaredOnType ("System", "Type")
@@ -370,8 +371,9 @@ namespace Mono.Linker.Dataflow
 				// System.Type.GetMethod (string, BindingFlags)
 				// System.Type.GetMethod (string, Type[])
 				// System.Type.GetMethod (string, Type[], ParameterModifier[])
-				// System.Type.GetMethod (string, BindingFlags, Binder, Type[], ParameterModifier[]) 6
-				// System.Type.GetMethod (string, BindingFlags, Binder, CallingConventions, Type[], ParameterModifier[]) 7
+				// System.Type.GetMethod (string, BindingFlags, Binder, Type[])
+				// System.Type.GetMethod (string, BindingFlags, Binder, Type[], ParameterModifier[])
+				// System.Type.GetMethod (string, BindingFlags, Binder, CallingConventions, Type[], ParameterModifier[])
 				// System.Type.GetMethod (string, int, Type[])
 				// System.Type.GetMethod (string, int, Type[], ParameterModifier[]?)
 				// System.Type.GetMethod (string, int, BindingFlags, Binder?, Type[], ParameterModifier[]?)
@@ -983,6 +985,7 @@ namespace Mono.Linker.Dataflow
 
 				//
 				// GetConstructor (Type[])
+				// GetConstructor (BindingFlags, Type[])
 				// GetConstructor (BindingFlags, Binder, Type[], ParameterModifier [])
 				// GetConstructor (BindingFlags, Binder, CallingConventions, Type[], ParameterModifier [])
 				//
@@ -1030,6 +1033,7 @@ namespace Mono.Linker.Dataflow
 				// GetMethod (string, BindingFlags)
 				// GetMethod (string, Type[])
 				// GetMethod (string, Type[], ParameterModifier[])
+				// GetMethod (string, BindingFlags, Binder, Type[])
 				// GetMethod (string, BindingFlags, Binder, Type[], ParameterModifier[])
 				// GetMethod (string, BindingFlags, Binder, CallingConventions, Type[], ParameterModifier[])
 				// GetMethod (string, int, Type[])
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1002,6 +1002,7 @@ namespace Mono.Linker.Dataflow
 
 						int? ctorParameterCount = parameters.Count switch {
 							1 => (methodParams[1] as ArrayValue)?.Size.AsConstInt (),
+							2 => (methodParams[2] as ArrayValue)?.Size.AsConstInt (),
 							4 => (methodParams[3] as ArrayValue)?.Size.AsConstInt (),
 							5 => (methodParams[4] as ArrayValue)?.Size.AsConstInt (),
 							_ => null,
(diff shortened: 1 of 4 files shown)


test/Mono.Linker.Tests.Cases/Reflection/MethodUsedViaReflection.cs
```diff
@@ -484,10 +436,19 @@ public static int OnlyCalledViaReflection (int foo, int bar, int baz)
 			{
 				return 45;
 			}
+
+			[Kept]
+			[RecognizedReflectionAccessPattern]
+			public static void TestNameWithIntAndBindingFlags ()
+			{
+				var method = typeof (GetMethod_Name_GenericParameterCount_BindingAttr_Binder_Types_Modifiers).GetMethod ("OnlyCalledViaReflection", 1, BindingFlags.Public, null, new Type[] { }, null);
```
tlakollo, line 444: Can we add another BindingFlag to check that is actually using the BindingFlags?

test/Mono.Linker.Tests.Cases/Reflection/ConstructorUsedViaReflection.cs
```diff
@@ -10,226 +11,365 @@ public class ConstructorUsedViaReflection
...
-		static Type FindType ()
-		{
-			return null;
-		}
+			[Kept]
+			public static void Test ()
+			{
+				TestConstructorWithTypes_EmptyTypes ();
+				TestConstructorWithTypes_NonEmptyTypes ();
+				TestConstructorWithTypes_EmptyTypes_DataFlow (null);
```
tlakollo, line 56: Any specific reason you used null instead of a real type for dataflow tests?

tlakollo (approved):

MichalStrehovsky (approved):

test/Mono.Linker.Tests.Cases/Reflection/ConstructorUsedViaReflection.cs
vitek-karas: It doesn't matter - the test doesn't care what's passed in here. But I changed it so that if one day we actually execute the tests they don't crash.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1030,6 +1040,7 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 				// GetMethod (string, BindingFlags)
 				// GetMethod (string, Type[])
 				// GetMethod (string, Type[], ParameterModifier[])
+				// GetMethod (string, BindingFlags, Binder, Type[])
```
eerhardt, line 1043: (nit) this overload doesn't exist. The new one is

```
// GetMethod (string, BindingFlags, Type[])
```

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -370,8 +371,9 @@ static IntrinsicId GetIntrinsicIdForMethod (MethodDefinition calledMethod)
 				// System.Type.GetMethod (string, BindingFlags)
 				// System.Type.GetMethod (string, Type[])
 				// System.Type.GetMethod (string, Type[], ParameterModifier[])
-				// System.Type.GetMethod (string, BindingFlags, Binder, Type[], ParameterModifier[]) 6
-				// System.Type.GetMethod (string, BindingFlags, Binder, CallingConventions, Type[], ParameterModifier[]) 7
+				// System.Type.GetMethod (string, BindingFlags, Binder, Type[])
```
eerhardt, line 374: Same comment as below:

```
// System.Type.GetMethod (string, BindingFlags, Type[])
```

Issue #1758, opened 2021-01-15
Propagate certain annotations across Type.BaseType


vitek-karas:
If the type is annotated with `All` then the linker recursively applies this to all base types. As such linker could also annotate the return value of `Type.get_BaseType` with `All` to avoid unnecessary warnings.

Similarly for all `PublicMehotds`, `PublicFields`, and so on... those are also recursively applied to all base types.

Note that currently this is not true for `PrivateMethods` and similar, since in that case private members on base types are not marked (per reflection behavior of similar BindingFlags).

/cc @eerhardt @MichalStrehovsky

PR #1901, opened 2021-03-16
Propagate annotations over Type.BaseType

vitek-karas:
Some of the annotations apply with the same meaning on the base type - `All` and all of the `Public..`. It makes sense to propagate these over the call to `Type.BaseType`.

This change implements that. It also implements that if the input type is a known type the return value of `BaseType` is also know type - the actual base type of the type passed in.

Adds tests for both cases.

Fixes https://github.com/mono/linker/issues/1758.


test/Mono.Linker.Tests.Cases/DataFlow/TypeBaseTypeDataFlow.cs
```diff
@@ -0,0 +1,278 @@
...
+
+			TestCombinationOfPublicsIsPropagated (typeof (TestType));
+			TestCombinationOfNonPublicsIsNotPropagated (typeof (TestType));
+			TestCombinationOfPublicAndNonPublicsPropagatesPublicOnly (typeof (TestType));
+
+			TestNoAnnotation (typeof (TestType));
+			TestAnnotatedAndUnannotated (typeof (TestType), typeof (TestType), 0);
+			TestNull ();
+
+			Mixed_Derived.Test (typeof (TestType), 0);
```
mateoatr, line 46: Should this method be called with 1 and 2 as well?

mateoatr (approved):

test/Mono.Linker.Tests.Cases/DataFlow/TypeBaseTypeDataFlow.cs
vitek-karas: The tests actually never run the code, and since we don't do cross-method analysis the actual value here doesn't matter at all. (There's a plan to actually run the tests, but it still would not matter much).

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1169,6 +1175,59 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+				// Type.BaseType
+				//
+				case IntrinsicId.Type_get_BaseType: {
+						foreach (var value in methodParams[0].UniqueValues ()) {
+							if (value is LeafValueWithDynamicallyAccessedMemberNode dynamicallyAccessedMemberNode) {
+								DynamicallyAccessedMemberTypes propagatedMemberTypes = DynamicallyAccessedMemberTypes.None;
+								if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes == DynamicallyAccessedMemberTypes.All)
+									propagatedMemberTypes = DynamicallyAccessedMemberTypes.All;
+								else {
+									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicConstructors))
```
MichalStrehovsky, line 1188: I don't think we propagate constructors to base types.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1169,6 +1175,59 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicEvents))
+										propagatedMemberTypes |= DynamicallyAccessedMemberTypes.PublicEvents;
+
+									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicFields))
+										propagatedMemberTypes |= DynamicallyAccessedMemberTypes.PublicFields;
+
+									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicMethods))
+										propagatedMemberTypes |= DynamicallyAccessedMemberTypes.PublicMethods;
+
+									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicNestedTypes))
```
MichalStrehovsky, line 1200: I don't think we do nested types either.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1169,6 +1175,59 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+										propagatedMemberTypes |= DynamicallyAccessedMemberTypes.PublicNestedTypes;
+
+									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicParameterlessConstructor))
```
MichalStrehovsky, line 1203: Neither this one.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1169,6 +1175,59 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+										propagatedMemberTypes |= DynamicallyAccessedMemberTypes.PublicParameterlessConstructor;
+
+									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicProperties))
+										propagatedMemberTypes |= DynamicallyAccessedMemberTypes.PublicProperties;
+								}
+
+								if (propagatedMemberTypes != DynamicallyAccessedMemberTypes.None)
+									methodReturnValue = MergePointValue.MergeValues (methodReturnValue, new MethodReturnValue (calledMethod.MethodReturnType, propagatedMemberTypes));
+								else
+									methodReturnValue = MergePointValue.MergeValues (methodReturnValue, new MethodReturnValue (calledMethod.MethodReturnType, DynamicallyAccessedMemberTypes.None));
```
MichalStrehovsky, line 1213:
```suggestion
								methodReturnValue = MergePointValue.MergeValues (methodReturnValue, new MethodReturnValue (calledMethod.MethodReturnType, propagatedMemberTypes));
```

test/Mono.Linker.Tests.Cases/Reflection/TypeBaseTypeUseViaReflection.cs
```diff
@@ -0,0 +1,42 @@
...
+		}
+
+		[Kept]
+		[KeptBaseType (typeof (KnownType_Base))]
+		class KnownType_Derived : KnownType_Base
+		{
+			[Kept]
+			public static void Test ()
+			{
+				typeof (KnownType_Derived).BaseType.GetMethod ("UsedViaReflection", BindingFlags.NonPublic | BindingFlags.Static);
```
MichalStrehovsky, line 38: Nit: this test would also succeed if we implemented BaseType as "return this".

Maybe add one more test that e.g. looks for a constructor on the base type.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: Good point - and the ones below as well - fixed

test/Mono.Linker.Tests.Cases/Reflection/TypeBaseTypeUseViaReflection.cs
vitek-karas: This uses private reflection, so calling it on the derived type should not find the method on the base type.
I added the `GetConstructor` as well.

MichalStrehovsky: 🤦 I see! Thanks

MichalStrehovsky (approved):

PR #1904, opened 2021-03-19
Fix automatic propagation in property setters

### LINKED ISSUE #1902: Annotating property indexers causes the linker to treat the key (index) to have the annotations too.
Consider this code:

```c#
public class MyTypeDictionary
{
  [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
  public Type this[string key] { get; set; }
}
```

With the annotation slapped to the indexer, what you really want is for the getter return value, as well as for the setter param value to get annotated. In reality what happens is that the whole property gets annotated and so the linker will treat the setter as having two params `(string key, Type value)` both of which it will consider as annotated meaning we would need to preserve their constructors. Problem being that if you then have some consuming code like:

```c#
MyTypeDictionary dict = new MyTypeDictionary();
dict[myStringParam] = typeof(StringBuilder);
```

to throw a linker warning since the key parameter (the string) is not annotated. For now, the workaround is to annotate the getter and setter individually like:

```c#
public Type? this[string key]
{
    [return: DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
    get;
    [param: DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
    set;
}
```

The code above will produce the following IL:

```
.method public hidebysig specialname 
    instance void set_Item (
        string key,
        class [System.Runtime]System.Type 'value'
    ) cil managed 
{
    .param [2]
        .custom instance void [System.Runtime]System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute::.ctor(valuetype [System.Runtime]System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes) = (
            01 00 01 00 00 00 00 00
        )
    // Method begins at RVA 0x12aa1a
    // Code size 15 (0xf)
```

which shows how only the value parameter gets annotated so that the linker will only throw warnings if Type param is not annotated.

tlakollo:
Fix automatic propagation in property setters
Add test

Fixes #1902

### DIFF
--- a/test/Mono.Linker.Tests.Cases/DataFlow/PropertyDataFlow.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/PropertyDataFlow.cs
@@ -144,7 +144,8 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 			instance.TestInstancePropertyWithStaticField ();
 			instance.TestPropertyWithDifferentBackingFields ();
 			instance.TestPropertyWithExistingAttributes ();
-			instance.TestPropertyWithIndexer (null);
+			instance.TestPropertyWithIndexerWithMatchingAnnotations (null);
+			instance.TestPropertyWithIndexerWithoutMatchingAnnotations (null);
 		}
 
 		class TestAutomaticPropagationType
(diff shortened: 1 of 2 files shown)


tlakollo:
The main branch of linker is broken after introducing #1852

test/Mono.Linker.Tests.Cases/DataFlow/PropertyDataFlow.cs
```diff
@@ -265,6 +267,27 @@ public void TestPropertyWithExistingAttributes ()
...
+			{
+				var propclass = new PropertyWithIndexer ();
+				propclass[1] = myType;
+				propclass[1].RequiresPublicConstructors ();
+			}
+
+			public class PropertyWithIndexer
+			{
+				[CompilerGenerated]
+				Type[] Property_Field;
```
vitek-karas, line 283: Nit: Just curious - I don't think the test needs the field to be compiler generated does it? It would work equally well if it was 
correctly annotated (just like the property).

test/Mono.Linker.Tests.Cases/DataFlow/PropertyDataFlow.cs
```diff
@@ -265,6 +267,27 @@ public void TestPropertyWithExistingAttributes ()
 				[param: DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicConstructors)]
 				set { PropertyWithExistingAttributes_Field = value; }
 			}
+
+			[UnrecognizedReflectionAccessPattern (typeof (PropertyWithIndexer), "Item.set", new Type[] { typeof (Type) }, messageCode: "IL2067")]
+			[LogDoesNotContain ("'Value passed to parameter 'index' of method 'Mono.Linker.Tests.Cases.DataFlow.PropertyDataFlow.TestAutomaticPropagationType.PropertyWithIndexer.Item.set'")]
+			public void TestPropertyWithIndexer (Type myType)
```
vitek-karas, line 273: Maybe there should be a second test which calls the setter with correctly annotated value - and validate that it doesn't warn. That's what the original bug is about, so we should have a test case for it.

vitek-karas (approved):

MichalStrehovsky (approved):
Thank you!

test/Mono.Linker.Tests.Cases/DataFlow/PropertyDataFlow.cs
```diff
@@ -265,6 +268,36 @@ public void TestPropertyWithExistingAttributes ()
 				[param: DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicConstructors)]
 				set { PropertyWithExistingAttributes_Field = value; }
 			}
+
+			public void TestPropertyWithIndexerWithMatchingAnnotations ([DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicConstructors)] Type myType)
```
vitek-karas, line 272: This needs RecognizedReflectionAccessPattern attribute - otherwise the test doesn't actually validate anything - it might generate warnings, but the test will not fail as is.

PR #1911, opened 2021-03-22
Simplify internal attributes handling by unifying them with custom attributes

marek-safar:


src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -14,34 +15,56 @@ namespace Mono.Linker
...
-			if (context.CustomAttributes.HasInternalAttributes (provider)) {
-				foreach (var internalAttribute in context.CustomAttributes.GetInternalAttributes (provider))
-					AddAttribute (ref _linkerAttributes, internalAttribute);
-			}
+
+			return new LinkerAttributesInformation (cache);
 		}
 
 		static void AddAttribute (ref Dictionary<Type, List<Attribute>> attributes, Attribute attributeValue)
 		{
```
vitek-karas, line 67: I think it's now called from just one place - so maybe we can inline this.

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -33,35 +33,44 @@ public void Parse (AttributeInfo xmlInfo)
...
 
+				TypeDefinition attributeType;
 				string internalAttribute = GetAttribute (iterator.Current, "internal");
 				if (!string.IsNullOrEmpty (internalAttribute)) {
-					ProcessInternalAttribute (provider, internalAttribute);
-					continue;
-				}
+					var removeAttributeType = typeof (RemoveAttributeInstancesAttribute);
+					var removeAttributeTypeAssembly = AssemblyDefinition.ReadAssembly (removeAttributeType.Assembly.Location);
+					var tr = removeAttributeTypeAssembly.MainModule.ImportReference (removeAttributeType);
```
vitek-karas, line 49: This loads linker's assembly into the closure of the app being trimmed. This might cause trouble - for example trying to run 6.0 linker on a 5.0 app might be problematic.
I think it would be easier/cleaner if we simply defined the attribute anew in one of the already loaded modules (CoreLib?).

vitek-karas (changes requested):

vitek-karas (approved):

Issue #1895, opened 2021-03-15
Type.GetType("TypeName") will pick the "first" such type found


vitek-karas:
If we're resolving type name -> `TypeDefinition` (for example `Type.GetType` does this), and the type name is not assembly qualified, we will iterate over all assemblies and search each one for that type name. Once we find the first, we'll use that.

There's no guarantee it's the right one (if there are multiple). It would be easy to write a counter example.

sbomer:
I believe for non-qualified type names, `GetType` will only look in the calling assembly or in corelib - we might want to do the same in the linker.

vitek-karas:
Yeah - I know, the problem is tricky though.

```C#
// Assembly A
public void Test()
{
    LoadRandomType("MyType");
}

// Assembly B
public object LoadRandomType([DynamicallyAccessedMembers(PublicConstructors)] string typeName)
{
    Activator.CreateInstance(Type.GetType(typeName));
}
```

The runtime behavior in this case is to search to `MyType` in assembly `B`. But currently (without cross-method analysis) there's no way for the linker to figure this out. At the time we're trying to resolve `MyType` we only see assembly `A`.

sbomer:
I'm curious if there are cases where we really need to support non-qualified type names flowing into `DynamicallyAccessedMembers`. If we don't have a way to find the right type, maybe it's better to warn instead of marking all types with the same name.

Anyone wrapping reflection APIs this way will already be facing the limitation that the wrapper only works for certain non-qualified type names (those that exist in the wrapper assembly or corelib) - it might be useful for this to produce a warning even independently of trimming. I think it should be possible to convert most of these cases to use `typeof` since the wrapper assembly and corelib will already be referenced. In cases where there are nested wrappers from different assemblies this might not be true, but this seems like a fragile setup in the first place.

eerhardt:
> I'm curious if there are cases where we really need to support non-qualified type names flowing into DynamicallyAccessedMembers

Check out these places from ComponentModel.TypeConverter:

https://github.com/dotnet/runtime/blob/82ca681cbac89d813a3ce397e0c665e6c051ed67/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/PropertyDescriptor.cs#L342-L374

[...]

PR #1920, opened 2021-03-25
Resolving a type represented by a type name string should mimic Type.GetType behavior

mateoatr:
See #1895. Mimic `Type.GetType` when resolving a type from a type name string. This is, if the type name is not assembly qualified, we will try to search for it in the caller's assembly, mscorlib and System.Private.CoreLib, in that same order. If the type is not found anywhere, linker will produce a warning.

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -1573,6 +1573,23 @@ void TestMethod(MethodInfo methodInfo)
 
 The assembly 'assembly' produced trim analysis warnings in the context of the app. This means the assembly has not been fully annotated for trimming. Consider contacting the library author to request they add trim annotations to the library. To see detailed warnings for this assembly, turn off grouped warnings by passing either `--singlewarn-` to show detailed warnings for all assemblies, or `--singlewarn- "assembly"` to show detailed warnings for that assembly. https://aka.ms/dotnet-illink/libraries has more information on annotating libraries for trimming.
 
+#### `IL2105`: Type 'type' was not found in the caller assembly nor in the base library. Type name strings used for dynamically accessing a type should be assembly qualified.
+
+Type name strings representing dynamically accessed types must be assembly qualified, otherwise linker will first search for the type name in the caller's assembly and then in System.Private.CoreLib.
+If the linker fails to resolve the type name, null will be returned.
+
+```C#
+void TestInvalidTypeName()
+{
+    RequirePublicMethodOnAType("Foo.Unqualified.TypeName");
+}
+void RequirePublicMethodOnAType(
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
+    string typeName)
+{
+}
+```
+
 ## Single-File Warning Codes
 
 #### `IL3000`: 'member' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1006,7 +1006,7 @@ namespace Mono.Linker.Dataflow
 						}
 						foreach (var typeNameValue in methodParams[0].UniqueValues ()) {
 							if (typeNameValue is KnownStringValue knownStringValue) {
-								TypeReference foundTypeRef = _context.TypeNameResolver.ResolveTypeName (knownStringValue.Contents, out AssemblyDefinition typeAssembly);
+								TypeReference foundTypeRef = _context.TypeNameResolver.ResolveTypeName (knownStringValue.Contents, callingMethodDefinition, out AssemblyDefinition typeAssembly, false);
 								TypeDefinition foundType = foundTypeRef?.ResolveToMainTypeDefinition ();
 								if (foundType == null) {
 									// Intentionally ignore - it's not wrong for code to call Type.GetType on non-existing name, the code might expect null/exception back.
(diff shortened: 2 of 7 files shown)


src/linker/Linker/TypeNameResolver.cs
```diff
@@ -31,13 +31,37 @@ public TypeReference ResolveTypeName (string typeNameString)
...
+			// search for the type in the calling object's assembly, mscorlib and SPCL, in that order.
+			AssemblyDefinition assemblyOfCaller = origin switch {
+				TypeReference tr => tr.Module.Assembly,
+				_ => origin.DeclaringType.Module.Assembly
+			};
+
+			if (assemblyOfCaller != null && TryResolveTypeName (assemblyOfCaller, parsedTypeName, out var typeRef))
+				return typeRef;
+
+			AssemblyDefinition mscorlibAssembly = _context.GetLoadedAssembly ("mscorlib");
```
MichalStrehovsky, line 44: `Type.GetType` only searches in whatever is the core library of the given runtime. The docs are confusing because they refer to both mscorlib and system.private.corelib, but the former is for .NET Framework, the latter is for Core.

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -31,13 +31,37 @@ public TypeReference ResolveTypeName (string typeNameString)
...
+			if (mscorlibAssembly != null && TryResolveTypeName (mscorlibAssembly, parsedTypeName, out var typeRefFromMscorlib))
+				return typeRefFromMscorlib;
+
+			AssemblyDefinition spclAssembly = _context.GetLoadedAssembly (PlatformAssemblies.CoreLib);
+			if (spclAssembly != null && TryResolveTypeName (spclAssembly, parsedTypeName, out var typeRefFromSPCL))
+				return typeRefFromSPCL;
 
+			_context.LogWarning ($"Type '{typeNameString}' could not be found in the caller's assembly, 'mscorlib' nor 'System.Private.CoreLib'",
```
MichalStrehovsky, line 52: We don't emit warnings like this because Type.GetType is often used in lightup (look for something using reflection and use it if it's there, fallback to something else if it isn't). This warning would be spurious.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -966,7 +966,7 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 						}
 						foreach (var typeNameValue in methodParams[0].UniqueValues ()) {
 							if (typeNameValue is KnownStringValue knownStringValue) {
-								TypeReference foundTypeRef = _context.TypeNameResolver.ResolveTypeName (knownStringValue.Contents);
+								TypeReference foundTypeRef = _context.TypeNameResolver.ResolveTypeName (knownStringValue.Contents, callingMethodDefinition);
```
MichalStrehovsky, line 969: IMHO this should be the only place that is allowed to call ResolveTypeName with an unqualified assembly. All the other callers of ResolveTypeName should require the string to be assembly qualified and warn if it isn't. https://github.com/mono/linker/issues/1895#issuecomment-800062109

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -31,13 +31,45 @@ public TypeReference ResolveTypeName (string typeNameString)
...
+			// assembly.
+			AssemblyDefinition assemblyOfCaller = origin switch {
+				TypeReference tr => tr.Module.Assembly,
+				_ => origin.DeclaringType.Module.Assembly
+			};
+
+			if (assemblyOfCaller != null && TryResolveTypeName (assemblyOfCaller, parsedTypeName, out var typeRef))
+				return typeRef;
+
+			if (Environment.Version.Major == 4) {
```
marek-safar, line 50: What is the intention (goal) of this check?

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -31,13 +31,45 @@ public TypeReference ResolveTypeName (string typeNameString)
 				return ResolveTypeName (null, assemblyQualifiedTypeName);
 			}
 
-			foreach (var assemblyDefinition in _context.GetReferencedAssemblies ()) {
-				var foundType = ResolveTypeName (assemblyDefinition, parsedTypeName);
-				if (foundType != null)
-					return foundType;
+			if (!isCallFromGetType) {
+				_context.LogWarning ($"Type name strings used for dynamically accessing a type must be assembly qualified.",
```
marek-safar, line 35: I don't follow the message, it says "must be assembly qualified" but it's only a warning?

mateoatr: Thanks! Changed the phrasing here. I've made this a warning since we previously used to go through all loaded assemblies searching for the referenced type (which was wrong); so there could be users who will now have breaking apps with no good way to figure out what is going on. That being said -- users shouldn't lean on the linker behavior to pull a type out of nowhere, and the documentation already specifies that type names should be fully qualified, otherwise `Type.GetType` will only search for it in the calling object's assembly and SPCL (https://docs.microsoft.com/en-us/dotnet/api/system.type.gettype?view=net-5.0)... so I'm also happy to not warn here.

docs/error-codes.md
```diff
@@ -1569,6 +1569,23 @@ void TestMethod(MethodInfo methodInfo)
 }
 ```
 
+#### `IL2104`: Type name strings used for dynamically accessing a type must be assembly qualified.
```
marek-safar, line 1572: Update the warning message to match the code

docs/error-codes.md
```diff
@@ -1569,6 +1569,23 @@ void TestMethod(MethodInfo methodInfo)
...
+
+Type name strings representing types with dynamically accessed members must be assembly qualified, otherwise the linker will not be able to figure out from which assembly the type should be resolved.
```
marek-safar, line 1574: It'd be nice to document the look up logic order, e.g. first is current assembly and then corelib.

marek-safar (approved):

test/Mono.Linker.Tests.Cases/DataFlow/AssemblyQualifiedNameDataflow.cs
```diff
@@ -50,6 +51,14 @@ static void TestConstructors ()
...
 		}
 
+		[ExpectedWarning ("IL2104",
+			"Type 'System.Invalid.TypeName' was not found in the caller assembly nor in the base library. " +
+			"Type name strings used for dynamically accessing a type should be assembly qualified.")]
+		static void TestUnqualifiedTypeNameWarns ()
+		{
+			RequirePublicConstructors ("System.Invalid.TypeName");
+		}
+
```
vitek-karas, line 61: Please also add a test for this happening in the `TypeConverter` attribute - as that is a special code path in the mark step (doesn't go through data flow).

I filed https://github.com/mono/linker/issues/1942 to remove this as a special case, but for now we should add a test for this.

vitek-karas (approved):

PR #1921, opened 2021-03-26
Merge SingleFileAnalyzer into RequiresAssemblyFilesAnalyzer and add capability to suppress IL3000/IL3001 with RequiresAssemblyFilesAnalyzer

tlakollo:
Merge SingleFileAnalyzer into RequiresAssemblyFilesAnalyzer
Add new logic for suppressing warnings on SingleFile dangerous cases if RequiresAssemblyFilesAttribute is used on caller
Add test for suppression


test/ILLink.RoslynAnalyzer.Tests/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -19,6 +19,33 @@ static Task VerifySingleFileAnalyzer (string source, params DiagnosticResult[] e
 				expected);
 		}
 
+		// TODO: Remove this once we have the new attribute available for linker testing.
+		static Task VerifySingleFileAnalyzerWithRequiresAssemblyFilesAttribute (string source, params DiagnosticResult[] expected)
+		{
+			source = @"
+using System.Reflection;
```
tlakollo, line 26: @vitek-karas I thought that we had the attribute already available since we already bump the linker to net6.0, but if I delete this block of code Roslyn will complain about System.Diagnostics.CodeAnalysis not having RequiresAssemblyFilesAttribute. I wasn't sure if I was missing anything 😵

test/ILLink.RoslynAnalyzer.Tests/AvoidAssemblyLocationInSingleFileTests.cs
```diff
@@ -19,6 +19,33 @@ static Task VerifySingleFileAnalyzer (string source, params DiagnosticResult[] e
...
+namespace System.Diagnostics.CodeAnalysis
+{
+#nullable enable
+    [AttributeUsage(AttributeTargets.Constructor |
+                    AttributeTargets.Event |
+                    AttributeTargets.Method |
+                    AttributeTargets.Property,
+                    Inherited = false,
+                    AllowMultiple = false)]
+    public sealed class RequiresAssemblyFilesAttribute : Attribute
```
tlakollo, line 36: @vitek-karas I thought that we had the attribute already available since we already bump the linker to net6.0, but if I delete this block of code Roslyn will complain about System.Diagnostics.CodeAnalysis not having RequiresAssemblyFilesAttribute. I wasn't sure if I was missing anything 😵

mateoatr: It is already checked-in in dotnet/runtime and I can see it in System.Private.CoreLib.dll; not sure how the analyzer resolves it though, I think @agocke might have an idea.

mateoatr: I opened #1922, which should solve this, PTAL. When compiling the analyzer tests, we use .NET 5 assemblies, see:
https://github.com/mono/linker/blob/000432728ae8603d7ac0519d844bc260eee4910e/test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier%601.cs#L53

This is the reason why `RequiresAssemblyFilesAttribute` can't be used.

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -95,6 +152,39 @@ public override void Initialize (AnalysisContext context)
 							url));
 					}
 				}
+
+				static bool Contains<T, TComp> (ImmutableArray<T> list, T elem, TComp comparer)
```
mateoatr, line 156: I think we should move this and bellow methods to another file containing helper methods.

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -46,9 +72,36 @@ public override void Initialize (AnalysisContext context)
...
+				}
+
+				var properties = propertiesBuilder.ToImmutable ();
+				var methods = methodsBuilder.ToImmutable ();
+
 				context.RegisterOperationAction (operationContext => {
 					var methodInvocation = (IInvocationOperation) operationContext.Operation;
-					CheckCalledMember (operationContext, methodInvocation.TargetMethod);
+					var targetMethod = methodInvocation.TargetMethod;
+					if (!operationContext.ContainingSymbol.HasAttribute (RequiresAssemblyFilesAttribute) && Contains (methods, targetMethod, SymbolEqualityComparer.Default)) {
```
mateoatr, line 100: Nit: maybe add a comment that this check and the one we have for properties can be removed once/if we annotate dotnet/runtime?

mateoatr (approved):

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
tlakollo: Even with the annotation in dotnet/runtime this code still needs to exist, the description provided by IL3000 and IL3001 is something the users can action and we already have documented. We don't want to produce IL3002 for everything, so these warnings (IL3000 and IL3001) still need to exist and be triggered in case someone calls directly the public API. The only difference would be that they can be suppressed by using the RequiresAssemblyFiles attribute.

mateoatr: We can always decide on which diagnostic to print inside CheckCalledMember, also `RequiresAssemblyFilesAttribute` has a `Message` property that can give further instructions to users.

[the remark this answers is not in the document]
tlakollo: I didn't found a good way to use the ImmutableArray inside CheckCalledMember so I passed them as parameters, after that all the checks are now in CheckCalledMember. I also added a test for when the ContainingSymbol is a property accessor and some code to fix it.

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
```diff
@@ -49,5 +51,38 @@ internal static bool TryGetAttributeWithMessageOnCtor (this ISymbol symbol, stri
 
 			return false;
 		}
+
+		internal static bool Contains<T, TComp> (ImmutableArray<T> list, T elem, TComp comparer)
```
mateoatr, line 55: These functions aren't extending anything, we should put them in a new file.

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -87,57 +86,63 @@ public override void Initialize (AnalysisContext context)
...
-					CheckCalledMember (operationContext, eventRef.Member);
+					CheckCalledMember (operationContext, eventRef.Member, properties, methods);
 				}, OperationKind.EventReference);
 
 				static void CheckCalledMember (
 					OperationAnalysisContext operationContext,
-					ISymbol member)
+					ISymbol member,
+					ImmutableArray<IPropertySymbol> dangerousProperties,
+					ImmutableArray<IMethodSymbol> dangerousMethods)
```
mateoatr, line 128: Why not a single `ImmutableArray<ISymbol>` instead.

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
tlakollo: Now there is a file called ImmutableArrayOperations

PR #1922, opened 2021-03-26
Compile roslyn analyzer tests using repo's assembly refs

mateoatr:
This change will allow us to test our roslyn analyzers with the latest dotnet/runtime bits.

### DIFF
--- a/test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
@@ -39,26 +39,13 @@ namespace ILLink.RoslynAnalyzer.Tests
 			(string, string)[]? globalAnalyzerOptions = null)
 			=> CreateCompilation (CSharpSyntaxTree.ParseText (src), globalAnalyzerOptions);
 
-		static List<MetadataReference> GetLatestNETCoreReferenceAssemblies ()
+		static List<MetadataReference> GetSystemPrivateCoreLibMetadataReference ()
 		{
 			FileInfo fiSPCL = new FileInfo (typeof (int).Assembly.Location);
-			if (fiSPCL == null || fiSPCL.DirectoryName == null)
-				throw new FileNotFoundException ("Could not find directory containing System.Private.CoreLib.dll.");
+			if (fiSPCL == null)
+				throw new FileNotFoundException ("Could not find System.Private.CoreLib.dll.");
 
-			DirectoryInfo netCoreAppDirectory = new DirectoryInfo (fiSPCL.DirectoryName);
-			string[] assemblies = Directory.GetFiles (netCoreAppDirectory.FullName, "System.*.dll");
-
-			List<MetadataReference> metadataReferences = new List<MetadataReference> ();
-			foreach (var assemblyLocation in assemblies) {
-				try {
-					var assemblyName = AssemblyName.GetAssemblyName (assemblyLocation);
-					metadataReferences.Add (MetadataReference.CreateFromFile (assemblyLocation));
-				} catch (BadImageFormatException) {
-					// Although we filter out assemblies not starting with 'System.', there can be assemblies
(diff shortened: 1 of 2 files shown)


test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
```diff
@@ -38,19 +39,42 @@ public static DiagnosticResult Diagnostic (DiagnosticDescriptor descriptor)
...
 		{
+			FileInfo fiSPCL = new FileInfo (typeof (int).Assembly.Location);
+			if (fiSPCL == null || fiSPCL.DirectoryName == null)
+				throw new FileNotFoundException ("Could not find directory containing System.Private.CoreLib.dll.");
+
+			DirectoryInfo netCoreAppDirectory = new DirectoryInfo (fiSPCL.DirectoryName);
+			string[] assemblies = Directory.GetFiles (netCoreAppDirectory.FullName, "System.*.dll");
+
 			TestCaseUtils.GetDirectoryPaths (out _, out string testAssemblyPath);
 			var expectationsPath = Path.Combine (testAssemblyPath, "Mono.Linker.Tests.Cases.Expectations.dll");
```
tlakollo, line 52: We dont use this anymore

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
```diff
@@ -38,19 +39,38 @@ public static DiagnosticResult Diagnostic (DiagnosticDescriptor descriptor)
 			(string, string)[]? globalAnalyzerOptions = null)
 			=> CreateCompilation (CSharpSyntaxTree.ParseText (src), globalAnalyzerOptions);
 
+		static List<MetadataReference> GetLatestNETCoreReferenceAssemblies ()
+		{
+			FileInfo fiSPCL = new FileInfo (typeof (int).Assembly.Location);
```
tlakollo, line 44: Do we really need to filter and check everything?
Why not to give the metadata references without modifying them, like:
```
var referenceSPC = MetadataReference.CreateFromFile (typeof (object).Assembly.Location);
if (referenceSPC is null)
  throw new FileNotFoundException ("Could not find System.Private.CoreLib.dll");
var comp = CSharpCompilation.Create (
  assemblyName: Guid.NewGuid ().ToString ("N"),
	syntaxTrees: new SyntaxTree[] { src },
	references: new[] { referenceSPC },
	new CSharpCompilationOptions (OutputKind.DynamicallyLinkedLibrary));
```

mateoatr: That would only add a reference to `System.Private.CoreLib`; it'd probably work for the tests that we currently have, but if someone were to add more complex tests cases, say having something like:

```C#
List<Type> types = new List<Type> { typeof(MyType) };
```

it'd fail... also, this doesn't seem to take much time, so I'd rather keep it this way.

tlakollo (approved):
Looks good to me

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
mateoatr: Wrote too fast 🙃. Everything seems to work fine even when just referencing SPCL, so I took your advice. I'll let @agocke comment on this.

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
```diff
@@ -5,6 +5,7 @@
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Loader;
```
tlakollo, line 8: Seems like this is not being used, right?

PR #1923, opened 2021-03-26
Improve external custom attribute annotations support

### LINKED ISSUE #1915: LinkAttributes doesn't support custom attribute overloading
It's not possible to insert certain overloads of DynamicDependency using the LinkAttributes file.
For example:
```C#
public class ReferencedType {
  // Attribute trying to be inserted in the code via xml
  // [DynamicDependency (DynamicallyAccessedMemberTypes.All, NonUsedType)]
  public static void Main () {}
}

public class NonUsedType {
  public static void PleasePreserveThisMethod () {}
}
```
The main problem is that the xml only supports to write something like the following:
```XML
<linker>
  <assembly fullname="SomeAssembly">
    <type name="ReferencedType">
      <method signature="System.Void Main()">
        <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicDependency">
          <argument>All</argument>
          <argument>NonUsedType</argument>
        </attribute>
      </method>
    </type>
  </assembly>
</linker>
```
The previous code assumes that argument one is of type MemberTypes. In practice, the linker will take the first constructor that matches the number of arguments, in this case, it will take the overload of DynamicDependency that takes as the first argument a string and the second argument a type.
```C#
// This is the constructor the linker finds just by searching a constructor with 2 arguments
public DynamicDependencyAttribute(string memberSignature, Type type);

public DynamicDependencyAttribute(DynamicallyAccessedMemberTypes memberTypes, Type type);
```
We should have a way to declare the type of argument to override the default behavior of the linker.

marek-safar:
This change makes external custom attribute values logic deterministic
by always using constructors which match the input argument types.

Fixes #1915


test/Mono.Linker.Tests.Cases/DataFlow/XmlAnnotations.cs
```diff
@@ -13,7 +13,6 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 {
 	[SkipKeptItemsValidation]
 	[SetupLinkAttributesFile ("XmlAnnotations.xml")]
-	[ExpectedWarning ("IL2031", "System.DoesNotExistAttribute", FileName = "XmlAnnotations.xml")]
 	[LogDoesNotContain ("IL2067: Mono.Linker.Tests.Cases.DataFlow.XmlAnnotations.ReadFromInstanceField():*", true)]
 	class XmlAnnotations
```
tlakollo, line 17: We should add a test case similar to the one in #1915, DynamicDependency not only makes use of the overload feature that this PR fixes, but also interacts with the caches in the linker so it triggers an action that can be verified by the ResultChecker

test/Mono.Linker.Tests.Cases/DataFlow/XmlAnnotations.xml
```diff
@@ -13,51 +13,51 @@
...
-            <argument>PublicConstructors</argument>
+            <argument type="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes">PublicConstructors</argument>
           </attribute>
         </parameter>
       </method>
       <method signature="SpacesBetweenParametersWrongArgument (System.Type, System.Boolean)">
         <parameter name="type">
           <attribute fullname="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute">
-            <argument>NonValidArgument</argument>
+            <argument type="System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes">0</argument>
```
tlakollo, line 34: We need to test that a warning is produced when someone passes something for which the constructor was not found. In this case, 0 means MemberTypes.None so it should be a valid entry

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -57,43 +57,83 @@ IEnumerable<CustomAttribute> ProcessAttributes (XPathNavigator nav, ICustomAttri
...
 			return attributes;
+
+			static string FormatCustomAttribute (CustomAttribute ca)
+			{
+				StringBuilder sb = new StringBuilder ();
+				sb.Append (ca.Constructor.GetDisplayName ());
+				sb.Append (" { args: ");
+				for (int i = 0; i < ca.ConstructorArguments.Count; ++i) {
+					var caa = ca.ConstructorArguments[i];
+					sb.Append ($"{caa.Type.GetDisplayName ()} {caa.Value}");
```
tlakollo, line 75: You need to append the comma before the display name or the format will be wrong

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/DataFlow/XmlAnnotations.cs
marek-safar: Added extra test for DynamicDependency with Type argument into relevant tests

test/Mono.Linker.Tests.Cases/DataFlow/XmlAnnotations.xml
marek-safar: See the added test in the PR, there are tests for invalid arguments which are errors that would be the same case here so the whole test fixture would fail. There is no validation for this test so not sure what it's supposed to check.

src/linker/Linker.Steps/LinkAttributesParser.cs
marek-safar: It's an arbitrary format for log, see the comma handling after type+constant pair

tlakollo: Yes
What Im saying is that in i=0 you append the type name+value
In i=1 you append type name+value and then the coma
leaving you with "type name+valuetype name+value, "
the first strings will be together but there will be an extra comma at the end

test/Mono.Linker.Tests.Cases/DataFlow/XmlAnnotations.xml
tlakollo: Yeah, I see that now we have a error test file, seems fine to me to have the failing test there then

src/linker/Linker.Steps/LinkAttributesParser.cs
marek-safar: makes sense now

PR #1925, opened 2021-03-27
Remove automatic default serializable constructor marking

marek-safar:


marek-safar:
Merging to check the impact on libraries

PR #1929, opened 2021-03-29
Add support for annotations in classes/interfaces/structs in the linker

tlakollo:
Adds support for `DynamicallyAccessedMembers` on types (class, struct, interface).

The desired behavior is that calling `object.GetType()` on a statically typed instance will look for the annotation on the static type (or its base class or interface). If it finds it there, it will apply the annotation to the static type and all of its derived types (which are marked). Then it will treat the return value of the `object.GetType()` as a `Type` annotated with the found annotation.

The implementation is somewhere in between the `MarkStep` and the `ReflectionMethodBodyScanner` so I put it into a new class `DynamicallyAccessedMembersTypeHierarchy`. It stores a cache of type -> annotation. This cache is built from `MarkType` for all types regardless if the reflection pattern to access it was seen or not. It also stores information if the annotation has already been applied (if so, newly marked types to which that annotation also applied will get it applied on them as well - this means there's no need for post-processing).

Added lot of tests to cover interesting cases.


src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -0,0 +1,197 @@
...
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using Mono.Cecil;
+using Mono.Linker.Steps;
+
+namespace Mono.Linker.Dataflow
+{
+	class DynamicallyAccessedMembersTypeHierarchy
```
vitek-karas, line 13: Naming: I would welcome any suggestion for a better name...

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -0,0 +1,197 @@
...
+		//   - Marked types which are not in the cache don't have any annotation
+		// Interface types
+		//   - All interface types accessible from marked types are stored in the cache
+		//   - If the interface type doesn't have annotation the value None is stored here
+		//
+		// It's not possible to use the marking as a filter for interfaces in the cache
+		// because interfaces are marked late and in effectively random order.
+		// For this cache to be effective we need to be able to fill it for all base types and interfaces
+		// of a type which is currently being marked - at which point the interfaces are not yet marked.
+		readonly Dictionary<TypeDefinition, (DynamicallyAccessedMemberTypes annotation, bool applied)> _typesInDynamicallyAccessedMembersHierarchy;
```
vitek-karas, line 39: This cache will include all interfaces reachable from marked types - which can be relatively large number (1000s). It is built regardless of the annotations being used or even present, since it acts as a negative cache as well.

Alternatively this could be implemented by re-computing the annotation for interface each type it's needed - basically trade memory or CPU. I went with the cache as some kind of cache would still be needed for the `applied` bit, and the memory impact doesn't seem that bad (couple of KB).

test/Mono.Linker.Tests.Cases/Reflection/ObjectGetType.cs
```diff
@@ -4,61 +4,792 @@
...
+			}
+
+			[Kept]
+			[KeptMember (".ctor()")]
+			class ClassImplementingAnnotatedInterface : IBasicAnnotatedInterface
+			{
+				[Kept]
+				public void UsedMethod () { }
+				[Kept] // The type is not sealed, so trimmer will apply the annotation from the interface
+				public void UnsedMethod () { }
```
tlakollo, line 111: Nit: Unsed instead of Unused in all the test inside BasicAnnotationWithNoDerivedClasses

test/Mono.Linker.Tests.Cases/DataFlow/TypeDataflow.cs
```diff
@@ -0,0 +1,127 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
tlakollo, line 1: All the cases inside here all already covered in ObjectGetType tests so it's safe to delete this file

test/Mono.Linker.Tests.Cases/Reflection/ObjectGetType.cs
```diff
@@ -4,61 +4,792 @@
```
tlakollo, line 4: Nit: Add license

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -910,36 +918,42 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
-							if (canUse) {
-								methodReturnValue = new SystemTypeValue (staticType);
+								// We ignore the fact that the type can be annotated (see below for hanling of annotated types)
+								// This means the annotations (if any) won't be applied - instead we rely on the exact knowledge
+								// of the type. So for example even if the type is annotated with PublicMethods
+								// but the code calls GetProperties on it - it will work - mark properties, don't mark methods
+								// since we ignored the fact that it's annotated.
+								// This can be seen a little bit as a violation of the annotation, but we already have similar cases
+								// where a parameter is annotated and if something in the method sets a specific known type to it
+								// we will also make it just work, even if the annotation doesn't match he usage.
```
tlakollo, line 939: nit: doesn't match he usage -> doesn't match the usage

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -910,36 +918,42 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
 
-							if (!canUse) {
 								// We can allow Object.GetType to be modeled as System.Delegate because we keep all methods
 								// on delegates anyway so reflection on something this approximation would miss is actually safe.
-								canUse = staticType.IsTypeOf ("System", "Delegate");
-							}
 
-							if (canUse) {
-								methodReturnValue = new SystemTypeValue (staticType);
+								// We ignore the fact that the type can be annotated (see below for hanling of annotated types)
```
tlakollo, line 932: nit: hanling -> handling

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -0,0 +1,197 @@
...
+				}
+			}
+
+			Debug.Assert (!apply || annotation != DynamicallyAccessedMemberTypes.None);
+
+			if (apply) {
+				// One of the base/interface types is already marked as having the annotation applied
+				// so we need to apply the annotation to this type as well
+				var reflectionMethodBodyScanner = new ReflectionMethodBodyScanner (_context, _markStep);
+				var reflectionPatternContext = new ReflectionPatternContext (_context, true, type, type);
```
MichalStrehovsky, line 96:
```suggestion
				using var reflectionPatternContext = new ReflectionPatternContext (_context, true, type, type);
```

MichalStrehovsky (approved):
Looks good otherwise!

test/Mono.Linker.Tests.Cases/Reflection/ObjectGetType.cs
```diff
@@ -4,61 +4,792 @@
...
+			public static void Test ()
+			{
+				// Make sure the interface is kept
+				var i = typeof (IAnnotatedInterface);
+				GetInstance ().GetType ().GetMethod ("InterfaceMethod");
+			}
+		}
+
+		[Kept]
+		class AllAnnotationsAreApplied
```
marek-safar, line 652: How well does the code handle complex or conflicting hierarchies? I couldn't find any test and it could be worth adding as well.

Examples like

```c#
[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.All)]
class X : A, I
{
}

[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.None)]
class A : I
{
	public virtual void M ();
}

[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.None)]
interface I
{
	void M ();
}
```

or 

```c#
[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.None)]
class C : B
{
	public override  void Foo () {}
}

[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.All)]
class B : A
{
	public override  void Foo () {}
}

[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.None)]
class A
{
	public virtual void Foo () {}
}
```

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
vitek-karas: It's actually not possible to use `using` since we're passing it as `ref` (and the compiler checks that the values in `using` is immutable).
But it's a valid concern that we're not calling `Dispose` even though it's just a debug check. I modified all call sites to call `Dispose` - without the try/finally as it doesn't seem necessary - But I can be convinced to add that as well.

test/Mono.Linker.Tests.Cases/Reflection/ObjectGetType.cs
vitek-karas: Good suggestions - I added a bit more complicated case of both samples above and one more around diamond shapes with interfaces.

PR #1930, opened 2021-03-30
Expression.Call can act as MakeGenericMethod so make it perform same validation

### LINKED ISSUE #1819: Expression.Call intrinsic must check for empty types parameter
``` C#
public void GenericMethod<[DynamicallyAccessedMembers(PublicProperties)] T>() {}

public void Test()
{
    Expression.Call(typeof(TestType), nameof(GenericMethod), new Type[] { typeof(AnotherType) });
}
```

This needs to warn:
* It's effectively the same as `MakeGenericMethod`
* It may NOT warn if we can resolve the method in question and prove that it has no annotations

vitek-karas:
Also improved validation done by MakeGenericMethod to avoid warning if no type arguments are passed in.
Added tests.

Fixes #1819

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -797,7 +797,7 @@ namespace Mono.Linker.Dataflow
 								foreach (var stringParam in methodParams[1].UniqueValues ()) {
 									if (stringParam is KnownStringValue stringValue) {
 										foreach (var method in systemTypeValue.TypeRepresented.GetMethodsOnTypeHierarchy (m => m.Name == stringValue.Contents, bindingFlags)) {
-											ValidateGenericMethodInstantiation (ref reflectionContext, method, methodParams[2], calledMethod);
+											ValidateGenericMethodInstantiation (ref reflectionContext, method, calledMethod);
 											MarkMethod (ref reflectionContext, method);
 										}
 
@@ -1611,7 +1611,7 @@ namespace Mono.Linker.Dataflow
 
 						foreach (var methodValue in methodParams[0].UniqueValues ()) {
 							if (methodValue is SystemReflectionMethodBaseValue methodBaseValue) {
-								ValidateGenericMethodInstantiation (ref reflectionContext, methodBaseValue.MethodRepresented, methodParams[1], calledMethod);
+								ValidateGenericMethodInstantiation (ref reflectionContext, methodBaseValue.MethodRepresented, calledMethod);
 								reflectionContext.RecordHandledPattern ();
 							} else if (methodValue == NullValue.Instance) {
 								reflectionContext.RecordHandledPattern ();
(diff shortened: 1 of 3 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1588,30 +1608,31 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 				//
 				case IntrinsicId.MethodInfo_MakeGenericMethod: {
 						reflectionContext.AnalyzingPattern ();
+						if (callingMethodDefinition.Name == "TestWithRequirementsButNoTypeArguments")
```
tlakollo, line 1611: Debug extra stuff

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1588,30 +1608,31 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
 								reflectionContext.RecordHandledPattern ();
 							} else if (methodValue == NullValue.Instance) {
 								reflectionContext.RecordHandledPattern ();
 							} else {
-								// There is a generic parameter which has some requirements on input types.
-								// For now we don't support tracking actual array elements, so we can't validate that the requirements are fulfilled.
-								reflectionContext.RecordUnrecognizedPattern (
-									2060, string.Format (Resources.Strings.IL2060,
-										DiagnosticUtilities.GetMethodSignatureDisplayName (calledMethod)));
+								if (hasTypeArguments) {
```
MichalStrehovsky, line 1622: Is it worth it to add big comment blocks and extra code just to avoid generating a warning when someone calls MakeGenericMethod with an empty array?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -2151,6 +2166,35 @@ void MarkEventsOnTypeHierarchy (ref ReflectionPatternContext reflectionContext,
...
+		{
+			if (!genericMethod.HasGenericParameters)
+				return;
+
+			// If there are no type arguments there's nothing to validate.
+			// This case will typically fail at runtime
+			// since the call requires that all generic parameters have arguments to them.
+			// But it's not the job of the linker to detect possible runtime error cases
+			// the code may well expect to get an exception in this case.
+			if ((typeArgumentsValue as ArrayValue)?.Size.AsConstInt () == 0)
```
MichalStrehovsky, line 2183: Will this be `false` if `methodParams[1]` is not `ArrayValue`?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: Probably not - originally I did it because I've done it for `Expression.Call` (where it's much more common). I'll remove this code.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: The left side should evaluate to `null` and `null != 0`.

MichalStrehovsky: We could probably delete this for the same reasons as https://github.com/mono/linker/pull/1930#discussion_r606120623.

This would fail at runtime if this is non-null and doesn't match the number of required arguments in general, not just for zero. It feels low value compared to the amount of necessary comments and code.

vitek-karas: I guess this makes sense - since it will never really work at runtime. Removed.

PR #1932, opened 2021-03-31
Preserve types for serialization

sbomer:
This adds a flag to keep types for serialization under certain conditions. The details are described in the included docs. See also the comments in SerializationHelper.cs for rationale and a comparison with xamarin-android.

@vitek-karas in particular we discussed only preserving the discovered types instead of marking them, but xamarin-android used to mark types in non-SDK "link" assemblies (in [ApplyPreserveAttribute](https://github.com/xamarin/xamarin-android/blob/681887ebdbd192ce7ce1cd02221d4939599ba762/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs#L17), which uses this logic from [ApplyPreserveAttributeBase](https://github.com/mono/linker/blob/release/5.0/src/tuner/Mono.Tuner/ApplyPreserveAttributeBase.cs#L129) for example). So to match that I am marking all discovered types for now, which should enable deserialization scenarios - PTAL and let me know if this seems reasonable to you.

@LakshanF PTAL (I can't add you to the reviewers)


src/linker/Linker.Steps/InternalSubStepsDispatcher.cs
```diff
@@ -0,0 +1,18 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+using Mono.Cecil;
+using Mono.Collections.Generic;
```
marek-safar, line 8: Looks like redundant usings

src/linker/Linker.Steps/InternalSubStepsDispatcher.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Mono.Linker.Steps
+{
+	public class InternalSubStepsDispatcher : MarkSubStepsDispatcher
```
marek-safar, line 12: Wouldn't be better to "fix" MarkSubStepsDispatcher ?

src/linker/Linker/PreserveSerialization.cs
```diff
@@ -0,0 +1,125 @@
...
+using Mono.Linker.Steps;
+
+namespace Mono.Linker
+{
+
+	// This discovers types attributed with certain serialization attributes, to match the old behavior
+	// of xamarin-android. It is not meant to be complete. Unlike xamarin-andorid:
+	// - this will only discover attributed types in marked assemblies
+	// - this will discover types in non-"link" assemblies as well
+	public class PreserveSerialization : BaseSubStep
```
marek-safar, line 19: I think this should be inside .Steps namespace and use SubStep suffix

src/linker/Linker/SerializationHelper.cs
```diff
@@ -0,0 +1,123 @@
...
+	//
+	// Discover attributed "roots" the same way.
+	//
+	// For each "root":
+	//   recursively scan types of properties and fields (including generic arguments)
+	//     for each recursive type:
+	//       mark the type, and set TypePreserve.All
+	//         this handles the cases where XA used to set TypePreserve.All, preserve default ctors, and mark or conditionally preserve fields/methods (including property and event methods)
+
+	public class SerializationHelper
```
marek-safar, line 46: You know I hate Helper names ;-). What about SerializationMarker instead ?

src/linker/Linker/Driver.cs
```diff
@@ -335,6 +335,12 @@ protected int SetupContext (ILogger customLogger = null)
 
 						continue;
 
+					case "--keep-serialization":
```
marek-safar, line 338: What is the thinking here? What do you need the explicit option for?

I'm wondering if we could do this differently and make all serialization logic dependant on some internal feature flag which would be set automatically or explicitly if something goes wrong. This would help us a lot with size because we could automatically remove also other serialization attributes and we already have a mechanism via `ILLink.LinkAttributes.xml` to do that.

src/linker/Linker/PreserveSerialization.cs
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		enum SerializerKind
+		{
+			None,
+			XmlSerializer,
+			DataContractSerializer,
+		}
+
+		static bool IsPreservedSerializationAttribute (ICustomAttributeProvider provider, CustomAttribute attribute, out SerializerKind serializerKind)
```
marek-safar, line 87: I don't think we should use attributes to detect which serializer is needed. We want to trim all these and other serialization-related attributes whenever it's possible.

src/linker/Linker/Driver.cs
sbomer: The thinking is that the heuristics are optional in case some scenarios don't want the unpredictable behavior or size impact, or want to deal with specific serialized types manually.

> make all serialization logic dependant on some internal feature flag which would be set automatically or explicitly if something goes wrong

If by "something goes wrong" you mean the linker misses some types that are serialized by the app, I don't see how we could tell. Or do you mean something like: if we see a call to `XmlSerializer.ctor` we activate the serialization logic?

src/linker/Linker/PreserveSerialization.cs
sbomer: The reason I'm checking for attributes is because we're concerned about back-compat with xamarin-android, and they used to do this. If we're ok with relaxing this, one idea is to make it conditional on marking a serializer ctor.

src/linker/Linker.Steps/InternalSubStepsDispatcher.cs
sbomer: Made it non-abstract and made the ctor public.

src/linker/Linker/SerializationHelper.cs
sbomer: Good idea, fixed :)

src/linker/Linker/Driver.cs
marek-safar: >  in case some scenarios don't want the unpredictable behavior or size impact

I think we should use the existing feature switch mechanism for that

> Or do you mean something like: if we see a call to XmlSerializer.ctor we activate the serialization logic?

Correct for example someone having serialization logic inside generated SRE code, basically some odd corner cases.

src/linker/Linker/PreserveSerialization.cs
marek-safar: >  If we're ok with relaxing this, one idea is to make it conditional on marking a serializer ctor.

The context back then was different so whatever logic that makes more sense in .NET6 world is preferable.

src/linker/Linker/Driver.cs
sbomer: I don't think this should be a feature switch since it doesn't match the existing pattern:
- There won't be a runtime feature switch for this so there won't be a predictable way to turn off support for these serializers (unless there are plans I'm not aware of)
- Turning this on on just makes it more likely to work but doesn't guarantee it (so a switch like "XmlSerializer.IsSupported" would be misleading)

marek-safar: > There won't be a runtime feature switch for this 

Why not? I'd love to have it to be able to conditional serialization dependant code and attributes with it.

docs/serialization.md
```diff
@@ -0,0 +1,91 @@
+# Serialization
+
+The linker cannot analyze the patterns typically used by reflection-based serializers. Such serializers should be annotated with `RequiresUnreferencedCodeAttribute`, and using them in a trimmed app will likely not work (or will work unpredictably). Set `SuppressTrimAnalysisWarnings` to `false` to see static analysis warnings for these patterns.
```
agocke, line 3: Is the `SuppressTrimAnalysisWarnings` part necessary now?

src/linker/Linker/Driver.cs
agocke: Hm, not clear on what value the feature switch is providing. If you want to remove XmlSerializer from your app it seems like the best way to do that is actually removing references to it. The feature switch seems more likely to cause problems instead of solve them.

src/linker/Linker/PreserveSerialization.cs
sbomer: Changed so that we only mark things if we marked a serializer ctor.

src/linker/Linker/Driver.cs
sbomer: With the existing pattern, I guess a feature switch would just make all of the serializer APIs throw or nop when called, right? I don't think that's a great experience. I could see value in a "feature switch" that gives build-time errors for calls to a serializer, and allows the linker to predictably remove the APIs, but that's another discussion... ;)

That aside, in this change we just need a way to turn off the heuristics. My thinking is that even if we add some kind of feature switch, it should at least be possible to turn off the heuristics when the serializer feature is enabled - that's why I'd prefer it to be a separate argument.

docs/serialization.md
sbomer: I initially wanted to mention it in case people run into issues for xamarin-android apps, where warnings are suppressed by default, but I changed it to just point to the docs for warnings which should stay up-to-date wrt defaults.

src/linker/Linker/Driver.cs
marek-safar: The suggestion for feature switch was a conceptual thing, in the same way, we are keeping weak serialization dependencies we could remove them but we don't have infrastructure in linker to do that yet so this will have to wait.

> That aside, in this change we just need a way to turn off the heuristics.

Then let's use the option name which reflects that. Considering we want to have the heuristic enabled by default we could add something like `--disable-serialization-discovery`

docs/serialization.md
```diff
@@ -0,0 +1,98 @@
...
+- Type graph: recursive logic to build a set of types to consider for serialization, starting from the roots
+- Preservation logic: what the linker does with the discovered types
+
+## Root type discovery
+
+The heuristics will discover types that satisfy _all_ of the following criteria:
+
+- The type is used
+
+  There must be a statically discoverable reference to the type. In other words, if running the linker without the serialization heuristics removes a given type, then the heuristics will not keep it either.
```
vitek-karas, line 28:
```suggestion
  There must be a statically discoverable reference to the type. In other words, if running the linker without the serialization heuristics removes a given type, then the heuristics will not even consider such type to become a root.
```

We might keep more types due to the type graph phase.

docs/serialization.md
```diff
@@ -0,0 +1,98 @@
...
+
+- The type or one of its members must be attributed with a serializer-specific attribute.
+
+  See the sections below about the attributes you can use for each serializer. The attribute must be present on the type, or one of its defined members, including fields/properties/methods/events, public or private, though the serializers may not define attributes that can be placed on all member kinds. It is not enough to place an attribute on a base type or on a member defined by a base type.
+
+Note that passing a type directly to a serializer constructor is _not_ enough to keep it. We do not use dataflow to discover types. For example:
+
+```csharp
+new XmlSerializer (typeof (RootType)); // Will not keep RootType
```
vitek-karas, line 37:
```suggestion
new XmlSerializer (typeof (RootType)); // Will not consider RootType as one of the root types
```

Linker will preserve the RootType since it's referenced.

docs/serialization.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+This pattern will not necessarily keep the type passed into the constructor, even though it is statically analyzable in theory.
```
vitek-karas, line 40:
```suggestion
This pattern will not consider the type passed into the constructor as a root type, even though it is statically analyzable in theory.
```

docs/serialization.md
```diff
@@ -0,0 +1,98 @@
...
+## Preservation logic
+
+For each discovered type (including root types and the recursive type graph), the linker marks the type and the following members:
+
+- Public instance properties
+  - including public or private getters and setters for such properties
+- Public instance fields
+- Public parameterless instance constructors
+
+Note that in general, private members and static members are not preserved, nor are methods or events (other than the mentioned constructor).
```
vitek-karas, line 81: What happens if I put `[DataMember]` on a private field - I would assume we preserve it... but this sentence seems to say otherwise.

docs/serialization.md
```diff
@@ -0,0 +1,98 @@
...
+
+As a last resort, the linker does have limited heuristics that can be enabled to keep _some_ of the types and members required for serialization, but this provides no correctness guarantees; apps which use reflection-based serializers are still considered "broken" as far as the static analysis can tell, and it is up to you to make sure that the app works as intended.
+
+## History
+
+The linker has historically been used for Xamarin scenarios that use reflection-based serializers like XmlSerializer, since before the introduction of the trim analysis warnings. There were limited heuristics to satisfy some simple uses of serializers. To provide backwards compatibility for such scenarios, the linker has built-in heuristics that makes some simple cases "just work", albeit in an opaque and unpredictable way.
+
+Consider disabling this behavior by passing `--disable-serialization-discovery` if possible, but it may be necessary when using legacy serializers that don't provide source generators or a similar solution that is statically analyzable. The following is a description of the heuristics for anyone who is unfortunate enough to have to rely on this behavior.
+
+## Heuristics
```
vitek-karas, line 15: I think we should also mention that the logic only kicks in if the .ctor for a given serializer is marked. It's kind of obvious for apps which use serialization, but I think it's worth mentioning for linker documentation purposes.

src/linker/Linker.Steps/DiscoverSerializationHandler.cs
```diff
@@ -0,0 +1,144 @@
...
+	// - this will only discover attributed types that are marked
+	// - this will discover types in non-"link" assemblies as well
+	public class DiscoverSerializationHandler : IMarkHandler
+	{
+		LinkContext _context;
+
+		public void Initialize (LinkContext context, MarkContext markContext)
+		{
+			_context = context;
+			markContext.RegisterMarkTypeAction (type => ProcessType (type));
```
vitek-karas, line 25:
```suggestion
			markContext.RegisterMarkTypeAction (ProcessType);
```

src/linker/Linker.Steps/DiscoverSerializationHandler.cs
```diff
@@ -0,0 +1,144 @@
...
+			markContext.RegisterMarkMethodAction (method => CheckForSerializerActivation (method));
```
vitek-karas, line 26:
```suggestion
			markContext.RegisterMarkMethodAction (CheckForSerializerActivation);
```

src/linker/Linker.Steps/DiscoverSerializationHandler.cs
```diff
@@ -0,0 +1,144 @@
...
+		}
+
+		void CheckForSerializerActivation (MethodDefinition method)
+		{
+			var type = method.DeclaringType;
+
+			if (!_context.SerializationMarker.IsActive (SerializerKind.DataContractSerializer) &&
+				method.IsConstructor && !method.IsStatic &&
+				type.Namespace == "System.Runtime.Serialization" &&
+				type.Name == "DataContractSerializer") {
```
vitek-karas, line 36: What about `DataContractJsonSerializer` (note that it doesn't derive from `DataContractSerializer`)?

src/linker/Linker/DependencyInfo.cs
```diff
@@ -131,6 +131,11 @@ public enum DependencyKind
 		DisablePrivateReflectionRequirement = 79, // null -> DisablePrivateReflectionAttribute type/methods (note that no specific source is reported)
 		DynamicInterfaceCastableImplementation = 80, // type -> type is marked with IDynamicInterfaceCastableImplementationAttribute and implements the provided interface
 		AlreadyMarked = 82, // null -> member that has already been marked for a particular reason (used to propagate reasons internally, not reported)
+
+		DataContractSerialized = 83, // entry type for DataContract serialization
+		XmlSerialized = 84, // entry type for XML serialization
+		RecursiveType = 85, // recursive type kept dup to serialization handling
```
vitek-karas, line 137: Nit: I would probably rename to something like SerializedRecursiveType - just to make it clear it's for serialization only

src/linker/Linker/SerializationMarker.cs
```diff
@@ -0,0 +1,202 @@
...
+
+			TrackedTypes.Remove (serializerKind);
+		}
+
+		public void MarkRecursiveMembers (TypeReference typeRef, in DependencyInfo reason)
+		{
+			MarkRecursiveMembersInternal (typeRef, reason);
+		}
+
+		void MarkRecursiveMembersInternal (TypeReference typeRef, in DependencyInfo reason)
```
vitek-karas, line 131: Why do we need the internal version if the public one is exactly the same thing?

src/linker/Linker/SerializationMarker.cs
```diff
@@ -0,0 +1,202 @@
...
+			while (typeRef is GenericInstanceType git) {
+				if (git.HasGenericArguments) {
+					foreach (var argType in git.GenericArguments)
+						MarkRecursiveMembersInternal (argType, new DependencyInfo (DependencyKind.GenericArgumentType, typeRef));
+				}
+				_context.Tracer.AddDirectDependency (typeRef, typeReason, marked: false);
+				typeReason = new DependencyInfo (DependencyKind.ElementType, typeRef);
+				typeRef = (typeRef as TypeSpecification).ElementType;
+			}
+			// This doesn't handle other TypeSpecs. We are only matching what xamarin-android used to do.
```
vitek-karas, line 146: I think we should explicitely handle arrays. The code already does this, because (Array[int]).Resolve() actually resolves to Int32 (and not Array)!. So either own up to this behavior by adding a comment or handle it explicitly in code.

src/linker/Linker/SerializationMarker.cs
```diff
@@ -0,0 +1,202 @@
...
+
+			TypeDefinition type = typeRef.Resolve ();
+			if (type == null)
+				return;
+
+			_context.Annotations.Mark (type, typeReason);
```
vitek-karas, line 152: Should we move this below the hashset check? I know it will work either way, but since we already have one hashset check, we might as well avoid a second one.

src/linker/Linker.Steps/MarkSubStepsDispatcher.cs
```diff
@@ -27,7 +27,7 @@ public abstract class MarkSubStepsDispatcher : IMarkHandler
 		List<ISubStep> on_properties;
 		List<ISubStep> on_events;
 
-		protected MarkSubStepsDispatcher (IEnumerable<ISubStep> subSteps)
+		public MarkSubStepsDispatcher (IEnumerable<ISubStep> subSteps)
```
vitek-karas, line 30: Is this necessary for the rest of the change - I could not see how we need that.

docs/serialization.md
sbomer: Definitely - I forgot to update it with the recent changes.

src/linker/Linker/SerializationMarker.cs
sbomer: The issue is not arrays but other obscure cases like function pointers (that's the one I can think of) - I actually have a test where we can hit the function pointer case, but I'm not sure it's worth handling here. I will update the comment to at least call out how it works for arrays.

src/linker/Linker.Steps/MarkSubStepsDispatcher.cs
sbomer: It was based on PR feedback in a previous iteration that used MarkSubStepsDispatcher for the serializer logic. I think it's an improvement, but I can separate it if you prefer - LMK.

src/linker/Linker/SerializationMarker.cs
sbomer: I'd prefer it above, so that we get trace output if multiple serializers cause the same type to be discovered as a root.

src/linker/Linker/SerializationMarker.cs
sbomer: Removing (I think it was left over from a previous iteration that treated root types differently)

docs/serialization.md
sbomer: Currently it causes the type to be considered a root, but the field won't actually be kept. I think I can fix this without building serializer-specific logic into the recursive phase by marking such members up-front.

src/linker/Linker/SerializationMarker.cs
sbomer: Actually with the fix for https://github.com/mono/linker/pull/1932#discussion_r614053345 this will remain.

docs/serialization.md
sbomer: Fixed in the latest iteration (see the commit comments and doc updates)

docs/serialization.md
sbomer: Updated, PTAL

vitek-karas (approved):
Nice!

src/linker/Linker.Steps/DiscoverSerializationHandler.cs
```diff
@@ -32,8 +32,8 @@ void CheckForSerializerActivation (MethodDefinition method)
 
 			if (!_context.SerializationMarker.IsActive (SerializerKind.DataContractSerializer) &&
 				method.IsConstructor && !method.IsStatic &&
-				type.Namespace == "System.Runtime.Serialization" &&
-				type.Name == "DataContractSerializer") {
+				((type.Namespace == "System.Runtime.Serialization" && type.Name == "DataContractSerializer") ||
+				(type.Namespace == "System.Runtime.Serialization.Json" && type.Name == "DataContractJsonSerializer"))) {
```
vitek-karas, line 36: I think this is also mentioned in the doc - so we should update the doc.

PR #1935, opened 2021-04-01
Extend attributes removal to support values matching

marek-safar:
This is useful for key/value pair like custom attributes removal

### DIFF
--- a/docs/data-formats.md
+++ b/docs/data-formats.md
@@ -544,7 +544,9 @@ instead of specifying `fullname` in the attribute as described in the following 
 </linker>
 ```
 
-In some cases it's useful to remove only specific instances of same attribute. This can be achieved by specifying values of the arguments to match
+In some cases, it's useful to remove only specific usage of the attribute. This can be achieved by specifying the value
+or values of the arguments to match. In the example below only `System.Reflection.AssemblyMetadataAttribute` custom attributes
+with the first argument equal to `RemovableValue` will be removed.
 
 ```xml
 <linker>
--- a/docs/data-formats.md
+++ b/docs/data-formats.md
@@ -528,17 +528,33 @@ attributes are applied.
 <attribute fullname="SomecustomAttribute" feature="EnableOptionalFeature" featurevalue="false"/>
 ```
 
-### Special custom attributes
+### Removing custom attributes
 
-Also if the attribute is used in a type, a special property can be used to specify that the type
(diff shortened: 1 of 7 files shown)


docs/data-formats.md
```diff
@@ -528,22 +528,38 @@ attributes are applied.
...
-  <assembly fullname="*"> 
+  <assembly fullname="*">
     <type fullname="System.Runtime.CompilerServices.NullableAttribute">
       <attribute internal="RemoveAttributeInstances" feature="EnableOptionalFeature" featurevalue="false" />
     </type>
   </assembly>
 </linker>
 ```
 
+In some cases it's useful to remove only specific instances of same attribute. This can be achieved by specifying values of the arguments to match
```
vitek-karas, line 547: This should say that the sample will only remove `[AssemblyMetadataValue("RemovableValue")]` but leave alone for example `[AssemblyMetadataValue("KeptValue")]`.

test/Mono.Linker.Tests.Cases/LinkAttributes/LinkerAttributeRemovalConditional.cs
```diff
@@ -0,0 +1,91 @@
...
+		[Kept]
+		[KeptAttributeAttribute (typeof (TestConditionalRemoveAttribute))]
+		[TestConditionalRemove (1, true)]
+		static void Kept_3 ()
+		{
+		}
+
+		[Kept]
+		[KeptAttributeAttribute (typeof (TestConditionalRemoveAttribute))]
+		[TestConditionalRemove ("remove", 1)]
```
vitek-karas, line 49: Please add a comment why this is not getting removed - the only reason I found possible was that the parameter is typed `object` in this case, which probably doesn't match the `string` we have in the XML. If that's the case I find this rather weird, because the values match exactly.

vitek-karas (approved):
Aside from the weird behavior of object<>String, this looks good

docs/data-formats.md
marek-safar: Updated the explanation

test/Mono.Linker.Tests.Cases/LinkAttributes/LinkerAttributeRemovalConditional.cs
marek-safar: Good observation. I changed the tests to be easier to understand.

vitek-karas (approved):

PR #1936, opened 2021-04-01
Revert "Change behavior of copy action to not rewrite assemblies"

mateoatr:
Reverts mono/linker#1869

Reverting from now. I'm still working on an edge case that I found while linking a blazorwasm app just after `System.Runtime.CompilationServices.Unsafe` was marked as `IsTrimmable`: custom attributes having enum arguments living in assemblies that are forwarded seem to have problems -- so far, I've been able to circumvent this by marking the forwarded scopes of CAs arguments, but I still need to add appropriate tests and measure impact with this changes.


sbomer:
Also worth noting that System.Runtime.CompilationServices.Unsafe was one of the things keeping the dependency on System.Runtime (the other being the app itself). Marking it IsTrimmable will probably shave more size off System.Runtime.

sbomer (approved):

marek-safar:
Makes sense (I didn't know about that so sorry for the extra work)

PR #1937, opened 2021-04-01
Support collapsing trim analysis warnings

sbomer:
This adds an option to allow collapsing trim analysis warnings either globally or per-assembly (collapsing means that there will be only one warning shown for an assembly that produces trim analysis warnings).

The flag `--collapse` collapses warnings for all assemblies, `--collapse AssemblyName` only for that assembly. This also supports passing `--collapse-` similar to warnaserror. The default in the command-line will be not to collapse any warnings, but the SDK will pass the right flags to show detailed warnings for the project and projectreferences, but collapsed warnings for packagereferences or framework libraries (for example: `--collapse --collapse- MyProject --collapse- MyProjectRef`).

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -1569,9 +1569,9 @@ void TestMethod(MethodInfo methodInfo)
 }
 ```
 
-#### `IL2104`: Assembly 'assembly' produced trim warnings
+#### `IL2104`: Assembly 'assembly' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries
 
-The assembly 'assembly' produced trim analysis warnings in the context of the app. This means the assembly has not been fully annotated for trimming. Consider contacting the library author to request they add trim annotations to the library. To see detailed warnings for this assembly, turn off warning collapsing by passing either `--collapse-` to show detailed warnings for all assemblies, or `--collapse- "assembly"` to show detailed warnings for that assembly.
+The assembly 'assembly' produced trim analysis warnings in the context of the app. This means the assembly has not been fully annotated for trimming. Consider contacting the library author to request they add trim annotations to the library. To see detailed warnings for this assembly, turn off warning collapsing by passing either `--collapse-` to show detailed warnings for all assemblies, or `--collapse- "assembly"` to show detailed warnings for that assembly. https://aka.ms/dotnet-illink/libraries has more information on annotating libraries for trimming.
 
 ## Single-File Warning Codes
 
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -1569,6 +1569,10 @@ void TestMethod(MethodInfo methodInfo)
 }
 ```
 
+#### `IL2104`: Assembly 'assembly' produced trim warnings
+
+The assembly 'assembly' produced trim analysis warnings in the context of the app. This means the assembly has not been fully annotated for trimming. Consider contacting the library author to request they add trim annotations to the library. To see detailed warnings for this assembly, turn off warning collapsing by passing either `--collapse-` to show detailed warnings for all assemblies, or `--collapse- "assembly"` to show detailed warnings for that assembly.
+
 ## Single-File Warning Codes
 
 #### `IL3000`: 'member' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -1571,7 +1571,7 @@ void TestMethod(MethodInfo methodInfo)
 
 #### `IL2104`: Assembly 'assembly' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries
 
-The assembly 'assembly' produced trim analysis warnings in the context of the app. This means the assembly has not been fully annotated for trimming. Consider contacting the library author to request they add trim annotations to the library. To see detailed warnings for this assembly, turn off warning collapsing by passing either `--collapse-` to show detailed warnings for all assemblies, or `--collapse- "assembly"` to show detailed warnings for that assembly. https://aka.ms/dotnet-illink/libraries has more information on annotating libraries for trimming.
+The assembly 'assembly' produced trim analysis warnings in the context of the app. This means the assembly has not been fully annotated for trimming. Consider contacting the library author to request they add trim annotations to the library. To see detailed warnings for this assembly, turn off grouped warnings by passing either `--singlewarn-` to show detailed warnings for all assemblies, or `--singlewarn- "assembly"` to show detailed warnings for that assembly. https://aka.ms/dotnet-illink/libraries has more information on annotating libraries for trimming.
 
 ## Single-File Warning Codes
 
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -23,7 +23,7 @@ namespace ILLink.Tasks
 		///       UnusedInterfaces
 		///       IPConstProp
 		///       Sealer
-		///   Optional metadata "CollapseTrimWarnings" may also be set to "True"/"False" to control
+		///   Optional metadata "TrimmerSingleWarn" may also be set to "True"/"False" to control
 		///   whether the linker produces granular warnings for this assembly.
 		///   Maps to '-reference', and possibly '--action', '--enable-opt', '--disable-opt', '--verbose'
 		/// </summary>
@@ -76,11 +76,11 @@ namespace ILLink.Tasks
 		bool? _treatWarningsAsErrors;
 
 		/// <summary>
-		/// Collapse all warnings to produce one per assembly.
-		/// Maps to '--collapse' if true, '--collapse-' if false.
+		/// Produce at most one trim analysis warning per assembly.
+		/// Maps to '--singlewarn' if true, '--singlewarn-' if false.
 		/// </summary>
-		public bool CollapseWarnings { set => _collapseWarnings = value; }
-		bool? _collapseWarnings;
+		public bool SingleWarn { set => _singleWarn = value; }
+		bool? _singleWarn;
 
 		/// <summary>
 		/// The list of warnings to report as errors.
@@ -297,11 +297,11 @@ namespace ILLink.Tasks
 				args.AppendLine ();
 			}
 
-			if (_collapseWarnings is bool collapse) {
-				if (collapse)
-					args.AppendLine ("--collapse");
+			if (_singleWarn is bool generalSingleWarn) {
+				if (generalSingleWarn)
+					args.AppendLine ("--singlewarn");
 				else
-					args.AppendLine ("--collapse-");
+					args.AppendLine ("--singlewarn-");
 			}
 
 			HashSet<string> assemblyNames = new HashSet<string> (StringComparer.OrdinalIgnoreCase);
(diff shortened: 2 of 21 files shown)


src/ILLink.Tasks/LinkTask.cs
```diff
@@ -23,7 +23,9 @@ public class ILLink : ToolTask
 		///       UnusedInterfaces
 		///       IPConstProp
 		///       Sealer
-		///   Maps to '-reference', and possibly '--action', '--enable-opt', '--disable-opt'
+		///   Optional metadata "CollapseTrimWarnings" may also be set to "True"/"False" to control whether
```
mateoatr, line 26:
```suggestion
		///   Optional metadata "CollapseTrimWarnings" may also be set to "True"/"False" to control
```

src/linker/Linker/Driver.cs
```diff
@@ -500,6 +500,41 @@ protected int SetupContext (ILogger customLogger = null)
 
 						continue;
 
+					case "--collapse":
+					case "--collapse+": {
+							string assemblyName = GetNextStringValue ();
+							if (!string.IsNullOrEmpty (assemblyName)) {
+								if (!IsValidAssemblyName (assemblyName)) {
```
mateoatr, line 507: Maybe I'm missing something, but I don't think this will ever evaluate true. Unrelated to your changes, but I'm also a little bit confused why we need this `IsValidAssemblyName` method? It doesn't seem to be evaluating the validity of anything.

src/linker/Linker/Driver.cs
```diff
@@ -500,6 +500,41 @@ protected int SetupContext (ILogger customLogger = null)
...
+								context.CollapseWarnings.Clear ();
+							}
+
+							continue;
+						}
+
+					case "--collapse-": {
+							string assemblyName = GetNextStringValue ();
+							if (!string.IsNullOrEmpty (assemblyName)) {
+								if (!IsValidAssemblyName (assemblyName)) {
```
mateoatr, line 524: Same as above.

src/linker/Linker/MessageContainer.cs
```diff
@@ -132,6 +132,16 @@ private static MessageContainer CreateWarningMessageContainer (LinkContext conte
 			if (version > context.WarnVersion)
 				return Empty;
 
+			if (origin.MemberDefinition != null) {
+				var assembly = origin.MemberDefinition.DeclaringType.Module.Assembly;
+				var assemblyName = assembly.Name.Name;
+				if (context.AreWarningsCollapsed (assemblyName)) {
+					if (context.WarningAssemblies.Add (assemblyName))
+						context.LogWarning ($"Assembly '{assemblyName}' produced trim warnings", 2104, context.GetAssemblyLocation (assembly));
```
mateoatr, line 140: Should we set the origin to be the assembly so that the warning can be easily suppressed?
```C#
[assembly: UnconditionalSuppressMessage ("SuppressThis", "IL2104")]
```

mateoatr: I'm not sure what should we do here, maybe it's better to leave it as is but I'm curious what the experience should be for a user that's consuming a third-party library that is not yet annotated with `IsTrimmable`, do we expect them to not trim the library at all after seeing this warning, or should we instruct them to turn on granular warnings to decide whether or not they're using code that is dangerous to trim?  And for the later case I suppose that after deciding, the user would get rid of this warning by setting `<NoWarn>IL2104</NoWarn>`?

sbomer: > what the experience should be for a user that's consuming a third-party library that is not yet annotated with `IsTrimmable`
The defaults ("copy") will warning for everything in the assembly, so they'd have a few options:
- stop trimming the app
- trim the assembly and hope the warning goes away (which it might if it originates from unused code)
- turn on granular warnings and investigate

We don't have currently have a good UX around dealing with warnings in assemblies without source code, so they would have to either use NoWarn to suppress the (granular or collapsed) warnings globally, or figure out how to use the attribute XML.

This does set the origin to the assembly path, so you at least get warnings like:
```
/home/sven/.nuget/packages/depnuget/1.0.0/lib/net6.0/depnuget.dll : warning IL2104: Assembly 'depnuget' produced trim warnings
```

src/linker/Linker/Driver.cs
sbomer: Thanks. I think the intention of IsValidAssemblyName is to catch empty strings passed on the command-line, so I updated the outer check to check null only. This way if you pass `--collapse ""` you get a warning about the assembly name.

sbomer:
@agocke, @vitek-karas PTAL in particular at the wording of the collapsed warning.

docs/error-codes.md
```diff
@@ -1569,6 +1569,10 @@ void TestMethod(MethodInfo methodInfo)
 }
 ```
 
+#### `IL2104`: Assembly 'assembly' produced trim warnings
+
+The assembly 'assembly' produced trim analysis warnings in the context of the app. This means the assembly has not been fully annotated for trimming. Consider contacting the library author to request they add trim annotations to the library. To see detailed warnings for this assembly, turn off warning collapsing by passing either `--collapse-` to show detailed warnings for all assemblies, or `--collapse- "assembly"` to show detailed warnings for that assembly.
```
agocke, line 1574: Seems long. I wonder if we can shorten it without distorting the message too much.

> The assembly 'assembly' is not trim-safe and may reference members that are trimmed away in the final app. Consider contacting the library author to request they add trim annotations to the library. For more information, see \<link to docs>

I'm not sure seeing the detailed warnings is that helpful, so I would suggest putting that in the docs, along with instructions for the library author who may get this info after the user contacts them.

agocke: My mistake, I think I read this as the warning message, when it's just the warning description.

In that case, I think the message `Assembly 'assembly' produced trim warnings` is too brief. We should either include a link for more information, or include a recommendation to contact the author in the message directly.

vitek-karas (approved):

docs/error-codes.md
```diff
@@ -1569,6 +1569,10 @@ void TestMethod(MethodInfo methodInfo)
 }
 ```
 
+#### `IL2104`: Assembly 'assembly' produced trim warnings
```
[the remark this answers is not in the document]
sbomer: @samsp-msft This change is making it so that app authors will get at most one trim analysis warning from nuget libraries by default. PTAL at the wording of the warning message and aka.ms link, currently:
```
`IL2104`: Assembly 'assembly' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries
```
The link will point to docs that:
- recommend the app author contact the library author about making the library trim-friendly
- instructions for library authors to make their libraries trim-friendly

Note we have an existing link, https://aka.ms/dotnet-illink, which is the landing page for the linker docs, hence the "/libraries".

src/linker/Linker/Driver.cs
```diff
@@ -1188,6 +1223,8 @@ static void Usage ()
 			Console.WriteLine ("                              VERSION is an integer in the range 0-9999.");
 			Console.WriteLine ("  --warnaserror[+|-]        Report all warnings as errors");
 			Console.WriteLine ("  --warnaserror[+|-] WARN   Report specific warnings as errors");
+			Console.WriteLine ("  --collapse[+|-]           Collapse analysis warnings to one per assembly");
```
marek-safar, line 1226: To me `--collapse System.Runtime` option is anything but a warning setting. We should change the option to be more descriptive because if I understood the code correctly it does not collapse anything but replaces any warning with a generic warning. What about something like `--warnasone` or `--singlewarn` ?

src/linker/Linker/MessageContainer.cs
```diff
@@ -132,6 +132,16 @@ private static MessageContainer CreateWarningMessageContainer (LinkContext conte
 			if (version > context.WarnVersion)
 				return Empty;
 
+			if (origin.MemberDefinition != null) {
+				var assembly = origin.MemberDefinition.DeclaringType.Module.Assembly;
+				var assemblyName = assembly.Name.Name;
+				if (context.AreWarningsCollapsed (assemblyName)) {
```
marek-safar, line 138: Shouldn't this be the first check?

src/linker/Linker/MessageContainer.cs
```diff
@@ -132,6 +132,16 @@ private static MessageContainer CreateWarningMessageContainer (LinkContext conte
 			if (version > context.WarnVersion)
 				return Empty;
 
+			if (origin.MemberDefinition != null) {
```
marek-safar, line 135: How many warnings have origin.MemberDefinition == null?

src/linker/Linker/MessageContainer.cs
```diff
@@ -132,6 +132,16 @@ private static MessageContainer CreateWarningMessageContainer (LinkContext conte
...
+				var assembly = origin.MemberDefinition.DeclaringType.Module.Assembly;
+				var assemblyName = assembly.Name.Name;
+				if (context.AreWarningsCollapsed (assemblyName)) {
+					if (context.AssembliesWithGeneratedCollapsedWarnings.Add (assemblyName))
+						context.LogWarning ($"Assembly '{assemblyName}' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries", 2104, context.GetAssemblyLocation (assembly));
```
marek-safar, line 140: Is the sentence correct? The assembly itself didn't produce any warnings, right? Something like `Trimming warnings for assembly '{assemblyName} were suppressed.' would make more sense to me.

marek-safar (changes requested):

src/linker/Linker/MessageContainer.cs
sbomer: It's a little hard to tell because we pass around the source location through multiple helpers, but I believe the trim analysis warnings will have a MemberDefinition. We don't have a source location for assembly-level attributes, but I can't think of a case where we can produce dataflow warnings from these.

I made the check more specific (checking the subcategory) and added an assert for this.

src/linker/Linker/MessageContainer.cs
sbomer: Not sure what you mean exactly, but https://github.com/mono/linker/pull/1937#discussion_r607275309 may address it. This is only trying to collapse the trim analysis warnings (which we should be able to "blame" on a particular assembly).

src/linker/Linker/MessageContainer.cs
sbomer: How do you feel about "Assembly 'asm' _has_ trim warnings"?

I'd avoid "suppressed" since that has an existing specific meaning. It's true that the assembly isn't technically producing warnings (the analysis logic is) - but I also don't like the passive voice "Trimming warnings were produced...". And "Trim analysis produced warnings for assembly 'asm'" seems a bit wordy and puts the source of the warning at the end of the message. Other verbs we could use are "caused" or "originated".

I'm having trouble picking the best wording here. Maybe @agocke has suggestions.

src/linker/Linker/Driver.cs
sbomer: Changed to `--singlewarn`. Thanks!

src/linker/Linker/MessageContainer.cs
sbomer: Oh, and regarding the suggestion in https://github.com/mono/linker/pull/1937#discussion_r606382945:
> The assembly 'assembly' is not trim-safe...

I am not sure if we want to talk about it in terms of "safety" since that has other connotations. IIRC @jkotas made this point in past discussions.

marek-safar (approved):

PR #1938, opened 2021-04-01
Move remaining special serialization handling under same guard check

marek-safar:


vitek-karas (approved):
Looks good, but I'm not sure if it does what we discussed. I though we wanted to only mark these things in the library mode. If I read it correctly this will mark them always. (But it does make it more consistent, which is good)

marek-safar:
> If I read it correctly this will mark them always. (But it does make it more consistent, which is good)

Only for pre-net6 configuration to keep the previous behaviour

PR #1941, opened 2021-04-05
Copy action behavior

mateoatr:
This PR contains all changes previously introduced in #1869 and fixes an edge case that caused us to revert the merged PR:

> SweepStep removes the (unused) enum forwarder before it tries to fix the scope of the attribute arg, so fixing the scope fails

Final size impact on a blazorwasm app:

Total size: 2240 KB (used to be 2228 KB).

```diff
   17408 Microsoft.AspNetCore.Components.Web.dll
   48128 Microsoft.AspNetCore.Components.WebAssembly.dll
  121856 Microsoft.AspNetCore.Components.dll
    6144 Microsoft.Extensions.Configuration.Abstractions.dll
    8704 Microsoft.Extensions.Configuration.Json.dll
    7680 Microsoft.Extensions.Configuration.dll
   12800 Microsoft.Extensions.DependencyInjection.Abstractions.dll
   38400 Microsoft.Extensions.DependencyInjection.dll
   24064 Microsoft.Extensions.Logging.Abstractions.dll
   17408 Microsoft.Extensions.Logging.dll
   14848 Microsoft.Extensions.Options.dll
    8704 Microsoft.Extensions.Primitives.dll
    8704 Microsoft.JSInterop.WebAssembly.dll
   30720 Microsoft.JSInterop.dll
   17408 System.Collections.Concurrent.dll
   66560 System.Collections.Immutable.dll
!  22016 System.Collections.dll (before: 22528)
    4608 System.ComponentModel.dll
   13824 System.Console.dll
   25088 System.Linq.dll
!  17920 System.Memory.dll (before: 18432)
   10752 System.Net.Http.Json.dll
  138752 System.Net.Http.dll
    7168 System.Net.Primitives.dll
    7680 System.ObjectModel.dll
!1116672 System.Private.CoreLib.dll (before: 1117184)
   21504 System.Private.Runtime.InteropServices.JavaScript.dll
   65536 System.Private.Uri.dll
    3072 System.Runtime.CompilerServices.Unsafe.dll
-   5632 System.Runtime.dll
   29696 System.Text.Encodings.Web.dll
  239104 System.Text.Json.dll
   17408 foo.dll
+  21860 foo.pdb (before: 13032)
```


marek-safar (approved):

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -63,6 +63,17 @@ protected override void Process ()
 			foreach (var assembly in assemblies)
 				UpdateAssemblyReferencesToRemovedAssemblies (assembly);
 
+			// Update scopes before removing any type forwarder.
+			foreach (var assembly in assemblies) {
+				switch (Annotations.GetAction (assembly)) {
+				case AssemblyAction.CopyUsed:
```
sbomer, line 69: Updating scopes in a copyused assembly is new, but I think this is a good change.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -172,22 +182,14 @@ protected void ProcessAssemblyAction (AssemblyDefinition assembly)
 				break;
 
 			case AssemblyAction.CopyUsed:
-				Annotations.SetAction (assembly, AssemblyAction.Copy);
-				goto case AssemblyAction.Copy;
+				AssemblyAction assemblyAction = AssemblyAction.Copy;
```
sbomer, line 185: If we are updating scopes in a copyused assembly, we shouldn't change its action to copy.

src/linker/Linker.Steps/SweepStep.cs
sbomer: On second thought it might be best to leave it as before purely for back-compat. Some scenarios might depend on the old behavior that copies "copyused" assemblies in some cases (when there were no removed references).

sbomer (approved):
A couple more comments around the copyused edge case, otherwise LGTM. Thanks a lot!

src/linker/Linker.Steps/SweepStep.cs
mateoatr: Thanks. Makes sense -- I changed this so we keep same behavior for copyused assemblies.

PR #1945, opened 2021-04-08
Use CodeFixProvider project as analyzer package project

agocke:
Bundles both the analyzer and codefix together as analyzers.

### DIFF
--- a/src/ILLink.CodeFix/ILLink.CodeFixProvider.csproj
+++ b/src/ILLink.CodeFix/ILLink.CodeFixProvider.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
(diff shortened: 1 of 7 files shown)


src/ILLink.CodeFix/ILLink.CodeFixProvider.csproj
```diff
@@ -2,15 +2,33 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <IsPackable>false</IsPackable>
```
sbomer, line 5: I think we should set `AutoGenerateAssemblyVersion` to follow [this](https://github.com/dotnet/arcade/blob/6cc4c1e9e23d5e65e88a8a57216b3d91e9b3d8db/Documentation/CorePackages/Versioning.md#recommended-settings) arcade recommendation.

sbomer (approved):
Thanks!

PR #1946, opened 2021-04-09
First draft of blog post on fixing warnings

agocke:


docs/fixing-warnings.md
```diff
@@ -0,0 +1,186 @@
...
+
+There are two big categories of warnings which you will likely see:
+
+ 1. `RequiresUnreferencedCode`
+ 2. `DynamicallyAccessedMembers`
+
+### RequiresUnreferencedCode
+
+`RequiresUnreferencedCode` is simple: it's an attribute that can be placed on methods to indicate
+that the method is not trim-safe, meaning that it might use reflection or some other mechanism
```
vitek-karas, line 54: Based on previous discussions let's try to stay away from "Safe", or "Safety" terms. They tend to imply security related meaning... which would be confusing at best.
We've been trying to use the term "trim-compatible" instead.

docs/fixing-warnings.md
```diff
@@ -0,0 +1,186 @@
...
+
+[UnconditionalSuppressMessage("AssemblyLoadTrimming", "IL2026:RequiresUnreferencedCode",
+    Justification = "I've verified the loaded assembly only adds integers, so it's safe")]
+void TestMethod()
+{
+    MethodWithAssemblyLoad(); // Warning suppressed
+}
+```
+
+`UnconditionalSuppressMessage` is preserved into IL, so the trimmer can see the suppression even
```
vitek-karas, line 96: I think we should mention the relation to `SuppressMessageAttribute` - you kind of already do that, I would just make it more explicit. And maybe also mention that pragmas won't work either.

docs/fixing-warnings.md
```diff
@@ -0,0 +1,186 @@
...
+trimmed away, you can also suppress the warning using
+[UnconditionalSuppressMessageAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.unconditionalsuppressmessageattribute?view=net-5.0).
+For example:
+
+```C#
+[RequiresUnreferencedCode("Use 'MethodFriendlyToTrimming' instead")]
+void MethodWithAssemblyLoad() { ... }
+
+[UnconditionalSuppressMessage("AssemblyLoadTrimming", "IL2026:RequiresUnreferencedCode",
+    Justification = "I've verified the loaded assembly only adds integers, so it's safe")]
```
sbomer, line 89: I think it's more important that the assembly and its required code is kept, something like "I've verified the loaded assembly is kept by the trimmer for other reasons". (And any marked code in it would produce warnings if it had non-understood patterns).

docs/fixing-warnings.md
```diff
@@ -0,0 +1,186 @@
...
+string s = Console.ReadLine();
+Type type = Type.GetType(s);
+foreach (var m in type.GetMethods())
+{
+    Console.WriteLine(m.Name);
+}
+```
+
+In this example, `Type.GetType` dynamically requests a type with an unknown name, and then prints
+the names of all of its members. Because there's no way to know at publish time what type name is
```
tlakollo, line 26: Nit: Since there is a GetMembers() that gets the names of all of its members, we may want to rephrase it as "all of its methods" or "all of its method members" instead

docs/fixing-warnings.md
```diff
@@ -0,0 +1,189 @@
...
+trimmed away, you can also suppress the warning using
+[UnconditionalSuppressMessageAttribute](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.unconditionalsuppressmessageattribute?view=net-5.0).
+For example:
+
+```C#
+[RequiresUnreferencedCode("Use 'MethodFriendlyToTrimming' instead")]
+void MethodWithAssemblyLoad() { ... }
+
+[UnconditionalSuppressMessage("AssemblyLoadTrimming", "IL2026:RequiresUnreferencedCode",
+    Justification = "Everything referenced in the loaded assembly is manually preserved, so it's safe")]
```
tlakollo, line 90: Should we explain/link to how to preserve things manually. If I were to read this blog post looking for answers on how to solve my current broken app I would see that I can:
- Use other function that is trim compatible (which is not always possible)
- Mark my method with RequiresUnreferencedCode which changes the warning location but my app would still be broken
- Preserve manually and Suppress the warning, the blog explains how to suppress the warning, but not how to preserve manually.

agocke: A good question, but not really a post I'm prepared to write.

tlakollo: Well just in case someone reads these comments, an XML descriptor file or DynamicDependencyAttribute could help to preserve stuff on the app

vitek-karas (approved):

tlakollo (approved):
Looks good to me

docs/fixing-warnings.md
```diff
@@ -0,0 +1,189 @@
+
+# Trim warnings in .NET 6
+
+In .NET Core 3.1 and 5.0 we introduced trimming as a new preview feature for self-contained .NET
+core applications. Conceptually the feature is very simple: when publishing the application the
+.NET SDK analyzes the entire application, including the .NET framework, and removes all unused
```
marek-safar, line 6: I'm not sure what "including the .NET framework" is intended to explain

docs/fixing-warnings.md
```diff
@@ -0,0 +1,189 @@
...
+code. In the time trimming has been in preview, we've learned that trimming is very powerful --
+it can reduce application size by half or more. However, we've also learned about the
+difficulties in safely trimming applications.
+
+The most difficult question in trimming is what is unused, or more precisely, what is used. For
+most standard C# code this is trivial -- the trimmer can easily walk method calls, field and
```
marek-safar, line 12: What is "the trimmer" ? We have not been using that term and most developers are familiar with linker term. Do we really have to confuse the community with another new name?

docs/fixing-warnings.md
```diff
@@ -0,0 +1,189 @@
...
+trimmer encounters code like that, that's when you should expect a trim warning.
+
+There are two big categories of warnings which you will likely see:
+
+ 1. `RequiresUnreferencedCode`
+ 2. `DynamicallyAccessedMembers`
+
+### RequiresUnreferencedCode
+
+`RequiresUnreferencedCode` is simple: it's an attribute that can be placed on methods to indicate
```
marek-safar, line 53: It could reference MSDN documentation at https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.requiresunreferencedcodeattribute

docs/fixing-warnings.md
```diff
@@ -0,0 +1,189 @@
...
+`UnconditionalSuppressMessage` is like `SuppressMessage` but it is preserved into IL, so the
+trimmer can see the suppression even after build and publish. `SuppressMessage` and `#pragma`
+directives are only present in source so they can't be used to silence warnings from the
+trimmer. Be very careful when suppressing trim warnings: it's possible that the call may be
+trim-compatible now, but as you change your code that may change and you may forget to review all
+the suppressions.
+
+### DynamicallyAccessedMembers
+
+`DynamicallyAccessedMembers` is usually about reflection. Unlike `RequiresUnreferencedCode`,
```
marek-safar, line 106: Add a reference to MSDN documentation for DynamicallyAccessedMembers

docs/fixing-warnings.md
```diff
@@ -0,0 +1,189 @@
...
+
+As with `RequiresUnreferencedCode`, adding `RequiresUnreferencedCode` or
+`UnconditionalSuppressMessage` attributes also works, but none of these options make the code
+compatible with trimming, while adding `DynamicallyAccessedMembers` does.
+
+## Conclusion
+
+This description should cover the most common situations you end up in while trimming your
+application. Over time we'll continue to improve the diagnostic experience and provide more streamlined
+ways to make your code trim-compatible.
```
marek-safar, line 184: > Over time we'll continue to improve the diagnostic experience and provide more streamlined
ways to make your code trim-compatible.

Is this claim correct?

docs/fixing-warnings.md
vitek-karas: In all locations other than `mono/linker` we've been trying to avoid the term "linker" because most people understand it as the "platform linker". We use "trim tool" or sometimes "trimmer" (I would prefer "trim tool" if possible).
For the same reason we use "trimming" and not "linking" everywhere.

marek-safar: > we've been trying to avoid the term "linker" because most people understand it as the "platform linker"

Are you sure? Few references to public documentation that clearly uses IL Linker term to describe the tool.

- https://docs.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/configure-linker?view=aspnetcore-3.1
- https://www.telerik.com/blogs/reduce-self-contained-app-size-with-latest-net-core-features
- https://www.meziantou.net/optimizing-a-blazor-webassembly-application-size.htm
- https://montemagno.com/optimizing-xamarin-apps-libraries-with-the-linker/
- https://docs.telerik.com/devtools/xamarin/knowledge-base/common-understanding-ui-for-xamarin-and-linker-behavior

vitek-karas: I'm intentionally "ignoring" the mono/Xamarin side of things - those will likely use "linker" a lot. Most of our docs and blog posts use "Trimming" - we also use it everywhere in the messages (RUC) in the framework. I don't see us going back to "linking".

That said - I understand the confusion for Xamarin users - we might want to mention this on the first use of the term.

marek-safar: I was not referring to trimming but to "trimmer". Even .NET Core MSDN documentation is written in the way that it uses the term since 3.1.

> .NET Core 3+ ships with an IL Linker which allows you to reduce the size of the self-contained application. It automatically determines which assemblies are needed and copies only them.

I'm really not sure we have to rename everything to cause more confusion.

MichalStrehovsky: I would probably just not give it a name here. We can update the sentence to say "For most standard C# code this is trivial -- ~~the trimmer~~we can easily walk method calls, field and"

That said,

https://en.wikipedia.org/wiki/Linker_(computing)

This is the definition of the term used in computer engineering. It's not what IL Linker is doing. It's confusing to anyone who has ever used a linker in the past.

Reusing this term in .NET was a mistake. We should have gone with something like [Remover 8000](https://github.com/mono/embeddinator-4000). Now Xamarin often needs to qualify the word "linker" with the word "native" when referring to what everyone outside .NET calls "linker". It's a pretty sad place to be in and puts us (the authors of .NET) into a bad light (trying to steal a well-established term).

I've seen this cause confusion in the past. "Linker failed. Oh, not that one, the other one."

We should phase this name out before it moves out of the niche parts of .NET (both Xamarin and Blazor-WASM are still niches).

docs/fixing-warnings.md
```diff
@@ -0,0 +1,189 @@
...
+    // can break functionality when trimming application code. Use 'MethodFriendlyToTrimming' instead.
+    MethodWithAssemblyLoad();
+}
+```
+
+There aren't many workarounds for `RequiresUnreferencedCode`. The best way is to avoid calling
+the method at all when trimming and use something else which is trim-compatible. If you're
+writing a library and it's not in your control whether or not to call the method and you just
+want to communicate to *your* caller, you can also add `RequiresUnreferencedCode` to your own
+method. This silences all trimming warnings in your code, but will produce a warning whenever
```
samsp-msft, line 77: Why does this produce a warning rather than an error? If the method is marked that its not going to work, and is called, then why only warn?

docs/fixing-warnings.md
```diff
@@ -0,0 +1,189 @@
...
+`DynamicallyAccessMembers` come in? What happens if the reflection is split across methods?
+
+```C#
+void M1()
+{
+    M2(typeof(System.Tuple));
+}
+void M2(Type type)
+{
+    // Trim analysis warning IL2070: net6.Program.M2(Type): 'this' argument does not satisfy
```
samsp-msft, line 142: Move out the error into a block to show what the compiler would say

docs/fixing-warnings.md
```diff
@@ -0,0 +1,189 @@
...
+    Console.WriteLine(m.Name);
+}
+```
+
+This would be fine. Here the trimmer can see the exact type being referenced: `System.Tuple`. Now
+it can use flow analysis to determine that it needs to keep all public methods. So where does
+`DynamicallyAccessMembers` come in? What happens if the reflection is split across methods?
+
+```C#
+void M1()
```
samsp-msft, line 136: M1 seems wrong to me, Method1 would read better

samsp-msft:
A couple of niggles, but LGTM.

docs/fixing-warnings.md
agocke: I meant to emphasize that we're trimming away the core framework, but it's probably not necessary.

docs/fixing-warnings.md
agocke: My use of the term "trimmer" here is in its most generic sense -- it is the thing which "trims." The other documentation about a specific tool which is responsible for trimming is incorrect in this case because the only supported method of trimming applications in .NET 6 will be passing `PublishTrimmed=true` to the SDK. Rather than talk specifically about the SDK, I wanted to just use a generic term.

I don't really see the need to get more specific here and talk about any specific tool. The term "linker" is certainly incorrect for .NET 6 since we do not plan on shipping a separate linker tool for supported scenarios.

docs/fixing-warnings.md
vitek-karas: Because in almost all cases the method might work in some cases and might not in others. For example `XmlSerializer.Serialize(object o)` - it's definitely not trim compatible, but if you make sure that the type passed in is preserved enough - it will work just fine.

We use this attribute basically to say "If you ignore this, you're on your own and will have to test at runtime".

There are also cases where the method is sooo close to being trim compatible except for some weird special case - we can't mark it as compatible (since it's not 100%), but it almost always is.

And as with any other trim analysis - the linker plays it super safe, the annotations do that as well. We don't want to block people completely if we are too careful.

docs/fixing-warnings.md
agocke: Yup! Better warning messages, analyzers for `DynamicallyAccessedMembers`, better perf, etc

docs/fixing-warnings.md
agocke: Pretty much everything in trimming is a warning, because many things just happen to work. For instance, if you call a method that has RUC does a lot of reflection on its parameter, but the parameter type is in user code (and therefore untrimmed) and nothing important from the framework was removed, it all just works.

Of course, this is just luck, which is why the warning is there.

docs/fixing-warnings.md
agocke: Sorry, not sure what you mean by this. You mean extract the comment into a separate code block?

agocke:
@samsp-msft Any idea where this should go? I'd like to put it somewhere more official than the linker docs. Either MS docs or blog post. Thoughts?

docs/fixing-warnings.md
samsp-msft: Yes, and you can mark it as console, something like:

-----

```c#
void M1()
{
    M2(typeof(System.Tuple));
}
void M2(Type type)
{
    var methods = type.GetMethods();
    ...
}
```

If you compile, you'll see the following warning:

```console
Trim analysis warning IL2070: net6.Program.M2(Type): 'this' argument does not satisfy
'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethods()'. The
parameter 'type' of method 'net6.Program.M2(Type)' does not have matching annotations. The
source value must declare at least the same requirements as those declared on the target
location it is assigned to.
```

samsp-msft:
> @samsp-msft Sam Spencer FTE Any idea where this should go? I'd like to put it somewhere more accessible than the linker docs. Either MS docs or blog post. Thoughts?

It reads more like a blog post than docs, but doesn't have the preamble of a blog post. I'd suggest either
- Turn it into an update on trimming blog post - put a bit more up front about changes in trimming
- or put it in the trimming docs - https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options
- ideally, both

marek-safar (approved):

agocke:
@samsp-msft I think this is ready. Not sure what kind of preface you would want to add, but can we start moving this to a blog site?

PR #1950, opened 2021-04-09
Remove some TODO comments - instead file issues

vitek-karas:
Also resolved one of the TODOs.

### DIFF
--- a/src/linker/Linker.Steps/LinkAttributesParser.cs
+++ b/src/linker/Linker.Steps/LinkAttributesParser.cs
@@ -280,6 +280,7 @@ namespace Mono.Linker.Steps
 
 				return new CustomAttributeArgument (typeref, type);
 			default:
+				// No support for null and arrays, consider adding - mono/linker/issues/1957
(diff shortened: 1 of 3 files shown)


src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -281,8 +280,6 @@ CustomAttributeArgument[] ReadCustomAttributeArguments (XPathNodeIterator iterat
 
 				return new CustomAttributeArgument (typeref, type);
 			default:
-				// TODO: Add support for null values
```
marek-safar, line 284: I'd keep it, these TODOs are still valid

marek-safar (approved):

src/linker/Linker.Steps/LinkAttributesParser.cs
vitek-karas: They are - but they're basically a feature request. I'm just not a fan of TODOs in merged code - it's either "bug", "future feature" or "nice to have". But on their own they're just confusing - when I find one, I always try to figure out which one it is and if I should care or not.

PR #1955, opened 2021-04-10
Support running a basic analysis of generic instantiations created through reflection.

jkoritzinsky:
Enable a basic analysis for arrays of `System.Type` passed to `Type.MakeGenericType` and `MethodInfo.MakeGenericMethod`	. This analysis works for arrays created within the method with a constant size where all array elements are known.

This allows for common usages of `Type.MakeGenericType` and `MethodInfo.MakeGenericMethod` that let the compiler implicitly create an array for them (since the parameter is `params`) to get basic analysis.

Also enable analysis for ldelem when it's a known array with a known value at a given known index.

In combination with #1930, we could also enable this validation for `Expression.Call` on generic types.


jkoritzinsky:
Some of the data flow tests are now failing since more object are now analyzable (since arrays can be analyzed now).

I'm not sure the best way to change the test code to still be unanalyzable.

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -1239,23 +1252,36 @@ public override bool Equals (ValueNode other)
...
+				yield return new ArrayValue (sizeConst, ElementType, IndexValues);
 		}
 
 		protected override ValueNode ChildAt (int index)
 		{
 			if (index == 0) return Size;
+			if (index - 1 <= IndexValues.Count)
+				return IndexValues[index].Value;
+
 			throw new InvalidOperationException ();
```
vitek-karas, line 1285: Should we hard-fail in this case? I would expect us to return unknown value here.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -894,5 +894,61 @@ private static VariableDefinition GetLocalDef (Instruction operation, Collection
...
+		{
+			StackSlot indexToLoadFrom = PopUnknown (currentStack, 1, methodBody, operation.Offset);
+			StackSlot arrayToLoadFrom = PopUnknown (currentStack, 1, methodBody, operation.Offset);
+			if (arrayToLoadFrom.Value is not ArrayValue arr) {
+				PushUnknown (currentStack);
+				return;
+			}
+			int? index = indexToLoadFrom.Value.AsConstInt ();
+			if (index == null) {
+				PushUnknown (currentStack);
```
MichalStrehovsky, line 938: Should we also erase the whole array if this was ldelema? (This feels equivalent to a store with unknown index.)

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -894,5 +894,61 @@ private static VariableDefinition GetLocalDef (Instruction operation, Collection
...
+				if (array is ArrayValue arrValue) {
+					if (indexToStoreAtInt == null) {
+						// Since we can't know the current index we're storing the value at, clear all indices.
+						// That way we won't accidentally think we know the value at a given index when we cannot.
+						foreach (var knownIndex in arrValue.IndexValues.Keys) {
+							StoreMethodLocalValue (arrValue.IndexValues, UnknownValue.Instance, knownIndex, curBasicBlock);
+						}
+					} else {
+						// When we know the index, we can record the value at that index.
+						StoreMethodLocalValue (arrValue.IndexValues, valueToStore.Value, indexToStoreAtInt.Value, curBasicBlock);
```
MichalStrehovsky, line 918: Wondering if we should impose a limit on the number of indices we can model (say, 32?) - there are lots of big initialized arrays in .NET, most of which we don't care about.

MichalStrehovsky (commented):
The annoying thing about modeling arrays is that they're reference types. As soon they "escape" the current analysis, we should technically assume their value is unknown. I'm not sure whether we care in practice, but it's worth at least capturing somewhere.

```csharp
object[] foo = new object[123];
Populate(foo);
// The values in foo might have changed
```

test/Mono.Linker.Tests.Cases/Reflection/ExpressionCallString.cs
```diff
@@ -253,11 +254,10 @@ static void TestMethodWithoutRequirements ()
 			}
 
 			[Kept]
-			[ExpectedWarning ("IL2060", "Expression::Call")]
 			static void TestMethodWithRequirements ()
 			{
-				// This must warn - as this is dangerous
-				Expression.Call (typeof (TestGenericMethods), nameof (GenericMethodWithRequirements), new Type[] { GetUnknownType () });
+				// This may not warn - as it's safe
+				Expression.Call (typeof (TestGenericMethods), nameof (GenericMethodWithRequirements), new Type[] { GetUnknownTypeWithRequrements () });
```
vitek-karas, line 260: Can you keep both cases? Otherwise I don't see a case where we get an unknown type in the array.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -820,7 +820,13 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
 									if (stringParam is KnownStringValue stringValue) {
 										foreach (var method in systemTypeValue.TypeRepresented.GetMethodsOnTypeHierarchy (m => m.Name == stringValue.Contents, bindingFlags)) {
-											ValidateGenericMethodInstantiation (ref reflectionContext, method, methodParams[2], calledMethod);
+											if (!ValidateGenericMethodInstantiation (ref reflectionContext, method, methodParams[2], calledMethod)) {
+												reflectionContext.RecordUnrecognizedPattern (
+												2060, string.Format (Resources.Strings.IL2060,
+													DiagnosticUtilities.GetMethodSignatureDisplayName (calledMethod)));
+											} else {
+												reflectionContext.RecordHandledPattern ();
+											}
```
vitek-karas, line 829: Why not do this inside the ValidateGenericMethodInstantiation - the two callsites look exactly the same.

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -1239,23 +1252,36 @@ public override bool Equals (ValueNode other)
...
+				yield return new ArrayValue (sizeConst, ElementType, IndexValues);
 		}
 
 		protected override ValueNode ChildAt (int index)
 		{
 			if (index == 0) return Size;
+			if (index - 1 <= IndexValues.Count)
+				return IndexValues.Values.ElementAt (index - 1).Value;
+
 			throw new InvalidOperationException ();
```
vitek-karas, line 1285: Sorry if I missed it - what was the resolution of my question? I don't think we should fail hard if we can avoid it.

vitek-karas (commented):
I think we should add tests for the handling of cases like:
* Multiple values for the array
* Multiple values for one element in an array
* Accessing uninitialized array element
* Verifying that calling methods with the array as parameter "invalidates" it

src/linker/Linker.Dataflow/ValueNode.cs
jkoritzinsky: I realized that there was a bug in the ChildAt function for fetching from the IndexValues dictionary. The idea I had for what the "children" collection is for the ArrayValue object is to have each `index-1` element be the i-th element of the Values collection and to match other nodes with children, to throw an `InvalidOperationException` when an out of range child is requested. To fetch a value at a particular index, the user should use the `IndexValues` property instead.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
jkoritzinsky: Initially I was going to reuse `ValidateGenericMethodInstantiation` with the `MakeGenericType` case, but I've since realized that doesn't quite work right. I'll refactor this into there.

test/Mono.Linker.Tests.Cases/Reflection/ExpressionCallString.cs
jkoritzinsky: I'll add in a new case that generates the correct "cannot be analyzed" semantics.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -2229,16 +2217,23 @@ void MarkEventsOnTypeHierarchy (ref ReflectionPatternContext reflectionContext,
...
-				return true;
+			if (!genericMethod.HasGenericParameters) {
+				reflectionContext.RecordHandledPattern ();
+			}
 
-			return AnalyzeGenericInstatiationTypeArray (genericParametersArray, ref reflectionContext, reflectionMethod, genericMethod.GenericParameters);
+			if (!AnalyzeGenericInstatiationTypeArray (genericParametersArray, ref reflectionContext, reflectionMethod, genericMethod.GenericParameters)) {
+				reflectionContext.RecordUnrecognizedPattern (
+2060, string.Format (Resources.Strings.IL2060,
+DiagnosticUtilities.GetMethodSignatureDisplayName (reflectionMethod)));
```
vitek-karas, line 2233: The formatting looks weird.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -702,20 +702,29 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
-										//  without any good reason to do so.
+								if (AnalyzeGenericInstatiationTypeArray (methodParams[1], ref reflectionContext, calledMethodDefinition, typeValue.TypeRepresented.GenericParameters)) {
+									reflectionContext.RecordHandledPattern ();
+								} else {
+									bool hasUncheckedAnnotation = false;
+									foreach (var genericParameter in typeValue.TypeRepresented.GenericParameters) {
+										if (_context.Annotations.FlowAnnotations.GetGenericParameterAnnotation (genericParameter) != DynamicallyAccessedMemberTypes.None ||
+											(genericParameter.HasDefaultConstructorConstraint && !typeValue.TypeRepresented.IsTypeOf ("System", "Nullable`1"))) {
+											// There is a generic parameter which has some requirements on the input types.
+											// For now we don't support tracking actual array elements, so we can't validate that the requirements are fulfilled.
```
vitek-karas, line 713: The comment is probably out of date now.

vitek-karas (approved):

jkoritzinsky:
I don't have permissions to merge. @vitek-karas can you merge this PR for me when you have a chance?

PR #1956, opened 2021-04-12
Continue updating forwarder scopes even if a non resolvable forwarder is found

mateoatr:
Should fix https://github.com/dotnet/runtime/pull/50846

We mark all type forwarders for copy assemblies during MarkStep; this may end up marking forwarders that are not resolvable (i.e., `td = forwarder.Resolve()` will be null) -- which is fine. During SweepStep we update scopes of forwarders so that we point to the correct assembly ref, although, we might end up returning early if there was a forwarder which could not be resolved:

https://github.com/mono/linker/blob/eded7e5b8aaa09d0d4c9564b5419ab35a057507c/src/linker/Linker.Steps/SweepStep.cs#L681-L687

If we return early, some forwarders that are actually used, might have assembly refs pointing to nowhere.

### DIFF
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -68,7 +68,7 @@ namespace Mono.Linker.Steps
 				switch (Annotations.GetAction (assembly)) {
 				case AssemblyAction.Link:
 				case AssemblyAction.Save:
-					SweepAssemblyReferences (assembly, Annotations);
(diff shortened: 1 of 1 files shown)


sbomer (approved):
Thanks for finding and fixing! If it's easy enough, might be nice to add a regression test (but I wouldn't bother if it takes more than a few minutes).

marek-safar:
The problem is still there. Same failure on https://github.com/dotnet/runtime/pull/51162

PR #1961, opened 2021-04-12
Suppress dependencies for analyzer package

agocke:
Fixes problem in https://github.com/dotnet/sdk/pull/16820

The build is automatically inserting a package dependency on "ILLink.RoslynAnalyzer" (the name of the dependent project) but both assemblies are included in the same package. Adding the `SuppressDependenciesWhenPacking` seems to do the trick.


sbomer (approved):
Ah, makes sense. PrivateAssets on the packagereference might be another way to do it IIRC.

agocke:
It might, but I do actually want the DLL to be output, just not included as a package reference.

agocke:
Checked -- yearh PrivateAssets also works. I think this flag sets the defaults for all project references, and `privateassets` would let you customize per-reference. I don't really have a preference either way.

PR #1962, opened 2021-04-13
Fixes a nullref and spurious warnings

vitek-karas:
Both of these issues were found by the same test, so that's why fixing them together.

* Null ref in the recently added single-warning logic. If the origin of the warning is a top-level TypeDefinition, the code nullrefs.
* If there's an attribute type which should be removed (like NullableAttribute) in a copy assembly, we keep the type. Since all of its members are also marked the code will call MarkType many times on it - the current guard which prevents spurious 2045 warnings doesn't work in this case since the source location member is null (in theory it should be the "copy assembly", but currently we can't represent that). To prevent the warnings it's better to rely on reason.Source as that will be the member always.

Tweaked existing tests to cover this case and also added one more test case.

### DIFF
--- a/src/linker/Linker/MessageContainer.cs
+++ b/src/linker/Linker/MessageContainer.cs
@@ -134,8 +134,9 @@ namespace Mono.Linker
 
 			if (subcategory == MessageSubCategory.TrimAnalysis) {
 				Debug.Assert (origin.MemberDefinition != null);
-				var assembly = origin.MemberDefinition?.DeclaringType.Module.Assembly;
-				var assemblyName = assembly?.Name.Name;
+				var declaringType = origin.MemberDefinition?.DeclaringType ?? (origin.MemberDefinition as TypeDefinition);
+				var assembly = declaringType?.Module?.Assembly;
+				var assemblyName = assembly?.Name?.Name;
 				if (assemblyName != null && context.IsSingleWarn (assemblyName)) {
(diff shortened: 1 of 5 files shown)


vitek-karas:
Note: The nullref is a recent regression which crashes the linker - so we absolutely need to fix that soon.

marek-safar (approved):

src/linker/Linker/MessageContainer.cs
```diff
@@ -134,8 +134,9 @@ private static MessageContainer CreateWarningMessageContainer (LinkContext conte
 
 			if (subcategory == MessageSubCategory.TrimAnalysis) {
 				Debug.Assert (origin.MemberDefinition != null);
-				var assembly = origin.MemberDefinition?.DeclaringType.Module.Assembly;
-				var assemblyName = assembly?.Name.Name;
+				var declaringType = origin.MemberDefinition?.DeclaringType ?? (origin.MemberDefinition as TypeDefinition);
+				var assembly = declaringType?.Module?.Assembly;
```
sbomer, line 138: Do these need `?` for every dereference? I thought the first `?` would cause the whole thing to return null if declaringType was null.

sbomer (approved):
Thanks for catching this!

src/linker/Linker/MessageContainer.cs
vitek-karas: You're right... I was just... "let's git rid of null ref" :wink:

Issue #1912, opened 2021-03-23
RUC analyzer does not warn about delegate conversion


agocke:
E.g.

```C#
[RequiresUnreferencedCode("")]
void M1() { }
void M2()
{
    Action a = M1; // should warn
    Action b = () => M1(); // should warn
}
```

MichalStrehovsky:
Dupe of #1764? "We should probably extend this to delegates pointing to such methods (although that one has a natural extension - we could require the delegate invoke method itself to be annotated same as the target)."

agocke:
Yeah, they're maybe dups -- this bug was specific to the analyzer, but if the linker is missing similar analysis it makes sense to do both together. 

My suggestion is simpler as well-- warn at the point of delegate conversion, instead of Invoke, so we don't have to do alias tracking for delegate instances. From what I've seen it's often not a worse experience anyway because users often want to see the origination of the problem instead of the final invocation.

MichalStrehovsky:
> My suggestion is simpler as well-- warn at the point of delegate conversion, instead of Invoke

Yup, this is also what I want linker to do. The comment about invoke method is just that we could potentially allow delegate conversion if the delegate itself has matching annotations. So for example:

```csharp
delegate ConstructorInfo GimmeConstructor([DynamicallyAccessedMembers(Constructors)] Type theType);
```

Could be used to construct a delegate pointing to an annotated method. The annotations at the delegate use site naturally fall out.

Unfortunately, it doesn't look like C# allows placing attributes on the Invoke method itself, so the natural extension for RequiresUnreferencedCode doesn't look possible. But it could work at least for the flow annotations if we have the need.

tlakollo:
While reviewing the tests and adding some more for RAF I realized that both Linker and Analyzer already have support for this so I will close this PR and add lazy tests in #2113

PR #1964, opened 2021-04-15
Fix delegate conversion on analyzers

tlakollo:
Fix delegate conversion on analyzers
Add tests
Related to https://github.com/mono/linker/issues/1912


src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -118,6 +118,12 @@ public override void Initialize (AnalysisContext context)
 					CheckCalledMember (operationContext, eventRef.Member, dangerousPatterns);
 				}, OperationKind.EventReference);
 
+				context.RegisterOperationAction (operationContext => {
+					var delegateCreation = (IDelegateCreationOperation) operationContext.Operation;
+					var target = (IMethodReferenceOperation) delegateCreation.Target;
```
Youssef1313, line 123: This cast doesn't feel safe. the target may be `IAnonymousFunctionOperation` for something like `Action x = () => { };` or can even be a none operation for invalid code.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -72,6 +72,12 @@ public override void Initialize (AnalysisContext context)
 						CheckMethodOrCtorCall (operationContext, prop.SetMethod);
 				}, OperationKind.PropertyReference);
 
+				context.RegisterOperationAction (operationContext => {
+					var delegateCreation = (IDelegateCreationOperation) operationContext.Operation;
+					var target = (IMethodReferenceOperation) delegateCreation.Target;
```
Youssef1313, line 77: Same applies here.

agocke: This diff makes me think we need to share more between the two analyzers. They look like they're doing pretty much exactly the stuff.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -72,6 +72,12 @@ public override void Initialize (AnalysisContext context)
 						CheckMethodOrCtorCall (operationContext, prop.SetMethod);
 				}, OperationKind.PropertyReference);
 
+				context.RegisterOperationAction (operationContext => {
```
agocke, line 75: This diff makes me think we need to share more between the two analyzers. They look like they're doing pretty much exactly the stuff.

test/ILLink.RoslynAnalyzer.Tests/RequiresAssemblyFilesAnalyzerTests.cs
```diff
@@ -352,5 +352,52 @@ public void M()
 
 			return VerifyRequiresAssemblyFilesAnalyzer (src);
 		}
+
+		[Fact]
```
agocke, line 356: Same with the tests. Can we start to share stuff here?

tlakollo: I open https://github.com/mono/linker/issues/1986 to track the refactoring work in a separate PR since I think it would have a lot of feedback.

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
```diff
@@ -234,7 +240,12 @@ Action M2()
 					// /0/Test0.cs(12,28): warning IL2026: Using method 'C.M1()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. message.
 					VerifyCS.Diagnostic().WithSpan(12, 28, 12, 32).WithArguments("C.M1()", "message", "")
 				},
-				fixedExpected: Array.Empty<DiagnosticResult> ());
+				fixedExpected: Array.Empty<DiagnosticResult> (),
+				// This test needs to declare the number of iterations it will run, by default the number of iterations is calculated counting the number of DiagnosticResults
```
agocke, line 244: Note sure I understand the comment, lemme see if I can rephrase:

The default iterations for the codefix is the number of diagnostics (1 in this case),
but since the fixer introduces a new diagnostic in the first iteration, it needs to run
twice, so we set the number of iterations to 2.

agocke (approved):
LGTM aside from small question about comment

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
tlakollo: I like your comment, I will change the comment

PR #1966, opened 2021-04-15
Delete CodeBase and EscapedCodeBase from analyzer

tlakollo:
Delete CodeBase and EscapedCodeBase from the analyzer since they are obsolete patterns. See [file](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime/ref/System.Runtime.cs)

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
@@ -76,7 +76,6 @@ namespace ILLink.RoslynAnalyzer
 				var assemblyType = compilation.GetTypeByMetadataName ("System.Reflection.Assembly");
 				if (assemblyType != null) {
 					// properties
-					var property = ImmutableArrayOperations.TryGetSingleSymbol<IPropertySymbol> (assemblyType.GetMembers ("Location"));
 					ImmutableArrayOperations.AddIfNotNull (dangerousPatternsBuilder, ImmutableArrayOperations.TryGetSingleSymbol<IPropertySymbol> (assemblyType.GetMembers ("Location")));
 
 					// methods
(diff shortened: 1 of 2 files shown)


mateoatr:
These are obsolete since .NET 5. And even though marked obsolete, users can still use them, no? I guess I don't see why we should remove this tests -- maybe we should once the APIs are gone.

tlakollo:
> These are obsolete since .NET 5. And even though marked obsolete, users can still use them, no? I guess I don't see why we should remove this tests -- maybe we should once the APIs are gone.

Since they have an obsolete attribute in the ref assembly, we will warn that these shouldn't be used, and the property will throw an exception (https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.codebase?view=net-5.0). I think having the obsolete it's enough.

mateoatr:
If the user silences `SYSLIB0012` for any reason, and is using these APIs in a single file app, we should still warn (and have a test for this).

agocke:
The message given is wrong -- these don't return empty string, they throw an exception. You effectively can't use these properties in single-file.

agocke:
If we really feel the need to duplicate the messages it seems like these APIs aren't important enough to be worth special casing. We could probably just put `RequiresAssemblyFiles` on them.

mateoatr:
I don't  think we need to special case this. But we shouldn't remove the tests, and we shouldn't remove the diagnostics until we have annotated the APIs in the runtime.

agocke:
I'm fine with keeping the tests, but we don't have, and don't plan to have, tests to validate that things marked with `RequiresAssemblyFiles` produce analyzer warnings in the linker repo. Mostly for the simple reason that that's not a unit test, that's an integration test.

So why should we have an integration test for these APIs in particular? When we special case the analyzer for particular APIs, sure, but isn't our job just to make sure the analyzer works properly with the attribute and let the runtime decide what to attribute?

mateoatr:
> So why should we have an integration test for these APIs in particular? When we special case the analyzer for particular APIs, sure, but isn't our job just to make sure the analyzer works properly with the attribute and let the runtime decide what to attribute?

Agreed. We should annotate and then this can go away.

agocke:
Aside: if someone suppresses `SYSLIB0012` they're only suppressing warnings on these exact 2 APIs, so I'm not worried about any suppressions coming into play here.

tlakollo:
I will then suppress the warning for IL3000 in https://github.com/dotnet/runtime/blob/main/src/libraries/System.ComponentModel.Composition/src/System/ComponentModel/Composition/Hosting/AssemblyCatalog.cs#L559
I don't think is worth having a special casing in the analyzer only for CodeBase.get than later on needs to be deleted
Will mark the CodeBase getter with RequiresAssemblyFiles

PR #1968, opened 2021-04-15
Check for namespace in GetDisplayName

### LINKED ISSUE #1967: Warnings include '.' for names of types in global namespace
For types in the global namespace, the warnings are printing the type name with an extra `.`:

```csharp
class Program
{
    static void Main() {}
    static void Activate(Type t) {
        // IL2067: .Program.Activate(Type): '#0' argument does not satisfy...
        Activator.CreateInstance(t);
    }
}
```

mateoatr:
Fix #1967. If a type lives in the global namespace, `GetDisplayName` will prepend a period `<emptystring>.MyType.MyMember`.

### DIFF
--- a/src/linker/Linker/TypeReferenceExtensions.cs
+++ b/src/linker/Linker/TypeReferenceExtensions.cs
@@ -11,8 +11,11 @@ namespace Mono.Linker
 		public static string GetDisplayName (this TypeReference type)
 		{
 			var builder = GetDisplayNameWithoutNamespace (type);
-			builder.Insert (0, ".");
-			builder.Insert (0, type.GetNamespaceDisplayName ());
+			var namespaceDisplayName = type.GetNamespaceDisplayName ();
+			if (!string.IsNullOrEmpty (namespaceDisplayName)) {
+				builder.Insert (0, ".");
+				builder.Insert (0, namespaceDisplayName);
+			}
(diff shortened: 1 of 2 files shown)


src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -11,8 +11,11 @@ public static class TypeReferenceExtensions
 		public static string GetDisplayName (this TypeReference type)
 		{
 			var builder = GetDisplayNameWithoutNamespace (type);
-			builder.Insert (0, ".");
-			builder.Insert (0, type.GetNamespaceDisplayName ());
+			var namespaceDisplayName = type.GetNamespaceDisplayName ();
```
sbomer, line 14: nit: I'd prefer to use Append over Delete

sbomer (approved):
Thanks! Mind adding a test for this?

src/linker/Linker/TypeReferenceExtensions.cs
mateoatr: Yeah -- I'd also prefer using append in the whole `GetDisplayName` logic, I remember trying to use append instead but (IIRC) it got difficult in presence generics. I opened https://github.com/mono/linker/issues/1975.

PR #1974, opened 2021-04-16
Change category from SingleFile to Publish

tlakollo:
Fixes https://github.com/dotnet/docs/issues/23790


tlakollo:
After discussing this with Andy we decided to instead change the documentation to reflect SingleFile

PR #1979, opened 2021-04-20
Unified handling of Resolve calls

### LINKED ISSUE #1953: Sharing state between custom steps
It should be possible to use multiple custom steps with shared state. The way xamarin-android does this is:
1. [Register](https://github.com/xamarin/xamarin-android/blob/681887ebdbd192ce7ce1cd02221d4939599ba762/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets#L61) a single custom step ([SetupStep](https://github.com/xamarin/xamarin-android/blob/main/src/Microsoft.Android.Sdk.ILLink/SetupStep.cs)) via --custom-step
2. SetupStep creates multiple new steps, some with shared [state](https://github.com/xamarin/xamarin-android/blob/d21a42c9e328236cc4fd2f28eecc563e9c48dead/src/Microsoft.Android.Sdk.ILLink/SetupStep.cs#L40)
3. SetupStep [uses](https://github.com/xamarin/xamarin-android/blob/d21a42c9e328236cc4fd2f28eecc563e9c48dead/src/Microsoft.Android.Sdk.ILLink/SetupStep.cs#L20) reflection to add these steps to the linker pipeline

We should provide a way to share state that doesn't require such workarounds. Some options are:
- Make Pipeline available in the ref assembly so adding steps doesn't require reflection
- Design a new api with the same purpose, and make that available in the ref assembly
- Require shared state to be static on one of the custom steps, and fix https://github.com/mono/linker/issues/1314

If we disallow shared state, then xamarin-android could work around it by grouping steps with shared state together, but this isn't the best for code organization, and means that these steps will run in order, without other steps in-between (limiting what custom steps can do without more workarounds).

marek-safar:
This change introduces a commonplace for resolving definition for types,
methods and fields. This is useful for following reasons

- Speeds up linker by 20% for default Blazor app (more for large apps)
- Any custom step can avoid building local mapping cache
- Custom steps could support `--skip-unresolved` linker option
- Consistent error handling for unresolved references

Most of the changes are just mechanical method replacement and
Link context passing.

Contributes to #918
Fixes #1953

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -178,7 +178,13 @@ namespace Mono.Linker.Dataflow
 		protected override ValueNode GetMethodParameterValue (MethodDefinition method, int parameterIndex)
 		{
 			DynamicallyAccessedMemberTypes memberTypes = _context.Annotations.FlowAnnotations.GetParameterAnnotation (method, parameterIndex);
-			return new MethodParameterValue (this, method, parameterIndex, memberTypes, DiagnosticUtilities.GetMethodParameterFromIndex (method, parameterIndex));
+
+			var staticType =
+				!method.HasImplicitThis () ? ResolveToTypeDefinition (method.Parameters[parameterIndex].ParameterType) :
+				parameterIndex == 0 ? method.DeclaringType :
+				ResolveToTypeDefinition (method.Parameters[parameterIndex - 1].ParameterType);
+
+			return new MethodParameterValue (staticType, parameterIndex, memberTypes, DiagnosticUtilities.GetMethodParameterFromIndex (method, parameterIndex));
 		}
 
 		protected override ValueNode GetFieldValue (MethodDefinition method, FieldDefinition field)
@@ -193,7 +199,7 @@ namespace Mono.Linker.Dataflow
 
 			default: {
 					DynamicallyAccessedMemberTypes memberTypes = _context.Annotations.FlowAnnotations.GetFieldAnnotation (field);
-					return new LoadFieldValue (this, field, memberTypes);
+					return new LoadFieldValue (ResolveToTypeDefinition (field.FieldType), field, memberTypes);
 				}
 			}
 		}
@@ -948,7 +954,7 @@ namespace Mono.Linker.Dataflow
 						foreach (var valueNode in methodParams[0].UniqueValues ()) {
 							TypeDefinition staticType = valueNode.StaticType;
 							if (staticType is null) {
-								// We don�t know anything about the type GetType was called on. Track this as a usual �result of a method call without any annotations�
+								// We don't know anything about the type GetType was called on. Track this as a usual result of a method call without any annotations
 								methodReturnValue = MergePointValue.MergeValues (methodReturnValue, new MethodReturnValue (this, calledMethod.MethodReturnType, DynamicallyAccessedMemberTypes.None));
 							} else if (staticType.IsSealed || staticType.IsTypeOf ("System", "Delegate")) {
 								// We can treat this one the same as if it was a typeof() expression
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -17,41 +17,34 @@ namespace Mono.Linker.Dataflow
 {
 	class ReflectionMethodBodyScanner : MethodBodyScanner
 	{
-		readonly LinkContext _context;
 		readonly MarkStep _markStep;
 
 		public static bool RequiresReflectionMethodBodyScannerForCallSite (LinkContext context, MethodReference calledMethod)
 		{
-			MethodDefinition methodDefinition = calledMethod.Resolve ();
-			if (methodDefinition != null) {
-				return
-					GetIntrinsicIdForMethod (methodDefinition) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel ||
-					context.Annotations.FlowAnnotations.RequiresDataFlowAnalysis (methodDefinition) ||
-					context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (methodDefinition);
-			}
+			MethodDefinition methodDefinition = context.TryResolveMethodDefinition (calledMethod);
+			if (methodDefinition == null)
+				return false;
 
-			return false;
+			return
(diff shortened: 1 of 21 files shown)


src/linker/Linker/LinkContext.cs
```diff
@@ -663,6 +663,169 @@ public TargetRuntimeVersion GetTargetRuntimeVersion ()
...
+		{
+			if (methodReference is MethodDefinition md)
+				return md;
+
+			if (methodReference is null)
+				return null;
+
+			if (methodresolveCache.TryGetValue (methodReference, out md)) {
+				if (md == null && !IgnoreUnresolved)
+					ReportUnresolved (methodReference);
```
vitek-karas, line 681: I wonder if we should cache that we already reported the unresolved reference - to avoid repeating the same warning multiple times. We could store a sentinel value in the cache as a mark of that happening.

src/linker/Linker/LinkContext.cs
```diff
@@ -663,6 +663,169 @@ public TargetRuntimeVersion GetTargetRuntimeVersion ()
...
+
+			if (typeReference is TypeSpecification ts) {
+				if (typeReference is FunctionPointerType) {
+					td = null;
+				} else {
+					//
+					// It returns element-type for arrays and also element type for wrapping types like ByReference, PinnedType, etc
+					//
+					td = TryResolveTypeDefinition (ts.GetElementType ());
+				}
```
vitek-karas, line 800: Should we create an issue to handle at least some of these at the callsite - resolving to element-type is dangerous. Not counting that it also hides generic instantiations.

src/linker/Linker/LinkContext.cs
```diff
@@ -663,6 +663,169 @@ public TargetRuntimeVersion GetTargetRuntimeVersion ()
...
+				td = typeReference.Resolve ();
+			}
+
+			typeresolveCache.Add (typeReference, td);
+			return td;
+		}
+
+		public TypeDefinition TryResolveTypeDefinition (AssemblyDefinition assembly, string typeNameString)
+		{
+			// It could be cached if shows up on fast path
```
vitek-karas, line 811:
```suggestion
			// It could be cached if it shows up on fast path
```

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1790,17 +1777,17 @@ TypeDefinition GetDebuggerAttributeTargetType (CustomAttribute ca, AssemblyDefin
 		{
 			foreach (var property in ca.Properties) {
 				if (property.Name == "Target")
-					return ((TypeReference) property.Argument.Value)?.Resolve ();
+					return _context.ResolveTypeDefinition ((TypeReference) property.Argument.Value);
```
vitek-karas, line 1780: Should this be "Try" instead?

vitek-karas (approved):

src/linker/Linker/LinkContext.cs
marek-safar: I actually experimented with it a bit and I think this logic is fine. Most code still needs to resolve to the element so the separation would not have benefits beyond perhaps type system purity.

src/linker/Linker.Steps/MarkStep.cs
marek-safar: Makes sense (resolved)

src/linker/Linker/LinkContext.cs
marek-safar: Added extra cache for that

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -955,8 +948,8 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 						foreach (var valueNode in methodParams[0].UniqueValues ()) {
 							TypeDefinition staticType = valueNode.StaticType;
 							if (staticType is null) {
-								// We don�t know anything about the type GetType was called on. Track this as a usual �result of a method call without any annotations�
-								methodReturnValue = MergePointValue.MergeValues (methodReturnValue, new MethodReturnValue (calledMethod.MethodReturnType, DynamicallyAccessedMemberTypes.None));
+								// We don�t know anything about the type GetType was called on. Track this as a usual �result of a method call without any annotations�
```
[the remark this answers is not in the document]
marek-safar: Fixed

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -833,14 +833,14 @@ public override bool Equals (ValueNode other)
...
+		public MethodParameterValue (MethodBodyScanner scanner, MethodDefinition method, int parameterIndex, DynamicallyAccessedMemberTypes dynamicallyAccessedMemberTypes, IMetadataTokenProvider sourceContext)
 		{
 			Kind = ValueNodeKind.MethodParameter;
 			StaticType = method.HasImplicitThis ()
 				? (parameterIndex == 0
 					? method.DeclaringType
-					: method.Parameters[parameterIndex - 1].ParameterType.ResolveToMainTypeDefinition ())
-				: method.Parameters[parameterIndex].ParameterType.ResolveToMainTypeDefinition ();
+					: scanner.ResolveToTypeDefinition (method.Parameters[parameterIndex - 1].ParameterType))
+				: scanner.ResolveToTypeDefinition (method.Parameters[parameterIndex].ParameterType);
```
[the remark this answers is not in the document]
marek-safar: OK, moved the logic to the caller

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -905,10 +901,10 @@ protected override string NodeToString ()
 	/// </summary>
 	class MethodReturnValue : LeafValueWithDynamicallyAccessedMemberNode
 	{
-		public MethodReturnValue (MethodReturnType methodReturnType, DynamicallyAccessedMemberTypes dynamicallyAccessedMemberTypes)
+		public MethodReturnValue (MethodBodyScanner scanner, MethodReturnType methodReturnType, DynamicallyAccessedMemberTypes dynamicallyAccessedMemberTypes)
```
vitek-karas, line 904: Last one to move into the caller

marek-safar: Yeah, working on it as it can be simplified differently and I didn't want to mix it with the previous ones

marek-safar: Done

vitek-karas (approved):

Issue #1973, opened 2021-04-16
COM warnings from generated descriptors 


sbomer:
On linux the linker produces warnings like the following: resource ILLink.Descriptors.xml in System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e : warning IL2008: Could not resolve type 'System.StubHelpers.MngdSafeArrayMarshaler' [/home/sven/tmp/trimmed/trimmed.csproj]
 
I think they are coming from the runtime-generated descriptors which reference COM, even though COM is only enabled on windows. We may want to take DefineConstants into account when parsing header files in https://github.com/mono/linker/blob/main/src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs.

/cc @LakshanF @vitek-karas

vitek-karas:
For Preview 4 we suppressed these via https://github.com/dotnet/sdk/pull/16979.
After Preview 4 we need to fix this and remove the suppressions.

vitek-karas:
This has been fixed both in runtime and SDK - there should be no implicit suppressions for .NET 6 apps anymore.

PR #1980, opened 2021-04-20
Implement support for ifdefs in header file parsing

vitek-karas:
The CreateRuntimeRootDescriptorFile task parses some header files from the CoreCLR and produces an XML descriptor. This change adds support for understanding simple ifdefs.

Also added tests.

This is the mono/linker part of fix for #1973

### DIFF
--- a/src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
+++ b/src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
@@ -39,6 +39,8 @@ namespace ILLink.Tasks
 		[Required]
 		public ITaskItem ILLinkTrimXmlFilePath { get; set; }
 
+		public ITaskItem[] DefineConstants { get; set; }
+
 		/// <summary>
 		///   The path to the file to generate.
 		/// </summary>
@@ -55,6 +57,7 @@ namespace ILLink.Tasks
 		readonly Dictionary<string, string> namespaceDictionary = new Dictionary<string, string> ();
 		readonly Dictionary<string, string> classIdsToClassNames = new Dictionary<string, string> ();
 		readonly Dictionary<string, ClassMembers> classNamesToClassMembers = new Dictionary<string, ClassMembers> ();
+		readonly HashSet<string> defineConstants = new HashSet<string> ();
 
 		public override bool Execute ()
 		{
@@ -81,6 +84,8 @@ namespace ILLink.Tasks
 				Log.LogError ("File " + rexcepFilePath + " doesn't exist.");
 				return false;
 			}
+
+			ParseDefineConstants ();
 
 			var iLLinkTrimXmlFilePath = ILLinkTrimXmlFilePath.ItemSpec;
 			if (!File.Exists (iLLinkTrimXmlFilePath)) {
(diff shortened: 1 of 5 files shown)


vitek-karas:
The dotnet/runtime counterpart to this change is here: https://github.com/dotnet/runtime/pull/51562

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -336,5 +396,15 @@ string GetFullClassName (string classNamespace, string className)
...
 		}
+
+		void ParseDefineConstants ()
+		{
+			if (DefineConstants is not null) {
+				foreach (var item in DefineConstants)
+					defineConstants.Add (item.ItemSpec.Trim ());
+			}
+
+			defineConstants.Add ("FOR_ILLINK");
```
sbomer, line 407: Instead of building this in, might it be better to pass it in from runtime?

marek-safar (approved):
Looks fine to me but not sure if we need it. I think we have task that can merge XMLs files together in runtime already

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -132,9 +137,64 @@ void ProcessNamespaces (string namespaceFile)
 		void ProcessMscorlib (string typeFile)
```
sbomer, line 137: Should any of the other files get the same treatment? rexcep.h has a few conditional ones:
https://github.com/dotnet/runtime/blob/58b25dfa8fae5f5a382a6ffbe89e9aee479eb31c/src/coreclr/vm/rexcep.h#L181-L183

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -132,9 +137,64 @@ void ProcessNamespaces (string namespaceFile)
...
+			int linenum = 0;
 			string classId;
 
+			Stack<bool?> activeSections = new Stack<bool?> ();
 			foreach (string def in types) {
+				linenum++;
+				{
+					int ifdefLength = 0;
+					bool negativeIfDef = false;
+					if (def.StartsWith ("#ifdef ")) {
```
jkotas, line 149: Would it be better to just run the C/C++ preprocessor on the file instead of building a poor man's version of the preprocessor into this task?

vitek-karas: I've investigated what it would take to use C/C++ preprocessor:
* We can't really run it during coreclr build - it would mean that building CoreLib requires output from CoreCLR build - which is a new dependency and the infra team really doesn't like this (it would also mean that `dotnet build` on CoreLib would basically never work fully).
* Running CMake from CoreLib build is a possibility, but it's complex and probably slow (we already have 3 places we start native build pipeline in the runtime repo, this would introduce a new one - while we're trying hard to go down to just 1 place ideally).
* That leaves running the preprocessor directly (without CMake) from the CoreLib build. This is probably not that hard since we have scripts to find the compiler path - [Windows to get cl.exe on PATH](https://github.com/dotnet/runtime/blob/main/eng/native/init-vs-env.cmd), [Linux, macOS to get clang path into CC/CXX](https://github.com/dotnet/runtime/blob/main/eng/native/init-compiler.sh). That said it's still not exactly easy as we would be adding a new thing into the pipeline for CoreLib - the risk here is that we have to construct the compiler command line by hand - making that work on all the platforms could be time consuming.
  * This also comes with a downside - we would not run it in the same environment as within CoreCLR build, for example we would have to propagate the defines manually.

In the meantime, there's a request to add additional logic to this process to be able to specify that some of the roots should be added behind a feature switch (built-in COM support), which puts additional pressure on making this happen.

So the proposal is:
* Keep the current "hacky" solution (merge this PR and then the one in runtime)
* Define new macros for the feature switch support so that we express that in a C++ syntax - but still use the "hacky" parser for it
* File an issue to improve this in the future - documenting the knowledge from the investigation done so far.

What do you think @jkotas ?

/cc @LakshanF

jkotas: Sounds good. Thank you for looking into this.

> File an issue to improve this in the future

The ultimate plan would be to switch to the same plan that Mono is on: Capture the dependencies introduced by the VM side explicitly by regular DynamicDependency attributes (and probably also convert more of the VM logic to C# so that it is maintainable). This whole task can then disapper.

vitek-karas: I filed https://github.com/dotnet/runtime/issues/51991 to track this

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
vitek-karas: Good point.

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
vitek-karas: Good point

vitek-karas:
> Looks fine to me but not sure if we need it. I think we have task that can merge XMLs files together in runtime already

The `corelib.h` contains several ifdefs which mostly surround COM related stuff. We don't want to root these on non-Windows. There are also architecture specific APIs there - same story.

sbomer (approved):

PR #1983, opened 2021-04-21
Fix generic parameter validation in MakeGenericMethod and similar

vitek-karas:
The validator must ignore methods which don't have any requirements. So even if the passed in array doesn't match the method, it needs to not warn in such cases. This is because we can't guarantee we always find the exact method info to work on, instead we compensate by considering potentially multiple candidates. Some of which may not even be generic, or may not have any requirements and so on. In such cases the validator should only warn if there's potentially a real problem - not if the reflection API will fail (that's a non-failure for the validator).

This should fix the problem in https://github.com/dotnet/runtime/pull/51558#issuecomment-823430400.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1723,8 +1723,10 @@ namespace Mono.Linker.Dataflow
 		{
 			bool hasRequirements = false;
 			foreach (var genericParameter in genericParameters) {
-				if (_context.Annotations.FlowAnnotations.GetGenericParameterAnnotation (genericParameter) != DynamicallyAccessedMemberTypes.None)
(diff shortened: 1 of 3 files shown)


vitek-karas:
@jkoritzinsky

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1721,6 +1721,16 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 
 		private bool AnalyzeGenericInstatiationTypeArray (ValueNode arrayParam, ref ReflectionPatternContext reflectionContext, MethodReference calledMethod, IList<GenericParameter> genericParameters)
 		{
+			bool hasRequirements = false;
+			foreach (var genericParameter in genericParameters) {
+				if (_context.Annotations.FlowAnnotations.GetGenericParameterAnnotation (genericParameter) != DynamicallyAccessedMemberTypes.None)
+					hasRequirements = true;
```
marek-safar, line 1727: missing break ?

marek-safar (approved):

PR #1987, opened 2021-04-22
Fix issue with forwarders in copyused assemblies

sbomer:
@vitek-karas noticed an issue where we stopped updating typerefs from copyused assemblies in some cases. This fixes the issue by always updating scopes in copyused assemblies.

This will be subtly different from the behavior before @mateoatr's changes in case the copyused assembly doesn't reference a removed assembly (we used to not rewrite references in that case, but now we will). I think this is the right thing to do since I believe the old behavior could have caused bugs similar to this one.


marek-safar (approved):

mateoatr (approved):
LGTM. Thanks!

PR #1988, opened 2021-04-22
Trim parameter names metadata when they are not needed.

marek-safar:
The total saving for Blazor template app is about 135k uncompressed. The
names are still kept for reflection and debugging scenarios.

I had to update the verification infrastructure to make the testing
easier when the names are not available.

### DIFF
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -415,6 +415,8 @@ namespace ILLink.Tasks
 				}
 			}
 
+			bool debugger_supported = true;
+
 			if (FeatureSettings != null) {
 				foreach (var featureSetting in FeatureSettings) {
 					var feature = featureSetting.ItemSpec;
@@ -422,8 +424,14 @@ namespace ILLink.Tasks
 					if (String.IsNullOrEmpty (featureValue))
 						throw new ArgumentException ("feature settings require \"Value\" metadata");
 					args.Append ("--feature ").Append (feature).Append (' ').AppendLine (featureValue);
+
+					if (feature == "System.Diagnostics.Debugger.IsSupported" && bool.TryParse (featureValue, out bool fValue))
+						debugger_supported = fValue;
 				}
 			}
+
+			if (debugger_supported)
+				args.AppendLine ("--keep-metadata all");
 
 			if (_removeSymbols == false)
 				args.AppendLine ("-b");
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -213,6 +213,11 @@ namespace ILLink.Tasks
 		/// </summary>
 		public ITaskItem[] CustomSteps { get; set; }
 
+		/// <summary>
+		///   A list selected metadata which should not be trimmed. It maps to 'keep-metadata' option
+		/// </summary>
+		public ITaskItem[] KeepMetadata { get; set; }
+
 		private const string DotNetHostPathEnvironmentName = "DOTNET_HOST_PATH";
 
 		private string _dotnetPath;
@@ -415,8 +420,6 @@ namespace ILLink.Tasks
 				}
 			}
 
-			bool debugger_supported = true;
-
 			if (FeatureSettings != null) {
 				foreach (var featureSetting in FeatureSettings) {
 					var feature = featureSetting.ItemSpec;
@@ -424,14 +427,13 @@ namespace ILLink.Tasks
 					if (String.IsNullOrEmpty (featureValue))
 						throw new ArgumentException ("feature settings require \"Value\" metadata");
 					args.Append ("--feature ").Append (feature).Append (' ').AppendLine (featureValue);
-
-					if (feature == "System.Diagnostics.Debugger.IsSupported" && bool.TryParse (featureValue, out bool fValue))
-						debugger_supported = fValue;
 				}
 			}
 
-			if (debugger_supported)
-				args.AppendLine ("--keep-metadata all");
+			if (KeepMetadata != null) {
+				foreach (var metadata in KeepMetadata)
+					args.Append ("--keep-metadata ").AppendLine (Quote (metadata.ItemSpec));
+			}
 
 			if (_removeSymbols == false)
 				args.AppendLine ("-b");
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -41,6 +41,7 @@ namespace Mono.Linker.Steps
 	{
 		AssemblyDefinition[] assemblies;
 		readonly bool sweepSymbols;
+		bool sweep_debug_support;
 		readonly HashSet<AssemblyDefinition> BypassNGenToSave = new HashSet<AssemblyDefinition> ();
 
 		public SweepStep (bool sweepSymbols = true)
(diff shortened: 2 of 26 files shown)


src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -50,6 +51,8 @@ public SweepStep (bool sweepSymbols = true)
 
 		protected override void Process ()
 		{
+			sweep_debug_support = Context.HasFeatureValue (FeatureSettings.DebuggerIsSupported, false);
```
vitek-karas, line 54: I'm wondering if we should add a new feature for this type of optimization. Only for the linker. And then in SDK default it when the Debugger.IsSupported is off. Just so that there's a way to control it specifically without controlling the rest of the debugger support changes.

test/Mono.Linker.Tests.Cases.Expectations/Assertions/RemovedNameValueAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Mono.Linker.Tests.Cases.Expectations.Assertions
+{
+	/// <summary>
+	/// Verifies that literal value of the member is removed
```
vitek-karas, line 6:
```suggestion
	/// Verifies that name of the member is removed
```

"literal value" to me reads like `_field = "literal value"`.

vitek-karas (approved):
I assume Blazor app still works after trimming with this. We will also have to try how this impacts CoreCLR and Xamarin.

MichalStrehovsky:
I'm wondering whether it would be worth it to consider another angle:

* Remove the names unconditionally (no feature switch needed)
* Mark the obvious APIs (`Exception.TargetSite`, `StackFrame.GetMethod`, `MethodBase.GetCurrentMethod`) as `RequiresUnreferencedCode` with a message describing that various _unspecified_ parts of the metadata could be missing to enable future optimizations.
* Don't do this for generic type parameters because marking `Object.GetType` as `RequiresUnreferencedCode` wouldn't be acceptable

The advantage is that we get a more consistent trimming experience that is not based on the SDK. The SDKs having different defaults has been scaring me for a while and I'm worried it could come back to bite us. It would be better not to have too many _observable_ behavioral differences between various SDKs.

marek-safar:
> Remove the names unconditionally (no feature switch needed)

That's what it does. The names are removed unconditionally only when debugging support is enabled it keeps everything for debugger-like tools to allow users to show everything.

> Mark the obvious APIs with a message 

What would be the benefit of that? These APIs are usually wrapped with some high-level api like ToString

> The advantage is that we get a more consistent trimming experience that is not based on the SDK.

The behaviour is the default unless you opt-out with a switch I'm planning to add based on Vitek's feedback

MichalStrehovsky:
> What would be the benefit of that? These APIs are usually wrapped with some high-level api like ToString

`Console.WriteLine(MethodBase.GetCurrentMethod().GetParameters()[0]);` is going to print different results based on whether the app is trimmed or not - if we're going through the effort of keeping the names for methods that are obviously reflected on - but what about methods unobviously reflected on? The easiest solution is to make sure there's a warning for that - that's our license to have behavioral differences.

> That's what it does. The names are removed unconditionally only when debugging support is enabled it keeps everything for debugger-like tools to allow users to show everything.

If that switch will be enabled by default in some SDKs, I still have the same concern. If we don't intend to enable it by default anywhere, then that addresses my concern.

marek-safar:
> The easiest solution is to make sure there's a warning for that - that's our license to have behavioral differences.

I'm having a hard time imagining where to put the warning and what it should say. From your list
	- MethodBase.GetCurrentMethod that method should work fine all the time
	- StackFrame.GetMethod could have a different output for ToString (but still valid) but we cannot add a warning to this method
	- Exception.TargetSite similar to the first one. This one could be fine if you don't poke into parameter Infos on the method or call ToString.

If I understood your suggestion you would annotate `ParameterInfo::Name` property with something like "This property can return a null value" and suppress the warning everywhere because there is no other way to resolve the warning?

> If that switch will be enabled by default in some SDKs, I still have the same concern.

It's not by default but it's by default for all release builds, does that meet your default anywhere?

src/linker/Linker.Steps/SweepStep.cs
marek-safar: Yeah, added an extra option to allow control of what will be kept.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -415,16 +415,24 @@ protected override string GenerateResponseFileCommands ()
...
+
 			if (FeatureSettings != null) {
 				foreach (var featureSetting in FeatureSettings) {
 					var feature = featureSetting.ItemSpec;
 					var featureValue = featureSetting.GetMetadata ("Value");
 					if (String.IsNullOrEmpty (featureValue))
 						throw new ArgumentException ("feature settings require \"Value\" metadata");
 					args.Append ("--feature ").Append (feature).Append (' ').AppendLine (featureValue);
+
+					if (feature == "System.Diagnostics.Debugger.IsSupported" && bool.TryParse (featureValue, out bool fValue))
```
sbomer, line 428: I don't think the task is the right place for logic like this - the task should just translate options from MSBuild language into command-line args. Instead this should either be built-in to the linker, or moved to MSBulid targets.

src/linker/Linker/MetadataTrimming.cs
```diff
@@ -0,0 +1,17 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Mono.Linker
+{
+	[Flags]
+	public enum MetadataTrimming
+	{
+		None = 1,
```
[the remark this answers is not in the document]
marek-safar: Fixed

src/linker/Linker/Driver.cs
```diff
@@ -1088,6 +1101,22 @@ protected bool GetOptimizationName (string text, out CodeOptimizations optimizat
 			return false;
 		}
 
+		bool GetMetadataTrimming (string text, out MetadataTrimming metadataTrimming)
```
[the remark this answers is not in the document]
marek-safar: Renamed

src/linker/Linker/Driver.cs
```diff
@@ -1088,6 +1101,22 @@ protected bool GetOptimizationName (string text, out CodeOptimizations optimizat
...
+		{
+			switch (text.ToLowerInvariant ()) {
+			case "all":
+				metadataTrimming = MetadataTrimming.Any;
+				return true;
+			case "parametername":
+				metadataTrimming = MetadataTrimming.ParameterName;
+				return true;
+			}
```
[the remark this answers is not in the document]
marek-safar: ok, added

src/ILLink.Tasks/LinkTask.cs
marek-safar: Replaces with explicit task setting

src/linker/Linker/MetadataTrimming.cs
```diff
@@ -9,7 +9,7 @@ namespace Mono.Linker
 	[Flags]
 	public enum MetadataTrimming
 	{
-		None = 1,
+		None = 0,
 		ParameterName = 1 << 1,
```
vitek-karas, line 13: This can be just `1` now

vitek-karas (approved):

PR #1992, opened 2021-04-26
Add tests for object.GetType over generic type arguments

vitek-karas:
This is currently unsupported, so linker should warn, but we didn't have any tests for it.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -956,7 +956,7 @@ namespace Mono.Linker.Dataflow
 							// Note that valueNode can be statically typed in IL as some generic argument type.
 							// For example:
 							//   void Method<T>(T instance) { instance.GetType().... }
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -953,6 +953,21 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 				//
 				case IntrinsicId.Object_GetType: {
 						foreach (var valueNode in methodParams[0].UniqueValues ()) {
+							// Note that valueNode can be statically typed in IL as some generic argument type.
+							// For example:
+							//   void Method<T>(T instance) { instance.GetType().... }
+							// Currently this case will end up with StaticType - since there's no typedef for the generic argument type.
```
MichalStrehovsky, line 959:
```suggestion
							// Currently this case will end up with null StaticType - since there's no typedef for the generic argument type.
```

MichalStrehovsky (approved):

PR #1998, opened 2021-04-27
Implements constant return value propagation for methods with parameters

marek-safar:


vitek-karas:
@MichalStrehovsky 

I remember we discussed this a lot when we added the existing limited support for methods with parameters - and there was a relatively good reason not to do better.

What is the use case for this - where do we need it?
I would not be too worried if we scoped this only to explicitly substituted methods, but allowing full constant prop across methods with arbitrary parameters feels wrong. For example, if the method has a `ref` parameter and mutates it - I know we're going to still call it, but still - it feels risky.

I think this needs a good reason to add and probably lot more tests.

MichalStrehovsky:
I can't come up with examples of when ref parameters specifically would be risky, but the `Rewrite any arguments pushed to stack to nop to allow possible body reduction into constant result` sounds concerning. Do we know the pushes are not side-effecting?

marek-safar:
> What is the use case for this - where do we need it?

We cannot evaluate as much as we could even for dotnet/runtime code. For example with this change, this [condition](https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/AsyncTaskMethodBuilderT.cs#L219-L221) can be now removed because it calls into [method](https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/AsyncMethodBuilderCore.cs#L80) which couldn't be evaluated as const returning before.

> I remember we discussed this a lot when we added the existing limited support for methods with parameters - and there was a relatively good reason not to do better.

Where do you see the risk? The logic for computing the constant value abort when argument value is used I'd like to understand what could be risky here.

> allowing full constant prop across methods with arbitrary parameters feels wrong. For example, if the method has a ref parameter and mutates it - I know we're going to still call it, but still - it feels risky.

That's not what this change does. It only expands support of return constant value propagation over calls with parameters.

> Rewrite any arguments pushed to stack to nop to allow possible body reduction into constant result sounds concerning. Do we know the pushes are not side-effecting?

Why is it concerning? This rewrite is done only for the purpose of the return value evaluation. It does not have an effect on the final IL. 

> I think this needs a good reason to add and probably lot more tests.

I originally wanted but as the change is very narrow we already had test coverage for it. Could you please suggest what other tests would you like to see added?

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
```diff
@@ -521,6 +512,107 @@ void RewriteConditionToNop (int index)
...
+
+						delta -= ms.Parameters.Count;
+						break;
+					}
+
+					Debug.Fail (instruction.Operand?.ToString ());
+					unknown = true;
+					return 0;
+				default:
+					throw new NotImplementedException (instruction.OpCode.StackBehaviourPop.ToString ());
```
vitek-karas, line 581: I would be hesitant to crash the linker on this - just set unknown to true and return. Maybe Debug.Fail so that we catch this in tests.

vitek-karas: Main reason - there's no workaround for crashing the linker. If I hit this and the optimization will not work - well not great, but I can still work with my app. If it crashes the linker, I have no way to trim my app.

vitek-karas (commented):
We should add tests for more complex parameters. Things like:
- passing a return value of a method as a param
- conditions in the param

And set it up such that we have to try to get "past" it - that is make this complex thing the last one pushed on the stack and make sure there's something else before it. You added code for going backwards finding all the parameter setup instructions, but the tests all use simple constants and fields only.

vitek-karas:
What is the impact of this change - how much does it help? (just curious)

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
marek-safar: this should happen only for unknown opcodes but addressed anyway.

marek-safar:
> We should add tests for more complex parameters.

Added a few more, let me know if there is some other interesting il patter to cover

> all use simple constants and fields only

There are already some tests, see e.g. https://github.com/marek-safar/linker/blob/constprop/test/Mono.Linker.Tests.Cases/UnreachableBlock/MethodWithParametersSubstitutions.cs#L145

> What is the impact of this change - how much does it help? (just curious)

Low single digit Kb for console app but we are experimenting with more stuff for Xamarin which could benefit from this too

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
```diff
@@ -524,6 +515,111 @@ void RewriteConditionToNop (int index)
...
+				if (stackDepth != 0) {
+					Debug.Fail ("Invalid IL?");
+					return;
+				}
+
+				while (start_index != index)
+					RewriteToNop (++start_index);
+			}
+
+			static int GetStackBehaviourDelta (Instruction instruction, out bool unknown)
```
vitek-karas, line 541: This doesn't work correctly when there are branches involved. For example:

```C#
TestCall(a == 0 ? 1 : 0);
```

In IL this looks something like:
```
ldarg.0
beq.s IL_002
ldc.i4.1
br.s IL_003
ldc.i4.0 // <- Stops here
call TestCall
```

The algorithm here will walk back but will basically stop on the first instruction since the stack depth will reach 0. The other code in the branches above it will be ignored (and not overwritten with nops). Similarly if there's another parameter above the branches, the two branches will "clear" the stack and we will never get to the other parameter.

This probably doesn't "matter" as all this means is that we won't be able to fully optimize the method (the method will probably never be detected as returning constant value). In that sense I'm fine with this behavior, but I think we should document it. Large comment which says that this is only an approximation, some examples when it breaks and  an explanation why it doesn't matter and what happens if it doesn't work.

marek-safar: It still works when branches are involved but as you said it won't be able to optimize that. Added comment and tests case in case we add the support one day.

vitek-karas (approved):

PR #2000, opened 2021-04-28
Revert "Upgrade linker TFM to net6.0 (#1996)"

vitek-karas:
This reverts commit bf37b2909bbfb2942173e7b383642c9b32ae0e79.

I intentionally kept the new version of SDK in the repo - we need at least P4 anyway due to tests added for APIs added in that release. Ideally we would keep the SDK in sync with dotnet/runtime, but that really hard to do if linker needs to react to API changes in the runtime repo.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <RunAnalyzers>true</RunAnalyzers>
-    <TargetFramework>net6.0</TargetFramework>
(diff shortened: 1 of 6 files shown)


marek-safar (approved):
Is it worth backporting the API changes to P4 so we don't have to wait till July for this to land?

vitek-karas:
We would still have to wait at least 1 month. I'll try to work around this in the linker - it's basically just an enum value - we can just use raw numbers in the worst case.

PR #2002, opened 2021-04-28
Add UnconditionalSuppressMessage codefixer

tlakollo:
Add UnconditionalSuppressMessage codefixer
Add CodeFixProviderOperations to share more code between code fixers
Add test for UnconditionalSuppressMessage


src/ILLink.CodeFix/CodeFixProviderOperations.cs
```diff
@@ -0,0 +1,97 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Simplification;
+
+namespace ILLink.CodeFix
+{
+	class CodeFixProviderOperations
+	{
+
+		internal static async Task<Document> AddRequiresAttribute (
```
agocke, line 19: Why move this into its own file?

src/ILLink.CodeFix/UnconditionalSuppressMessageCodeFixProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+
+		public sealed override FixAllProvider GetFixAllProvider ()
+		{
+			// See https://github.com/dotnet/roslyn/blob/master/docs/analyzers/FixAllProvider.md for more information on Fix All Providers
+			return WellKnownFixAllProviders.BatchFixer;
+		}
+
+		public sealed override async Task RegisterCodeFixesAsync (CodeFixContext context)
+		{
+			var root = await context.Document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false);
```
agocke, line 39: Is this code identical to the code above, except for the attribute targets and the name of the attribute?

src/ILLink.CodeFix/UnconditionalSuppressMessageCodeFixProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+
+			var title = diagnostic.Descriptor.Title.ToString (CultureInfo.CurrentUICulture);
+			var ruleIdText = string.IsNullOrWhiteSpace (title) ? diagnostic.Id : string.Format ("{0}:{1}", diagnostic.Id, title);
+			var ruleId = generator.LiteralExpression (ruleIdText);
+			var ruleIdArgument = generator.AttributeArgument (ruleId);
+
+			var justificationExpr = generator.LiteralExpression ("<Pending>");
+			var justificationArgument = generator.AttributeArgument ("Justification", justificationExpr);
+
+			SyntaxNode[] attrArgs = new[] { categoryArgument, ruleIdArgument, justificationArgument };
```
agocke, line 90: This method also looks very similar, except for the syntax of the attribute arguments. You could add a new abstract base class `AddAttributeToParentCodeFixProvider`, and then have both of the code fixers derive from it. Then, have an abstract method that constructs the syntax for the arguments, and override it for each of the code fix providers.

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
```diff
@@ -5,120 +5,37 @@
...
-				context.RegisterCodeFix (
-					CodeAction.Create (
-						title: s_title,
-						createChangedDocument: c => AddRequiresUnreferencedCode (
-							context.Document, root, targetNode, declarationSyntax, symbol!, c),
-						equivalenceKey: s_title),
-					diagnostic);
-
-			}
+			await BaseRegisterCodeFixesAsync (context, AttributeableParentTargets.Method, RequiresUnreferencedCodeAnalyzer.FullyQualifiedRequiresUnreferencedCodeAttribute, s_title);
```
agocke, line 27: Don't call `await` on Task-returning functions as the last statement, just return the result of the call and change the parent method to not be `async`.

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
```diff
@@ -5,120 +5,37 @@
...
 		}
 
-		private static async Task<Document> AddRequiresUnreferencedCode (
-			Document document,
-			SyntaxNode root,
-			SyntaxNode targetNode,
-			CSharpSyntaxNode containingDecl,
-			ITypeSymbol requiresUnreferencedCodeSymbol,
-			CancellationToken cancellationToken)
+		internal override SyntaxNode[] SetAttributeArguments (SemanticModel semanticModel, SyntaxNode targetNode, CSharpSyntaxNode containingDecl, SyntaxGenerator generator, Diagnostic diagnostic)
```
agocke, line 30: `Set` methods set particular state. I think you want `Get` since you're returning the result.

test/ILLink.RoslynAnalyzer.Tests/UnconditionalSuppressMessageCodeFixTests.cs
```diff
@@ -0,0 +1,512 @@
...
+build_property.{MSBuildPropertyOptionNames.EnableTrimAnalyzer} = true")));
+			if (numberOfIterations != null) {
+				test.NumberOfIncrementalIterations = numberOfIterations;
+				test.NumberOfFixAllIterations = numberOfIterations;
+			}
+			test.FixedState.ExpectedDiagnostics.AddRange (fixedExpected);
+			return test.RunAsync ();
+		}
+
+		static Task VerifyUnconditionalSuppressMessageCodeFixWithRAF (
```
agocke, line 94: Is the above method identical except for the name of the MSBuild property name? If so, consider extracting a helper.

test/ILLink.RoslynAnalyzer.Tests/UnconditionalSuppressMessageCodeFixTests.cs
```diff
@@ -0,0 +1,512 @@
...
+				test.NumberOfIncrementalIterations = numberOfIterations;
+				test.NumberOfFixAllIterations = numberOfIterations;
+			}
+			test.FixedState.ExpectedDiagnostics.AddRange (fixedExpected);
+			return test.RunAsync ();
+		}
+
+
+		[Fact]
+		public async Task RequiresUnreferencedCodeDiagnosticFix ()
```
agocke, line 120: This method name doesn't mention what fix is being used. I'd call it something like `SuppressRucFixer`

test/ILLink.RoslynAnalyzer.Tests/UnconditionalSuppressMessageCodeFixTests.cs
```diff
@@ -0,0 +1,512 @@
...
+    [RequiresUnreferencedCode(""message"")]
+    public int M1() => 0;
+    int M2() => M1();
+}
+class D
+{
+    public int M3(C c) => c.M1();
+    public class E
+    {
+        public int M4(C c) => c.M1();
```
agocke, line 135: A lot of tests with different visibility makes sense for RUC since it applies different things, but `UnconditionalSuppress` doesn't care about visibility, so I'm unclear on what this test is supposed to demonstrate.

test/ILLink.RoslynAnalyzer.Tests/UnconditionalSuppressMessageCodeFixTests.cs
```diff
@@ -0,0 +1,512 @@
...
+    }
+}
+public class E
+{
+    public class F
+    {
+        public int M5(C c) => c.M1();
```
agocke, line 213: Same here, why do I care that the original class `D` is internal but `E` is `public`?

test/ILLink.RoslynAnalyzer.Tests/UnconditionalSuppressMessageCodeFixTests.cs
tlakollo: The type each function is dealing with, so I  create a generic helpers

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -209,5 +260,102 @@ internal static void GetDirectoryPaths (out string rootSourceDirectory, out stri
...
+			{typeof(RequiresUnreferencedCodeAnalyzer), RequiresUnreferencedCodeAnalyzer.DiagnosticId},
+			{typeof(RequiresAssemblyFilesAnalyzer), RequiresAssemblyFilesAnalyzer.IL3002}
+		};
+
+		private static string GetDiagnosticId<TAnalyzer> () where TAnalyzer : DiagnosticAnalyzer
+		{
+			if (diagnosticDict.ContainsKey (typeof (TAnalyzer))) {
+				return diagnosticDict[typeof (TAnalyzer)];
+			} else {
+				throw new ArgumentException ($"Couldn't retrieve diagnostic id data for unrecognized Analyzer {typeof (TAnalyzer).Name}");
```
agocke, line 325:
```suggestion
			var t = typeof(TAnalyzer);
			if (t == typeof(RequiresUnreferencedCodeAnalyzer)) {
			    return RequiresUnreferencedCodeAnalyzer.DiagnosticId;
			else if (t == typeof(RequiresAssemblyFilesAnalyzer)) {
			    return RequiresAssemblyFilesAnalyzer.IL3002;
			}
		        throw new ArgumentException ($"Couldn't retrieve diagnostic id data for unrecognized Analyzer {typeof (TAnalyzer).Name}");
```

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -209,5 +260,102 @@ internal static void GetDirectoryPaths (out string rootSourceDirectory, out stri
...
+
+		private static Dictionary<Type, string> attributeDefinitionDict = new Dictionary<Type, string>
+		{
+			{typeof(RequiresUnreferencedCodeAnalyzer), requiresUnreferencedCodeAttributeDefinition},
+			{typeof(RequiresUnreferencedCodeCodeFixProvider), requiresUnreferencedCodeAttributeDefinition},
+			{typeof(RequiresAssemblyFilesAnalyzer), requiresAssemblyFilesAttributeDefinition},
+			{typeof(UnconditionalSuppressMessageCodeFixProvider), unconditionalSuppressMessageAttributeDefinition}
+		};
+
+		private static string GetAttributeDefinition<T> ()
```
agocke, line 337: Same pattern as above

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -209,5 +260,102 @@ internal static void GetDirectoryPaths (out string rootSourceDirectory, out stri
...
+				TestCaseUtils.UseMSBuildProperties (GetAnalyzerMSBuildPropertyOption<TAnalyzer> ()),
+				baselineExpected);
+
+		internal static DiagnosticResult GetDiagnosticResult<TAnalyzer> ()
+			where TAnalyzer : DiagnosticAnalyzer, new()
+		{
+			return CSharpAnalyzerVerifier<TAnalyzer>.Diagnostic (GetDiagnosticId<TAnalyzer> ());
+		}
+
+		private static Dictionary<Type, string> diagnosticDict = new Dictionary<Type, string>
```
agocke, line 314: The JIT is pretty smart and can often turn `typeof(Generic) == typeof(X)` into a JIT-time constant expression. Worst case, for two or three items it's still much faster to do if-else.

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -209,5 +260,102 @@ internal static void GetDirectoryPaths (out string rootSourceDirectory, out stri
...
+		private static Dictionary<Type, string> msbuildOptionsDict = new Dictionary<Type, string>
+		{
+			{typeof(RequiresUnreferencedCodeAnalyzer), MSBuildPropertyOptionNames.EnableTrimAnalyzer},
+			{typeof(RequiresAssemblyFilesAnalyzer), MSBuildPropertyOptionNames.EnableSingleFileAnalyzer}
+		};
+
+		private static string GetAnalyzerMSBuildPropertyOption<TAnalyzer> () where TAnalyzer : DiagnosticAnalyzer
+		{
+			if (msbuildOptionsDict.ContainsKey (typeof (TAnalyzer))) {
+				return msbuildOptionsDict[typeof (TAnalyzer)];
```
agocke, line 355: This is usually the wrong way to use a `Dictionary`. If you want try to retrieve an item you should almost always use

```C#
if (msbuildOptionsDict.TryGetValue(key, out var value)) {
   return value;
}
...
```

instead of doing a double lookup.

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -209,5 +260,102 @@ internal static void GetDirectoryPaths (out string rootSourceDirectory, out stri
...
+			DiagnosticResult[] baselineExpected,
+			DiagnosticResult[] fixedExpected,
+			int? numberOfIterations = null)
+			where TAnalyzer : DiagnosticAnalyzer, new()
+			where TCodeFix : CodeFixProvider, new()
+		{
+			string analyzerAttribute = GetAttributeDefinition<TAnalyzer> ();
+			string codeFixAttribute = GetAttributeDefinition<TCodeFix> ();
+			var attributeDefinitions = analyzerAttribute;
+			if (string.Equals (analyzerAttribute, codeFixAttribute) == false) {
```
agocke, line 276: `== false`

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -7,15 +7,66 @@
...
+{
+	[AttributeUsage(AttributeTargets.Constructor | AttributeTargets.Event | AttributeTargets.Method | AttributeTargets.Property, Inherited = false, AllowMultiple = false)]
+	public sealed class RequiresAssemblyFilesAttribute : Attribute
+	{
+		public RequiresAssemblyFilesAttribute () { }
+		public string? Message { get; set; }
+		public string? Url { get; set; }
+	}
+}";
+		private const string _requiresUnreferencedCodeAttributeDefinition = @"
```
agocke, line 37: Why are we defining this and `UnconditionalSuppress` ourselves?

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -209,5 +260,87 @@ internal static void GetDirectoryPaths (out string rootSourceDirectory, out stri
...
+			DiagnosticResult[] fixedExpected,
+			int? numberOfIterations = null)
+			where TAnalyzer : DiagnosticAnalyzer, new()
+			where TCodeFix : CodeFixProvider, new()
+		{
+			string analyzerAttribute = GetAttributeDefinition<TAnalyzer> ();
+			string codeFixAttribute = GetAttributeDefinition<TCodeFix> ();
+			var attributeDefinitions = analyzerAttribute;
+			if (!string.Equals (analyzerAttribute, codeFixAttribute)) {
+				attributeDefinitions += codeFixAttribute;
```
agocke, line 277: Do we need this to be conditional?

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -209,5 +260,87 @@ internal static void GetDirectoryPaths (out string rootSourceDirectory, out stri
...
+				TestCaseUtils.UseMSBuildProperties (GetAnalyzerMSBuildPropertyOption<TAnalyzer> ()),
+				baselineExpected);
+
+		internal static DiagnosticResult GetDiagnosticResult<TAnalyzer> ()
+			where TAnalyzer : DiagnosticAnalyzer, new()
+		{
+			return CSharpAnalyzerVerifier<TAnalyzer>.Diagnostic (GetDiagnosticId<TAnalyzer> ());
+		}
+
+		private static string GetDiagnosticId<TAnalyzer> () where TAnalyzer : DiagnosticAnalyzer
```
agocke, line 314: Looking more closely, this method's design seems busted. If the analyzer declares only one diagnostic descriptor then we can use the `CSharpAnalyzerVerifier<TAnalyzer>.Diagnostic()` method. The point of the method taking an ID is that we don't know which ID we're testing for. Below we hardcode `IL3002`, but what about `IL3000` and `IL3001`?

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
```diff
@@ -139,20 +95,23 @@ public class F
...
-	// /0/Test0.cs(27,10): error CS7036: There is no argument given that corresponds to the required formal parameter 'message' of 'RequiresUnreferencedCodeAttribute.RequiresUnreferencedCodeAttribute(string)'
-	DiagnosticResult.CompilerError("CS7036").WithSpan(27, 10, 27, 36).WithArguments("message", "System.Diagnostics.CodeAnalysis.RequiresUnreferencedCodeAttribute.RequiresUnreferencedCodeAttribute(string)"),
-			}
-	);
+			return TestCaseUtils.VerifyCodeFix<RequiresUnreferencedCodeAnalyzer, RequiresUnreferencedCodeCodeFixProvider> (
+				test,
+				fixtest,
+				baselineExpected: new[] {
+					// /0/Test0.cs(9,17): warning IL2026: Using method 'C.M1()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. message.
+					TestCaseUtils.GetDiagnosticResult<RequiresUnreferencedCodeAnalyzer> ().WithSpan (9, 17, 9, 21).WithArguments ("C.M1()", "message", ""),
```
agocke, line 103: Why is this necessary? `VerifyCS.Diagnostic()` doesn't take a diagnostic ID.

agocke: I see -- you want to generalize this in the future for all `Requires` analyzers. Let's save that for a separate PR.

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
tlakollo: We define our own analyzer verifier so we pass the metadata references into our compilation https://github.com/mono/linker/blob/main/test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier%601.cs#L48 in the case of the codefix we rely on the code in Roslyn which has references to Microsoft.NETCore.App.Ref.3.1.0 so we needed to include our own definition

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
tlakollo: Yes, adding two times the same definition will make the compiler to error on ambiguity

test/ILLink.RoslynAnalyzer.Tests/RequiresAssemblyFilesAnalyzerTests.cs
```diff
@@ -2,20 +2,11 @@
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.Testing;
 using Xunit;
-using VerifyCS = ILLink.RoslynAnalyzer.Tests.CSharpAnalyzerVerifier<
```
agocke, line 5: I think you can revert all changes to this file.

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
```diff
@@ -4,61 +4,15 @@
 
 using System;
 using System.Threading.Tasks;
+using ILLink.CodeFix;
 using Microsoft.CodeAnalysis.Testing;
 using Microsoft.CodeAnalysis.Text;
 using Xunit;
-using VerifyCS = ILLink.RoslynAnalyzer.Tests.CSharpCodeFixVerifier<
```
agocke, line 10: Same here, I think you can revert most, if not all changes.

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
agocke: Let's just update the references to 5.0 then.

tlakollo: I was experimenting with the SolutionTransforms and I just manage to do the same as we do on analyzers tests, get the metadata from the SPC in the linker which is 6.0

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
tlakollo: I still see value in using a single generic VerifyDiagnostic that is shared between the analyzers, any reason to revert the changes?

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpCodeFixVerifier`2.cs
```diff
@@ -32,6 +33,7 @@ public Test ()
 					compilationOptions = compilationOptions!.WithSpecificDiagnosticOptions (
 						compilationOptions.SpecificDiagnosticOptions.SetItems (CSharpVerifierHelper.NullableWarnings));
 					solution = solution.WithProjectCompilationOptions (projectId, compilationOptions);
+					solution = solution.WithProjectMetadataReferences (projectId, new List<MetadataReference> { MetadataReference.CreateFromFile (typeof (object).Assembly.Location) });
```
agocke, line 36: I don't understand why this is necessary...

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
agocke: Commented offline: there's a lot of benefit in sharing, but I think we should introduce a new base class for the tests, which is a bigger change. We could do it in this PR, but probably easier to do it in a new PR and get this one merged.

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpCodeFixVerifier`2.cs
tlakollo: Well, it swaps the metadataReferences in the project from netcoreapp3.1 to whatever the linker tests are running in this case netcoreapp6.0. If we use the available [ReferenceAssemblies.Net.Net50](https://github.com/dotnet/roslyn-sdk/blob/main/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/ReferenceAssemblies.cs#L868) the RequiresAssemblyFiles attribute would be missing since it was introduced in 6.0 and there is no ReferenceAssemblies.Net.Net60

tlakollo: Not sure if the methodology is to use ReferenceAssemblies.Net.Net50 and add the additional references manually until net6.0 gets published the ReferenceAssemblies list gets updated and then we can update everything. For me, it seemed like running against the tfm of the tests is not a bad option.

agocke: Ah, I see, you're adding another corelib as a reference (System.Private.CoreLib). That's pretty messy. Let's see if we can do this properly. In `VerifyCodeFixAsync` we create a new `Test` -- I think we can set `ReferenceAssemblies` to `ReferenceAssemblies.Net50`. We'll still have to define the `RequiresAssemblyFiles` attribute ourselves, but that doesn't seem to be a big deal.

tlakollo: Ok I will change both the analyzer and codefix references assemblies to use ReferenceAssemblies.Net.Net50, which would mean we revert https://github.com/mono/linker/pull/1922

mateoatr: If we still want to test that specific APIs produce correct warnings (e.g., `Assembly.Location`) we have to keep the md reference, no? At least until we annotate everything in the runtime and we can start using `ReferenceAssemblies.Net60`.

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
```diff
@@ -42,12 +42,15 @@ public static DiagnosticResult Diagnostic (DiagnosticDescriptor descriptor)
 			SyntaxTree src,
 			(string, string)[]? globalAnalyzerOptions = null)
 		{
+			TestCaseUtils.GetDirectoryPaths (out _, out string testAssemblyPath);
+			var expectationsPath = Path.Combine (testAssemblyPath, "Mono.Linker.Tests.Cases.Expectations.dll");
```
agocke, line 46: This doesn't seem bad, but we could use `typeof(<some-type-we-know).Assembly.Location` here. We know it's safe because we wrote that assembly and know that it doesn't have a ref assembly, and using `typeof` would prevent this from failing by us not building the dependent assembly. If we can use a `typeof` then clearly the assembly was built and referenced.

agocke (approved):
LGTM aside from small comment.

Nice change, lots of code consolidation here!

Issue #2001, opened 2021-04-28
Correctly perform analysis for compiler-generated methods


agocke:
This a blanket issue for analyzing compiler generated code, like async/iterator state machines and nested functions (lambdas, local functions).

The first task is to design the expected user behavior. For instance, if you call a method annotated with `RequiresUnreferencedCode` in a local function, does placing `RequiresUnreferencedCode` on the containing method suppress the warning? Or do you need to place `RequiresUnreferencedCode` on the local function? The C# compiler is somewhat inconsistent about requirements here -- if you place `unsafe` on a method, the local function is considered `unsafe` as well. However, if you use `await` in a local function, making the parent method `async` is not sufficient; the local function must also be declared `async`.

It's up to us to decide the expected user experience.

For data flow analysis, we need to decide if the user experience should be closer to the source code representation, or to the behavior between methods. The linker doesn't normally do interprocedural analysis, and neither does the C# compiler, but for local functions C# does do interprocedural flow analysis for definite assignment. If we want the experience of captured variables to be closer to the experience of local variables, we likely need to do interprocedural analysis. If we think the experience of fields is sufficient, then we don't.

Once we decide on a user experience then need to examine technical requirements: do we think that attributes can be propagated automatically? Or do we want to try to rebuild some of the knowledge of source code from the IL? If we do need to rebuild some knowledge of source code, is there something we need or want from the C# compiler?

marek-safar:
Please keep in mind that the linker is not the C# compiler post analyzer but a generic IL linking/trimming tool. The same experience should be available to developers writing F# code.

agocke:
Sure -- once we decide on the user experience we can design a technical solution which will hopefully be implementable by both the C# and F# compilers. Of course, support for any particular language will be dependent on the language's compiler implementing the feature.

MichalStrehovsky:
The .NET Native analyzer hardcoded the shape of e.g. `async` state machines and was able to propagate dataflow through that. Given .NET Native is where the dataflow analysis originated, we could very cheaply steal that. I intentionally didn't port that part over because I pretty strongly think we should not do that. But it's temptingly cheap, so calling this out.

vitek-karas:
https://github.com/mono/linker/pull/2075 implements support for warning suppression via `UnconditionalSuppressMessage` or `RequiresUnreferencedCode` on iterator and async method bodies.

jtschuster:
[...]

PR #2010, opened 2021-05-03
Compiler generated code handling

vitek-karas:
Also adds tests for the existing patterns - all of them fail. Either lack of implementation or lack of known outcome. The tests are there as they are the complete examples (not just the short versions in the doc).

This is meant to start the discussion on https://github.com/mono/linker/issues/2001 - determine desired behaviors and eventually also about possible solutions.


docs/design/compiler-generated-code-handling.md
```diff
@@ -0,0 +1,336 @@
...
+{
+    type.GetMethod ("BeforeIteratorMethod");
+    yield return 1;
+    type.GetMethod ("AfterIteratorMethod");
+}
+```
+
+Should not produce any warnings, because the `type` variable is properly annotated.
+
+#### Intrinsic data flow
```
marek-safar, line 84: These are all C# specific examples, maybe you could group them that way.

docs/design/compiler-generated-code-handling.md
```diff
@@ -0,0 +1,336 @@
+# Handling of compiler generated code
+
+Modern compilers provide language features which require lot of fancy code generation by the compiler. Not just pure IL generation, but producing new types, methods and fields. An example is `async`/`await` in C# which turns the body of the method into a separate class which implements a state machine.
+
+Lot of the trimming logic relies on attributes authored by the developer. These provide hints to the trimmer especially around ares which are otherwise problematic, like reflection. For example see [reflection-flow](reflection-flow.md) for an example of such attribute.
+
+## Problem
+
+User authored attributes are not propagated to the compiler generated code. For example:
```
marek-safar, line 9: Which languages suffer from this problem?

vitek-karas: C# - is the one I know of. VB probably as well (didn't try).
While I agree that we should try to come up with solutions which could work for all languages, I think it's OK to start by focusing on C# first.

marek-safar: I don't have a problem focusing on C# but the wording used in the design doc should indicate that more clearly.

MichalStrehovsky:
Thank you for writing these down!

I think another interesting aspect is how we handle lack of annotations.

E.g. would we want to tell the user that for this:

```csharp
static void TestLambdaWithCapture(int p)
{
    Action a = () => MethodRequiresUnreferencedCode(p);
}
```

the problem is in the TestLambdaWithCapture method and not with the compiler-generated closure class?

docs/design/compiler-generated-code-handling.md
```diff
@@ -0,0 +1,336 @@
+# Handling of compiler generated code
+
+Modern compilers provide language features which require lot of fancy code generation by the compiler. Not just pure IL generation, but producing new types, methods and fields. An example is `async`/`await` in C# which turns the body of the method into a separate class which implements a state machine.
+
+Lot of the trimming logic relies on attributes authored by the developer. These provide hints to the trimmer especially around ares which are otherwise problematic, like reflection. For example see [reflection-flow](reflection-flow.md) for an example of such attribute.
```
eerhardt, line 5:
```suggestion
Lot of the trimming logic relies on attributes authored by the developer. These provide hints to the trimmer especially around areas which are otherwise problematic, like reflection. For example see [reflection-flow](reflection-flow.md) for an example of such attribute.
```

docs/design/compiler-generated-code-handling.md
```diff
@@ -0,0 +1,336 @@
...
+}
+```
+
+Similarly to the A5 case, the trimmer could propagate the warning suppression to the local function. Unless the function already has suppressions.
+**Question Q1b**: Should method body attributes propagate to local functions? It's possible to add the attribute manually to the local function, so maybe we should simply rely on that.
+
+### A7 - Data flow annotations with local function
+
+```csharp
+static void TestParameterInLocalFunction ([DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicMethods)] Type type)
```
eerhardt, line 194: Do we also need to explicitly call out generic type annotations? Or is it assumed to be the same as this case?

```C#
static void TestParameterInLocalFunction<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] T>()
{
    LocalFunction ();
    void LocalFunction ()
    {
        typeof(T).GetMethod ("InLocalMethod");
    }
}
```

eerhardt:
> **Open question Q1a**: Should method body attributes propagate to lambdas?
> **Question Q1b**: Should method body attributes propagate to local functions?

(My opinion) To a developer these look like different methods. Assuming the C# 10 feature gets in for lambda method attributes, I don't feel it would be necessary to propagate the attributes to lambdas and local functions. But obviously for iterators and async methods, these don't appear to be different methods to a developer. Which is why the attributes applying to the "whole method" (from a dev's point of view) makes sense there.

docs/design/compiler-generated-code-handling.md
vitek-karas: I tried to add text which describes that this should ideally be for all compilers, but that we're focusing on C# right now.

vitek-karas:
@MichalStrehovsky I added a new section about method names in warnings. It's a very good point!
That said I think we've got it already covered for the most part:
* If there are symbols we already point to the correct source line. There are some tweaks we should do, but in the end the warning should look almost identical to what the analyzer produces - meaning it doesn't include the name of the method at all - it solely relies on the source line info for that.
* If there are no symbols, we show the IL method name currently. We can discuss if it's worth trying to improve on this - but it should be a much less important case, since we expect most warnings to have symbols (and those which don't should be collapsed).

vitek-karas:
@eerhardt I'm yet to write down recommended solution and such... but to comment on your recommendation that lambdas and local functions should not need automatic propagation because developers can explicitly annotate them:

In general I agree, but it might be pretty tricky to achieve. Closure+Lambdas and iterators look very similar at the IL level - we would need some new information from the compiler to be able to differentiate (as far as I can tell). But that's an implementation detail. More importantly `SuppressMessageAttribute` and `UnconditionalSuppressMessageAttribute` in Roslyn already applies to lambdas and local functions - it effectively propagates (I assume the latter was modeled by the former in that behavior). So linker should (basically must) implement the same behavior - unless we want to change the behavior in Roslyn.

We could still do something different for RUC, but I'm leaning towards keeping it simple (and thus consistent).

docs/design/compiler-generated-code-handling.md
vitek-karas: Great point - I added this to all the sections as yet another case.

MichalStrehovsky:
> @MichalStrehovsky I added a new section about method names in warnings

There's also a question about dataflow annotations:

```csharp
static IEnumerable<int> TestParameter(Type type)
{
    type.GetMethod("BeforeIteratorMethod");
    yield return 1;
    type.GetMethod("AfterIteratorMethod");
}
```

(I took your example and removed the annotation on `Type`.)

In the ideal world, this would produce a warning saying that parameter `type` doesn't have a PublicMethods annotation (but the way to get to that warning is... complex...).

agocke:
One interesting thing to note: if the code generated ends up in types, we could unambiguously refer to the type via typeof. But if the user code ends up in a bare method, there's no equivalent encoding for method signatures, including custom modifiers.

test/Mono.Linker.Tests.Cases/DataFlow/DataFlowInAsyncCode.cs
```diff
@@ -0,0 +1,57 @@
...
+	public class DataFlowInAsyncCode
+	{
+		public static void Main ()
+		{
+			TestParameter (typeof (TestClass));
+			TestLocalVariable ();
+			TestGenericParameter<TestClass> ();
+		}
+
+		static async void TestParameter ([DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicMethods)] Type type)
```
agocke, line 26: I checked and it looks like `async` and `iterator` methods each get an attribute, `Async/IteratorStateMachine` along with a `typeof` pointing to the definition. We could consider generalizing that in some way.

agocke:
@marek-safar Do you happen to have a design doc for VB or F# for the other linker features? We recently discovered serious deficiencies with VB, so I wonder if the rest of the linker was designed with other languages in mind.

marek-safar:
> @marek-safar Do you happen to have a design doc for VB or F# for the other linker features?

In pre-netcore version of linker we didn't add any language-specific features. There were added later. In our previous testing, we generally focused on C# and F# only.

vitek-karas:
I added first draft for recommended solutions. Currently it's very high-level.

vitek-karas:
I'm going to merge this so that we have at least some documentation in the repo around this.
Part of this was already implemented in .NET 6. The .NET 7 part is being worked on.

PR #2015, opened 2021-05-05
Add RequiresAssemblyFiles CodeFix

### LINKED ISSUE #1986: Refactor Requires Analyzer Code
Given that RequiresUnreferencedCode and RequiresAssemblyFiles analyzers are very similar, do the refactoring of both to share more code and tests.

### LINKED ISSUE #2005: Adding RequiresAssemblyFiles in an Event does not suppress the warning
Using the following code:
```C#
using System;

public class C
{
    [RequiresAssemblyFiles(Message = "message")]
    public int M1() => 0;

    public event EventHandler E1
    {
        add
        {
            var a = M1();
        }
        remove { }
    }
}
```
Should lead to a warning in the add event property, which later on should be able to be suppressed by adding an attribute to the event, like:
```C#
using System;

public class C
{
    [RequiresAssemblyFiles(Message = "message")]
    public int M1() => 0;

    [RequiresAssemblyFiles()]
    public event EventHandler E1
    {
        add
        {
            var a = M1();
        }
        remove { }
    }
}
```
The RequiresAssemblyFiles will continue to warn in the `var a = M1();` assign operation, contrary to the use of UnconditionalSuppressMessage which will silence the warning

tlakollo:
Add RequiresAssemblyFiles CodeFix
Add tests
Fixes #2005
Fixes #1986
Add other fixes that exist in RequiresUnreferencedCode but weren't on RequiresAssemblyFiles


src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -134,17 +142,53 @@ public override void Initialize (AnalysisContext context)
 					CheckCalledMember (operationContext, methodSymbol, dangerousPatterns);
 				}, OperationKind.DelegateCreation);
 
+				static void CheckStaticConstructors (OperationAnalysisContext operationContext,
```
mateoatr, line 145: We could abstract this method away and share it between the analyzers.

src/ILLink.CodeFix/RequiresAssemblyFilesCodeFixProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		internal override SyntaxNode[] GetAttributeArguments (SemanticModel semanticModel, SyntaxNode targetNode, CSharpSyntaxNode containingDecl, SyntaxGenerator generator, Diagnostic diagnostic)
+		{
+			var containingSymbol = semanticModel.GetDeclaredSymbol (containingDecl);
+			var name = semanticModel.GetSymbolInfo (targetNode).Symbol?.Name;
+			if (string.IsNullOrEmpty (name) || HasPublicAccessibility (containingSymbol!)) {
+				return Array.Empty<SyntaxNode> ();
+			} else {
+				return new[] { generator.AttributeArgument ("Message", generator.LiteralExpression ($"Calls {name}")) };
```
mateoatr, line 37: Should also add the `Message` argument in the RUC code fixer?

src/ILLink.CodeFix/RequiresAssemblyFilesCodeFixProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+	public class RequiresAssemblyFilesCodeFixProvider : BaseAttributeCodeFixProvider
+	{
+		private const string s_title = "Add RequiresAssemblyFiles attribute to parent method";
+
+		public sealed override ImmutableArray<string> FixableDiagnosticIds
+			=> ImmutableArray.Create (RequiresAssemblyFilesAnalyzer.IL3000, RequiresAssemblyFilesAnalyzer.IL3001, RequiresAssemblyFilesAnalyzer.IL3002);
+
+		public sealed override async Task RegisterCodeFixesAsync (CodeFixContext context)
+		{
+			await BaseRegisterCodeFixesAsync (context, AttributeableParentTargets.Method | AttributeableParentTargets.Property | AttributeableParentTargets.Event, RequiresAssemblyFilesAnalyzer.FullyQualifiedRequiresAssemblyFilesAttribute, s_title);
```
mateoatr, line 27: nit: should we add `AttributableParentTargets.Constructor`? It doesn't matter that much for now, since both RAF and RUC can target constructors _and_ methods, but we could need more granular analysis in the future (we can ignore this for now if it isn't trivial to differentiate between methods/ctors).

src/ILLink.CodeFix/RequiresAssemblyFilesCodeFixProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Editing;
+
+namespace ILLink.CodeFix
+{
+	[ExportCodeFixProvider (LanguageNames.CSharp, Name = nameof (RequiresAssemblyFilesCodeFixProvider)), Shared]
+	public class RequiresAssemblyFilesCodeFixProvider : BaseAttributeCodeFixProvider
+	{
+		private const string s_title = "Add RequiresAssemblyFiles attribute to parent method";
```
mateoatr, line 20: Why not move from having these strings in the code to using a resource file (this also applies to the `RequiresUnreferenceCode` code fix).

src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
```diff
@@ -108,7 +108,7 @@ internal enum AttributeableParentTargets
 
 		internal abstract SyntaxNode[] GetAttributeArguments (SemanticModel semanticModel, SyntaxNode targetNode, CSharpSyntaxNode declarationSyntax, SyntaxGenerator generator, Diagnostic diagnostic);
 
-		internal static bool HasPublicAccessibility (IMethodSymbol? m)
+		internal static bool HasPublicAccessibility (ISymbol? m)
 		{
```
mateoatr, line 112: Should we add a null check here?

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
tlakollo: I'm testing abstracting to a base class the common code in the analyzers, but I was expecting to open a different PR for that

src/ILLink.CodeFix/RequiresAssemblyFilesCodeFixProvider.cs
tlakollo: This is also included in the refactor code I'm preparing, if you want I can include the changes and make a review of both adding the codefix and refactoring things

mateoatr: I'm fine with having another PR for addressing the RUC changes.

src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
agocke: A property check is an implicit `null` check (note, there are no nullability warnings from here-on).

src/ILLink.CodeFix/RequiresAssemblyFilesCodeFixProvider.cs
agocke: Personally I would add the complexity only when we need it, but changing from "Method" to reflect that it encompasses everything isn't bad.

test/ILLink.RoslynAnalyzer.Tests/RequiresAssemblyFilesAnalyzerTests.cs
```diff
@@ -34,6 +48,30 @@ public sealed class RequiresAssemblyFilesAttribute : Attribute
...
+		{
+			var test = new VerifyCS.Test {
+				TestCode = source + rafDef,
+				FixedCode = fixedSource + rafDef,
+			};
+			test.ExpectedDiagnostics.AddRange (baselineExpected);
+			test.TestState.AnalyzerConfigFiles.Add (
+						("/.editorconfig", SourceText.From (@$"
+is_global = true
+build_property.{MSBuildPropertyOptionNames.EnableSingleFileAnalyzer} = true")));
```
agocke, line 66: `EnableSingleFileAnalyzer`? These tests probably shouldn't be passing 😄

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -134,17 +142,53 @@ public override void Initialize (AnalysisContext context)
...
+					}
+				}
+
 				static void CheckCalledMember (
 					OperationAnalysisContext operationContext,
 					ISymbol member,
 					ImmutableArray<ISymbol> dangerousPatterns)
 				{
+					// Find containing symbol
+					ISymbol? containingSymbol = null;
```
agocke, line 168: Isn't this `FindAttributableParent`?

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -134,17 +142,53 @@ public override void Initialize (AnalysisContext context)
 					CheckCalledMember (operationContext, methodSymbol, dangerousPatterns);
 				}, OperationKind.DelegateCreation);
 
+				static void CheckStaticConstructors (OperationAnalysisContext operationContext,
+					ImmutableArray<IMethodSymbol> constructors)
+				{
+					foreach (var constructor in constructors) {
+						if (constructor.Parameters.Length == 0 && constructor.HasAttribute (RequiresAssemblyFilesAttribute) && constructor.MethodKind == MethodKind.StaticConstructor) {
+							if (constructor.TryGetAttributeWithMessageOnCtor (FullyQualifiedRequiresAssemblyFilesAttribute, out AttributeData? requiresUnreferencedCode)) {
```
agocke, line 150: Why the nested `if`?

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -134,17 +142,53 @@ public override void Initialize (AnalysisContext context)
 					CheckCalledMember (operationContext, methodSymbol, dangerousPatterns);
 				}, OperationKind.DelegateCreation);
 
+				static void CheckStaticConstructors (OperationAnalysisContext operationContext,
+					ImmutableArray<IMethodSymbol> constructors)
+				{
+					foreach (var constructor in constructors) {
+						if (constructor.Parameters.Length == 0 && constructor.HasAttribute (RequiresAssemblyFilesAttribute) && constructor.MethodKind == MethodKind.StaticConstructor) {
```
agocke, line 149: Why are we checking to see if the argument is a static constructor? Isn't that all this method takes?

tlakollo: The way we call the function is with static constructors, but the method is open to receive MethodSymbol so I double-check that we have a constructors. Just in case someone calls it with methods, having said that, we own the analyzers so I don't know if it makes sense to assume that someone else will call this method in a different way since is private

test/ILLink.RoslynAnalyzer.Tests/RequiresAssemblyFilesAnalyzerTests.cs
tlakollo: I agree that shouldnt be passing because we are missing setting IncludeAllContentForSelfExtract to true. Otherwise, we set this MSBuild property and check for it in [RequiresAssemblyFilesAnalyzer.cs](https://github.com/mono/linker/blob/main/src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs#L67)

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
agocke: I don't think it's worth guarding private methods -- but in the case you need to, this is an example of a programming bug, not a recoverable error, so the only correct way to handle is to throw an uncaught exception (crash the process).

I'd recommend Joe Duffy's write-up here on the different between expected failures and logic errors: http://joeduffyblog.com/2016/02/07/the-error-model/#reliability-fault-tolerance-and-isolation

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
tlakollo: I don't think they are the same FindAttributableParent search for a Method/Property/Event/Field to put the attribute on, this function only looks for local functions and lambdas to warn and otherwise falls back to operationContext.ContainingSymbol

test/ILLink.RoslynAnalyzer.Tests/RequiresAssemblyFilesAnalyzerTests.cs
agocke: Ah, my mistake, I thought we changed the name -- but it's just the attribute.

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
agocke: Lambdas can't have attributes so I'm not clear on what this is supposed to be doing.

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -134,17 +142,53 @@ public override void Initialize (AnalysisContext context)
...
-					if (operationContext.ContainingSymbol.HasAttribute (RequiresAssemblyFilesAttribute))
+					if (containingSymbol.HasAttribute (RequiresAssemblyFilesAttribute))
 						return;
 					// In case ContainingSymbol is a property accesor check also for RequiresAssemblyFilesAttribute in the associated property
-					if (operationContext.ContainingSymbol is IMethodSymbol containingSymbol && (containingSymbol.MethodKind == MethodKind.PropertyGet || containingSymbol.MethodKind == MethodKind.PropertySet)
-						&& containingSymbol.AssociatedSymbol!.HasAttribute (RequiresAssemblyFilesAttribute)) {
+					if (containingSymbol is IMethodSymbol methodSymbol &&
+						(methodSymbol.MethodKind == MethodKind.PropertyGet || methodSymbol.MethodKind == MethodKind.PropertySet ||
+						methodSymbol.MethodKind == MethodKind.EventAdd || methodSymbol.MethodKind == MethodKind.EventRemove)
+						&& methodSymbol.AssociatedSymbol!.HasAttribute (RequiresAssemblyFilesAttribute)) {
```
agocke, line 191: I think you could probably just say `methodSymbol.AssociatedSymbol?.HasAttribute(RequiresAssemblyFilesAttribute)` and it would work fine.

src/ILLink.CodeFix/RequiresAssemblyFilesCodeFixProvider.cs
tlakollo: I just remember that I tried this, so this AttributeArgument will create a named argument something like
```C#
[RequiresUnreferencedCodeAttribute("message")]
    public int M1() => 0;
[RequiresUnreferencedCode(Message = "Calls M1")]
    int M2() => M1();
```
which will end up generating `error CS0617: 'Message' is not a valid named attribute argument.`
We would need to change the RUC definition to have Message as a named optional argument

src/ILLink.CodeFix/RequiresAssemblyFilesCodeFixProvider.cs
tlakollo: I changed the name to MethodOrConstructor

src/ILLink.CodeFix/xlf/Resources.cs.xlf
```diff
@@ -0,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.2" xsi:schemaLocation="urn:oasis:names:tc:xliff:document:1.2 xliff-core-1.2-transitional.xsd">
+  <file datatype="xml" source-language="en" target-language="cs" original="../Resources.resx">
+    <body>
+      <trans-unit id="RequiresAssemblyFilesCodeFixTittle">
```
agocke, line 5: We don't have localization so you can probably turn this off (I think there's a property in the analyzer project that's set that does this, something like `<EnableXlfLocalization>false`

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -4,22 +4,20 @@
...
-using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
-using Microsoft.CodeAnalysis.Operations;
 
 namespace ILLink.RoslynAnalyzer
 {
 	[DiagnosticAnalyzer (LanguageNames.CSharp)]
-	public class RequiresUnreferencedCodeAnalyzer : DiagnosticAnalyzer
+	public class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
```
agocke, line 13: `sealed class`

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -0,0 +1,193 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace ILLink.RoslynAnalyzer
+{
+	[DiagnosticAnalyzer (LanguageNames.CSharp)]
+	public abstract class RequiresAnalyzerBase : DiagnosticAnalyzer
+	{
+		protected abstract string RequiresAttributeName { get; }
```
agocke, line 17:
```suggestion
		private protected abstract string RequiresAttributeName { get; }
```

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -0,0 +1,193 @@
...
+
+		protected abstract string RequiresAttributeFullyQualifiedName { get; }
```
agocke, line 19:
```suggestion
		private protected abstract string RequiresAttributeFullyQualifiedName { get; }
```

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -0,0 +1,193 @@
...
+
+		protected abstract DiagnosticTargets AnalyzerDiagnosticTargets { get; }
```
agocke, line 21:
```suggestion
		private protected abstract DiagnosticTargets AnalyzerDiagnosticTargets { get; }
```

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -0,0 +1,193 @@
...
+
+		protected abstract DiagnosticDescriptor RequiresDiagnosticRule { get; }
```
agocke, line 23:
```suggestion
		private protected abstract DiagnosticDescriptor RequiresDiagnosticRule { get; }
```

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -56,120 +55,77 @@ public sealed class RequiresAssemblyFilesAnalyzer : DiagnosticAnalyzer
...
-				if (string.Equals (includesAllContent?.Trim (), "true", StringComparison.OrdinalIgnoreCase))
-					return;
+		protected override DiagnosticDescriptor RequiresDiagnosticRule => s_requiresAssemblyFilesRule;
 
-				var dangerousPatternsBuilder = ImmutableArray.CreateBuilder<ISymbol> ();
+		protected override bool VerifyMSBuildOptions (AnalyzerOptions options, Compilation compilation)
+		{
+			var isSingleFileAnalyzerEnabled = options.GetMSBuildPropertyValue (MSBuildPropertyOptionNames.EnableSingleFileAnalyzer, compilation);
+			if (!string.Equals (isSingleFileAnalyzerEnabled?.Trim (), "true", StringComparison.OrdinalIgnoreCase))
+				return true;
```
agocke, line 70: This feels backwards. My intuition is that a function called `Verify` would return `true` on success and `false` on failure, but this seems to do the opposite.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -0,0 +1,193 @@
...
+
+		private string GetUrlFromAttribute (AttributeData? requiresAssemblyFilesAttribute)
+		{
+			var url = requiresAssemblyFilesAttribute?.NamedArguments.FirstOrDefault (na => na.Key == "Url").Value.Value?.ToString ();
+			return string.IsNullOrEmpty (url) ? "" : " " + url;
+		}
+
+		protected abstract bool TryGetRequiresAttribute (ISymbol member, out AttributeData? requiresAttribute);
+
+		protected abstract bool ReportDangerousPatternDiagnostic (OperationAnalysisContext operationContext, ImmutableArray<ISymbol> dangerousPatterns, ISymbol member);
```
agocke, line 187: Some of these probably deserve doc comments. It's not clear to me what each method is supposed to do.

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -56,120 +55,77 @@ public sealed class RequiresAssemblyFilesAnalyzer : DiagnosticAnalyzer
...
-							operationContext.Operation.Syntax.GetLocation (),
-							member.OriginalDefinition.ToString (),
-							message,
-							url));
-					}
+		protected override bool TryGetRequiresAttribute (ISymbol member, out AttributeData? requiresAttribute)
+		{
+			requiresAttribute = null;
+			foreach (var _attribute in member.GetAttributes ()) {
+				if (_attribute.AttributeClass is var attrClass && attrClass != null &&
```
agocke, line 115: `
```suggestion
				if (_attribute.AttributeClass is { } attrClass &&
```

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -0,0 +1,193 @@
...
+
+		protected abstract ImmutableArray<ISymbol> GetDangerousPatterns (Compilation compilation);
```
agocke, line 189: Is this only used for `RequiresAssemblyFiles`?

tlakollo: Yes, is logic that handles special cases in Single file that generate IL3000 and IL3001

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -0,0 +1,193 @@
...
+					if (ctor is not null) {
+						CheckCalledMember (operationContext, ctor, dangerousPatterns);
+					}
+				}, OperationKind.ObjectCreation);
+
+				context.RegisterOperationAction (operationContext => {
+					var fieldAccess = (IFieldReferenceOperation) operationContext.Operation;
+					if (fieldAccess.Field.ContainingType is INamedTypeSymbol { StaticConstructors: var ctors } &&
+						!SymbolEqualityComparer.Default.Equals (operationContext.ContainingSymbol.ContainingType, fieldAccess.Field.ContainingType)) {
+						CheckStaticConstructors (operationContext, ctors);
```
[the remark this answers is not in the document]
tlakollo: There is another call in CheckCalledMember that handles other IObjectCreationOperations

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -56,120 +55,77 @@ public sealed class RequiresAssemblyFilesAnalyzer : DiagnosticAnalyzer
...
+			if (string.Equals (includesAllContent?.Trim (), "true", StringComparison.OrdinalIgnoreCase))
+				return true;
+			return false;
+		}
 
-				var assemblyType = compilation.GetTypeByMetadataName ("System.Reflection.Assembly");
-				if (assemblyType != null) {
-					// properties
-					ImmutableArrayOperations.AddIfNotNull (dangerousPatternsBuilder, ImmutableArrayOperations.TryGetSingleSymbol<IPropertySymbol> (assemblyType.GetMembers ("Location")));
+		protected override ImmutableArray<ISymbol> GetDangerousPatterns (Compilation compilation)
```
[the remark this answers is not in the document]
tlakollo: I could rename it to SpecialIncompatibleMembers

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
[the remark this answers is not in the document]
agocke: I think we should do all-or-nothing here. Either no support for static constructors, or full support (including in the linker). That should probably go into another PR because it's not going to be a small change.

src/ILLink.CodeFix/xlf/Resources.cs.xlf
agocke: These files should be deleted.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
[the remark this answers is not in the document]
agocke: Still need to remove all static constructor handling

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -0,0 +1,207 @@
...
+				}
+			}
+			return false;
+		}
+
+		protected abstract bool VerifyAttributeArguments (AttributeData attribute);
+
+		protected abstract bool ReportSpecialIncompatibleMembersDiagnostic (OperationAnalysisContext operationContext, ImmutableArray<ISymbol> specialIncompatibleMembers, ISymbol member);
+
+		protected abstract ImmutableArray<ISymbol> GetSpecialIncompatibleMembers (Compilation compilation);
```
agocke, line 203: Since this is optional, maybe make it virtual and return an empty array?

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpCodeFixVerifier`2.cs
```diff
@@ -21,7 +21,7 @@ namespace ILLink.RoslynAnalyzer.Tests
 	/// <typeparam name="TTest">The test implementation to use.</typeparam>
 	public partial class CSharpCodeFixVerifier<TAnalyzer, TCodeFix>
 		   where TAnalyzer : DiagnosticAnalyzer, new()
-		   where TCodeFix : CodeFixProvider, new()
+		   where TCodeFix : Microsoft.CodeAnalysis.CodeFixes.CodeFixProvider, new()
```
agocke, line 24: Why?

tlakollo: By inserting a resource file in ILLink.CodeFixProvider I needed to add it as part of the using statements, creating ambiguity on what is a CodeFixProvider

tlakollo:
If we remove the static constructor handling some of the tests in RequiresCapability will fail, but I just noticed that Marek disabled those test for being unreliable

agocke:
Ah, yes, we should re-enable those. What test fails though? Does the linker already check this functionality?

tlakollo:
I'm not sure which test was failing on CI, but if I remove the static constructor code it will fail on 3 of the RequiresCapability tests
TestStaticCctorRequiresUnreferencedCode
TestStaticCtorMarkingIsTriggeredByFieldAccess
TestStaticCtorTriggeredByMethodCall
Those were added around the same time they were added in the analyzer, @mateoatr  made the change to support these scenarios in the linker

agocke:
Awesome -- didn't realize this had already been implemented in the linker. Yeah, then we should keep this and add test coverage in the analyzer tests.

I forgot the linker tests were disabled though, I'll look at that.

src/ILLink.CodeFix/Resources.resx
```diff
@@ -0,0 +1,129 @@
...
+  <resheader name="version">
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="RequiresAssemblyFilesCodeFixTittle" xml:space="preserve">
```
agocke, line 120: `Title`

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -0,0 +1,249 @@
...
+
+		private protected abstract DiagnosticDescriptor RequiresDiagnosticRule { get; }
+
+		public override void Initialize (AnalysisContext context)
+		{
+			context.EnableConcurrentExecution ();
+			context.ConfigureGeneratedCodeAnalysis (GeneratedCodeAnalysisFlags.ReportDiagnostics);
+			context.RegisterCompilationStartAction (context => {
+				var compilation = context.Compilation;
+				if (VerifyMSBuildOptions (context.Options, compilation) == false)
```
agocke, line 31: `== false`

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
agocke: This doesn't seem resolved to me. Maybe call it `IsAnalyzerEnabled` to be clear on the purpose of the check?

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
agocke: Maybe it would be better if this were called `IsAnalyzerEnabled`? If a function returns a bool the standard C# convention is to use a prefix like `Is` or `Has`. `Verify` sounds more like a `void` function that performs some sort of side effect.

agocke (approved):
LGTM!

Nice work, looks like a lot of cleaner sharing between the analyzers now

PR #2017, opened 2021-05-06
Avoid showing hidden lines in warning location

vitek-karas:
In specific cases (see the test added) there might be a warning comming from compiler generated code with no user code to assign to it. In such case the compiler generates a sequence point with a special line number, marking it as "hidden code". Showing that line number is weird and doesn't make any sense. Especially if the warning comes from method which has several overloads, it's basically impossible to tell which one it was.

So instead use the first sequence point with non-hiden line number.

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Logging/SourceLines.cs
+++ b/test/Mono.Linker.Tests.Cases/Logging/SourceLines.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 using Mono.Linker.Tests.Cases.Expectations.Metadata;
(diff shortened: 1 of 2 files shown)


marek-safar (approved):

mateoatr (approved):

test/Mono.Linker.Tests.Cases/Logging/SourceLines.cs
```diff
@@ -8,27 +9,42 @@ namespace Mono.Linker.Tests.Cases.Logging
...
 			UnrecognizedReflectionPattern ();
+			GenericMethodIteratorWithRequirement<SourceLines> ();
 		}
 
 		[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicConstructors)]
-		private static Type type;
+		static Type type;
 
-		private static Type GetUnknownType ()
+		static Type GetUnknownType ()
```
kasperk81, line 26: Was this change needed for line numbers to show up (should have same meaning in IL)?

vitek-karas: It's just style - in this repo the coding style is to not specify `private` because it's the default. I think for this one I simply ran `lint.cmd` and it changed it like that.

PR #2023, opened 2021-05-11
Implement `DynamicallyAccessedMemberTypes.Interfaces`

vitek-karas:
Implement marking logic for Interfaces
- Transitively mark all interface implementations on a type
- Mark all interface implementations on all base types as well

Implement annotation propagation over `Type.GetInterface`

Add tests for marking behavior as well as annotation behavior

Workaround: Currently the Interfaces enum value can't be used by the linker itself
since it can't yet rely on high enough version of framework. Worked around this
by adding an overlay class and a const value. Once linker is upgraded to high enough
framework version, this workaround should be removed.

### DIFF
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
@@ -10,12 +10,19 @@ using Mono.Cecil;
 
 namespace Mono.Linker
 {
+	// Temporary workaround - should be removed once linker can be upgraded to build against
+	// high enough version of the framework which has this enum value.
+	internal static class DynamicallyAccessedMemberTypesOverlay
+	{
+		public const DynamicallyAccessedMemberTypes Interfaces = (DynamicallyAccessedMemberTypes) 0x2000;
+	}
+
 	internal static class DynamicallyAccessedMembersBinder
 	{
-		// Returns the members of the type bound by memberTypes. For MemberTypes.All, this returns a single null result.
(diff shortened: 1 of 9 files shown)


MichalStrehovsky (approved):

src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -298,5 +310,22 @@ public static IEnumerable<EventDefinition> GetEventsOnTypeHierarchy (this TypeDe
 				onBaseType = true;
 			}
 		}
+
+		public static IEnumerable<InterfaceImplementation> GetAllInterfaceImplementations (this TypeDefinition type, LinkContext context)
```
marek-safar, line 314: Is it expected that it could return the same interface multiple times?

marek-safar: You could consider moving this to TypeDefinitionExtensions

marek-safar (approved):

src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
vitek-karas: Yes - it's OK if it returns the same interface multiple times. It was not worth optimizing (MarkType will almost immediately return if the type is already marked so adding another layer of cache here is not worth it).

For this reason I don't want to move this to TypeDefinitionsExtensions - it's not "Clean" - but for this purpose it doesn't have to be.

PR #2029, opened 2021-05-12
Try to re-enable RequiresCapability tests

agocke:
Found some non-determinism in finding the right file -- it should now look through all the files.

### DIFF
--- a/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
-using System.Linq;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -14,40 +14,33 @@
 
 namespace ILLink.RoslynAnalyzer.Tests
 {
-	public class TestCaseUtils
+	public abstract class TestCaseUtils
```
tlakollo, line 17: Why is this now an abstract class?

agocke: It only contains helpers -- it probably should have been an abstract class in the beginning.

In fact, everything in it is static. It should be a static class, but XUnit requires that all `MemberData` work against members in the same class, so the only way to reuse the functionality is to put it in a base class and use inheritance.

test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
```diff
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections.Generic;
+using System.Linq;
```
tlakollo, line 6: Nit: additional using

tlakollo (approved):
Looks good to me

PR #2032, opened 2021-05-15
Remove custom marking of nested types for event sources

### LINKED ISSUE #1949: Remove special handling of EventSource
Linker has some special handling of EventSource types:
https://github.com/mono/linker/blob/55bd0ebd00ddbf2cd817314b1cf6824003cfb463/src/linker/Linker.Steps/MarkStep.cs#L1701-L1705

With the changes to support EventSource fully with annotations, this should be removed once the necessary framework changes are made.

marek-safar:
Fixes #1949


marek-safar:
@vitek-karas when should the nested types be kept? It does not look like the libraries were annotated yet for the tests we have but I'm not sure they are correct.

vitek-karas (approved):
Ideally we should test downlevel behavior. But let's track that in here: https://github.com/mono/linker/issues/2040

Issue #1282, opened 2020-06-18
Consider stripping names of fields that don't matter


MichalStrehovsky:
Now that linker has a pretty good idea what fields are accessed through reflection (and warns whenever its not sure), we could consider stripping names of fields that are not observable.

A field name that is not observable could be defined as:
* Private field, or field on a private nested type
* Internal field, or field on an internal type when no InternalsVisibleTo is present
* Field that linker didn't see as used through reflection
* Not on a generic type (because of MemberRefs to TypeSpecs).
* If we have whole program view, we could extend to publics that are not accessed from other assemblies in the closure.

Stripping name means setting accessibility to `compilercontrolled` (`IsCompilerControlled = true` in Cecil) and setting name to null.

I experimentally tried this on Mono's WASM CoreLib and saw 1.3% savings. I didn't actually try whether things work. `compilercontrolled` is such a corner case that I wouldn't be surprised if there's bugs.

It would be more significant if we could do this to methods and types, but it's too easy to get to names of those without linker realizing the name is observable (e.g. `o.GetType().FullName` for types, or `new StackTrace().ToString()` for method names).

benaadams:
> It would be more significant if we could do this to methods and types, but it's too easy to get to names of those without linker realizing the name is observable 

Could strip method names marked with `System.Diagnostics.StackTraceHiddenAttribute` https://github.com/dotnet/runtime/issues/29681; though probably not significant?

MichalStrehovsky:
> Could strip method names marked with System.Diagnostics.StackTraceHiddenAttribute dotnet/runtime#29681; though probably not significant?

That would help a little bit, but we would also have to prove the address of the method wasn't taken (e.g. we didn't create a delegate that someone could call `GetMethodInfo` on - because that becomes visible too).

PR #2033, opened 2021-05-15
Trim field names in compiler generated types

marek-safar:
Contributes to #1282

Saves about 10k (uncompressed) for default Blazor template

### DIFF
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -307,7 +307,7 @@ namespace Mono.Linker.Steps
 		protected virtual void SweepType (TypeDefinition type)
 		{
 			if (type.HasFields)
-				SweepCollectionWithCustomAttributes (type.Fields);
+				SweepFields (type);
 
 			if (type.HasMethods)
 				SweepMethods (type.Methods);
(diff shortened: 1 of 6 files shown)


MichalStrehovsky:
What does this do to:

```csharp
using System;

var g = new Gen<string>();
var e = g.Enumerate(1, 2).GetEnumerator();
e.MoveNext();

class Gen<T>
{
    public IEnumerable<int> Enumerate(int x, int y)
    {
        yield return x;
        yield return y;
    }
}
```

I expect this program to be broken with this optimization.

I tried to capture the rules of when and how this can be done safely in https://github.com/mono/linker/issues/1282 (it also covers more cases than just fields on types marked with CompilerGeneratedAttribute).

marek-safar:
Yep, that fails with `System.MissingFieldException` but I think excluding fields with TypeSpec field type should covert that, right?

MichalStrehovsky:
Yes, that would fix that one.

But CompilerGeneratedAttribute is a general purpose attribute that is not specific to stuff generated by e.g. Roslyn. People can just slap it on stuff e.g. generated by source generators or other tools. There's no guarantees that there wouldn't be another assembly trying to access the field through a MemberRef, hitting MissingFieldException again. We need to make sure the field is not accessible from outside the assembly (or that no other assembly accesses it, if we have a global view).

I'm not sure if there's a way to approximate #1282 without actually doing everything in #1282.

marek-safar:
> There's no guarantees that there wouldn't be another assembly trying to access the field through a MemberRef, hitting MissingFieldException again.

Why does MemberRef matter? I thought that if we record all fields used via TypeSpec to require name there should be no MissingFieldException.

MichalStrehovsky:
TypeSpec is only used for generic classes, which is why the above repro program breaks (the yield return state machine is generic).

I meant the general:

## Assembly A

```csharp
[CompilerGenerated]
public class Foo
{
    public int Field;
}
```

## Assembly B

```csharp
Foo.Field = 1234;
```

There's no TypeSpec.

PR #2038, opened 2021-05-17
Add DynamicallyAccessedMembers analyzer 

mateoatr:
This PR adds basic support for `DynamicallyAccessedMembers` in the linker's Roslyn analyzer. It focuses on the diagnostics concerning unmatched member annotations (warnings [IL2067](https://github.com/mono/linker/blob/main/docs/error-codes.md#il2067-trim-analysis-target-parameter-argument-does-not-satisfy-dynamicallyaccessedmembersattribute-in-call-to-target-method-the-parameter-source-parameter-of-method-source-method-does-not-have-matching-annotations-the-source-value-must-declare-at-least-the-same-requirements-as-those-declared-on-the-target-location-it-is-assigned-to) to [IL2091](https://github.com/mono/linker/blob/main/docs/error-codes.md#il2091-trim-analysis-target-generic-parameter-generic-argument-does-not-satisfy-dynamicallyaccessedmembersattribute-in-target-method-or-type-the-generic-parameter-source-target-parameter-of-source-method-or-type-does-not-have-matching-annotations-the-source-value-must-declare-at-least-the-same-requirements-as-those-declared-on-the-target-location-it-is-assigned-to)). The idea is to have this merged into a feature branch that can be merged with master once the analyzer is in a good shape.

- [ ] Validate scenarios where no warnings should be produced.
- [ ] Add support for `UnrecognizedReflectionAccessPattern` in the analyzer tests infra.
   - [ ] Validate analyzer with linker's DataFlow tests.
- [ ] Consolidate resource files in one place.


src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,683 @@
...
+namespace ILLink.RoslynAnalyzer
+{
+	[DiagnosticAnalyzer (LanguageNames.CSharp)]
+	public class DynamicallyAccessedMembersAnalyzer : DiagnosticAnalyzer
+	{
+		internal const string DynamicallyAccessedMembers = nameof (DynamicallyAccessedMembers);
+		internal const string DynamicallyAccessedMembersAttribute = nameof (DynamicallyAccessedMembersAttribute);
+
+		#region SupportedDiagnostics
+		public const string IL2067 = "IL2067";
```
agocke, line 24: We should give this identifiers real names so we can easily differentiate in code.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,683 @@
...
+			string missingMemberTypes,
+			SyntaxNodeAnalysisContext context,
+			bool sourceIsMethodReturnType = false)
+		{
+			Debug.Assert (!string.IsNullOrEmpty (missingMemberTypes));
+
+			switch ((sourceSymbol, targetSymbol)) {
+			case (IParameterSymbol sourceParameter, IParameterSymbol targetParameter):
+				context.ReportDiagnostic (Diagnostic.Create (
+					s_IL2067,
```
agocke, line 494: Is the only difference in many of these diagnostics that the source is one type of symbol and the target is another type of symbol?

This seems like a difficult design. As a user this would mean that if I want to silence all the conversion warnings in a method I would have to individually silence each warning code, but I don't see what value I get from having a specific warning code for each one.

These all feel like instances of the same warning.

agocke: At minimum, we should try to extract some of this code into a helper, but really I think we should try to figure out why we need this many warning codes.

agocke: Ah I'm looking at the messages now -- some of them are significantly different. We could probably share more but I see why it's not obvious how to do so. Probably a good investment though.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,683 @@
...
+			s_IL2070, s_IL2071, s_IL2072, s_IL2073, s_IL2074, s_IL2075, s_IL2076, s_IL2077, s_IL2078, s_IL2079,
+			s_IL2080, s_IL2081, s_IL2082, s_IL2083, s_IL2084, s_IL2085, s_IL2086, s_IL2087, s_IL2088, s_IL2089,
+			s_IL2090, s_IL2091);
+		#endregion
+
+		public override void Initialize (AnalysisContext context)
+		{
+			context.EnableConcurrentExecution ();
+			context.ConfigureGeneratedCodeAnalysis (GeneratedCodeAnalysisFlags.ReportDiagnostics);
+			context.RegisterSyntaxNodeAction (DynamicallyAccessedMembersAnalyze,
```
agocke, line 312: I think an IOperation context makes more sense here -- syntax seems hard because you may have a lot of different syntaxes for the same semantic operation.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,683 @@
...
+				typeArgumentSymbol.TryGetDynamicallyAccessedMemberTypes (out var damtOnTypeArgument);
+				if (!SourceHasMatchingAnnotations (damtOnTypeArgument, damtOnTypeParameter, out var missingAnnotationsOnTypeArgument)) {
+					ReportUnmatchedAnnotations (invocationExpression.GetLocation (),
+						typeArgumentSymbol, typeParameterSymbol, missingAnnotationsOnTypeArgument, context);
+					continue;
+				}
+			}
+
+			// For all parameters in the called method, check if the passed arguments have matching annotations.
+			for (int i = 0; i < invokedMethodSymbol.Parameters.Length; i++) {
```
agocke, line 366: I think what you want to use here is an `IInvocationOperation`, because then you can loop through the `IArgumentOperations`. Otherwise you'll be stuck trying to match up arguments with parameters, which is just insanely difficult given things like named arguments, optional parameters, `params` arguments, etc.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,683 @@
...
+	[DiagnosticAnalyzer (LanguageNames.CSharp)]
+	public class DynamicallyAccessedMembersAnalyzer : DiagnosticAnalyzer
+	{
+		internal const string DynamicallyAccessedMembers = nameof (DynamicallyAccessedMembers);
+		internal const string DynamicallyAccessedMembersAttribute = nameof (DynamicallyAccessedMembersAttribute);
+
+		#region SupportedDiagnostics
+		public const string IL2067 = "IL2067";
+		static readonly DiagnosticDescriptor s_IL2067 = new (
+			IL2067,
```
vitek-karas, line 26: Rant: This feels like a great opportunity for some kind of source generator - having to write this much code is crazy.

For now - can we refactor this a bit - create a helper which only takes the parts which are different and call that instead of repeating the same parameters over and over.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
vitek-karas: We have a helper in linker to produce the right warning.

The many different warning codes were introduce actually according to your (@agocke) guidance that each unique message should have its own warning code. It's true that silencing them all is tricky, but I see that as a benefit (you should not want to do that manually, have the `SuppressTrimAnalysisWarnings` for it) - the simplest way to silence them in code is to add `RequiresUnreferencedCode` onto the method which produces them.

PR #2039, opened 2021-05-17
Add line number/position info to warnings from XML files

vitek-karas:
This basically improves the precision of errors we report from XML files by providing sourceline/column information.

Required refactoring how we create XML, since I wanted only one place to actually parse the XML (since we need to do it in a special way to get line/position info).

Also required to change to use XDocument instead of XPathDocument, as the latter doesn't store line/position info. This will slightly increase memory consumption for any XML processed, but since that number is already very low and we don't expect it to go high, it's not a problem.


sbomer (approved):
Nice!

marek-safar (approved):

PR #2041, opened 2021-05-18
Add line info in IL2001 and IL2002

mateoatr:
Following #2039, this adds line source line info. for warnings IL2001 and IL2002. This will cause the warnings to be triggered even in cases where the annotated types are trimmed.

Note that we still have to keep producing these warnings in `ApplyPreserveInfo` for cases where a user applies a `TypePreserve` using the Annotations API:

https://github.com/mono/linker/blob/30f2498c2a3de1f7e236d5793f5f1aca6e5ba456/src/linker/Linker.Steps/MarkStep.cs#L2427-L2434

### DIFF
--- a/src/linker/Linker.Steps/DescriptorMarker.cs
+++ b/src/linker/Linker.Steps/DescriptorMarker.cs
@@ -101,8 +101,21 @@ namespace Mono.Linker.Steps
 			Debug.Assert (ShouldProcessElement (nav));
 
 			TypePreserve preserve = GetTypePreserve (nav);
-			if (preserve != TypePreserve.Nothing)
-				_context.Annotations.SetPreserve (type, preserve);
+			switch (preserve) {
+			case TypePreserve.Fields when !type.HasFields:
+				LogWarning ($"Type {type.GetDisplayName ()} has no fields to preserve", 2001, nav);
+				break;
+
+			case TypePreserve.Methods when !type.HasMethods:
+				LogWarning ($"Type {type.GetDisplayName ()} has no methods to preserve", 2002, nav);
+				break;
+
+			default:
+				if (preserve != TypePreserve.Nothing)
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Steps/DescriptorMarker.cs
```diff
@@ -101,8 +101,21 @@ protected override void ProcessType (TypeDefinition type, XPathNavigator nav)
...
+			case TypePreserve.Fields when !type.HasFields:
+				LogWarning ($"Type {type.GetDisplayName ()} has no fields to preserve", 2001, nav);
+				break;
+
+			case TypePreserve.Methods when !type.HasMethods:
+				LogWarning ($"Type {type.GetDisplayName ()} has no methods to preserve", 2002, nav);
+				break;
+
+			default:
+				if (preserve != TypePreserve.Nothing)
```
sbomer, line 114: I'd just replace this with a `TypePreserve.All` case.

test/Mono.Linker.Tests.Cases/LinkXml/LinkXmlErrorCases.cs
```diff
@@ -28,14 +28,12 @@ public static void Main ()
 		}
 
 		[Kept]
-		[ExpectedWarning ("IL2001", "TypeWithNoFields")]
```
sbomer, line 31: Could we check that the XML produces IL2001/IL2002 instead?

sbomer (commented):
I like reporting warnings up-front for the XML, but I would prefer to be consistent between custom steps and the XML. Maybe we should consistently report the warnings from `Annotations.SetPreserve (typeWithoutFields, TypePreserve.Fields)` instead.

sbomer (approved):
LGTM, thanks!

vitek-karas (approved):

marek-safar (approved):

PR #2042, opened 2021-05-18
Flow C++/CLI assemblies into the linker

mateoatr:
This will stop the linker from dropping C++/CLI assemblies passed to `ComputeManageAssemblies` and save us some time by not loading the `ModuleDefinition` to check if the assembly is MC++.

Note that for any C++/CLI assembly that is passed to the linker, its action will always be `copy`.


jkoritzinsky:
We need to make sure that the C++/CLI assemblies won’t be round-tripped through Cecil and will just be copied directly from source to destination. They have dependencies on RVA locations, which Cecil does not support expressing.

vitek-karas:
@jkoritzinsky this:
>Note that for any C++/CLI assembly that is passed to the linker, its action will always be copy.

`copy` has been fixed in the linker to mean "don't touch this" - so we now guarantee that C++/CLI assemblies will not roundtrip through Cecil.

src/linker/Linker/LinkContext.cs
```diff
@@ -308,7 +308,18 @@ public void RegisterAssembly (AssemblyDefinition assembly)
...
+			}
+
+			static bool IsCPPCLIAssembly (ModuleDefinition module)
+			{
+				foreach (var type in module.Types)
+					if (type.Namespace == "<CppImplementationDetails>" ||
+						type.Namespace == "<CrtImplementationDetails>")
+						return true;
+
+				return false;
```
vitek-karas, line 322: Why not move this into the `CalculateAssemblyAction` - that name actually suggests it would do things like this (unlike `RegisterAssembly).

vitek-karas (commented):
Can we come up with some test? Even if it involves checked in binary...

src/linker/Linker/LinkContext.cs
```diff
@@ -308,7 +308,18 @@ public void RegisterAssembly (AssemblyDefinition assembly)
 		{
 			if (SeenFirstTime (assembly)) {
 				SafeReadSymbols (assembly);
-				Annotations.SetAction (assembly, CalculateAssemblyAction (assembly));
+				Annotations.SetAction (assembly,
+					IsCPPCLIAssembly (assembly.MainModule) ? AssemblyAction.Copy : CalculateAssemblyAction (assembly));
```
marek-safar, line 312: I think this logic needs more tweaks. Do we really want to change the action all the time? What if someone sets "skip" action for the assembly?

I think we should downgrade to copy only for action we cannot handle and issue a warning that about the limitation.

vitek-karas: > I think we should downgrade to copy only for action we cannot handle and issue a warning that about the limitation.

I agree with the first part - if somebody sets `skip` it should stick, but we should change `link` and `copyused` to `copy`.
I don't think we should warn always - the way it works currently, the SDK relies on linker doing the right thing, so we would be warning basically by default.
Maybe we should warn only if the action was specified explicitly for the assembly, but not warn if we use the default action (if that's possible).

sbomer: I think it'd be a good idea to warn if we are changing the action from an explicit action passed on the command-line.

mateoatr: Added a warn for the case where a non-skip/copy action was specified.

sbomer (approved):
Otherwise LGTM

docs/error-codes.md
```diff
@@ -1590,6 +1590,9 @@ void RequirePublicMethodOnAType(
 }
 ```
 
+#### `IL2106`: Invalid assembly action '{action}' in for assembly '{assembly}'. C++/CLI assemblies can only be copied or skipped.
```
vitek-karas, line 1593:
```suggestion
#### `IL2106`: Invalid assembly action '{action}' specified for assembly '{assembly}'. C++/CLI assemblies can only be copied or skipped. 
```

PR #2044, opened 2021-05-18
Parsing update for native header files with feature switch

LakshanF:
Added support to parse native header files that have the [feature switch macro](https://github.com/dotnet/runtime/pull/52940) defined. Should be a no-op if the feature switch macro is not defined in a header file

### DIFF
--- a/src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
+++ b/src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
@@ -56,24 +56,30 @@ namespace ILLink.Tasks
 
 		/// <summary>
 		/// Helper utility to track feature switch macros in header file
-		/// This type is used in a dictionary as a key and only uses
-		/// the feature name since value and default are not expected to change
+		/// This type is used in a dictionary as a key
 		/// </summary>
-		class FeatureSwitchMembers
+		readonly struct FeatureSwitchMembers
 		{
-			public string feature;
-			public string featureValue;
-			public string featureDefault;
+			public string Feature { get; }
+			public string FeatureValue { get; }
+			public string FeatureDefault { get; }
+
+			public FeatureSwitchMembers (string feature, string featureValue, string featureDefault)
+			{
+				Feature = feature;
+				FeatureValue = featureValue;
+				FeatureDefault = featureDefault;
+			}
 
 			public override int GetHashCode ()
 			{
-				return feature.GetHashCode ();
+				return HashCode.Combine (Feature, FeatureValue, FeatureDefault);
 			}
 
 			public override bool Equals (object obj)
 			{
-				FeatureSwitchMembers other = obj as FeatureSwitchMembers;
-				return other.feature.Equals (feature);
+				FeatureSwitchMembers other = (FeatureSwitchMembers) obj;
+				return other.Feature.Equals (Feature) && other.FeatureValue.Equals (FeatureValue) && other.FeatureDefault.Equals (FeatureDefault);
 			}
 		}
 
@@ -163,8 +169,7 @@ namespace ILLink.Tasks
 			string[] types = File.ReadAllLines (typeFile);
 			DefineTracker defineTracker = new DefineTracker (defineConstants, Log, typeFile);
 			string classId;
-			bool insideFeatureSwitch = false;
-			FeatureSwitchMembers currentFeatureSwitch = null;
+			FeatureSwitchMembers? currentFeatureSwitch = null;
 
 			foreach (string def in types) {
 				if (defineTracker.ProcessLine (def) || !defineTracker.IsActiveSection)
@@ -172,26 +177,22 @@ namespace ILLink.Tasks
 
 				//We will handle BEGIN_ILLINK_FEATURE_SWITCH and END_ILLINK_FEATURE_SWITCH
 				if (def.StartsWith ("BEGIN_ILLINK_FEATURE_SWITCH")) {
-					if (insideFeatureSwitch) {
+					if (currentFeatureSwitch.HasValue) {
 						Log.LogError ($"Could not figure out feature switch status in '{typeFile}' for line {def}");
 					} else {
-						insideFeatureSwitch = true;
 						char[] separators = { ',', '(', ')', ' ', '\t', '/' };
 						string[] featureSwitchElements = def.Split (separators, StringSplitOptions.RemoveEmptyEntries);
-						currentFeatureSwitch = new FeatureSwitchMembers ();
-						currentFeatureSwitch.feature = featureSwitchElements[1];
-						currentFeatureSwitch.featureValue = featureSwitchElements[2];
-						currentFeatureSwitch.featureDefault = featureSwitchElements[3];
-						if (!featureSwitchMembers.ContainsKey (currentFeatureSwitch)) {
-							featureSwitchMembers.Add (currentFeatureSwitch, new Dictionary<string, ClassMembers> ());
+						currentFeatureSwitch = new FeatureSwitchMembers (featureSwitchElements[1], featureSwitchElements[2], featureSwitchElements[3]);
+						if (!featureSwitchMembers.ContainsKey (currentFeatureSwitch.Value)) {
+							featureSwitchMembers.Add (currentFeatureSwitch.Value, new Dictionary<string, ClassMembers> ());
 						}
 					}
 				}
 				if (def.StartsWith ("END_ILLINK_FEATURE_SWITCH")) {
-					if (!insideFeatureSwitch) {
+					if (!currentFeatureSwitch.HasValue) {
 						Log.LogError ($"Could not figure out feature switch status in '{typeFile}' for line {def}");
 					} else {
-						insideFeatureSwitch = false;
+						currentFeatureSwitch = null;
 					}
 				}
 
@@ -206,7 +207,7 @@ namespace ILLink.Tasks
 					classId = defElements[1];               // APP_DOMAIN
 					string classNamespace = defElements[2]; // System
 					string className = defElements[3];      // AppDomain
-					AddClass (classNamespace, className, classId, false, insideFeatureSwitch, currentFeatureSwitch);
+					AddClass (classNamespace, className, classId, false, currentFeatureSwitch);
 				} else if (def.StartsWith ("DEFINE_CLASS_U(")) {
 					// E.g., DEFINE_CLASS_U(System,                 AppDomain,      AppDomainBaseObject)
 					string classNamespace = defElements[1]; // System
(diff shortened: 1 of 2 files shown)


src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -54,9 +54,33 @@ class ClassMembers
 			public HashSet<string> fields;
 		}
 
+		/// <summary>
+		/// Helper utility to track feature switch macros in header file
+		/// This type is used in a dictionary as a key and only uses
+		/// the feature name since value and default are not expected to change
+		/// </summary>
+		class FeatureSwitchMembers
```
mateoatr, line 62: Based on the above summary sounds like this could be a `readonly struct`.

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -288,34 +403,48 @@ void OutputXml (string iLLinkTrimXmlFilePath, string outputFileName)
 			doc.Save (outputFileName);
 		}
 
-		void AddClass (string classNamespace, string className, string classId, bool keepAllFields = false)
+		void AddClass (string classNamespace, string className, string classId, bool keepAllFields = false, bool featureSwitchOn = false, FeatureSwitchMembers featureSwitch = null)
```
vitek-karas, line 406: Why not use `featureSwitch == null` as the indicator for FS presence - I don't see a need for another parameter.

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -288,34 +403,48 @@ void OutputXml (string iLLinkTrimXmlFilePath, string outputFileName)
...
+						classNamesToClassMembers[fullClassName] = members;
+					}
+					members.keepAllFields |= keepAllFields;
+				} else {
+					Dictionary<string, ClassMembers> currentFeatureSwitchMembers = featureSwitchMembers[featureSwitch];
+					if (!currentFeatureSwitchMembers.TryGetValue (fullClassName, out ClassMembers members)) {
+						members = new ClassMembers ();
+						currentFeatureSwitchMembers[fullClassName] = members;
+					}
+					members.keepAllFields |= keepAllFields;
```
vitek-karas, line 425: This can be moved outside of the if.

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -239,6 +290,70 @@ void OutputXml (string iLLinkTrimXmlFilePath, string outputFileName)
...
+					foreach (string typeName in featureSwitchMembers[fsMembers].Keys) {
+						XmlNode typeNode = doc.CreateElement ("type");
+						XmlAttribute typeFullName = doc.CreateAttribute ("fullname");
+						typeFullName.Value = typeName;
+						typeNode.Attributes.Append (typeFullName);
+
+						ClassMembers members = featureSwitchMembers[fsMembers][typeName];
+
+						if (members.keepAllFields) {
+							XmlAttribute preserve = doc.CreateAttribute ("preserve");
```
vitek-karas, line 323: Can we deduplicate the code here? I would expect the logic how to write ClassMembers to be identical with or without feature switch.

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -54,9 +54,33 @@ class ClassMembers
 			public HashSet<string> fields;
 		}
 
+		/// <summary>
+		/// Helper utility to track feature switch macros in header file
+		/// This type is used in a dictionary as a key and only uses
+		/// the feature name since value and default are not expected to change
```
vitek-karas, line 60: > the feature name since value and default are not expected to change

I think this is a dangerous assumption to make. Please change this so that all three values are part of the key.

vitek-karas (commented):
Please add tests. (There is an existing test already, so adding to it or adding a new one should not be that hard)

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -357,85 +322,92 @@ void OutputXml (string iLLinkTrimXmlFilePath, string outputFileName)
...
 
-					if (!members.keepAllFields && (members.fields != null)) {
-						foreach (string field in members.fields) {
-							XmlNode fieldNode = doc.CreateElement ("field");
-							XmlAttribute fieldName = doc.CreateAttribute ("name");
-							fieldName.Value = field;
-							fieldNode.Attributes.Append (fieldName);
-							typeNode.AppendChild (fieldNode);
-						}
+		static void AddXmlTypeNode (XmlDocument doc, XmlNode assemblyNode, string typeName, Dictionary<string, ClassMembers> classToClassMems)
```
vitek-karas, line 330: Why not just pass `ClassMembers` - we already pass the type name, so the additional lookup is unnecessary (and the callers all do a foreach so they can directly specify the value without lookup as well.

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -139,11 +169,33 @@ void ProcessMscorlib (string typeFile)
...
 					continue;
 
+				//We will handle BEGIN_ILLINK_FEATURE_SWITCH and END_ILLINK_FEATURE_SWITCH
+				if (def.StartsWith ("BEGIN_ILLINK_FEATURE_SWITCH")) {
+					if (currentFeatureSwitch.HasValue) {
+						Log.LogError ($"Could not figure out feature switch status in '{typeFile}' for line {def}");
+					} else {
+						char[] separators = { ',', '(', ')', ' ', '\t', '/' };
+						string[] featureSwitchElements = def.Split (separators, StringSplitOptions.RemoveEmptyEntries);
+						currentFeatureSwitch = new FeatureSwitchMembers (featureSwitchElements[1], featureSwitchElements[2], featureSwitchElements[3]);
```
vitek-karas, line 185: we should check that the returned number of segments is as expected and fail otherwise... to avoid surprises.

vitek-karas (approved):
Looks good.

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -313,7 +317,7 @@ void OutputXml (string iLLinkTrimXmlFilePath, string outputFileName)
 					featureAssemblyNode.Attributes.Append (featureDefault);
 
 					foreach (string typeName in featureSwitchMembers[fsMembers].Keys) {
-						AddXmlTypeNode (doc, featureAssemblyNode, typeName, featureSwitchMembers[fsMembers]);
+						AddXmlTypeNode (doc, featureAssemblyNode, typeName, featureSwitchMembers[fsMembers][typeName]);
```
vitek-karas, line 320: This can be rewritten like this to simplify it even more:
```C#
foreach (var type in featureSwitchMembers[fsMembers]) {
    AddXmlTypeNode (doc, featureAssemblyNode, type.Key, type.Value);
}
```

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -316,8 +316,8 @@ void OutputXml (string iLLinkTrimXmlFilePath, string outputFileName)
 					featureDefault.Value = fsMembers.FeatureDefault;
 					featureAssemblyNode.Attributes.Append (featureDefault);
 
-					foreach (string typeName in featureSwitchMembers[fsMembers].Keys) {
-						AddXmlTypeNode (doc, featureAssemblyNode, typeName, featureSwitchMembers[fsMembers][typeName]);
+					foreach (var type in featureSwitchMembers[fsMembers]) {
+						AddXmlTypeNode (doc, featureAssemblyNode, type.Key, type.Value);
```
vitek-karas, line 320: Same can be done for the other callsite of AddXmlTypeNode

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -239,83 +295,130 @@ void OutputXml (string iLLinkTrimXmlFilePath, string outputFileName)
...
-					if (members.keepAllFields) {
-						XmlAttribute preserve = doc.CreateAttribute ("preserve");
-						preserve.Value = "fields";
-						typeNode.Attributes.Append (preserve);
-					} else if ((members.fields == null) && (members.methods == null)) {
-						XmlAttribute preserve = doc.CreateAttribute ("preserve");
-						preserve.Value = "nothing";
-						typeNode.Attributes.Append (preserve);
-					}
+				AddXmlTypeNode (doc, assemblyNode, typeName, classNamesToClassMembers[typeName]);
```
vitek-karas, line 329: This is the second callsite - again the foreach above can be done over the dictionary - getting KeyValuePair and using that instead of repeating the lookup here.

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -239,83 +299,130 @@ void OutputXml (string iLLinkTrimXmlFilePath, string outputFileName)
...
 		{
 			string className = classIdsToClassNames[classId];
+			ClassMembers members;
 
-			ClassMembers members = classNamesToClassMembers[className];
+			if (!featureSwitch.HasValue) {
+				members = classNamesToClassMembers[className];
+			} else {
+				members = featureSwitchMembers[featureSwitch.Value][className];
+			}
```
sbomer, line 417: Would it be worth combining featureSwitchMembers and classNamesToClassMembers into a Dictionary<FeatureSwitchMembers?, ...> where a null key is just the default (no feature settings)? That would simplify some of the dictionary logic.

sbomer:
In case you don't know - you can run lint.cmd/sh to fix formatting issues

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -54,9 +54,43 @@ class ClassMembers
...
+		/// Helper utility to track feature switch macros in header file
+		/// This type is used in a dictionary as a key
+		/// </summary>
+		readonly struct FeatureSwitchMembers
+		{
+			public string Feature { get; }
+			public string FeatureValue { get; }
+			public string FeatureDefault { get; }
+			// Unique value to track the key
+			private readonly String _key;
```
mateoatr, line 67: Do we need this field? Why not implement `FeatureSwitchMembers : IEquatable<FeatureSwitchMembers>` and have
```C#
public bool Equals (FeatureSwitchMembers other) => (Feature, FeatureValue, FeatureDefault) == (other.Feature, other.FeatureValue, other.FeatureDefault);

public override bool Equals (object obj) => obj is FeatureSwitchMembers featureSwitchMembers && Equals (featureSwitchMembers);
```

LakshanF: I wanted to avoid the tuple path when I realized HashCode.Combine() was only available in later .NET versions. And reverted to the simplest key implementation

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
LakshanF: If you don't have a strong objection, I prefer to keep these 2 separate for the time being. Should we re-visit feature switch to add member level support, I expect the design to be revamped and revisit this.

Issue #1985, opened 2021-04-21
Analyzer needs to warn on virtual methods


tlakollo:
Currently, the RUC analyzer and RAF analyzer don't check for other virtual calls, for example:
```C#
// On one file
public virtual void VirtualMethod() => null

// On some other file
[RequiresAttributeFiles]
public override void VirtualMethod(){ 
   SomeCode
}

public void M1(){
    VirtualMethod(); //This doesnt warn on the analyzer
}
```
The behavior should be to warn on all implementations of the method

PR #2046, opened 2021-05-21
Match Attributes for virtual methods and overrides

tlakollo:
Add analyzer logic to detect when a method has RUC or RAF but its base method or override method doesn't
Use IL2046 to display errors for RUC
Use IL3003 to display errors for RAF
Add IL3003 to error-codes.md
Add tests
Fixes https://github.com/mono/linker/issues/1985


src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -126,6 +134,28 @@ public override void Initialize (AnalysisContext context)
 						ReportRequiresDiagnostic (operationContext, member, requiresAttribute);
 					}
 				}
+
+				void CheckOverrides (
+					SymbolAnalysisContext symbolAnalysisContext,
+					IMethodSymbol method)
+				{
+					if (!method.HasAttribute (RequiresAttributeName)) {
+						while (method.OverriddenMethod != null) {
```
sbomer, line 143: Since this gets called for each method in the compilation, would it be enough to just check the immediate overridden method?

I think we also will need to look for matching annotations on implemented interface methods.

docs/error-codes.md
```diff
@@ -1636,4 +1636,22 @@ void RequirePublicMethodOnAType(
       // can break functionality when embedded in a single-file app. Use 'MethodFriendlyToSingleFile' instead. http://help/assemblyfiles
       MethodWithAssemblyFilesUsage();
   }
+  ```
+
+#### `IL3003`: Presence of 'RequiresAssemblyFilesAttribute' on method 'method' doesn't match overridden method 'base method'. All overridden methods must have 'RequiresAssemblyFilesAttribute'.
```
mateoatr, line 1641: nit: Going forward, if we add any other `Requires...` attribute, we should re-use this diagnostic in the analyzer, so I think we should make it specific to `RequiresAssemblyFilesAttribute` here.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
mateoatr: Yeah, we should test things like:

```C#

interface IRUC
{
	[RequiresUnreferencedCode("RUC Message")]
	public void RUC();
}

class Imp : IRUC
{
	public void RUC()
	{
	}
}
```

Produce `IL2046` (in the analyzer).

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -126,6 +147,28 @@ public override void Initialize (AnalysisContext context)
...
 				}
+
+				void CheckOverridesAndInterfaces (
+					SymbolAnalysisContext symbolAnalysisContext,
+					ISymbol member)
+				{
+					ISymbol? overriddenMember;
+					ImmutableArray<ISymbol> interfacesImplementations;
+					if ((member.IsVirtual || member.IsOverride) && member.TryGetOverriddenMember (out overriddenMember)) {
+						if (!member.HasAttribute (RequiresAttributeName) && overriddenMember!.HasAttribute (RequiresAttributeName))
```
mateoatr, line 158: I think you could fold these four if-else statements by using xor and deciding in `ReportMatchOverrideOrInterfaceDiagnostic` which of the two symbols is to blame (it will always be the one without the attribute).

```suggestion
						if (member.HasAttribute (RequiresAttributeName) ^ overriddenMember!.HasAttribute (RequiresAttributeName))
```

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
```diff
@@ -19,5 +21,31 @@ internal static bool HasAttribute (this ISymbol symbol, string attributeName)
...
+			if (symbol.Kind != SymbolKind.Method && symbol.Kind != SymbolKind.Property && symbol.Kind != SymbolKind.Event)
+				return false;
+
+			var containingType = symbol.ContainingType;
+			var query = from iface in containingType.AllInterfaces
+						from interfaceMember in iface.GetMembers ()
+						let impl = containingType.FindImplementationForInterfaceMember (interfaceMember)
+						where SymbolEqualityComparer.Default.Equals (symbol, impl)
+						select interfaceMember;
+			interfaces = query.ToImmutableArray ();
```
mateoatr, line 47: We can iterate through the interface members here to save us from Linq :)

agocke: I would probably treat implicit interface implementations separately since they're tricky to access. Instead of looking at each method, then seeing if it's an interface implementation, why not register a handler for every type, and loop over all implicit interface implementations that are in the same type, and check each one there?

You can probably check explicit implementations along with overrides because they're accessed pretty similarly.

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
```diff
@@ -541,9 +541,173 @@ public static void M2()
...
 				// (10,20): warning IL2026: Using method 'C.M1()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Message from --C.M1--.
-				VerifyCS.Diagnostic ().WithSpan (10, 20, 10, 22).WithArguments ("C.M1()", " Message from --C.M1--.", ""),
+				VerifyCS.Diagnostic (RequiresUnreferencedCodeAnalyzer.IL2026).WithSpan (10, 20, 10, 22).WithArguments ("C.M1()", " Message from --C.M1--.", ""),
 				// (11,26): warning IL2026: Using method 'C.M1()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Message from --C.M1--.
-				VerifyCS.Diagnostic ().WithSpan (11, 26, 11, 30).WithArguments ("C.M1()", " Message from --C.M1--.", ""));
+				VerifyCS.Diagnostic (RequiresUnreferencedCodeAnalyzer.IL2026).WithSpan (11, 26, 11, 30).WithArguments ("C.M1()", " Message from --C.M1--.", ""));
+		}
+
+		[Fact]
+		public Task OverrideHasAttributeButBaseDoesnt ()
```
agocke, line 550: Time to move these tests to a base class, I think -- unless you want to add the same ones for `RequiresAssemblyFiles`

agocke:
I'm surprised not to see any tests being unskipped for the linker -- since we should be running the analyzer over the linker tests I assumed we would be seeing those warnings now

mateoatr:
I see you added some of the changes discussed in #2077, could we get these in a separate PR to make it easier to review 😄?

src/ILLink.Shared/SharedStrings.resx
```diff
@@ -141,4 +141,28 @@
   <data name="RequiresAssemblyFilesMessage" xml:space="preserve">
     <value>Using member '{0}' which has 'RequiresAssemblyFilesAttribute' can break functionality when embedded in a single-file app.{1}{2}</value>
   </data>
+  <data name="BaseRequiresMismatchMessage" xml:space="preserve">
+    <value>Base member '{1}' with '{0}' has a derived member '{2}' without '{0}'. Add the '{0}' to '{1}'.</value>
```
agocke, line 145: If we want to keep one error code and just have a bunch of different messages, we could have something like
```suggestion
  <data name="BaseRequiresMismatchMessage" xml:space="preserve">
    <value>{0}</value>
   </data>
```

and then when we create a diagnostic we would manually pull out the messages below and call

```C#
CreateDiagnostic(RequiresMismatchMessage, Location, string.Format(SharedStrings.BaseMismatchRequiresMessage, ..., ..., ...);
```

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -187,6 +239,28 @@ private void ReportRequiresDiagnostic (OperationAnalysisContext operationContext
 				url));
 		}
 
+		private void ReportMismatchInAttributesDiagnostic (SymbolAnalysisContext symbolAnalysisContext, ISymbol member, ISymbol baseMember)
+		{
+			string message = string.Empty;
+			if (!member.HasAttribute (RequiresAttributeName) && baseMember.IsVirtual)
+				message = $"Base member '{baseMember.ToDisplayString ()}' with 'RequiresUnreferencedCodeAttribute' has a derived member '{member.ToDisplayString ()}' without 'RequiresUnreferencedCodeAttribute'";
```
agocke, line 246: These messages should still be in the resources file, they just shouldn't have separate diagnostics associated.

src/linker/Linker.Steps/ValidateVirtualMethodAnnotationsStep.cs
```diff
@@ -40,12 +40,22 @@ protected override void Process ()
...
-				annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (baseMethod))
-				Context.LogWarning (
-					$"Presence of 'RequiresUnreferencedCodeAttribute' on method '{method.GetDisplayName ()}' doesn't match overridden method '{baseMethod.GetDisplayName ()}'. " +
-					$"All overridden methods must have 'RequiresUnreferencedCodeAttribute'.",
-					2046, method, subcategory: MessageSubCategory.TrimAnalysis);
+			bool methodHasAttribute = annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method);
+			if (methodHasAttribute != annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (baseMethod)) {
+				string message = string.Empty;
+				if (!methodHasAttribute && !baseMethod.DeclaringType.IsInterface)
+					message = $"Base member '{baseMethod.GetDisplayName ()}' with 'RequiresUnreferencedCodeAttribute' has a derived member '{method.GetDisplayName ()}' without 'RequiresUnreferencedCodeAttribute'";
```
agocke, line 47: Same here.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -187,6 +239,31 @@ private void ReportRequiresDiagnostic (OperationAnalysisContext operationContext
...
+			if (!member.HasAttribute (RequiresAttributeName) && baseMember.IsVirtual)
+				message = string.Format (SharedStrings.BaseRequiresMismatchMessage, var0, var1, var2);
+			else if (member.HasAttribute (RequiresAttributeName) && baseMember.IsVirtual)
+				message = string.Format (SharedStrings.DerivedRequiresMismatchMessage, var0, var1, var2);
+			else if (!member.HasAttribute (RequiresAttributeName) && !baseMember.IsVirtual)
+				message = string.Format (SharedStrings.InterfaceRequiresMismatchMessage, var0, var1, var2);
+			else if (member.HasAttribute (RequiresAttributeName) && !baseMember.IsVirtual)
+				message = string.Format (SharedStrings.ImplementationRequiresMismatchMessage, var0, var1, var2);
+			if (string.IsNullOrEmpty (message)) {
+				return;
```
agocke, line 257: How could this happen?

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -187,6 +239,31 @@ private void ReportRequiresDiagnostic (OperationAnalysisContext operationContext
...
+			string var2 = baseMember.ToDisplayString ();
+			if (!member.HasAttribute (RequiresAttributeName) && baseMember.IsVirtual)
+				message = string.Format (SharedStrings.BaseRequiresMismatchMessage, var0, var1, var2);
+			else if (member.HasAttribute (RequiresAttributeName) && baseMember.IsVirtual)
+				message = string.Format (SharedStrings.DerivedRequiresMismatchMessage, var0, var1, var2);
+			else if (!member.HasAttribute (RequiresAttributeName) && !baseMember.IsVirtual)
+				message = string.Format (SharedStrings.InterfaceRequiresMismatchMessage, var0, var1, var2);
+			else if (member.HasAttribute (RequiresAttributeName) && !baseMember.IsVirtual)
+				message = string.Format (SharedStrings.ImplementationRequiresMismatchMessage, var0, var1, var2);
+			if (string.IsNullOrEmpty (message)) {
```
agocke, line 256: Instead of checking the value, just use `else`, and don't assign `message` at the beginning of the method, so definite assignment will catch if we forget to set it.

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -54,7 +55,17 @@ public sealed class RequiresAssemblyFilesAnalyzer : RequiresAnalyzerBase
 			isEnabledByDefault: true,
 			helpLinkUri: "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/il3002");
 
-		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create (s_locationRule, s_getFilesRule, s_requiresAssemblyFilesRule);
+		static readonly DiagnosticDescriptor s_RequiresAttributeMismatch = new DiagnosticDescriptor (
```
agocke, line 58:
```suggestion
		static readonly DiagnosticDescriptor s_requiresAttributeMismatch = new DiagnosticDescriptor (
```

src/linker/Linker.Steps/ValidateVirtualMethodAnnotationsStep.cs
```diff
@@ -40,12 +40,24 @@ protected override void Process ()
...
-					$"Presence of 'RequiresUnreferencedCodeAttribute' on method '{method.GetDisplayName ()}' doesn't match overridden method '{baseMethod.GetDisplayName ()}'. " +
-					$"All overridden methods must have 'RequiresUnreferencedCodeAttribute'.",
-					2046, method, subcategory: MessageSubCategory.TrimAnalysis);
+			bool methodHasAttribute = annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method);
+			if (methodHasAttribute != annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (baseMethod)) {
+				string message = string.Empty;
+				string var0 = nameof (RequiresUnreferencedCodeAttribute);
+				string var1 = method.GetDisplayName ();
+				string var2 = baseMethod.GetDisplayName ();
+				if (!methodHasAttribute && !baseMethod.DeclaringType.IsInterface)
```
agocke, line 49: This code looks identical. Let's share it.

agocke (commented):
Looks like the tests still need to move to a base class

src/linker/Linker.Steps/ValidateVirtualMethodAnnotationsStep.cs
tlakollo: SharedStrings is in the limbo between two contexts Mono.Linker and ILlink.RoslynAnalyzer, I had two options to share it:
- Make ILLink.Shared a project
- Create a resource using reflection

I implemented the second only because I feel like the push is to not have an additional project, although I would prefer an additional project over the risk of getting things based on the executing assembly.

tlakollo: > Looks like the tests still need to move to a base class

I didn't create a base class since after #2077 these tests won't need a base class anymore. When I implementing #2077 I will create a base class to share everything left in the RequiresUnreferencedCodeAnalyzerTests and RequiresAssemblyFilesTests

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
tlakollo: By the way the conditionals are constructed this should never happen

src/ILLink.Shared/MessageFormat.cs
```diff
@@ -20,5 +27,19 @@ public static string FormatRequiresAttributeUrlArg (string? url)
...
+			if (!memberHasAttribute && condition)
+				return string.Format (sharedStrings.GetString ("BaseRequiresMismatchMessage")!, var0, var1, var2);
+			else if (memberHasAttribute && condition)
+				return string.Format (sharedStrings.GetString ("DerivedRequiresMismatchMessage")!, var0, var1, var2);
+			else if (!memberHasAttribute && !condition)
+				return string.Format (sharedStrings.GetString ("InterfaceRequiresMismatchMessage")!, var0, var1, var2);
+			else if (memberHasAttribute && !condition)
+				return string.Format (sharedStrings.GetString ("ImplementationRequiresMismatchMessage")!, var0, var1, var2);
+			else
+				return string.Empty;
```
agocke, line 42: if something should never happen, we should throw an exception to indicate that we don't expect this to be a legal state.

src/ILLink.Shared/MessageFormat.cs
```diff
@@ -20,5 +27,19 @@ public static string FormatRequiresAttributeUrlArg (string? url)
...
+		public static string FormatRequiresAttributeMismatch (bool memberHasAttribute, bool condition, string var0, string var1, string var2)
+		{
+			if (!memberHasAttribute && condition)
+				return string.Format (sharedStrings.GetString ("BaseRequiresMismatchMessage")!, var0, var1, var2);
+			else if (memberHasAttribute && condition)
+				return string.Format (sharedStrings.GetString ("DerivedRequiresMismatchMessage")!, var0, var1, var2);
+			else if (!memberHasAttribute && !condition)
+				return string.Format (sharedStrings.GetString ("InterfaceRequiresMismatchMessage")!, var0, var1, var2);
+			else if (memberHasAttribute && !condition)
+				return string.Format (sharedStrings.GetString ("ImplementationRequiresMismatchMessage")!, var0, var1, var2);
```
agocke, line 40:
```suggestion
                        string resourceId = (memberHasAttribute, condition) switch {
                            (false, true) => "BaseRequiresMismatchMessage",
                            (true, true) => "DerivedRequiresMismatchMessage",
                            (false, false) =>  "InterfaceRequiresMismatchMessage",
                            (false, true) =>  "ImplementationRequiresMismatchMessage"
                         };	
			return string.Format(sharedStrings.GetString(resourceId)!, var0, var1, var2));
```

src/ILLink.Shared/MessageFormat.cs
```diff
@@ -20,5 +27,19 @@ public static string FormatRequiresAttributeUrlArg (string? url)
 				arg2 = " " + url;
 			return arg2;
 		}
+
+		public static string FormatRequiresAttributeMismatch (bool memberHasAttribute, bool condition, string var0, string var1, string var2)
```
agocke, line 31: What does "condition" mean? Is there a better name?

src/ILLink.Shared/MessageFormat.cs
```diff
@@ -1,10 +1,17 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Resources;
+
 namespace ILLink.Shared
 {
 	internal static class MessageFormat
 	{
+		static readonly ResourceManager sharedStrings = new ResourceManager (Path.GetFileNameWithoutExtension (Assembly.GetExecutingAssembly ().GetManifestResourceNames ().FirstOrDefault (name => name.Contains ("SharedStrings")))!, Assembly.GetExecutingAssembly ());
```
agocke, line 13: This shouldn't be necessary, why can't you call `SharedStrings` directly? Namespace name differences? We can probably resolve this at compile time without reflection.

tlakollo: So when is in the context of ILLink.RoslynAnalyzers then the resource being used is ILLink.RoslynAnalyzers.SharedStrings but when the context is Mono.Linker the resource being used is Mono.Linker.SharedStrings
I think if there is a middle layer project like ILLink.Shared and then we take ILLink.Shared.SharedStrings everything would be fine. But without an additional project the only other way I found to get the resource was using reflection

agocke: You can fix this by changing the csprojs as follows:

```diff
-    <EmbeddedResource Include="../ILLink.Shared/SharedStrings.resx" GenerateSource="true" />
+    <EmbeddedResource Include="../ILLink.Shared/SharedStrings.resx">
+      <GenerateSource>true</GenerateSource>
+      <Namespace>ILLink.Shared</Namespace>
+    </EmbeddedResource>
```

src/ILLink.Shared/MessageFormat.cs
tlakollo: Changed for "isInterface"

agocke (approved):
LGTM

PR #2047, opened 2021-05-21
Bump versions

marek-safar:


vitek-karas (approved):

eng/Versions.props
```diff
@@ -18,12 +18,12 @@
     <MicrosoftNETCoreILAsmVersion>$(MicrosoftNETSdkILPackageVersion)</MicrosoftNETCoreILAsmVersion>
     <!-- These should match the SDK version at https://github.com/dotnet/sdk/blob/master/eng/Versions.props -->
```
akoeplinger, line 19: should we remove this comment since we don't match with sdk anymore?

marek-safar: They do match (hence this PR) or do you see some which don't?

akoeplinger: MicrosoftBuildFrameworkVersion and MicrosoftBuildUtilitiesCoreVersion are `16.10.0-preview-21112-04` / `17.0.0-preview-21267-01` here but 15.4.8 in the sdk: https://github.com/dotnet/sdk/blob/69fa43ea727df04cc69b0aa1f05d1278b1a5f158/eng/Versions.props#L21-L22

marek-safar: ok, updated the comment

PR #2048, opened 2021-05-22
Add support for UnrecognizedReflectionAccessPattern in analyzer test infra

mateoatr:
This PR adds support for `UnrecognizedReflectionAccessPattern` in the analyzer test infra and refactors the existing code in ILLink.RoslynAnalyzer.Tests.TestCaseUtils.

### DIFF
--- a/test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
@@ -58,7 +58,7 @@ namespace ILLink.RoslynAnalyzer.Tests
 			test.ValidateAttributes (attrs);
 		}
 
-		
+
 
 		public static readonly ImmutableDictionary<string, List<string>> s_testFiles = GetTestFilesByDirName ();
 
@@ -81,7 +81,6 @@ namespace ILLink.RoslynAnalyzer.Tests
 
 		public static void GetDirectoryPaths (out string rootSourceDirectory, out string testAssemblyPath)
 		{
-
 #if DEBUG
 			var configDirectoryName = "Debug";
 #else
--- a/test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
@@ -7,9 +7,11 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
+using System.Text;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
 using Xunit;
 
 namespace ILLink.RoslynAnalyzer.Tests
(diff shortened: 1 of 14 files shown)


test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -34,106 +35,240 @@ public static IEnumerable<object[]> GetTestData (string testSuiteName)
...
+				returnTypeName = semanticModel.GetTypeInfo ((TypeOfExpressionSyntax) returnTypeArgument!.Expression).Type!.GetDisplayName ();
+
+			MemberDeclarationSyntax sourceMember = attributeArguments.Ancestors ().OfType<MemberDeclarationSyntax> ().First ();
+			if (semanticModel.GetDeclaredSymbol (sourceMember) is not ISymbol memberSymbol)
+				return;
+
+			string expectedSourceMemberName = memberSymbol!.GetDisplayName ();
+			// Build the member signature to match.
+			var sb = new StringBuilder ();
+			if (string.IsNullOrEmpty (returnTypeName))
```
tlakollo, line 202: It should be the opposite logic, append the returnTypeName when not nullOrEmpty
```suggestion
			if (!string.IsNullOrEmpty (returnTypeName))
```

tlakollo (commented):
I didn't saw that we are handling the message, there is logic in TryGetNamedArgument to retrieve it but we don't ask for it and we don't check it in the diagnostics. Although I haven't seen that we use it in the tests, and in general by the way it's implemented in the [ResultChecker](https://github.com/mono/linker/blob/main/test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs#L924) makes me think that maybe we will never use it

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -34,106 +35,240 @@ public static IEnumerable<object[]> GetTestData (string testSuiteName)
...
+			var expectedReflectionMemberType = attributeArguments.Arguments[0]!.Expression as TypeOfExpressionSyntax;
+			var expectedReflectionTypeName = semanticModel.GetTypeInfo (expectedReflectionMemberType!).Type!.GetDisplayName ();
+			sb.Append (expectedReflectionTypeName);
+
+			var expectedReflectionMember = semanticModel.GetConstantValue (attributeArguments.Arguments[1]!.Expression);
+			if (expectedReflectionMember.HasValue) {
+				string? expectedReflectionMemberName = expectedReflectionMember.Value as string;
+				sb.Append ("::").Append (expectedReflectionMemberName);
+
+				var parameterTypes = attributeArguments.Arguments[2]!.DescendantNodes ().OfType<ArrayCreationExpressionSyntax> ().FirstOrDefault ();
```
tlakollo, line 214: I wonder how this will behave when Arguments[2] is null, like in this [attribute](https://github.com/mono/linker/blob/main/test/Mono.Linker.Tests.Cases/DataFlow/FieldDataFlow.cs#L62)

mateoatr: > I wonder how this will behave when Arguments[2] is null, like in this attribute.

I think there's no point on passing null here, all reflection callsites should have a name (and we should check for it) -- makes me think that we should probably add a null check in the attribute.

> I see some logic in the TryGetNamedArgument to prevent having null, but then it's not used.

I think I don't understand why this isn't being used?

tlakollo: But Arguments[2] is not the name is the reflectionMethodParameters, that's why the link to the example in the example there is no Arguments[2] so it's null

mateoatr: > But Arguments[2] is not the name is the reflectionMethodParameters

My bad, of course. When this parameter is null we should simply not check (just what the logic in ResultChecker does today). I rebased on top of main to use GetAttributeArguments and make things simpler :)

vitek-karas (approved):

tlakollo (approved):

PR #2051, opened 2021-05-24
Fix a problem where generating an empty FS assembly element roots the whole assembly

vitek-karas:
Descriptors have the behavior that specifying just the assembly element without any children will root the entire assembly (and everything in it). The same applies to types (specifying just the type element will mark the entire type).
With the new feature switches, we have to be careful to not generate an empty assembly element, otherwise we will accidentally root the entire corelib.

Fixes the problem seen in https://github.com/dotnet/runtime/pull/53144.


LakshanF (approved):
Thanks for fixing this!

PR #2053, opened 2021-05-24
Stop duplicating trailing period in IL2026

### LINKED ISSUE #2052: RequiresUnreferencedCode messages have two periods
When getting a warning for using a method annotated with `RequiresUnreferencedCode`, if the message to the `RequiresUnreferencedCode` attribute has a trailing period, the warning message ends up having 2 periods at the end of the sentence. For example:

`C:\git\runtime\src\libraries\System.Text.Json\src\System\Text\Json\Nodes\JsonNode.Dynamic.cs(21,13): Trim analysis warning IL2026: System.Text.Json.Nodes.JsonNode.System.Dynamic.IDynamicMetaObjectProvider.GetMetaObject(Expression): Using method 'System.Text.Json.Nodes.JsonNode.CreateDynamicObject(Expression,JsonNode)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Using JsonNode instances as dynamic types is not compatible with trimming. It can result in non-primitive types being serialized, which may have their members trimmed.. [C:\git\runtime\src\libraries\src.proj]`

Note the last line:

`It can result in non-primitive types being serialized, which may have their members trimmed..`

We should check for a trailing period, and if it is there, not append another period to the message.

mateoatr:
Fixes #2052.


src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -50,7 +50,10 @@ protected override bool IsAnalyzerEnabled (AnalyzerOptions options, Compilation
 		protected override string GetMessageFromAttribute (AttributeData? requiresAttribute)
 		{
 			var message = (string) requiresAttribute!.ConstructorArguments[0].Value!;
-			return message != string.Empty ? " " + message + "." : message;
+			if (!string.IsNullOrEmpty (message))
+				message = $" {message}{(message.TrimEnd ().EndsWith (".") ? "" : ".")}";
```
tlakollo, line 54: We should add the same to RAF

mateoatr: Nice catch. Added.

tlakollo (approved):

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -113,7 +113,10 @@ protected override bool ReportSpecialIncompatibleMembersDiagnostic (OperationAna
 		protected override string GetMessageFromAttribute (AttributeData? requiresAttribute)
 		{
 			var message = requiresAttribute?.NamedArguments.FirstOrDefault (na => na.Key == "Message").Value.Value?.ToString ();
-			return string.IsNullOrEmpty (message) ? "" : $" {message}.";
+			if (!string.IsNullOrEmpty (message))
+				message = $" {message}{(message!.TrimEnd ().EndsWith (".") ? "" : ".")}";
```
agocke, line 117: Should we just stop printing a period all-together? Examining the message string used by the code feels like an anti-pattern

mateoatr: I don't have a personal opinion on this one. I agree that examining the message string feels a bit wrong, so we can stop adding the period.

agocke: That would be my preferred solution. @eerhardt Any objections?

eerhardt: Note that there are a lot of existing warning messages without a trailing period:

https://source.dot.net/#System.Private.CoreLib/RequiresUnreferencedCodeAttribute.cs,6e9970f8b7a80375,references

image

image

If we decide to not append a period, we should have a plan for those.

agocke: My plan: so what? 😄 It doesn't really look bad to me.

eerhardt: As long as we never plan to append more characters after these messages (in any context) then I'm fine with not appending a period and not fixing up the existing messages.

agocke: That seems like a fine design principle to me: the user message is always the last thing in the message string.

tlakollo: What about the cases in which we use a url? The whole process of examining the message and adding spaces or periods was introduced because RAF could have a Message, url, both, or none. And in the case of RUC you still can have the url appended after the message. 
Do we feel ok with
``` c#
[RequiresUnreferencedCode("Types and members the loaded module depends on might be removed", Url="https:/helpurl")]
// warning IL2026: Using member 'C.M1()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when embedded in an app. Types and members the loaded module depends on might be removed https://helpurl
```

tlakollo: On the same line, do we want to keep the resources for RAF and RUC in sync?
Part of the issue that generated #2052 was that now both code messages in the resource file use:
```
Using member '{0}' which has 'RequiresAssemblyFilesAttribute' can break functionality when embedded in a single-file app.{1}{2}
Using method '{0}' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code.{1}{2}
```
This makes the format when creating tests in the analyzer explicitly express spaces and periods consistently between RUC and RAF
```C#
WithArguments ("C.M1()", " Message from attribute.", " https://helpurl")
// Notice the spaces at the beginning of the message and url
```
I think if we keep them in sync we could even end up using one resource title and message for both requires attributes

mateoatr: > Do we feel ok with

I don't see a problem with this -- the user is the one responsible of adding a message/url here.

> I think if we keep them in sync we could even end up using one resource title and message for both requires attributes

I think the title of both messages must remain as is, since one talks about dynamically accessed code while the other talks about single-file incompatible code. For the message, I think we could share it between both attributes, and also change the RUC's message from "`Using method...`" to "`Using member...`" now that the attribute can be used on classes.

agocke: > The whole process of examining the message and adding spaces or periods was introduced because RAF could have a Message, url, both, or none.

Specifically, I think it's fine to examine whether or not a message *exists*, I would just try to avoid analyzing the message contents.

Including the URL afterwards does contradict my earlier statement, but the missing "." between the message and the URL still doesn't bother me too much.

MichalStrehovsky: FWIW, we just append the dot conditionally in NativeAOT - dotnet/runtimelab#701. I forgot to file that here after it was fixed there. The missing dot looks sloppy and I can't really come up with a scenario where "analyzing" the string could do something bad.

vitek-karas: The only case where this might be problematic is if the errors are localized and there's some locale which doesn't use `.` to end a sentence. Localization in general is a good reason to try to not reason about content of any string - in this case not only the string we get from the attribute, but also the string we get from the resources.

What if we rely on the resources string to do the separation. Maybe with newlines, or something similar. Basically use a different mechanism to clearly separate the URL from the message. We can have two different resources, one for the case where there's no URL, and another where there is a URL.

MichalStrehovsky: We're concatenating a string in an unknown language. The language might be written right-to-left and mismatch how illink is localized. The dot is a miniscule problem compared to what this will do to the message.

We consciously scoped glob/loc out using the precedent of ObsoleteAttribute.

agocke: >  The dot is a miniscule problem compared to what this will do to the message.

This is a compelling argument. I'm convinced that shaving of a `.` is an acceptable solution, although I'd still prefer if we avoided looking at the user string.

I also don't really like the problem we're working around: in my opinion the framework should have a uniform style for these messages that's applied everywhere. Either all the messages should end with a period, or none of them should.

eerhardt: FWIW - I'd be totally fine with saying "The linker doesn't append a period to the message. If _you_ want a period in the message, you should have added it." where _you_ is the developer defining the `RequiresUnreferencedCode` attribute. And then we take the stance of "All our framework defined messages should have a period". And for other developers (outside of the framework), they do what they want.

MichalStrehovsky: Basically, I think it's important that messages from our tools don't look sloppy because it might affect how people perceive the tool (sloppy messages -> sloppy tool -> sloppy .NET).

It would be nice if we can solve this for all messages - not just our own. But if we go with fixing just our own, I think it would be really nice if there at least was a Roslyn analyzer for that, if it's not too hard to make one.

agocke: It sounds like we're converging towards the solution in this PR: always add a period, and remove any duplicates. Any objections?

vitek-karas: No objections from me.

mateoatr: > always add a period, and remove any duplicates

So basically the changes implemented in this PR?

The other option is to stop looking into the user-given string (needed to remove duplicates) and handle the responsibility of adding (or not -- depending on the language of the user) a period to the message.

> It would be nice if we can solve this for all messages - not just our own. But if we go with fixing just our own, I think it would be really nice if there at least was a Roslyn analyzer for that, if it's not too hard to make one.

I agree we should have an analyzer do some basic string formatting validation for things that will end up being read by the user.

agocke: Yup, what's in this PR looks good. @mateoatr thanks for hosting this discussion 😄

agocke (approved):

PR #2054, opened 2021-05-25
Introduce scopes into MarkStep to track the location for warnings

vitek-karas:
Marking needs to know where to report a warning if it comes up. So far this has been done by passing around a refence to the member (type, method, ...) typically in a parameter called `sourceLocation`. Going forward this is problematic:
* It's verbose, we're passing this around lot of methods in the MarkStep, without needing to modify it.
* It's hard to enforce rules about it since there's no central location where handling of these things is done.
* With the coming changes for compiler generated code, we will need to pass around more state
* It's technically imprecise, in some cases we can point to exact instruction as the location, but the current mechanism has no way to represent that.

This change introduces a new class `MarkStepScope` which uses a stack to track the active scopes. This state is a member on `MarkStep` and thus available to all methods on it, without the need to pass parameters around. It also creates a single location which can:
* Track more state when needed
* Handle state transitions when needed

This change is largely just a refactoring. There should be almost no functional changes. The exception is a fix for RUC on an attribute in generic parameter. This adds a test case for it as well.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2645,10 +2645,8 @@ namespace Mono.Linker.Steps
 			Annotations.MarkIndirectlyCalledMethod (method);
 		}
 
-		protected virtual MethodDefinition MarkMethod (MethodReference reference, DependencyInfo reason, MessageOrigin? origin = null)
+		protected virtual MethodDefinition MarkMethod (MethodReference reference, DependencyInfo reason)
 		{
-			using var localScope = origin.HasValue ? _scopeStack.PushScope (origin.Value) : null;
-
 			(reference, reason) = GetOriginalMethod (reference, reason);
 
 			if (reference.DeclaringType is ArrayType arrayType) {
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -50,7 +50,7 @@ namespace Mono.Linker.Steps
 		protected Queue<AttributeProviderPair> _assemblyLevelAttributes;
 		readonly List<AttributeProviderPair> _ivt_attributes;
 		protected Queue<(AttributeProviderPair, DependencyInfo, MarkScopeStack.Scope)> _lateMarkedAttributes;
-		protected List<TypeDefinition> _typesWithInterfaces;
+		protected List<(TypeDefinition, MarkScopeStack.Scope)> _typesWithInterfaces;
 		protected HashSet<AssemblyDefinition> _dynamicInterfaceCastableImplementationTypesDiscovered;
 		protected List<TypeDefinition> _dynamicInterfaceCastableImplementationTypes;
 		protected List<(MethodBody, MarkScopeStack.Scope)> _unreachableBodies;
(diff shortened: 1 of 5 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -682,7 +692,7 @@ void ProcessOverride (OverrideInformation overrideInformation)
 			// Only track instantiations if override removal is enabled and the type is instantiated.
 			// If it's disabled, all overrides are kept, so there's no instantiation site to blame.
 			if (_context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, method) && isInstantiated) {
-				MarkMethod (method, new DependencyInfo (DependencyKind.OverrideOnInstantiatedType, method.DeclaringType), null);
+				MarkMethod (method, new DependencyInfo (DependencyKind.OverrideOnInstantiatedType, method.DeclaringType), new MessageOrigin (method.DeclaringType));
```
mateoatr, line 695: Could we use the stack here as well so that we get rid of the last parameter (origin) in MarkMethod?

mateoatr (approved):
Looks pretty good!

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Not really - there's nothing on the stack when this is called - this is because virtual override processing is delayed and done outside of normal marking (from `ProcessVirtualMethods`). We could setup something onto the stack, but the most natural thing would the method itself. It's unclear if that's better scope that the type. Since virtual methods are largely marked because of their type (it's instantiated and so on).

vitek-karas: And I immediately prove myself wrong.
Turns out we have to propagate the scope from the `ProcessMethod` to these delayed processing queues, otherwise the upcoming compiler generated code handling would not work (since it needs a parent scope at all times).

So you were right.

sbomer:
Nice! I like the use of IDisposable.

I never fully understood the differences between the source locations and the dependency info. Is the difference mainly that dependency info captures more granular info about type specifications? For example for a call to Foo<Bar>.Baz(), the dependency for Bar is Caller -> Foo<Bar>.Baz() -> Foo<Bar> -> Bar, whereas for the source location I think we just pass through the instruction location and use that as the source for Bar.

The two feel like they are trying to accomplish very similar things - just tracking the context that led us to mark some piece of IL. In some cases we need to know the relevant source location, and in other cases (dependency tracing) we might want more granular info, but the "context" should be the same. This looks like it opens the possibility to combine twe two.

I have mixed feelings about making the current context implicit in the MarkStep methods. I do like that it removes an argument from many of the Mark* methods, but it would be easier to reason about if there were no way for a callee to mess up the caller's context by mutating global state. The invariant checking in the scope stack will help, but I would personally rather have the context passed in as an argument (ideally combined with dependency info).

vitek-karas:
The way I think about it is that there are two cases:
* Method which doesn't care about the current scope and doesn't modify it any way (nor it should). Passing this around in such a method is pure noise
* Method which cares about the current scope
  * It reads it - in this case, passing it in is no different from reading it from global state. The method has to trust the passed in (or global) value anyway.
  * It modifies it - in this case it doesn't matter what the previous value is as the method is going to set a new one. (This will not be 100% true with compiler generated code handling, but it's close enough).

So the only problem could happen if somebody calls PushScope, but doesn't dispose the returned value. That's a clear bug. 
I don't feel this is that risky and it simplifies passing stuff around.

I agree that it might make sense to somehow combine the scope and dependency info tracking. And it might change my view on passing the scope around as parameter. I don't know yet. I'm trying to scope the work on this a little, otherwise the compiler generated code change will never happen (it's already rather complex unfortunately).

sbomer (approved):
Makes sense. I think combining the two would be a stronger reason to use a separate parameter in the future. The implementation LGTM.
> I'm trying to scope the work on this a little

I see what you did there ;)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2570,33 +2624,35 @@ protected void MarkStaticFields (TypeDefinition type, in DependencyInfo reason)
...
+			MarkMethod (method, reason);
 			Annotations.MarkIndirectlyCalledMethod (method);
 		}
 
-		protected virtual MethodDefinition MarkMethod (MethodReference reference, DependencyInfo reason, MessageOrigin? origin)
+		protected virtual MethodDefinition MarkMethod (MethodReference reference, DependencyInfo reason, MessageOrigin? origin = null)
 		{
-			(reference, reason) = GetOriginalMethod (reference, reason, origin?.MemberDefinition);
+			using var localScope = origin.HasValue ? _scopeStack.PushScope (origin.Value) : null;
+
```
mateoatr, line 2651:
```suggestion
		protected virtual MethodDefinition MarkMethod (MethodReference reference, DependencyInfo reason)
		{
```

PR #2057, opened 2021-05-26
Fix validation of ExpectedWarning attribute

vitek-karas:
Probably due to copy/paste error we've validated the attribute by looking at all warning messages at once. This means that potentially any warning can be matched to the attribute even if its location doesn't match.

This change fixes that but uncovers some small test issues:
* The expected warnings happen in different places - fixed that
* Some warnings can't be reported by the Roslyn Analyzer as they require global view - added a mechanism to mark these

Refactored a bit of the Roslyn analyzer validation code to support named properties in attributes.

### DIFF
--- a/test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
+++ b/test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
@@ -732,7 +732,7 @@ namespace Mono.Linker.Tests.TestCasesRunner
 										if (mc.Origin?.MemberDefinition?.FullName == attrProvider.FullName)
 											return true;
 
-										if (loggedMessages.Any (m => m.Text.Contains (attrProvider.FullName)))
+										if (mc.Text.Contains (attrProvider.FullName))
 											return true;
 
 										return false;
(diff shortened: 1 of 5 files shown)


test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -729,10 +729,16 @@ void VerifyLoggedMessages (AssemblyDefinition original, LinkerTestLogger logger,
 										if (sourceColumn != null && mc.Origin?.SourceColumn != sourceColumn.Value)
 											return false;
 									} else {
+										if (attrProvider.FullName.Contains ("PropertyWithUnsupportedType"))
+											System.Diagnostics.Debug.WriteLine ("");
+
```
mateoatr, line 734: Do we need this? Looks like it was left here accidentally.

mateoatr (approved):

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
vitek-karas: Story of me debugging the linker :-(

Issue #2059, opened 2021-05-27
`All` annotation does not preserve members on base type if the type is from "copy" assembly


vitek-karas:
Repro:

```C#
// Base.dll - "link" action
class Base { private void Method() {} }

// Test.dll - "copy" action
class TestType : Base {}

void Test() {
    RequireAll(typeof(TestType);
}

void RequireAll([DynamicallyAccessedMembers.All] Type t) {}
```

This should preserve `Base.Method` since `All` should apply to base classes as well. But currently this doesn't work. If the `Test.dll` is "link" action, it will work.

PR #2060, opened 2021-05-27
Fix applying All annotation on a type from "copy" assembly

vitek-karas:
"Copy" assembly calls `MarkEntireType` on all types in the assembly, but it specified to not include base and interface types. We cache if a type was marked as a whole, so to not repeat the process (and also to avoid potential recursion). But the cache doesn't differentiate if the type was mark with its base/interface types or not.

So in the "copy" assembly case, the type is first marked without base/interface marking, and that is cached. Later on we apply the All annotation to the type, which calls `MarkEntireType` again, but since it's already cached it doesn't do anything. But All annotation should preserve base/interface types as well, which will not happen in this case.

Fixed this by making the cache aware of the base/interface flag and repeat the marking if base/interface is requested even if the type has already been marked as a whole but without base/interface marking.

Added a test case.

Fixes https://github.com/mono/linker/issues/2059


mateoatr (approved):
LGTM.

PR #2063, opened 2021-05-27
Revert "Remove custom marking of nested types for event sources (#2032)"

LakshanF:
We still cannot take out the special handling for EventSource due to [test failures ](https://dev.azure.com/dnceng/public/_build/results?buildId=1159451&view=ms.vss-test-web.build-test-results-tab&runId=34991568&paneView=debug&resultId=163140)in runtime repo. Will reopen #1949

This reverts commit 2cf308289ea6140bbee332318e3ac45c1a6d0342.


sbomer (approved):
I don't have all of the context on the EventSource plan, but the revert LGTM.

Issue #2065, opened 2021-05-28
Data flow annotation on unsupported return type doesn't warn


vitek-karas:
```C#
[return: DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicMethods)]
static object UnsupportedReturnType () => null;
```

This should produce a warning. For example the same but on method parameter does produce a warning:
```C#
// IL2098
private void UnsupportedParameterType ([DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicFields)] object p1)
```

PR #2066, opened 2021-05-28
Warn when method return type is annotated but it's not Type or String

vitek-karas:
Also added a small fix in the test infra

Fixes https://github.com/mono/linker/issues/2065

### DIFF
--- a/test/Mono.Linker.Tests.Cases/DataFlow/MethodReturnParameterDataFlow.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/MethodReturnParameterDataFlow.cs
@@ -38,6 +38,8 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 			instance.PropagateReturnToReturn (0);
 
 			instance.ReturnWithRequirementsAlwaysThrows ();
+
+			UnsupportedReturnType ();
 		}
 
 		static Type NoRequirements ()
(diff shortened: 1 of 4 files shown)


test/Mono.Linker.Tests.Cases/DataFlow/MethodReturnParameterDataFlow.cs
```diff
@@ -189,6 +191,10 @@ Type ReturnWithRequirementsAlwaysThrows ()
 			throw new NotImplementedException ();
 		}
 
+		[ExpectedWarning ("IL2106", nameof (UnsupportedReturnType))]
```
mateoatr, line 194: This and #2048 make me wonder how much do we gain from keeping both `ExpectedWarning` and `URAP` -- I get that `URAP` is more granular and only produced in very specific scenarios, but can't we just add a mechanism for keeping the origin of the produced warnings when testing and then just use ExpectedWarning` everywhere? Furthermore, I think this could allow us get rid of `ReflectionPatternContext` (letting us simplify some of the code in ReflectionMethodBodyScanner and the test infra).
Thinking out loud here, since there might be something I'm completely ignoring.

mateoatr (approved):

test/Mono.Linker.Tests.Cases/DataFlow/MethodReturnParameterDataFlow.cs
vitek-karas: You're not wrong at all. The `IReflectionPatternRecorder` should be removed. I added it a long time ago when the DataFlow work started, but the design has changed since. Its original purpose was to separate the analysis from reporting. We thought that reporting the warnings would be a complex piece of logic (bubbling warnings up the virtual call stacks and so on).

Currently it's basically useless - the only implementation is the logger, which doesn't need that complexity, and the test infra - which can migrate to `ExpectedWarning` just like you describe.

PR #2070, opened 2021-05-31
Use full MessageOrigin when reporting data flow warnings

vitek-karas:
Recent chaneg introduced a scope stack to marking which carries `MessageOrigin` around so that warnings can be reported with precise origin. The one exception to this was the `ReflectionMethodBodyScanner`.

This change integrates the scope stack with `ReflectionMethodBodyScanner`. The visible result is more precise reporting of some warnings. This change will be needed for the work around compiler generated code (since then MessageOrigin will carry even more information than it does currently).

The main API change is addition of the `MessageOrigin` parameter to the `IReflectionPatternRecorder.UnrecognizedReflectionAccessPattern`. There are plans to remove this interface completely, so there' not much need to clean it up right now (so I left the potentially duplicated `origin` and `source` arguments as is).

The rest of the change is just correctly flowing the information around.

I also improved the test infra to be able to correctly handle `ExpectedWarning` with source/line info on all warnings.

There are visible changes in the `SourceLines` test where some warnings changed location (more precise now).

### DIFF
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
@@ -14,6 +14,7 @@ namespace Mono.Linker.Dataflow
 	{
 		readonly LinkContext _context;
 		readonly MarkStep _markStep;
+		readonly MarkScopeStack _scopeStack;
 
 		// Cache of DynamicallyAccessedMembers annotations applied to types and their hierarchies
 		// Values
@@ -38,10 +39,11 @@ namespace Mono.Linker.Dataflow
 		// of a type which is currently being marked - at which point the interfaces are not yet marked.
 		readonly Dictionary<TypeDefinition, (DynamicallyAccessedMemberTypes annotation, bool applied)> _typesInDynamicallyAccessedMembersHierarchy;
 
-		public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep markStep)
+		public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep markStep, MarkScopeStack scopeStack)
 		{
 			_context = context;
 			_markStep = markStep;
+			_scopeStack = scopeStack;
 			_typesInDynamicallyAccessedMembersHierarchy = new Dictionary<TypeDefinition, (DynamicallyAccessedMemberTypes, bool)> ();
 		}
 
@@ -106,8 +108,9 @@ namespace Mono.Linker.Dataflow
 			if (apply) {
 				// One of the base/interface types is already marked as having the annotation applied
 				// so we need to apply the annotation to this type as well
-				var reflectionMethodBodyScanner = new ReflectionMethodBodyScanner (_context, _markStep);
-				var reflectionPatternContext = new ReflectionPatternContext (_context, true, type, type);
+				using var _ = _scopeStack.PushScope (new MessageOrigin (type));
+				var reflectionMethodBodyScanner = new ReflectionMethodBodyScanner (_context, _markStep, _scopeStack);
+				var reflectionPatternContext = new ReflectionPatternContext (_context, true, _scopeStack.CurrentScope.Origin, type);
 				reflectionMethodBodyScanner.ApplyDynamicallyAccessedMembersToType (ref reflectionPatternContext, type, annotation);
 				reflectionPatternContext.Dispose ();
 			}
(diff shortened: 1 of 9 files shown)


src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -106,8 +108,9 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
 			if (apply) {
 				// One of the base/interface types is already marked as having the annotation applied
 				// so we need to apply the annotation to this type as well
-				var reflectionMethodBodyScanner = new ReflectionMethodBodyScanner (_context, _markStep);
-				var reflectionPatternContext = new ReflectionPatternContext (_context, true, type, type);
+				using var _ = _scopeStack.PushScope (new MessageOrigin (type));
+				var reflectionMethodBodyScanner = new ReflectionMethodBodyScanner (_context, _markStep, _scopeStack);
+				var reflectionPatternContext = new ReflectionPatternContext (_context, true, _scopeStack.CurrentScope.Origin, type);
```
marek-safar, line 113: Would it make sense to add overload with ReflectionPatternContext parameter?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -63,10 +64,11 @@ bool ShouldEnableReflectionPatternReporting (MethodDefinition method)
 			return !_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method);
 		}
 
-		public ReflectionMethodBodyScanner (LinkContext context, MarkStep parent)
+		public ReflectionMethodBodyScanner (LinkContext context, MarkStep parent, MarkScopeStack scopeStack)
```
marek-safar, line 67: nit: should be `in`

src/linker/Linker.Dataflow/ReflectionPatternContext.cs
```diff
@@ -25,21 +25,22 @@ struct ReflectionPatternContext : IDisposable
 		bool _patternReported;
 #endif
 
-		public IMemberDefinition Source { get; private set; }
+		public MessageOrigin Origin { get; private set; }
```
marek-safar, line 28: nit: better would be `public MessageOrigin Origin { get; init; }`

marek-safar (approved):

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
vitek-karas: Sorry - I don't follow. Overload of which function call here?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: `MarkScopeStack` is a `class` - does `in` do anything interesting on a ref type?

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
marek-safar: Of the constructors or make static factory methods. Something like this 

for types
```
reflectionPatternContext = ReflectionPatternContext.ForType (_context, _scopeStack, type);
```
for methods
```
reflectionPatternContext = ReflectionPatternContext.ForMethod (_context, _scopeStack, method);
```

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
marek-safar: My bad for some reason I though it's a struct

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
vitek-karas: Definitely possible, but I don't see the benefits. There's nothing different between methods/types/fields in this case. The only advantage would be to potentially restrict what can be passed in a bit more (not just random "member", but "method/type/field" only).

marek-safar: Fair enough, it was just a suggestion

Issue #2058, opened 2021-05-26
Add support for abstract static interface members


agocke:
The linker shouldn't trim the interface definition when trimming in library mode, and we likely need to support the same implementation checks that we do for other interface members.

The minimum bar for .NET 6 is to ensure that runtime libraries are trim safe with this feature. The following check list tracks the tasks that are needed for the full support of this feature

- [x] Trimmer support for the IL changes. Specifically, the trimmer matches the runtime [changes](https://github.com/dotnet/runtime/pull/52173).

> - [x] The runtime tests can be found [here](https://github.com/dotnet/runtime/tree/main/src/tests/Loader/classloader/StaticVirtualMethods) which contains tests in IL for `TypeHierarchy`, `GenericContext`, `InterfaceVariance`, and negative tests.

- [x] Add C# tests to the trimmer that tests this feature comprehensively (will require an updated compiler with the static interface support enabled)
- [x] Reflection support, specifically `MakeGenericType `and `MakeGenericMethod`.
- [x] Trimmer annotations are correctly propagated.
  - [x] Validate DAM on type implementing static interface, or on interface type itself keeps interface members
  - [x] Validate DAM on generic in interface method signature, or on interface type gets applied
  - [x] Validate matching of DAM/RUC on interface method and implementing overrides
- ~~[ ] https://github.com/dotnet/linker/issues/2733 - potential improvement as this could reduce size~~

agocke:
@LakshanF Has anything been finished here?

LakshanF:
The IL support

agocke:
+ @jtschuster who was looking at some of these problems. Looks like runtime is still hitting issues -- we should probably try to run the linker against https://github.com/dotnet/runtime/pull/68612 directly and see if we can find and solve the problem

jtschuster:
I've confirmed that the linker removes all unused `INumber<>` interfaces in a hello world application.
cc @vitek-karas

PR #2072, opened 2021-06-01
Initial work to support virtual static interface methods

LakshanF:
Partial work for static interface support as tracked by issue #2058. 

The initial PR is a draft one and is WIP with the the primary goal of getting feedback. This PR supports the [runtime minimum test case](https://github.com/dotnet/runtime/blob/main/src/tests/Loader/classloader/StaticVirtualMethods/TypeHierarchy/TypeHierarchyTest_minimal.il) for type hierarchy. Issue #2058 tracks the full support tasks but the minimum bar work for trimmer should include the following.

All static overrides are done exclusively via [`MethodImpl `](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#ii1032-the-override-directive)records. (which translates to the .override directives). The linker has good existing support for this that helps with most of the runtime tests that are used to test this feature (see below)

- [Done] Full runtime support as tracked by the runtime tests added: 

> - [Done] [This](https://github.com/dotnet/runtime/blob/main/src/tests/Loader/classloader/StaticVirtualMethods/TypeHierarchy/TypeHierarchyTest.il) runtime test tracks support for 3 configurations. The 3 follow the below patterns (Interface with static method, Base Class and a derived class that extends the base class, Derived class is called with the interface method). Contains 276 Scenarios

> - [Done] Generic context: 1536 scenarios with generic, non-generic, value type and class combinations

> - [Done] Variance: 10 scenarios with interface variance

- [ ] Tests in C# with the new compiler (adding tests to the repo might need to wait until the compiler gets checked in)


src/linker/Linker/TypeMapInfo.cs
```diff
@@ -299,9 +300,14 @@ IEnumerable<InterfaceImplementation> GetDefaultInterfaceImplementations (TypeDef
 
 		MethodDefinition TryMatchMethod (TypeReference type, MethodReference method)
 		{
+			var isStaticInterfaceMethod = (method as MethodDefinition)?.IsStatic;
 			foreach (var candidate in type.GetMethods (context)) {
 				var md = context.TryResolve (candidate);
-				if (md?.IsVirtual != true)
+				//Static interface methods - they are not virtual
+				bool isMethodStaticInterface = false;
```
vitek-karas, line 307: The naming is confusing - what's the difference between `isStaticInterfaceMethod` and `isMethodStaticInterface`? They both read the same to me.

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/DefaultInterfaceMethods/Dependencies/StaticInterfaceImple.il
```diff
@@ -0,0 +1,153 @@
...
+  } // end of method MethodImplOnDerived
+} // end of class DerivedScenario3
+
+.class public auto ansi TestEntrypoint
+       extends [mscorlib]System.Object
+{
+  .method public static void Test_Scenario1() cil managed noinlining
+  {
+    constrained. DerivedScenario1
+    call int32 InterfaceScenario1::Method()
```
vitek-karas, line 118: Just want to double check that the linker works as I think it does:
* This should be the only reference to `DerivedScenario1` the linker sees 
  * Assumes that we correctly process `constrained. ` instructions - can you double check what happens for these?
  * This ends up marking `DerivedScenario1` without anything in it
* The second instruction is treated as normal call which will mark both `InterfaceScenario1` and also `Method` on it
* Since `Method` is evaluated as a base method in a virtual hierarchy (sort of), we end up marking it on `DerivedScenario1` as well

Right?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -671,9 +671,11 @@ void ProcessOverride (OverrideInformation overrideInformation)
 
 			var isInstantiated = Annotations.IsInstantiated (method.DeclaringType);
 
+			var isStaticInterfaceOverride = @base.IsStatic && method.IsStatic;
```
vitek-karas, line 674: I think it's worth a comment what's going on here.
What path is a static interface method expected to take through this method - the method is already pretty complex, and it seems we rely on relatively fragile things for the static interface method to make it through and return `true`,

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -671,9 +671,11 @@ void ProcessOverride (OverrideInformation overrideInformation)
...
+
 			// We don't need to mark overrides until it is possible that the type could be instantiated
 			// Note : The base type is interface check should be removed once we have base type sweeping
-			if (IsInterfaceOverrideThatDoesNotNeedMarked (overrideInformation, isInstantiated))
+			if (!isStaticInterfaceOverride && IsInterfaceOverrideThatDoesNotNeedMarked (overrideInformation, isInstantiated))
 				return;
 
 			if (!isInstantiated && !@base.IsAbstract && _context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, method))
```
vitek-karas, line 681: (Can't comment on code below this line directly):
Naming consideration: Currently all of this code uses the term 'virtual method' - which is correct for normal virtual methods, it's a slight stretch for interface methods (but they do act like virtual methods for the most part), but now it's a rather larger stretch for static interface methods - those are definitely no virtual in the typical sense of the word.

Not for this checking, but I think we should consider changing the name to something more descriptive... (although not sure what it should be). At least we need to have a good comment somewhere describing what the logic covers and how it's supposed to work high-level.

Honestly it might be wort a doc in the linker repo. Virtual method and interface marking is one of the most complex behaviors in the linker.

src/linker/Linker/TypeMapInfo.cs
```diff
@@ -299,9 +301,20 @@ IEnumerable<InterfaceImplementation> GetDefaultInterfaceImplementations (TypeDef
 
 		MethodDefinition TryMatchMethod (TypeReference type, MethodReference method)
 		{
+			var isStaticInterfaceMethod = context.TryResolve (method)?.IsStatic;
+			var typeDefintion = context.TryResolve (type);
+			if (isStaticInterfaceMethod == true && !context.Annotations.IsFromStaticInterface (typeDefintion)) {
+				context.Annotations.ImplementsStaticInterface (typeDefintion);
```
vitek-karas, line 307: If we need to cache this, it should be done in `AnnotateMethods` - not in `TryMatchMethod` - ideally `Try` methods should have no side-effects.

Also technically - `TryMatchMethod` is called for both virtual and interface hierarchies. I think we need this new logic only in the interface hierarchy processing.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -558,7 +558,7 @@ void ProcessMarkedTypesWithInterfaces ()
 			foreach (var type in typesWithInterfaces) {
 				// Exception, types that have not been flagged as instantiated yet.  These types may not need their interfaces even if the
 				// interface type is marked
-				if (!Annotations.IsInstantiated (type) && !Annotations.IsRelevantToVariantCasting (type))
+				if (!Annotations.IsFromStaticInterface (type) && !Annotations.IsInstantiated (type) && !Annotations.IsRelevantToVariantCasting (type))
```
vitek-karas, line 561: I assume this is a perf optimization, right?

src/linker/Linker/Annotations.cs
```diff
@@ -244,6 +245,17 @@ public bool IsInstantiated (TypeDefinition type)
 			return marked_instantiated.Contains (type);
 		}
 
+		public void ImplementsStaticInterface (TypeDefinition type)
+		{
+			implements_static_interface_methods.Add (type);
+		}
+
+		public bool IsFromStaticInterface (TypeDefinition type)
```
MichalStrehovsky, line 253: When I first saw this I didn't know what to expect from a method with this name.

Consider naming these `void MarkImplementsStaticInterfaceMethods(TypeDefinition)` and `bool ImplementsStaticInterfaceMethod(TypeDefinition)`.

But I'm not even sure we really need them.

src/linker/Linker/TypeMapInfo.cs
```diff
@@ -166,7 +167,8 @@ void MapVirtualMethods (TypeDefinition type)
 				return;
 
 			foreach (MethodDefinition method in type.Methods) {
-				if (!method.IsVirtual)
+				// This could be a static interface method implementation
+				if (!method.IsStatic && !method.IsVirtual)
```
MichalStrehovsky, line 171:
```suggestion
				if (method.IsStatic || !method.IsVirtual)
```

I find this one easier to read because there's less negation. Also, it can be evaluated faster.

src/linker/Linker/TypeMapInfo.cs
```diff
@@ -299,9 +301,20 @@ IEnumerable<InterfaceImplementation> GetDefaultInterfaceImplementations (TypeDef
 
 		MethodDefinition TryMatchMethod (TypeReference type, MethodReference method)
 		{
+			var isStaticInterfaceMethod = context.TryResolve (method)?.IsStatic;
```
MichalStrehovsky, line 304: I wouldn't expect us to run TryMatchMethod for static interface methods. Those are not implemented by a name/sig match and this search would return bogus results:

```csharp
interface IFoo
{
    static virtual void Blah();
}

class Foo : IFoo
{
    // This method is unrelated to IFoo.Blah
    public static void Blah() => throw null;

    // This one implements it
    static void IFoo.Blah() => throw null;
}
```

We should have an early out for static methods before this path is reached.

src/linker/Linker.Steps/MarkStep.cs
LakshanF: Ended up not using this path as you suggested and relying on MapVirtualMethods()'s MapOverrides (after by passing the short circuit for the static method check). All static overrides are done exclusively via [MethodImpl ](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#ii1032-the-override-directive)records. (Which translates to .override directives)

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/StaticInterfaceMethods/Dependencies/TypeHierarchyBasic.il
```diff
@@ -3,45 +3,17 @@
...
-
 .class interface public abstract auto ansi InterfaceScenario1
 {
   .method public newslot virtual abstract static int32 Method() cil managed noinlining
   {
   } // end of method Method
 } // end of class InterfaceScenario1
 .class public auto ansi BaseScenario1
-       extends [System.Runtime]System.Object
+       extends [mscorlib]System.Object
```
vitek-karas, line 16: Why are we targeting mscorlib? These tests are never expected to work on anything but .NET 6+, so there's no reason to use mscorlib.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -558,7 +558,7 @@ void ProcessMarkedTypesWithInterfaces ()
 			foreach (var type in typesWithInterfaces) {
 				// Exception, types that have not been flagged as instantiated yet.  These types may not need their interfaces even if the
 				// interface type is marked
-				if (!Annotations.IsFromStaticInterface (type) && !Annotations.IsInstantiated (type) && !Annotations.IsRelevantToVariantCasting (type))
+				if (!Annotations.IsInstantiated (type) && !Annotations.IsRelevantToVariantCasting (type))
```
vitek-karas: I would expect this to break the functionality. Static interface method types should not be instantiated (hopefully the tests don't do that) and vast majority of them won't be relevant to variant casting either. So I would expect these to hit the 'continue' here and "escape".
Meaning we'd be left with a type which doesn't implement the static interface after trimming. The tests probably don't validate this yet.

MichalStrehovsky: Isn't this path used for lazy marking of stuff that `IsInterfaceOverrideThatDoesNotNeedMarked` allowed to be lazy? Since we reject static methods being done lazily, we shouldn't need to do anything for static methods here.

src/linker/Linker/TypeMapInfo.cs
```diff
@@ -166,10 +167,12 @@ void MapVirtualMethods (TypeDefinition type)
 				return;
 
 			foreach (MethodDefinition method in type.Methods) {
-				if (!method.IsVirtual)
+				// This could be a static interface method implementation that has a .override
+				if (!(method.IsStatic || method.IsVirtual))
```
MichalStrehovsky, line 171: I see, we want to remember the overrides but not to base mapping.

I would still do something along the lines of this.

```csharp
if (!method.IsVirtual)
    continue;

if (!method.IsStatic)
    MapVirtualMethod (method);

if (method.HasOverrides)
    MapOverrides (method);
```

Maybe it's just me, but it took me a while to realize when the `(!(method.IsStatic || method.IsVirtual))` condition is **not taken** (since this is just a continue, I'm more interested in when this evaluates to false).

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/StaticInterfaceMethods/Dependencies/TypeHierarchyBasic.il
tlakollo: When referencing System.Runtime the test would generate errors stating that the assembly System.Runtime couldn't be found. I guess in part is inexperience with the IL testing in linker. Not sure if there is something else that needs to be made in the testing infra to be able to reference System.Runtime instead of mscorlib

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Maybe I don't understand this correctly. I think the difference is marking methods versus marking types and interface implementations.

`ProcessOverride` (called via `ProcessVirtualMethod`) will handle the methods, but it doesn't really mark types (only indirectly). `ProcessMarkedTypesWithInterfaces` on the other hand deals only with types and more importantly with `InterfaceImplementation`. The effect is probably very subtle and won't be visible in most tests, but for example `MarkInterfaceImplementation` will mark the attributes on the interface impl, which would not happen for static interfaces if I read it correctly.

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/StaticInterfaceMethods/Dependencies/TypeHierarchyBasic.il
vitek-karas: That's because there's `.assembly extern mscorlib { }` above. If you dump simple .NET Core app it will have `.assembly extern System.Runtime { }` instead.

In IL you have to declare assembly references, and only the declared ones will be "usable".

tlakollo: That line existed in the test file and was the one generating a roslyn error

vitek-karas: hmm - Roslyn error? OK - I would have to repro this I guess, but the current state is not great. (I assume the tests in runtime repo are not targeting mscorlib, right?)

tlakollo: I experimented a little bit more with it
If you only reference `.assembly extern System.Runtime { }` and then use it in the test, you will get the following failure:
```
System.AggregateException : One or more errors occurred. (Roslyn compilation errors: (20,5): error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Runtime, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
    )
      ----> System.Exception : Roslyn compilation errors: (20,5): error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'System.Runtime, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
```
But if you put a version and publickeytoken
```
.assembly extern System.Runtime
{
  // I took those values from a dotnet app that I build
  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A)
  .ver 6:0:0:0
}
```
Then the test will pass

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/DefaultInterfaceMethods/Dependencies/StaticInterfaceImple.il
LakshanF: Yes!

- The `.constrained` instruction ends up marking `DerivedScenario1 `as you indicated above (as well as explore base types).
- The next instruction, `call`, results in marking the `InterfaceScenario1 `and `Method` in it as you expected.
- The actual method is `overriden `in BaseScenario1. Since we marked this method earlier in `MapOverrides` at `TypeMapInfo`, I assume everything works correctly since the full type hierarchy (`InterfaceScenario1`, `BaseScenario1`, `DerivedScenario1` and related methods) have been correctly marked. 
 
The post trim code looks correct.

LakshanF:
The linker changes were minimal and was primarily to by pass short circuits that we had for static methods. Some highlights

- Linker already has support for the runtime [implementation ](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#ii1032-the-override-directive). Needed the short circuits to be by passed
- Inspection of the code as a result of the short circuit by pass seems we do the right thing
- Ported the 1822 IL tests from the runtime to the linker test bed
- Did a quick test run with coreclr and library tests on the runtime repo with the linker changes to check for regressions and found none

I believe this work is sufficient to unblock the runtime library failures

tannergooding:
I was able to validate this locally and things look to be working as expected (for Compiler + Runtime + Libraries changes)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -711,6 +711,10 @@ bool IsInterfaceOverrideThatDoesNotNeedMarked (OverrideInformation overrideInfor
 			if (!overrideInformation.IsOverrideOfInterfaceMember || isInstantiated)
 				return false;
 
+			// This is a static interface method
```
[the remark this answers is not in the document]
sbomer: I think we can now get to `ProcessVirtualMethod` for the overriding non-virtual static method (at the end of `ProcessOverride`). Could you add a check there, with an explanation of why it's needed, something like this?
```csharp
// The override may be non-virtual if it is an override of a static virtual interface method
if (method.IsVirtual)
  ProcessVirtualMethod (method);
```

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -711,6 +711,10 @@ bool IsInterfaceOverrideThatDoesNotNeedMarked (OverrideInformation overrideInfor
...
+			if (overrideInformation.Override.IsStatic && overrideInformation.Base.IsStatic)
```
sbomer, line 715: I would personally make this check as specific as possible - I'd check that the base is abstract and the override is non-virtual (I believe that's always the case unless I misread the spec).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2770,6 +2774,7 @@ protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo
 				Debug.Assert (Annotations.IsMarked (method));
 				break;
 			default:
+
```
sbomer, line 2777: nit: remove extra whitespace

src/linker/Linker/TypeMapInfo.cs
```diff
@@ -118,6 +118,7 @@ protected virtual void MapType (TypeDefinition type)
 
 		void MapInterfaceMethodsInTypeHierarchy (TypeDefinition type)
 		{
+
```
sbomer, line 121: nit: remove extra whitespace

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/StaticInterfaceMethods/TypeHierarchyBasic.cs
```diff
@@ -0,0 +1,34 @@
...
+	[KeptTypeInAssembly ("library.dll", "InterfaceScenario1")]
+	[KeptMemberInAssembly ("library.dll", "BaseScenario1", "Method()")]
+	[KeptTypeInAssembly ("library.dll", "DerivedScenario1")]
+	[KeptTypeInAssembly ("library.dll", "InterfaceScenario2")]
+	[KeptMemberInAssembly ("library.dll", "BaseScenario2", "Method()")]
+	[KeptTypeInAssembly ("library.dll", "DerivedScenario2")]
+	[KeptTypeInAssembly ("library.dll", "InterfaceScenario3")]
+	[KeptTypeInAssembly ("library.dll", "BaseScenario3")]
+	[KeptMemberInAssembly ("library.dll", "DerivedScenario3", "MethodImplOnDerived()")]
+
```
sbomer, line 24: Could you also add a few `KeptInterfaceOnTypeInAssembly` checks to test that the interfaceimpls are kept?

sbomer (approved):
A few minor comments, otherwise LGTM. Thank you!

Issue #2073, opened 2021-06-01
Provide a way for custom steps to resolve assemblies


sbomer:
Xamarin-macios uses reflection to get a linker method that loads assemblies:

https://github.com/xamarin/xamarin-macios/blob/882dde132a0b893c1e929969d7472e93ec7745c8/tools/dotnet-linker/Steps/CollectAssembliesStep.cs#L19

We should provide some way to load new assemblies so that reflection isn't required here. I think rather than making `GetReferencedAssemblies` public we should give custom steps more granular control over individual assembly resolution. Maybe we could make AssemblyResolver public, or implement `IAssemblyResolver` on `LinkContext`. The latter would nicely match what we did with `IMetadataResolver`.

PR #2074, opened 2021-06-02
Expose Resolve helper for assemblies

sbomer:
Fixes https://github.com/mono/linker/issues/2073

### DIFF
--- a/src/linker/ref/Linker/LinkContext.cs
+++ b/src/linker/ref/Linker/LinkContext.cs
@@ -2,8 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Mono.Cecil;
 using System;
+using Mono.Cecil;
 
 namespace Mono.Linker
 {
--- a/src/linker/ref/Linker/LinkContext.cs
+++ b/src/linker/ref/Linker/LinkContext.cs
@@ -3,11 +3,12 @@
 // See the LICENSE file in the project root for more information.
 
 using Mono.Cecil;
+using System;
 
 namespace Mono.Linker
 {
 
-	public class LinkContext : IMetadataResolver
+	public class LinkContext : IMetadataResolver, IAssemblyResolver, IDisposable
 	{
 		internal LinkContext () { }
 		public AnnotationStore Annotations { get { throw null; } }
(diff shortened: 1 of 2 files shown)


vitek-karas (approved):
:shipit:

src/linker/ref/Linker/LinkContext.cs
```diff
@@ -2,12 +2,13 @@
...
 // See the LICENSE file in the project root for more information.
 
+using System;
 using Mono.Cecil;
 
 namespace Mono.Linker
 {
 
-	public class LinkContext : IMetadataResolver
+	public class LinkContext : IMetadataResolver, IAssemblyResolver, IDisposable
```
marek-safar, line 11: What is the scenario where LinkContext would be used as IAssemblyResolver ?

I don't think we want to expose IDisposable, if custom step calls Dispose bad things will most likely happen

marek-safar (changes requested):

src/linker/ref/Linker/LinkContext.cs
sbomer: It would be used here to resolve references up-front: https://github.com/xamarin/xamarin-macios/blob/882dde132a0b893c1e929969d7472e93ec7745c8/tools/dotnet-linker/Steps/CollectAssembliesStep.cs#L19

They need to do this for https://github.com/xamarin/xamarin-macios/issues/11449 and https://github.com/xamarin/xamarin-macios/issues/11447 (the IsDirectBinding part).

IDisposable is required by IAssemblyResolver, so if we want to avoid it we'd need to expose a Resolve helper without implementing IAssemblyResolver.

marek-safar: Sorry, I don't follow how is this related to LinkContext implementing IAssemblyResolver. Presumably, you plan to use it somewhere but what will will look like?

sbomer: What's needed is just some way to load references up-front, for optimizations that rely on global information. The issue is that there's no supported way for a custom step to tell the linker to load assembly references (all references, or even an individual reference) using the built-in assembly resolver.

The way it will be used is in a custom step that just recursively loads references for each already-loaded assembly. A subsequent step will scan for global information across all loaded assemblies.

Implementing IAssemblyResolver on LinkContext is just one way to make that possible as mentioned in https://github.com/mono/linker/issues/2073. Another solution would be to make GetReferencedAssemblies public, or to add a Resolve helper without implementing IAssemblyResolver.

sbomer: I changed it to add a Resolve helper without implementing IAssemblyResolver per discussion.

marek-safar (approved):

PR #2075, opened 2021-06-02
Warning suppressions in compiler generated code

vitek-karas:
This change implements the following high-level functionality:
* `UnconditionalSuppressMessage` applies to the entire method body even for iterators, async methods
* `RequiresUnreferencedCode` automatically suppresses trim analysis warnings from entire method body for iterators, async methods

Solution approach:
* Detect compiler generated code by using the `IteratorStateMachineAttribute`, `AsyncStateMachineAttribute` and `AyncIteratorStateMachineAttribute`.
  * When a method which is compiler generated (detected by looing for `<` in its name) is processed the original "user method" is determined by looking for a method with the "state machine" attribute pointing to the compiler generated type.
  * This information is cached to avoid duplication of relatively expensive detection logic.
* Looks for `UnconditionalSuppressMessage` and `RequriesUnreferencedCode` in:
  * If the warning origin is not a compiler generated code - simply use the warning origin (method) - existing behavior
  * If the warning origin is compiler generated use both the origin as well as the user defined method (non-compiler-generated) which is the source of the code for the compiler generated origin

This is done by storing additional `SuppressionContextMember` on `MessageOrigin` which should always point to non-compiler-generated item.

Added lot of new tests for these scenarios.

This implements warning suppression part of https://github.com/mono/linker/issues/2001 for state machines.

### DIFF
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -1613,7 +1613,7 @@ class Test
 
   ```C#
   class <compiler_generated_state_machine>_type {
-      void MoveNext ()
+      void MoveNext()
       {
           // This should normally produce IL2026
           CallSomethingWhichRequiresUnreferencedCode ();
@@ -1622,7 +1622,7 @@ class Test
 
   [RequiresUnreferencedCode ("")] // This should suppress all warnings from the method
   [IteratorStateMachine(typeof(<compiler_generated_state_machine>_type))]
-  IEnumerable<int> UserDefinedMethod ()
+  IEnumerable<int> UserDefinedMethod()
   {
       // Uses the state machine type
       // The IL2026 from the state machine should be suppressed in this case
@@ -1630,12 +1630,11 @@ class Test
 
   // IL2107: Methods 'UserDefinedMethod' and 'SecondUserDefinedMethod' are both associated with state machine type '<compiler_generated_state_machine>_type'.
   [IteratorStateMachine(typeof(<compiler_generated_state_machine>_type))]
-  IEnumerable<int> SecondUserDefinedMethod ()
+  IEnumerable<int> SecondUserDefinedMethod()
   {
       // Uses the state machine type
       // The IL2026 from the state should be reported in this case
   }
-
   ```
 
 ## Single-File Warning Codes
--- a/docs/error-codes.md
+++ b/docs/error-codes.md
@@ -1561,36 +1561,36 @@ class Test
 
 #### `IL2103`: Trim analysis: Value passed to the 'propertyAccessor' parameter of method 'System.Linq.Expressions.Expression.Property(Expression, MethodInfo)' cannot be statically determined as a property accessor
 
-The value passed to the `propertyAccessor` parameter of `Expression.Property(expression, propertyAccessor)` was not recognized as a property accessor method. Trimmer can't guarantee the presence of the property.
+- The value passed to the `propertyAccessor` parameter of `Expression.Property(expression, propertyAccessor)` was not recognized as a property accessor method. Trimmer can't guarantee the presence of the property.
 
-```C#
-void TestMethod(MethodInfo methodInfo)
-{
-  // IL2103: Value passed to the 'propertyAccessor' parameter of method 'System.Linq.Expressions.Expression.Property(Expression, MethodInfo)' cannot be statically determined as a property accessor.
-  Expression.Property(null, methodInfo);
-}
-```
+  ```C#
+  void TestMethod(MethodInfo methodInfo)
+  {
+    // IL2103: Value passed to the 'propertyAccessor' parameter of method 'System.Linq.Expressions.Expression.Property(Expression, MethodInfo)' cannot be statically determined as a property accessor.
+    Expression.Property(null, methodInfo);
+  }
+  ```
 
 #### `IL2104`: Assembly 'assembly' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries
 
-The assembly 'assembly' produced trim analysis warnings in the context of the app. This means the assembly has not been fully annotated for trimming. Consider contacting the library author to request they add trim annotations to the library. To see detailed warnings for this assembly, turn off grouped warnings by passing either `--singlewarn-` to show detailed warnings for all assemblies, or `--singlewarn- "assembly"` to show detailed warnings for that assembly. https://aka.ms/dotnet-illink/libraries has more information on annotating libraries for trimming.
+- The assembly 'assembly' produced trim analysis warnings in the context of the app. This means the assembly has not been fully annotated for trimming. Consider contacting the library author to request they add trim annotations to the library. To see detailed warnings for this assembly, turn off grouped warnings by passing either `--singlewarn-` to show detailed warnings for all assemblies, or `--singlewarn- "assembly"` to show detailed warnings for that assembly. https://aka.ms/dotnet-illink/libraries has more information on annotating libraries for trimming.
 
 #### `IL2105`: Type 'type' was not found in the caller assembly nor in the base library. Type name strings used for dynamically accessing a type should be assembly qualified.
 
-Type name strings representing dynamically accessed types must be assembly qualified, otherwise linker will first search for the type name in the caller's assembly and then in System.Private.CoreLib.
-If the linker fails to resolve the type name, null will be returned.
+- Type name strings representing dynamically accessed types must be assembly qualified, otherwise linker will first search for the type name in the caller's assembly and then in System.Private.CoreLib.
+  If the linker fails to resolve the type name, null will be returned.
 
-```C#
-void TestInvalidTypeName()
-{
-    RequirePublicMethodOnAType("Foo.Unqualified.TypeName");
-}
-void RequirePublicMethodOnAType(
-    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
-    string typeName)
-{
-}
-```
+  ```C#
+  void TestInvalidTypeName()
+  {
+      RequirePublicMethodOnAType("Foo.Unqualified.TypeName");
+  }
+  void RequirePublicMethodOnAType(
+      [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
+      string typeName)
+  {
+  }
+  ```
 
 #### `IL2106`: Trim analysis: Return type of method 'method' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to properties of type 'System.Type' or 'System.String'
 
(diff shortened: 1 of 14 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -59,9 +59,23 @@ public static bool RequiresReflectionMethodBodyScannerForAccess (LinkContext con
...
+			// since that attribute automatically suppresses all trim analysis warnings.
+			// Check both the immediate origin method as well as suppression context method
+			// since that will be different for compiler generated code.
+			IMemberDefinition suppressionContextMember = _scopeStack.CurrentScope.Origin.SuppressionContextMember;
+			if (suppressionContextMember is MethodDefinition suppressionContextMethod &&
+				_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (suppressionContextMethod))
+				return false;
+
+			IMemberDefinition originMember = _scopeStack.CurrentScope.Origin.MemberDefinition;
+			if (suppressionContextMember != originMember && originMember is MethodDefinition originMethod &&
```
sbomer, line 74: Why do we need to check both? I would have thought that just checking SuppressionContextMember would be enough.

src/linker/Linker/MessageOrigin.cs
```diff
@@ -95,8 +105,15 @@ public override string ToString ()
 		public int CompareTo (MessageOrigin other)
 		{
 			if (MemberDefinition != null && other.MemberDefinition != null) {
-				return (MemberDefinition.DeclaringType?.Module?.Assembly?.Name?.Name, MemberDefinition.DeclaringType?.Name, MemberDefinition?.Name).CompareTo
+				int result = (MemberDefinition.DeclaringType?.Module?.Assembly?.Name?.Name, MemberDefinition.DeclaringType?.Name, MemberDefinition?.Name).CompareTo
```
sbomer, line 108: nit: does this need a `?` in every dereference? If the declaring type was non-null I think the rest is ok.

sbomer: Also, does this need a case for when the MemberDefinition is a type?

src/linker/Linker/MessageOrigin.cs
```diff
@@ -95,8 +105,15 @@ public override string ToString ()
...
 					((other.MemberDefinition.DeclaringType?.Module?.Assembly?.Name?.Name, other.MemberDefinition.DeclaringType?.Name, other.MemberDefinition?.Name));
+				if (result != 0)
+					return result;
+
+				if (ILOffset != null && other.ILOffset != null)
+					return ILOffset.Value.CompareTo (other.ILOffset);
+
+				return ILOffset == null ? 1 : -1;
```
sbomer, line 116: Shouldn't this return 0 when both offsets are null?

sbomer (commented):
Very nice!

src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -0,0 +1,66 @@
...
+			switch (GetAssumedLanguage (memberReference)) {
+			case Language.CSharp:
+				if (memberReference.Name.Contains ('<'))
+					return true;
+				break;
+
+			case Language.FSharp:
+				if (memberReference.Name.Contains ('@'))
+					return true;
+				break;
```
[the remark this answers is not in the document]
vitek-karas: I'll double check, but my understanding is that the compiler will use the same naming scheme for VB and C#...

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas:
```C#
[UnconditionalSuppressMessage("IL2026")]
void Test()
{
    LocalFunction ();
   
    [UnconditionalSuppressMessage("IL2071")]
    void LocalFunction() {
        // Do something which produces 2026 and also 2071
    }
}
```

In this case the `LocalFunction` is a compiler generated code (the actual name is something long and weird). But it does have a user provided attribute on it - so we need to look there. But at the same time, the attribute applied to the outer method still needs to apply to the body of `LocalFunction` as well (this is how Roslyn implements `UnconditionalSuppressMessage`)

MichalStrehovsky:
Does this work if I nest local functions?

```csharp
Level1();

static void Level1()
{
    Level2();

    static void Level2()
    {
        Level3();

        static void Level3()
        {

        }
    }
}
```

I general, I would kind of prefer we do something more conservative, like look for AsyncStateMachineAttribute. I know that wouldn't solve location functions. But it would make the compiler feature request more urgent. If we add a workaround, however miserable it is, it is easy to brush off the compiler feature request with "they're unblocked and shipped this way".

If I understand this correctly - we figure out the "owning user code" by looking at the code that triggered need for the method body we're looking at - I'm a bit worried about whether a situation can occur where a single generated method is referred to from multiple user method bodies - would the annotation propagate from user code we saw first only?

I'm not confident I know all the situations where C# generates these helpers to be able to say these assumptions are safe. And I know nothing about the F# compiler.

vitek-karas:
The intent is definitely for it to work in nested local functions - I just haven't written all the tests yet.

Your other points are all valid:
* Use more conservative approach - I actually wouldn't mind too much relying on the attributes. But @agocke didn't feel that these are OK to consume outside of the compiler/debugger. Note that it probably would make the compiler support more urgent because we would have a solution for iterators and async (via the attributes) and local functions and lambdas have an easy workaround - add the attribute on the local function - it would not match roslyn/analyzer behavior but it's a workaround just like any other.
* You are right that this sort of makes the assumption that the compiler generated code is only used from one "user method". If this is not true, we would apply the "first callsite only" rules.
* I discussed the C# one with @agocke and we're pretty confident that it should work for C#. F# is different and honestly it's more of a best effort (which is not much) currently since trimming F# is something which is already broken in many places anyway.

MichalStrehovsky:
Will this work if the reason why we're marking the compiler-generated thing is reflection/DynamicDependency/ILLinkTrim?

I'm thinking of scenario where we include an entire type with all its compiler-generated members because it was subject to something like `DAMT.All` - the reason why we are including it would be whatever brought `.All` on the type - not the method actually calling into compiler-generated code.

agocke:
Looking for `Async` or `IteratorStateMachine` is fine, but I thought about the attributes-to-local-functions problems -- I don't see how it could work for those cases because there's no type to point to, and the method name might be ambiguous for overloads.

agocke:
Ah, but if we don't need to identify the caller and only need to know that the method was generated by the compiler, that's different...

We would need a new attribute though, because the `CompilerGenerated` attribute is used by debuggers for a totally different purpose.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
sbomer: Thanks!

src/linker/Linker/MessageOrigin.cs
vitek-karas: I'm kind of split on this - I agree that it should not happen, but if it does it will hard crash the linker - with no really good reason (sorting warnings is not a critical feature anyway). I'd rather leave it as is for now.

vitek-karas: I fixed the type case (good catch)

src/linker/Linker/MessageOrigin.cs
vitek-karas: Fixed

vitek-karas:
I think this is now ready for final review.

The actually implemented solution is not based on heuristics as that lead to lot of trouble when code is accessed via reflection (in which case we can't rely on mark ordering to determine the "user method").

This now only works on async and iterator method bodies. Lambdas and local functions are not handled.
The mapping is done via the compiler generated attributes `IteratorStateMachine` and similar. The new implementation is completely independent from mark ordering and works even when the compiler generated code is referenced only via reflection.

docs/error-codes.md
```diff
@@ -1604,6 +1604,39 @@ void RequirePublicMethodOnAType(
...
+  Only a meta-sample:
+
+  ```C#
+  class <compiler_generated_state_machine>_type {
+      void MoveNext ()
+      {
+          // This should normally produce IL2026
+          CallSomethingWhichRequiresUnreferencedCode ();
+      }
+  }
```
MichalStrehovsky, line 1621: Since this is public facing documentation, we should format this the way normal people expect C# code to be formatted.

MichalStrehovsky (approved):

docs/error-codes.md
vitek-karas: argh...

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2716,12 +2716,20 @@ void ProcessRequiresUnreferencedCode (MethodDefinition method, DependencyKind de
...
+			// Check both the origin's member (the direct method which contains the call to RUC)
+			// as well as the suppression context member (the "user code" source in case of compiler generated code)
+			IMemberDefinition suppressionContextMember = currentOrigin.SuppressionContextMember;
+			if (suppressionContextMember != null &&
+				Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (suppressionContextMember))
+				return;
+
+			IMemberDefinition originMember = currentOrigin.MemberDefinition;
+			if (suppressionContextMember != originMember && originMember != null &&
+				Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (originMember))
```
sbomer, line 2732: Can these two checks be extracted into a helper and shared with https://github.com/mono/linker/pull/2075/files#diff-565c4d3a1fda40fe49e36a001479e217a6a484722f6e97ee132c3ddcefdbd585R64?

src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -0,0 +1,94 @@
...
+
+		public CompilerGeneratedState (LinkContext context)
+		{
+			_context = context;
+			_compilerGeneratedTypeToUserCodeMethod = new Dictionary<TypeDefinition, MethodDefinition> ();
+			_typesWithPopulatedCache = new HashSet<TypeDefinition> ();
+		}
+
+		static bool HasRoslynCompilerGeneratedName (IMemberDefinition member) =>
+			member.Name.Contains ('<') || member.DeclaringType == null || HasRoslynCompilerGeneratedName (member.DeclaringType);
```
sbomer, line 25: This can be changed to take a TypeDefinition I think. Also, I'm not sure I understand the DeclaringType logic - should it be this instead?

```suggestion
			member.Name.Contains ('<') || (member.DeclaringType != null && HasRoslynCompilerGeneratedName (member.DeclaringType));
```

sbomer (approved):

src/linker/Linker/CompilerGeneratedState.cs
vitek-karas: You're right on both accounts - fixed.

PR #2076, opened 2021-06-02
Handle different dynamic member access for multiple branches

kant2002:
This case assume that all code branches should have same minimal dynamic access pattern

Was discovered as part of https://github.com/dotnet/runtimelab/issues/1187

Issue happens when 2 code blocks has diferent requirements for dynamic code access, and this produce MergePoint. This case was not handled.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1777,27 +1777,17 @@ namespace Mono.Linker.Dataflow
 			}
 
 			// Validate that the return value has the correct annotations as per the method return value annotations
-			if (returnValueDynamicallyAccessedMemberTypes != 0 && methodReturnValue != null) {
-				if (methodReturnValue is LeafValueWithDynamicallyAccessedMemberNode methodReturnValueWithMemberTypes) {
-					if (!methodReturnValueWithMemberTypes.DynamicallyAccessedMemberTypes.HasFlag (returnValueDynamicallyAccessedMemberTypes))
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1784,6 +1784,18 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 				} else if (methodReturnValue is SystemTypeValue) {
 					// SystemTypeValue can fullfill any requirement, so it's always valid
 					// The requirements will be applied at the point where it's consumed (passed as a method parameter, set as field value, returned from the method)
+				} else if (methodReturnValue is MergePointValue mergePointValue) {
```
MichalStrehovsky, line 1787: I think we can unify this with the if checks above:

Starting a line 1780:

```csharp
// Validate that the return value has the correct annotations as per the method return value annotations
if (returnValueDynamicallyAccessedMemberTypes != 0) {
    foreach (var uniqueValue in methodReturnValue.UniqueValues () {
      // put lines 1781 - 1789 of the original file here, but replace methodReturnValue with uniqueValue
    }
}
```

That way we don't need to duplicate the checks. Calling UniqueValues on null or on somehting that is not mergepoint is fine.

vitek-karas (approved):
It's unfortunate that the test currently doesn't validate anything because linker repo doesn't use framework which has the return value annotation yet. But I think this should work correctly.

Thanks a lot for the fix!

vitek-karas:
Can you please run `lint.cmd` or `lint.sh` from the root of the repo and commit/push any changes it makes?

kant2002:
@vitek-karas done!

Issue #2004, opened 2021-04-29
Opaque warnings for unsupported features without managed callers


sbomer:
By design, turning on startup hook support produces a warning. However, the warning shown by default is not helpful.

~~First, it is collapsed:~~ (fixed by https://github.com/mono/linker/pull/2087)
```
dotnet publish -r linux-x64 -p:PublishTrimmed=true -p:StartupHookSupport=true
```
```
...System.Private.CoreLib.dll : warning IL2104: Assembly 'System.Private.CoreLib' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries
```

~~If you uncollapse it,~~ the warning mentions implementation details of the startup hook:
```
dotnet publish -r linux-x64 -p:PublishTrimmed=true -p:StartupHookSupport=true -p:TrimmerSingleWarn=false
```
```
ILLink : Trim analysis warning IL2026: System.StartupHookProvider.ProcessStartupHooks(): Using method 'System.StartupHookProvider.CallStartupHook(StartupHookProvider.StartupHookNameOrPath)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The StartupHookSupport feature switch has been enabled for this app which is being trimmed. Startup hook code is not observable by the trimmer and so required assemblies, types and members may be removed.
```

Only the last part is useful to the app developer.  Note that there is a dependency from the XML descriptor -> ProcessStartupHook -> [RUC] CallStartupHook, and the warning is produced for the call to CallStartupHook.

I believe we need a way to mark some APIs such that they:
- produce warnings that "blame" the feature switch settings, even if they are never called by any other methods (only rooted due to the feature setting)
- do not look like they are warning about implementation details of corelib
- ~~don't get collapsed by default~~

For startup hook, I think it should only say "The StartupHookSupport feature switch has been enabled for this app which is being trimmed. Startup hook code is not observable by the trimmer and so required assemblies, types and members may be removed."

Startup hooks happen to not have public APIs, but the native hosting APIs will have a similar problem, and they do have public UnmanagedCallersOnly APIs. The current approach will show warnings like "public API calls internal method with RequiresUnreferencedCode" - it would be better not to mention the internal APIs.

@LakshanF

vitek-karas:
[...]

PR #2087, opened 2021-06-09
Always show IL2026 for IsTrimmable assemblies

sbomer:
This lets RUC warnings intentionally left in IsTrimmable-attributed assemblies survive collapsing, so that the user-readable message is shown.

This is a workaround for the lack of a more well-defined model for "global" warnings (such as RUC methods behind a disabled feature switch, or RUC code that's part of an override which _may_ be called virtually). In the framework libraries we took the approach of moving `RUC` to helper methods that always have a direct callsite in managed code, and we gave the attribute a message indicating the global problem. These warnings will no longer be collapsed, but they still have a message and origin mentioning a call from non-RUC to RUC inside of the library. 

Addresses part of https://github.com/mono/linker/issues/2004.

### DIFF
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -72,6 +72,7 @@ namespace Mono.Linker
 		readonly CustomAttributeSource _customAttributes;
 		readonly List<MessageContainer> _cachedWarningMessageContainers;
 		readonly ILogger _logger;
+		readonly Dictionary<AssemblyDefinition, bool> _isTrimmable;
 
 		public Pipeline Pipeline {
 			get { return _pipeline; }
@@ -210,6 +211,7 @@ namespace Mono.Linker
 			_parameters = new Dictionary<string, string> (StringComparer.Ordinal);
 			_customAttributes = new CustomAttributeSource (this);
 			_cachedWarningMessageContainers = new List<MessageContainer> ();
+			_isTrimmable = new Dictionary<AssemblyDefinition, bool> ();
 			FeatureSettings = new Dictionary<string, bool> (StringComparer.Ordinal);
 
 			SymbolReaderProvider = new DefaultSymbolReaderProvider (false);
@@ -424,12 +426,15 @@ namespace Mono.Linker
 			}
 		}
 
-		bool IsTrimmable (AssemblyDefinition assembly)
+		public bool IsTrimmable (AssemblyDefinition assembly)
 		{
-			if (!assembly.HasCustomAttributes)
-				return false;
+			if (_isTrimmable.TryGetValue (assembly, out bool isTrimmable))
+				return isTrimmable;
 
-			bool isTrimmable = false;
+			if (!assembly.HasCustomAttributes) {
+				_isTrimmable.Add (assembly, false);
+				return false;
+			}
 
 			foreach (var ca in assembly.CustomAttributes) {
 				if (!ca.AttributeType.IsTypeOf<AssemblyMetadataAttribute> ())
@@ -450,6 +455,7 @@ namespace Mono.Linker
 				isTrimmable = true;
 			}
 
+			_isTrimmable.Add (assembly, isTrimmable);
 			return isTrimmable;
 		}
 
--- a/src/linker/Linker/MessageContainer.cs
+++ b/src/linker/Linker/MessageContainer.cs
@@ -132,22 +132,41 @@ namespace Mono.Linker
 			if (version > context.WarnVersion)
 				return Empty;
 
-			if (subcategory == MessageSubCategory.TrimAnalysis) {
-				Debug.Assert (origin.MemberDefinition != null);
-				var declaringType = origin.MemberDefinition?.DeclaringType ?? (origin.MemberDefinition as TypeDefinition);
-				var assembly = declaringType.Module.Assembly;
-				var assemblyName = assembly?.Name.Name;
-				if (assemblyName != null && context.IsSingleWarn (assemblyName)) {
-					if (context.AssembliesWithGeneratedSingleWarning.Add (assemblyName))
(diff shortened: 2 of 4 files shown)


vitek-karas (approved):
Looks good but please try to run the linker on an app with for example enabled startup hook, to see if it actually works the way we think it should.

I assume the followup is to remove the detailed message origin and maybe only report an assembly, or maybe no origin if collapsing is turned on for the affected assembly. We will also probably have to change the warning message in some way.

mateoatr (approved):
LGTM. Should we use an out `MessageContainer` param in `TryLogSingleWarning` instead of going through `LogWarning` wheel a second time?

sbomer:
> I assume the followup is to remove the detailed message origin and maybe only report an assembly, or maybe no origin if collapsing is turned on for the affected assembly.

I wasn't planning on doing this - the warning is still an IL2026 which has a documented message format, and that message doesn't make much sense without an origin - unless we wanted to give it a different warning code? Then I'm not sure what would be a good message/origin. The linker doesn't really have any more context except that the RUC method was called.

Only printing the RUC user message without an origin might be a bit mysterious as to where the warning is coming from. This feels like it would be better handled by specialized warning codes for different circumstances along the lines we've been discussing.

sbomer:
> Should we use an out MessageContainer param in TryLogSingleWarning instead of going through LogWarning a second time?

I think I'd want it to go through LogWarning again because it's kind of a separate warning. Any of the usual warning filtering should be applied - for example if there were a way to suppress assembly-level warnings, it should be possible to suppress IL2104.

sbomer:
> please try to run the linker on an app with for example enabled startup hook, to see if it actually works the way we think it should.

before:
```
sven@localhost StartupHook]$ dotnet publish -r linux-x64 -p:PublishTrimmed=true -p:StartupHookSupport=true
Microsoft (R) Build Engine version 17.0.0-preview-21304-01+a5ea6d2ca for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored /home/sven/tmp/StartupHook/StartupHook.csproj (in 209 ms).
  You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
  StartupHook -> /home/sven/tmp/StartupHook/bin/Debug/net6.0/linux-x64/StartupHook.dll
/home/sven/.nuget/packages/microsoft.netcore.app.runtime.linux-x64/6.0.0-preview.6.21306.1/runtimes/linux-x64/lib/net6.0/System.Private.CoreLib.dll : warning IL2104: Assembly 'System.Private.CoreLib' produced trim warnings. For more information see https://aka.ms/dotnet-illink/libraries [/home/sven/tmp/StartupHook/StartupHook.csproj]
  Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
  StartupHook -> /home/sven/tmp/StartupHook/bin/Debug/net6.0/linux-x64/publish/
```
after:
```
[sven@localhost StartupHook]$ dotnet publish -r linux-x64 -p:PublishTrimmed=true -p:StartupHookSupport=true -p:ILLinkTasksAssembly=/home/sven/linker/artifacts/bin/ILLink.Tasks/Debug/net5.0/ILLink.Tasks.dll
Microsoft (R) Build Engine version 17.0.0-preview-21304-01+a5ea6d2ca for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored /home/sven/tmp/StartupHook/StartupHook.csproj (in 222 ms).
  You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
  StartupHook -> /home/sven/tmp/StartupHook/bin/Debug/net6.0/linux-x64/StartupHook.dll
ILLink : Trim analysis warning IL2026: System.StartupHookProvider.ProcessStartupHooks(): Using method 'System.StartupHookProvider.CallStartupHook(StartupHookProvider.StartupHookNameOrPath)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The StartupHookSupport feature switch has been enabled for this app which is being trimmed. Startup hook code is not observable by the trimmer and so required assemblies, types and members may be removed. [/home/sven/tmp/StartupHook/StartupHook.csproj]
  Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
  StartupHook -> /home/sven/tmp/StartupHook/bin/Debug/net6.0/linux-x64/publish/
```

eerhardt:
> cc @eerhardt in case you have opinions on this.

I don't think I have hugely strong opinions on it. On one hand I do like it says 

> `System.StartupHookProvider.ProcessStartupHooks(): Using method 'System.StartupHookProvider.CallStartupHook(StartupHookProvider.StartupHookNameOrPath)....`,

so a savvy developer could go looking into our code for more information about this (comments in the code, git blame, linked issues and PRs, etc).

I almost think the only real piece of that message that doesn't provide value in this case is `which has 'RequiresUnreferencedCodeAttribute' `. Beyond those 3 words, I think the message reads pretty well.

marek-safar (approved):

vitek-karas:
Definitely a much better experience. And I would be perfectly OK shipping it this way.

One day I think we should improve on this:
* For "normal" developer the warning source is "magic" - either nothing at all, or just CoreLib would probably be better
* The "technical" part might be interesting but only for people who are willing to dig into it, which is probably not very common.

Ideally I think the warning should be basically:
```console
ILLink : Trim analysis warning IL2026: The StartupHookSupport feature switch has been enabled for this app which is being trimmed. Startup hook code is not observable by the trimmer and so required assemblies, types and members may be removed. http://dot.net/illink/startuphook.
```

That is for cases where we collapse warnings. If I disable warning collapse, I should get the warning in the current form with the detailed technical description and true origin.

Issue #2090, opened 2021-06-13
CallSite.ReturnType.MetadataType sometimes gives incorrect results


LakshanF:
`MethodBodyScanner `has a check for IL Opcode calli for [void](https://github.com/mono/linker/blob/caeaf2a3fb3f636805fdd4881df4f9a539fff8f6/src/linker/Linker.Dataflow/MethodBodyScanner.cs#L535) and` ReturnType.MetadataType` gives incorrect results. Specifically, for a call on `System.DateTime::UpdateLeapSecondCacheAndReturnUtcNow()` returns `OptionalModifier`. `signature.ReturnType.GetElementType ().MetadataType` return the correct `MetadataType.Void`

Relevant Code is below
[C#](https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/libraries/System.Private.CoreLib/src/System/DateTime.Windows.cs#L189):    `s_pfnGetSystemTimeAsFileTime(&fileTimeNow);`

IL: `calli      unmanaged cdecl void modopt(System.Runtime.CompilerServices.CallConvSuppressGCTransition)(uint64*)`

EXPECT: `MetadataType.Void`

MichalStrehovsky:
Oh, Cecil.

We probably have this problem wherever `.MetadataType` is used on something that we didn't call `Resolve` on because there can be a modifier. C# started putting them on a lot of things lately.

LakshanF:
This behavior seems to be by design

Issue #1989, opened 2021-04-22
Linker warning IL2050 doesn't get suppressed by adding RequiresUnreferencedCode


joperezr:
While resolving the linker warnings for Microsoft.VisualBasic.Core I found an instance of warning IL2050 here:

https://github.com/dotnet/runtime/blob/21dcc7385ea41beffe19913f2c6b765728d5ee48/src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/Helpers/UnsafeNativeMethods.vb#L31-L36

The Linker shows the following message:

```
Microsoft.VisualBasic.CompilerServices.UnsafeNativeMethods.VariantChangeType(Object&,Object&,Int16,Int16): P/invoke method 'Microsoft.VisualBasic.CompilerServices.UnsafeNativeMethods.VariantChangeType(Object&,Object&,Int16,Int16)' declares a parameter with COM marshalling. Correctness of COM interop cannot be guaranteed after trimming. Interfaces and interface members might be removed.
```

While the warning makes sense, I was surprised that this warning wouldn't be automatically suppressed by adding RequiresUnreferencedCode attribute on that method. In my mind, when resolving Linker warnings there are pretty much 3 different actions to take:
 - Rewrite the statement to make it safe so that the linker won't warn.
 - Manually Suppress the error with UnconditionallySuppressMessage (maybe along with adding DynamicDependency when appropriate) if the callsite is deemed safe and the linker is warning as a false positive.
 - Add RequiresUnreferencedCode attribute so that you can bubble up the warning to your callers.

For IL2050, there doesn't seem to be a way of doing that 3rd action, which there should be one in my opinion (either via RequiresUnreferencedCode or a new attribute).

cc: @MichalStrehovsky @eerhardt @LakshanF @vitek-karas @sbomer

MichalStrehovsky:
I have mixed feeling about having RequiresUnreferencedCode suppressing COM warnings. Conceptually, it's similar, but hitting COM codepaths after trimming leads to a lot more sadness than regular RequiresUnreferencedCode (the failures modes might be a lot more inscrutable - including wrong methods being called with wrong calling conventions). It's to the point that we should consider IL2050 to be an error.

I wonder if we can somehow connect this to https://github.com/dotnet/runtime/issues/50500 so that if the app is trimmed with default settings, COM would be disabled and VariantChangeType never reached (therefore never generating a warning). If COM would be opted in manually, it's fine to warn from an implementation detail.

[...]

PR #2091, opened 2021-06-14
Handle PInvoke analysis similar to other methods

LakshanF:
This fixes `PInvoke` analysis similar to other methods as discussed in issue #1989. 

- We now generate the warnings from the `callsite` as opposed to for the `PInvoke `call itself. We do this by moving the analysis to `MethodBodyScanner `and `ReflectionMethodBodyScanner`
- Dangerous `PInvokes `(the ones that we generate warnings) have an implicit relationship with built-in COM interop support. Specifically, [IsComInterop](https://github.com/mono/linker/blob/caeaf2a3fb3f636805fdd4881df4f9a539fff8f6/src/linker/Linker.Steps/MarkStep.cs#L3103)method in the linker tries to match the logic in [MarshalInfo::MarshalInfo](https://github.com/dotnet/runtime/blob/fef9b592b28bc4b906091b1361f097da63bf418c/src/coreclr/vm/mlinfo.cpp#L1086). This means that when `BuiltInComInteropSupport `property is false (`netcore `app default), dangerous `PInvokes `will not work. Created a doc [bug ](https://github.com/dotnet/docs/issues/24645)to track this.
- Needed to change the `calli `instruction handling since we dont always detect methods that return void (issue #2090)
- Updated `ComPInvokeWarning `test to now check for warnings from the call sites

Plan regarding the OOB library warnings

- Runtime libraries fix plan: We have around 40 new warnings since we moved the warning location. Creating a draft PR in the runtime repo that will address this that can be used when the linker gets merged

### DIFF
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -526,14 +526,13 @@ namespace Mono.Linker.Dataflow
 						}
 
 						// Pop arguments
-						PopUnknown (currentStack, signature.Parameters.Count, methodBody, operation.Offset);
+						if (signature.Parameters.Count > 0)
+							PopUnknown (currentStack, signature.Parameters.Count, methodBody, operation.Offset);
 
 						// Pop function pointer
 						PopUnknown (currentStack, 1, methodBody, operation.Offset);
 
-						// Push return value - BUG: signature.ReturnType.MetadataType sometime doesn't show void for valid cases
-						// Revisit after https://github.com/mono/linker/issues/2090
-						if (signature.ReturnType.GetElementType ().MetadataType != MetadataType.Void)
+						if (GetReturnTypeWithoutModifiers (signature.ReturnType).MetadataType != MetadataType.Void)
 							PushUnknown (currentStack);
 					}
 					break;
@@ -568,7 +567,9 @@ namespace Mono.Linker.Dataflow
 					break;
 
 				case Code.Ret: {
-						bool hasReturnValue = methodBody.Method.ReturnType.MetadataType != MetadataType.Void;
+
+						bool hasReturnValue = GetReturnTypeWithoutModifiers (methodBody.Method.ReturnType).MetadataType != MetadataType.Void;
+
 						if (currentStack.Count != (hasReturnValue ? 1 : 0)) {
 							WarnAboutInvalidILInMethod (methodBody, operation.Offset);
 						}
@@ -893,7 +894,7 @@ namespace Mono.Linker.Dataflow
 					else
 						methodReturnValue = newObjValue;
 				} else {
-					if (calledMethod.ReturnType.MetadataType != MetadataType.Void) {
+					if (GetReturnTypeWithoutModifiers (calledMethod.ReturnType).MetadataType != MetadataType.Void) {
 						methodReturnValue = UnknownValue.Instance;
 					}
 				}
@@ -907,6 +908,14 @@ namespace Mono.Linker.Dataflow
 					MarkArrayValuesAsUnknown (arr, curBasicBlock);
 				}
 			}
+		}
+
+		protected static TypeReference GetReturnTypeWithoutModifiers (TypeReference returnType)
+		{
+			while (returnType is IModifierType) {
+				returnType = ((IModifierType) returnType).ElementType;
+			}
+			return returnType;
 		}
 
 		// Array types that are dynamically accessed should resolve to System.Array instead of its element type - which is what Cecil resolves to.
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -531,8 +531,9 @@ namespace Mono.Linker.Dataflow
 						// Pop function pointer
 						PopUnknown (currentStack, 1, methodBody, operation.Offset);
 
-						// Push return value
-						if (signature.ReturnType.MetadataType != MetadataType.Void)
+						// Push return value - BUG: signature.ReturnType.MetadataType sometime doesn't show void for valid cases
+						// Revisit after https://github.com/mono/linker/issues/2090
+						if (signature.ReturnType.GetElementType ().MetadataType != MetadataType.Void)
 							PushUnknown (currentStack);
 					}
 					break;
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -37,10 +37,31 @@ namespace Mono.Linker.Dataflow
 			if (methodDefinition == null)
 				return false;
 
-			return
-				GetIntrinsicIdForMethod (methodDefinition) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel ||
+			bool isPInvokeImpl = methodDefinition.IsPInvokeImpl;
+			if (isPInvokeImpl) {
+				// HACK - we don't want to scan C++/CLI assemblies
+				if (IsCPPCLIAssembly (methodDefinition.DeclaringType.Module)) {
+					isPInvokeImpl = false;
+				}
+				// @TODO - there must be a better way to find out methods from C++/CLI assemblies
+				// Revisit after discussing with the team
+				static bool IsCPPCLIAssembly (ModuleDefinition module)
+				{
+					foreach (var type in module.Types)
+						if (type.Namespace == "<CppImplementationDetails>" ||
+							type.Namespace == "<CrtImplementationDetails>")
+							return true;
+
+					return false;
+				}
+			}
+
+			bool requiredCheck = GetIntrinsicIdForMethod (methodDefinition) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel ||
 				context.Annotations.FlowAnnotations.RequiresDataFlowAnalysis (methodDefinition) ||
-				context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (methodDefinition);
+				context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (methodDefinition)
+				|| isPInvokeImpl;
+
+			return requiredCheck;
 		}
 
 		public static bool RequiresReflectionMethodBodyScannerForMethodBody (FlowAnnotations flowAnnotations, MethodDefinition methodDefinition)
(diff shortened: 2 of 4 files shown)


src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -531,8 +531,9 @@ public void Scan (MethodBody methodBody)
 						// Pop function pointer
 						PopUnknown (currentStack, 1, methodBody, operation.Offset);
 
-						// Push return value
-						if (signature.ReturnType.MetadataType != MetadataType.Void)
+						// Push return value - BUG: signature.ReturnType.MetadataType sometime doesn't show void for valid cases
+						// Revisit after https://github.com/mono/linker/issues/2090
+						if (signature.ReturnType.GetElementType ().MetadataType != MetadataType.Void)
```
MichalStrehovsky, line 536: We don't want to treat `void*` as `void` here. AFAIK GetElementType will strip off the pointerness. This should `Resolve` the return type or whatever is the convenient thing to do for it in Cecil.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -37,10 +37,31 @@ public static bool RequiresReflectionMethodBodyScannerForCallSite (LinkContext c
...
-			return
-				GetIntrinsicIdForMethod (methodDefinition) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel ||
+			bool isPInvokeImpl = methodDefinition.IsPInvokeImpl;
+			if (isPInvokeImpl) {
+				// HACK - we don't want to scan C++/CLI assemblies
+				if (IsCPPCLIAssembly (methodDefinition.DeclaringType.Module)) {
+					isPInvokeImpl = false;
+				}
+				// @TODO - there must be a better way to find out methods from C++/CLI assemblies
+				// Revisit after discussing with the team
```
MichalStrehovsky, line 47: The proper way would be to detect the things that are actually problematic for us:

* the assembly has a p/invoke with a null entrypoint (indicates we're calling native code within the current module), or
* there's any `vtfixup`s (see ECMA-335 spec).

The p/invoke check is not very cheap in Cecil (not sure if we can enumerate all the p/invokes in the module). It is cheap to check with System.Reflection.Metadata.

This is probably the best we can do right now.

Neither of those would detect things like https://twitter.com/MStrehovsky/status/1384221192702218250 which IL Linker is also going to damage, but we probably have to just live with that.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -37,10 +37,31 @@ public static bool RequiresReflectionMethodBodyScannerForCallSite (LinkContext c
 			if (methodDefinition == null)
 				return false;
 
-			return
-				GetIntrinsicIdForMethod (methodDefinition) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel ||
+			bool isPInvokeImpl = methodDefinition.IsPInvokeImpl;
+			if (isPInvokeImpl) {
+				// HACK - we don't want to scan C++/CLI assemblies
```
MichalStrehovsky, line 42: Why are we avoiding C++/CLI?

LakshanF: Primarily as a convenience in the initial support. Our existing analysis of the method body instructions seem inadequate from the few tests we had (IL scanning errors) and given native code interaction to be very common in C++/CLI, I avoided adding support in this iteration.

MichalStrehovsky: If we are concerned the method body scanner could crash, should we also block scanning for reflection patterns then?

C++/CLI can still do stuff that would trigger the method body scanner (call into reflection intrinsics, or RUC code). If we are afraid we would crash, we need to do something about that. All those crashes are just bugs - there's nothing fundamentally hard in scanning the IL bodies of C++/CLI assemblies.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
sbomer: `Resolve` will also strip off the pointerness (it does `GetElementType`). I think we need to check the `GetElementType` only if `returnType.IsOptionalModifier`.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -654,6 +731,29 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 				calledMethodDefinition,
 				operation);
 
+			if (calledMethodDefinition.IsPInvokeImpl) {
+				// Is the PInvoke dangerous?
+				bool comDangerousMethod = IsComInterop (calledMethodDefinition.MethodReturnType, calledMethodDefinition.ReturnType);
+				foreach (ParameterDefinition pd in calledMethodDefinition.Parameters) {
+					TypeReference paramTypeReference = pd.ParameterType;
+					if (paramTypeReference is TypeSpecification) {
+						paramTypeReference = (paramTypeReference as TypeSpecification).ElementType;
```
sbomer, line 740: I see that this is just copying existing logic from MarkStep, but I don't understand why we look at the element type here. Is this again to handle modifiers? Maybe it would be better to handle this kind of logic inside of the `IsComInterop` helper.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: If we think C++/CLI is broken, OK - let's fix it, but I don't see a reason to try to "make up for it" in one specific area. I personally would remove any C++/CLI specific handling here - unless we need new logic to correctly detect PInvokes in C++/CLI (per the other discussion).

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -37,10 +37,31 @@ public static bool RequiresReflectionMethodBodyScannerForCallSite (LinkContext c
...
+						if (type.Namespace == "<CppImplementationDetails>" ||
+							type.Namespace == "<CrtImplementationDetails>")
+							return true;
+
+					return false;
+				}
+			}
+
+			bool requiredCheck = GetIntrinsicIdForMethod (methodDefinition) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel ||
 				context.Annotations.FlowAnnotations.RequiresDataFlowAnalysis (methodDefinition) ||
```
vitek-karas, line 60: If the PInvoke check is expensive, consider adding it into the FlowAnnotations - that has a cache already built-in. We would only need to pay the price once per method then.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -654,6 +731,29 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+					}
+				}
+
+				if (comDangerousMethod) {
+					reflectionContext.AnalyzingPattern ();
+					reflectionContext.RecordUnrecognizedPattern (2050, $"P/invoke method '{calledMethodDefinition.GetDisplayName ()}' declares a parameter with COM marshalling. Correctness of COM interop cannot be guaranteed after trimming. Interfaces and interface members might be removed.");
+				}
+
+				// We did not handle this call
+				return false;
```
vitek-karas, line 754: This will prevent RUC on PInvoke method to work. I can imagine cases where that might be useful. Not counting that there's no reason for it to NOT work.

I would prefer if we move this check to the default branch in the switch.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
MichalStrehovsky: I think it helps us to deal with arrays, but yes, it might be better to unwrap that in IsComInterop helper.

I was scratching my head why I wrote it that way but this code has been there for 4 years, I just failed to notice that paramTypeDefinition is not actually the parameter type definition.

LakshanF: Removed the spec check since Resolve in the `IsComInterop `helper seems to handle this

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1829,6 +1729,23 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 					break;
 
 				default:
+
+					if (calledMethodDefinition.IsPInvokeImpl) {
+						// Is the PInvoke dangerous?
+						bool comDangerousMethod = IsComInterop (calledMethodDefinition.MethodReturnType, calledMethodDefinition.ReturnType);
+						foreach (ParameterDefinition pd in calledMethodDefinition.Parameters) {
+							TypeReference paramTypeReference = pd.ParameterType;
+							if (paramTypeReference != null) {
```
vitek-karas, line 1738: Can it actually happen that the type is null?
I don't think we check this anywhere else....

test/Mono.Linker.Tests.Cases/Interop/PInvoke/Warnings/ComPInvokeWarning.cs
```diff
@@ -15,39 +15,175 @@ class ComPInvokeWarning
...
 		[UnconditionalSuppressMessage ("trim", "IL2050")]
+		static void Call_CanSuppressWarningOnReturnType ()
+		{
+			CanSuppressWarningOnReturnType ();
+		}
 		[DllImport ("Foo")]
 		static extern IFoo CanSuppressWarningOnReturnType ();
 
-		[ExpectedWarning ("IL2050")] // Issue https://github.com/mono/linker/issues/1989
 		[RequiresUnreferencedCode ("test")]
```
vitek-karas, line 134: Can you please also add a test that RUC on the Pinvoke itself works as expected?
I know it's not exactly what this fix is about, but it's very related.

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/Interop/PInvoke/Warnings/ComPInvokeWarning.cs
```diff
@@ -140,6 +141,17 @@ static void Call_CanSuppressWithRequiresUnreferencedCode ()
...
+		static void Call_CanSuppressPInvokeWithRequiresUnreferencedCode ()
+		{
+			CanSuppressPInvokeWithRequiresUnreferencedCode (null);
+		}
+
+		[RequiresUnreferencedCode ("test")]
+		[DllImport ("Foo")]
+		static extern void CanSuppressPInvokeWithRequiresUnreferencedCode (IFoo foo);
+
+
```
vitek-karas, line 154: I would add one more where we expect the warning (not suppress it) just to make sure it works correctly.

PR #2098, opened 2021-06-16
Start sharing message strings between the analyzer and the linker

agocke:


src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -191,7 +192,7 @@ private void ReportRequiresDiagnostic (OperationAnalysisContext operationContext
 		private string GetUrlFromAttribute (AttributeData? requiresAssemblyFilesAttribute)
 		{
 			var url = requiresAssemblyFilesAttribute?.NamedArguments.FirstOrDefault (na => na.Key == "Url").Value.Value?.ToString ();
```
tlakollo, line 194: I guess here there are two things wrong first the parameter of the method is called requiresAssemblyFilesAttribute instead of requiresAttribute which is probably my fault, and second the function is called GetRucUrlArg (and there is a GetRucMessageArg). GetUrlFromAttribute is in the base because is generic to whichever RequiresAttribute so calling a method with Ruc on the name sounds funky maybe we should call them GetRequiresUrlArg/GetRequiresMessageArg

src/ILLink.Shared/ILLink.Shared.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
marek-safar, line 1: Is it really worth adding complexities with managing a shared library for 4 lines of c# code that could be easily shared differently?

tlakollo: They share the resource file, and while at this moment only contains IL2026, the amount of possible shared information will increase. Only from other PR that I have open it's other 4 diagnostics that could be shared and support for DynamicallyAccessMembers in the analyzer would make sharing the diagnostics even more important.

marek-safar: Was there any problem sharing the resource file as a build input?

agocke: Not sure what you mean by that -- you mean sharing by including files into both projects?

agocke: Changed to compile the shared files in separately. It's a little bit tricky when dealing with shared resources, but presumably we shouldn't have to touch that very often.

marek-safar (approved):

tlakollo (approved):

Issue #986, opened 2020-03-10
Create a list of error/warning codes


mateoatr:
Related to #769 

By adhering to the format used by MSBuild to output warnings and errors to the standard output (which give us for free the capacity to surface these warnings/errors inside IDEs such as Visual Studio), we are required to define (at least) the following items:

Origin: tool name or filename, more broadly what is producing the warning/error. I think that if we are to make the warnings actionable by the user, we should show the filename where the warning is coming from together with the line and column.

Code: list of error/warning codes specific to the illinker, these have to be prefixed with a string that identifies the linker. There's an [ongoing discussion](https://github.com/dotnet/roslyn/issues/40351) related to this in the Roslyn repo.

Some examples of the MSBuild message format together with the complete list of its parameters can be found [here](https://docs.microsoft.com/en-us/archive/blogs/msbuild/msbuild-visual-studio-aware-error-messages-and-message-formats).

marek-safar:
I suspect we'll have to create 2 code groups (similar to what csc does with CSxxxx and CAxxxx) to cover cases where linker produces error/warning and linker analyzer produces error/warning

mateoatr:
Trying to move forward with this, here's a table with all different messages that we currently log via `LogMessage` inside the linker, together with a proposed set of `category`, `subcategory` and error/warning `code` to use for these.
| Text | Proposed category | Proposed subcategory | Proposed code | Location in codebase |
|-|-|-|-|-|
| `Could not resolve '{assemblyName}' assembly dependency` | Warning | Unresolved dependency | 6001 | MarkStep |
| `Could not resolve '{typeName}' type dependency` | Warning | Unresolved dependency | 6002 | MarkStep |
| `Could not resolve dependency member '{member}' declared in type '{td.FullName}'` |Error | Unresolved dependency | 0001 | MarkStep |
| `Could not resolve type '{fullname}' for substitution` | Error || 0002 | BodySubstituterStep |
| `Could not match assembly '{name.FullName}' for substitution` | Error ||0003| BodySubstituterStep |
| `Could not find method '{signature}' for substitution` | Error ||0004|BodySubstituterStep |
| `Could not find field '{name}' for substitution.` | Error || 0005 | BodySubstituterStep |
| `Invalid value for '{signature}' stub` | Error || 0006 | BodySubstituterStep |
| `Unknown body modification '{action}' for '{signature}'` | Error || 0007 | BodySubstituterStep |
[...]

PR #2101, opened 2021-06-18
Add enum to hold new diagnostic IDs

agocke:


src/ILLink.Shared/DiagCode.cs
```diff
@@ -0,0 +1,13 @@
+
+namespace ILLink.Shared
+{
+	internal enum DiagCode
+	{
+		WRN_UsingRequiresUnreferencedCode = 2026
```
tlakollo, line 6: Just double-checking, the point would be to use this enum even if the analyzer doesn't use the id, meaning it doesn't have to be shared between linker and analyzer to be in the enum and therefore we can migrate whichever id from error-codes here

MichalStrehovsky: There was a discussion about having an enum for error codes in the past and it was deliberately not done: https://github.com/mono/linker/issues/986#issuecomment-610228419.

agocke: I think that discussion is a bit obsolete now because the linker repo is no longer where we want to document error codes, so whatever pattern we choose in code should be purely for our benefit. The documentation for warnings should be on docs.microsoft.com (I'll try to start moving stuff over shortly).

This PR is still in draft because I haven't quite figured out the most useful pattern here.

PR #2108, opened 2021-06-23
Better tests for various RUC scenarios and fixes

vitek-karas:
Adds several new tests for various RUC scenarios.
Cleans up the RUC tests to correctly baseline all warnings it produces.
Fixes a product issue with RUC on attribute property setter.
Small refactoring in the product.


marek-safar (approved):

vitek-karas:
I think this is now finally ready for the review. I added a lot more test cases and reworked the implementation.
Previously we had a select few cases where we did warn. This actually missed quit a few cases. I went through almost all possible values and instead picked those for which we don't want to warn.

agocke (approved):

agocke:
LGTM, merging so we can rebase on top

PR #2112, opened 2021-06-24
Add bootstrap run on Linux

agocke:
Adds support for a bootstrap flag on Linux, which builds the linker and uses
it to trim a test copy of csc. The trimmed csc is then used to build the
linker and tests. This should act as a real-world trimming test.


eng/azure-pipelines.yml
```diff
@@ -101,6 +101,28 @@ stages:
           parameters:
             enablePublishTestResults: true
 
+      - ${{ if eq(variables.officialBuild, 'false') }}:
```
marek-safar, line 104: Why is it not included in the existing --integrationTest ?

Directory.Build.props
```diff
@@ -27,6 +27,14 @@
     <DebugType Condition=" '$(IsReferenceAssembly)' == 'true' ">none</DebugType>
   </PropertyGroup>
 
+  <Import Project="$(BootstrapLinkerPath)/build/Microsoft.NET.ILLink.Tasks.props" Condition="'$(BootstrapLinkerPath)' != ''" />
```
marek-safar, line 30: Wondering if this would be better only under test/TestCsc

test/TestCsc/csc.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information. -->
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net6.0</TargetFrameworks>
+    <ServerGarbageCollection>true</ServerGarbageCollection>
+    <PublishTrimmed>true</PublishTrimmed>
```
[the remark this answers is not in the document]
agocke: Yup, it goes into `artifacts/bin/csc/Debug/net6.0/linux-x64/publish/`. This is because of the automatic Directory.Build.props importing.

eng/azure-pipelines.yml
agocke: You mean why doesn't the existing ``integrationTest` imply bootstrap? Good question. First reason is that I didn't know about that flag until I started on adding the pipeline, which I did last.

I think I might be able to integrate it. Let me take a shot at it.

eng/Version.Details.xml
```diff
@@ -20,5 +20,9 @@
       <SourceBuild RepoName="runtime" />
       -->
     </Dependency>
+    <Dependency Name="Microsoft.CodeAnalysis.CSharp" Version="4.0.0-2.21323.33">
+      <Uri>dotnet/roslyn</Uri>
```
akoeplinger, line 24: I know darc can fix the missing sha but not sure it can fix the Uri so probably safer to use `https://github.com/dotnet/roslyn`

eng/azure-pipelines.yml
```diff
@@ -101,6 +101,27 @@ stages:
...
+        - job: Bootstrap
+          condition: eq(variables.officialBuild, 'false')
+          pool:
+            name: Hosted Ubuntu 1604
```
akoeplinger, line 108: do you really need Ubuntu 16.04? AzDO throws up warnings now that it will be removed soon: https://github.com/actions/virtual-environments/issues/3287

```suggestion
            vmImage: ubuntu-20.04
```

agocke:
@marek-safar Anything else?

build.sh
```diff
@@ -11,6 +31,42 @@ while [[ -h $source ]]; do
...
+
+# Import Arcade functions
+. "$scriptroot/eng/common/tools.sh"
+
+function BuildBootstrapCompiler {
+  InitializeDotNetCli true
+  # Build linker
+  echo "Building illink"
+
+  local dir="$artifacts_dir/bootstrap"
```
sbomer, line 44: I'd prefer to do this in msbuild so that it works on all platforms. Ideally you'd be able to run `dotnet test test/Mono.Linker.Tests/Mono.Linker.Tests.csproj /p:BootstrapCsc=true` or similar and it would set up the bootstrap compiler.

test/TestCsc/Program.cs
```diff
@@ -0,0 +1,33 @@
...
+                textWriter,
+                new NoopLoader());
+        }
+    }
+
+	internal sealed class NoopLoader : IAnalyzerAssemblyLoader
+	{
+	    public void AddDependencyLocation (string fullPath) { }
+
+	    public Assembly? LoadFromPath (string fullPath) => null;
```
sbomer, line 31: nit: indentation looks a bit off. Maybe this could be added to the linted files.

test/TestCsc/Csc.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.CodeAnalysis.CommandLine;
+
+namespace Microsoft.CodeAnalysis.CSharp.CommandLine
+{
+    internal sealed class Csc : CSharpCompiler
+    {
+        internal Csc(string responseFile, BuildPaths buildPaths, string[] args, IAnalyzerAssemblyLoader analyzerLoader)
+            : base(CSharpCommandLineParser.Default, responseFile, args, buildPaths, Environment.GetEnvironmentVariable("LIB"), analyzerLoader)
```
sbomer, line 15: What's the LIB env var used for?

build.sh
agocke: Per our previous conversation, I'd rather not do this in MSBuild as it would involve either creating another project file, which seems like overkill to me, or having a circular dependency where the linker is dependent upon itself.

This is expressible in MSBuild, but it doesn't seem like a particularly clear use of it.

Re: xplat -- happy to port this to our `cmd` files as well. Getting linux working was just the first and easiest step.

sbomer: I'd still prefer for this to work with `dotnet` tooling, but it's probably more important to get the extra validation working. And I can see that MSBuild might not be the clearest way to write it. :)

sbomer (approved):

marek-safar:
Closing as it does not seem relevant anymore

PR #2113, opened 2021-06-25
Test restructure

### LINKED ISSUE #2077: Analyzer testing restructure
Given the increase in the number of tests for the different analyzers, we have now repeated several of the tests to cover different attribute usages. In general, we have two approaches to reduce the number of repeated tests:
- Share tests between the linker and the analyzer. Given that ideally the linker and the analyzer should produce the same diagnostics sharing the tests between the linker and analyzer makes sense. The problem is that we can only share tests for things that are tested in the linker repo, this means that we cannot test RAF as part of the linker tests since the linker doesn't conduct a global analysis of this attribute.
- Share tests between analyzers. The RAF and RUC analyzer tests are similar and therefore we can use a base class to share the test between them.

The proposal is an idea originally by @mateoatr to modify the ExpectedWarning attribute to be able to differentiate between a warning that is generated by the linker, the analyzer, or both. This way we could have analyzer validation inside linker tests and cover analyzer scenarios by reusing tests in the linker. Consider the following example:

```C#
[ExpectedWarning ("IL2026", "Message for --RequiresWithMessageOnly--.")]
// ExpectedTypes includes something like Analyzer, Linker, and All values, All is the default
[ExpectedWarning ("IL3002", ExpectedBy = ExpectedTypes.Analyzer]
static void TestRequiresWithMessageOnlyOnMethod ()
{
	RequiresWithMessageOnly ();
}

[RequiresUnreferencedCode ("Message for --RequiresWithMessageOnly--")]
[RequiresAssemblyFiles]
static void RequiresWithMessageOnly () {}
```
The result checker in the linker will check if the ExpectedTypes.Linker is active, otherwise; it would disregard the warning. We could do something similar for the analyzer testing infrastructure to test only what is marked with ExpectedTypes.Analyzer. Explaining the code scenario the warning IL2026 would be expected for both the analyzer and the linker, meaning that both testing infrastructures would check for the warning being generated. In the case of IL3002 the result checker in linker will bypass the diagnostic but the test infrastructure of the analyzer will check for the diagnostic.

This would simplify the testing since there would be only 1 testing file in the linker for requirescapability that tests RAF and RUC on both linker and analyzer. 

There would be still analyzer testing since the CodeFixers cannot be tested in the linker infrastructure.

### LINKED ISSUE #2107: Support analyzer tests with additional assemblies
Currently it seems that we don't have much coverage for analyzer tests of interactions between multiple assemblies. Specifically for the RUC analyzer, test cases for code which calls RUC annotated methods from different assembly.

Linker has some of these tests in `RequiresUnreferencedCodeCapability` but the analyzer test infra doesn't handle `SetupCompileBefore` and similar attributes, so those cases has to be disabled for analyzer.

tlakollo:
Add ProducedBy enum to define where the diagnostic can be generated
ExpectedWarningAttribute and LogContainsAttribute now have a property to define where the diagnostic can be generated using the ProducedBy enum
Modified the different tests environments to support the new property, also test will only warn if the id is supported by the current running analyzer
RunTest function now accepts additional references to be inserted before compilation
RequiresUnreferencedCodeCapability.cs file is now only RequiresCapability.cs since is now agnostic of which kind of RequiresAttribute is testing
Fixes #2077
Fixes #2107


test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
```diff
@@ -40,7 +40,7 @@ public static DiagnosticResult Diagnostic (DiagnosticDescriptor descriptor)
 			=> CreateCompilation (CSharpSyntaxTree.ParseText (src), globalAnalyzerOptions, additionalReferences);
 
 		public static async Task<Compilation> GetCompilation (string source, IEnumerable<MetadataReference>? additionalReferences = null)
-			=> (await CSharpAnalyzerVerifier<RequiresAssemblyFilesAnalyzer>.CreateCompilation (source, additionalReferences: additionalReferences ?? Array.Empty<MetadataReference> ())).Compilation.Compilation;
+			=> (await CSharpAnalyzerVerifier<TAnalyzer>.CreateCompilation (source, additionalReferences: additionalReferences ?? Array.Empty<MetadataReference> ())).Compilation.Compilation;
```
agocke, line 43: I think the fact that `TAnalyzer` is completely meaningless here means we should probably move a base `CreateCompilation` helper out of this class and into some neutral test file. There are no analyzers on a base compilation, so taking in type parameters is just confusing.

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
```diff
@@ -90,8 +99,15 @@ private void ValidateExpectedWarningAttribute (AttributeSyntax attribute)
...
 		{
-			var arg = Assert.Single (TestCaseUtils.GetAttributeArguments (attribute));
-			var text = TestCaseUtils.GetStringFromExpression (arg.Value);
+			var args = TestCaseUtils.GetAttributeArguments (attribute);
+			var text = TestCaseUtils.GetStringFromExpression (args["#0"]);
+
+			if (args.TryGetValue ("ProducedBy", out var diagnosticProducedBy) &&
+				diagnosticProducedBy is MemberAccessExpressionSyntax memberAccessExpression &&
+				memberAccessExpression.Expression.ToString () == "ProducedBy" &&
+				memberAccessExpression.Name.Identifier.ValueText == "Linker") {
```
agocke, line 108: Extract shared method for the two places you use this.

agocke:
```suggestion
			if (args.TryGetValue ("ProducedBy", out var diagnosticProducedBy) &&
				diagnosticProducedBy is MemberAccessExpressionSyntax memberAccessExpression &&
				memberAccessExpression.Expression.ToString () == nameof(ProducedBy) &&
				memberAccessExpression.Name.Identifier.ValueText == nameof(ProducedBy.Linker)) {
```

test/Mono.Linker.Tests.Cases/RequiresCapability/Dependencies/RequiresAttributeInCopyAssembly.cs
```diff
@@ -9,48 +9,56 @@
...
-	public class RequiresUnreferencedCodeInCopyAssembly
+	public class RequiresAttributeInCopyAssembly
 	{
-		public RequiresUnreferencedCodeInCopyAssembly ()
+		public RequiresAttributeInCopyAssembly ()
 		{
 		}
 
 		[RequiresUnreferencedCode ("Message for --Method--")]
+		[RequiresAssemblyFiles (Message = "Message for --Method--")]
```
vitek-karas, line 19: The fact that the two attributes can't look the same is really not ideal. I filed https://github.com/dotnet/runtime/issues/55274 to track changing it.

test/Mono.Linker.Tests.Cases/RequiresCapability/Dependencies/RequiresAttributeInCopyAssembly.cs
```diff
@@ -9,48 +9,56 @@
...
 		public void Method ()
 		{
 		}
 
 		[RequiresUnreferencedCode ("Message for --UncalledMethod--")]
+		[RequiresAssemblyFiles (Message = "Message for --UncalledMethod--")]
```
vitek-karas, line 25: This looks fragile - in that if somebody adds a new test case and forgets to add both attributes we're loosing coverage (and it's really hard to find). Maybe we can "fix" this by:
* Adding a unit test which goes over the test classes in the right namespace and validates that if one attribute is there, there's also the second one
* Would need some way to tell this test that there are cases where it's intentional that there's a difference.

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -659,6 +659,13 @@ void VerifyKeptAllTypesAndMembersInAssembly (AssemblyDefinition linked)
 			}
 		}
 
+		bool IsProducedOnlyByAnalyzer (CustomAttribute attr)
```
vitek-karas, line 662: Similarly to the analyzer I would prefer this to work the other way round (test for ProducedBy Linker)

vitek-karas: In fact it seems the code actually does it already, in which case I would just rename the method and swap the return value meaning. The linker tests should only care about "linker" stuff - and "ignore" anything else.

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -671,6 +678,9 @@ void VerifyLoggedMessages (AssemblyDefinition original, LinkerTestLogger logger,
 						case nameof (LogContainsAttribute): {
 								var expectedMessage = (string) attr.ConstructorArguments[0].Value;
 
+								if (IsProducedOnlyByAnalyzer (attr))
```
mateoatr, line 681: Should we move this check just before the switch?

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -703,6 +713,10 @@ void VerifyLoggedMessages (AssemblyDefinition original, LinkerTestLogger logger,
 								if (!expectedWarningCode.StartsWith ("IL")) {
 									Assert.Fail ($"The warning code specified in {nameof (ExpectedWarningAttribute)} must start with the 'IL' prefix. Specified value: '{expectedWarningCode}'.");
 								}
+
+								if (IsProducedOnlyByAnalyzer (attr))
+									break;
```
mateoatr, line 718: Same as above.

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
```diff
@@ -56,16 +57,32 @@ public void ValidateAttributes (List<AttributeSyntax> attributes)
...
 
 			if (!expectedWarningCode.StartsWith ("IL"))
 				return;
 
-			if (args.TryGetValue ("GlobalAnalysisOnly", out var globalAnalysisOnly) &&
-				globalAnalysisOnly is LiteralExpressionSyntax { Token: { Value: true } })
+			bool isSupportedDiagnostic = false;
+			foreach (var supportedDiagnostic in Compilation.Analyzers.Single ().SupportedDiagnostics) {
+				if (supportedDiagnostic.Id == expectedWarningCode)
+					isSupportedDiagnostic = true;
```
mateoatr, line 83: Break loop here?

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
```diff
@@ -75,13 +92,10 @@ private void ValidateExpectedWarningAttribute (AttributeSyntax attribute)
 
 			Assert.True (
 				DiagnosticMessages.Any (mc => {
-					if (mc.Id != expectedWarningCode)
-						return true;
```
mateoatr, line 79: Why don't we keep verifying that the warning codes match?

tlakollo: Just realized that I had an idea and implemented another. Basically, I wanted to do this check along with checking that the id is supported by the current analyzer. Need to go over it again to make it work correctly.

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -661,13 +661,22 @@ void VerifyKeptAllTypesAndMembersInAssembly (AssemblyDefinition linked)
 			}
 		}
 
+		bool IsProducedByLinker (CustomAttribute attr)
+		{
+			var propertyObject = attr.GetPropertyValue ("ProducedBy");
+			ProducedBy diagnosticProducedBy = propertyObject is null ? ProducedBy.LinkerAndAnalyzer : (ProducedBy) propertyObject;
```
mateoatr, line 667: Nit: 
```suggestion
			return propertyObject is null ? true : ((ProducedBy) propertyObject).HasFlag (ProducedBy.Linker);
```

test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
```diff
@@ -13,20 +16,27 @@ namespace ILLink.RoslynAnalyzer.Tests
 	/// </summary>
 	public class LinkerTestCases : TestCaseUtils
 	{
-		[Theory]
-		[MemberData (nameof (TestCaseUtils.GetTestData), parameters: nameof (RequiresCapability))]
-		public void RequiresCapability (MethodDeclarationSyntax m, List<AttributeSyntax> attrs)
+		private static Lazy<IEnumerable<MetadataReference>> _additionalReferences = new Lazy<IEnumerable<MetadataReference>> (LoadDependencies);
+
+		private static IEnumerable<MetadataReference> LoadDependencies ()
```
mateoatr, line 21: Why not put this in LinkerTestCases.cs? It feels like `GetReferenceFilesByDirName` should only be used there and have the dependencies resolved before compilation (ideally through looking at the linker attributes that we use to add assemblies to the compilation).

test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
```diff
@@ -13,20 +16,27 @@ namespace ILLink.RoslynAnalyzer.Tests
...
-			// There is a discrepancy between the way linker and the analyzer represent the location of the error,
-			// linker will point to the method caller and the analyzer will point to a line of code.
-			// The TestTypeIsBeforeFieldInit scenario is supported by the analyzer, just the diagnostic message is different
-			// We verify the analyzer generating the right diagnostic in RequiresUnreferencedCodeAnalyzerTests.cs
-			case "TestTypeIsBeforeFieldInit":
-				return;
+			List<MetadataReference> additionalReferences = new List<MetadataReference> ();
+			var s_refFiles = GetReferenceFilesByDirName ();
+			foreach (var refFile in s_refFiles["Dependencies"]) {
+				if (refFile.Contains ("RequiresCapability"))
```
mateoatr, line 26: It feels somewhat fragile looking into the string of a filename here.

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -48,16 +49,25 @@ public static IEnumerable<object[]> GetTestData (string testSuiteName)
...
-					// Currently, the analyzer's test infra only understands these attributes when placed on methods.
+					// Currently, the analyzer's test infra only understands these attributes when placed on methods and properties.
 					case "ExpectedWarning":
 					case "LogContains":
 					case "LogDoesNotContain":
-						return attr.Ancestors ().OfType<MemberDeclarationSyntax> ().First ().IsKind (SyntaxKind.MethodDeclaration);
+						var ancestor = attr.Ancestors ().OfType<MemberDeclarationSyntax> ().First ();
+						if (ancestor.IsKind (SyntaxKind.MethodDeclaration) || ancestor.IsKind (SyntaxKind.PropertyDeclaration) || ancestor.IsKind (SyntaxKind.GetAccessorDeclaration) || ancestor.IsKind (SyntaxKind.SetAccessorDeclaration))
+							return true;
+						return false;
```
mateoatr, line 70: Nit: I think it helps readability to have one line between the `if () <exp>; <exp>;`. I've noticed this pattern in the codebase, although I didn't find any coding style doc (searched in runtime repo) that clearly states this.

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -68,11 +78,27 @@ static bool IsWellKnown (AttributeSyntax attr)
 			}
 		}
 
-		public static void RunTest<TAnalyzer> (MemberDeclarationSyntax m, List<AttributeSyntax> attrs, params (string, string)[] MSBuildProperties)
+		public static async Task<Compilation> GetCompilation (string src)
+		{
+			var srctree = CSharpSyntaxTree.ParseText (src);
+			var mdRef = MetadataReference.CreateFromFile (typeof (Mono.Linker.Tests.Cases.Expectations.Metadata.BaseMetadataAttribute).Assembly.Location);
```
mateoatr, line 84: Nit: Why not have a using and put `typeof (BaseMetadataAttribute)` here?

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -68,11 +78,27 @@ static bool IsWellKnown (AttributeSyntax attr)
...
+			var comp = CSharpCompilation.Create (
```
mateoatr, line 85: This could probably be a good place to call a method that pulls in dependencies. Probably by understanding the `SetupCompileBefore`/`SetupCompileAfter` attributes.

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -95,6 +121,38 @@ public static void RunTest<TAnalyzer> (MemberDeclarationSyntax m, List<Attribute
...
+				} else {
+					sources = new List<string> () { file };
+					builder[dirName] = sources;
+				}
+			}
+
+			return builder.ToImmutable ();
+		}
+
+		public static IEnumerable<string> GetReferenceFiles ()
```
mateoatr, line 141: This is only called by `GetReferenceFilesByDirName`, should we collapse these two methods into just one?

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -95,6 +121,38 @@ public static void RunTest<TAnalyzer> (MemberDeclarationSyntax m, List<Attribute
...
+		{
+			GetDirectoryPaths (out var rootSourceDir, out _);
+
+			foreach (var subDir in Directory.EnumerateDirectories (rootSourceDir, "*", SearchOption.AllDirectories)) {
```
mateoatr, line 145: Why not use `Directory.GetDirectories(rootSourceDir)` instead?

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
```diff
@@ -56,16 +57,35 @@ public void ValidateAttributes (List<AttributeSyntax> attributes)
 			}
 		}
 
+		private bool IsProducedByAnalyzer (Dictionary<string, ExpressionSyntax> args)
+		{
+			return !args.TryGetValue ("ProducedBy", out var diagnosticProducedBy) || (
```
mateoatr, line 62: Could we do something similar to what we have in `IsProducedByLinker` instead of having this very long expression?

tlakollo:
Closing in favor of #2336

Issue #1174, opened 2020-05-06
DynamicallyAccessMemberType.PublicNestedTypes behavior discussion


vitek-karas:
If the `DynamicallyAccessedMemberTypes.PublicNestedTypes` is used (same applies to `NonPublicNestedTypes`) we will only "mark" the nested types - which includes the bare minimum about the type. So it will make the `Type.GetNestedType` work, but it's very unlikely that the calling code will end up working overall as the type as marked is effectively useless (no methods, no `.ctors`, nothing really).

It's true that in the current state the user code would get lot of warnings anyway as any action on the type returned by `Type.GetNestedType` will result in a warning. So in that sense the behavior is still correct, but it makes the `PublicNestedTypes`/`NonPublicNestedType enum` values borderline useless.

/cc @MichalStrehovsky @sbomer @tlakollo @marek-safar @eerhardt

vitek-karas:
Possible change proposed by @MichalStrehovsky : The semantic that would make sense is to keep the nested type at the `All` level. The `GetNested*` APIs are used rarely enough that the size overhead won't matter (close to 0% on all overloads at https://apisof.net/), the result would be usable and easy to reason about (one can call `GetNested*` recursively and still get something useful out), and it would work for the one "meaningful" use case of `GetNested` in `CoreLib` (event source enumerates fields on nested types with a magic name).

vitek-karas:
If we go with the approach of marking the entire nested type I assume it should not be recursive to second level nested types, but it should probably be "recursive" to the inheritance (so base classes of the first level nested type).

Note: `DynamicallyAccessedMemberTypes.All` is "recursive" in sense of inheritance, it will bring in all the base types of that type as `All` as well.

MichalStrehovsky:
> If we go with the approach of marking the entire nested type I assume it should not be recursive to second level nested types, but it should probably be "recursive" to the inheritance (so base classes of the first level nested type).

I would just make it mean `DynamicallyAccessedMemberTypes.All` on the nested types (i.e. the return value of `GetNestedType` will be annotated as returning a `Type` annotated `DynamicallyAccessedMemberTypes.All`).

This implies keeping the type fully intact (so base classes, all members (including other nested types), interfaces, etc. are all kept).

marek-safar:
> The GetNested* APIs are used rarely enough that the size overhead won't matter (close to 0% on all overloads at apisof.net)

[...]

PR #2114, opened 2021-06-25
Add a library mode for type hierarchy marking

LakshanF:
We added type hierarchy support to the linker via #1929 but the annotations to the derived types were only triggered via `object.GetType()` in code. This causes problems in library mode linking, especially for internal types when the assembly does not have code that triggers `object.GetType()`. Library mode linking can remove members in such cases which might be needed later by application code that invokes the removed internal code via reflection. A specific example is an internal type like [NetEventSource](https://github.com/dotnet/runtime/blob/3924d0308110f6365e47e784b8c7cf1e8f7efa2b/src/libraries/Common/src/System/Net/Logging/NetEventSource.Common.cs#L39) in `System.Net.Mail.dll` library, where the library does not have code that triggers `object.GetType()`. Library mode linking removes nested types in such cases and application code patterns like [this ](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Mail/tests/Functional/LoggingTest.cs#L25) will fail in such cases.

Unfortunately, we cannot yet remove the special case handling that we currently have for `EventSource `in the trimmer with this change as discussed in #1949. This is because some library code relies in the trimmer to remove unused code in custom `EventSources `and that goal conflicts with this change since this change will keep all code in custom `EventSources`. There is a runtime issue [54859](https://github.com/dotnet/runtime/issues/54859) that tracks this. This change moves the EventSource special handling to library mode

- This fix is to allow a new `CodeOptimizations`, `OptimizeTypeHierarchyAnnotations`, that will be disabled in library mode linking. Disabling the optimization will cause types to be marked when `object.GetType()` is not seen and the type is used.
       - `EventSource `types will not be impacted due to the size concerns expressed above
- We continue to special case EventSource in the trimmer. This is done by disabling a new optimization `CodeOptimizations`, `RemoveEventSourceSpecialHandling` in library mode.
       - Type hierarchy annotation in `EventSource `types will not be triggered without `object.GetType()`
       - Also removed the special handling of the `EventDataAttribute `at `MarkTypeSpecialCustomAttributes` in this mode since `EventSource` type has annotations to keep all methods. 
       - We are keeping the special handling of EventSource for earlier runtimes than .NET6.
- Fixed the existing test case since it was not correctly testing `EventSource`
- Added some tests to cover missing scenarios for library mode linking

### DIFF
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
@@ -91,6 +91,9 @@ namespace Mono.Linker.Dataflow
 
 			Debug.Assert (!apply || annotation != DynamicallyAccessedMemberTypes.None);
 
+			// If OptimizeTypeHierarchyAnnotations is disabled, we will apply the annotations without seeing object.GetType()
+			apply |= (annotation != DynamicallyAccessedMemberTypes.None) && !_context.IsOptimizationEnabled (CodeOptimizations.OptimizeTypeHierarchyAnnotations, type);
+
 			// Store the results in the cache
 			// Don't store empty annotations for non-interface types - we can use the presence of the row
 			// in the cache as indication of it instead.
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
@@ -92,7 +92,9 @@ namespace Mono.Linker.Dataflow
 			Debug.Assert (!apply || annotation != DynamicallyAccessedMemberTypes.None);
 
 			// If OptimizeTypeHierarchyAnnotations is disabled, we will apply the annotations without seeing object.GetType()
-			apply |= (annotation != DynamicallyAccessedMemberTypes.None) && !_context.IsOptimizationEnabled (CodeOptimizations.OptimizeTypeHierarchyAnnotations, type);
+			// Unfortunately, we cannot apply the annotation to type derived from EventSource - Revisit after https://github.com/dotnet/runtime/issues/54859
+			apply |= (annotation != DynamicallyAccessedMemberTypes.None) && !_context.IsOptimizationEnabled (CodeOptimizations.OptimizeTypeHierarchyAnnotations, type)
+				&& !_context.IsOptimizationEnabled (CodeOptimizations.RemoveEventSourceSpecialHandling, type) && !BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context);
 
 			// Store the results in the cache
 			// Don't store empty annotations for non-interface types - we can use the presence of the row
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
@@ -96,7 +96,7 @@ namespace Mono.Linker.Dataflow
 			// Unfortunately, we cannot apply the annotation to type derived from EventSource - Revisit after https://github.com/dotnet/runtime/issues/54859
 			// Breaking the logic to make it easier to maintain in the future since the logic is convoluted
 			// DisableEventSourceSpecialHandling is closely tied to a type derived from EventSource and should always go together
-			// However, logically it should be possible to use DisableEventSourceSpecialHandling to control OptimizeTypeHierarchyAnnotations
+			// However, logically it should be possible to use DisableEventSourceSpecialHandling to allow marking types derived from EventSource when OptimizeTypeHierarchyAnnotations is disabled
 			apply |= applyOptimizeTypeHierarchyAnnotations && (_context.DisableEventSourceSpecialHandling || !BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context));
 
 			// Store the results in the cache
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
@@ -92,9 +92,12 @@ namespace Mono.Linker.Dataflow
 			Debug.Assert (!apply || annotation != DynamicallyAccessedMemberTypes.None);
 
 			// If OptimizeTypeHierarchyAnnotations is disabled, we will apply the annotations without seeing object.GetType()
+			bool applyOptimizeTypeHierarchyAnnotations = (annotation != DynamicallyAccessedMemberTypes.None) && !_context.IsOptimizationEnabled (CodeOptimizations.OptimizeTypeHierarchyAnnotations, type);
 			// Unfortunately, we cannot apply the annotation to type derived from EventSource - Revisit after https://github.com/dotnet/runtime/issues/54859
-			apply |= (annotation != DynamicallyAccessedMemberTypes.None) && !_context.IsOptimizationEnabled (CodeOptimizations.OptimizeTypeHierarchyAnnotations, type)
-				&& !_context.IsOptimizationEnabled (CodeOptimizations.RemoveEventSourceSpecialHandling, type) && !BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context);
+			// Breaking the logic to make it easier to maintain in the future since the logic is convoluted
+			// DisableEventSourceSpecialHandling is closely tied to a type derived from EventSource and should always go together
+			// However, logically it should be possible to use DisableEventSourceSpecialHandling to control OptimizeTypeHierarchyAnnotations
+			apply |= applyOptimizeTypeHierarchyAnnotations && (_context.DisableEventSourceSpecialHandling || !BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context));
 
 			// Store the results in the cache
 			// Don't store empty annotations for non-interface types - we can use the presence of the row
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1776,8 +1776,8 @@ namespace Mono.Linker.Steps
 			MarkSerializable (type);
 
 			// This marks static fields of KeyWords/OpCodes/Tasks subclasses of an EventSource type.
-			// The special handling of EventSource is not needed in .NET6 since DynamicallyAccessedMembers can be applied to types
-			if (_context.GetTargetRuntimeVersion () < TargetRuntimeVersion.NET6 && BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context)) {
+			// The special handling of EventSource is still needed in .NET6 in library mode
+			if ((_context.GetTargetRuntimeVersion () < TargetRuntimeVersion.NET6 || !_context.IsOptimizationEnabled (CodeOptimizations.RemoveEventSourceSpecialHandling, type)) && BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context)) {
 				MarkEventSourceProviders (type);
 			}
 
@@ -1912,8 +1912,8 @@ namespace Mono.Linker.Steps
 				case "DebuggerTypeProxyAttribute" when attrType.Namespace == "System.Diagnostics":
 					MarkTypeWithDebuggerTypeProxyAttribute (type, attribute);
 					break;
-				// The special handling of EventSource is not needed in .NET6 since DynamicallyAccessedMembers can be applied to types
-				case "EventDataAttribute" when attrType.Namespace == "System.Diagnostics.Tracing" && _context.GetTargetRuntimeVersion () < TargetRuntimeVersion.NET6:
+				// The special handling of EventSource is still needed in .NET6 in library mode
+				case "EventDataAttribute" when attrType.Namespace == "System.Diagnostics.Tracing" && (_context.GetTargetRuntimeVersion () < TargetRuntimeVersion.NET6 || !_context.IsOptimizationEnabled (CodeOptimizations.RemoveEventSourceSpecialHandling, type)):
 					if (MarkMethodsIf (type.Methods, MethodDefinitionExtensions.IsPublicInstancePropertyMethod, new DependencyInfo (DependencyKind.ReferencedBySpecialAttribute, type)))
 						Tracer.AddDirectDependency (attribute, new DependencyInfo (DependencyKind.CustomAttribute, type), marked: false);
 					break;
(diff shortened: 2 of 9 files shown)


src/linker/Linker.Dataflow/ReflectionPatternContext.cs
```diff
@@ -99,7 +99,7 @@ public void RecordUnrecognizedPattern (int messageCode, string message)
 		public void Dispose ()
 		{
 #if DEBUG
-			if (_patternAnalysisAttempted && !_patternReported)
+			if (_patternAnalysisAttempted && !_patternReported && !_context.IsOptimizationEnabled (CodeOptimizations.ApplyTypeHierarchyAnnotation, (MemberReference) MemberWithRequirements))
```
vitek-karas, line 102: This is wrong - if there's a place where this throws we should fix that place and not relax this check. Reading the code I don't see a need for this (I don't see the place where this would throw).

src/linker/Linker/LinkContext.cs
```diff
@@ -984,5 +984,6 @@ public enum CodeOptimizations
 		RemoveSubstitutions = 1 << 21,
 		RemoveLinkAttributes = 1 << 22,
 		RemoveDynamicDependencyAttribute = 1 << 23,
+		ApplyTypeHierarchyAnnotation = 1 << 24,
```
vitek-karas, line 987: I think the new annotation should have the opposite meaning - if it's turned on, we only apply hierarchy annotations when we see the usage. Library mode would disable such optimization. (It would also fit better with it being an "optimization").

src/linker/Linker.Steps/RootAssemblyInputStep.cs
```diff
@@ -71,6 +71,10 @@ protected override void Process ()
 
 				MarkAndPreserve (assembly, preserve_library);
 
+				// Enable type heirarchy apply in library mode to annotate derived types eagerly
+				// Otherwise, type annotation will only be applied with calls to object.GetType()
+				Context.Optimizations.Enable (CodeOptimizations.ApplyTypeHierarchyAnnotation, assembly.Name.Name);
+
```
vitek-karas, line 77: With the opposite meaning this can be set in the below bulk statement.

src/linker/Linker/LinkContext.cs
vitek-karas: Maybe just call it `OptimizeTypeHierarchyAnnotations` or something like that - (and add a comment here describing what is the effect).

test/Mono.Linker.Tests.Cases/BCLFeatures/ETW/CustomEventSource.cs
```diff
@@ -21,24 +22,30 @@ public static void Main ()
...
 	class MyCompanyEventSource : EventSource
 	{
+		[KeptMember (".ctor()")]
 		[Kept]
 		public class Keywords
 		{
 			[Kept]
 			public const EventKeywords Page = (EventKeywords) 1;
 
+			[Kept]
```
vitek-karas, line 32: I think we should link to https://github.com/mono/linker/issues/1174 - since once that is fixed, the annotation in EventSource can become just `PublicMethods | PublicNestedTypes` I think (or something similar, but much smaller than `All`).

test/Mono.Linker.Tests.Cases/Reflection/ObjectGetTypeLibraryMode.cs
```diff
@@ -0,0 +1,154 @@
...
+			[Kept]
+			public static void Test ()
+			{
+				TestInterface (new ClassImplementingNoAnnotatedInterface ());
+				TestClass (new BasicNoAnnotatedClass ());
+				TestStruct (new BasicNoAnnotatedStruct ());
+			}
+		}
+
+
```
vitek-karas, line 152: Nit: Extra new lines

test/Mono.Linker.Tests.Cases/Reflection/ObjectGetTypeLibraryMode.cs
```diff
@@ -0,0 +1,154 @@
...
+				[Kept]
+				public void UsedMethod () { }
+				[Kept] // The type is not sealed, so trimmer will apply the annotation from the interface
+				public void UnusedMethod () { }
+			}
+
+			[Kept]
+			static void TestInterface (ClassImplementingAnnotatedInterface classImplementingInterface)
+			{
+			}
```
vitek-karas, line 50: Nit: It would be a little bit cleaner to just new-up the instance inside this method - instead of passing it as a parameter (as is, this method doesn't nothin, even during marking it won't do anything interesting).

test/Mono.Linker.Tests.Cases/Mono.Linker.Tests.Cases.csproj
```diff
@@ -58,8 +58,6 @@
   </ItemGroup>
   <ItemGroup>
     <Folder Include="Reflection\Dependencies\" />
-    <Folder Include="BCLFeatures\" />
-    <Folder Include="BCLFeatures\ETW\" />
```
vitek-karas, line 62: I assume the change also deletes the files and folders (GitHub doesn't show this in the PR I guess).

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -105,7 +105,7 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
 			// method to look for annotations. If the newly marked type derives from the one we're processing
 			// it will rely on the cache to know if it's annotated - so the record must be in the cache
 			// before it happens.
-			if (apply) {
+			if (apply || (annotation != DynamicallyAccessedMemberTypes.None && _context.IsOptimizationEnabled (CodeOptimizations.ApplyTypeHierarchyAnnotation, type))) {
```
vitek-karas, line 108: An alternative solution would be to set `apply` to true (if the optimization is disabled) on line 56. That would have the added benefit, that the caches would be "in-sync" with the behavior (the cache would store `apply=true` as well). Would also mean that finding the actual `GetType` call for the hierarchy would not do anything (since the cache would tell it that it's already applied) - unlike as-is where that would try to apply the annotation again.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1769,11 +1769,6 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 
 			MarkSerializable (type);
 
-			// This marks static fields of KeyWords/OpCodes/Tasks subclasses of an EventSource type.
```
MichalStrehovsky, line 1772: I think we can't physically delete the support for this at this point - we probably need to keep it under `if (_context.GetTargetRuntimeVersion () < TargetRuntimeVersion.NET6)` same as #1692 did with the serializer support.

LakshanF: Thanks! I'm not very familiar with the conditions where the linker runs in old runtime versions. 

What does this mean with this fix on types with the annotations? Is it safe to assume that pre-NET6 versions would not have annotations on types?

vitek-karas:
We can't take this as is based on the discussion in dotnet/runtime#54859.
For event source we want to keep the existing one-off behavior in library mode (in normal mode we do want to use the type hierarchy marking as it's today).
The problem is that there are other types in the framework which make use of type hierarchy marking - for example https://github.com/dotnet/runtime/blob/46e5fe1c9a4584b006601d46063b353338219af0/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/MaskedTextProvider.cs#L22

For those we want to take this change. I think we'll have to make a "hack" and intentionally ignore the DAM on `EventSource` in the linker when in library mode. I don't like that solution, but I don't have anything better for now...

LakshanF:
Unfortunately, internal custom `EventSources `like [NetEventSource](https://github.com/dotnet/runtime/blob/385b6f3e5fd6fbf08a8a82bde5a5f57f18018c2a/src/libraries/Common/src/System/Net/Logging/NetEventSource.Common.cs#L39) make it challenging to have this changes unless the library fixes to have only the code they need for each library tracing (or special case `EventSource `as suggested above). As things stand now, there are 23 `[NonEvent]` methods to 5 `[Event]` methods in `NetEventSource`. This change will keep the 23 methods that are not needed and this `NetEventSource `source is used in multiple net libraries.

Will move the special casing of `EventSource `in the trimmer to a library mode as suggested by @marek-safar in a new PR

vitek-karas:
What about the other use cases for type hierarchy marking - without the proper fix for library mode, it's bound to break (either now, or soon).

LakshanF:
Unfortunately, rebasing caused some changes that are not part of this fix to be displayed

src/linker/Linker/LinkContext.cs
```diff
@@ -986,5 +988,17 @@ public enum CodeOptimizations
...
+
+		/// <summary>
+		/// Option to apply annotations to type heirarchy
+		/// Enable type heirarchy apply in library mode to annotate derived types eagerly
+		/// Otherwise, type annotation will only be applied with calls to object.GetType()
+		/// </summary>
+		OptimizeTypeHierarchyAnnotations = 1 << 24,
+
+		/// <summary>
+		/// Pption to not special case EventSource
```
vitek-karas, line 1000:
```suggestion
		/// Option to not special case EventSource
```

vitek-karas (approved):
Looks good - please validate the impact on runtime repo before merging (running the runtime repo build with the new linker should basically make no difference)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1776,7 +1776,8 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 			MarkSerializable (type);
 
 			// This marks static fields of KeyWords/OpCodes/Tasks subclasses of an EventSource type.
-			if (BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context)) {
+			// The special handling of EventSource is still needed in .NET6 in library mode
+			if ((_context.GetTargetRuntimeVersion () < TargetRuntimeVersion.NET6 || !_context.IsOptimizationEnabled (CodeOptimizations.RemoveEventSourceSpecialHandling, type)) && BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context)) {
```
sbomer, line 1780: Could we avoid spreading these policy decisions throughout the code, and instead move the TFM check to where we decide which optimizations are enabled?

src/linker/Linker/LinkContext.cs
```diff
@@ -986,5 +988,17 @@ public enum CodeOptimizations
...
+		/// Option to apply annotations to type heirarchy
+		/// Enable type heirarchy apply in library mode to annotate derived types eagerly
+		/// Otherwise, type annotation will only be applied with calls to object.GetType()
+		/// </summary>
+		OptimizeTypeHierarchyAnnotations = 1 << 24,
+
+		/// <summary>
+		/// Option to not special case EventSource
+		/// </summary>
+		RemoveEventSourceSpecialHandling = 1 << 25,
```
sbomer, line 1002: For the special-case handling of serialization and custom operators, we made them separate command-line options instead of optimizations. Should we do the same here for consistency?

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -91,6 +91,11 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
 
 			Debug.Assert (!apply || annotation != DynamicallyAccessedMemberTypes.None);
 
+			// If OptimizeTypeHierarchyAnnotations is disabled, we will apply the annotations without seeing object.GetType()
+			// Unfortunately, we cannot apply the annotation to type derived from EventSource - Revisit after https://github.com/dotnet/runtime/issues/54859
+			apply |= (annotation != DynamicallyAccessedMemberTypes.None) && !_context.IsOptimizationEnabled (CodeOptimizations.OptimizeTypeHierarchyAnnotations, type)
+				&& !_context.IsOptimizationEnabled (CodeOptimizations.RemoveEventSourceSpecialHandling, type) && !BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context);
```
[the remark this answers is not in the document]
LakshanF: Yes, the hack to avoid special handling of `EventSource` is not easy to follow (wish the library source gets fixed so we can avoid this). We wish to apply only if the optimization mode, `OptimizeTypeHierarchyAnnotations`, is disabled. The last 2 are to ensure that `EventSource` types don't follow this.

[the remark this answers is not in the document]
sbomer: I think it should be
```suggestion
				&& !(_context.IsOptimizationEnabled (CodeOptimizations.RemoveEventSourceSpecialHandling, type) && BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context));
```
so that if we are removing event source handling, it'll still apply the annotations to non-eventsource types.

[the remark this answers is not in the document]
sbomer: Based on discussions it sounds like I misunderstood the intention - `RemoveEventSourceSpecialHandling == false` means "treat EventSource specially, including by not applying DAMT type hierarchy annotations". In that case I think the logic should be:
```suggestion
				&& (_context.IsOptimizationEnabled (CodeOptimizations.RemoveEventSourceSpecialHandling, type) || !BCL.EventTracingForWindows.IsEventSourceImplementation (type, _context));
```

And a better name might be "DisableEventSourceDiscovery" or "DisableEventSourceSpecialHandling" - how do those sound to you?

src/linker/Linker/LinkContext.cs
LakshanF: Moved this away from as an optimization

src/linker/Linker.Steps/MarkStep.cs
LakshanF: I prefer to keep this here since we will need to treat `EventSource `as special for lower TFM as long as we support backward compat in the trimmer

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -91,6 +91,14 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
 
 			Debug.Assert (!apply || annotation != DynamicallyAccessedMemberTypes.None);
 
+			// If OptimizeTypeHierarchyAnnotations is disabled, we will apply the annotations without seeing object.GetType()
+			bool applyOptimizeTypeHierarchyAnnotations = (annotation != DynamicallyAccessedMemberTypes.None) && !_context.IsOptimizationEnabled (CodeOptimizations.OptimizeTypeHierarchyAnnotations, type);
+			// Unfortunately, we cannot apply the annotation to type derived from EventSource - Revisit after https://github.com/dotnet/runtime/issues/54859
+			// Breaking the logic to make it easier to maintain in the future since the logic is convoluted
+			// DisableEventSourceSpecialHandling is closely tied to a type derived from EventSource and should always go together
+			// However, logically it should be possible to use DisableEventSourceSpecialHandling to control OptimizeTypeHierarchyAnnotations
```
sbomer, line 99: I find the comment a bit misleading since DisableEventSourceSpecialHandling doesn't change the value of OptimizeTypeHierarchyAnnotations - I would just say that DisableEventSourceSpecialHandling controls whether we skip event sources when applying annotations (in the case where we aren't optimizing type hierarchy annotations).

sbomer (approved):
One last nit, otherwise LGTM. Thank you!

PR #2121, opened 2021-06-30
Use net6.0 reference assemblies in analyzer tests

agocke:
The version is currently hardcoded in TestCaseUtils.cs, but since
these are reference assemblies, as long as the surface area isn't
predicted to change, we can update as necessary.

### DIFF
--- a/test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
@@ -17,7 +17,7 @@ using Xunit;
 
 namespace ILLink.RoslynAnalyzer.Tests
 {
-    public abstract class TestCaseUtils
+	public abstract class TestCaseUtils
 	{
 		public static readonly ReferenceAssemblies Net6PreviewAssemblies =
 			new ReferenceAssemblies (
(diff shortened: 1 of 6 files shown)


test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -2,22 +2,40 @@
...
 using Xunit;
 
 namespace ILLink.RoslynAnalyzer.Tests
 {
 	public abstract class TestCaseUtils
 	{
+		public static readonly ReferenceAssemblies Net6PreviewAssemblies =
+			new ReferenceAssemblies (
+				"net6.0",
+				new PackageIdentity ("Microsoft.NETCore.App.Ref", "6.0.0-preview.5.21224.4"),
```
tlakollo, line 25: I'm not very familiar with ReferenceAssemblies creation but I assume the package identity version doesn't need to be maintained unless we actually need newer ReferenceAssemblies, we just took one version and rely on it even if the linker global.json changes, right?

agocke: Yup, that's the idea. Unless there's a breaking change in the reference assemblies, we shouldn't care which exact version we use.

tlakollo (approved):

Issue #2122, opened 2021-06-30
Latest linker is raising unexpected warnings in System.Text.Json


eerhardt:
With change #2108, we are now getting unexpected ILLink warnings in System.Text.Json. This is blocking merging the latest linker into dotnet/runtime (https://github.com/dotnet/runtime/pull/54907):

```
ILLink: Trim analysis warning IL2026: System.Text.Json.Nodes.JsonArray: Using method 'System.Text.Json.Nodes.JsonNode.CreateDynamicObject(Expression,JsonNode)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Using JsonNode instances as dynamic types is not compatible with trimming. It can result in non-primitive types being serialized, which may have their members trimmed.
ILLink: Trim analysis warning IL2026: System.Text.Json.Nodes.JsonArray: Using method 'System.Text.Json.Nodes.JsonArray.Add<T>(T)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Creating JsonValue instances with non-primitive types is not compatible with trimming. It can result in non-primitive types being serialized, which may have their members trimmed.
ILLink: Trim analysis warning IL2026: System.Text.Json.Nodes.JsonArray: Using method 'System.Text.Json.Nodes.JsonNode.TrySetMemberMethodInfo.get' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. JSON serialization and deserialization might require types that cannot be statically analyzed. Use the overload that takes a JsonTypeInfo or JsonSerializerContext, or make sure all of the required types are preserved.
ILLink: Trim analysis warning IL2026: System.Text.Json.Serialization.Metadata.JsonTypeInfo: Using method 'System.Text.Json.Serialization.Metadata.JsonTypeInfo.JsonTypeInfo(Type,JsonSerializerOptions)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. JSON serialization and deserialization might require types that cannot be statically analyzed. Use the overload that takes a JsonTypeInfo or JsonSerializerContext, or make sure all of the required types are preserved.
[...]

PR #2124, opened 2021-06-30
Don't warn for RUC from DebuggerDisplayAttribute

sbomer:
Fixes https://github.com/mono/linker/issues/2122


agocke (approved):
LGTM. I think this is a corner case that is probably safe not to warn about

Issue #1821, opened 2021-02-10
Preserving operator methods necessary for System.Linq.Expressions


eerhardt:
### Background

There are currently 2 ILLink warnings left in System.Linq.Expressions that both boil down to the same pattern.

When you create a `Binary` or `Unary` Expression, and don't specify a `MethodInfo`, `System.Linq.Expressions` will try to find an operator on the Type corresponding to the ExpressionType you specified.

For example, if you call `Expression.Equal(left, right)`, will call the [following code](https://github.com/dotnet/runtime/blob/6885777ebbd906bb4e2a306635bb028004764eaa/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/BinaryExpression.cs#L948-L955)

```C#
        public static BinaryExpression Equal(Expression left, Expression right, bool liftToNull, MethodInfo? method)
        {
            ExpressionUtils.RequiresCanRead(left, nameof(left));
            ExpressionUtils.RequiresCanRead(right, nameof(right));
            if (method == null)
            {
                return GetEqualityComparisonOperator(ExpressionType.Equal, "op_Equality", left, right, liftToNull);
            }
```

[`GetEqualityComparisonOperator`](https://github.com/dotnet/runtime/blob/6885777ebbd906bb4e2a306635bb028004764eaa/src/libraries/System.Linq.Expressions/src/System/Linq/Expressions/BinaryExpression.cs#L1048) will look for the `op_Equality` operator on both the `left` and `right` types. This cannot be expressed in the existing trimming annotations because we don't have a `System.Type` parameter to annotate. `left` and `right` are `System.Linq.Expressions.Expression` instances.

This same pattern applies to pretty much all Binary and Unary expressions:

* Equal
* NotEqual
* Add
* Subtract
* Divide
* Modulo
* Multiply
* Power
* LeftShift
* RightShift
* And
* Or
* Not
* Convert
* Increment
* Decrement
* etc

### Proposal

We should add a new rule in the `mono/linker` that when `System.Linq.Expressions` is used in an app, it should preserve all user defined operators on Types that are being preserved. This will allow the above Expression code to keep working in a trimmed application.

In the `System.Linq.Expressions` library, we can then suppress the remaining warnings saying "The trimmer won't trim operators on preserved Types when Expressions are involved. It is safe to try to find these operator methods."

cc @vitek-karas @MichalStrehovsky @marek-safar

MichalStrehovsky:
[...]

PR #2125, opened 2021-06-30
Preserve custom operators

sbomer:
This will keep custom operators on marked types whenever System.Linq.Expressions
is used, and the operator input types are marked.

The behavior is enabled by default, and can be disabled by passing
--disable-operator-discovery.

Addresses https://github.com/mono/linker/issues/1821

### DIFF
--- a/src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
+++ b/src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
@@ -11,14 +11,14 @@ namespace Mono.Linker.Steps
 	public class DiscoverOperatorsHandler : IMarkHandler
 	{
 		LinkContext _context;
-		bool seenLinqExpressions;
+		bool _seenLinqExpressions;
 		readonly HashSet<TypeDefinition> _trackedTypesWithOperators;
-		Dictionary<TypeDefinition, HashSet<MethodDefinition>> _pendingOperatorsForType;
+		Dictionary<TypeDefinition, List<MethodDefinition>> _pendingOperatorsForType;
 
-		Dictionary<TypeDefinition, HashSet<MethodDefinition>> PendingOperatorsForType {
+		Dictionary<TypeDefinition, List<MethodDefinition>> PendingOperatorsForType {
 			get {
 				if (_pendingOperatorsForType == null)
-					_pendingOperatorsForType = new Dictionary<TypeDefinition, HashSet<MethodDefinition>> ();
+					_pendingOperatorsForType = new Dictionary<TypeDefinition, List<MethodDefinition>> ();
 				return _pendingOperatorsForType;
 			}
 		}
@@ -41,8 +41,8 @@ namespace Mono.Linker.Steps
 			// Check for custom operators and either:
 			// - mark them, if Linq.Expressions was already marked, or
 			// - track them to be marked in case Linq.Expressions is marked later
-			var hasOperators = ProcessCustomOperators (type, mark: seenLinqExpressions);
-			if (!seenLinqExpressions) {
+			var hasOperators = ProcessCustomOperators (type, mark: _seenLinqExpressions);
+			if (!_seenLinqExpressions) {
 				if (hasOperators)
 					_trackedTypesWithOperators.Add (type);
 				return;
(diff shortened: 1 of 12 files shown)


src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
```diff
@@ -0,0 +1,180 @@
...
+					pendingOperators = new HashSet<MethodDefinition> ();
+					_pendingOperatorsForType.Add (otherType, pendingOperators);
+				}
+				pendingOperators.Add (method);
+			}
+			return hasCustomOperators;
+		}
+
+		TypeDefinition _int32;
+		TypeDefinition Int32 {
```
marek-safar, line 91: This should not be necessary as you can use much cheaper `MetadataType` on TypeReference and compare with `MetadataType.Int32`

src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
```diff
@@ -0,0 +1,180 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class DiscoverOperatorsHandler : IMarkHandler
+	{
+		LinkContext _context;
```
marek-safar, line 12: a few could be reasonly

src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
```diff
@@ -0,0 +1,180 @@
...
+		{
+			_context = context;
+			_trackedTypesWithOperators = new HashSet<TypeDefinition> ();
+			_pendingOperatorsForType = new Dictionary<TypeDefinition, HashSet<MethodDefinition>> ();
+			markContext.RegisterMarkTypeAction (ProcessType);
+		}
+
+		void ProcessType (TypeDefinition type)
+		{
+			CheckForLinqExpressions (type);
```
marek-safar, line 27: Shouldn't this return if the type is not SLE?

sbomer: It shouldn't return but it can skip some logic - good catch. Fixed.

src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
sbomer: _context can't be since it's not passed to the ctor, but I made the HashSet readonly. I changed the Dictionary to be lazily initialized.

src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
sbomer: Fixed. I am still using similar logic for Nullable<T> - is that ok?

marek-safar: Yep, for nullable this makes sense

src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using Mono.Cecil;
+
+namespace Mono.Linker.Steps
+{
+	public class DiscoverOperatorsHandler : IMarkHandler
+	{
+		LinkContext _context;
+		bool seenLinqExpressions;
```
vitek-karas, line 14:
```suggestion
		bool _seenLinqExpressions;
```

src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
```diff
@@ -0,0 +1,215 @@
...
+		readonly HashSet<TypeDefinition> _trackedTypesWithOperators;
+		Dictionary<TypeDefinition, HashSet<MethodDefinition>> _pendingOperatorsForType;
```
vitek-karas, line 16: I would use `Dictionary<TypeDefinition, List<MethodDefinition>>` as far as I can tell we don't do `Contains` checks on the value collection, we only enumerate over it. So hash set is not appropriate (and a tiny bit more expensive). And we should never try to add a duplicate to that list since we only process each type once.

vitek-karas (approved):

PR #2129, opened 2021-07-02
Check target of module/assembly level suppressions

mateoatr:
The `Scope` parameter is only relevant in the context of global suppressions (put in a module or an assembly); and the only scopes that we support are `module`, `type` and `member`. Whenever the scope is set to a value other than these three, we log a message stating that the used argument is not supported -- this was changed to log a warning instead.

Unconditional suppressions put on a module/assembly that specify a target are registered as if they were put on the target (e.g., on a type, field, etc.); thus, when looking if a warning is suppressed on a global level, we must make sure to only look in global suppressions that do not specify a target -- this was fixed.

### DIFF
--- a/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
+++ b/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
@@ -173,14 +173,17 @@ namespace Mono.Linker
 				if (!TryDecodeSuppressMessageAttributeData (instance, out info))
 					continue;
 
-				// If the scope is missing we treat the suppression as if it was placed on the module.
 				var scope = info.Scope?.ToLower ();
-				if (scope == "module" || scope == null) {
+				if (info.Target == null && (scope == "module" || scope == null)) {
 					AddSuppression (info, provider);
 					continue;
 				}
 
 				switch (scope) {
+				case "module":
+					AddSuppression (info, provider);
+					break;
+
 				case "type":
 				case "member":
 					foreach (var result in DocumentationSignatureParser.GetMembersForDocumentationSignature (info.Target, module))
@@ -188,7 +191,8 @@ namespace Mono.Linker
 
 					break;
 				default:
-					_context.LogWarning ($"Invalid scope '{info.Scope}' used in 'UnconditionalSuppressMessageAttribute' on module '{module.Name}'.",
+					_context.LogWarning ($"Invalid scope '{info.Scope}' used in 'UnconditionalSuppressMessageAttribute' on module '{module.Name}' " +
+						$"with target '{info.Target}'.",
 						2108, _context.GetAssemblyLocation (module.Assembly));
 					break;
 				}
--- a/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
+++ b/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
@@ -175,7 +175,7 @@ namespace Mono.Linker
 					continue;
 
 				// If the scope is missing we treat the suppression as if it was placed on the module.
-				var scope = info.Scope ?.ToLower ();
+				var scope = info.Scope?.ToLower ();
 				if (info.Target == null && (scope == "module" || scope == null)) {
 					AddSuppression (info, provider);
 					continue;
--- a/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
+++ b/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
@@ -90,7 +90,8 @@ namespace Mono.Linker
 				return false;
 
 			return _suppressions.TryGetValue (provider, out var suppressions) &&
-				suppressions.TryGetValue (id, out info);
+				suppressions.TryGetValue (id, out info) &&
+				(provider is ModuleDefinition || provider is AssemblyDefinition ? info.Target == null : true);
 		}
 
 		static bool TryDecodeSuppressMessageAttributeData (CustomAttribute attribute, out SuppressMessageInfo info)
@@ -120,7 +121,7 @@ namespace Mono.Linker
 				foreach (var p in attribute.Properties) {
 					switch (p.Name) {
 					case ScopeProperty:
-						info.Scope = (p.Argument.Value as string)?.ToLower ();
+						info.Scope = p.Argument.Value as string;
 						break;
 					case TargetProperty:
 						info.Target = p.Argument.Value as string;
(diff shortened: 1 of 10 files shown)


src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -173,23 +174,26 @@ public void LookForModuleLevelAndGlobalSuppressions (ModuleDefinition module, IC
...
 
-				if (info.Target == null || info.Scope == null)
+				if (info.Target == null || scope == null)
 					continue;
 
-				switch (info.Scope) {
+				switch (scope) {
 				case "type":
 				case "member":
 					foreach (var result in DocumentationSignatureParser.GetMembersForDocumentationSignature (info.Target, module))
```
sbomer, line 190: Should we be validating that the scope kind matches the target, and warning if you for example use scope "type" but specify a member signature as the target?

I also think we should make it a warning if the target doesn't match anything.

mateoatr: Feels a little bit as an overkill since we'll end up running the DocumentationSignatureParse, but we can add the check. Related to this, I opened https://github.com/mono/linker/issues/2130.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -173,23 +174,26 @@ public void LookForModuleLevelAndGlobalSuppressions (ModuleDefinition module, IC
...
+				// If the scope is missing we treat the suppression as if it was placed on the module.
+				var scope = info.Scope?.ToLower ();
+				if (info.Target == null && (scope == "module" || scope == null)) {
 					AddSuppression (info, provider);
 					continue;
 				}
 
-				if (info.Target == null || info.Scope == null)
+				if (info.Target == null || scope == null)
 					continue;
```
sbomer, line 185: Should null Target or Scope be warnings too?

mateoatr: I don't see why we should. Both parameters are only relevant for global suppressions, which should always look like `[assembly: Unc...`/`[module: Unc...`, and then I think it's reasonable to put the suppression on the module whenever neither are specified. Having a non-module scope without a target could be a case where I think a warning makes sense, e.g. `[module: USM("", "", Scope = "type")]`, I haven't tried this but I think this would actually throw from DocumentationSignatureParser :/

eerhardt:
> Unconditional suppressions put on a module/assembly that specify a target are registered as if they were put on the target (e.g., on a type, field, etc.); thus, when looking if a warning is suppressed on a global level, we must make sure to only look in global suppressions that do not specify a target 

I wonder if we really need to support global suppressions for a whole assembly. It feels like too easy of a trap to fall into.

Maybe if we need to support this, we could write a special rule into the Roslyn analyzer that warns for `UnconditionalSuppressMessage` attributes at the global level that don't have `Target` specified. Then someone could suppress that warning with a `#pragma` if they really, really wanted to suppress a warning across a whole assembly.

vitek-karas:
> Maybe if we need to support this, we could write a special rule into the Roslyn analyzer that warns for UnconditionalSuppressMessage attributes at the global level that don't have Target specified. Then someone could suppress that warning with a #pragma if they really, really wanted to suppress a warning across a whole assembly.

I like this - also as a way to gather more data - if there are actual use cases for assembly level suppressions, we can always just remove this analyzer warning (or scope it better).

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
vitek-karas: We should add tests for all these cases - to validate the current behavior. And if some of those behaviors seem broken, either fix it or file new bugs. But have them "verified" by tests regardless.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -90,7 +90,8 @@ bool IsSuppressedOnElement (int id, ICustomAttributeProvider provider, out Suppr
 				return false;
 
 			return _suppressions.TryGetValue (provider, out var suppressions) &&
-				suppressions.TryGetValue (id, out info);
+				suppressions.TryGetValue (id, out info) &&
+				(provider is ModuleDefinition || provider is AssemblyDefinition ? info.Target == null : true);
```
vitek-karas, line 94: Why is this necessary - the code in `LookForModuleLevelAndGlobalSuppressions` will only add the suppression to the list if
* Target is null and provider is module or assembly
* Target is not null, but then provider is not module or assembly

In which case do we need this additional condition?

mateoatr: Good call. I was hitting this whenever a global suppression was being added from `MarkStep` https://github.com/mono/linker/blob/41fd293dda27c099b063f51d8917e607271a508e/src/linker/Linker.Steps/MarkStep.cs#L811-L812
Changed things there so that that path isn't followed for global suppressions, otherwise we'd hit this (having suppressions where the provider is a module/assembly and it has a target).

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: Added some tests and changed the DocumentationSignatureParser so that we don't throw when `module` or `id` are null.

vitek-karas (approved):
Looks good to me.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -173,23 +173,23 @@ public void LookForModuleLevelAndGlobalSuppressions (ModuleDefinition module, IC
 				if (!TryDecodeSuppressMessageAttributeData (instance, out info))
 					continue;
 
-				if (info.Target == null && (info.Scope == "module" || info.Scope == null)) {
+				// If the scope is missing we treat the suppression as if it was placed on the module.
+				var scope = info.Scope?.ToLower ();
+				if (scope == "module" || scope == null) {
 					AddSuppression (info, provider);
```
sbomer, line 179: This looks like it will make it easy to accidentally suppress all warnings in a module if you specify a target but forget the scope - we should add a test for this.

mateoatr: Yup, it will follow the same path as TargettedModuleSuppressionWithUnmatchedScope.cs and suppress all warnings in the module. I'll add another test to make this explicit in the case where a scope isn't passed.

sbomer: I think it would be worth making this a warning instead. If I add:
```csharp
[module: UnconditionalSuppressMessage ("Test", "IL2026", Target ="M:MyLib.Program.MyDangerousMethod")]
```
That looks like it suppresses warnings only on `MyDangerousMethod`. Because it does suppress the warning from that method, it's all too easy to assume that it works as expected and leave it as-is, hiding future warnings if the author introduces new RUC code.

Before this change, the suppression would at least be ignored. If you'd rather scope the warning behavior out of this change, we should continue ignoring such suppressions, and track adding a warning for it (I think it's not covered by the existing issues, but I might have missed it).

mateoatr: > Because it does suppress the warning from that method, it's all too easy to assume that it works as expected and leave it as-is, hiding future warnings if the author introduces new RUC code.
>
> Before this change, the suppression would at least be ignored.

Ah, yeah. Reverted -- we'll keep ignoring these scenarios.

sbomer (approved):
Thank you!

PR #2133, opened 2021-07-06
Enhance NestedTypes in DynamicallyAccessedMemberTypes

### LINKED ISSUE #1174: DynamicallyAccessMemberType.PublicNestedTypes behavior discussion

LakshanF:
We don't do much for nested type options in `DynamicallyAccessedMemberTypes`, forcing the much broader scope `All `to be used if we want to preserve members inside nested types. This fixes by keeping all the members for nested types

Fixes #1174

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Reflection/MembersUsedViaReflection.cs
+++ b/test/Mono.Linker.Tests.Cases/Reflection/MembersUsedViaReflection.cs
@@ -138,7 +138,9 @@ namespace Mono.Linker.Tests.Cases.Reflection
 			[Kept]
 			public static class PublicNestedType
 			{
+				[Kept]
 				public static int _nestedPublicField;
+				[Kept]
 				public static void NestedPublicMethod ()
 				{ }
 			}
@@ -193,7 +195,9 @@ namespace Mono.Linker.Tests.Cases.Reflection
 			public static class PublicNestedType
 			{
 				// PublicNestedType should be kept but linker won't mark anything besides the declaration
+				[Kept]
 				public static int _nestedPublicField;
+				[Kept]
 				public static void NestedPublicMethod ()
 				{ }
 			}
@@ -258,7 +262,9 @@ namespace Mono.Linker.Tests.Cases.Reflection
 			public static class PublicNestedType
 			{
 				// PublicNestedType should be kept but linker won't mark anything besides the declaration
+				[Kept]
 				public static int _nestedPublicField;
+				[Kept]
 				public static void NestedPublicMethod ()
 				{ }
 			}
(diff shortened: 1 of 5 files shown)


sbomer:
Should this really keep everything on the nested types? I would not expect `PublicNestedTypes` to recursively keep private nested types for example.

LakshanF:
I was primarily going with the discussion on the issue for the fix from @MichalStrehovsky and the view that there are not wide usage of `DAM `with `NestedTypes `in the library

I looked into this change since I was curious if this change would have an impact on the library size due to `EventSource` proliferations in our library but the size wins are minimal since we need to keep most of parent type members. But created the PR since the consensus in the issue was to fix it.

The trimmer is not able to reason on what is needed for internal types that are decorated with `DAM `in the library mode since the programming paradigm relies on reflection in application runtime to enable the event source. I think the solution is for custom library source code such as `NetEventSource` to be refactored to use only what is needed instead of the current model. The runtime [issue ](https://github.com/dotnet/runtime/issues/54859) tracks this but that work might not happen in this release.

vitek-karas (approved):
Assuming we agree on the behavior (as per the discussion on the issue) the change looks good.
This should be followed by changing the intrinsic behavior of `GetNestedType` which currently only propagates the `All` annotation to the return value if the source type was annotated with `All`. With this change, it can now propagate `All` always. In fact we may want to change the runtime and add the annotation on the return value directly.

MichalStrehovsky (approved):
Looks good modulo Vitek's comment about the GetNestedType intrinsic.

Sven is right that we might be able to not mark non-public nested types in the deeper levels, but I'm not sure if it will result in savings in practice and it looks like it would complicate the implementation quite a bit.

```csharp
class Foo
{
    public class PublicNestedAndReflected
    {
        protected class ProtectedNestedButNotReflected { }
    }
}
```

If we ask for all public nested types on Foo, we might be able to get rid of ProtectedNestedButNotReflected. But I think it's very unlikely that a fully-preserved PublicNestedType would have no references to ProtectedNestedButNotReflected (not many other classes have access to this nested class since it's not public - chances are pretty high we would end up keeping it anyway).

test/Mono.Linker.Tests.Cases/Reflection/MembersUsedViaReflection.cs
```diff
@@ -313,7 +319,9 @@ private void PrivateMethod ()
 			public static class PublicNestedType
 			{
 				// PublicNestedType should be kept but linker won't mark anything besides the declaration
```
tlakollo, line 321: Nit: We should delete these comments in our testing given that now the link will mark more than the declaration

PR #2137, opened 2021-07-08
Move EventSource special handling to library mode

LakshanF:
Trimmer cannot remove special handling of `EventSource` without some work from the library side due to some [size concerns](https://github.com/dotnet/runtime/issues/54859). This fix moves the `EventSource` special handling to library mode since the general case will be handled via the type hierarchy support feature.


LakshanF:
Will move this to the #2114 since that makes more sense as suggested by https://github.com/mono/linker/pull/2114#issuecomment-876634420

Issue #2138, opened 2021-07-09
Broken type references when using copyused mode


vitek-karas:
Repro is pretty simple:

New console app:
```C#
using System;
using System.Runtime.InteropServices;

namespace ReproCopyUsedBug
{
    class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine(RuntimeInformation.FrameworkDescription);
            Console.WriteLine(RuntimeEnvironment.GetRuntimeDirectory());
        }
    }
}
```

Publish with:
```console
dotnet publish -r win-x64 -p:PublishTrimmed=true -p:TrimMode=copyused
```

Try to run:
```console
Unhandled exception.
   Cannot print exception string because Exception.ToString() failed.
```

The problem is in `System.Runtime.InteropServices.RuntimeInformation.dll`, the `RuntimeInformation.get_FrameworkDescription` calls `CustomAttributeExtensions.GetCustomAttribute<T>(...)`. The type ref on that dll is pointing to `System.Runtime.dll`, which has a type forwarder for it which points to `CoreLib`.

When trimmed in copyused mode, we rewrite the typereference to point directly to `CoreLib` and remove the typeforwarder from `System.Runtime.dll`, but then when deciding how to output the `System.Runtime.InteropServices.RuntimeInformation` we mark it as "copy" and effectively throw away any changes (including the rewrite of the type ref).

Note that this works if you omit the call to `RuntimeEnvironment` since now we will remove `System.Runtime.InteropServices.dll` completely, which will force us to rewrite `System.Runtime.InteropServices.RuntimeInformation` as that has an assembly ref to it.

vitek-karas:
@sbomer, can you please take a look? This is currently blocking @odhanson as this is a regression from .NET 6 Preview 3 (where it worked).

vitek-karas:
Also - any idea about a potential workaround... ?

odhanson:
Thanks @vitek-karas

sbomer:
I don't have any ideas for a workaround unfortunately.

Thanks for the repro. The copyused behavior has always seemed inconsistent to me - sometimes we would copy, sometimes we would rewrite their typerefs, and the decision was based on whether the assembly referenced any removed assembly. But it's possible that the assembly doesn't reference any removed assembly, yet still references a removed forwarder in a kept assembly.

I can see two possible fixes:
1. Always save copyused assemblies
[...]

PR #2140, opened 2021-07-10
Fix writing of updated copyused scopes

sbomer:
Fixes https://github.com/mono/linker/issues/2138

### DIFF
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -65,11 +65,14 @@ namespace Mono.Linker.Steps
 
 			// Update scopes before removing any type forwarder.
 			foreach (var assembly in assemblies) {
-				switch (Annotations.GetAction (assembly)) {
+				var action = Annotations.GetAction (assembly);
+				switch (action) {
 				case AssemblyAction.CopyUsed:
 				case AssemblyAction.Link:
 				case AssemblyAction.Save:
-					SweepAssemblyReferences (assembly);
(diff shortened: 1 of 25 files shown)


vitek-karas (approved):
Looks good to me - @mateoatr please take a look as well (since you actually know the copyused better than me)

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -696,12 +711,17 @@ void UpdateTypeScope (Collection<ExportedType> forwarders)
 					if (td == null) {
 						// Forwarded type cannot be resolved but it was marked
 						// linker is running in --skip-unresolved true mode
+						var anr = (AssemblyNameReference) f.Scope;
+						f.Scope = importer.ImportReference (anr);
```
mateoatr, line 715: Why is this needed?

sbomer: This was for consistency with how we treat unresolved typereferences here:
https://github.com/mono/linker/blob/73fbf96cb452f291cceba684034b8548d54d82eb/src/linker/Linker.Steps/SweepStep.cs#L908-L918

I think continue is right since we may be able to resolve other forwarders if not this one.

mateoatr (approved):

odhanson:
@vitek-karas @sbomer are we planning to merge this into preview 7 ?

vitek-karas:
As far as I can tell this is in Preview 7.

Issue #2136, opened 2021-07-07
Unnecessary warnings for DiagnosticSourceEventSource


eerhardt:
`dotnet publish` the following application:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
    <RuntimeIdentifier>win-x64</RuntimeIdentifier>
    <PublishTrimmed>true</PublishTrimmed>
  </PropertyGroup>

</Project>
```

```C#
using System;
using System.Xml;

class Program
{
    static void Main(string[] args)
    {
        XmlDocument doc = new();
        doc.LoadXml("<hello />");
        Console.WriteLine(doc.InnerXml);
    }
}
```

### Expected results
There shouldn't be any ILLink warnings, since I am not using code that is unsafe.

### Actual results

```
ILLink : Trim analysis warning IL2026: System.Diagnostics.DiagnosticSourceEventSource: Using method 'System.Diagnostics.DiagnosticSourceEventSource.TransformSpec.Morph(Object)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The type of object being written to DiagnosticSource cannot be discovered statically. [C:\DotNetTest\HelloWorld\HelloWorld.csproj]
ILLink : Trim analysis warning IL2026: System.Diagnostics.DiagnosticSourceEventSource: Using method 'System.Diagnostics.DiagnosticSourceEventSource.TransformSpec.PropertySpec.Fetch(Object)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The type of object being written to DiagnosticSource cannot be discovered statically. [C:\DotNetTest\HelloWorld\HelloWorld.csproj]
ILLink : Trim analysis warning IL2026: System.Diagnostics.DiagnosticSourceEventSource: Using method 'System.Diagnostics.DiagnosticSourceEventSource.FilterAndTransform.Morph(Object)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The type of object being written to DiagnosticSource cannot be discovered statically. [C:\DotNetTest\HelloWorld\HelloWorld.csproj]
[...]

PR #2143, opened 2021-07-13
Ruc on classes

### LINKED ISSUE #1742: Allow RequiresUnreferencedCode on classes and structs
Some types are inherently not trim compatible. Take, for example, `BinaryFormatter`. Virtually anything you do with the type is not going to be trim compatible.

Today, we only allow `RequiresUnreferencedCodeAttribute` on methods and constructors.

https://github.com/dotnet/runtime/blob/d2e1a86434d1c0b74d75604b4cf9fdca100108c3/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/RequiresUnreferencedCodeAttribute.cs#L14-L15

When trying to annotate the implementation of `BinaryFormatter`, it is causing almost every method in the library to be marked as `RequiresUnreferencedCode`. This is a lot of changes when just putting the attribute on a whole class would be a lot easier.

We should change the ILLinker to support `RequiresUnreferencedCodeAttribute` on classes and structs, and possibly interfaces.

@agocke @vitek-karas @MichalStrehovsky

tlakollo:
Add support for RequiresUnreferencedCode on classes
 Added a cache to fast lookup for RequiresUnreferencedCode on the type hierarchy
 Added special case for warning in constructors while marking types
 Added logic to handle suppressions of other trim analysis warnings when the declaring type has RUC
 Added tests

Fixes #1742

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1754,6 +1754,11 @@ namespace Mono.Linker.Steps
 			foreach (Action<TypeDefinition> handleMarkType in _markContext.MarkTypeActions)
 				handleMarkType (type);
 
+			if (type.BaseType != null &&
+				!_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (type) &&
+				_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (_context.TryResolve (type.BaseType)))
+				CheckAndReportRequiresUnreferencedCode (type);
+
 			MarkType (type.BaseType, new DependencyInfo (DependencyKind.BaseType, type));
 
 			// The DynamicallyAccessedMembers hiearchy processing must be done after the base type was marked
@@ -2785,15 +2790,32 @@ namespace Mono.Linker.Steps
 
 			IMemberDefinition suppressionContextMember = currentOrigin.SuppressionContextMember;
 			if (suppressionContextMember != null &&
-				Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (suppressionContextMember))
+				(Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (suppressionContextMember) ||
+				(suppressionContextMember.DeclaringType != null &&
+				Annotations.HasRequiresUnreferencedCodeOnTypeHyerarchy (suppressionContextMember.DeclaringType))))
 				return true;
 
 			IMemberDefinition originMember = currentOrigin.MemberDefinition;
 			if (suppressionContextMember != originMember && originMember != null &&
-				Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (originMember))
+				(Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (originMember) ||
+				(originMember.DeclaringType != null &&
+				Annotations.HasRequiresUnreferencedCodeOnTypeHyerarchy (originMember.DeclaringType))))
 				return true;
 
 			return false;
+		}
+
+		internal void CheckAndReportRequiresUnreferencedCode (TypeDefinition type)
+		{
+			var currentOrigin = _scopeStack.CurrentScope.Origin;
+
+			// If the caller of a type is already marked with `RequiresUnreferencedCodeAttribute` a new warning should not
+			// be produced for the callee.
+			if (ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ())
+				return;
+
+			if (Annotations.TryGetLinkerAttribute (_context.TryResolve (type.BaseType), out RequiresUnreferencedCodeAttribute requiresUnreferencedCodeOnTypeHierarchy))
+				ReportRequiresUnreferencedCode (type.BaseType.GetDisplayName (), requiresUnreferencedCodeOnTypeHierarchy, currentOrigin);
 		}
 
 		internal void CheckAndReportRequiresUnreferencedCode (MethodDefinition method)
@@ -2805,13 +2827,24 @@ namespace Mono.Linker.Steps
 			if (ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ())
 				return;
 
-			if (Annotations.TryGetLinkerAttribute (method, out RequiresUnreferencedCodeAttribute requiresUnreferencedCode)) {
-				string formatString = SharedStrings.RequiresUnreferencedCodeMessage;
-				string arg1 = MessageFormat.FormatRequiresAttributeMessageArg (requiresUnreferencedCode.Message);
-				string arg2 = MessageFormat.FormatRequiresAttributeUrlArg (requiresUnreferencedCode.Url);
-				string message = string.Format (formatString, method.GetDisplayName (), arg1, arg2);
-				_context.LogWarning (message, 2026, currentOrigin, MessageSubCategory.TrimAnalysis);
+			if (method.IsStatic || method.IsConstructor) {
+				if (Annotations.TryGetRequiresUnreferencedCodeAttributeOnTypeHierarchy (method.DeclaringType, out RequiresUnreferencedCodeAttribute requiresUnreferencedCodeOnTypeHierarchy) && requiresUnreferencedCodeOnTypeHierarchy != null) {
+					ReportRequiresUnreferencedCode (method.GetDisplayName (), requiresUnreferencedCodeOnTypeHierarchy, currentOrigin);
+					return;
+				}
 			}
+
+			if (Annotations.TryGetLinkerAttribute (method, out RequiresUnreferencedCodeAttribute requiresUnreferencedCode))
+				ReportRequiresUnreferencedCode (method.GetDisplayName (), requiresUnreferencedCode, currentOrigin);
+		}
+
+		private void ReportRequiresUnreferencedCode (string displayName, RequiresUnreferencedCodeAttribute requiresUnreferencedCode, MessageOrigin currentOrigin)
+		{
+			string formatString = SharedStrings.RequiresUnreferencedCodeMessage;
+			string arg1 = MessageFormat.FormatRequiresAttributeMessageArg (requiresUnreferencedCode.Message);
+			string arg2 = MessageFormat.FormatRequiresAttributeUrlArg (requiresUnreferencedCode.Url);
+			string message = string.Format (formatString, displayName, arg1, arg2);
+			_context.LogWarning (message, 2026, currentOrigin, MessageSubCategory.TrimAnalysis);
 		}
 
 		protected (MethodReference, DependencyInfo) GetOriginalMethod (MethodReference method, DependencyInfo reason)
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2792,14 +2792,14 @@ namespace Mono.Linker.Steps
 			if (suppressionContextMember != null &&
 				(Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (suppressionContextMember) ||
 				(suppressionContextMember.DeclaringType != null &&
-				Annotations.HasRequiresUnreferencedCodeOnTypeHyerarchy (suppressionContextMember.DeclaringType))))
+				Annotations.HasRequiresUnreferencedCodeOnTypeHierarchy (suppressionContextMember.DeclaringType))))
 				return true;
 
 			IMemberDefinition originMember = currentOrigin.MemberDefinition;
 			if (suppressionContextMember != originMember && originMember != null &&
 				(Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (originMember) ||
 				(originMember.DeclaringType != null &&
-				Annotations.HasRequiresUnreferencedCodeOnTypeHyerarchy (originMember.DeclaringType))))
+				Annotations.HasRequiresUnreferencedCodeOnTypeHierarchy (originMember.DeclaringType))))
 				return true;
 
 			return false;
--- a/src/linker/Linker/Annotations.cs
+++ b/src/linker/Linker/Annotations.cs
@@ -58,6 +58,7 @@ namespace Mono.Linker
 		protected readonly Dictionary<IMemberDefinition, List<MethodDefinition>> preserved_methods = new Dictionary<IMemberDefinition, List<MethodDefinition>> ();
 		protected readonly HashSet<IMetadataTokenProvider> public_api = new HashSet<IMetadataTokenProvider> ();
 		protected readonly Dictionary<AssemblyDefinition, ISymbolReader> symbol_readers = new Dictionary<AssemblyDefinition, ISymbolReader> ();
+		readonly Dictionary<TypeDefinition, Attribute> ruc_on_type_hierarchy = new Dictionary<TypeDefinition, Attribute> ();
 		readonly Dictionary<IMemberDefinition, LinkerAttributesInformation> linker_attributes = new Dictionary<IMemberDefinition, LinkerAttributesInformation> ();
 		readonly Dictionary<object, Dictionary<IMetadataTokenProvider, object>> custom_annotations = new Dictionary<object, Dictionary<IMetadataTokenProvider, object>> ();
 		protected readonly Dictionary<AssemblyDefinition, HashSet<EmbeddedResource>> resources_to_remove = new Dictionary<AssemblyDefinition, HashSet<EmbeddedResource>> ();
@@ -550,6 +551,31 @@ namespace Mono.Linker
 		public bool SetPreservedStaticCtor (TypeDefinition type)
 		{
 			return marked_types_with_cctor.Add (type);
+		}
+
+		public bool HasRequiresUnreferencedCodeOnTypeHyerarchy (TypeDefinition type)
+		{
+			if (!ruc_on_type_hierarchy.TryGetValue (type, out var hasRucOnTypeHierarchy))
+				return TryGetRequiresUnreferencedCodeAttributeOnTypeHierarchy (type, out var _);
+			return hasRucOnTypeHierarchy is not null;
+		}
+
+		public bool TryGetRequiresUnreferencedCodeAttributeOnTypeHierarchy (TypeDefinition type, out RequiresUnreferencedCodeAttribute attribute)
+		{
+			if (!ruc_on_type_hierarchy.TryGetValue (type, out var output)) {
+				if (TryGetLinkerAttribute (type, out RequiresUnreferencedCodeAttribute currentTypeAttribute)) {
+					ruc_on_type_hierarchy.Add (type, currentTypeAttribute);
+					attribute = currentTypeAttribute;
+					return true;
+				} else if (type.DeclaringType != null && TryGetRequiresUnreferencedCodeAttributeOnTypeHierarchy (type.DeclaringType, out var declaringTypeAttribute)) {
+					ruc_on_type_hierarchy.Add (type, declaringTypeAttribute);
+					attribute = declaringTypeAttribute;
+					return true;
+				} else
+					ruc_on_type_hierarchy.Add (type, null);
+			}
+			attribute = (RequiresUnreferencedCodeAttribute) output;
+			return attribute is not null;
 		}
 
 		public bool HasLinkerAttribute<T> (IMemberDefinition member) where T : Attribute
--- a/src/linker/Linker/Annotations.cs
+++ b/src/linker/Linker/Annotations.cs
@@ -553,7 +553,7 @@ namespace Mono.Linker
 			return marked_types_with_cctor.Add (type);
 		}
 
-		public bool HasRequiresUnreferencedCodeOnTypeHyerarchy (TypeDefinition type)
+		public bool HasRequiresUnreferencedCodeOnTypeHierarchy (TypeDefinition type)
 		{
 			if (!ruc_on_type_hierarchy.TryGetValue (type, out var hasRucOnTypeHierarchy))
 				return TryGetRequiresUnreferencedCodeAttributeOnTypeHierarchy (type, out var _);
--- a/test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
+++ b/test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
@@ -79,6 +79,7 @@ namespace Mono.Linker.Tests.Cases.RequiresCapability
 			OnEventMethod.Test ();
 			AccessThroughNewConstraint.Test ();
 			AccessThroughLdToken.Test ();
+			RequiresOnClass.Test ();
 		}
 
 		[ExpectedWarning ("IL2026", "Message for --RequiresWithMessageOnly--.")]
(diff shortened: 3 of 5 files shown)


src/linker/Linker/Annotations.cs
```diff
@@ -552,6 +553,31 @@ public bool SetPreservedStaticCtor (TypeDefinition type)
 			return marked_types_with_cctor.Add (type);
 		}
 
+		public bool HasRequiresUnreferencedCodeOnTypeHierarchy (TypeDefinition type)
```
MichalStrehovsky, line 556: Why do we need to collect this on the type hierarchy?

I would expect:

```csharp
[RUC]
class Foo
{
}

class Bar : Foo
{
}
```

To just give me a warning when we're analyzing Bar's constructor (it's going to call into Foo's constructor and that one is RUC).

It's also how the C# compiler handles the `[Obsolete]` attribute. It doesn't auto propagate to derived classes.

The cache costs extra complexity and memory and I'm not sure it's giving us anything useful.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2805,13 +2827,24 @@ internal void CheckAndReportRequiresUnreferencedCode (MethodDefinition method)
 			if (ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ())
 				return;
 
-			if (Annotations.TryGetLinkerAttribute (method, out RequiresUnreferencedCodeAttribute requiresUnreferencedCode)) {
-				string formatString = SharedStrings.RequiresUnreferencedCodeMessage;
-				string arg1 = MessageFormat.FormatRequiresAttributeMessageArg (requiresUnreferencedCode.Message);
-				string arg2 = MessageFormat.FormatRequiresAttributeUrlArg (requiresUnreferencedCode.Url);
-				string message = string.Format (formatString, method.GetDisplayName (), arg1, arg2);
-				_context.LogWarning (message, 2026, currentOrigin, MessageSubCategory.TrimAnalysis);
+			if (method.IsStatic || method.IsConstructor) {
```
MichalStrehovsky, line 2830: We might need to think about how this applies to static constructors.

Is this going to warn anywhere?

```csharp
[RUC]
class Foo
{
    public static object Instance;

    static Foo()
    {
        Instance = Activator.CreateInstance(Type.GetType(WebClient.DownloadString("https//example.com/")));
    }
}

class Program
{
    static void Main() => Console.WriteLine(Foo.Instance.ToString());
}
```

The more I think about it, the more I feel we should disallow placing RUC on static constructors because it makes things _too damn complicated_.

src/linker/Linker/Annotations.cs
tlakollo: That's true, maybe I should have used other name for the functions and cache, but the current changes don't auto propagate between derived classes the cache is only going over nested types. For the scenario you mentioned there will only be a warning in `class Bar : Foo` because it calls Foo constructor just like [Obsolete] and is also one of the tests

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Given that we can now put RUC on the type itself - the value of supporting it on static .ctor seems questionable. The RUC on static .ctor should produce warnings in almost all the cases where RUC on type will (I know that there are some potential exceptions, where the runtime is allowed to call a static method without calling the static .ctor). From a UX perspective I don't see much of a difference.

I think it should be fine to remove support for RUC on static .ctor once we have this merged. Note that we should see how many cases in libraries use RUC on static .ctor today.

vitek-karas: I would expect the above to warn in `Main` - that said this is something which currently probably doesn't work and we'll have to add that (check for RUC on declaring type when accessing static fields). Obsolete already handles this as expected (access to any static member issues the warning).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1754,6 +1754,11 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 			foreach (Action<TypeDefinition> handleMarkType in _markContext.MarkTypeActions)
 				handleMarkType (type);
 
+			if (type.BaseType != null &&
+				!_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (type) &&
+				_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (_context.TryResolve (type.BaseType)))
+				CheckAndReportRequiresUnreferencedCode (type);
```
vitek-karas, line 1760: With the discussion from https://github.com/mono/linker/issues/2136 in mind - I think we should use a different warning code in this case. The reason is that otherwise it's tricky to suppress this warning correctly.

The warning should be reported on the derived type itself, so suppressing it there is the only viable option, but that also means that the suppression is effective for everything in the type. If this were just `IL2026` suppressing it would suppress all `IL2026` warnings in the type and in all its methods. We should use a new warning code just for this case so that suppressing it is more precise. It is still not 100% precise as it would suppress the warning on all nested types as well, but that's probably good enough.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2785,17 +2790,34 @@ internal bool ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ()
...
+		{
+			var currentOrigin = _scopeStack.CurrentScope.Origin;
+
+			// If the caller of a type is already marked with `RequiresUnreferencedCodeAttribute` a new warning should not
+			// be produced for the callee.
+			if (ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ())
+				return;
+
+			if (Annotations.TryGetLinkerAttribute (_context.TryResolve (type.BaseType), out RequiresUnreferencedCodeAttribute requiresUnreferencedCodeOnTypeHierarchy))
+				ReportRequiresUnreferencedCode (type.BaseType.GetDisplayName (), requiresUnreferencedCodeOnTypeHierarchy, currentOrigin);
```
vitek-karas, line 2818: I think the origin here is potentially wrong. The warning about the base type should be always reported on the derived type, but the scope stack will probably point to the location of a reference to the derived type.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2785,17 +2790,34 @@ internal bool ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ()
...
-				Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (originMember))
+				(Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (originMember) ||
+				(originMember.DeclaringType != null &&
+				Annotations.HasRequiresUnreferencedCodeOnTypeHierarchy (originMember.DeclaringType))))
 				return true;
 
 			return false;
 		}
 
+		internal void CheckAndReportRequiresUnreferencedCode (TypeDefinition type)
```
vitek-karas, line 2808: I would either inline this into `MarkType` (it's not that long) or rename it to make it very specific about base type handling.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2785,17 +2790,34 @@ internal bool ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ()
...
+		{
+			var currentOrigin = _scopeStack.CurrentScope.Origin;
+
+			// If the caller of a type is already marked with `RequiresUnreferencedCodeAttribute` a new warning should not
+			// be produced for the callee.
+			if (ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ())
```
vitek-karas, line 2814: This will probably also not handle suppressions correctly since the current scope will not be right.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -783,5 +784,142 @@ public static void Test ()
...
+				public void NonStaticMethod () { }
+
+				public class NestedClass
+				{
+					public static void NestedStaticMethod () { }
+				}
+
+				public static void TestSuppressions (Type[] types)
+				{
+					// RequiresUnreferencedCode in the class suppresses other RequiresUnreferencedCode messages
```
vitek-karas, line 806:
```suggestion
					// StaticMethod is a static method on a RUC annotated type, so it should warn. But RequiresUnreferencedCode in the class suppresses other RequiresUnreferencedCode messages
```

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1754,6 +1754,11 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 			foreach (Action<TypeDefinition> handleMarkType in _markContext.MarkTypeActions)
 				handleMarkType (type);
 
+			if (type.BaseType != null &&
+				!_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (type) &&
+				_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (_context.TryResolve (type.BaseType)))
```
sbomer, line 1759: I think this check needs to look for RUC not just on the type, but on declaring types in case they are nested. For example:

```csharp
[RUC]
public class Outer {
  public class Base {}
}

class Derived : Outer.Base {} // should warn
```

agocke: I think it's slightly trickier than that if we want to follow Obsolete:

```C#
using System;
[Obsolete]
public class C {
    public class D { }
    public void M() {
        var d = new D(); // no warning
    }
}

class E
{
    public static void M()
    {
        var d = new C.D(); // warning
    }
}
```

src/linker/Linker.Steps/MarkStep.cs
agocke: I agree that once we have `RUC` on types, it makes a lot less sense and is a lot more complicated to make it work for static constructors.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: I think we should use Obsolete as a good way to tell what is a good UX - and then ask if it's something which is reasonable to achieve in the linker/analyzer and if it's worth it.

This example from @agocke should already work that way, because the first call (where it should not warn) will be suppressed by the RUC being on the parent type.

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: Placing Obsolete on a static constructor doesn't seem to do anything. It's not a behavior we can be okay with for RUC.

I was more thinking along the lines of:

* RUC on a class doesn't apply to the cctor (it will not suppress anything within the cctor)
* Placing RUC on a cctor is always a warning

We cannot deprecate RUC on a cctor in favor of RUC on a class because we don't allow RUC on structs and interfaces but those can have cctors.

Allowing RUC on a cctor complicates things by _a lot_. I'm not sure we handle them:

```csharp
// Where do we warn?
((IFoo)new A()).Frob();

interface IFoo
{
    static IFoo
    {
        // This code will run when the default implementation on IFoo is used
    }
    void Frob() { }
}

interface IBar : IFoo
{
    static IBar()
    {
        // This will run when the default implementation is used
    }
    void IFoo.Frob() { }
}

class A : IFoo { }
```

```csharp
// Where do we warn?
_ = Foo.MyField;

// Where do we warn?
typeof(Foo).GetField("MyField").GetValue(null);

class Foo
{
    static Foo()
    {
        // Will run if a field is accessed
    }

    public static int MyField;
}
```

Of course there will be a customer who really really wants to put this on a cctor. We tell them to lazy-initialize using their own devices.

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: For this example, where do you expect to see the warning?

```csharp
[RUC]
public class Outer {
  public class Base {}
}

class Derived : Outer.Base {} // should warn
```

I would expect to see one for the Derived's constructor because it's calling the constructor on Outer.Base. Does that match your expectations?

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: > We cannot deprecate RUC on a cctor in favor of RUC on a class because we don't allow RUC on structs and interfaces but those can have cctors.

Why do you think it's important to support RUC on .cctor for struct/interface? I'm sure I can come up with a reason, but at the same time, I think I would be OK simply not supporting that at all.
I personally would go for the "simple" solution - RUC on classes (as per this PR), warn on RUC on .cctor (everywhere) and ignore it.

> RUC on a class doesn't apply to the cctor (it will not suppress anything within the cctor)

The only case I think of why this might be a good thing is if there's a case where the .cctor is executed by we won't be able to warn due to access to the type. But I can't think of such a case... (doesn't mean it's not there)

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: I would expect to see a warning that `Derived` derives from `Outer.Base` which is RUC. And ideally no other warnings (byt we might need to do some work to not generate more).

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: Yes, I'm trying to go for the simple solution as well - the above was my proposal for that:

* Disallow explicit RUC on cctor.
* RUC on a class doesn't imply RUC behavior on it's cctor.

vitek-karas: > RUC on a class doesn't imply RUC behavior on it's cctor.

I assume this means "don't do anything special about .cctor". I still don't understand why we would not want to auto-suppress warnings from within the .cctor if the type is RUC annotated (we will do that for all methods on the type).

MichalStrehovsky: > I assume this means "don't do anything special about .cctor". I still don't understand why we would not want to auto-suppress warnings from within the .cctor if the type is RUC annotated (we will do that for all methods on the type).

Because it's equivalent to allowing RUC on cctor itself and brings all the problems listed above (and more).

It would suppress the warning about Type.GetType here.

```csharp
[RUC]
class Foo
{
    public static object Instance;

    static Foo()
    {
        Instance = Activator.CreateInstance(Type.GetType(WebClient.DownloadString("https//example.com/")));
    }
}

class Program
{
    static void Main() => Console.WriteLine(Foo.Instance.ToString());
}
```

Here are two options to solve this:

* Don't let the class RUC apply to the static constructor. People get a warning in the static constructor and have to deal with it.
* Let the RUC apply to static constructor as well. Now ILLink is on the hook to analyze access to static fields as well. We also have to make sure warnings are generated when reflection-accessing those fields. Feels like a bunch of work.

agocke: Would you also not suppress warnings from static field initializers?

agocke: Even without `RUC` on static constructors, the `Obsolete` behavior is to warn on access to any static member, including a static field:

```C#
public class C {
    public void M() {
        int x = D.X; // warning
    }
}

[Obsolete]
public class D
{
    public static int X;
}
```

I don't think changing the static constructor behavior saves us from implementing support for checking static field access.

MichalStrehovsky: > I don't think changing the static constructor behavior saves us from implementing support for checking static field access.

Static field access doesn't _have to_ warn. RUC is different than Obsolete in the sense that we're declaring the class is doing something trimming incompatible. But field access is normally not trimming incompatible.

It's also tricky to map C#'s warning semantics to IL - Roslyn is actually always warning about the type - not about the member. That's why it also warns on a cast to the type, or warns when declaring a local. We're not going to warn on those I assume.

agocke: Agreed on the concept, but if the actual action that's creating the warning is accessing the type (which triggers running the static constructor), then the warning is in the appropriate place.

src/linker/Linker.Steps/MarkStep.cs
tlakollo: Now it warns in the following way
```C#
[RUC]
public class Outer {
  public class Base {}
}

// IL2109: Type 'Derived' derives from 'Outer.Base' which has 'RequiresUnreferencedCodeAttribute'.
class Derived : Outer.Base {}
```

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1771,6 +1766,22 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 			MarkCustomAttributes (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
 			MarkSecurityDeclarations (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
 
+			if (type.ToString ().Contains ("Requires"))
+				Debugger.Break ();
+
```
vitek-karas, line 1771: Debugging left-over :-)

src/linker/Linker/Annotations.cs
```diff
@@ -553,29 +552,45 @@ public bool SetPreservedStaticCtor (TypeDefinition type)
...
-			if (!ruc_on_type_hierarchy.TryGetValue (type, out var hasRucOnTypeHierarchy))
-				return TryGetRequiresUnreferencedCodeAttributeOnTypeHierarchy (type, out var _);
-			return hasRucOnTypeHierarchy is not null;
+			if (HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (type))
+				return true;
+			while (type.DeclaringType != null) {
+				if (HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (type))
+					return true;
+				type = type.DeclaringType;
+			}
```
vitek-karas, line 568: For simplicity I would just call TryGet... here and discard the attribute value.

src/linker/Linker.Steps/MarkStep.cs
tlakollo: Let's say that is because it was a partial commit xD

src/linker/Linker.Steps/MarkStep.cs
tlakollo: I have some scenarios with suppressions that are working fine, although is not extensive. Do you have a particular scenario that you would like me to test?

src/linker/Linker.Steps/MarkStep.cs
tlakollo: I removed the Suppress function call all along, for this specific scenario I think the normal suppression rules don't apply.
Added a test case where I suppress IL2109 by checking that the parent class has RUC, which I think is the actual suppression that applies for IL2109

src/linker/Linker.Steps/MarkStep.cs
sbomer: > Roslyn is actually always warning about the type - not about the member. That's why it also warns on a cast to the type, or warns when declaring a local. We're not going to warn on those I assume.

What would be the problem with warning for these patterns? It doesn't seem like you can do anything too useful with just casts/local declarations, unless you also try to instantiate the type or call static methods.

MichalStrehovsky: > What would be the problem with warning for these patterns? It doesn't seem like you can do anything too useful with just casts/local declarations, unless you also try to instantiate the type or call static methods.

It will produce a lot more _unnecessary_ warnings.

I think there's a lot of value in avoiding too much unnecessary noise. It we can avoid generating a warning for something that is not actually unsafe, we should. The point of the warnings it to prevent different behavior after trimming, not "to be consistent with `ObsoleteAttribute`". If the thing illink is warning about can __under no circumstances__ cause a runtime difference, that warning is useless and noise for the developer.

(That's why I'm also asking about the static fields.)

agocke: > Don't let the class RUC apply to the static constructor. People get a warning in the static constructor and have to deal with it.

The main problem is this. How do we expect people to deal with calls to RUC methods in static initializers? It seems like it would imply a significant code change, mostly to reduce our implementation complexity.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -783,5 +784,182 @@ public static void Test ()
...
+
+				public ClassWithRequiresUnreferencedCode () { }
+
+				public static void StaticMethod () { }
+
+				public void NonStaticMethod () { }
+
+				public class NestedClass
+				{
+					public static void NestedStaticMethod () { }
```
agocke, line 803: Is there a test where the nested method inside a nested class calls something with RUC?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2810,13 +2827,24 @@ internal void CheckAndReportRequiresUnreferencedCode (MethodDefinition method)
...
-				string message = string.Format (formatString, method.GetDisplayName (), arg1, arg2);
-				_context.LogWarning (message, 2026, currentOrigin, MessageSubCategory.TrimAnalysis);
+			if (method.IsStatic || method.IsConstructor) {
+				if (Annotations.TryGetEffectiveRequiresUnreferencedCodeAttributeOnType (method.DeclaringType, out RequiresUnreferencedCodeAttribute requiresUnreferencedCodeOnTypeHierarchy) && requiresUnreferencedCodeOnTypeHierarchy != null) {
+					ReportRequiresUnreferencedCode (method.GetDisplayName (), requiresUnreferencedCodeOnTypeHierarchy, currentOrigin);
+					return;
+				}
 			}
+
+			if (Annotations.TryGetLinkerAttribute (method, out RequiresUnreferencedCodeAttribute requiresUnreferencedCode))
```
agocke, line 2837: Can the `out` parameter be `null` here?

tlakollo: Only if we return false in which case we will not execute the code following. Otherwise, the LinkerAttributes cache when gets generated checks for the creation of the right attribute with the right parameters.

src/linker/Linker/Annotations.cs
```diff
@@ -552,6 +552,41 @@ public bool SetPreservedStaticCtor (TypeDefinition type)
...
+		/// <summary>
+		/// Looks if there is a RequiresUnreferencedCodeAttribute on the <paramref name="type"/> or any of its declaring types and returns the attribute information
+		/// on <paramref name="attribute"/>
+		/// </summary>
+		/// <param name="type">Type to start the search of the attribute</param>
+		/// <param name="attribute">Information about the found RequiresUnreferencedCode Attribute</param>
+		/// <returns>Returns true along with the RequiresUnreferencedCodeAttribute if found, otherwise returns false</returns>
+		public bool TryGetEffectiveRequiresUnreferencedCodeAttributeOnType (TypeDefinition type, out RequiresUnreferencedCodeAttribute attribute)
+		{
+			do {
```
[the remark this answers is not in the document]
tlakollo: Yup, I was thinking on calling always TryGetEffectiveRequiresUnreferencedCodeAttributeOnType and discard the attribute for the cases I do HasEffectiveRequiresUnreferencedCodeOnType. Don't have a strong opinion for either way.

src/linker/Linker.Steps/MarkStep.cs
tlakollo: https://github.com/mono/linker/blob/2174dc1bf8217ab8f5b77b0fcc0ab0cbf1973a85/src/linker/Linker/LinkerAttributesInformation.cs#L34
https://github.com/mono/linker/blob/2174dc1bf8217ab8f5b77b0fcc0ab0cbf1973a85/src/linker/Linker/LinkerAttributesInformation.cs#L88

agocke: Can we `#enable nullable` for these sections so we have proper tracking? If this can only return `null` when false, then I think the null check on line 2831 is always true

src/linker/Linker/Annotations.cs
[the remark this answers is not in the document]
agocke: I don't see why we wouldn't call the other method and discard the extra item rather than duplicate the code.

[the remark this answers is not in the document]
agocke: Sorry, I don't think I was clear -- you could have kept the other method, and had it just call this one with a discard. My objection was to duplicating the while loop, not having two methods.

src/linker/Linker.Steps/MarkStep.cs
tlakollo: Aghh, that was some remaining code of a previous implementation, deleted the check

agocke (approved):
LGTM, thanks!

src/linker/Linker/Annotations.cs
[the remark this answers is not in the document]
tlakollo: I dont know it will be just
```C#
public bool HasEffectiveRequiresUnreferencedCodeOnType (TypeDefinition type) => TryGetEffectiveRequiresUnreferencedCodeAttributeOnType (type, out RequiresUnreferencedCodeAttribute _);
```
I think I prefer just call it

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: > The main problem is this. How do we expect people to deal with calls to RUC methods in static initializers? It seems like it would imply a significant code change, mostly to reduce our implementation complexity.

Yup, same as we scoped out RUC on structs and interfaces. I don't even know how we would be able to support the semantics of "cctor is considered RUC" there (the problem is that for structs and interfaces, the cctor is going to run when virtual methods get invoked - Equals/GetHashCode/ToString/interface methods - so we would have to mark as RUC those too; but that immediately causes a warning because the RUC annotation is inconsistent across virtuals and we can't guarantee it would warn - we can ignore the problem for classes because the cctor is triggered from the ctor there). I'm sure we would come up with some solution. But also it's very complex.

If we say RUC on struct and interface cctors is never going to work (we're also going to block it if people explicily add it to their cctor), we are already blocking a scenario because of complexity. People will have to work around if they hit it. We can document it.

Allowing RUC on class cctors (either explicityly placed on the cctor or propagated from the class-level cctor) is also pretty complex if we want to do it in a way that it doesn't produce meaningless warnings. The workarounds in the absence of that RUC handling are the same as for interfaces and structs.

vitek-karas:
I think @MichalStrehovsky has a great point in https://github.com/mono/linker/issues/2151. We should also change the behavior for that.

Issue #1764, opened 2021-01-19
Linker should warn when reflection-accessing annotated methods


MichalStrehovsky:
When we have a method that is an intrinsic for flow analysis purposes, or has parameters marked as `[DynamicallyAccessedMembers]`, or is marked `[RequiresUnreferencedCode]`, linker should warn when we see that method being reflection accessed.

The problem this is trying to avoid is someone e.g. locating `MakeGenericMethod` (that currently warns if used statically) and reflection-invoking it with some parameters. The initial reflection access should be a warning since illink is not capable of figuring out the dataflow into the method. Someone could be violating the annotations, resulting in a broken app without warnings.

See https://github.com/dotnet/runtime/pull/47134/files#diff-e09e7f6880244829d1c3b16e19b451ab04d059a7df0d8a367cd3fd08230c9096R64.

We should probably extend this to delegates pointing to such methods (although that one has a natural extension - we could require the delegate invoke method itself to be annotated same as the target).

vitek-karas:
Isn't this basically a dupe of https://github.com/mono/linker/issues/1607 ?

MichalStrehovsky:
It wasn't clear to me what the scope of that issue is - the original issue talks about warning when a `RequiresUnreferencedCode`-annotated method somehow ends up in the final output.

I added the "methods that have DynamicallyAccessedMembers annotations" as a comment, but the original issue already had a pretty big scope and there was no reaction to that comment, so I filed a new one to make sure that aspect is not missed either. In that case we don't want to warn whenever such method ends up in the output - we specifically want to warn if we see it gets added as "visible to reflection". The place where it gets implemented might be different.

vitek-karas:
I'm fine keeping it as separate issues.

Why do you think it's important to recognize "visible to reflection"? Let's say I have a method `CallEverything([DAM(All)] Type type)`. And I root it in my descriptor, because it's going to be called from a place linker can't see (for example native code - mono can do this and probably does in some interop code). I think we should warn in this case - since linker can't reason about its usage and thus it's potentially dangerous.

MichalStrehovsky:
Would we warn if such method is in a `copy` or `copyused` assembly? We don't know why someone made it `copy` or `copyused` - it's possible they did it because there will be native code reflection-invoking everything in the assembly. Or maybe they just did it because they are doing some reflection and they can't be bothered to specify what exactly they reflect on.

[...]

PR #2145, opened 2021-07-14
Warn when accessing DAM annotated method via reflection

vitek-karas:
Any access to DAM annotated method must be validated by the trimmer. Normal calls and such are handled already, but direct reflection access or other indirect accesses can lead to trimmer allowing potential invocation of DAM annotated method without validating the annotations. This change will warn whenever such potential "leak" happens.

This applies to method parameter, method return value and field annotations.
Uses the same infra as RUC already just adds additional checks.

Tests for the new cases.

Fixes https://github.com/mono/linker/issues/1764

### DIFF
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -75,7 +75,11 @@ namespace Mono.Linker.Dataflow
 
 		public bool ShouldWarnWhenAccessedForReflection (MethodDefinition method) =>
 			GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation) &&
-				(annotation.ParameterAnnotations != null || annotation.ReturnParameterAnnotation != DynamicallyAccessedMemberTypes.None);
+			(annotation.ParameterAnnotations != null || annotation.ReturnParameterAnnotation != DynamicallyAccessedMemberTypes.None);
+
+		public bool MethodHasNoAnnotatedParameters (MethodDefinition method) =>
+			GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation) &&
+			annotation.ParameterAnnotations == null;
 
 		public bool ShouldWarnWhenAccessedForReflection (FieldDefinition field) =>
 			GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out _);
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -37,20 +37,30 @@ namespace Mono.Linker.Dataflow
 		/// </summary>
 		/// <param name="parameterIndex">Parameter index in the IL sense. Parameter 0 on instance methods is `this`.</param>
 		/// <returns></returns>
-		public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition method, int parameterIndex) =>
-			(GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation) && annotation.ParameterAnnotations != null)
-				? annotation.ParameterAnnotations[parameterIndex]
-				: DynamicallyAccessedMemberTypes.None;
+		public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition method, int parameterIndex)
+		{
+			if (GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation) &&
+				annotation.ParameterAnnotations != null)
+				return annotation.ParameterAnnotations[parameterIndex];
 
-		public DynamicallyAccessedMemberTypes GetReturnParameterAnnotation (MethodDefinition method) =>
-			GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation)
-				? annotation.ReturnParameterAnnotation
-				: DynamicallyAccessedMemberTypes.None;
+			return DynamicallyAccessedMemberTypes.None;
+		}
 
-		public DynamicallyAccessedMemberTypes GetFieldAnnotation (FieldDefinition field) =>
-			GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out var annotation)
-				? annotation.Annotation
-				: DynamicallyAccessedMemberTypes.None;
+		public DynamicallyAccessedMemberTypes GetReturnParameterAnnotation (MethodDefinition method)
+		{
+			if (GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation))
+				return annotation.ReturnParameterAnnotation;
+
+			return DynamicallyAccessedMemberTypes.None;
+		}
+
+		public DynamicallyAccessedMemberTypes GetFieldAnnotation (FieldDefinition field)
+		{
+			if (GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out var annotation))
+				return annotation.Annotation;
+
+			return DynamicallyAccessedMemberTypes.None;
+		}
 
 		public DynamicallyAccessedMemberTypes GetTypeAnnotation (TypeDefinition type) =>
 			GetAnnotations (type).TypeAnnotation;
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -23,57 +23,37 @@ namespace Mono.Linker.Dataflow
 			_hierarchyInfo = new TypeHierarchyCache (context);
 		}
 
-		public bool RequiresDataFlowAnalysis (MethodDefinition method)
-		{
-			return GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out _);
-		}
+		public bool RequiresDataFlowAnalysis (MethodDefinition method) =>
+			GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out _);
 
-		public bool RequiresDataFlowAnalysis (FieldDefinition field)
-		{
-			return GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out _);
-		}
+		public bool RequiresDataFlowAnalysis (FieldDefinition field) =>
+			GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out _);
 
-		public bool RequiresDataFlowAnalysis (GenericParameter genericParameter)
-		{
-			return GetGenericParameterAnnotation (genericParameter) != DynamicallyAccessedMemberTypes.None;
-		}
+		public bool RequiresDataFlowAnalysis (GenericParameter genericParameter) =>
+			GetGenericParameterAnnotation (genericParameter) != DynamicallyAccessedMemberTypes.None;
 
 		/// <summary>
 		/// Retrieves the annotations for the given parameter.
 		/// </summary>
 		/// <param name="parameterIndex">Parameter index in the IL sense. Parameter 0 on instance methods is `this`.</param>
 		/// <returns></returns>
-		public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition method, int parameterIndex)
-		{
-			if (GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation) && annotation.ParameterAnnotations != null) {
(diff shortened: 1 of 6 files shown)


vitek-karas:
I ran this on dotnet/runtime and it found a couple of places where it warns. Fixing them now as most are easy.
The result of this was the tweak to not warn on non-virtual return annotations - since that occurs in runtime and is just tedious (as it's clearly not a problem). I'll like a draft PR to runtime with the fixes here once ready.

vitek-karas:
The changes to runtime are being done here: https://github.com/dotnet/runtime/compare/main...vitek-karas:FixReflectionOnDAMImpact?expand=1

Everything except EventSource is already resolved.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1557,6 +1557,22 @@ void MarkField (FieldDefinition field, in DependencyInfo reason)
 				Annotations.Mark (field, reason);
 			}
 
+			switch (reason.Kind) {
+			case DependencyKind.AccessedViaReflection:
```
sbomer, line 1561: I know we use this kind of pattern in a few places already (I added some of them myself :frowning:) but I think we should avoid it. Special-casing behavior for some "reason kinds" isn't very maintainable - it requires examining all callers of the method to see which cases should be treated specially. I would prefer to add a separate parameter to MarkField that indicates what the optional behavior is, and pass it in from some callers only.

vitek-karas: In general I agree, but it would require non-trivial effort. Not for fields, but for methods. The entire RUC and DAM detection in `MarkMethod` currently relies on this. If we were to fix it, I think we should do it in both places.

I did go through all the callers for both methods and fields, it's tedious I agree, but fixing it would require lot more work. I didn't want to block this change on it. Mainly because I want to avoid passing many different parameters to Mark methods. And `MarkMethod` would need at least 2 more (RUC and DAM handling). So in my mind a cleaner solution would be to do this in the planned refactoring of scope stacks and `DependencyInfo` into one parameter. With that we could relatively easily add more fields to the structure we will pass around and store additional information that way.

vitek-karas:
This is ready for review.

That said we may want to hold off on merging this until we have https://github.com/mono/linker/issues/2136 resolved since this causes new warnings in EventSource which will require different supressions once that change is implemented.

vitek-karas:
Ping. Could somebody please review this?

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -23,57 +23,37 @@ public FlowAnnotations (LinkContext context)
...
-			return DynamicallyAccessedMemberTypes.None;
-		}
-
-		public DynamicallyAccessedMemberTypes GetReturnParameterAnnotation (MethodDefinition method)
-		{
-			if (GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation)) {
-				return annotation.ReturnParameterAnnotation;
-			}
+		public DynamicallyAccessedMemberTypes GetParameterAnnotation (MethodDefinition method, int parameterIndex) =>
+			(GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation) && annotation.ParameterAnnotations != null)
```
MichalStrehovsky, line 41: Feels like this one is forcing things into following a pattern but hurting readability in the process.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1557,6 +1557,22 @@ void MarkField (FieldDefinition field, in DependencyInfo reason)
...
+			case DependencyKind.DynamicDependency:
+			case DependencyKind.DynamicallyAccessedMember:
+			case DependencyKind.InteropMethodDependency:
+				if (_context.Annotations.FlowAnnotations.ShouldWarnWhenAccessedForReflection (field))
+					_context.LogWarning (
```
MichalStrehovsky, line 1566: Do we need a call to ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2773,24 +2789,51 @@ void ProcessRequiresUnreferencedCode (MethodDefinition method, DependencyKind de
...
+
+				default:
+					break;
+				}
+
+				// If the method only has annotation on the return value and it's not virtual avoid warning.
+				// Return value annotations are "consumed" by the caller of a method, and as such there is nothing
+				// wrong calling these dynamically. The only problem can happen if something overrides a virtual
+				// method with annotated return value at runtime - in this case the trimmer can't validate
+				// that the method will return only types which fulfill the annotation's requirements.
```
MichalStrehovsky, line 2827: We should also disallow this if the return value is a `ref` because one can just create a delegate from that and modify whatever the byref is pointing to without anyone checking.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2773,24 +2789,51 @@ void ProcessRequiresUnreferencedCode (MethodDefinition method, DependencyKind de
...
+			// (else we will produce warning IL2046 or IL2092 or some other warning).
+			// When marking override methods via DynamicallyAccessedMembers, we should only issue a warning for the base method.
+			if (dependencyKind == DependencyKind.DynamicallyAccessedMember &&
+				method.IsVirtual &&
+				Annotations.GetBaseMethods (method) != null)
+				return;
+
+			CheckAndReportRequiresUnreferencedCode (method);
+
+			if (_context.Annotations.FlowAnnotations.ShouldWarnWhenAccessedForReflection (method)) {
```
MichalStrehovsky, line 2807: We should make sure this also covers all the methods that we handle with intrinsics - last time I looked into it, not everything we handle intrinsically was properly annotated: https://github.com/dotnet/runtime/issues/53121

MichalStrehovsky (approved):
Looks good otherwise!

src/linker/Linker.Dataflow/FlowAnnotations.cs
vitek-karas: I was on the fence about these - so if others like the more verbose shape... I changed the once which have conditions in them to normal if pattern.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Great catch - thanks a lot!

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: I'm not sure what we can do about this (other then going over the list once again and fixing the annotations).

The way it is now -> if there's an intrinsic handling which leads to warnings and the method has no annotation, it will not be caught.

If I change this to also look at all intrinsically recognized methods it would report on perfectly safe things, since we intrinsically handle patterns which do not produce any warnings.

I welcome any ideas here...

One idea:
Modify the switch in https://github.com/mono/linker/blob/4526d388102db2dc481522d9dee92b8f32414b23/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs#L304-L627 to return a boolen which indicates if the method should be handled here - meaning calling the method without further checks is potentially dangerous.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Can you give me an example? You mean something like:

```C#
[DAM(PublicMethods)]
Type _typeField;

[return: DAM(PublicMethods)]
ref Type GetRefType() { return ref _typeField; }

// But now what?
```

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: I would fix any of the necessary annotations in the runtime repo and not worry about anything else. Then we just add: "whenever an intrinsic is introduced, make sure it's annotated" into our list of things to look at when reviewing code.

Should be futureproof enough, including the scenarios of "using new illlink with an old framework".

src/linker/Linker.Steps/MarkStep.cs
MichalStrehovsky: Along the lines of:

```csharp
delegate ref Type Break();

var b = (Break)typeof(Foo).GetMethod("GetRefType").CreateDelegate(typeof(Break));
ref x = b();

// Nobody told us to keep methods on Bar
x = typeof(Bar);
```

vitek-karas: But `x` has no annotations on it (even implicitly), so if I add one more line:
```
x.GetMethods(); // Warns, since x doesn't have any annotations
```

MichalStrehovsky: `_typeField` is annotated though so we can get methods on it without warnings.

vitek-karas: I see - the magical line is:
```C#
_typeField.GetMethods(); // This doesn't warn, but it now contains typeof(Bar) which nobody checked
```

vitek-karas: This feels like a bigger problem - if I don't annotate the return value it also doesn't warn - and for that I don't need a delegate to break it:
```C#
[DAM(PublicMethods)]
Type _fieldType;

ref Type GetRefType() { return _fieldType; }

ref Type x = ref GetRefType();
x = typeof(Bar);
_fieldType.GetMethods();
```

This doesn't warn at all.

vitek-karas: In short it seems that we don't correctly handle ref return values.
And I didn't look at ref locals either (maybe there are issues there as well).

MichalStrehovsky: Ah yeah. Good catch. That one looks out of scope for this pull request.

sbomer, line 2827: I'm not sure I understand why virtual methods are considered specially here. How would you override it at runtime without producing analysis warnings elsewhere, for example when instantiating the overriding type? We can't check annotations on any methods created at runtime so I would expect this to fall into the same bucket.

vitek-karas: The only way to override a method at runtime is via something like `TypeBuilder` which will need the `MethodInfo` of the base method to override. So we're guarding all the ways to get that `MethodInfo` - which is what this code does. The problem is that the other code in the app may call the base method with its annotation and assume that any type it gets as the return value will fulfill the annotation - but if that value comes from the runtime generated method, linker can't see it and validate - so we warn here.

Non-virtual methods can't be overriden, so there's no way (other than the ref return mentioned above) to change the implementation. Linker checks all implementations to validate that they fulfill the annotation requirements on return values - so no problem there. (Note that interface methods are also considered virtual here and the logic will apply to them as well).

MichalStrehovsky: It might help to add a blurb like this to the comment:

```csharp
class Foo
{
    [return: DAM(Fields)]
    public abstract Type GetTypeWithFields();
}
```

Then we reflection emit a class that derives from `Foo` and add a virtual method named `GetTypeWithFields` with an implementation that is equivalent to `return SomeOtherType.SomeMethod()`.

Since illink didn't see this, it wouldn't make sure whatever SomeMethod returns meets the requirements of the return type.

Our check ensures reflection.emit won't be able to derive from Foo without raising a warning (Ref.Emit requires the base type be annotated as DAM.All).

vitek-karas: I added a long description with a sample - hopefully it will make it a bit clearer.

vitek-karas: We probably didn't do full analysis of this use case, but I would think that getting to a bad place through ref emit will be pretty hard (as in, we've done a good job guarding ourselves). The logic is basically that in order to call any method from a ref emit code, you must be able to reflection access that method (get a MethodInfo). All methods which lead to dangerous places should be annotated and thus should produce warnings when accessed through reflection. So even code which would use reflection from the ref emit code should still cause warnings, since it will have to get MethodInfo of the reflection APIs, which will produce warnings (since those APIs are annotated).

I think we have potential holes around things like out parameters, by ref values and so on - which is basically partially what this discussion is about, but ignoring those cases I can't think of another break (which means absolutely nothing... since I didn't realize the by ref problems either).

PR #2146, opened 2021-07-14
Fix Activator.CreateInstance case

sbomer:
When no ctor parameters are passed, and non-public binding flags are
used, we only need to require parameterless public constructors. Fixes one of the issues in https://github.com/dotnet/runtime/pull/55636.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1596,10 +1596,10 @@ namespace Mono.Linker.Dataflow
 								var requiredMemberTypes = GetDynamicallyAccessedMemberTypesFromBindingFlagsForConstructors (bindingFlags);
 
 								// Special case the public parameterless constructor if we know that there are 0 args passed in
-								if (ctorParameterCount == 0 &&
-									requiredMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicConstructors) &&
-									!requiredMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.NonPublicConstructors))
-									requiredMemberTypes = DynamicallyAccessedMemberTypes.PublicParameterlessConstructor;
+								if (ctorParameterCount == 0 && requiredMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicConstructors)) {
+									requiredMemberTypes &= ~DynamicallyAccessedMemberTypes.PublicConstructors;
+									requiredMemberTypes |= DynamicallyAccessedMemberTypes.PublicParameterlessConstructor;
+								}
 
 								RequireDynamicallyAccessedMembers (ref reflectionContext, requiredMemberTypes, value, calledMethod.Parameters[0]);
 							}
--- a/test/Mono.Linker.Tests.Cases/Reflection/ActivatorCreateInstance.cs
+++ b/test/Mono.Linker.Tests.Cases/Reflection/ActivatorCreateInstance.cs
@@ -55,6 +55,7 @@ namespace Mono.Linker.Tests.Cases.Reflection
 			TestNullArgsOnKnownType ();
 			TestNullArgsOnAnnotatedType (typeof (TestType));
 			TestNullArgsNonPublicOnly (typeof (TestType));
+			TestNullArgsNonPublicWithNonPublicAnnotation (typeof (TestType));
 
 			CreateInstanceWithGetTypeFromHierarchy.Test ();
 		}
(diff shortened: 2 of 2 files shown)


agocke:
Test case?

sbomer:
Yup, I just wanted to prioritize figuring out the other warnings.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1596,10 +1596,10 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 								var requiredMemberTypes = GetDynamicallyAccessedMemberTypesFromBindingFlagsForConstructors (bindingFlags);
 
 								// Special case the public parameterless constructor if we know that there are 0 args passed in
-								if (ctorParameterCount == 0 &&
-									requiredMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicConstructors) &&
-									!requiredMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.NonPublicConstructors))
-									requiredMemberTypes = DynamicallyAccessedMemberTypes.PublicParameterlessConstructor;
+								if (ctorParameterCount == 0 && requiredMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicConstructors)) {
+									requiredMemberTypes &= ~DynamicallyAccessedMemberTypes.PublicConstructors;
+									requiredMemberTypes |= DynamicallyAccessedMemberTypes.PublicParameterlessConstructor;
```
vitek-karas, line 1601: I must be missing something :-(

I think this is basically a no-op. `PublicConstructors = 3`, `PublicParameterlessConstructor = 1`, so `HasFlags(PublicConstructors)` should imply `PublicParameterlessConstructor`.
I wonder how come the original test doesn't fail with this change. The test `TestNullArgsOnAnnotatedType` should not work with this change if I understand it correctly.

sbomer: Sorry, I didn't describe the problem very clearly. This is to address the case where you use Public | NonPublic binding flags. The special case before this change will kick in only when the requirements have Public & !NonPublic - so they wouldn't apply. But we still don't need to keep all public ctors (we know there's only one parameter) - the Public requirement should be downgraded to PublicParameterless, regardless of whether NonPublic ctors were required.

`TestNullArgsOnAnnotatedType` will still hit this special case because it doesn't use NonPublic binding flags.

For more context, we started getting this warning:
```
/_/src/libraries/System.Data.Common/src/System/Data/DataSet.cs(1100,13): error IL2072: (NETCORE_ENGINEERING_TELEMETRY=Build) System.Data.DataSet.CreateInstanceOfThisType(): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors' in call to 'System.Object System.Activator::CreateInstance(System.Type,System.Boolean)'. The return value of method 'System.Type System.Object::GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
```

This is for the CreateInstance call here:

test/Mono.Linker.Tests.Cases/Reflection/ActivatorCreateInstance.cs
```diff
@@ -532,6 +533,15 @@ private static void TestNullArgsOnKnownType ()
 			Activator.CreateInstance (type, BindingFlags.NonPublic | BindingFlags.Instance, null, null, CultureInfo.InvariantCulture);
 		}
 
+		[Kept]
+		[ExpectedNoWarnings]
+		private static void TestNullArgsNonPublicWithNonPublicAnnotation (
+			[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicParameterlessConstructor | DynamicallyAccessedMemberTypes.NonPublicConstructors),
+			KeptAttributeAttribute (typeof (DynamicallyAccessedMembersAttribute))] Type type)
+		{
+			Activator.CreateInstance (type, nonPublic: true);
```
agocke, line 542: How does this test work? I assume we're trying to assert that the given type has its non-public constructors preserved, but how are we confirming that we're also not marking the other public constructors?

Wasn't the problem in the PR that we were marking extra things and getting additional warnings?

vitek-karas: Not really - the problem is that PublicParameterlessConstructor | NonPublicConstructors was an annotation which was not considered sufficient for calling CreateInstance(type, true). Which it should be.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: I was reading this wrong. I get it now - sorry - no idea what I was thinking before.

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/Reflection/ActivatorCreateInstance.cs
sbomer: > Wasn't the problem in the PR that we were marking extra things and getting additional warnings?

I may have given you the wrong impression when we chatted - this additional warning was because we tried to _require_ (not mark) extra things on the input (which was an abstract DAMT value). But the extra requirements also would have resulted in extra things being marked for concrete input types (that just wasn't the failure mode we happened to hit).

PR #2147, opened 2021-07-14
Add source generator for analyzer diagnostics

mateoatr:
This PR introduces a source generator to be used on the different ILLinker analyzers to get the supported diagnostics for types that derive from `DiagnosticAnalyzer`. It's meant to be used together with the new attribute `AddSupportedDiagnostic`; where all the information needed to generate a given diagnostic should be specified. This ought to save us from rewriting over and over some boilerplate code such as:

```C#
static readonly DiagnosticDescriptor s_locationRule = new DiagnosticDescriptor (
	IL3000,
	new LocalizableResourceString (nameof (SharedStrings.AvoidAssemblyLocationInSingleFileTitle),
		SharedStrings.ResourceManager, typeof (SharedStrings)),
	new LocalizableResourceString (nameof (SharedStrings.AvoidAssemblyLocationInSingleFileMessage),
		SharedStrings.ResourceManager, typeof (SharedStrings)),
	DiagnosticCategory.SingleFile,
	DiagnosticSeverity.Warning,
	isEnabledByDefault: true,
	helpLinkUri: "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/il3000");
```

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
@@ -19,7 +19,7 @@ namespace ILLink.RoslynAnalyzer
 	[AddSupportedDiagnostic ("IL3002", "RequiresAssemblyFiles",
 		Category = DiagnosticCategory.SingleFile,
 		HelpLinkURI = "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/il3002")]
-	[AddSupportedDiagnostic ("IL3003", "RequiresAttributeMismatch")]
+	[AddSupportedDiagnostic ("IL3003", "RequiresAttributeMismatch", Category = DiagnosticCategory.SingleFile)]
 	[DiagnosticAnalyzer (LanguageNames.CSharp)]
 	public sealed class RequiresAssemblyFilesAnalyzer : RequiresAnalyzerBase
 	{
--- a/src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
@@ -5,67 +5,29 @@
 using System;
 using System.Collections.Immutable;
 using System.Linq;
-using ILLink.Shared;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 
 namespace ILLink.RoslynAnalyzer
 {
+	[AddSupportedDiagnostic ("IL3000", "AvoidAssemblyLocationInSingleFile",
+		Category = DiagnosticCategory.SingleFile,
+		HelpLinkURI = "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/il3000")]
+	[AddSupportedDiagnostic ("IL3001", "AvoidAssemblyGetFilesInSingleFile",
+		Category = DiagnosticCategory.SingleFile,
+		HelpLinkURI = "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/il3001")]
+	[AddSupportedDiagnostic ("IL3002", "RequiresAssemblyFiles",
+		Category = DiagnosticCategory.SingleFile,
+		HelpLinkURI = "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/il3002")]
+	[AddSupportedDiagnostic ("IL3003", "RequiresAttributeMismatch")]
 	[DiagnosticAnalyzer (LanguageNames.CSharp)]
 	public sealed class RequiresAssemblyFilesAnalyzer : RequiresAnalyzerBase
 	{
-		public const string IL3000 = nameof (IL3000);
-		public const string IL3001 = nameof (IL3001);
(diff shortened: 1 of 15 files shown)


src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -5,77 +5,39 @@
...
+	[AddSupportedDiagnostic ("IL3000", "AvoidAssemblyLocationInSingleFile",
+		Category = DiagnosticCategory.SingleFile,
+		HelpLinkURI = "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/il3000")]
+	[AddSupportedDiagnostic ("IL3001", "AvoidAssemblyGetFilesInSingleFile",
+		Category = DiagnosticCategory.SingleFile,
+		HelpLinkURI = "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/il3001")]
+	[AddSupportedDiagnostic ("IL3002", "RequiresAssemblyFiles",
+		Category = DiagnosticCategory.SingleFile,
+		HelpLinkURI = "https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/il3002")]
+	[AddSupportedDiagnostic ("IL3003", "RequiresAttributeMismatch")]
```
tlakollo, line 22: This will default to category trimming and is still a SingleFile warning

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -94,10 +56,10 @@ protected override ImmutableArray<ISymbol> GetSpecialIncompatibleMembers (Compil
 
 			var assemblyType = compilation.GetTypeByMetadataName ("System.Reflection.Assembly");
 			if (assemblyType != null) {
-				// properties
+				// Properties
 				ImmutableArrayOperations.AddIfNotNull (dangerousPatternsBuilder, ImmutableArrayOperations.TryGetSingleSymbol<IPropertySymbol> (assemblyType.GetMembers ("Location")));
 
-				// methods
+				// Methods
```
tlakollo, line 62: Maybe its a good time to delete the CodeBase and EscapeCodeBase special handling? The runtime is already annotated

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
mateoatr: I simply put the category that it has today:

https://github.com/mono/linker/blob/2174dc1bf8217ab8f5b77b0fcc0ab0cbf1973a85/src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs#L64

Do you think we should change it then?

tlakollo: Yikes, me putting comments for my own mistakes 😂
Yes, this is not a trimming warning is a mismatch between RequiresAssemblyFiles so it should have the SingleFile category

src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
mateoatr: Nice. Shouldn't we remove all special cases? `GetFile`, `GetFiles` and `Location` are also annotated on the runtime, so I think we could get rid of this method.

tlakollo: I'm up for it, but now that I think more, all of this would be a breaking change for people that had suppressed IL3000 and IL3001 and now will start having IL3002. On the other hand, it will even simplify the current analyzer code we would only generate IL3002 and IL3003. But I'm not very good at dimensioning if it's worth it the break, so I will leave @agocke to comment on that

agocke: Don't we have a customized message for `Location`? Using the attribute for `GetFile` and `GetFiles` seems fine, but `Assembly.Location` seems like it might be worth a special message, if we can give a good one.

tlakollo: Well, we control the message field in RAF, is that enough? Or we can change the Documentation URLs for IL3000 and IL3001 to not show the warning code (since they wouldn't be generated anymore) and pass them in RAF as URLs

tlakollo: By the way, I gave a quick look at the runtime and in System.Reflection.Assembly the GetFile/GetFiles/Location are not annotated with RAF

mateoatr: > By the way, I gave a quick look at the runtime and in System.Reflection.Assembly the GetFile/GetFiles/Location are not annotated with RAF

My bad. Yeah, just saw, some of these have RUC on them.

tlakollo: Having said that we could annotate the runtime if we want

PR #2148, opened 2021-07-15
Fix warning suppressions on types for type hierarchy

sbomer:
Currently it's not possible to suppress warnings that originate from types in most cases, because we don't process the suppression attributes until after we've processed the type for type hierarchy related warnings.

https://github.com/mono/linker/pull/2114 changes the library-mode behavior of type hierarchy marking to apply the DAMT annotation when the type with the annotation is first seen. So far we have been reporting an inconsistent warning origin for these (sometimes the GetType call, sometimes the type - see https://github.com/mono/linker/issues/2136). But by applying the annotations when the type is first marked, we are now reporting warnings from the type, which is introducing some of the additional warnings in https://github.com/dotnet/runtime/pull/55636 that were previously suppressed on the GetType call.

This change makes it possible to suppress the warnings on the originating type. I don't think it's complete since there might be cases where the type is derived from an attribute type and also has attribute instances of that attribute - which would mark the declaring type before processing type hierarchy annotations. But this workaround should be good enough in the meantime, in case we want to use type-level suppressions to unblock runtime dependency flow.

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyLibraryModeSuppressions.cs
+++ b/test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyLibraryModeSuppressions.cs
@@ -1,0 +1,51 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Helpers;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.Reflection
+{
+	[SetupLinkerArgument ("-a", "test.exe", "library")]
+	[ExpectedNoWarnings]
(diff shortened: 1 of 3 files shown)


test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyLibraryModeSuppressions.cs
```diff
@@ -0,0 +1,51 @@
...
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Helpers;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.Reflection
+{
+	[SetupLinkerArgument ("-a", "test.exe", "library")]
+	[ExpectedNoWarnings]
+	[KeptMember (".ctor()")]
+	public class TypeHierarchyLibraryModeSuppressions
```
vitek-karas, line 18: We should also have tests for non-library mode for this... which will probably have to be partially disabled due to the bug... I just want to make sure we will have coverage for that case.

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyLibraryModeSuppressions.cs
sbomer: I added some tests for non-library mode but they're pretty broken.

PR #2150, opened 2021-07-16
Add `dynamic` analyzer

mateoatr:
A `dynamic` type causes reflection to be performed on the runtime type, making it trimmer-incompatible. This analyzer will warn on any invocation on a dynamic type with `IL2026`.

Happy to read any suggestion(s) on interesting testing scenarios.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -216,13 +216,13 @@ namespace ILLink.RoslynAnalyzer
 
 				case ILocalFunctionOperation local when targets.HasFlag (DiagnosticTargets.MethodOrConstructor):
 					return local.Symbol;
-				
+
 				case IMethodBodyBaseOperation when targets.HasFlag (DiagnosticTargets.MethodOrConstructor):
 				case IPropertyReferenceOperation when targets.HasFlag (DiagnosticTargets.Property):
 				case IFieldReferenceOperation when targets.HasFlag (DiagnosticTargets.Field):
 				case IEventReferenceOperation when targets.HasFlag (DiagnosticTargets.Event):
 					return operationContext.ContainingSymbol;
-				
+
 				default:
 					parent = parent.Parent;
 					break;
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -25,6 +25,8 @@ namespace ILLink.RoslynAnalyzer
 
 		private protected abstract DiagnosticDescriptor RequiresAttributeMismatch { get; }
 
+		private protected virtual ImmutableArray<(Action<OperationAnalysisContext> Action, OperationKind[] OperationKind)> ExtraOperationActions { get; } = ImmutableArray<(Action<OperationAnalysisContext> Action, OperationKind[] OperationKind)>.Empty;
+
 		public override void Initialize (AnalysisContext context)
 		{
 			context.EnableConcurrentExecution ();
@@ -113,6 +115,10 @@ namespace ILLink.RoslynAnalyzer
 					CheckCalledMember (operationContext, methodSymbol, incompatibleMembers);
 				}, OperationKind.DelegateCreation);
 
+				// Register any extra operation actions supported by the analyzer.
+				foreach (var extraOperationAction in ExtraOperationActions)
+					context.RegisterOperationAction (extraOperationAction.Action, extraOperationAction.OperationKind);
+
 				void CheckStaticConstructors (OperationAnalysisContext operationContext,
 					ImmutableArray<IMethodSymbol> staticConstructors)
 				{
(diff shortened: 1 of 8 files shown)


mateoatr:
Update: checking that the warning isn't emitted whenever the invocation on a dynamic type happens inside a member marked with RUC.

tlakollo (approved):

src/ILLink.RoslynAnalyzer/Constants.cs
```diff
@@ -0,0 +1,24 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace ILLink.RoslynAnalyzer
+{
+	public static class Constants
+	{
+		public static class WarningCodes
+		{
+			public const string IL2026 = nameof (IL2026);
```
tlakollo, line 11: For me seems like this list of warnings will be replaced in the future with a shared enum with the linker, but for now seems ok.

mateoatr: Yup. That's the plan.

src/ILLink.Shared/SharedStrings.resx
```diff
@@ -159,4 +159,10 @@
   <data name="InterfaceRequiresMismatchMessage" xml:space="preserve">
     <value>Interface member '{2}' with '{0}' has an implementation member '{1}' without '{0}'</value>
   </data>
+  <data name="DynamicTypeInvocationMessage" xml:space="preserve">
+    <value>Invoking members on dynamic types is not trimming safe. Types or member might have been removed by the trimmer.</value>
```
agocke, line 163:
```suggestion
    <value>Invoking members on dynamic types is not trimming-compatible. Types or member might have been removed by the trimmer.</value>
```

I think the terminology we're using now is "trimming-compatible"

src/ILLink.RoslynAnalyzer/DynamicTypeAnalyzer.cs
```diff
@@ -0,0 +1,50 @@
...
+		{
+			context.EnableConcurrentExecution ();
+			context.ConfigureGeneratedCodeAnalysis (GeneratedCodeAnalysisFlags.ReportDiagnostics);
+			context.RegisterCompilationStartAction (context => {
+				var compilation = context.Compilation;
+				if (!context.Options.MSBuildPropertyValueIsTrue (MSBuildPropertyOptionNames.EnableTrimAnalyzer, compilation))
+					return;
+
+				context.RegisterOperationAction (operationContext => {
+					var dynamicTypeInvocation = (IDynamicInvocationOperation) operationContext.Operation;
```
agocke, line 39: Heh forgot there was a special operation for dynamic invocation -- that certainly makes things simpler.

src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
```diff
@@ -9,6 +9,8 @@
     <!-- There are currently no translations, so the satellite assemblies are a waste of space. -->
     <EnableXlfLocalization>false</EnableXlfLocalization>
     <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
+    <!-- Let different analyzers emit the same diagnostic id. -->
+    <NoWarn>RS1019</NoWarn>
```
agocke, line 13: It may be worth it to just fold this into the RequiresUnreferencedCode analyzer to avoid the complication

mateoatr: Folded. For this I added
```C#
private protected abstract ImmutableArray<(Action<OperationAnalysisContext> Action, OperationKind[] OperationKind)> ExtraOperationActions
```
Into `RequiresAnalyzerBase` so that any analyzer can pass the operation actions that it will register. Since we cannot pass a diagnostic id string to the AnalyzerVerifier (used by the tests) when there's more than one supported diagnostic using the same id, I had to make the diagnostic descriptors of the RUC analyzer public. I'll hold off this PR until we settle a way for generating/retrieving diagnostic descriptors.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -24,6 +24,8 @@ public abstract class RequiresAnalyzerBase : DiagnosticAnalyzer
 
 		private protected abstract DiagnosticDescriptor RequiresAttributeMismatch { get; }
 
+		private protected abstract ImmutableArray<(Action<OperationAnalysisContext> Action, OperationKind[] OperationKind)> ExtraOperationActions { get; }
```
agocke, line 27: Make this `virtual` and return `Empty` because it's optional?

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -126,15 +132,18 @@ public override void Initialize (AnalysisContext context)
...
 
 					// Do not emit any diagnostic if caller is annotated with the attribute too.
 					if (containingSymbol.HasAttribute (RequiresAttributeName))
 						return;
+
 					// Check also for RequiresAttribute in the associated symbol
-					if (containingSymbol is IMethodSymbol methodSymbol && methodSymbol.AssociatedSymbol is not null && methodSymbol.AssociatedSymbol!.HasAttribute (RequiresAttributeName)) {
+					if (containingSymbol is IMethodSymbol methodSymbol &&
+						methodSymbol.AssociatedSymbol is not null &&
+						methodSymbol.AssociatedSymbol!.HasAttribute (RequiresAttributeName))
```
agocke, line 144:
```suggestion
					if (containingSymbol is IMethodSymbol { AssociatedSymbol: { } associated }&&
						associated.HasAttribute (RequiresAttributeName))
```

src/ILLink.RoslynAnalyzer/DiagnosticTargets.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace ILLink.RoslynAnalyzer
+{
+	[Flags]
+	public enum DiagnosticTargets
```
agocke, line 10: Any reason why we're splitting this out? I liked having the type grouped up with the only method that it's being used in.

src/ILLink.RoslynAnalyzer/OperationAnalysisContextExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+	internal static class OperationAnalysisContextExtensions
+	{
+		/// <summary>
+		/// Finds the symbol of the caller to the current operation, helps to find out the symbol in cases where the operation passes
+		/// through a lambda or a local function.
+		/// </summary>
+		/// <param name="operationContext">Analyzer operation context to retrieve the current operation.</param>
+		/// <param name="targets">Scope of the attribute to search for callers.</param>
+		/// <returns>The symbol of the caller to the operation</returns>
+		public static ISymbol GetContainingSymbol (this OperationAnalysisContext operationContext, DiagnosticTargets targets)
```
agocke, line 23: Was this moved because it's being used in more places?

mateoatr: I moved it since it feels like a method that is useful more scenarios other than `Requires...`. Both this and `DiagnosticTargets` can be kept inside `RequiresAnalyzerBase` if we want -- we'd just need to make this function visible to derived classes.

mateoatr:
Responded to feedback and updated to use the new DiagnosticDescriptors class. Unfortunately, looks like we are not getting any CI info back from AzDO. PTAL @agocke.

agocke (approved):
LGTM aside from nit

PR #2152, opened 2021-07-19
Fixes a null reference when trying to get RUC on a type

vitek-karas:
If the type in question has unresolvable base type, we null ref.


src/linker/Linker/Annotations.cs
```diff
@@ -561,14 +561,17 @@ public bool SetPreservedStaticCtor (TypeDefinition type)
...
+			if (type == null)
+				return false;
+
 			do {
 				if (TryGetLinkerAttribute (type, out RequiresUnreferencedCodeAttribute declaringTypeAttribute)) {
 					attribute = declaringTypeAttribute;
 					return true;
 				}
 				type = type.DeclaringType;
 			} while (type != null);
```
marek-safar, line 574: What about changing do-while to `while (type != null)` instead?

vitek-karas: Fixed

marek-safar (approved):

PR #2155, opened 2021-07-19
Add ILLink.Shared shared project

mateoatr:
We already share code between analyzer and linker, this only adds a shared project that makes intentions more transparent and files easily discoverable inside VS. Moving forward with the analyzer, I think we'd like to share more and more code between the two; for one, I'd like to share as much as possible related to the diagnostic creation / warnings.

### DIFF
--- a/illink.sln
+++ b/illink.sln
@@ -33,6 +33,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "tlens", "src\tlens\tlens.cs
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ILLink.CodeFixProvider", "src\ILLink.CodeFix\ILLink.CodeFixProvider.csproj", "{6D20F334-B7E4-4585-854B-8A0E2B29B4AA}"
 EndProject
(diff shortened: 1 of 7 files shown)


tlakollo:
So initially the ILLink.Shared was a project, but was decided to not keep it as one see: https://github.com/mono/linker/pull/2098#discussion_r653308654
So far I think the additional resources work, is there anything different you want to make that would be easier if we had a project?

mateoatr:
> So initially the ILLink.Shared was a project, but was decided to not keep it as one see: [#2098 (comment)](https://github.com/mono/linker/pull/2098#discussion_r653308654)
> So far I think the additional resources work, is there anything different you want to make that would be easier if we had a project?

If we have a supported way of sharing files within projects, I don't see any reason not to use it. I plan on adding a bunch more of diagnostics to the SharedStrings.resx file as part of the DAM analyzer (and hopefully, more code related to how we log/process diagnostics).

Also, it's nice having the file structure as clear as:
image

vs:
image

tlakollo:
I'm actually fine with having an additional project, and indeed looks easier to work with the files

tlakollo (approved):
There is some shared project logic I'm not very familiar with, but the changes look good

PR #2162, opened 2021-07-23
Fix warning origin for DAM on types

sbomer:
Fixes https://github.com/mono/linker/issues/2136.

The warnings pretty much work as described in https://github.com/mono/linker/issues/2136#issuecomment-879146153, and the scenario from https://github.com/mono/linker/issues/2136#issuecomment-881681867 will warn on the derived type. There's still an issue where the member warnings can't be suppressed on the member because we haven't marked custom attributes at that point.

### DIFF
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -128,7 +128,6 @@ namespace Mono.Linker.Dataflow
 
 		public DynamicallyAccessedMemberTypes ApplyDynamicallyAccessedMembersToTypeHierarchy (
 			ReflectionMethodBodyScanner reflectionMethodBodyScanner,
-			ref ReflectionPatternContext reflectionPatternContext,
 			TypeDefinition type)
 		{
 			Debug.Assert (_context.Annotations.IsMarked (type));
@@ -142,6 +141,8 @@ namespace Mono.Linker.Dataflow
 				return annotation;
 
 			// Apply the effective annotation for the type
+			using var _ = _scopeStack.PushScope (new MessageOrigin (type));
+			var reflectionPatternContext = new ReflectionPatternContext (_context, true, _scopeStack.CurrentScope.Origin, type);
 			reflectionMethodBodyScanner.ApplyDynamicallyAccessedMembersToType (ref reflectionPatternContext, type, annotation);
 
 			// Mark it as applied in the cache
@@ -174,7 +175,7 @@ namespace Mono.Linker.Dataflow
 					break;
 
 				foreach (var candidateType in candidateTypes) {
-					ApplyDynamicallyAccessedMembersToTypeHierarchyInner (reflectionMethodBodyScanner, ref reflectionPatternContext, candidateType);
+					ApplyDynamicallyAccessedMembersToTypeHierarchyInner (reflectionMethodBodyScanner, candidateType);
 				}
 			}
 
@@ -183,7 +184,6 @@ namespace Mono.Linker.Dataflow
 
 		bool ApplyDynamicallyAccessedMembersToTypeHierarchyInner (
 			ReflectionMethodBodyScanner reflectionMethodBodyScanner,
-			ref ReflectionPatternContext reflectionPatternContext,
 			TypeDefinition type)
 		{
 			(var annotation, var applied) = GetCachedInfoForTypeInHierarchy (type);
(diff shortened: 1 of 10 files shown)


sbomer:
I'm not seeing an easy way to fix the suppression issue so I'd like to leave that for a separate change (filed https://github.com/mono/linker/issues/2163)

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -142,6 +141,8 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
 				return annotation;
 
 			// Apply the effective annotation for the type
+			using var _ = _scopeStack.PushScope (new MessageOrigin (type));
+			var reflectionPatternContext = new ReflectionPatternContext (_context, true, _scopeStack.CurrentScope.Origin, type);
```
vitek-karas, line 145: The `true` here has a substantial effect on the RUC and DAM interactions. At the very least this should have a comment on why. It took me a long time to decide what behavior I like, so I posted it as a comment on the main review thread (it's long).

vitek-karas:
Actually the interaction with RUC go so complex that I just added a design doc (it will be easier to discuss that in its own PR): https://github.com/mono/linker/pull/2168

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
sbomer: The true isn't really new - we have been reporting these warnings - but the origin is different than before. I added a comment here, and in MarkStep where the interesting decisions about RUC are made.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -75,6 +75,16 @@ public DynamicallyAccessedMemberTypes GetTypeAnnotation (TypeDefinition type)
...
 
+		public bool DoesMemberAccessRequireDynamicallyAccessedMembers (IMemberDefinition provider) =>
+			provider switch {
+				MethodDefinition method =>
+					GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var annotation) &&
+					(annotation.ParameterAnnotations != null || annotation.ReturnParameterAnnotation != DynamicallyAccessedMemberTypes.None),
+				FieldDefinition field =>
+					GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out _),
+				_ => false
+			};
```
vitek-karas, line 86: In my change there are very similar methods added as well. Please make sure we keep only one copy of these :smile:

test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyReflectionWarnings.cs
```diff
@@ -0,0 +1,444 @@
...
+			)
+			{ }
+		}
+
+		[Kept]
+		[KeptAttributeAttribute (typeof (DynamicallyAccessedMembersAttribute))]
+		[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicMethods)]
+		class AnnotatedPublicMethods
+		{
+			[Kept]
```
vitek-karas, line 95: I would add a DAM field and check that it doesn't produce a warning in this case (if you already have test cases for All and PublicMethods and so on).

test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyReflectionWarnings.cs
```diff
@@ -0,0 +1,444 @@
...
+			[KeptAttributeAttribute (typeof (RequiresUnreferencedCodeAttribute))]
+			[ExpectedWarning ("IL2112", "--RUC on AnnotatedBase--")]
+			[RequiresUnreferencedCode ("--RUC on AnnotatedBase--")]
+			public void RUCMethod () { }
+		}
+
+		[KeptBaseType (typeof (AnnotatedBase))]
+		[ExpectedWarning ("IL2113", "--RUC on AnnotatedBase--")]
```
vitek-karas, line 204: Ideally we would not produce this warning - there's nothing wrong the derived type did in this case - all the blame should be on the base type - which already generates a warning for this anyway.

I think it's OK to not fix this if it's hard, but logically it should not happen.

vitek-karas: (As in create an issue if it's too hard to fix)

test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyReflectionWarnings.cs
```diff
@@ -0,0 +1,444 @@
...
+		{
+			[Kept]
+			[KeptAttributeAttribute (typeof (RequiresUnreferencedCodeAttribute))]
+			[ExpectedWarning ("IL2112", "--RUC on DerivedFromAnnotatedBase--")]
+			[RequiresUnreferencedCode ("--RUC on DerivedFromAnnotatedBase--")]
+			public void RUCMethod () { }
+		}
+
+		[KeptMember (".ctor()")]
+		[KeptMember ("get_DAMTVirtualProperty()")]
```
vitek-karas, line 215: Nit: You can add `[Kept]` onto the getter I think - it should work the same.

vitek-karas (approved):
Looks good!

test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyReflectionWarnings.cs
sbomer: I'll leave that for a follow-up: https://github.com/mono/linker/issues/2175

test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyReflectionWarnings.cs
sbomer: Done!

PR #2164, opened 2021-07-23
Avoid the propagation of RequiresUnreferencedCode to nested classes

### LINKED ISSUE #2151: Implications of RUC on nested types
We currently consider types nested under RUC types as RUC themselves.

This means:

```csharp
[RUC]
class Foo
{
    // [RUC] is implied
    class Nested { }

    // Treat as RUC as well?
    struct Bar { }

    // Treat as RUC as well?
    interface Baz { }

    // Treat as RUC as well?
    deleate Woo();

    // Also RUC?
    enum MyEnum { One, Two }
}
```

I see two options:

1. Treat RUC the same as 99% other attributes that don't get inherited from their owning type.
2. Have non-class nested types "break" the RUC inheritance chain (types nested under them no longer have RUC either, including if there's any classes) and document the weirdness.

I prefer the first option because it's simpler and I like simple, but I also don't know what lead us to deciding we want it to be inherited in the first place (the inheritance wouldn't be my first choice).

Cc @tlakollo @vitek-karas @agocke

tlakollo:
Avoid the propagation of RequiresUnreferencedCode to nested classes
When suppressing warnings only ask the declaring type instead of the effective type
Fixes #2151

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1768,10 +1768,7 @@ namespace Mono.Linker.Steps
 				MarkType (type.DeclaringType, new DependencyInfo (DependencyKind.DeclaringType, type));
 			MarkSecurityDeclarations (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
 
-			if (((type.DeclaringType is not null &&
-				!_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (type.DeclaringType)) ||
-				type.DeclaringType is null) &&
-				type.BaseType is not null &&
+			if (type.BaseType is not null &&
 				!_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (type) &&
 				_context.Annotations.TryGetLinkerAttribute (_context.TryResolve (type.BaseType), out RequiresUnreferencedCodeAttribute effectiveRequiresUnreferencedCode)) {
 
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1768,9 +1768,13 @@ namespace Mono.Linker.Steps
 				MarkType (type.DeclaringType, new DependencyInfo (DependencyKind.DeclaringType, type));
 			MarkSecurityDeclarations (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
 
-			if (type.BaseType != null &&
-				!_context.Annotations.TryGetEffectiveRequiresUnreferencedCodeAttributeOnType (type, out RequiresUnreferencedCodeAttribute _) &&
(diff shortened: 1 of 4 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1768,9 +1768,13 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 				MarkType (type.DeclaringType, new DependencyInfo (DependencyKind.DeclaringType, type));
 			MarkSecurityDeclarations (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
 
-			if (type.BaseType != null &&
-				!_context.Annotations.TryGetEffectiveRequiresUnreferencedCodeAttributeOnType (type, out RequiresUnreferencedCodeAttribute _) &&
-				_context.Annotations.TryGetEffectiveRequiresUnreferencedCodeAttributeOnType (_context.TryResolve (type.BaseType), out RequiresUnreferencedCodeAttribute effectiveRequiresUnreferencedCode)) {
+			if (((type.DeclaringType is not null &&
+				!_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (type.DeclaringType)) ||
```
vitek-karas, line 1772: Just to capture our offline discussion:
```C#
[RequiresUnreferencedCode("test")]
class RUCBase {}

[RequiresUnreferencedCode("test")]
class OuterType
{
    // Expecting warning here - as the RUC on OuterType should not apply to anything on nested types
    class Derived : RUCBase
    {
    }
}
```

agocke (approved):
LGTM

tlakollo:
Changed the different checks to call DoesMethodRequireUnreferencedCode to prevent having inconsistent calls, also moved the DoesMethodRequireUnreferencedCode method from MarkStep to the Annotations file

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2903,13 +2903,13 @@ internal bool ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ()
...
-			if (suppressionContextMember != null &&
-				Annotations.DoesMethodRequireUnreferencedCode (suppressionContextMember, out _))
+			if (suppressionContextMember is not null && suppressionContextMember is MethodDefinition &&
+				Annotations.DoesMethodRequireUnreferencedCode ((MethodDefinition) suppressionContextMember, out _))
 				return true;
 
 			IMemberDefinition originMember = currentOrigin.MemberDefinition;
-			if (suppressionContextMember != originMember && originMember != null &&
-				Annotations.DoesMethodRequireUnreferencedCode (originMember, out _))
+			if (originMember is not null && originMember is MethodDefinition && suppressionContextMember != originMember &&
```
mateoatr, line 2911: It is sufficient to have `originMember is MethodDefinition` (same above for suppressionContextMember.

PR #2170, opened 2021-07-26
Share DiagnosticStrings add mechanism to generate a DiagnosticDescriptor given a DiagnosticId

mateoatr:
Share `DiagnosticString`, `DiagnosticId`, and `DiagnosticCategory` between linker and analyzer. `DiagnosticString` receives a readable diagnostic id (a resource key which maps to exactly one `ILXXXX` code) and returns its diagnostic message format; this way we might move warnings in the linker side to a simpler logging mechanism: `context.Log(DiagnosticId, params string[] args, ...)`. 

The Roslyn Analyzer uses this to get a diagnostic descriptor given a diagnostic id; this saves some diagnostic descriptors handwriting; useful going forward with supporting more diagnostics on the analyzers.


src/ILLink.Shared/DiagnosticId.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace ILLink.Shared
+{
+	public enum DiagnosticId
+	{
+		// Linker diagnostic ids.
+		RequiresUnreferencedCode = 2026,
+		RequiresAttributeMismatch = 2046,
```
tlakollo, line 7: Nit: Maybe is better to rename this to RequiresUnreferencedCodeAttributeMistmatch to match the RequiresAssemblyFilesAttributeMismatch format and be more verbose about the difference
```suggestion
		RequiresUnreferencedCodeAttributeMismatch = 2046,
```

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -13,30 +13,13 @@ namespace ILLink.RoslynAnalyzer
...
-			IL2046,
-			new LocalizableResourceString (nameof (SharedStrings.RequiresAttributeMismatchTitle),
-			SharedStrings.ResourceManager, typeof (SharedStrings)),
-			new LocalizableResourceString (nameof (SharedStrings.RequiresAttributeMismatchMessage),
-			SharedStrings.ResourceManager, typeof (SharedStrings)),
-			DiagnosticCategory.Trimming,
-			DiagnosticSeverity.Warning,
-			isEnabledByDefault: true);
+		static readonly DiagnosticDescriptor s_requiresUnreferencedCodeRule = DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.RequiresUnreferencedCode);
+		static readonly DiagnosticDescriptor s_requiresAttributeMismatch = DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.RequiresAttributeMismatch,
```
tlakollo, line 20: Is there any reason why we called the constructor with parameters instead of only passing DiagnosticId.RequiresAttributeMismatch?

illink.sln
```diff
@@ -36,6 +36,11 @@ EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "ILLink.Shared", "src\ILLink.Shared\ILLink.Shared.shproj", "{92F5E753-2179-46DC-BDCE-736858C18DC7}"
 EndProject
 Global
+	GlobalSection(SharedMSBuildProjectFiles) = preSolution
+		src\ILLink.Shared\ILLink.Shared.projitems*{92f5e753-2179-46dc-bdce-736858c18dc7}*SharedItemsImports = 13
```
tlakollo, line 40: To educate myself whats this code doing?

tlakollo (approved):
Definitely a very clever of solving the diagnostics, thanks Mateo!

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
mateoatr: None! This was a left over of a previous commit (sigh).

src/ILLink.Shared/DiagnosticId.cs
mateoatr: Changed it together we the warning message, PTAL.

src/ILLink.Shared/DiagnosticString.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Resources;
+
+namespace ILLink.Shared
+{
+	public readonly struct DiagnosticString
+	{
+		readonly string _titleFormat;
+		readonly string _messageFormat;
+		static readonly ResourceManager _resourceManager = SharedStrings.ResourceManager;
```
agocke, line 9: Why do we need a static field here? It seems like a local variable would suffice

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
```diff
@@ -234,7 +235,8 @@ private static string FormatDiagnostics (ImmutableArray<DiagnosticAnalyzer> anal
 				var applicableAnalyzer = analyzers.FirstOrDefault (a => a.SupportedDiagnostics.Any (dd => dd.Id == diagnosticsId));
 				if (applicableAnalyzer != null) {
 					var analyzerType = applicableAnalyzer.GetType ();
-					var rule = diagnostics[i].HasLocation && applicableAnalyzer.SupportedDiagnostics.Length == 1 ? string.Empty : $"{analyzerType.Name}.{diagnosticsId}";
+					var rule = diagnostics[i].HasLocation &&
+						applicableAnalyzer.SupportedDiagnostics.Length == 1 ? string.Empty : $"DiagnosticId.{(DiagnosticId) Int32.Parse (diagnosticsId.Substring (2))}";
```
agocke, line 239: Extract helper method for this usage and below?

src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
```diff
@@ -0,0 +1,68 @@
...
+					diagnosticString.GetMessageFormat (),
+					diagnosticCategory ?? GetDiagnosticCategory (diagnosticId),
+					diagnosticSeverity,
+					isEnabledByDefault,
+					helpLinkUri);
+			}
+
+			return new DiagnosticDescriptor (diagnosticId.AsString (),
+				lrsTitle!,
+				lrsMessage!,
```
agocke, line 45: What if one of the title or message is `null`?

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
```diff
@@ -19,7 +20,7 @@ namespace ILLink.CodeFix
 	public class RequiresUnreferencedCodeCodeFixProvider : BaseAttributeCodeFixProvider
 	{
 		public sealed override ImmutableArray<string> FixableDiagnosticIds
-			=> ImmutableArray.Create (RequiresUnreferencedCodeAnalyzer.IL2026);
+			=> ImmutableArray.Create (DiagnosticId.RequiresUnreferencedCode.AsString ());
```
agocke, line 23: Consider storing the `SupportedDiagnostics` of the `RequiresUnreferencedCodeAnalyzer` as a static property, then calling that property to get the full list of diagnostic IDs for this.

Same idea for `RequiresAssemblyFilesCodeFixProvider`.

agocke (approved):
Minor comments. LGTM otherwise

illink.sln
mateoatr: This code was generated by VS. Couldn't find anything particularly interesting; looks like the first entry (13) is referencing the shared project itself and the bellow ones are the projects that use the shared proj.

Issue #2163, opened 2021-07-23
Warnings for DAM on type can't be suppressed at the member level


sbomer:
With the fix for https://github.com/mono/linker/issues/2136, we will start producing warnings for reflection access to annotated members by DAMT on types, where the warning origin will be the annotated member itself. These warnings aren't always suppressible on the member itself.

Architectural limitations make this a bit hard to solve (or maybe I'm lacking some creativity - ideas welcome!):
- We need to warn for every potential access to the member (from DAMT annotated types in this case)
- We decide whether a warning is suppressed when generating it
- We only want to scan the member's attributes once
  - And this should happen after marking the member "processed", otherwise it can trigger re-processing of the member leading to stack overflow.
  - But at the point where we process methods we have lost information about what originally caused the method to be kept

I think this points to two possible changes that would fix the suppressions:
- When we unify the DependencyInfo/ScopeStack infrastructure we should try to track enough information to decide what kind of warnings to report
- We should consider not treating warnings suppressed until we are ready to emit them, as @agocke has suggested before

MichalStrehovsky:
I haven't been a big fan of seeing DAM processing getting moved to MarkStep and relying on marking order/DependencyInfo/ScopeStack. It doesn't translate well to the parallel processing model in NativeAOT.

A custom attribute walk is less expensive than the walk over all instructions that we're doing. It feels like we're making things more complicated just to avoid walking attributes multiple times.

E.g. just yesterday I was scratching my head how I would port https://github.com/mono/linker/pull/2070 to NativeAOT and I'm pretty much decided I'm just going to skip it. The extra information that we're piping through can be computed anytime we need it without having to worry we forgot to update the state (e.g. `UpdateCurrentScopeInstructionOffset` looks kind of like a red flag to me).

From my outside view, it looks like things took the wrong turn when we were about to do warning suppressions from compiler generated code using marking order and that approach ended up not being workable, but we left in preparatory changes that would not be needed - it feels like the design decisions in #2054 and co were largely based on the fact that we wanted marking order to do suppressions in addition to what's in custom attributes.

Now we have all these problems that whether a warning gets suppressed depends on the marking order - when the suppressions would be clear if we could just walk things and look at their attributes.

vitek-karas:
There's another example of basically the same underlying problem: https://github.com/dotnet/runtime/issues/56252
[...]

Issue #2095, opened 2021-06-16
Wrong message about duplicate suppression attributes


vitek-karas:
If a type is marked as a whole (via `MarkEntireType`) and it's also marked the normal way (`MarkType`) its custom attributes will be processed twice which leads to registering the same suppression attributes twice. This generates an Info message like:

```console
ILLink: Element <typename> has more than one unconditional suppression. Note that only the last one is used.
```

Why is this a problem:
* Noise - the output is wrong and confusing. But it's only visible when `--verbose` is turned on, which is not the default.
* Performance - there's no need to process the attributes twice.

There's a bigger underlying problem which we've ran into couple of times already:
`MarkEntireType` is not a superset of `MarkType` and so it doesn't mark the type as processed. Right now this seems to work almost always but mostly by accident because it's VERY likely that `MarkEntireType` will eventually call `MarkType` as a result of marking members on the type.

There might be cases where this is wrong (types without members for example might cause trouble in some specific cases).

vitek-karas:
https://github.com/mono/linker/pull/2096 adds a test for the repro case.

PR #2171, opened 2021-07-27
Check attribute suppressions lazily

sbomer:
Previously attribute suppressions were populated while marking
the attributes. This meant that any suppression checks before the
attributes got marked would not see the suppressions. Doing this
lazily fixes the problem, but may result in scanning the attributes
multiple times.

Fixes https://github.com/mono/linker/issues/2163
FIxes https://github.com/mono/linker/issues/2095
Fixes https://github.com/dotnet/runtime/issues/56252

This also fixes an issue where we were not respecting
UnconditionalSuppressMessageAttribute injected via
XML for anything other than assembly/module level attributes.

### DIFF
--- a/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
+++ b/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Mono.Cecil;
@@ -212,7 +212,7 @@ namespace Mono.Linker
 			}
 		}
 
-		public static bool TypeRefHasUnconditionalSuppressions (TypeReference typeRef)
+		static bool TypeRefHasUnconditionalSuppressions (TypeReference typeRef)
 		{
 			return typeRef.Name == "UnconditionalSuppressMessageAttribute" &&
 				typeRef.Namespace == "System.Diagnostics.CodeAnalysis";
--- a/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
+++ b/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
@@ -22,7 +22,7 @@ namespace Mono.Linker
 			InitializedAssemblies = new HashSet<AssemblyDefinition> ();
 		}
 
-		public void AddSuppression (CustomAttribute ca, ICustomAttributeProvider provider)
+		void AddSuppression (CustomAttribute ca, ICustomAttributeProvider provider)
 		{
 			SuppressMessageInfo info;
 			if (!TryDecodeSuppressMessageAttributeData (ca, out info))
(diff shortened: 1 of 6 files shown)


src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -90,7 +90,19 @@ bool IsSuppressedOnElement (int id, ICustomAttributeProvider provider, out Suppr
...
+				return false;
+
+			foreach (var ca in _context.CustomAttributes.GetCustomAttributes (provider)) {
+				if (TypeRefHasUnconditionalSuppressions (ca.Constructor.DeclaringType) &&
+					provider is not ModuleDefinition or AssemblyDefinition)
+					AddSuppression (ca, provider);
+			}
+
+			return _suppressions.TryGetValue (provider, out suppressions) &&
 				suppressions.TryGetValue (id, out info);
```
mateoatr, line 106: Do we have to do this a second time? Can't we keep info of whether we saw any suppression on the above loop and return that?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -22,7 +22,7 @@ public UnconditionalSuppressMessageAttributeState (LinkContext context)
 			InitializedAssemblies = new HashSet<AssemblyDefinition> ();
 		}
 
-		public void AddSuppression (CustomAttribute ca, ICustomAttributeProvider provider)
```
mateoatr, line 25: Nit: Now that we moved the CA walk here, should we make `TypeRefHasUnconditionalSuppressions` private as well (same applies to `LookForModuleLevelAndGlobalSuppressions`.

mateoatr (approved):
LGTM!

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
sbomer: We would need to track the decoded info as we go in addition to whether we saw any so it would look something like this: https://github.com/sbomer/linker/compare/suppressionsFix...sbomer:suppressionsFixFeedback?expand=1. I personally find this less readable but LMK which you prefer.

vitek-karas: I think it's OK to probe the cache twice. We could only call the `suppression.TryGetValue` once, but it's not a big deal.

test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInMembersAndTypes.cs
```diff
@@ -103,9 +103,7 @@ public void SuppressionHasNullParameters ()
 
 	class SuppressOnTypeMarkedEntirely
 	{
-		// https://github.com/mono/linker/issues/2095
-		// [LogDoesNotContain (nameof(TypeWithSuppression) + " has more than one unconditional suppression")]
-		[LogContains (nameof (TypeWithSuppression) + " has more than one unconditional suppression")]
+		[LogDoesNotContain (nameof (TypeWithSuppression) + " has more than one unconditional suppression")]
```
vitek-karas, line 106: Can you please also add a test case for https://github.com/dotnet/runtime/issues/56252.
Basically a "copy" assembly which has compiler generated method with a suppression - in that case we don't guarantee order of marking between the method and compiler generated code - and that broke suppressions.

vitek-karas (approved):
Other than the test for https://github.com/dotnet/runtime/issues/56252 this looks good!

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
mateoatr: Thanks for the example! I agree it's not a big deal and makes it less readable. I'm fine with either approach.

PR #2176, opened 2021-07-28
Type hierarchy merge

sbomer:
This contains the changes from https://github.com/mono/linker/pull/2145 and https://github.com/mono/linker/pull/2162. PTAL at the last commit @vitek-karas. Once we have the runtime changes ready I would suggest we merge your change in, then I will rebase my original PR with this change included and we can check it in immediately (that way the repo commit history will be cleaner since the changes are logically separate).

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyReflectionWarnings.cs
+++ b/test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyReflectionWarnings.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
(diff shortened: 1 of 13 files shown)


test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyReflectionWarnings.cs
```diff
@@ -146,9 +166,13 @@ class AnnotatedPublicProperties
 			[KeptAttributeAttribute (typeof (DynamicallyAccessedMembersAttribute))]
 			[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicMethods)]
 			public static string DAMProperty {
-				// Property access reports warnings on getter/setter
-				[ExpectedWarning ("IL2114", nameof (AnnotatedPublicProperties), nameof (DAMProperty) + ".get")]
+				[Kept]
+				// No warning for getter since return value is not annotated
```
vitek-karas, line 170: The return value is annotated, but because the getter is not virtual it doesn't matter. If the property was virtual this should/would warn.

vitek-karas (approved):
Looks good

sbomer:
@vitek-karas here are the updated runtime suppressions: https://github.com/dotnet/runtime/compare/main...sbomer:updateSuppressionsForReflectionAccess?expand=1.

I took a few of them straight from your changes in https://github.com/dotnet/runtime/compare/main...vitek-karas:FixReflectionOnDAMImpact?expand=1.

PR #2177, opened 2021-07-28
Add document explaining RUC on types usage and scenarios

tlakollo:


docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
...
+class ClassWithRequiresUnreferencedCode
+{
+    public static void StaticMethod () { }
+
+    public void NonStaticMethod () { }
+}
+
+static void TestRequiresUnreferencedCodeInClass ()
+{
+    // IL2026: TestRequiresUnreferencedCodeInClass(): Using method 'ClassWithRequiresUnreferencedCode.StaticMethod()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Message for --ClassWithRequiresUnreferencedCode--.
```
vitek-karas, line 18: More of a side note - we should probably tweak the message for IL2026 to include the option of having RUC on the type.

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
...
+static void TestRequiresOnBaseButNotOnDerived ()
+{
+    DerivedWithoutRequires.StaticMethodInInheritedClass ();
+    DerivedWithoutRequires.StaticMethod ();
+}
+```
+As the previous example ilustrates the DerivedType without RequiresUnreferencedCode will generate a warning only for deriving from ClassWithRequiresUnreferencedCode. Additionally, by calling some of the methods we can see the difference on behavior on methods declared on the derived type vs methods on the base type. Calling a method in the derived type will not generate a warning since the derived type is not annotated, while calling a method for which the base type will be called will generate IL2026.
+
+Arguably we could say that the warning in the type declaration is not necessary and that we could rely on the warning being generated in the constructor instead, but this could lead to some behaviors in which we can call a method without ever using the constructor.
```
vitek-karas, line 134: The argument was that there's nothing wrong about not getting a warning in this case. The Derived.SomeMethod doesn't have any dangerous code in it - if it did it would get a warning of its own.

Assuming we can "trust" in us detecting all calls to base (specifically that `.ctor` on Derived will always call `.ctor` on Base), then I don't see a need for a warning on inheritance.

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
...
+- The attribute is too dificult to detect when the static constructor is being called. This has the advantage of not having to understand the rules for .cctor invocation nor to detect static field accesses
+- Placing RUC on a .cctor would generate a warning saying that is not valid to put RUC on a .cctor
+- RUC on types would not silence warnings from static constructor methodbody
+- The only options available for a developer would be to silence the warnings (seems not the best approach since we know there is a warning inside the methodbody) or refactor the code in such a way the .cctor no longer has to deal with the warning. But then we must rewrite code when cctor has RUC code
+  - how? what are some common patterns and how do we rewrite them?
+  - rewriting field to lazy-init property is a breaking change
+  - rewriting field to Lazy<T> will still warn from the cctor, won't solve the problem
+
+### Approach 2
+Allow RUC on static constructors. Need to detect anything that can trigger cctor and warn if cctor is RUC
```
vitek-karas, line 185: There is a 3rd option - don't allow RUC on `.cctor`, but when RUC is on type, treat it as if it's on `.cctor` as well. The "advantage" of this approach is that it feels more predictable. Even if linker can understand all the various rules around `.cctor` invocation, normal developers won't. So putting RUC on `.cctor` would generate warnings from seemingly "random" places for the developer. If instead we "force" the RUC on type in these cases, it's a more predictable behavior.

mateoatr: Agreed. This approach would also make things more intuitive when unconditionally suppressing the warnings; otherwise, I think we'd also need to decide how suppressions work on RUC annotated types. It would be somewhat weird if RUC on a type silenced all warnings on calls to instance members but .cctors produced warnings from different places -- I'd think that most devs would then try to unconditionally suppress on the type (which, as of today, would work).

sbomer: I think with both Approach 2 or this 3rd option we have to at least warn on the accesses that might trigger a cctor. For example, either way:
- For beforefieldinit types, static field access should warn
- For non-beforefieldinit types, access to static field/method or ctor should warn
  - (If we allow RUC on structs, then access to non-static methods on structs should warn also, but this seems really hard: https://github.com/mono/linker/pull/2143#discussion_r671595464)

I think where we have some freedom is:
- Should we try to detect the first field access (at least within a method)? Or warn on all of them?
- Should we conservatively warn on static method access or ctors even for beforefieldinit types?

Or going even further, should we warn on every reference to a type (that isn't inside of the RUC scope)?:
- Access to virtual/instance methods
- Access to instance fields
- `typeof(Foo)` reference to the type?
- Deriving from a RUC type?

The "warn on every reference" behavior seems closest to the Obsolete behavior, and I find it the most intuitive, but it will produce more warnings. What's not clear to me is whether the extra warnings will be problematic - see the discussion here: https://github.com/mono/linker/pull/2143#discussion_r671025458.

sbomer (commented):
Really nice, thanks for writing this down!

docs/design/RUC-on-type.md
vitek-karas: Originally I though that it would make sense to model Obsolete behavior, but it's lot less close to what we want than I expected. Warning on instance members feels like just noise - since we will warn on .ctors no matter what.

I don't think we should warn on typeof(Foo) - one reason is that linker may actually decide to remove the type if it's not used anywhere else.

Deriving from RUC type - I would start without warning if we can get away with it - that is if we can guarantee that derived .ctor will always have a call to base .ctor. I don't know IL well enough to answer that though.

For now I would not bother trying to detect first access to anything - it would have to be only inside a single method's body and the noise is not that big in my opinion. It's similar to what we do with RUC methods - if you call the same RUC method multiple times from the same method body, you'll get multiple warnings.

I would start by not differentiation on beforefieldinit - while linker currently tracks this fully, I don't think the added value is worth it. And in some way this could create more confusion (why do I get warning on type A and not B even though I'm doing the exact same thing?).

sbomer: I should clarify that even within the "warn on every reference" there are different degrees - for example Obsolete doesn't warn on access to instance fields/methods and I think we shouldn't either. But it does warn on local declarations or default value expressions for example.

Interestingly Obsolete on type seems to warn in all cases that could trigger a static constructor (as far as I can tell), but not when the cctor itself is Obsolete - there it warns for calling the ctor, but not for accessing static fields or methods:

```csharp
var c = new C(); // warning here, but not below
var f = C.f;
C.M();

public class C {
    [Obsolete]
    public C() {}
    
    public static void M() {}
    
    public static int f;
}
```

We definitely want to warn here to avoid holes (if supporting RUC cctors) even though Obsolete doesn't.

I understand the general point in https://github.com/mono/linker/pull/2143#discussion_r671025458, but still don't understand what is the value in allowing `typeof(RUCType)` or locals of type annotated with RUC (I'm having a hard time imagining what you could usefully do with these). But I do think there is value in the conceptual simplicitly of the principle that anything annotated with RUC will be removed by the linker, if there are no warnings or suppressions.

> I don't think we should warn on typeof(Foo) - one reason is that linker may actually decide to remove the type if it's not used anywhere else.

The typeof(Foo) would result in the type being kept - I assume you mean that the linker may decide to remove its members?

sbomer (commented):
We should add another section about the virtual method validation: https://github.com/mono/linker/issues/2179

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
+# RequiresUnreferencedCode on Types
+The RequiresUnreferencedCode being allowed on types its a feature that allows developers to quickly mark entire classes as dangerous for trimming. instead of having to mark every method in the class. This document will go over whats the behavior of certain scenarios when the RequiresUnreferencedCode attribute is placed on a type
+
+## RequiresUnreferencedCode on a type warns on instance constructor or static method access
+Having RequiresUnreferencedCode on the type implies that any instance constructor or static method is considered dangerous for trimming. Following are examples of this behavior:
```
agocke, line 5: Let's be a bit more technical. My description would be:

If `RequiresUnreferencedCode` is present on a class declaration, the behavior of the following members behave as if they were annotated directly:

- Static methods
- Instance constructors
- Static properties
- Static events

Members which are not affected are:

- Nested types
- Instance members

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
+# RequiresUnreferencedCode on Types
```
agocke, line 1: Let's make this a spec for `RequiresUnreferencedCode`, all together. No need to just talk about types.

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
...
+The RequiresUnreferencedCode being allowed on types its a feature that allows developers to quickly mark entire classes as dangerous for trimming. instead of having to mark every method in the class. This document will go over whats the behavior of certain scenarios when the RequiresUnreferencedCode attribute is placed on a type
```
agocke, line 2:
```suggestion
RequiresUnreferencedCode is a feature for providing and suppressing diagnostics related to trimming. In general, when accessing a member attributed with `RequiresUnreferencedCode`, a warning should be produced at the access location that the member is not compatible with trimming. In addition, when a member is annotated with `RequiresUnreferencedCode`, trimming compatibility warnings will be suppressed in the body of that member.
```

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
...
+    // IL2026: TestRequiresUnreferencedCodeInClass(): Using method 'ClassWithRequiresUnreferencedCode.StaticMethod()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Message for --ClassWithRequiresUnreferencedCode--.
+    ClassWithRequiresUnreferencedCode.StaticMethod ();
+
+    ClassWithRequiresUnreferencedCode.NestedClass.StaticMethod ();
+}
+
+```
+In the previous example calling ClassWithRequiresUnreferencedCode.StaticMethod () is considered dangerous since the declaring type of the static method is annotated with RequiresUnreferencedCode. For ClassWithRequiresUnreferencedCode.NestedClass.StaticMethod () the declared type NestedClass is not considered dangerous since it doesn't contain the RequiresUnreferencedCode on it nor the method contains the attribute.
+
+If we want ClassWithRequiresUnreferencedCode.NestedClass.StaticMethod () to generate the warning we would need to annotate either the NestedClass type or the ClassWithRequiresUnreferencedCode.NestedClass.StaticMethod static method.
```
agocke, line 86: I think much of this text is more of a guide than a spec. There's nothing wrong with guides, but specs can be much shorter and just list the exact behavior that the implementers need to know.

docs/design/RUC-on-type.md
agocke: > I think where we have some freedom is:

> Should we try to detect the first field access (at least within a method)? Or warn on all of them?

This sounds like it's purely about reducing repeated warnings. I would count this as the same question as, "should we warn on repeated calls to the same static method in particular method body?"

> Should we conservatively warn on static method access or ctors even for beforefieldinit types?

For simplicity and easy porting to a Roslyn analyzer, I would say yes.

> Or going even further, should we warn on every reference to a type (that isn't inside of the RUC scope)?:

> Access to virtual/instance methods
> Access to instance fields
> typeof(Foo) reference to the type?
> Deriving from a RUC type?

I don't think these questions are useful without a discussion of the hole being plugged. Access to virtual/instance methods and instance fields seem not worthy of a warning, for example, because the user must have already received a warning on calling the constructor of the type. On the other hand, deriving from an RUC type seems purely like a consequence of warning on constructor calls.

vitek-karas: To me this is about approach to warning on RUC on type.

"type approach" - We either say - any access to the type is problematic (very similar to what `Obsolete` does in the analyzer) - the implementation of this would warn from `MarkType`. But that runs into some interesting problems like warning when accessing instance members and so on. This has lot more chances of producing lot of warning noise.

"members approach" - Or we state that only access to some of the type's members is problematic (static things and .ctors) - the implementation of this would basically not warn from `MarkType` at all. This has already mentioned potential issues with analysis holes.

There are some interesting corner cases in both approaches, but overall it's pretty simple distinction. Based on our discussions of what the desired behavior should be, I'm leaning towards the "members" approach. I assume we can figure out the holes in analysis and fix them. Either approach should work reasonably well for the analyzer (although it's true that members approach is probably a bit more complicated to implement).

> The typeof(Foo) would result in the type being kept - I assume you mean that the linker may decide to remove its members?

You're right that `typeof(Foo)` will mark the type. I got it confused with `instance is Foo` - this will [not mark](https://github.com/mono/linker/blob/aec70d0ae5a5d304176bb9831f5ffdb6e13d3568/src/linker/Linker.Steps/MarkStep.cs#L3474) `Foo` in all cases.

docs/design/RUC-on-type.md
vitek-karas: I would not block on this for now - we need to resolve RUC-on-type behavior soon, so that we can implement it for .NET 6. Documenting the existing behavior of RUC on methods is definitely something we should do, but not part of this particular effort.

docs/design/RUC-on-type.md
MichalStrehovsky: It would be nice if we could avoid behaviors that look like "weird magic".

```csharp
// Doesn't warn
Console.WriteLine(Foo.X);

[RUC]
class Foo
{
    public static int X;
    public static int Y;
}
```

Versus:

```csharp
// Now warns because we initialize an unrelated variable?
Console.WriteLine(Foo.X);

[RUC]
class Foo
{
    public static int X;
    public static int Y = 1234;
}
```

To achieve that, I think we should warn on any static field access, whether there is or isn't a cctor (if we go with the option that RUC on type applies to the cctor as well).

MichalStrehovsky: > Interestingly Obsolete on type seems to warn in all cases that could trigger a static constructor (as far as I can tell), but not when the cctor itself is Obsolete - there it warns for calling the ctor, but not for accessing static fields or methods

The example that Sven used places the Obsolete on the instance constructor, not static constructor. As far as I can tell, placing Obsolete on a static constructor has no effect at anything.

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
...
+    // Dangerous code that generetares a trimming warning
+
+static void TestSuppressionOnMethod ()
+{
+    // IL2026: TestSuppressionOnMethod(): Using method 'MethodWithDangerousCode()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. This method calls dangeours code.
+    MethodWithDangerousCode()
+}
+```
+
+Similarly RequiresUnreferencedCode on a type suppresses other warnings, with the difference that it will silence warnings from all methods in the type and attributes (except maybe static constructors, pending discussion)
```
MichalStrehovsky, line 43: I have mixed feelings about suppressing warnings from attributes. Assume `Problematic` is a problematic attribute. Maybe because the attribute itself is marked as RUC.

```csharp
[RUC]
[Problematic]
class Foo { }
```

Now I do: `typeof(Foo).GetCustomAttribute(ProblematicAttribute)`. Where do I get a warning that we just executed trim-unfriendly code?

There will be similar problems with custom attributes on instance methods; static methods will be covered because we first need to reflect on the method itself and that's a warning if the method is static; maybe we could warn on reflection-accessing instance methods of RUC types and it will be fixed there too?.

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
...
+```C#
+var t = Type.GetType("Derived").GetMethod("SomeMethod").Invoke(null, null);
+
+[RequiresUnreferencedCode("This base type is dangerous")]
+class Base {}
+
+class Derived : Base 
+{
+    static void SomeMethod() {}
+}
```
MichalStrehovsky, line 145: Is this example really problematic?

Derived.SomeMethod is not marked as RUC because we didn't propagate RUC from Base. So no warnings are suppressed in it. If it didn't generate any warnings, it will be safe to reflection-invoke. If it generated warnings, it just means that the class author has to go and either selectively annotate SomeMethod as RUC, or RUC annotate the whole type. Then the reflection will warn too.

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,189 @@
...
+{
+    var classGeneric1 = new GenericClass1<ClassWithRequiresUnreferencedCode> ();
+    var classGeneric2 = new GenericClass2<ClassWithRequiresUnreferencedCode> ();
+    var classGeneric3 = new GenericClass3<ClassWithRequiresUnreferencedCode> ();
+}
+```
+For scenarios in GenericClass1 and GenericClass2 the new constraint is present which means that the constructor of the ClassWithRequiresUnreferencedCode will be called and a warning with IL2026 will be produced. In the GenericClass3 the new constraint is not present therefore the warning will not be produced.
+
+## Structs and interfaces
+We have decided to exclude to do any analysis if structs or interfaces are inside of a class marked with RequiresUnreferencedCode, this is to follow the targets of the RequiresUnreferencedCode attribute which at this moment they dont support the attribute on a struct or an interface.
```
MichalStrehovsky, line 169:
```suggestion
We have decided not to allow `RequiresUnreferencedCode` on structs and interfaces because the semantics is hard to guard - both structs and interfaces can be "instantiated" without running an explicit constructor and at that point all instance methods on them are callable, even indirectly through virtual dispatch.
```

docs/design/RUC-on-type.md
MichalStrehovsky: Thinking about it a bit more, the "members approach" makes it easier to pinpoint source locations that people can just click on.

"Types approach" would often require us to point at "line where a local is declared", "method parameter", "base type of a type". We don't have source information for that. So the warning message we would generate for places that violate the rules would have to be in the "SomeType.SomeMember in assembly Foo" category - that one is not a great user experience.

sbomer: > The example that Sven used places the Obsolete on the instance constructor, not static constructor. As far as I can tell, placing Obsolete on a static constructor has no effect at anything.

🤦 thanks for catching my typo. That makes more sense than the weird behavior I thought I was seeing.

docs/design/RUC-on-type.md
tlakollo: Deleted the scenario and the documentation about warning with IL2109

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,397 @@
+# RequiresUnreferencedCode on Types
+`RequiresUnreferencedCode` is a feature for providing and suppressing diagnostics related to trimming. In general, when accessing a member attributed with `RequiresUnreferencedCode`, a warning should be produced at the access location that the member is not compatible with trimming. In addition, when a member is annotated with `RequiresUnreferencedCode`, trimming compatibility warnings will be suppressed in the body of that member.
+
+In the case in which `RequiresUnreferencedCode` is present on a class declaration, the following members behave as if they were annotated directly with the attribute:
+- Static methods
+- Constructors
+- Static properties
+- Static events
```
vitek-karas, line 8: I completely missed the events - we don't have any code to handle those, right? This will the first time we do any warnings on events (Data flow annotations don't apply to events, and you can't put RUC directly on event). But I think the solution should be very similar to fields.

docs/design/RUC-on-type.md
```diff
@@ -0,0 +1,397 @@
...
+{
+  [RequiresUnreferencedCode("Message")]
+  public void TestMethod () { }
+}
+```
+Since `RequiresUnreferencedCode` cannot be placed in interfaces
+## Static interface methods
+
+```C#
+// IL2026: Invoker<Foo>.Invoke(): Using member 'Foo.StaticMethod()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Message for --ClassWithRequiresUnreferencedCode--.
```
vitek-karas, line 209: Nit: I would expect this warning to be reported on the `Invoker<T>.Invoke` method (open generic), not on the specific instantiation... (also the location should match that).

PR #2180, opened 2021-07-29
Fix suppression check with assembly suppressions

sbomer:
The change in https://github.com/mono/linker/pull/2171 was incorrect
because it didn't account for the possibility that the suppressions
cache already contains assembly or module suppressions for the provider.
(The same cache is used for both - the provider here is the suppression
target, not the attribute provider).

This fixes the check by re-using the cache to also track whether we have
scanned for suppression attributes on the provider. This caching is
necessary since we warn about duplicate suppressions.

Providers without any suppressions may still be scanned multiple times
since we don't cache the negative result.

I encountered this while investigating warnings in runtime, but it appears to
have been the result of an inconsistent state in my case - I don't believe
this change is necessary to fix runtime in response to the reflection warning
changes.

### DIFF
--- a/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
+++ b/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
@@ -13,35 +13,26 @@ namespace Mono.Linker
 		internal const string MessageIdProperty = "MessageId";
 
 		readonly LinkContext _context;
-		readonly Dictionary<ICustomAttributeProvider, (bool ScannedProvider, Dictionary<int, SuppressMessageInfo> Suppressions)> _suppressions;
+		readonly Dictionary<ICustomAttributeProvider, Dictionary<int, SuppressMessageInfo>> _suppressions;
 		HashSet<AssemblyDefinition> InitializedAssemblies { get; }
 
 		public UnconditionalSuppressMessageAttributeState (LinkContext context)
 		{
 			_context = context;
-			_suppressions = new Dictionary<ICustomAttributeProvider, (bool, Dictionary<int, SuppressMessageInfo>)> ();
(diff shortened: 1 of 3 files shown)


src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -34,14 +35,13 @@ void AddSuppression (CustomAttribute ca, ICustomAttributeProvider provider)
 		void AddSuppression (SuppressMessageInfo info, ICustomAttributeProvider provider)
 		{
 			if (!_suppressions.TryGetValue (provider, out var suppressions)) {
-				suppressions = new Dictionary<int, SuppressMessageInfo> ();
+				suppressions = (ScannedProvider: false, Suppressions: new Dictionary<int, SuppressMessageInfo> ());
```
vitek-karas, line 38: Did you consider just always scanning the provider when adding a record to the cache?
It is potentially "slightly" wasteful, but it would reduce the complexity of this code.
I personally would prefer that to the "half populated" cache solution.

sbomer: I took the approach we discussed where we populate the cache for all members referenced by the assembly-level attributes. PTAL.

sbomer:
@vitek-karas PTAL when you get a chance

vitek-karas (approved):
Looks good.
@mateoatr please take a look as well, since you wrote the first version of this.

mateoatr (approved):
LGTM. Thanks for adding the useful comments! Just a nit, the method access modifiers look a little bit odd on the `UnconditionalSuppressMessageAttributeState` class (I know this is my fault). For instance, I think the "Decode..." methods should be private. Could we change these?

PR #2183, opened 2021-07-30
Attribute mismatch in combination with RUC on type

### LINKED ISSUE #2179: RUC on type should affect virtual/override validation
```csharp
class Base {
	[RUC("RUC")]
	public virtual void Method() {}
}

[RUC("RUC")]
class Derived : Base {
	public override void Method() {}
}
```

Currently this warns about a mismatch on the base/override, but I think it shouldn't.

tlakollo:
Check also the declaring type when checking for attribute mismatch
Add tests
Fixes #2179

### DIFF
--- a/src/linker/Linker.Steps/ValidateVirtualMethodAnnotationsStep.cs
+++ b/src/linker/Linker.Steps/ValidateVirtualMethodAnnotationsStep.cs
@@ -41,8 +41,8 @@ namespace Mono.Linker.Steps
 		void ValidateMethodRequiresUnreferencedCodeAreSame (MethodDefinition method, MethodDefinition baseMethod)
 		{
 			var annotations = Context.Annotations;
-			bool methodHasAttribute = annotations.DoesMethodRequireUnreferencedCode (method, out _);
-			if (methodHasAttribute != annotations.DoesMethodRequireUnreferencedCode (baseMethod, out _)) {
+			bool methodHasAttribute = annotations.IsMethodInRequiresUnreferencedCodeScope (method);
+			if (methodHasAttribute != annotations.IsMethodInRequiresUnreferencedCodeScope (baseMethod)) {
 				string message = MessageFormat.FormatRequiresAttributeMismatch (methodHasAttribute,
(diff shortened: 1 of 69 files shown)


src/linker/Linker.Steps/ValidateVirtualMethodAnnotationsStep.cs
```diff
@@ -41,8 +41,8 @@ protected override void Process ()
 		void ValidateMethodRequiresUnreferencedCodeAreSame (MethodDefinition method, MethodDefinition baseMethod)
 		{
 			var annotations = Context.Annotations;
-			bool methodHasAttribute = annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method);
-			if (methodHasAttribute != annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (baseMethod)) {
+			bool methodHasAttribute = annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method) || method.DeclaringType is not null && annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method.DeclaringType);
+			if (methodHasAttribute != annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (baseMethod) || baseMethod.DeclaringType is not null && annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (baseMethod.DeclaringType)) {
```
sbomer, line 45: We already have a few places that perform similar checks to ask if a method is RUC (either directly or due to the type annotations). Elsewhere we still call TryGetEffectiveRequiresUnreferencedCodeAttributeOnType to deal with nested types but I think we are getting rid of that. Could we introduce a helper that encapsulates this logic to make sure it's consistent everywhere we call it?

sbomer (commented):
Thanks!

src/linker/Linker.Steps/ValidateVirtualMethodAnnotationsStep.cs
tlakollo: Replaced with DoesMethodRequireUnreferencedCode method

sbomer (approved):
Thank you!

PR #2184, opened 2021-07-31
Add DynamicallyAccessedMembers analyzer

mateoatr:
This PR supercedes #2038, the main differences are that this one uses an operation context where previous one used to work with syntax nodes (harder to reason about and error prone), and that this PR uses the new mechanism for grabbing diagnostic descriptors (it also updates how we generate `DAM` related warnings on the linker side -- this has been previously proposed but the main reason not to do it was that it would make it harder to customize the produced messages; since we now rely on grabbing the correct string format from the shared strings given a diagnostic id, this is no longer a problem, we can edit these strings in the resource file whenever we want).

Just like previous draft, this focuses on diagnostics concerning mismatched `DynamicallyAccessedMember` annotations (warnings [IL2067](https://github.com/mono/linker/blob/main/docs/error-codes.md#il2067-trim-analysis-target-parameter-argument-does-not-satisfy-dynamicallyaccessedmembersattribute-in-call-to-target-method-the-parameter-source-parameter-of-method-source-method-does-not-have-matching-annotations-the-source-value-must-declare-at-least-the-same-requirements-as-those-declared-on-the-target-location-it-is-assigned-to) to [IL2091](https://github.com/mono/linker/blob/main/docs/error-codes.md#il2091-trim-analysis-target-generic-parameter-generic-argument-does-not-satisfy-dynamicallyaccessedmembersattribute-in-target-method-or-type-the-generic-parameter-source-target-parameter-of-source-method-or-type-does-not-have-matching-annotations-the-source-value-must-declare-at-least-the-same-requirements-as-those-declared-on-the-target-location-it-is-assigned-to)) and the idea is still to have a feature branch with this analyzer until we are comfortable merging it with main (mainly, be 100% sure that only valid warning scenarios will produce diagnostics, and handle the scenarios where code flow analysis is needed).


src/ILLink.RoslynAnalyzer/DynamicallyAccessedMemberTypes.cs
```diff
@@ -0,0 +1,100 @@
...
+namespace System.Diagnostics.CodeAnalysis
+{
+	/// <summary>
+	/// Specifies the types of members that are dynamically accessed.
+	///
+	/// This enumeration has a <see cref="FlagsAttribute"/> attribute that allows a
+	/// bitwise combination of its member values.
+	/// </summary>
+	[Flags]
+#if SYSTEM_PRIVATE_CORELIB
```
agocke, line 13: I don't think this should ever be true

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,253 @@
...
+{
+	[DiagnosticAnalyzer (LanguageNames.CSharp)]
+	public class DynamicallyAccessedMembersAnalyzer : DiagnosticAnalyzer
+	{
+		internal const string DynamicallyAccessedMembers = nameof (DynamicallyAccessedMembers);
+		internal const string DynamicallyAccessedMembersAttribute = nameof (DynamicallyAccessedMembersAttribute);
+
+		static ImmutableArray<DiagnosticDescriptor> GetSupportedDiagnostics ()
+		{
+			var diagDescriptorsList = new List<DiagnosticDescriptor> ();
```
agocke, line 25: Use `ImmutableArray.CreateBuilder` when building Immutable arrays

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,253 @@
...
+			for (int i = 2067; i <= 2091; i++)
```
agocke, line 26: Can we the `DiagnosticId`s here (casted to `int`)? So there are names instead of numbers?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,253 @@
...
+				SymbolKind.Field => new string[] { symbol.GetDisplayName () },
+				SymbolKind.Method => new string[] { symbol.GetDisplayName () },
+				SymbolKind.TypeParameter => new string[] { symbol.GetDisplayName (), symbol.ContainingSymbol.GetDisplayName () },
+				_ => throw new NotImplementedException ($"Unsupported source or target symbol {symbol}.")
+			});
+
+			return args;
+		}
+
+		static ISymbol? GetSymbolFromOperation (IOperation? operation) =>
```
agocke, line 212: Maybe `TryGet` to signal that it might return `null`?

In general, this is an impossible function to write because IOperations can represent 0 or more symbols, so possibly failing seems like the right semantic.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,253 @@
...
+			if (GetSymbolFromOperation (returnOperation) is not ISymbol returnedSymbol)
+				return;
+
+			returnedSymbol.TryGetDynamicallyAccessedMemberTypes (out var damtOnReturnedValue);
+			if (SourceHasMatchingAnnotations (damtOnReturnedValue, damtOnReturnType, out var missingAnnotations))
+				return;
+
+			var diag = GetDiagnosticId (returnedSymbol.Kind, context.ContainingSymbol.Kind, true);
+			var diagArgs = GetDiagnosticArguments (returnedSymbol.Kind == SymbolKind.NamedType ? context.ContainingSymbol : returnedSymbol, context.ContainingSymbol, missingAnnotations);
+			context.ReportDiagnostic (Diagnostic.Create (DiagnosticDescriptors.GetDiagnosticDescriptor (diag), returnOperation.Syntax.GetLocation (), diagArgs));
```
agocke, line 114: What if the parent has `RequiresUnreferencedCode`?

agocke (commented):
This change is potentially disruptive enough that I think we need to test it against a local copy of `dotnet/runtime` and makes sure there are no unexpected warnings

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
```diff
@@ -22,6 +22,46 @@ internal static bool HasAttribute (this ISymbol symbol, string attributeName)
...
+			var damAttributeName = DynamicallyAccessedMembersAnalyzer.DynamicallyAccessedMembersAttribute;
+			AttributeData? dynamicallyAccessedMembers = null;
+			foreach (var _attribute in symbol.GetAttributes ())
+				if (_attribute.AttributeClass is var attrClass && attrClass != null &&
+					attrClass.HasName (damAttributeName)) {
+					dynamicallyAccessedMembers = _attribute;
+					break;
+				}
+
+			dynamicallyAccessedMemberTypes = (DynamicallyAccessedMemberTypes) dynamicallyAccessedMembers?.ConstructorArguments[0].Value!;
```
agocke, line 40: This cast looks wrong. If this can be null it should be `(DynamicallyAccessMemberTypes?)`

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
mateoatr: So have twenty-somethings lines of `diagDescriptorsList.Add(... (DiagnosticId.<name>))`?

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
```diff
@@ -22,6 +22,46 @@ internal static bool HasAttribute (this ISymbol symbol, string attributeName)
...
+		internal static bool TryGetDynamicallyAccessedMemberTypes (this ISymbol symbol, out DynamicallyAccessedMemberTypes? dynamicallyAccessedMemberTypes)
+		{
+			dynamicallyAccessedMemberTypes = null;
+			if (!symbol.HasAttribute (DynamicallyAccessedMembersAnalyzer.DynamicallyAccessedMembersAttribute))
+				return false;
+
+			var damAttributeName = DynamicallyAccessedMembersAnalyzer.DynamicallyAccessedMembersAttribute;
+			AttributeData? dynamicallyAccessedMembers = null;
+			foreach (var _attribute in symbol.GetAttributes ())
+				if (_attribute.AttributeClass is var attrClass && attrClass != null &&
```
sbomer, line 34: Can the attrClass ever be null? Just curious how that could happen.

Can this null check instead be `_attribute.AttributeClass?.HasName (damAttributeName) == true`?

The `HasName` check looks redundant with the `HasAttribute` check - could those be combined? Also should we include the namespace in `damAttributeName`?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
sbomer: I assume he means something like `for (int i = DiagnosticId.Start; i <= DiagnosticId.End; i++)`. Maybe we can add sentinel values equal to the first/last DAMT mismatch enum values.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,253 @@
...
+				ITypeOfOperation typeOf => typeOf.TypeOperand,
+				_ => null
+			};
+
+		static bool SourceHasMatchingAnnotations (
+			DynamicallyAccessedMemberTypes? sourceMemberTypes,
+			DynamicallyAccessedMemberTypes? targetMemberTypes,
+			out string missingMemberTypesString)
+		{
+			missingMemberTypesString = $"'{nameof (DynamicallyAccessedMemberTypes.All)}'";
```
sbomer, line 231: Might be better to set this to null if the annotations match.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,253 @@
...
+				IInstanceReferenceOperation instanceReference => instanceReference.Type,
+				IInvocationOperation invocation => invocation.TargetMethod,
+				IMemberReferenceOperation memberReference => memberReference.Member,
+				IParameterReferenceOperation parameterReference => parameterReference.Parameter,
+				IReturnOperation returnOp => GetSymbolFromOperation (returnOp.ReturnedValue),
+				ITypeOfOperation typeOf => typeOf.TypeOperand,
+				_ => null
+			};
+
+		static bool SourceHasMatchingAnnotations (
```
sbomer, line 226: Maybe rename to "SourceHasRequiredAnnotations"?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,253 @@
...
+			out string missingMemberTypesString)
+		{
+			missingMemberTypesString = $"'{nameof (DynamicallyAccessedMemberTypes.All)}'";
+			if (targetMemberTypes == null)
+				return true;
+
+			sourceMemberTypes ??= DynamicallyAccessedMemberTypes.None;
+			var missingMemberTypesList = Enum.GetValues (typeof (DynamicallyAccessedMemberTypes))
+				.Cast<DynamicallyAccessedMemberTypes> ()
+				.Where (damt => (damt & targetMemberTypes & ~sourceMemberTypes) == damt && damt != DynamicallyAccessedMemberTypes.None)
```
sbomer, line 238: Would it be possible to extract this logic from `RequireDynamicallyAccessedMembers` and share it between the linker and analyzer?

sbomer (commented):
Sorry it took me a while to get to this. Looking good!

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
mateoatr: Ah -- that makes sense. I don't know why I thought this referred to the below line.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,253 @@
...
+				if (SourceHasMatchingAnnotations (damtOnTypeArgument, damtOnTypeParameter, out var missingAnnotations))
+					continue;
+
+				var diag = GetDiagnosticId (arg.Kind, param.Kind);
+				var diagArgs = GetDiagnosticArguments (arg, param, missingAnnotations);
+				context.ReportDiagnostic (Diagnostic.Create (DiagnosticDescriptors.GetDiagnosticDescriptor (diag), invocationOperation.Syntax.GetLocation (), diagArgs));
+			}
+		}
+
+		static DiagnosticId GetDiagnosticId (SymbolKind source, SymbolKind target, bool targetsType = false)
```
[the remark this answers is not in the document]
mateoatr: Yes, that's the intention. I think I'd better use another parameter name for this to make it clearer, and maybe add a comment. The logic for (Method, Parameter) was a bit off because of the `targetsType` was passed in ProcessArgument, I already changed that.

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
mateoatr: Added `TryGetAttribute` to get a named attribute on a symbol and simplify with your suggestions. This way we avoid unnecessary null checks and don't do the redundant `HasName` checking.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
mateoatr: Good point. We could use the same logic here and in the ReflectionMethodBodyScanner.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+				return;
+
+			ProcessTypeArguments (context, invocationOperation);
+			ProcessArguments (context, invocationOperation);
+			if (!invocationOperation.TargetMethod.TryGetDynamicallyAccessedMemberTypes (out var damtOnCalledMethod))
+				return;
+
+			if (TryGetSymbolFromOperation (invocationOperation.Instance) is ISymbol instance &&
+				!instance.HasAttribute (RequiresUnreferencedCodeAnalyzer.FullyQualifiedRequiresUnreferencedCodeAttribute)) {
+				instance!.TryGetDynamicallyAccessedMemberTypes (out var damtOnCaller);
```
sbomer, line 87: This looks like `instance` will be the type of the "this" parameter - I think we want to be checking for annotations on the calling method, not the "this" type (which could be a different type entirely).

sbomer: Sorry, that was wrong - I think we want the annotation on the called method (which represents the annotation on its this parameter)

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+
+			return args;
+		}
+
+		static ISymbol? TryGetSymbolFromOperation (IOperation? operation) =>
+			operation switch {
+				IArgumentOperation argument => TryGetSymbolFromOperation (argument.Value),
+				IAssignmentOperation assignment => TryGetSymbolFromOperation (assignment.Value),
+				IConversionOperation conversion => conversion.Type,
+				IInstanceReferenceOperation instanceReference => instanceReference.Type,
```
sbomer, line 221: Maybe leave a comment here that explains we are using the type of the instance to represent `this` values which gets its annotations from the calling method.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+				IInvocationOperation invocation => invocation.TargetMethod,
```
sbomer, line 222: And here, explain that the target method represents the method return which may be annotated.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -0,0 +1,230 @@
...
+				IMemberReferenceOperation memberReference => memberReference.Member,
+				IParameterReferenceOperation parameterReference => parameterReference.Parameter,
+				IReturnOperation returnOp => TryGetSymbolFromOperation (returnOp.ReturnedValue),
+				ITypeOfOperation typeOf => typeOf.TypeOperand,
```
sbomer, line 226: Will the TypeOperand be a NamedType? I imagine this could cause ambiguities with the instanceReference case above. Could we just remove this case for now (I don't see where it is used)?

sbomer (commented):
Just took another look at this. Just a few more suggestions, then I think this looks good. I don't think we need to validate it on runtime yet - let's get this into a feature branch so we can iterate on it there.

sbomer:
@mateoatr I created a branch `feature/damAnalyzer` from main, and changed this PR to target it.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
mateoatr: This will always be a TypeParameter and is needed for cases where there's a `typeof` expression. I might be missing something here but I don't see why/how should we remove it.

sbomer (approved):
Thanks!

PR #2190, opened 2021-08-04
Fix nullref in operator handling

sbomer:
Fixes https://github.com/dotnet/runtime/issues/56819

The issue occurs when preserving an operator method with an argument like `Foo?[]`. We need to preserve the operator if `Foo` was marked. This was done by checking if the argument resolves to `Nullable<T>`, and if so, extracting the generic argument from the unresolved typeref. Cecil resolves `Foo?[]` to `Nullable<T>`, but we weren't prepared to handle this case because `Foo?[]` doesn't have generic arguments.

The fix unwraps the typereference until we get to the `Nullable<Foo>` so that we can extract the generic argument `Foo`.

I wasn't sure how to reproduce the original failing test in https://github.com/dotnet/runtime/issues/56819, but I see that the tests use this pattern. I did confirm that a local testcase reproduced the stacktrace from that issue.

### DIFF
--- a/src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
+++ b/src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
@@ -130,7 +130,18 @@ namespace Mono.Linker.Steps
 			// Unwrap Nullable<T>
 			Debug.Assert (typeDef.HasGenericParameters);
 			var nullableType = type as GenericInstanceType;
-			Debug.Assert (nullableType != null && nullableType.HasGenericArguments && nullableType.GenericArguments.Count == 1);
+			// The original type reference might be a TypeSpecification like array of Nullable<T>
+			// that we need to unwrap until we get to the Nullable<T>
+			while (nullableType == null) {
+				var typeSpec = type as TypeSpecification;
+				Debug.Assert (typeSpec != null);
+				if (typeSpec == null)
+					return null;
+
+				type = typeSpec.ElementType;
+				nullableType = type as GenericInstanceType;
+			}
+			Debug.Assert (nullableType.HasGenericArguments && nullableType.GenericArguments.Count == 1);
 			return _context.TryResolve (nullableType.GenericArguments[0]);
 		}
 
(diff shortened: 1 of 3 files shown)


src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
```diff
@@ -130,7 +130,18 @@ TypeDefinition NonNullableType (TypeReference type)
 			// Unwrap Nullable<T>
 			Debug.Assert (typeDef.HasGenericParameters);
 			var nullableType = type as GenericInstanceType;
-			Debug.Assert (nullableType != null && nullableType.HasGenericArguments && nullableType.GenericArguments.Count == 1);
+			// The original type reference might be a TypeSpecification like array of Nullable<T>
+			// that we need to unwrap until we get to the Nullable<T>
+			while (nullableType == null) {
+				var typeSpec = type as TypeSpecification;
+				Debug.Assert (typeSpec != null);
+				if (typeSpec == null)
```
agocke, line 138: This doesn't make sense. Either `null` is a valid value here and we should guard against it, or it's not and we should assert that it can never happen.

sbomer: The extra null check was there to keep the operator in case I'm wrong and there's a way for it to not be a TypeSpecification - but maybe it's better to let it throw a nullref so we'd catch the missing case.

Regarding your other feedback: there's an `IsTypeSpecification` check but it's not public: https://github.com/jbevain/cecil/blob/master/Mono.Cecil/TypeReference.cs#L322. I did change the GenericInstanceType type check to use `IsGenericInstanceType` instead.

agocke: Thanks -- I'm still trying to understand how Cecil works. So if you call `TryResolve` on the type reference that holds an array you get back the underlying array element type?

Is it the same for pointers? What about function pointers?

sbomer: Yup, it gives back the array element type. It's pretty confusing and we already have our own helper for cases where this isn't the right behavior:

https://github.com/mono/linker/blob/fb9ef5b60a953842a597f0c71ffb862ecaae4af7/src/linker/Linker.Dataflow/MethodBodyScanner.cs#L921-L930

Pointers do the same (Foo* gives Foo), and function pointers have element type null.

MichalStrehovsky (approved):

agocke (approved):
LGTM, thanks

And thanks for the Cecil lesson!

Issue #2160, opened 2021-07-22
Stripping parameter names for members accessed by DAM.All


sbomer:
Marking types for `DynamicallyAccessedMembers.All` or `NestedTypes` is stripping metadata for the accessed members (the repro requires the assembly to be trimmed, for example with `<TrimmerDefaultAction>link</TrimmerDefaultAction>`):
```csharp
using System;
using System.Diagnostics.CodeAnalysis;
using DAM = System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute;
using RUC = System.Diagnostics.CodeAnalysis.RequiresUnreferencedCodeAttribute;

public class Program {
	public static void Main() => RequireAll(typeof(Foo));

	static void RequireAll([DAM(DynamicallyAccessedMemberTypes.All)] Type t) {
		t.GetMethod("Method").Invoke (null, new object[] { "Hi" });
	}
}

public class Foo {
	public static void Method(string message) {
		throw new Exception(message);
	}
}
```
`dotnet run` (notice Foo.Method has the parameter name kept):
```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.Exception: Hi
   at Foo.Method(String message) in /home/sven/tmp/damt/th2/Program.cs:line 17
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Span`1& arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture) in System.Private.CoreLib.dll:token 0x6004db0+0x23
   at System.Reflection.MethodBase.Invoke(Object obj, Object[] parameters) in System.Private.CoreLib.dll:token 0x6004cb4+0x0
   at Program.RequireAll(Type t) in /home/sven/tmp/damt/th2/Program.cs:line 10
   at Program.Main() in /home/sven/tmp/damt/th2/Program.cs:line 7
```
running a trimmed app (notice Foo.Method is missing the parameter name):
```
Unhandled exception. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.Exception: Hi
   at Foo.Method(String ) in /home/sven/tmp/damt/th2/Program.cs:line 23
   --- End of inner exception stack trace ---
[...]

Issue #2159, opened 2021-07-22
Missing warnings for DynamicallyAccessedMemberTypes.All access to annotated methods


sbomer:
Due to the caching for `MarkEntireType`, we aren't producing warnings for all callsites:
```csharp
using System;
using System.Diagnostics.CodeAnalysis;
using DAM = System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute;
using RUC = System.Diagnostics.CodeAnalysis.RequiresUnreferencedCodeAttribute;

public class Program {
	public static void Main() {
		Test1();
		Test2();
	}

	static void Test1() => RequireAll(typeof(Foo)); // warns
	static void Test2() => RequireAll(typeof(Foo)); // no warning

	static void RequireAll([DAM(DynamicallyAccessedMemberTypes.All)] Type t) {}
}

public class Foo {
	[RUC("RUC")]
	void Method() {}
}
```
`NestedTypes` annotations will have similar problems.

PR #2191, opened 2021-08-05
Fix warnings for DAM.All

sbomer:
Fixes https://github.com/mono/linker/issues/2159.

This should also fix https://github.com/mono/linker/issues/2160 but I'm not sure we have a great way to test that. I will look into it more later today.

### DIFF
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
@@ -19,13 +19,13 @@ namespace Mono.Linker
 
 	internal static class DynamicallyAccessedMembersBinder
 	{
-		// Returns the members of the type bound by memberTypes. For DynamicallyAccessedMemberTypes.All, this returns a single null result.
-		// This sentinel value allows callers to handle the case where DynamicallyAccessedMemberTypes.All conceptually binds to the entire type
-		// including all recursive nested members.
+		// Returns the members of the type bound by memberTypes. For DynamicallyAccessedMemberTypes.All, this returns all members of the type and its
+		// nested types, plus the same or any base types or interfaces. The behavior for nested types
 		public static IEnumerable<IMetadataTokenProvider> GetDynamicallyAccessedMembers (this TypeDefinition typeDefinition, LinkContext context, DynamicallyAccessedMemberTypes memberTypes)
 		{
 			if (memberTypes == DynamicallyAccessedMemberTypes.All) {
-				yield return null;
+				foreach (var m in typeDefinition.GetAllOnType (context))
+					yield return m;
 				yield break;
 			}
 
@@ -65,13 +65,19 @@ namespace Mono.Linker
 			}
 
 			if (memberTypes.HasFlag (DynamicallyAccessedMemberTypes.NonPublicNestedTypes)) {
-				foreach (var t in typeDefinition.GetNestedTypesOnType (filter: null, bindingFlags: BindingFlags.NonPublic))
-					yield return t;
+				foreach (var nested in typeDefinition.GetNestedTypesOnType (filter: null, bindingFlags: BindingFlags.NonPublic)) {
+					yield return nested;
+					foreach (var m in nested.GetAllOnType (context))
+						yield return m;
+				}
 			}
 
 			if (memberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicNestedTypes)) {
-				foreach (var t in typeDefinition.GetNestedTypesOnType (filter: null, bindingFlags: BindingFlags.Public))
-					yield return t;
+				foreach (var nested in typeDefinition.GetNestedTypesOnType (filter: null, bindingFlags: BindingFlags.Public)) {
+					yield return nested;
+					foreach (var m in nested.GetAllOnType (context))
+						yield return m;
+				}
 			}
 
 			if (memberTypes.HasFlag (DynamicallyAccessedMemberTypes.NonPublicProperties)) {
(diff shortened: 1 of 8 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -311,77 +309,63 @@ protected bool IsFullyPreserved (TypeDefinition type)
...
 			// Prevent cases where there's nothing on the stack (can happen when marking entire assemblies)
 			// In which case we would generate warnings with no source (hard to debug)
 			using var _ = _scopeStack.CurrentScope.Origin.MemberDefinition == null ? _scopeStack.PushScope (new MessageOrigin (type)) : null;
 
-			MarkEntireTypeInternal (type, includeBaseAndInterfaceTypes, reason);
+			MarkEntireTypeInternal (type, reason);
 		}
 
-		void MarkEntireTypeInternal (TypeDefinition type, bool includeBaseAndInterfaceTypes, in DependencyInfo reason)
+		void MarkEntireTypeInternal (TypeDefinition type, in DependencyInfo reason)
```
mateoatr, line 321: Since you are modifying these methods, is there a reason for keeping both `MarkEntireType` and `MarkEntireTypeInternal`?

sbomer: I think this comment from `MarkEntireType` is still relevant and unnecessary for the recursive call: https://github.com/mono/linker/blob/ffe71535ba927280e6e93db0ca04dd99bde07637/src/linker/Linker.Steps/MarkStep.cs#L316-L318

mateoatr: I guess it doesn't matter much. But if we had it on `MarkEntireTypeInternal` it'd work the same way, since on a second we are guaranteed to have something on the stack (and so the `IDisposable` will be null).

mateoatr (approved):
LGTM!

src/linker/Linker.Steps/MarkStep.cs
sbomer: Good point, I took the suggestion. Thanks!

src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -327,5 +334,56 @@ public static IEnumerable<InterfaceImplementation> GetAllInterfaceImplementation
...
+			var baseType = context.TryResolve (type.BaseType);
+			if (baseType != null) {
+				foreach (var m in GetAllOnType (baseType, context, types))
+					yield return m;
+			}
+
+			if (type.HasInterfaces) {
+				foreach (var iface in type.Interfaces) {
+					yield return iface;
+					var interfaceType = context.Resolve (iface.InterfaceType);
```
MichalStrehovsky, line 362: Should this use `TryResolve`?

src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -327,5 +334,56 @@ public static IEnumerable<InterfaceImplementation> GetAllInterfaceImplementation
...
+		static IEnumerable<IMetadataTokenProvider> GetAllOnType (TypeDefinition type, LinkContext context, HashSet<TypeDefinition> types)
+		{
+			if (!types.Add (type))
+				yield break;
+
+			if (type.HasNestedTypes) {
+				foreach (var nested in type.NestedTypes) {
+					yield return nested;
+					foreach (var m in GetAllOnType (nested, context, types))
+						yield return m;
```
MichalStrehovsky, line 349: Nit: it might be better to just have an extra `List<IMetadataTokenProvider>` parameter that we add do instead of using `yield return`.

Nested/recursive yield returns have a funny debug experience.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -2308,10 +2307,6 @@ void MarkTypeForDynamicallyAccessedMembers (ref ReflectionPatternContext reflect
 				case InterfaceImplementation interfaceImplementation:
 					MarkInterfaceImplementation (ref reflectionContext, interfaceImplementation, dependencyKind);
 					break;
-				case null:
```
MichalStrehovsky, line 2311: This was confusing. I'm glad to see it gone.

MichalStrehovsky (approved):

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -2296,8 +2296,7 @@ void MarkTypeForDynamicallyAccessedMembers (ref ReflectionPatternContext reflect
 					MarkField (ref reflectionContext, field, dependencyKind);
 					break;
 				case TypeDefinition nestedType:
-					DependencyInfo nestedDependencyInfo = new DependencyInfo (dependencyKind, reflectionContext.Source);
-					reflectionContext.RecordRecognizedPattern (nestedType, () => _markStep.MarkEntireType (nestedType, includeBaseAndInterfaceTypes: true, nestedDependencyInfo));
+					MarkType (ref reflectionContext, nestedType, dependencyKind);
```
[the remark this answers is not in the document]
vitek-karas: Can you add a test case along these lines. The base type one would look like:

```C#
class Base<[DAM(All)] T> {}

class TestType : Base<TestType> {}
```

Similarly for interfaces.

Another one is attributes on members: We already have an issue with this on properties and other members - see https://github.com/mono/linker/issues/2196 (I just created this). The All is basically just another case of the same problem for these members.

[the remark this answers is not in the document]
sbomer: Yup, the first example is already covered by this test: https://github.com/mono/linker/blob/e21bb44ca909d33c18acf3ed0b68f94a413351a5/test/Mono.Linker.Tests.Cases/DataFlow/AttributeConstructorDataflow.cs#L96-L100.

I added the second testcase you suggested - I think generic type parameters shouldn't be an issue since the DAM binder only gives back members and doesn't look at the instantiated base generic type at all (instead it relies on normal MarkType logic to mark generic arguments).

Good find about attributes on properties/events.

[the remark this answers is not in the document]
vitek-karas: >  (instead it relies on normal MarkType logic to mark generic arguments).

That was the point - I don't think `MarkType` was written with reentrancy in mind - it seems to handle it, which is good, but I thought it would be good to have tests for it. Before data flow, there were only a few ways to get to the reentrancy, now there are many more.

But I'm fine if we try to cover this later on.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -127,7 +127,17 @@ public void ApplyDynamicallyAccessedMembersToType (ref ReflectionPatternContext
 			// Handle cases where a type has no members but annotations are to be applied to derived type members
 			reflectionPatternContext.RecordHandledPattern ();
 
-			MarkTypeForDynamicallyAccessedMembers (ref reflectionPatternContext, type, annotation, DependencyKind.DynamicallyAccessedMemberOnType);
+			// The annotations this type inherited from its base types should not produce
+			// warnings on the base members, since those are covered by warnings in the base type.
+			// For now, annotations that are inherited from base types but also declared explicitly on this type
+			// still warn about base members since the cache doesn't track enough information to tell the difference.
```
vitek-karas, line 133: I think this is fine - the only case where I can think of this making sense is if the derived type adds more annotations (base has all methods and derived type has both methods and fields) - but that is SO unlikely and rare - I don't care :-).

vitek-karas (approved):
Looks good.

I would like us to de-dupe the code between data flow and MarkEntireType, but it can be done in a followup change.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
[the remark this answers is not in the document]
sbomer: Yeah, it looks like it was at least given some thought but it's good to have more tests as we add more ways to get to MarkType. I did add the generic tests you suggested above so I think that is covered.

PR #2201, opened 2021-08-10
Warn on RUC annotated attribute ctors (analyzer)

### LINKED ISSUE #2094: RUC analyzer doesn't handle attribute .ctors
```C#
class AttributeWithRUC : Attribute
{
    [RequiresUnreferencedCode ("")]
    public AttributeWithRUC() {}
}

// Should produce IL2026
[AttributeWithRUC]
static void MethodWithAttributeWithRUC() {}

// Should not produce any warning since the IL2026 from the attribute should be suppressed by the RUC
[AttributeWithRUC]
static void RUCMethodWithAttributeWithRUC() {}
```

Analyzer doesn't seem to handle attribute instantiations as calls to the attribute .ctor. So if the attribute .ctor has RUC on it, it will not generate a warning.

mateoatr:
Fix #2094.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -43,11 +43,13 @@ namespace ILLink.RoslynAnalyzer
 				context.RegisterSymbolAction (symbolAnalysisContext => {
 					var methodSymbol = (IMethodSymbol) symbolAnalysisContext.Symbol;
 					CheckMatchingAttributesInOverrides (symbolAnalysisContext, methodSymbol);
+					CheckAttributeCtor (symbolAnalysisContext, methodSymbol);
 				}, SymbolKind.Method);
 
 				context.RegisterSymbolAction (symbolAnalysisContext => {
 					var typeSymbol = (INamedTypeSymbol) symbolAnalysisContext.Symbol;
 					CheckMatchingAttributesInInterfaces (symbolAnalysisContext, typeSymbol);
+					CheckAttributeCtor (symbolAnalysisContext, typeSymbol);
 				}, SymbolKind.NamedType);
 
 				if (AnalyzerDiagnosticTargets.HasFlag (DiagnosticTargets.Property)) {
@@ -175,6 +177,21 @@ namespace ILLink.RoslynAnalyzer
 						ReportMismatchInAttributesDiagnostic (symbolAnalysisContext, member, overriddenMember);
 				}
 
+				void CheckAttributeCtor (
+					SymbolAnalysisContext symbolAnalysisContext,
+					ISymbol symbol)
+				{
+					if (symbol.HasAttribute (RequiresAttributeName))
+						return;
+
+					foreach (var attr in symbol.GetAttributes ()) {
+						if (TryGetRequiresAttribute (attr.AttributeConstructor, out var requiresAttribute)) {
+							symbolAnalysisContext.ReportDiagnostic (Diagnostic.Create (RequiresDiagnosticRule,
+								symbol.Locations[0], attr.AttributeConstructor!.Name, GetMessageFromAttribute (requiresAttribute), GetUrlFromAttribute (requiresAttribute)));
+						}
+					}
+				}
+
 				void CheckMatchingAttributesInInterfaces (
 					SymbolAnalysisContext symbolAnalysisContext,
 					INamedTypeSymbol type)
@@ -281,9 +298,12 @@ namespace ILLink.RoslynAnalyzer
 		/// <param name="member">Symbol of the member to search attribute.</param>
 		/// <param name="requiresAttribute">Output variable in case of matching Requires attribute.</param>
 		/// <returns>True if the member contains a Requires attribute; otherwise, returns false.</returns>
-		private bool TryGetRequiresAttribute (ISymbol member, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttribute)
+		private bool TryGetRequiresAttribute (ISymbol? member, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttribute)
 		{
 			requiresAttribute = null;
+			if (member == null)
+				return false;
+
 			foreach (var _attribute in member.GetAttributes ()) {
 				if (_attribute.AttributeClass is { } attrClass &&
 					attrClass.HasName (RequiresAttributeFullyQualifiedName) &&
@@ -292,6 +312,7 @@ namespace ILLink.RoslynAnalyzer
 					return true;
 				}
 			}
+
 			return false;
 		}
 
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -44,27 +44,43 @@ namespace ILLink.RoslynAnalyzer
 					var methodSymbol = (IMethodSymbol) symbolAnalysisContext.Symbol;
 					CheckMatchingAttributesInOverrides (symbolAnalysisContext, methodSymbol);
 					CheckAttributeCtor (symbolAnalysisContext, methodSymbol);
+					foreach (var typeParameter in methodSymbol.TypeParameters)
+						CheckAttributeCtor (symbolAnalysisContext, typeParameter);
+
 				}, SymbolKind.Method);
 
 				context.RegisterSymbolAction (symbolAnalysisContext => {
 					var typeSymbol = (INamedTypeSymbol) symbolAnalysisContext.Symbol;
 					CheckMatchingAttributesInInterfaces (symbolAnalysisContext, typeSymbol);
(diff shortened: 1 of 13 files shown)


MichalStrehovsky:
Might be worth fixing the RUC-annotated-property case as well while you're at it:

```csharp
class AAttribute : Attribute
{
    public Type SomeProperty
    {
        [RUC]
        set { ... }
    }
}

[A(SomeProperty = typeof(Foo))]
class Foo { }
```

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -40,11 +40,13 @@ public override void Initialize (AnalysisContext context)
 				context.RegisterSymbolAction (symbolAnalysisContext => {
 					var methodSymbol = (IMethodSymbol) symbolAnalysisContext.Symbol;
 					CheckMatchingAttributesInOverrides (symbolAnalysisContext, methodSymbol);
+					CheckAttributeCtor (symbolAnalysisContext, methodSymbol);
 				}, SymbolKind.Method);
 
 				context.RegisterSymbolAction (symbolAnalysisContext => {
 					var typeSymbol = (INamedTypeSymbol) symbolAnalysisContext.Symbol;
 					CheckMatchingAttributesInInterfaces (symbolAnalysisContext, typeSymbol);
+					CheckAttributeCtor (symbolAnalysisContext, typeSymbol);
```
vitek-karas, line 49: If I read this correctly it will only work if the attribute is on a method or type(which doesn't seem to be covered by the test BTW). What if I put such attribute on anything else (field, property, method parameter, ...) - ideally we should generate the warning in all cases where the attribute is instantiated (liker will).

mateoatr:
This takes some of the changes from the test restructure PR (https://github.com/mono/linker/pull/2113). In particular, it grabs the `ProducedBy` enum and updates `ExpectedWarningAttribute` to use it instead of `GlobalAnalysisOnly`.
I also added support for `SetupCompileBefore` in the analyzer tests: we check for this attribute and, if found, we grab the sourcefile pointed by the attribute and use its syntax tree for the test compilation in the analyzer infra.
Previously, the analyzer tests only validated the `ExpectedWarning`, `LogContains` and `LogDoesNotContain` annotations when placed on method declarations; this PR also changes this so that it will now validate these attributes when put on any member kind.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -43,26 +43,44 @@ public override void Initialize (AnalysisContext context)
...
+
+				context.RegisterSymbolAction (symbolAnalysisContext => {
+					var propertySymbol = (IPropertySymbol) symbolAnalysisContext.Symbol;
+					if (AnalyzerDiagnosticTargets.HasFlag (DiagnosticTargets.Property)) {
 						CheckMatchingAttributesInOverrides (symbolAnalysisContext, propertySymbol);
-					}, SymbolKind.Property);
-				}
+					}
+
+					CheckAttributeCtor (symbolAnalysisContext, propertySymbol);
```
vitek-karas, line 68: Don't we need to do this for getter/setter as well (similar to what we do for events with add/remove).

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -125,6 +149,21 @@ public override void Initialize (AnalysisContext context)
 				foreach (var extraSyntaxNodeAction in ExtraSyntaxNodeActions)
 					context.RegisterSyntaxNodeAction (extraSyntaxNodeAction.Action, extraSyntaxNodeAction.SyntaxKind);
 
+				void CheckAttributeCtor (
```
vitek-karas, line 152: In a separate change:
This should be called `CheckAttributeInstantiation` and it should also handle the case where the instantiation sets a property on the attribute - as in that case the property setter is called and it can have RUC on it.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
vitek-karas: In fact, RUC is not allowed on the property itself, only on the methods...

vitek-karas (approved):

test/Mono.Linker.Tests.Cases.Expectations/Assertions/ExpectedWarningAttribute.cs
```diff
@@ -17,7 +17,7 @@ public ExpectedWarningAttribute (string warningCode, params string[] messageCont
 		public int SourceColumn { get; set; }
 
 		// Set to true if the warning only applies to global analysis (ILLinker, as opposed to Roslyn Analyzer)
```
agocke, line 19: Obsolete comment

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
```diff
@@ -37,22 +37,26 @@ public static DiagnosticResult Diagnostic (DiagnosticDescriptor descriptor)
...
 			SyntaxTree src,
 			(string, string)[]? globalAnalyzerOptions = null,
-			IEnumerable<MetadataReference>? additionalReferences = null)
+			IEnumerable<MetadataReference>? additionalReferences = null,
+			IEnumerable<SyntaxTree>? additionalSources = null)
 		{
 			var mdRef = MetadataReference.CreateFromFile (typeof (Mono.Linker.Tests.Cases.Expectations.Metadata.BaseMetadataAttribute).Assembly.Location);
 			additionalReferences ??= Array.Empty<MetadataReference> ();
+			List<SyntaxTree> sources = additionalSources != null ? additionalSources.ToList () : new List<SyntaxTree> ();
+			sources.Add (src);
```
agocke, line 56:
```suggestion
			var sources = new List (src);
			sources.AddRange (additionalSources ?? Array.Empty<SyntaxTree> ());
```

test/ILLink.RoslynAnalyzer.Tests/Verifiers/CSharpAnalyzerVerifier`1.cs
```diff
@@ -37,22 +37,26 @@ public static DiagnosticResult Diagnostic (DiagnosticDescriptor descriptor)
...
 			var comp = CSharpCompilation.Create (
 				assemblyName: Guid.NewGuid ().ToString ("N"),
-				syntaxTrees: new SyntaxTree[] { src },
+				syntaxTrees: sources.ToArray (),
```
agocke, line 59:
```suggestion
				syntaxTrees: sources,
```

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -68,11 +66,17 @@ static bool IsWellKnown (AttributeSyntax attr)
...
 		}
 
-		public static void RunTest<TAnalyzer> (MemberDeclarationSyntax m, List<AttributeSyntax> attrs, params (string, string)[] MSBuildProperties)
+		public static void RunTestAsync<TAnalyzer> (MemberDeclarationSyntax m, List<AttributeSyntax> attrs, params (string, string)[] MSBuildProperties)
 			where TAnalyzer : DiagnosticAnalyzer, new()
 		{
+			var testSyntaxTree = m.SyntaxTree.GetRoot ().SyntaxTree;
+			var testDependenciesSource = new List<SyntaxTree> ();
+			foreach (var testDependency in GetTestDependencies (testSyntaxTree))
+				testDependenciesSource.Add (CSharpSyntaxTree.ParseText (File.ReadAllText (testDependency)));
```
agocke, line 75:
```suggestion
			var testDependenciesSource = GetTestDependencies (testSyntaxTree)
			    .Select(testDependency => CSharpSyntaxTree.ParseText (File.ReadAllText (testDependency)));
```

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -68,11 +66,17 @@ static bool IsWellKnown (AttributeSyntax attr)
 			}
 		}
 
-		public static void RunTest<TAnalyzer> (MemberDeclarationSyntax m, List<AttributeSyntax> attrs, params (string, string)[] MSBuildProperties)
+		public static void RunTestAsync<TAnalyzer> (MemberDeclarationSyntax m, List<AttributeSyntax> attrs, params (string, string)[] MSBuildProperties)
```
agocke, line 69:
```suggestion
		public static void RunTest<TAnalyzer> (MemberDeclarationSyntax m, List<AttributeSyntax> attrs, params (string, string)[] MSBuildProperties)
```

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
mateoatr: This is needed for `RequiresAssemblyFilesAttribute`, which can be put on properties.

PR #2202, opened 2021-08-10
Target net6.0 for linker

vitek-karas:
Fixes https://github.com/mono/linker/issues/2118

### DIFF
--- a/src/ILLink.Tasks/ILLink.Tasks.csproj
+++ b/src/ILLink.Tasks/ILLink.Tasks.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <!-- Keep these in sync with _ILLinkTasksTFM in Sdk.props. -->
     <!-- Keep the netcoreapp TFM in sync with the Mono.Linker.csproj condition below. -->
-    <TargetFrameworks>net5.0;net472</TargetFrameworks>
+    <TargetFrameworks>net6.0;net472</TargetFrameworks>
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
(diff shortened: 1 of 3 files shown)


src/ILLink.Tasks/ILLink.Tasks.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <!-- Keep these in sync with _ILLinkTasksTFM in Sdk.props. -->
```
sbomer, line 3: Might be worth updating this comment.

sbomer (approved):

agocke (approved):
LGTM. Does this change anything since the SDK usually forces everyone to net6.0 in publishing anyway?

vitek-karas:
> Does this change anything since the SDK usually forces everyone to net6.0 in publishing anyway?

Not really - the only scenario affected is if you run linker directly. Nobody does that - except us when debugging the linker.

lewing:
This seems to be causing issues dotnet/sdk#19655 and dotnet/runtime#57203

vitek-karas:
Working on it...

mateoatr:
I'm taking a look. This appears to be caused by the analyzer not taking into account the `SuppressTrimAnalysisWarnings`, `ILLinkWarningLevel` properties.

mateoatr:
> I'm taking a look. This appears to be caused by the analyzer not taking into account the `SuppressTrimAnalysisWarnings`, `ILLinkWarningLevel` properties.

Ok, the error was about wrong path. Thanks for clarification @vitek-karas.

Issue #2181, opened 2021-07-30
linker producing bad code in finally block that causes AOT compiler to seg fault


stephentoub:
See https://github.com/dotnet/runtime/pull/56316#issuecomment-888757492.

vitek-karas:
What should be simple repro steps - requires macOS (I wasn't able to validate that it does repro yet):

Latest dotnet/runtime repo clone
Remove these lines: https://github.com/dotnet/runtime/blob/7249ec41e059c3900eaf361007e5725033320453/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/Http3RequestStream.cs#L130-L135

```console
./build.sh -Subset mono+libs -os iOSSimulator -arch x64
./dotnet.sh build /t:Test src/tests/FunctionalTests/iOS/Simulator/AOT /p:TargetOS=iOSSimulator /p:TargetArchitecture=x64
```

The last command should fail to AOT compile.

vitek-karas:
Just an update. The underlying problem is:
* Sometimes linker edits method bodies by replacing instructions with new instructions
* One such place is the "isinst" instruction - if the target type is not instantiated we replace the "isinst" with "pop, ldnull" (basically return false).
* We use Cecils ILProcessor to edit the method body
* This component simple replaces the instruction objects in the method body collection, unfortunately it doesn't patch other data structures.
* On such data structure is the exception handler records - after replace these continue to point to the old instruction which not in the method body anymore
* When such state is saved to IL, the catch/finally block "pointers/offsets" in the IL will end up with effectively random values (typically these just shift a little bit)

Mono AOT compiler is obviously sensitive to such corruption.

The case of the `Http3RequestStream` is about the "isinst" case above. We have a preliminary fix for it here: https://github.com/vitek-karas/linker/tree/FixBrokenFinally

Unfortunately there are other places where linker edits the method body which all have similar issues. One such place is constant propagation and branch removal - which is what feature switches rely on. That's probably where the CoreLib/ArrayPool issue is - but we were not able to get a repro of that to validate the theory.

We don't have a candidate fix for the feature switch corruption (nor we have a direct repro - even a constructed one, just yet).

vitek-karas:
The issue with `isinst` rewrite which affected `Http3RequestStream` is now fixed. The fix is targeted to `isinst` and like won't fix whatever problem is in the ArrayPool around finaly sections (although it's like the same root cause).

vitek-karas:
[...]

PR #2205, opened 2021-08-11
Fix broken try/catch/filter offsets after isinst optimization

vitek-karas:
The isinst optimization replaces that instruction with a `pop, ldnull` when the type is not instantiated. This changes the instruction and also changes the length of the instruction in that position. Cecil unfortunately doesn't update try/catch/filter references and they keep pointing to the old `isinst` instruction which not part of the method body anymore. When saving the assembly the offsets stores in the try/catch/filter records end up effective random and corrupted.

This is a short-term fix to unblock failures in runtime due to this problem. 

Medium term fix would be to carefully handle all IL replacements in the linker with regard to try/catch/filter records.

Ideally the long term fix would be to do this in Cecil in such a way that IL replacements would be correctly handled on their own.

This fixes the `Http3RequestStream` failures mentioned in https://github.com/mono/linker/issues/2181, but I was not able to confirm if this fixes the CoreLib ArrayPool issues as well (I think it will not).

### DIFF
--- a/test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
+++ b/test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
@@ -404,7 +404,7 @@ namespace Mono.Linker.Tests.TestCasesRunner
 
 			HashSet<(Instruction, Instruction)> existingTryBlocks = new HashSet<(Instruction, Instruction)> ();
 			foreach (var exHandler in body.ExceptionHandlers) {
-				if (!existingTryBlocks.Contains ((exHandler.TryStart, exHandler.TryEnd))) {
+				if (existingTryBlocks.Add ((exHandler.TryStart, exHandler.TryEnd))) {
 					InsertBeforeInstruction (exHandler.TryStart, ".try");
 					InsertBeforeInstruction (exHandler.TryEnd, ".endtry");
 				}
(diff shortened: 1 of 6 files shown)


test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
```diff
@@ -390,11 +390,44 @@ protected static void VerifyInstructions (MethodDefinition src, MethodDefinition
...
+		{
+			List<(Instruction, string)> result = new List<(Instruction, string)> (body.Instructions.Count);
+			for (int index = 0; index < body.Instructions.Count; index++) {
+				var instruction = body.Instructions[index];
+				result.Add ((instruction, FormatInstruction (instruction).ToLowerInvariant ()));
+			}
+
+			HashSet<(Instruction, Instruction)> existingTryBlocks = new HashSet<(Instruction, Instruction)> ();
+			foreach (var exHandler in body.ExceptionHandlers) {
+				if (!existingTryBlocks.Contains ((exHandler.TryStart, exHandler.TryEnd))) {
```
sbomer, line 407: nit: you could use `Add` here instead of `Contains` and `Add` below.

sbomer (approved):
LGTM!

vitek-karas:
The isinst optimization replaces that instruction with a pop, ldnull when the type is not instantiated. This changes the instruction and also changes the length of the instruction in that position. Cecil unfortunately doesn't update try/catch/filter references and they keep pointing to the old isinst instruction which not part of the method body anymore. When saving the assembly the offsets stores in the try/catch/filter records end up effective random and corrupted.

This is a short-term fix to unblock failures in runtime due to this problem.

Medium term fix would be to carefully handle all IL replacements in the linker with regard to try/catch/filter records.

Ideally the long term fix would be to do this in Cecil in such a way that IL replacements would be correctly handled on their own.

This fixes the Http3RequestStream failures mentioned in #2181, but I was not able to confirm if this fixes the CoreLib ArrayPool issues as well (I think it will not).

PR #2206, opened 2021-08-11
Use DAM binder for MarkEntireType

sbomer:
This changes MarkEntireType to reuse the recursive logic in the DynamicallyAccessedMembers binder, as suggested here: https://github.com/mono/linker/pull/2191#discussion_r685445252.

Ideally we could just call MarkType on the type and its nested types, but this would produce extra warnings in copy assemblies since MarkType can cause us to mark extra methods.

The cache is no longer necessary because we usually only call MarkEntireType once for a type. The exception is `PresereveDependencyAttribute` which is deprecated.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -59,7 +59,6 @@ namespace Mono.Linker.Steps
 		readonly List<(TypeDefinition Type, MethodBody Body, Instruction Instr)> _pending_isinst_instr;
 		UnreachableBlocksOptimizer _unreachableBlocksOptimizer;
 		MarkStepContext _markContext;
-		readonly HashSet<TypeDefinition> _entireTypesMarked;
 		DynamicallyAccessedMembersTypeHierarchy _dynamicallyAccessedMembersTypeHierarchy;
 		MarkScopeStack _scopeStack;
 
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -320,47 +316,42 @@ internal void MarkEntireType (TypeDefinition type, in DependencyInfo reason)
...
+					break;
+				case PropertyDefinition property:
+					MarkProperty (property, reason);
+					break;
+				case EventDefinition @event:
+					MarkEvent (@event, reason);
+					break;
+				case InterfaceImplementation iface:
+					MarkInterfaceImplementation (iface, new MessageOrigin (type));
+					break;
```
vitek-karas, line 354: If I read the code in `GetAllOnType` correctly, it will not return any interfaces when `declaredOnly:true`.

Means we're probably missing a test as well.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -320,47 +316,42 @@ internal void MarkEntireType (TypeDefinition type, in DependencyInfo reason)
...
+			void MarkTypeOnly (TypeDefinition type, in DependencyInfo reason)
+			{
+				Annotations.Mark (type, reason);
+				MarkCustomAttributes (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
+				MarkTypeSpecialCustomAttributes (type);
+				MarkGenericParameterProvider (type);
 			}
 
-			MarkGenericParameterProvider (type);
+			MarkTypeOnly (type, reason);
```
vitek-karas, line 330: This comment is about the previous change in this area:
We don't do anything with base type here - it's the behavior we have for copy assemblies, so that's fine, but I think it's worth adding a comment - reading the code it looks really weird.

Side note: the base type will end up being marked in pretty much all cases anyway, since we call MarkMethod which will in turn call MarkType on the declaring type, which will mark the base type as appropriate.

vitek-karas (changes requested):

PR #2207, opened 2021-08-12
Add static constructor behavior for RUC on type

tlakollo:
Add new error code for cases in which the static constructor is annotated with RUC
Move static constructor verification in MarkField before pushing a new stack to print warnings in the method caller instead of fields
Don't treat static constructors annotated with RUC as dangerous
Add test for several static constructor calls


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3007,6 +3006,11 @@ protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo
 				MarkRequirementsForInstantiatedTypes (method.DeclaringType);
 				Tracer.AddDirectDependency (method.DeclaringType, new DependencyInfo (DependencyKind.InstantiatedByCtor, method), marked: false);
 			}
+			if (method.IsStaticConstructor () && Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute>(method)) {
```
vitek-karas, line 3009: `else if` - the method can't be both instance and static .ctor :-)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3007,6 +3006,11 @@ protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo
...
+				string formatString = SharedStrings.RequiresUnreferencedCodeOnStaticConstructorMessage;
+				string message = string.Format (formatString, method.GetDisplayName ());
+				_context.LogWarning (message, 2116, _scopeStack.CurrentScope.Origin, MessageSubCategory.TrimAnalysis);
```
vitek-karas, line 3012: Please add this to the `error-codes.md`

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -882,6 +897,87 @@ public class NestedClass
...
+			[RequiresUnreferencedCode("Message for --StaticCtor--")]
+			class StaticCtor
+			{
+				static StaticCtor ()
+				{
+				}
+			}
+
+			[ExpectedWarning ("IL2026", "RequiresOnClass.StaticCtor.StaticCtor()", "Message for --StaticCtor--", GlobalAnalysisOnly = true)]
+			[ExpectedWarning ("IL2026", "RequiresOnClass.StaticCtor..cctor()", "Message for --StaticCtor--", GlobalAnalysisOnly = true)]
```
vitek-karas, line 909: I assume the first one should say `GlobalAnalysisOnly=false` - but it's probably not enough, since it will only show up in the analyzer, right?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1633,6 +1633,16 @@ void MarkField (FieldDefinition field, in DependencyInfo reason)
...
 			MarkType (field.DeclaringType, new DependencyInfo (DependencyKind.DeclaringType, field));
 
+			var parent = field.DeclaringType;
+			if (!Annotations.HasPreservedStaticCtor (parent)) {
+				var cctorReason = reason.Kind switch {
+					// Report an edge directly from the method accessing the field to the static ctor it triggers
+					DependencyKind.FieldAccess => new DependencyInfo (DependencyKind.TriggersCctorThroughFieldAccess, reason.Source),
+					_ => new DependencyInfo (DependencyKind.CctorForField, field)
+				};
+				MarkStaticConstructor (parent, cctorReason);
```
vitek-karas, line 1643: I'm a bit lost how this works.

I thought the idea was that we would warn about access to a static field - so I would expect a check for the field being static and its type having RUC and then warn.
But based on the tests, this warns on the static `.cctor` - which I don't get since below we explicitly exclude `.cctor` from methods which can "require unreferenced code" and thus it should not produce a warning on its own.

Also from the end-user behavior I would expect to get the warning on the field access talking about the field - not about the static `.cctor` - for two reasons - one is that the invocation of the static `.cctor` is invisible and second, in lot of cases the static `.cctor` is a compiler generated method.

tlakollo: The problematic part of static constructors I think is only when the static constructor is the origin of the warning because it might be a compiler-generated method. But in this case, that is not the origin, the origin is the method that calls the field access, this can be suppressed or continue to annotate with RUC. The reason is not warning with something like `Using method 'some field' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code` is because we don't support RUC on fields so not only is weird to read it will need new methods to handle it

vitek-karas: I think it's perfectly OK if we use slightly different warning message for that case (but still use the same warning code).

tlakollo: Added a special case for the case when we are going to warn on a field in CheckAndReportRequiresUnreferencedCode, let me know if this is the behavior you were expecting

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
tlakollo: Creating the new instance of the class is calling the .cctor and then .ctor, why would linker show only one warning?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3006,6 +3012,10 @@ protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo
 			if (method.IsInstanceConstructor ()) {
 				MarkRequirementsForInstantiatedTypes (method.DeclaringType);
 				Tracer.AddDirectDependency (method.DeclaringType, new DependencyInfo (DependencyKind.InstantiatedByCtor, method), marked: false);
+			} else if (method.IsStaticConstructor () && Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (method)) {
+				string formatString = SharedStrings.RequiresUnreferencedCodeOnStaticConstructorMessage;
+				string message = string.Format (formatString, method.GetDisplayName ());
+				_context.LogWarning (message, 2116, _scopeStack.CurrentScope.Origin, MessageSubCategory.TrimAnalysis);
```
mateoatr, line 3018: Please add the new ID to `DiagnosticId` and consider using it to retrieve the diagnostic string:

```suggestion
				_context.LogWarning (new DiagnosticString (DiagnosticId.RequiresUnreferencedCodeOnStaticConstructor), (int) DiagnosticId.RequiresUnreferencedCodeOnStaticConstructor, _scopeStack.CurrentScope.Origin, MessageSubCategory.TrimAnalysis);
```

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2959,7 +2963,8 @@ protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo
 			if (!_methodReasons.Contains (reason.Kind))
 				throw new InternalErrorException ($"Unsupported method dependency {reason.Kind}");
 #endif
-
+			if (method.FullName.Contains ("TestStaticCtorMarkingIsTriggeredByFieldAccessWrite"))
+				Debugger.Break ();
```
mateoatr, line 2967: I suppose this was used for debugging.
```suggestion
```

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1657,6 +1658,9 @@ void MarkField (FieldDefinition field, in DependencyInfo reason)
 				}
 			}
 
+			if (reason.Kind is DependencyKind.FieldAccess && field.IsStatic && Annotations.TryGetLinkerAttribute (field.DeclaringType, out RequiresUnreferencedCodeAttribute requiresUnreferencedCode))
+				ReportRequiresUnreferencedCode (field.GetDisplayName (), requiresUnreferencedCode, methodCaller);
```
vitek-karas, line 1662: This needs to happen before the call to `CheckProcessed` since we want to report the warning on every "call site", not just on the first one.

I played with this some more: https://github.com/tlakollo/linker/pull/1 - please take a look - fixed this issue and added more tests.

Note that I think this will need to warn on almost all Kind - reflection access, DAM access, ...

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -1016,15 +1143,197 @@ static void TestSuppressionsOnClass ()
...
+			[AttributeWithRUC (PropertyOnAttribute = 42)]
+			[ExpectedWarning ("IL2026", "AttributeWithRUC.AttributeWithRUC()", GlobalAnalysisOnly = true)]
+			static void KeepFieldOnAttribute () { }
+
+			[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.NonPublicFields)]
+			[RequiresUnreferencedCode ("This class is dangerous")]
+			class ClassImplementingAnnotatedInterface
+			{
+				[ExpectedWarning ("IL2112", "ClassImplementingAnnotatedInterface::publicField", GlobalAnalysisOnly = true)]
+				public static int publicField;
```
tlakollo, line 1326: I wasn't fully sure if the field is the best place to warn about IL2112

sbomer: This is at least consistent with what we do for `DAM`-annotated fields, so I think it's the right place for the warning. Though I agree this could be confusing.

vitek-karas: I think this is fine - it's the same for methods. The IL2112 and similar warnings are inherently confusing to some degree, hopefully they should be super rare.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1571,17 +1571,26 @@ static bool IsDeclaredWithinType (IMemberDefinition member, TypeDefinition type)
...
+						MessageFormat.FormatRequiresAttributeMessageArg (memberAttribute.Url));
 					var code = reportOnMember ? 2112 : 2113;
 					_context.LogWarning (message, code, origin, MessageSubCategory.TrimAnalysis);
 				}
 
+				if (member is MethodDefinition method && Annotations.DoesMethodRequireUnreferencedCode (method, out RequiresUnreferencedCodeAttribute methodAttribute)) {
+					ReportRequiresUnreferencedCodeWhileApplyingDynamicallyAccessedMembers (type.GetDisplayName (), method.GetDisplayName (), methodAttribute);
+				} else if (member is FieldDefinition field && field.IsStatic &&
+					Annotations.TryGetLinkerAttribute (field.DeclaringType, out RequiresUnreferencedCodeAttribute fieldAttribute) &&
+					!ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ()) {
```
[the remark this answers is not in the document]
vitek-karas: Good point - fixed.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1571,17 +1571,26 @@ static bool IsDeclaredWithinType (IMemberDefinition member, TypeDefinition type)
...
+						member,
+						MessageFormat.FormatRequiresAttributeMessageArg (memberAttribute.Message),
+						MessageFormat.FormatRequiresAttributeMessageArg (memberAttribute.Url));
 					var code = reportOnMember ? 2112 : 2113;
 					_context.LogWarning (message, code, origin, MessageSubCategory.TrimAnalysis);
 				}
 
+				if (member is MethodDefinition method && Annotations.DoesMethodRequireUnreferencedCode (method, out RequiresUnreferencedCodeAttribute methodAttribute)) {
+					ReportRequiresUnreferencedCodeWhileApplyingDynamicallyAccessedMembers (type.GetDisplayName (), method.GetDisplayName (), methodAttribute);
+				} else if (member is FieldDefinition field && field.IsStatic &&
```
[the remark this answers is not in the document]
vitek-karas: Good idea - done.

vitek-karas:
Still working on events as it should be simple addition.

vitek-karas:
I'm giving up on events - it WILL warn, just on the wrong place and multiple times... see https://github.com/mono/linker/issues/2218. Not just RUC on type issue, this is an overall issue for events.

sbomer (approved):
LGTM!

PR #2209, opened 2021-08-12
Special case `MakeGenericMethod/Type` in RUC analyzer

mateoatr:
Now that `MakeGenericMethod` and `MakeGenericType` methods were annotated in the runtime, we need to special case these in the analyzer to emit the right warning (IL2060 and IL2055, respectively).

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
@@ -72,6 +72,9 @@ namespace ILLink.RoslynAnalyzer
 		{
 			if (member is IMethodSymbol method && ImmutableArrayOperations.Contains (specialIncompatibleMembers, member, SymbolEqualityComparer.Default) &&
 				(method.Name == "MakeGenericMethod" || method.Name == "MakeGenericType")) {
+				// These two RUC-annotated APIs are intrinsically handled by the trimmer, which will not produce any
+				// RUC warning related to them. For unrecognized reflection patterns realted to generic type/method
+				// creation IL2055/IL2060 should be used instead.
 				return true;
 			}
 
--- a/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
@@ -131,8 +131,8 @@ namespace ILLink.RoslynAnalyzer
 			return false;
 		}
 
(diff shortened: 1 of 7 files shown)


src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -50,6 +52,37 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
...
+			}
+
+			return incompatibleMembers.ToImmutable ();
+		}
+
+		protected override bool ReportSpecialIncompatibleMembersDiagnostic (OperationAnalysisContext operationContext, ImmutableArray<ISymbol> specialIncompatibleMembers, ISymbol member)
+		{
+			if (member is IMethodSymbol method && ImmutableArrayOperations.Contains (specialIncompatibleMembers, member, SymbolEqualityComparer.Default)) {
+				if (method.Name == "MakeGenericType") {
+					operationContext.ReportDiagnostic (Diagnostic.Create (s_makeGenericTypeRule, operationContext.Operation.Syntax.GetLocation (), method.GetDisplayName ()));
```
vitek-karas, line 75: I wonder if we should simply not warn here.
The guiding principal for the analyzer should be:
"If the analyzer produces a warning -> linker will also produce that warning"

In this case this will now hold - linker has more complex handling of MakeGenericType/Method which will avoid warning in many cases. The analyzer would warn every time. I would much rather we special case this and not warn for now, until we can build enough smarts in the analyzer to match the linker behavior.

/cc @sbomer - we discussed the general behavior, so for context.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -50,6 +52,32 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
...
+			}
+
+			return incompatibleMembers.ToImmutable ();
+		}
+
+		protected override bool ReportSpecialIncompatibleMembersDiagnostic (OperationAnalysisContext operationContext, ImmutableArray<ISymbol> specialIncompatibleMembers, ISymbol member)
+		{
+			if (member is IMethodSymbol method && ImmutableArrayOperations.Contains (specialIncompatibleMembers, member, SymbolEqualityComparer.Default) &&
+				(method.Name == "MakeGenericMethod" || method.Name == "MakeGenericType")) {
+				return true;
```
vitek-karas, line 75: Can you please add a comment why we're doing this?

vitek-karas (approved):

PR #2210, opened 2021-08-13
Fix warnings for DAMT on interfaces

sbomer:
https://github.com/mono/linker/pull/2191 addressed issues with redundant
warnings from base types, but interfaces need to be handled specially.

There were a few issues with the interface handling in that change:
- The binding logic didn't include interfaces for DAM.All when
  declaredOnly was specified, as @vitek-karas pointed out in
   https://github.com/mono/linker/pull/2206#discussion_r687573854.
   This could have led to us not marking interfaces for types derived
   from a DAMT.All-annotated type. Either way, the binding logic
   should not select interface _members_ when declaredOnly is specified.
- The DAMT-on-type logic had special handling for base types, but
  didn't separately mark interface members, which is needed when
  applying annotations DAMT.All or DAMT.Interfaces. This could have led
  to missing interface impls or interface members for types which have
  (explicit or inherited) DAMT.All or DAMT.Interfaces annotations.

Separately, the bit operations on DAMT also didn't correctly handle cases
where there are bits shared by different enum members.

### DIFF
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
@@ -227,9 +227,6 @@ namespace Mono.Linker.Dataflow
 		{
 			Debug.Assert (annotation != DynamicallyAccessedMemberTypes.None);
 
-			if (type.Name == "AnnotatedAll")
-				System.Console.WriteLine ("Hey");
-
 			reflectionPatternContext.AnalyzingPattern ();
 			// Handle cases where a type has no members but annotations are to be applied to derived type members
 			reflectionPatternContext.RecordHandledPattern ();
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
@@ -120,7 +120,7 @@ namespace Mono.Linker.Dataflow
 				var reflectionMethodBodyScanner = new ReflectionMethodBodyScanner (_context, _markStep, _scopeStack);
 				// Set up a context to report warnings on access to annotated members, with the annotated type as the origin.
 				var reflectionPatternContext = new ReflectionPatternContext (_context, reportingEnabled: true, _scopeStack.CurrentScope.Origin, type);
-				reflectionMethodBodyScanner.ApplyDynamicallyAccessedMembersToType (ref reflectionPatternContext, type, annotation);
+				ApplyDynamicallyAccessedMembersToType (ref reflectionMethodBodyScanner, ref reflectionPatternContext, type, annotation);
 				reflectionPatternContext.Dispose ();
 			}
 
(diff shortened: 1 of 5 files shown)


src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -216,13 +216,63 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
...
 
 			return applied;
 		}
 
+		public void ApplyDynamicallyAccessedMembersToType (ref ReflectionMethodBodyScanner reflectionMethodBodyScanner, ref ReflectionPatternContext reflectionPatternContext, TypeDefinition type, DynamicallyAccessedMemberTypes annotation)
+		{
+			Debug.Assert (annotation != DynamicallyAccessedMemberTypes.None);
+
+			if (type.Name == "AnnotatedAll")
+				System.Console.WriteLine ("Hey");
```
vitek-karas, line 231: Left over from debugging :wink:

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -216,13 +216,63 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
...
+
+			reflectionPatternContext.AnalyzingPattern ();
+			// Handle cases where a type has no members but annotations are to be applied to derived type members
```
vitek-karas, line 234:
```suggestion
			// There are no cases where we don't handle the pattern - the method will always "deal with it"
```

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -216,13 +216,63 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
...
+			// But the annotations on base/interfaces are already applied so we don't need to apply those
+			// again (and should avoid doing so as it would produce extra warnings).
+			var baseType = _context.TryResolve (type.BaseType);
+			if (baseType != null) {
+				var baseAnnotation = GetCachedInfoForTypeInHierarchy (baseType);
+				var annotationToApplyToBase = ReflectionMethodBodyScanner.GetMissingMemberTypes (annotation, baseAnnotation.annotation);
+
+				// Apply any annotations that didn't exist on the base type to the base type.
+				// This may produce redundant warnings when the annotation is DAMT.All or DAMT.PublicConstructors and the base already has a
+				// subset of those annotations.
```
vitek-karas, line 247: Unfortunate, but I would not worry about it for now...

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -216,13 +216,63 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
...
+
+					// Apply All or Interfaces to the interface type.
+					// DAMT.All may produce redundant warnings from implementing types, when the interface type already had some annotations.
+					reflectionMethodBodyScanner.MarkTypeForDynamicallyAccessedMembers (ref reflectionPatternContext, interfaceType, annotationToApplyToInterfaces, DependencyKind.DynamicallyAccessedMemberOnType, declaredOnly: false);
+				}
+			}
+
+			// The annotations this type inherited from its base types or interfaces should not produce
+			// warnings on the respective base/interface members, since those are already covered by applying
+			// the annotations to those types.
```
vitek-karas, line 272:
```suggestion
			// The annotations this type inherited from its base types or interfaces should not produce
			// warnings on the respective base/interface members, since those are already covered by applying
			// the annotations to those types. So we only need to handle the members directly declared on this type.
```

vitek-karas (approved):

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
sbomer: Oops!

Issue #2211, opened 2021-08-13
Error when trying to trim an ASP.NET Core project


davidfowl:
Project:

```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <Nullable>enable</Nullable>
    <ImplicitUsings>enable</ImplicitUsings>
    <PublishTrimmed>true</PublishTrimmed>
  </PropertyGroup>

</Project>
```

Program.cs

```C#
var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();

app.MapGet("/", () => "Hello World!");

app.Run();
```

Command:

dotnet publish -r win-x64

Exception:

```
ILLink : error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues [C:\Users\david\source\repos\WebApplication520\WebApplication3\WebApplication3.csproj]
  Fatal error in IL Linker
  Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
     at Mono.Linker.MessageContainer.TryLogSingleWarning(LinkContext context, Int32 code, MessageOrigin origin, String subcategory)
     at Mono.Linker.MessageContainer.CreateWarningMessageContainer(LinkContext context, String text, Int32 code, MessageOrigin origin, WarnVersion version, String subcategory)
     at Mono.Linker.MessageContainer.CreateWarningMessage(LinkContext context, String text, Int32 code, MessageOrigin origin, WarnVersion version, String subcategory)
     at Mono.Linker.LinkContext.LogWarning(String text, Int32 code, MessageOrigin origin, String subcategory)
     at Mono.Linker.Steps.MarkStep.ReportRequiresUnreferencedCode(String displayName, RequiresUnreferencedCodeAttribute requiresUnreferencedCode, MessageOrigin currentOrigin)
     at Mono.Linker.Steps.MarkStep.CheckAndReportRequiresUnreferencedCode(MethodDefinition method)
     at Mono.Linker.Steps.MarkStep.ProcessAnalysisAnnotationsForMethod(MethodDefinition method, DependencyKind dependencyKind)
     at Mono.Linker.Steps.MarkStep.MarkMethod(MethodReference reference, DependencyInfo reason)
     at Mono.Linker.Steps.MarkStep.MarkIndirectlyCalledMethod(MethodDefinition method, DependencyInfo& reason)
[...]

PR #2212, opened 2021-08-13
Support assembly as message origin

sbomer:
https://github.com/mono/linker/issues/2211 is caused by an assembly-level `MetadataUpdateHandlerAttribute` (which requires DAMT.All) on Microsoft.AspNetCore.Http.Abstractions. Currently this seems to be an analysis hole - we don't show warnings for assembly-level attributes because we don't consider the assembly (or assembly attribute) as a possible message origin. As far as I know we don't have a targeted suppression mechanism for such warnings, but we need to figure out how to close this hole.

I started working on an approach that will at least show such warnings.

Fixes https://github.com/mono/linker/issues/2211


mateoatr:
> As far as I know we don't have a targeted suppression mechanism for such warnings

You mean having an assembly as the target of an UnconditionalSuppressMessage? If this is the case, I think we have avoided this since we don't want users to be able to cross-assembly suppress trimmer warnings.

sbomer:
I agree we don't want cross-assembly suppressions - but what might be nice is some way to suppress warnings that come from the current assembly's attributes (without also suppressing the same warning for all other attributes in the same assembly). @vitek-karas agreed we don't need that for now though.

sbomer:
@vitek-karas @mateoatr PTAL

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1362,7 +1368,7 @@ protected void MarkAssembly (AssemblyDefinition assembly, DependencyInfo reason)
 				return;
 
 			// We don't have a good origin for "assembly" level, so just use null for now
```
mateoatr, line 1370:
```suggestion
```

mateoatr (approved):
LGTM!

vitek-karas (approved):
Looks good

PR #2213, opened 2021-08-13
Implement a custom ILProcessor and make it auto-fix IL corruptions

vitek-karas:
Cecil doesn't handle IL editing well. It doesn't patch scopes and branches. Added a linker-private IL processor which does this.


src/linker/Linker/LinkerILProcessor.cs
```diff
@@ -0,0 +1,125 @@
...
+		public void Replace (int index, Instruction instruction) => Replace (_ilProcessor.Body.Instructions[index], instruction);
+
+		public void Remove (Instruction instruction)
+		{
+			int index = _ilProcessor.Body.Instructions.IndexOf (instruction);
+			if (index == -1)
+				throw new ArgumentOutOfRangeException (nameof (instruction));
+
+			Instruction nextInstruction = Instructions.Count == index + 1 ? null : Instructions[index + 1];
+			Instruction previousInstruction = index == 0 ? (Instructions.Count == 1 ? null : Instructions[index + 1]) : Instructions[index - 1];
```
sbomer, line 67: Is the index+1 case for previousInstruction intentional?

src/linker/Linker/LinkerILProcessor.cs
```diff
@@ -0,0 +1,125 @@
...
+
+			RedirectScopeStart (instruction, nextInstruction);
+			RedirectScopeEnd (instruction, previousInstruction);
```
sbomer, line 70: Should removing an instruction also fix up branch targets?

src/linker/Linker/LinkerILProcessor.cs
```diff
@@ -0,0 +1,125 @@
...
+		public void Emit (OpCode opcode) => Append (_ilProcessor.Create (opcode));
+		public void Emit (OpCode opcode, TypeReference type) => Append (_ilProcessor.Create (opcode, type));
+		public void Emit (OpCode opcode, MethodReference method) => Append (_ilProcessor.Create (opcode, method));
+		public void Emit (OpCode opcode, VariableDefinition variable) => Append (_ilProcessor.Create (opcode, variable));
+
+		public void Emit (OpCode opcode, string value) => Append (_ilProcessor.Create (opcode, value));
+		public void InsertBefore (Instruction target, Instruction instruction)
+		{
+			// So when inserting before, all pointers have to be updated to the new "before" instruction.
+			RedirectScopeStart (target, instruction);
```
sbomer, line 31: Should this also update branch targets?

sbomer (commented):
Have we considered upstreaming these improvements? It might be worth asking Jb about the reasons for not automatically updating branch/handler targets.

src/linker/Linker/LinkerILProcessor.cs
vitek-karas: Yes - if we're removing the first instruction, we have to re-point to "something", in this case the previous and next will be the same, but there's no other way really.

src/linker/Linker/LinkerILProcessor.cs
vitek-karas: Yes - it should - good catch!

src/linker/Linker/LinkerILProcessor.cs
vitek-karas: You're right.

vitek-karas:
We will try the upstream, but we can't wait for upstream (even if we authored the fix, typically it takes a while to get it approved/merged).

src/linker/Linker/LinkerILProcessor.cs
```diff
@@ -0,0 +1,127 @@
...
+			int index = _ilProcessor.Body.Instructions.IndexOf (instruction);
+			if (index == -1)
+				throw new ArgumentOutOfRangeException (nameof (instruction));
+
+			Instruction nextInstruction = Instructions.Count == index + 1 ? null : Instructions[index + 1];
+			Instruction previousInstruction = index == 0 ? (Instructions.Count == 1 ? null : Instructions[index + 1]) : Instructions[index - 1];
+
+			RedirectScopeStart (instruction, nextInstruction);
+			RedirectScopeEnd (instruction, previousInstruction);
+			ReplaceInstructionReference (instruction, nextInstruction ?? previousInstruction);
```
sbomer, line 72: If removing the last instruction from a body with >1 instructions, this rewrites the branch target to the previous instruction - but above, RedirectScopeStart redirects scopes to null. Shouldn't they both use the same new instruction? I'd be ok with just using null here.

src/linker/Linker/LinkerILProcessor.cs
sbomer: I'm a little confused by this - if removing the last instruction, we are redirecting scope starts to null. Shouldn't removing the first instruction be symmetric, and redirect scope ends to null?

src/linker/Linker/LinkerILProcessor.cs
vitek-karas: You're right.
This is basically a really weird case - the branch now points to nowhere (since there's nowhere to point to). But it can't be "Fixed" - we will get into invalid state no matter what.

src/linker/Linker/LinkerILProcessor.cs
vitek-karas: You're right. I think the correct behavior here is actually to remove the exception handler if this happens - basically if we try to redirect the "Start" of the exception handler to null, we should remove it.
That said this will never happen in the linker I think, branch removal handles removing of exception handlers on its own. So I would rather leave this as is - and just point it to null. Again - purposefully getting into invalid state, but so be it.

vitek-karas:
I was able to reproduce the failure from ArrayPool. So I added test cases in the TryFinally and TryCatch tests which repro the failure (they would put the `.endcatch` after the end of the method before the fix).

I also removed now unnecessary test attribute, since in my previous change the exception handlers are now validated within the instruction stream, so no need for a second attribute to validate these.

There are no thorough tests for the IL processor in the various corner cases, since we don't have a way in the linker to hit these currently. And I didn't want to try to add true unit tests - it didn't seem worth it.

I think this is now ready for review - since it provably fixes a real issue.
I was able to validate that with these fixes we don't corrupt the ArrayPool code anymore. I will try the full repro of that and update the PR here with the outcome.

vitek-karas:
@agocke We will want to take this for .NET 6 - not sure what you'll need for approval, please let me know.

src/linker/Linker/LinkerILProcessor.cs
```diff
@@ -0,0 +1,127 @@
...
+
+		public void Emit (OpCode opcode) => Append (_ilProcessor.Create (opcode));
+		public void Emit (OpCode opcode, TypeReference type) => Append (_ilProcessor.Create (opcode, type));
+		public void Emit (OpCode opcode, MethodReference method) => Append (_ilProcessor.Create (opcode, method));
+		public void Emit (OpCode opcode, VariableDefinition variable) => Append (_ilProcessor.Create (opcode, variable));
+
+		public void Emit (OpCode opcode, string value) => Append (_ilProcessor.Create (opcode, value));
+		public void InsertBefore (Instruction target, Instruction instruction)
+		{
+			// So when inserting before, all pointers have to be updated to the new "before" instruction.
```
sbomer, line 30: nit
```suggestion
			// When inserting before, all pointers have to be updated to the new "before" instruction.
```

sbomer (approved):

src/linker/Linker/LinkerILProcessor.cs
sbomer: Agreed - removing the handler might be reasonable, but I think it's ok not to do too much "magic" here if you are getting into such weird states. Same for the case where we remove the last instruction and it's a branch target - in this case you just need to know what you're doing.

vitek-karas:
I was able to verify that the repro case works with these changes - no mono AOT compiler failures. Will merge once CI passes.

PR #2221, opened 2021-08-20
Update ILLink.RoslynAnalyzer and ILLink.CodeFixProvider TFM to net6.0

mateoatr:


sbomer (approved):
LGTM. @agocke, any reason not to update the analyzers to target net6?

agocke (changes requested):
Can't do this -- analyzers must run against netstandard2.0

Issue #2225, opened 2021-08-23
IL Trimmer doesn't support net7.0


eerhardt:
See the code here:

https://github.com/mono/linker/blob/062c230b2674ea187da17d94742531e196e1e172/src/linker/Linker/LinkContext.cs#L701-L706

When updating dotnet/runtime to `net7.0`, we are running into new warnings because the following code is not working correctly:

https://github.com/mono/linker/blob/062c230b2674ea187da17d94742531e196e1e172/src/linker/Linker.Steps/MarkStep.cs#L1670-L1681

Warnings:

```
ILLink : Trim analysis error IL2026: System.Data.DataSet: Using method 'System.Data.DataSet.DataSet(SerializationInfo,StreamingContext)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Members from serialized types may be trimmed if not referenced directly. [C:\git\runtime6\src\libraries\src.proj]
ILLink : Trim analysis error IL2026: System.Data.DataTable: Using method 'System.Data.DataTable.DataTable(SerializationInfo,StreamingContext)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Members from serialized types may be trimmed if not referenced directly. [C:\git\runtime6\src\libraries\src.proj]
ILLink : Trim analysis error IL2026: System.Data.TypedTableBase`1: Using method 'System.Data.TypedTableBase<T>.TypedTableBase`1(SerializationInfo,StreamingContext)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Members from serialized types may be trimmed if not referenced directly. [C:\git\runtime6\src\libraries\src.proj]
```

cc @vitek-karas @sbomer @ViktorHofer

vitek-karas:
Working on this....

PR #2226, opened 2021-08-23
Make linker forward compatible with regards to framework versioning

vitek-karas:
Runtime is trying to update to .NET 7 versioning, but this breaks linker since it hardcodes knowledge of all versions.
This change makes it forward compatible by recognizing "higher than what I know".

No test as there's not a really good way to test this without going through a lot of trouble. Tested locally with runtime targeting 7.0 versioning.

Fixes https://github.com/mono/linker/issues/2225

### DIFF
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -45,9 +45,13 @@ namespace Mono.Linker
 
 	public enum TargetRuntimeVersion
 	{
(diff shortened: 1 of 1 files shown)


sbomer (approved):

MichalStrehovsky:
Is there any value in having an enum? It feels like it would be more futureproof (and straightforward) to just treat it as an integer (or System.Version if one feels fancy).

PR #2230, opened 2021-08-25
Validate unrecognized patterns against logged warnings

sbomer:
This changes the linker validation to check for unrecognized
reflection access patterns against logged messages. The plan is to do
the same for the analyzer tests, allowing us to make progress as we
replace UnrecognizedReflectionAccessPatternAttribute with
ExpectedWarningAttribute.

The UnrecognizedReflectionAccessPatternAttribute is now checked
against both the logged messages and recorded unrecognized reflection
patterns. We check that it matches the same number of patterns as
logged warnings for consistency. This required some changes to the
signature formats used for validation, since the signatures used in
this attribute now need to use the same format as the one we use for
logged warnings.

Changing the signature formats here uncovered some places where we
were still using the cecil signature format even for user-visible
warnings. These have been changed to use the format from
GetDisplayName.

Generic parameter dataflow validation treats the
UnrecognizedReflectionAccessPatternAttribute differently, using the
member name to refer to a generic type parameter. This inconsistency
has been lifted into a separate custom attribute argument that
specifies when the attribute refers to a generic
parameter. Non-generic method parameters had a similar problem, but
this was less common, so those usages were just replaced by
ExpectedWarningAttribute.


src/linker/Linker/MemberReferenceExtensions.cs
```diff
@@ -13,8 +15,15 @@ public static string GetDisplayName (this MemberReference member)
...
+			case IMemberDefinition memberDef:
+				var sb = new StringBuilder ();
+				if (memberDef.DeclaringType != null)
+					sb.Append (memberDef.DeclaringType.GetDisplayName ()).Append ('.');
+				sb.Append (memberDef.Name);
+				return sb.ToString ();
+
 			default:
-				return member.FullName;
+				throw new NotImplementedException ();
```
vitek-karas, line 26: I would probably still return `FullName` in this case (and assert).
Any exception from the linker is fatal to the publish - so unless we really can't figure out what to do, we should not hard fail. In this case the only failure is to show a nice name - so definitely not serious enough to fail the publish.

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -663,189 +662,368 @@ void VerifyKeptAllTypesAndMembersInAssembly (AssemblyDefinition linked)
...
+							List<MessageContainer> matchedMessages;
+							if ((bool) attr.ConstructorArguments[1].Value)
+								matchedMessages = loggedMessages.Where (m => Regex.IsMatch (m.ToString (), expectedMessage)).ToList ();
+							else
+								matchedMessages = loggedMessages.Where (m => m.ToString ().Contains (expectedMessage)).ToList (); ;
+							Assert.IsTrue (
+								matchedMessages.Count > 0,
+								$"Expected to find logged message matching `{expectedMessage}`, but no such message was found.{Environment.NewLine}Logged messages:{Environment.NewLine}{string.Join (Environment.NewLine, loggedMessages)}");
+
+							loggedMessages.Remove (matchedMessages.First ());
```
vitek-karas, line 701: Just to make sure - it was intentionall to require multiple attributes if the same message appears more than once, right?

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -663,189 +662,368 @@ void VerifyKeptAllTypesAndMembersInAssembly (AssemblyDefinition linked)
...
-						var matchedMessages = loggedMessages.Where (mc => {
-							if (mc.Category != MessageCategory.Warning || mc.Code != expectedWarningCodeNumber)
-								return false;
+							int expectedWarningCodeNumber = int.Parse (expectedWarningCode.Substring (2));
+							string expectedOrigin = null;
 
-							foreach (var expectedMessage in expectedMessageContains)
-								if (!mc.Text.Contains (expectedMessage))
-									return false;
+							var partialMatch = loggedMessages.Where (m => m.ToString ().Contains ("--StaticCtorTriggeredByFieldAccess--"));
```
vitek-karas, line 732: I assume this is left-over from debugging?

vitek-karas (approved):
Looks good!

PR #2236, opened 2021-08-27
Refactor PeVerifier

mrvoorhe:
We are migrating our unit test projects to run on net5.  However, we will continue to run our linker tests against a mono like profile (since that is what Unity is based on).

Eliminating PeVerifier entirely when building for net5 causes us problems.  We want to continue running it on the linker test executables.

This PR refactors how PeVerifier is made to do nothing when compiling with NETCOREAPP defined.  This approach lets the illink tests continue to operate as they do now.  While also giving us the ability to run our `Unity.Linker.Tests` project on net5, our test case assemblies compiled against a .NET Framework profile, and have our test framework continue to run PeVerifier

### DIFF
--- a/test/Mono.Linker.Tests/TestCasesRunner/PeVerifier.cs
+++ b/test/Mono.Linker.Tests/TestCasesRunner/PeVerifier.cs
@@ -1,9 +1,7 @@
-#if !NETCOREAPP
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


marek-safar:
LGTM but please run lint.sh to fix the build

mrvoorhe:
Thanks.  lint ran.  CI should be green this time.

PR #2242, opened 2021-08-30
Shorten paths

mrvoorhe:
I'm not entirely sure how the illink tests are not running into path length issues.

I see tests running with paths such as
```
C:\Unity\dev\unity-monolinker-2\artifacts\testcases\illink\PreserveDependencies\PreserveDependencyOnUnusedMethodInNonReferencedAssemblyWithEmbeddedXml\ref_resource_before_PreserveDependencyMethodInNonReferencedAssemblyLibrary\PreserveDependencyMethodInNonReferencedAssemblyLibrary.xml
```

Which is 285 characters.  Yet our test framework is not happy with these paths. 

This PR shortens a bunch of test and file names to get them under the max path length.

* Remove redundant name prefix in preserve dependency file names

* Remove redundant DynamicDependency prefix

* Shorten `ref_source` and `ref_resource` prefixes.  I don't love this one.  However, without it certain tests would require further name shortening.  Or some other scheme would have to be created to shorten the names needed for certain tests.

* Drop `EmbeddedLink` prefix

* Shorten some generic test names

* Shorten attribute test names

* Shorten some debugger attribute test names


mrvoorhe:
Opening this PR in case these changes are necessary.  Although I'm still curious why our test framework can't handle these paths yet illink can...Could it have to do with us still running our unit tests on .NET Framework whereas illink has been using .NET5 +?

mrvoorhe:
I think illink has been OK since the tests run on .NET Core.  In a cruel twist of irony, I was trying to take the test parallelization changes into our fork so that I could iterate faster migrating our `Unity.Linker.Tests` project to .NET 5 faster.  I'll migrate to net5 first and then give the test parallelization changes another try.

Closing this PR for now.

PR #2243, opened 2021-08-30
Normalize warning messages

mateoatr:
Surround member names printed inside warnings with "'" (e.g., `The 'memberName'...`).
End all warnings messages with a period.
Change user-facing mentions of "linker", "linking", etc. to "trimmer", "trimming", etc.


docs/design/constant-propagation-and-branch-removal.md
```diff
@@ -64,7 +64,7 @@ The implementation of this optimization is relatively complex since it's solving
...
 
-* Processing stack which stores ordered list of processing node, each node representing a method and additional data about it. The stack is processed by always taking the top of the stack and attempting to process that node. Nodes are always added to the top of the stack and are always removed from the top of the stack. In some cases nodes are "moved", that is a node which is not on the top of the stack is moved to the top of the stack. For this reason the stack is implemented as a linked list (so that it's easy to point to nodes in it as well as moves nodes around).
+* Processing stack which stores ordered list of processing node, each node representing a method and additional data about it. The stack is processed by always taking the top of the stack and attempting to process that node. Nodes are always added to the top of the stack and are always removed from the top of the stack. In some cases nodes are "moved", that is a node which is not on the top of the stack is moved to the top of the stack. For this reason the stack is implemented as a trimmed list (so that it's easy to point to nodes in it as well as moves nodes around).
```
sbomer, line 67: "linked list" :)
```suggestion
* Processing stack which stores ordered list of processing node, each node representing a method and additional data about it. The stack is processed by always taking the top of the stack and attempting to process that node. Nodes are always added to the top of the stack and are always removed from the top of the stack. In some cases nodes are "moved", that is a node which is not on the top of the stack is moved to the top of the stack. For this reason the stack is implemented as a linked list (so that it's easy to point to nodes in it as well as moves nodes around).
```

docs/illink-options.md
```diff
@@ -56,7 +56,7 @@ The linker can do the following things on all or individual assemblies
 - `copyused` - copy used assemblies to the output directory
 - `link` - link them to reduce their size
```
sbomer, line 57:
```suggestion
- `link` - trim them to reduce their size
```
Could you also look for other instances of "link"?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1787,10 +1787,10 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 				// or because of a copy assembly with a reference and so on) then we should not spam the warnings due to the type itself.
 				if (!(reason.Source is IMemberDefinition sourceMemberDefinition && sourceMemberDefinition.DeclaringType == type))
 					_context.LogWarning (
-						$"Attribute '{type.GetDisplayName ()}' is being referenced in code but the linker was " +
+						$"Attribute '{type.GetDisplayName ()}' is being referenced in code but the trimmer was " +
 						$"instructed to remove all instances of this attribute. If the attribute instances are necessary make sure to " +
-						$"either remove the linker attribute XML portion which removes the attribute instances, " +
-						$"or override the removal by using the linker XML descriptor to keep the attribute type " +
+						$"either remove the trimmed attribute XML portion which removes the attribute instances, " +
```
sbomer, line 1792:
```suggestion
						$"either remove the trimmer attribute XML portion which removes the attribute instances, " +
```

sbomer (approved):
I'm not sure we need to avoid using "linker" terminology when talking about illink.dll, but I don't think it hurts either. LGTM with the suggested fixes, but please get a second opinion on the wording change.

mateoatr:
> I'm not sure we need to avoid using "linker" terminology when talking about illink.dll, but I don't think it hurts either. LGTM with the suggested fixes, but please get a second opinion on the wording change.

Agreed. It's not that important how we refer to the trimmer in the docs inside this repo (I think it's fine if we keep using the term "linker"). The most important thing is to be consistent with the terminology on user-facing messages and on msdocs, where we should definitely use trimmer/trimming/trim* instead.

vitek-karas (approved):
I agree this is a bit tricky on the terminology - but I guess being consistent doesn't hurt.

agocke (approved):
I agree, not strictly necessary to change lines in the repo docs, but it's good to have consistency. Warning message fixups also look good.

PR #2244, opened 2021-09-01
Enable nullable warnings for linker

sbomer:
The next step will be to start enabling nullable annotations, which can be done incrementally. A few of the warnings caught real bugs. :)

@agocke I started looking into using wrapper types to ensure completeness in some cases. For example we sometimes rely on the fact that a known set of types implement `IMemberDefinition` or `ICustomAttributeProvider` to ensure the validity of a cast to a concrete type. However using wrapper types was starting to look pretty invasive so for now I've just replaced the relevant nullable warnings for patterns like `member as TypeReference` with explicit casts.


marek-safar (approved):

src/linker/Linker.Steps/DiscoverCustomOperatorsHandler.cs
```diff
@@ -133,7 +133,7 @@ TypeDefinition NonNullableType (TypeReference type)
 			// that we need to unwrap until we get to the Nullable<T>
 			while (!type.IsGenericInstance) {
 				Debug.Assert (type is TypeSpecification);
```
agocke, line 135: Not sure this assert is providing much value over doing the cast and getting a classcastexception

agocke (approved):
LGTM thanks!

vitek-karas (approved):

PR #2248, opened 2021-09-02
Stop using git protocol for submodule

akoeplinger:
GitHub is removing support for unencrypted git soon: https://github.blog/2021-09-01-improving-git-protocol-security-github/

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "cecil"]
 	path = external/cecil
-	url = git://github.com/mono/cecil.git
+	url = https://github.com/mono/cecil.git


marek-safar:
Could you backport this to release/6.0 too?

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "cecil"]
 	path = external/cecil
-	url = git://github.com/mono/cecil.git
+	url = https://github.com/mono/cecil.git
```
premun, line 3: @akoeplinger should this be updated or we just rely on GitHub's auto-redirect?

(just asking, have no preference here)

akoeplinger: yeah, good idea to be consistent: https://github.com/dotnet/linker/pull/3132

PR #2250, opened 2021-09-02
Update edited document before continuing code fixing

### LINKED ISSUE #2149: IL2026 should use custom fix all provider
https://github.com/dotnet/roslyn/issues/54602#issuecomment-876837220

mateoatr:
Fixes #2149.

### DIFF
--- a/src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
+++ b/src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
@@ -32,55 +32,52 @@ namespace ILLink.CodeFix
 
 		protected async Task BaseRegisterCodeFixesAsync (CodeFixContext context)
 		{
-			var root = await context.Document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false);
-
+			var document = context.Document;
+			var root = await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false);
 			var diagnostic = context.Diagnostics.First ();
-			var diagnosticSpan = diagnostic.Location.SourceSpan;
+			SyntaxNode targetNode = root!.FindNode (diagnostic.Location.SourceSpan);
+			if (FindAttributableParent (targetNode, AttributableParentTargets) is not SyntaxNode attributableNode)
+				return;
 
-			SyntaxNode targetNode = root!.FindNode (diagnosticSpan);
-			CSharpSyntaxNode? declarationSyntax = FindAttributableParent (targetNode, AttributableParentTargets);
-			if (declarationSyntax is not null) {
-				var semanticModel = await context.Document.GetSemanticModelAsync (context.CancellationToken).ConfigureAwait (false);
-				var symbol = semanticModel!.Compilation.GetTypeByMetadataName (FullyQualifiedAttributeName);
-				var document = context.Document;
-				var editor = new SyntaxEditor (root, document.Project.Solution.Workspace);
-				var generator = editor.Generator;
-				var attrArgs = GetAttributeArguments (semanticModel, targetNode, declarationSyntax, generator, diagnostic);
-				var codeFixTitle = CodeFixTitle.ToString ();
+			var model = await document.GetSemanticModelAsync (context.CancellationToken).ConfigureAwait (false);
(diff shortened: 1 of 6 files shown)


test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
```diff
@@ -144,10 +144,8 @@ public class F
 	VerifyCS.Diagnostic (DiagnosticId.RequiresUnreferencedCode).WithSpan (17, 31, 17, 37).WithArguments ("C.M1()", " message.", ""),
 	// /0/Test0.cs(24,31): warning IL2026: Using member 'C.M1()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. message.
 	VerifyCS.Diagnostic (DiagnosticId.RequiresUnreferencedCode).WithSpan (24, 31, 24, 37).WithArguments ("C.M1()", " message.", "")
-			}, new[] {
-	// /0/Test0.cs(27,10): error CS7036: There is no argument given that corresponds to the required formal parameter 'message' of 'RequiresUnreferencedCodeAttribute.RequiresUnreferencedCodeAttribute(string)'
```
agocke, line 148: I think these results were correct -- we don't want to fill in the message for `public` members, since that message will be visible to external users. The code owner should be responsible for writing a message for their public APIs.

agocke:
OK I finally figured this out -- the problem is not the lack of a FixAll Provider, it's that the SyntaxEditor is shared via lambda capturing among multiple different invocations of the analyzer. Since the SyntaxEditor is mutable, this causes a race in addition attributes to the tree.

Your change fixes the problem by avoiding sharing the editor instance across multiple instances.

It's probably hard to write a test to reveal this problem (and probably not worth it to write one). If you can narrow down the fix to just that area I think we should take this for 6.0

src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
```diff
@@ -32,55 +32,52 @@ public sealed override FixAllProvider GetFixAllProvider ()
...
-					Simplifier.AddImportsAnnotation);
+			var editor = await DocumentEditor.CreateAsync (document, cancellationToken).ConfigureAwait (false);
+			var generator = editor.Generator;
+			var attribute = CreateNewAttribute (generator, attributeSymbol, attributeArguments);
+			editor.AddAttribute (targetNode, attribute);
+			return editor.GetChangedDocument ();
+		}
 
-			editor.AddAttribute (containingDecl, newAttribute);
+		private static SyntaxNode CreateNewAttribute (
```
agocke, line 71: Is this extra method really necessary? It doesn't look like it's used anywhere else

src/ILLink.CodeFix/RequiresUnreferencedCodeCodeFixProvider.cs
```diff
@@ -32,15 +31,13 @@ public class RequiresUnreferencedCodeCodeFixProvider : BaseAttributeCodeFixProvi
 
 		public sealed override Task RegisterCodeFixesAsync (CodeFixContext context) => BaseRegisterCodeFixesAsync (context);
 
-		protected override SyntaxNode[] GetAttributeArguments (SemanticModel semanticModel, SyntaxNode targetNode, CSharpSyntaxNode containingDecl, SyntaxGenerator generator, Diagnostic diagnostic)
+		protected override SyntaxNode[] GetAttributeArguments (SyntaxGenerator syntaxGenerator, ISymbol attributableSymbol, ISymbol targetSymbol, Diagnostic diagnostic)
```
agocke, line 34: I think you can minimize changes to this API by moving `SyntaxGenerator` back to being the second-to-last parameter

agocke (approved):
LGTM aside from minor suggestions

src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
mateoatr: Removed.

agocke:
@mateoatr Can you backport to 6.0?

Issue #2251, opened 2021-09-02
Suppressing analyzer diagnostics in IDE should generate a GlobalSuppresions.cs file using `UnconditionalSuppressMessage`


mateoatr:
Currently, if a user chooses to suppress the diagnostics produced by the trimmer analyzer and clicks on the option given by the IDE, a global suppressions file will be generated with a list of assembly-level `SuppressMessage` attributes. However, this does not work for the diagnostics generated by the trimmer analyzer, since these must be unconditionally suppressed, otherwise the trimmer will keep emitting warnings when publishing. We should look into generating a global suppressions file that uses `UnconditionalSuppressMessage` instead.

Example of the GlobalSuppressions.cs file that is today generated:

```C#

// This file is used by Code Analysis to maintain SuppressMessage
// attributes that are applied to this project.
// Project-level suppressions either have no target or are given
// a specific target and scoped to a namespace, type, member, etc.

using System.Diagnostics.CodeAnalysis;

[assembly: SuppressMessage("Trimming", "IL2026:Methods annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code", Justification = "<Pending>")]
```

PR #2252, opened 2021-09-02
Tag analyzer diagnostics as non configurable

mateoatr:
This PR tags the set of diagnostics produced by the trimmer analyzer as non configurable so that the option to generate a global suppression file is not shown to the user (since this would not work for these diagnostics, see #2251).


agocke:
Unfortunately I don't really know a good way to test this, I assume you just looked at the VS UI and confirmed that the option was hidden?

mateoatr:
> Unfortunately I don't really know a good way to test this, I assume you just looked at the VS UI and confirmed that the option was hidden?

This is correct. I could not figure a way to have a written test for this either, so I just validated it manually.

agocke (approved):
LGTM

agocke:
Feel free to backport to 6.0 after merging

PR #2253, opened 2021-09-02
Nullable annotations part 1

sbomer:
Enable nullable annotations for LinkContext and the XML processing classes. I chose these as a starting point for no reason in particular, except that they produced a lot of warnings. Also includes other small fixes for things I noticed. In particular PTAL at the removed `ReportUnresolved` calls in `LinkContext.Resolve` - those looked like they were producing redundant warnings for cached `null` results.

### DIFF
--- a/src/linker/Linker.Steps/OutputWarningSuppressions.cs
+++ b/src/linker/Linker.Steps/OutputWarningSuppressions.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Diagnostics;
 using System.IO;
 
 namespace Mono.Linker.Steps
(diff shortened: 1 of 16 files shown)


src/linker/Linker.Steps/OutputWarningSuppressions.cs
```diff
@@ -15,6 +16,7 @@ protected override bool ConditionToProcess ()
 
 		protected override void Process ()
 		{
+			Debug.Assert (Context.WarningSuppressionWriter != null);
```
marek-safar, line 19: I think the better change would be `?.` bellow

src/linker/Linker/LinkContext.cs
```diff
@@ -527,7 +488,7 @@ public bool HasCustomData (string key)
 			return _parameters.ContainsKey (key);
 		}
 
-		public bool TryGetCustomData (string key, out string value)
+		public bool TryGetCustomData (string key, out string? value)
```
marek-safar, line 491: nit: value should be `[NotNullWhen(true)]`

vitek-karas (approved):

PR #2254, opened 2021-09-03
Warn on RUC annotated ctors invoked through `new()` constraint in analyzer

### LINKED ISSUE #2165: RUC analyzer doesn't produce a warning for type substitution with `new()` annotation
Example code:

```C#
[RequiresUnreferencedCode(...)]
class C
{
}

static void M<T>() where T : new()
{
}
M<C>(); // should produce a warning
```

mateoatr:
Fixes #2165.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
@@ -7,8 +7,9 @@ using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
 using ILLink.Shared;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
-using Microsoft.CodeAnalysis.Operations;
 
 namespace ILLink.RoslynAnalyzer
 {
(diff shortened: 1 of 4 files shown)


test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
```diff
@@ -778,8 +778,7 @@ class NewConstrainTestType
 
 			static void GenericMethod<T> () where T : new() { }
 
-			// https://github.com/mono/linker/issues/2117
-			[ExpectedWarning ("IL2026", "--NewConstrainTestType.ctor--", GlobalAnalysisOnly = true)]
+			[ExpectedWarning ("IL2026", "--NewConstrainTestType.ctor--")]
```
vitek-karas, line 781: We should also add a test when the new constraint is on a type parameter of a type (this one is on method).
For linker there's basically no difference, but for the analyzer there is a big one. I don't know if the analyzer will already handle this or not. Feel free to add it in a separate PR.

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresUnreferencedCodeCapability.cs
mateoatr: Added a test and made the changes in this same PR. PTAL.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -31,6 +33,49 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
...
+			var typeArgs = ImmutableArray<ITypeSymbol>.Empty;
+			if (operationContext.Operation is IObjectCreationOperation objectCreationOperation &&
+				objectCreationOperation.Type is INamedTypeSymbol objectType && objectType.IsGenericType) {
+				typeParams = objectType.TypeParameters;
+				typeArgs = objectType.TypeArguments;
+			} else if (operationContext.Operation is IInvocationOperation invocationOperation &&
+				invocationOperation.TargetMethod is IMethodSymbol targetMethod && targetMethod.IsGenericMethod) {
+				typeParams = targetMethod.TypeParameters;
+				typeArgs = targetMethod.TypeArguments;
+			}
```
vitek-karas, line 55: `else return` ? Would seem to make sense.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -31,6 +33,49 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
...
+			if (FindContainingSymbol (operationContext, DiagnosticTargets.All) is not ISymbol containingSymbol ||
+			containingSymbol.HasAttribute (RequiresUnreferencedCodeAttribute))
+				return;
+
+			var typeParams = ImmutableArray<ITypeParameterSymbol>.Empty;
+			var typeArgs = ImmutableArray<ITypeSymbol>.Empty;
+			if (operationContext.Operation is IObjectCreationOperation objectCreationOperation &&
+				objectCreationOperation.Type is INamedTypeSymbol objectType && objectType.IsGenericType) {
+				typeParams = objectType.TypeParameters;
+				typeArgs = objectType.TypeArguments;
```
vitek-karas, line 50: This only reacts to calling a constructor on a generic type.
Code like this should still warn though:
```C#
static class MyTest<T> where T : new ()
{
    static void DoNothing() {}
}

void Test()
{
    // IL2026
    MyTest.DoNothing<ClassWithRUCOnCtor>();
}
```

I must admit that I don't know enough about analyzers to tell what type of solution would be the best here. Linker will warn anywhere the generic is instantiated. Even cases like `typeof(MyType<ClassWithRUCOnCtor>)` will generate the warning.

This is obviously not that common, so it's OK to solve this later...

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -31,6 +33,49 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
 				operationContext.Operation.Syntax.GetLocation ()));
 		};
 
+		[SuppressMessage ("MicrosoftCodeAnalysisPerformance", "RS1008",
+			Justification = "This action is registered through a compilation start action, so that the instances which " +
+			"can register this operation action will not outlive a compilation's lifetime, avoiding the possibility of " +
+			"this data causing stale compilations to remain in memory.")]
```
vitek-karas, line 39: Can you please explain why this action causes the warning and the one above doesn't? I don't see how they're really different...

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
mateoatr: I moved the analysis context used here to a `SyntaxNodeAnalysisContext`; this way we look for syntax nodes that represent generic names and check, for their type parameters, if these have any constructor constraints. I think this would cover up all cases. I've added the scenario you outline above (and the analyzer warns now).

vitek-karas (approved):

Issue #2260, opened 2021-09-06
Linker fails when resolving type reference in sweep step


vitek-karas:
This was caused by #2234.

The problem is that first we call `SweepAssemblyReferences` on pretty much all assemblies. This will eventually call into `AssemblyReferenceCorrector.UpdateScopeOfTypeReference` which call `LinkContext.TryResolve`. In the failing case the type reference first points to "netstandard". It gets resolved to "System.Collections" and the type reference instance is overwritten to point to that. And the TryResolve cache remembers that this typeref points to "System.Collection" type-def.

Later on - we call ProcessAssemblyAction which in turn calls SweepAssembly on "System.Collection", this will remove the type-def from the assembly (not marked). This means that the type-def's Scope is reset to 'null'.

Now we eventually call ProcessAssemblyAction on the assembly which contained the type-ref. This will again end up calling `AssemblyReferenceCorrector.UpdateScopeOfTypeReference` and `TryResolve` on the same type-ref. Now the result is cached, so it returns the type-def which is now pointing to `null` scope. We try to "import" this type-def into the target assembly, and this fails, since it doesn't know where type type-def comes from.

Before the change, the `TryResolve` was a direct call to `Resolve`. In the first time around - it would return the same value (type-def). But second time around, the type-ref doesn't resolve anymore, since the type it points to has been removed, to `Resolve` returned null - and SweepStep has handling for this case - it instead resolves the assembly ref of the type-ref - which now points to "System.Collection" (was fixed up during the first pass) - so it imports that scope and leaves the type-ref pointing to that scope. This means it never actually relies on the type-ref to resolve - which it won't. In this case it's writing an unresolvable type-ref.

vitek-karas:
The full problem description

Setup
`Base.dll` - library which declares a `Base` type which defines an abstract `VMethod`
`Utils.dll` - library with several types, contains `RemovedType` and `UsedType`
`Implementation.dll` - library which provides an `Implementation` with `VMethod`. The body of `VMethod` references `RemovedType` and it's the only reference to that type in the app.

The assemblies are loaded by the linker in this order (this is easy to achieve).

MarkStep
* The uses `Base` and `Implementation` types. But it never instantiates `Implementation. It does call `Base.VMethod`.
* MarkStep will mark `Base`, `Implementation` and `Base.VMethod`.
* It will not see any direct reference to `Implementation.VMethod`. Since `Implementation` is not instantiated it will also try to remove the method's body
[...]

PR #2263, opened 2021-09-07
Add a test for #2260

vitek-karas:
This uncovers a null ref in sweepstep.
The test is order dependent:
- Assembly (librarywithnonemtpy) has a virtual method which referes to type from second library (library). This processes the type reference.
- The type is removed from library
- The virtual method is re-processed again -> hits a type def which does not belong to any assembly/scope

For this to happen the virtual method must be processed after the type has been removed.
This also requires the method to be meant for removal later on (the body will be rewritten to throw), since otherwise the type would have been marked.
This also requires the library with the removed type to be kept (so something else in it must be kept).

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Inheritance.AbstractClasses/NoKeptCtor/OverrideRemoval/Dependencies/OverrideOfAbstractIsKeptNonEmptyLibraryWithNonEmpty.cs
+++ b/test/Mono.Linker.Tests.Cases/Inheritance.AbstractClasses/NoKeptCtor/OverrideRemoval/Dependencies/OverrideOfAbstractIsKeptNonEmptyLibraryWithNonEmpty.cs
@@ -1,0 +1,23 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Mono.Linker.Tests.Cases.Inheritance.AbstractClasses.NoKeptCtor.OverrideRemoval.Dependencies
+{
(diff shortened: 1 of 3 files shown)


marek-safar (approved):

test/Mono.Linker.Tests.Cases/Inheritance.AbstractClasses/NoKeptCtor/OverrideRemoval/Dependencies/OverrideOfAbstractIsKeptNonEmptyLibraryWithNonEmpty.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Mono.Linker.Tests.Cases.Inheritance.AbstractClasses.NoKeptCtor.OverrideRemoval.Dependencies
+{
+	public class OverrideOfAbstractIsKeptNonEmptyLibraryWithNonEmpty :
```
sbomer, line 13: nit: I'm not sure what the "non-empty" in `NonEmptyLibrary` or `WithNonEmpty` refer to - maybe the names could be simplified?

sbomer (approved):

test/Mono.Linker.Tests.Cases/Inheritance.AbstractClasses/NoKeptCtor/OverrideRemoval/Dependencies/OverrideOfAbstractIsKeptNonEmptyLibraryWithNonEmpty.cs
vitek-karas: The idea was to keep the names starting with the name of the test - so that it's easy to tell which dependencies belong to which test.

PR #2264, opened 2021-09-07
Workaround NRE in SweepStep

vitek-karas:
Basically a workaround for #2260. This simply goes back to using the `Resolve` method and avoid the cache for now.

### DIFF
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -568,7 +568,9 @@ namespace Mono.Linker.Steps
 		{
 			readonly AssemblyDefinition assembly;
 			readonly DefaultMetadataImporter importer;
+#pragma warning disable IDE0052 // Remove unread private members
 			readonly ITryResolveMetadata resolver;
+#pragma warning restore IDE0052 // Remove unread private members
 
 			HashSet<TypeReference> updated;
 			bool changedAnyScopes;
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -568,7 +568,9 @@ struct AssemblyReferencesCorrector
 		{
 			readonly AssemblyDefinition assembly;
 			readonly DefaultMetadataImporter importer;
+#pragma warning disable IDE0052 // Remove unread private members
 			readonly ITryResolveMetadata resolver;
```
sbomer, line 572: Do we still think the `resolver` is going to be useful for `AssemblyReferencesCorrector`? Maybe we could just remove the field.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -925,7 +927,10 @@ void UpdateScopeOfTypeReference (TypeReference type)
 				//
 				// Resolve to type definition to remove any type forwarding imports
 				//
-				TypeDefinition td = resolver.TryResolve (type);
+				// Workaround for https://github.com/mono/linker/issues/2260
```
sbomer, line 930: nit: it would be helpful to add a short comment about why the cache shouldn't be used here - something like "This needs to use Cecil's Resolve because the LinkContext.Resolve cache has been invalidated by sweeping unused types".

sbomer (approved):
Thank you for fixing this!

PR #2275, opened 2021-09-16
Add COMAnalyzer

mateoatr:
This adds support for `IL2050`. It is intended to be used for all other diagnostics related to COM in the analyzers. I didn't add any new tests -- we are reusing the ones under `Interop`.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
@@ -1,0 +1,82 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Immutable;
+using ILLink.Shared;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace ILLink.RoslynAnalyzer
+{
+	[DiagnosticAnalyzer (LanguageNames.CSharp)]
+	public sealed class COMAnalyzer : DiagnosticAnalyzer
+	{
+		private const string DllImportAttribute = nameof (DllImportAttribute);
+		private const string MarshalAsAttribute = nameof (MarshalAsAttribute);
+
+		static readonly DiagnosticDescriptor s_correctnessOfCOMCannotBeGuaranteed = DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.CorrectnessOfCOMCannotBeGuaranteed,
+			helpLinkUri: "https://docs.microsoft.com/en-us/dotnet/core/deploying/trim-warnings/il2050");
+
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create (s_correctnessOfCOMCannotBeGuaranteed);
+
+		public override void Initialize (AnalysisContext context)
+		{
+			context.EnableConcurrentExecution ();
+			context.ConfigureGeneratedCodeAnalysis (GeneratedCodeAnalysisFlags.ReportDiagnostics);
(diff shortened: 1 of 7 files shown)


src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
```diff
@@ -0,0 +1,124 @@
...
+					typedConstant.Type != null && typedConstant.Type.IsUnmanagedType) {
+					var unmanagedType = typedConstant.Value;
+					switch (unmanagedType) {
+					case (int) UnmanagedType.IUnknown:
+					case (int) UnmanagedType.IDispatch:
+					case (int) UnmanagedType.Interface:
+						return true;
+
+					default:
+						break;
```
sbomer, line 68: The linker's logic has an explicit `nativeType == NativeType.None` check - do we need something like that? (Is there a case where the marshal info can change any of the below `true` returns to be not COM related?)

src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
```diff
@@ -0,0 +1,124 @@
...
+					}
+				}
+
+				if (IsInterface (symbol))
+					return true;
+
+				if (symbol is not IParameterSymbol parameterSymbol)
```
sbomer, line 75: Don't we need to consider the below cases also for return types? Or are those also represented as `IParameterSymbol`?

src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
```diff
@@ -0,0 +1,124 @@
...
+				} else if (IsInterface (parameterSymbol.Type)) {
+					// Interface types marshal as COM by default
+					return true;
+				} else if (parameterType.ContainingNamespace.Name == "System" &&
+					parameterType.Name == "MulticastDelegate") {
+					// Delegates are special cased by interop
+					return false;
+				} else if (parameterType.ContainingNamespace.Name == "System.Runtime.InteropServices" &&
+					(parameterType.Name == "CriticalHandle" || parameterType.Name == "SafeHandle")) {
+					// Subclasses of CriticalHandle and SafeHandle are special cased by interop
```
sbomer, line 100: Do we need to check for subclasses here?

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -83,12 +85,16 @@ public static void RunTest<TAnalyzer> (MemberDeclarationSyntax m, List<Attribute
 			var builder = ImmutableDictionary.CreateBuilder<string, List<string>> ();
 
 			foreach (var file in GetTestFiles ()) {
-				var dirName = Path.GetFileName (Path.GetDirectoryName (file))!;
-				if (builder.TryGetValue (dirName, out var sources)) {
+				var dirName = Path.GetDirectoryName (file);
```
sbomer, line 88: nit: Might just be me, but I find it confusing that `GetDirectoryName` gives back the directory path, not just the name. Would you mind renaming things to make this a bit more readable? For example `dirName` -> `directory`, maybe introduce a `parentDirectory` variable?

sbomer (commented):
Nice!

src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
mateoatr: I added a check to see if the typed constant value exists in `UnmanagedType` (which mimics the check of `nativeType == NativeType.None` in Cecil).

src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
mateoatr: Yup. Added a helper method in ISymbolExtensions for this. I moved the `IsInterface` method there as well.

sbomer (approved):
Thank you, LGTM!

marek-safar:
@mateoatr I had to revert this change as it caused test failures, please re-submit it after rebasing.

PR #2276, opened 2021-09-16
Mark forwarders for all TypeReferences

sbomer:
Fixes https://github.com/dotnet/linker/issues/2267.

This fixes the issue by taking the first approach mentioned in https://github.com/dotnet/linker/issues/2267#issuecomment-916379762. We now mark type forwarders referenced from any `TypeReference` we discover for a copy assembly, including `TypeReference`s to built-in types which only exist as bytes in metadata rather than typeref rows.

The walking logic in `TypeReferenceMarker` is copied from `AssemblyReferencesCorrector` in `SweepStep` and adjusted so that instead of updating the scopes of typereferences, it just marks any type forwarders they point to. The walking logic was duplicated to avoid touching `SweepStep` to reduce the risk of this change.

This includes a change to the test infrastructure to build tests against `System.Runtime`, which is necessary to reproduce the issue (since it has to do with references to builtin types forwarded from `System.Runtime` to `System.Private.CoreLib`). That change caused the compiler to produce different code for event methods - before the change, there were unnecessary locals in the generated code. Either way the unreachablebodies optimization removes all locals for unreachable event methods, which is why I removed the `ExpectLocalsModified` from a few testcases.

I plan to factor out the duplicated logic in a follow-up that we won't consider taking into .NET6.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1380,7 +1380,12 @@ namespace Mono.Linker.Steps
 			foreach (TypeDefinition type in module.Types)
 				MarkEntireType (type, new DependencyInfo (DependencyKind.TypeInAssembly, assembly));
 
-			// Mark scopes of type references and exported types by traversing the assembly.
+			foreach (ExportedType exportedType in module.ExportedTypes) {
+				MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.ExportedType, assembly));
+				MarkingHelpers.MarkForwardedScope (new TypeReference (exportedType.Namespace, exportedType.Name, module, exportedType.Scope));
+			}
+
+			// Mark scopes of type references by traversing the assembly.
 			new TypeReferenceMarker (assembly, MarkingHelpers).Process ();
 
 			// Also mark the scopes of metadata typeref rows to cover any not discovered by the traversal.
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1380,11 +1380,11 @@ namespace Mono.Linker.Steps
 			foreach (TypeDefinition type in module.Types)
 				MarkEntireType (type, new DependencyInfo (DependencyKind.TypeInAssembly, assembly));
 
-			foreach (ExportedType exportedType in module.ExportedTypes) {
-				MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.ExportedType, assembly));
-				MarkingHelpers.MarkForwardedScope (new TypeReference (exportedType.Namespace, exportedType.Name, module, exportedType.Scope));
-			}
+			// Mark scopes of type references and exported types by traversing the assembly.
+			new TypeReferenceMarker (assembly, MarkingHelpers).Process ();
 
+			// Also mark the scopes of metadata typeref rows to cover any not discovered by the traversal.
+			// This can happen when the compiler emits typerefs into IL which aren't strictly necessary per ECMA 335.
 			foreach (TypeReference typeReference in module.GetTypeReferences ())
 				MarkingHelpers.MarkForwardedScope (typeReference);
 		}
--- a/src/linker/Linker/TypeReferenceMarker.cs
+++ b/src/linker/Linker/TypeReferenceMarker.cs
@@ -22,7 +22,7 @@ namespace Mono.Linker
 			visited = null;
 		}
 
-		// Traverse the assembly and mark exported types, and the scopes of discovered type references.
+		// Traverse the assembly and mark the scopes of discovered type references (but not exported types).
 		// This includes scopes referenced by Cecil TypeReference objects that don't represent rows in the typeref table,
 		// such as references to built-in types, or attribute arguments which encode type references as strings.
 		public void Process ()
@@ -41,9 +41,6 @@ namespace Mono.Linker
 					WalkScopes (type);
 				}
 			}
-
-			if (mmodule.HasExportedTypes)
-				WalkTypeScope (mmodule.ExportedTypes);
 
 			visited = null;
 		}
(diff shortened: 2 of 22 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1380,11 +1380,11 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
 			foreach (TypeDefinition type in module.Types)
 				MarkEntireType (type, new DependencyInfo (DependencyKind.TypeInAssembly, assembly));
 
-			foreach (ExportedType exportedType in module.ExportedTypes) {
-				MarkingHelpers.MarkExportedType (exportedType, module, new DependencyInfo (DependencyKind.ExportedType, assembly));
-				MarkingHelpers.MarkForwardedScope (new TypeReference (exportedType.Namespace, exportedType.Name, module, exportedType.Scope));
-			}
```
vitek-karas, line 1386: I must admit I don't understand why we're removing marking of type forwarders... the rest of the changes in this method make sense, but this one I don't get.

sbomer: The forwarders are marked by the `TypeReferenceMarker` at https://github.com/dotnet/linker/pull/2276/files#diff-279c983ce836c70ed323727a69c23d650e18e177e3ec103cf15805ad69e9e017R151-R152.

I can move that back into this method to reduce the delta if you prefer.

vitek-karas: I would prefer we don't remove anything from the code which is there - ideally we should just add. I know that this should be the same... but still.

sbomer: Done!

src/linker/Linker/TypeReferenceMarker.cs
```diff
@@ -0,0 +1,343 @@
...
+
+					if (m.HasBody)
+						WalkTypeScope (m.Body);
+				}
+			}
+
+			if (typeDefinition.HasProperties) {
+				foreach (var p in typeDefinition.Properties) {
+					WalkCustomAttributesTypesScopes (p);
+					// p.PropertyType is not saved
```
vitek-karas, line 107: For future reference - this feels dangerous - not because of saving, but because of custom steps. We update the type refs in sweep step - but we don't update this one (intentionally). Again - nothing to do with this change, just that I noticed this...

vitek-karas (approved):

test/Mono.Linker.Tests/TestCasesRunner/TestCaseMetadaProvider.cs
```diff
@@ -164,16 +164,23 @@ public virtual IEnumerable<string> GetCommonReferencedAssemblies (NPath workingD
...
+			yield return Path.Combine (referenceDir, "System.Collections.dll");
+			yield return Path.Combine (referenceDir, "System.ComponentModel.TypeConverter.dll");
+			yield return Path.Combine (referenceDir, "System.Console.dll");
+			yield return Path.Combine (referenceDir, "System.Diagnostics.Tracing.dll");
+			yield return Path.Combine (referenceDir, "System.Linq.Expressions.dll");
+			yield return Path.Combine (referenceDir, "System.ObjectModel.dll");
+			yield return Path.Combine (referenceDir, "System.Reflection.Emit.dll");
+			yield return Path.Combine (referenceDir, "System.Runtime.dll");
+			yield return Path.Combine (referenceDir, "System.Runtime.Extensions.dll");
+			yield return Path.Combine (referenceDir, "System.Runtime.InteropServices.dll");
```
marek-safar, line 183: Can we keep the list small? Every added assembly will slow down every test run.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1385,6 +1385,11 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
 				MarkingHelpers.MarkForwardedScope (new TypeReference (exportedType.Namespace, exportedType.Name, module, exportedType.Scope));
 			}
 
+			// Mark scopes of type references by traversing the assembly.
+			new TypeReferenceMarker (assembly, MarkingHelpers).Process ();
```
marek-safar, line 1389: Which type references we are converting are missing in module.GetTypeReferences () ?

test/Mono.Linker.Tests/TestCasesRunner/TestCaseMetadaProvider.cs
sbomer: I removed a few from the common list and made them individual testcase references instead. System.Collections and System.Runtime.InteropServices are referenced by test dependencies, so I think they need to be common.

src/linker/Linker.Steps/MarkStep.cs
sbomer: `GetTypeReferences` misses ones like in the original issue, for example when a field type is a builtin type. There's no typeref row for the reference in metadata in that case.

vitek-karas:
@sbomer - can you please prepare a backport to 6.0 RC2 for this? My understanding is that Xamarin will need this - @rolfbjarne can you please confirm?

rolfbjarne:
> @sbomer - can you please prepare a backport to 6.0 RC2 for this? My understanding is that Xamarin will need this - @rolfbjarne can you please confirm?

Correct, we need this.

sbomer:
This went into the RC2 SDK with this change: https://github.com/dotnet/sdk/pull/21118.

vitek-karas:
Sorry - I didn't notice this PR was for the 6.0 branch.

rolfbjarne:
I can confirm this is in RC2, but it's not in the 6.0.1xx branch: https://github.com/dotnet/sdk/blob/7156781d4c687ddf6e0f24d244c0c239a92dd02e/eng/Version.Details.xml#L123-L131

vitek-karas:
@agocke - do you know what branch flows to the 6.0.1xx SDK? It might be that the answer is "none yet" and we would definitely need to fix that.

agocke:
Ah I thought there was automerge in sdk like there is in runtime. I'll insert the same build into both for now.

PR #2278, opened 2021-09-16
Enable nullable on annotations

tlakollo:
Enable nullable on Annotations.cs
Change the MarkStep condition to check for null after TryResolve

Fixes https://github.com/dotnet/linker/issues/2274
I couldn't add a test case because I couldn't come up with a scenario in which the TypeReference Type.BaseType is not null but resolving the TypeReference is null

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1817,9 +1817,10 @@ namespace Mono.Linker.Steps
 			MarkCustomAttributes (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
 			MarkSecurityDeclarations (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
 
-			if (type.BaseType is not null &&
+			var baseType = _context.TryResolve (type.BaseType);
+			if (baseType is not null &&
 				!_context.Annotations.HasLinkerAttribute<RequiresUnreferencedCodeAttribute> (type) &&
-				_context.Annotations.TryGetLinkerAttribute (_context.TryResolve (type.BaseType), out RequiresUnreferencedCodeAttribute effectiveRequiresUnreferencedCode)) {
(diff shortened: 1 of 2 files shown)


src/linker/Linker/Annotations.cs
```diff
@@ -244,7 +245,7 @@ public bool IsInstantiated (TypeDefinition type)
 			return marked_instantiated.Contains (type);
 		}
 
-		public void MarkRelevantToVariantCasting (TypeDefinition type)
+		public void MarkRelevantToVariantCasting (TypeDefinition? type)
```
sbomer, line 248: I'd prefer to make the type non-nullable, and require callers to decide what to do if for example a type doesn't resolve.

sbomer (approved):
Thanks, LGTM! I think we will need to make a separate change in the 6.0 branch that's a more targeted fix for the BaseType issue (since none of the nullable annotation changes exist in that branch).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1817,9 +1817,9 @@ protected internal virtual TypeDefinition MarkType (TypeReference reference, Dep
 			MarkCustomAttributes (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
 			MarkSecurityDeclarations (type, new DependencyInfo (DependencyKind.CustomAttribute, type));
 
-			if (type.BaseType is not null &&
+			if (_context.TryResolve (type.BaseType) is TypeDefinition baseType &&
```
agocke, line 1820: Is nullable actually enabled in this file?

tlakollo: Nullable is not enabled on MarkStep, but methods in Annotations.cs that have non-nullable parameters are flagged when called with a possible null. The warning is on callsite so the warning appears in MarkStep even if nullable has not been enabled.

PR #2287, opened 2021-09-21
Share logic for discovering typereferences

sbomer:
Factor out the logic for walking an assembly to discover typerefs into a new abstract class, `TypeReferenceWalker`.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1381,7 +1381,7 @@ namespace Mono.Linker.Steps
 				MarkEntireType (type, new DependencyInfo (DependencyKind.TypeInAssembly, assembly));
 
 			// Mark scopes of type references and exported types.
-			new TypeReferenceMarker (assembly, MarkingHelpers).Process ();
+			TypeReferenceMarker.MarkTypeReferences (assembly, MarkingHelpers);
 		}
 
 		class TypeReferenceMarker : TypeReferenceWalker
@@ -1389,10 +1389,15 @@ namespace Mono.Linker.Steps
 
 			readonly MarkingHelpers markingHelpers;
 
-			public TypeReferenceMarker (AssemblyDefinition assembly, MarkingHelpers markingHelpers)
+			TypeReferenceMarker (AssemblyDefinition assembly, MarkingHelpers markingHelpers)
 				: base (assembly)
 			{
 				this.markingHelpers = markingHelpers;
+			}
+
+			public static void MarkTypeReferences (AssemblyDefinition assembly, MarkingHelpers markingHelpers)
+			{
+				new TypeReferenceMarker (assembly, markingHelpers).Process ();
 			}
 
 			protected override void ProcessTypeReference (TypeReference type)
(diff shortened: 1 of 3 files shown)


mateoatr (approved):
LGTM!

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1380,18 +1380,42 @@ void MarkEntireAssembly (AssemblyDefinition assembly)
...
 
-			// Also mark the scopes of metadata typeref rows to cover any not discovered by the traversal.
-			// This can happen when the compiler emits typerefs into IL which aren't strictly necessary per ECMA 335.
-			foreach (TypeReference typeReference in module.GetTypeReferences ())
-				MarkingHelpers.MarkForwardedScope (typeReference);
+			protected override void ProcessExtra ()
+			{
+				// Also mark the scopes of metadata typeref rows to cover any not discovered by the traversal.
+				// This can happen when the compiler emits typerefs into IL which aren't strictly necessary per ECMA 335.
+				foreach (TypeReference typeReference in assembly.MainModule.GetTypeReferences ()) {
```
vitek-karas, line 1413: Side question (ignore for PR purposes): Would be interesting to know what would happen if we removed such type ref.  If it's not used, then it should not matter that we don't preserve the type forwarder for it. I guess I don't know what Cecil will do when saving the assembly (will it rebuild the type-ref table from scratch, I guess it has to), but then again that will not happen for "copy".

src/linker/Linker/TypeReferenceWalker.cs
```diff
@@ -7,25 +7,26 @@
...
 {
-	struct TypeReferenceMarker
+	abstract class TypeReferenceWalker
 	{
-		readonly AssemblyDefinition assembly;
-		readonly MarkingHelpers markingHelpers;
-		HashSet<TypeReference> visited;
+		protected readonly AssemblyDefinition assembly;
+
+		protected HashSet<TypeReference>? visited;
```
vitek-karas, line 18: Maybe turn this into a get-only property which doesn't return null (would throw otherwise, as it should not get called outside of the Process method).
Generally I don't like using the `!` in non-weird cases.

vitek-karas: Actually - since the usage of this class is always:
```C#
var w = new TypeReferenceWalker(....);
w.Process();
// throw away w
```

Maybe change the class to not allow calling Process multiple times - the derived class would expose a static "ProcessSomething" which internally calls constructor and the Process method - and the state can be initialized in ctor and made readonly.

vitek-karas (approved):
Looks good. Looking forward to fixing the SweepStep to not call this twice :-)

src/linker/Linker.Steps/MarkStep.cs
sbomer: Good point - I think there would be no harm in not marking these, at least for the final output (which will just have unused dangling typerefs). There could be issues like https://github.com/dotnet/linker/issues/2267 since the in-memory representation would be unresolvable, but arguably custom steps should never be looking at these typerefs anyway.

src/linker/Linker/TypeReferenceWalker.cs
sbomer: Great ideas, thank you!

PR #2292, opened 2021-09-23
Analyzer test infra improvements

sbomer:
This makes the analyzer test validation more similar to the linker test validation and adds extra reporting when tests fail. The change caught a few differences between the linker and analyzer behavior:
- Analyzer was producing cctor warnings - so I removed that
- Analyzer was producing duplicate warnings for RUC properties set in attributes
- Analyzer behavior for reference to a testcase dependency assembly was already fixed

As part of this I also had to tweak the logic for checking whether a diagnostic originates from a particular member.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -96,14 +96,6 @@ namespace ILLink.RoslynAnalyzer
 				}, OperationKind.ObjectCreation);
 
 				context.RegisterOperationAction (operationContext => {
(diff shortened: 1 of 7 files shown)


sbomer:
@mateoatr once this is ready I plan to merge the changes from main into the DAM feature branch. These changes should make it easier to add coverage by sharing tests with the linker. PTAL when you have a chance!

mateoatr:
This LGTM. I think we are almost on-par with what we have under `VerifyLoggedMessages` in the linker's result checker! It would be good to try to start sharing code between that and the analyzer's test checker.

mateoatr (approved):

PR #2293, opened 2021-09-24
Minor simplification of the type ref walking code

vitek-karas:
Make the walker a true use-once class.

### DIFF
--- a/src/linker/Linker/TypeReferenceWalker.cs
+++ b/src/linker/Linker/TypeReferenceWalker.cs
@@ -16,7 +16,7 @@ namespace Mono.Linker
 	{
 		protected readonly AssemblyDefinition assembly;
 
-		protected HashSet<TypeReference> Visited { get; init; } = new HashSet<TypeReference> ();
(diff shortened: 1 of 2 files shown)


vitek-karas:
I was reading the PR once more ... and I just had to do this :wink:

src/linker/Linker/TypeReferenceWalker.cs
```diff
@@ -16,21 +16,20 @@ abstract class TypeReferenceWalker
 	{
 		protected readonly AssemblyDefinition assembly;
 
-		HashSet<TypeReference>? visited;
-		protected HashSet<TypeReference> Visited => visited ?? throw new InvalidOperationException ();
+		protected HashSet<TypeReference> Visited { get; init; } = new HashSet<TypeReference> ();
```
sbomer, line 19: Why does it need to be init?

sbomer (approved):
Nice, thank you!

src/linker/Linker/TypeReferenceWalker.cs
vitek-karas: It doesn't - it just "reads" right - I'll change it.

PR #2296, opened 2021-09-28
Make BindingFlagsAreUnsupported more precise

MichalStrehovsky:
`IgnorableBindingFlags` are all new - we wouldn't consider them supported in the past. I'm also adding `BindingFlags.ExactBinding` in the first group - that one was considered unsupported in the past but this flag just disallows implicit widening and similar things that don't affect our binding logic (I don't think we would ever want to go in that territory).


vitek-karas (approved):
Either in this PR or in a subsequent one we should add tests for at least some of these cases.

MichalStrehovsky:
Tweaked the existing tests since they were failing due to InvokeMethod being recognized and added a new test that covers an unrecognized flag.

PR #2297, opened 2021-09-28
Simplify DAM analyzer and add conversion test

sbomer:
@mateoatr I was looking again at the analyzer and saw some opportunity to simplify. I didn't understand why we check for `Conversion` operations specifically at https://github.com/dotnet/linker/blob/feature/damAnalyzer/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs#L125
so I wrote a test that should show a case where we might want different behavior for a conversion operator.

For now this pipes the `targetIsMethodReturn` argument through a few helpers, but soon we should be able to get rid of it and have a uniform representation of the source/target values.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
@@ -1,10 +1,11 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics;
 using ILLink.Shared;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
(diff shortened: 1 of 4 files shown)


src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -51,25 +52,52 @@ public override void Initialize (AnalysisContext context)
...
+			};
+
 			if (Annotations.SourceHasRequiredAnnotations (damtOnSource, damtOnTarget, out var missingAnnotations))
 				return;
 
-			var diag = GetDiagnosticId (source.Kind, target.Kind);
+			var diag = GetDiagnosticId (source.Kind, target.Kind, targetIsMethodReturn);
 			var diagArgs = GetDiagnosticArguments (source.Kind == SymbolKind.NamedType ? context.ContainingSymbol : source, target, missingAnnotations);
-			context.ReportDiagnostic (Diagnostic.Create (DiagnosticDescriptors.GetDiagnosticDescriptor (diag), assignmentOperation.Syntax.GetLocation (), diagArgs));
+			context.ReportDiagnostic (Diagnostic.Create (DiagnosticDescriptors.GetDiagnosticDescriptor (diag), originOperation.Syntax.GetLocation (), diagArgs));
```
agocke, line 92: Is `originOperation` only used for the location? If so, consider just passing along a location for the diagnostic.

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -678,6 +678,43 @@ private void M1()
 				.WithArguments ("type", "System.C.M2(System.Type)", "System.C.M1()", "'DynamicallyAccessedMemberTypes.PublicMethods'"));
 		}
 
+		[Fact]
+		public Task ConversionOperation ()
```
agocke, line 682: Shouldn't this be in the shared tests?

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -678,6 +678,43 @@ private void M1()
...
+		{
+			var TargetParameterWithAnnotations = @"
+namespace System
+{
+    class ConvertsToType
+    {
+        public static implicit operator Type(ConvertsToType value) => typeof (ConvertsToType);
```
agocke, line 689: Is there a test for putting `[DAM]` on the return type of the implicit conversion method?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
sbomer: Good idea, changed to take just a location.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -51,25 +52,52 @@ public override void Initialize (AnalysisContext context)
...
 				return;
 
-			if (TryGetSymbolFromOperation (assignmentOperation.Target) is not ISymbol target ||
-				TryGetSymbolFromOperation (assignmentOperation.Value) is not ISymbol source)
-				return;
+			CheckAndReportDynamicallyAccessedMembers (source, target, context, originOperation);
+		}
 
-			if (!target.TryGetDynamicallyAccessedMemberTypes (out var damtOnTarget))
+		static void CheckAndReportDynamicallyAccessedMembers (IOperation sourceOperation, ISymbol target, OperationAnalysisContext context, IOperation originOperation, bool targetIsMethodReturn = false)
```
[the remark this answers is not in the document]
sbomer: Fixed

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
sbomer: It's similar to the case where the return type isn't annotated at all but I added this test to make sure it behaves as expected.

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
sbomer: It should be, but I think there's still a little work left for us to confidently share tests, so I've put it here for now. I don't think it hurts to have a few analyzer-only tests as a sanity check, as long as we go back and make sure the shared tests have full coverage later.

mateoatr (approved):
Much cleaner/clearer! Thanks!

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -709,12 +709,52 @@ private void M1()
...
+				System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicMethods)] Type type)
+        {
+        }
+    }
+}";
+
+			return VerifyDynamicallyAccessedMembersAnalyzer (string.Concat (GetSystemTypeBase (), AnnotatedConversionOperation),
+				VerifyCS.Diagnostic (DiagnosticId.DynamicallyAccessedMembersMismatchMethodReturnTypeTargetsParameter)
+				.WithSpan (185, 16, 185, 36)
+				.WithArguments ("type", "System.C.M2(System.Type)", "System.ConvertsToType.implicit operator System.Type(System.ConvertsToType)", "'DynamicallyAccessedMemberTypes.PublicMethods'"));
```
agocke, line 755: Might be worth testing a success case as well.

agocke (approved):
LGTM

agocke:
If there's still outstanding work for the test integration it would be good to have an issue to track that

sbomer:
Tracking issue for test integration: https://github.com/dotnet/linker/issues/2299 (some of it moved out from https://github.com/dotnet/linker/issues/2273)

PR #2300, opened 2021-09-30
Nullable annotations part 4 (final)

sbomer:
Enable nullable annotations for the entire linker project. With this I think we will see fewer nullref exceptions in the future, though there might still be some due to Cecil being null-oblivious.

One pattern I used often was to turn fields which are initialized after the constructor into nullable fields with a non-nullable getter that throws if the underlying field is null. Let me know if there are any concerns with that approach - it's used all over MarkStep for example.

### DIFF
--- a/src/ILLink.Shared/DiagnosticString.cs
+++ b/src/ILLink.Shared/DiagnosticString.cs
@@ -1,4 +1,6 @@
-﻿namespace ILLink.Shared
+﻿using System;
+
+namespace ILLink.Shared
 {
 	public readonly struct DiagnosticString
 	{
@@ -9,23 +11,23 @@
 		{
 			var resourceManager = SharedStrings.ResourceManager;
 			_titleFormat = resourceManager.GetString ($"{diagnosticId}Title") ?? string.Empty;
-			_messageFormat = resourceManager.GetString ($"{diagnosticId}Message") ?? string.Empty;
+			_messageFormat = resourceManager.GetString ($"{diagnosticId}Message") ?? throw new ArgumentException ($"Diagnostic ID {nameof (diagnosticId)} has no message.");
 		}
 
 		public DiagnosticString (string diagnosticResourceStringName)
 		{
 			var resourceManager = SharedStrings.ResourceManager;
 			_titleFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Title") ?? string.Empty;
-			_messageFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Message") ?? string.Empty;
+			_messageFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Message") ?? throw new ArgumentException ($"Diagnostic ID {nameof (diagnosticResourceStringName)} has no message.");
 		}
 
 		public string GetMessage (params string[] args) =>
-			string.Format (_messageFormat ?? string.Empty, args);
+			string.Format (_messageFormat, args);
 
 		public string GetMessageFormat () => _messageFormat;
 
 		public string GetTitle (params string[] args) =>
-			string.Format (_titleFormat ?? string.Empty, args);
+			string.Format (_titleFormat, args);
 
 		public string GetTitleFormat () => _titleFormat;
 	}
--- a/src/ILLink.Shared/DiagnosticString.cs
+++ b/src/ILLink.Shared/DiagnosticString.cs
@@ -11,14 +11,14 @@ namespace ILLink.Shared
 		{
 			var resourceManager = SharedStrings.ResourceManager;
 			_titleFormat = resourceManager.GetString ($"{diagnosticId}Title") ?? string.Empty;
-			_messageFormat = resourceManager.GetString ($"{diagnosticId}Message") ?? throw new ArgumentException ($"Diagnostic ID {nameof (diagnosticId)} has no message.");
+			_messageFormat = resourceManager.GetString ($"{diagnosticId}Message") ?? string.Empty;
 		}
 
 		public DiagnosticString (string diagnosticResourceStringName)
 		{
 			var resourceManager = SharedStrings.ResourceManager;
 			_titleFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Title") ?? string.Empty;
-			_messageFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Message") ?? throw new ArgumentException ($"Diagnostic ID {nameof (diagnosticResourceStringName)} has no message.");
+			_messageFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Message") ?? string.Empty;
 		}
 
 		public string GetMessage (params string[] args) =>
--- a/src/ILLink.Shared/DiagnosticString.cs
+++ b/src/ILLink.Shared/DiagnosticString.cs
@@ -8,24 +8,24 @@
 		public DiagnosticString (DiagnosticId diagnosticId)
 		{
 			var resourceManager = SharedStrings.ResourceManager;
-			_titleFormat = resourceManager.GetString ($"{diagnosticId}Title");
-			_messageFormat = resourceManager.GetString ($"{diagnosticId}Message");
+			_titleFormat = resourceManager.GetString ($"{diagnosticId}Title") ?? string.Empty;
+			_messageFormat = resourceManager.GetString ($"{diagnosticId}Message") ?? string.Empty;
 		}
 
 		public DiagnosticString (string diagnosticResourceStringName)
 		{
 			var resourceManager = SharedStrings.ResourceManager;
-			_titleFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Title");
-			_messageFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Message");
+			_titleFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Title") ?? string.Empty;
+			_messageFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Message") ?? string.Empty;
 		}
 
 		public string GetMessage (params string[] args) =>
-			string.Format (_messageFormat, args);
+			string.Format (_messageFormat ?? string.Empty, args);
 
 		public string GetMessageFormat () => _messageFormat;
 
 		public string GetTitle (params string[] args) =>
-			string.Format (_titleFormat, args);
+			string.Format (_titleFormat ?? string.Empty, args);
 
 		public string GetTitleFormat () => _titleFormat;
 	}
--- a/src/linker/Linker.Steps/BaseStep.cs
+++ b/src/linker/Linker.Steps/BaseStep.cs
@@ -27,6 +27,7 @@
 //
 
 using System;
+using System.Diagnostics;
 using Mono.Cecil;
 
 namespace Mono.Linker.Steps
@@ -37,7 +38,12 @@ namespace Mono.Linker.Steps
 
 		private LinkContext? _context;
 
-		public LinkContext Context => _context ?? throw new InvalidOperationException ();
+		public LinkContext Context {
+			get {
+				Debug.Assert (_context != null);
+				return _context;
+			}
+		}
 
 		public AnnotationStore Annotations {
 			get { return Context.Annotations; }
--- a/src/linker/Linker.Steps/BaseStep.cs
+++ b/src/linker/Linker.Steps/BaseStep.cs
@@ -35,21 +35,19 @@ namespace Mono.Linker.Steps
 	public abstract class BaseStep : IStep
 	{
 
-		private LinkContext _context;
+		private LinkContext? _context;
 
-		public LinkContext Context {
-			get { return _context; }
-		}
+		public LinkContext Context => _context ?? throw new InvalidOperationException ();
 
 		public AnnotationStore Annotations {
-			get { return _context.Annotations; }
+			get { return Context.Annotations; }
 		}
 
 		public Tracer Tracer {
-			get { return _context.Tracer; }
+			get { return Context.Tracer; }
 		}
 
-		public MarkingHelpers MarkingHelpers => _context.MarkingHelpers;
+		public MarkingHelpers MarkingHelpers => Context.MarkingHelpers;
 
 		public void Process (LinkContext context)
 		{
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2164,13 +2164,14 @@ namespace Mono.Linker.Steps
 						}
 					}
 
-					TypeDefinition? nType = type;
-					while (nType != null) {
+					while (true) {
 						// Currently if we don't understand the DebuggerDisplayAttribute we mark everything on the type
 						// This can be improved: dotnet/linker/issues/1873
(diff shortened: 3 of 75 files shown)


src/linker/Linker/DocumentationSignatureGenerator.PartVisitor.cs
```diff
@@ -147,7 +147,7 @@ public void VisitTypeReference (TypeReference typeReference, StringBuilder build
 				}
 
 				if (typeReference.IsNested) {
-					VisitTypeReference (typeReference.GetInflatedDeclaringType (resolver), builder, resolver);
+					VisitTypeReference (typeReference.GetInflatedDeclaringType (resolver)!, builder, resolver);
```
sbomer, line 150: Calling out this use of the null-forgiving operator - I'd appreciate more eyes on this one.

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
```diff
@@ -1403,6 +1412,8 @@ public bool Analyze ()
 					case Code.Nop:
 						continue;
 					case Code.Pop:
+						if (stack_instr == null)
+							Debug.Fail ("Invalid IL?");
 						stack_instr.Pop ();
```
mateoatr, line 1417: Should we catch the `InvalidOperationException` here and wrap it with a linker exception instead (and add the "Invalid IL message")?

src/linker/Linker/MarkingHelpers.cs
```diff
@@ -35,7 +35,9 @@ public void MarkForwardedScope (TypeReference typeReference)
 
 			if (typeReference.Scope is AssemblyNameReference) {
 				var assembly = _context.Resolve (typeReference.Scope);
-				if (assembly != null && assembly.MainModule.GetMatchingExportedType (_context.TryResolve (typeReference), out var exportedType))
+				if (assembly != null &&
+					_context.TryResolve (typeReference) is TypeDefinition typeDefinition &&
```
mateoatr, line 39: I think we already check for nullability in `GetMatchingExportedType`.

src/linker/Linker/MethodBodyScanner.cs
```diff
@@ -113,7 +114,8 @@ HashSet<TypeDefinition> AllPossibleStackTypes (MethodDefinition method)
 
 			foreach (Instruction instruction in body.Instructions) {
 				if (instruction.Operand is FieldReference fieldReference) {
-					AddIfResolved (types, context.TryResolve (fieldReference)?.FieldType);
+					if (context.TryResolve (fieldReference)?.FieldType is TypeReference fieldType)
+						AddIfResolved (types, fieldType);
```
mateoatr, line 118: Shouldn't we check that this is resolved in `AddIfResolved`? Otherwise we are resolving here just to resolve once more in `AddIfResolved` (and the name suggest the resolution should be made inside that method).

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -61,10 +61,12 @@ public static StringBuilder GetDisplayNameWithoutNamespace (this TypeReference t
 					break;
 				}
 
-				type = type.DeclaringType;
-				if (type == null)
+				var nType = type.DeclaringType;
+				if (nType == null)
```
mateoatr, line 65: Nit: why is this needed?

src/linker/Linker/XmlDependencyRecorder.cs
```diff
@@ -173,7 +179,7 @@ bool WillAssemblyBeModified (AssemblyDefinition assembly)
 			}
 		}
 
-		bool ShouldRecord (object o)
+		bool ShouldRecord (object? o)
```
mateoatr, line 182: Would it make sense to return false if `o` is null?

sbomer: It might, but `TokenString` supports logging edges with null source and I'd rather not change that here.

src/linker/Linker/TypeReferenceExtensions.cs
sbomer: I needed a separate variable that is nullable since the type is not. I've cleaned this up a bit.

src/linker/Linker/MethodBodyScanner.cs
sbomer: There are two steps - resolving the field itself, then the field's type - and the latter is already done inside of `AddIfResolved`.

src/linker/Linker/MarkingHelpers.cs
sbomer: I somewhat prefer to annotate the input parameter to `GetMatchingExportedType` as non-nullable, and require callers to decide what to do if a type doesn't resolve - LMK if you prefer making it nullable instead.

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
sbomer: I think we already do in https://github.com/dotnet/linker/blob/640878adf1e27cb79bea0a894033f85a84db208d/src/linker/Linker.Steps/MarkStep.cs#L515-L517. Is that what you had in mind?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1004,17 +1008,19 @@ protected virtual void MarkUserDependency (IMemberDefinition context, CustomAttr
...
 
-			if (MarkNamedField (td, member, new DependencyInfo (DependencyKind.PreservedDependency, ca)))
-				return;
+				if (MarkNamedField (td, member, new DependencyInfo (DependencyKind.PreservedDependency, ca)))
+					return;
+			}
 
-			_context.LogWarning (
+			Context.LogWarning (
 				$"Could not resolve dependency member '{member}' declared in type '{td.GetDisplayName ()}' specified in a 'PreserveDependency' attribute.", 2005, context);
```
tlakollo, line 1020: We will print empty here in the case member is null, not sure if it's important or if we need a specific error code for that

src/linker/Linker/DocumentationSignatureGenerator.PartVisitor.cs
tlakollo: At the end of GetInflatedDeclaringType method we never check if the resolved is null, the declaring type would be null if the resolved type is null
```C#
var resolved = resolver.TryResolve (type);
System.Diagnostics.Debug.Assert (resolved == type);
return resolved?.DeclaringType;
```
If resolved cannot be null then the return statement needs a null forgiving operator `return resolved!.DeclaringType;`
I'm not sure what is a PinnedType or a RequiredModifierType which I guess is the part that needs more attention though

src/linker/Linker/Driver.cs
```diff
@@ -864,12 +866,12 @@ protected bool AddMarkHandler (Pipeline pipeline, string arg)
 			return true;
 		}
 
-		bool TryGetCustomAssembly (ref string arg, out Assembly assembly)
+		bool TryGetCustomAssembly (ref string arg, [NotNullWhen (true)] out Assembly? assembly)
 		{
 			assembly = null;
 			int pos = arg.IndexOf (",");
 			if (pos == -1)
-				return true;
```
tlakollo, line 872: I guess I just want to understand this one better, what is a custom assembly for driver purposes and what is the format that the coma seems to be important

src/linker/Linker/MarkingHelpers.cs
mateoatr: > I somewhat prefer to annotate the input parameter to `GetMatchingExportedType` as non-nullable, and require callers to decide what to do if a type doesn't resolve

Makes sense. We can keep it this way :)

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
mateoatr: If `stack_instr == null` then `stack_instr.Pop()` will throw `InvalidOperationException`. I was thinking about throwing that with the "invalid IL?" message so that it's displayed to user. I don't mind keeping it this way since I wouldn't expect this to be hit much and either way the stack trace would give a good clue that there's something weird with the IL.

src/linker/Linker/MethodReferenceExtensions.cs
```diff
@@ -56,15 +56,15 @@ public static string GetDisplayName (this MethodReference method)
 		public static TypeReference GetReturnType (this MethodReference method, LinkContext context)
 		{
 			if (method.DeclaringType is GenericInstanceType genericInstance)
-				return TypeReferenceExtensions.InflateGenericType (genericInstance, method.ReturnType, context);
+				return TypeReferenceExtensions.InflateGenericType (genericInstance, method.ReturnType, context)!;
```
vitek-karas, line 59: Is there a reason `InflateGenericType` can return null? (it seems to handle null input, but that can be changed ... and it would make sense)

vitek-karas (approved):
Other than existing comments, this looks good!

src/linker/Linker/MethodReferenceExtensions.cs
tlakollo: The only way I see it not working is when is not able to resolve the GenericInstanceProvider.ElementType, does a GenericInstanceType always have a ElementType?

src/linker/Linker/TypeNameResolver.cs
```diff
@@ -98,7 +102,7 @@ TypeReference ResolveTypeName (AssemblyDefinition assembly, TypeName typeName)
 				if (genericTypeRef == null)
 					return null;
 
-				TypeDefinition genericType = _context.TryResolve (genericTypeRef);
+				TypeDefinition? genericType = _context.TryResolve (genericTypeRef);
```
tlakollo, line 105: Should we check for null here?

tlakollo (approved):

src/linker/Linker/TypeNameResolver.cs
sbomer: Good catch - this is the kind of case that unfortunately won't be caught by the analysis. Looking at it I think the `TryResolve` call is actually unnecessary since resolving the generic type above should always return a TypeDefinition, so I replaced it with an assert. PTAL @tlakollo and @mateoatr

src/linker/Linker.Steps/MarkStep.cs
sbomer: I think it's ok since this is the same thing that would be printed if the member signature were an empty string, and because PreserveDependency is deprecated so this entire code path is legacy anyway.

src/linker/Linker/DocumentationSignatureGenerator.PartVisitor.cs
sbomer: Yeah, I think when we get to that line it will always be a TypeDefinition. I changed the code a bit to make this more clear. Maybe there is a way to get here for sentinel types (I believe they're used to encode vararg signatures) but this likely isn't the only thing that doesn't work in that case.

src/linker/Linker.Steps/BaseStep.cs
```diff
@@ -35,21 +35,19 @@ namespace Mono.Linker.Steps
 	public abstract class BaseStep : IStep
 	{
 
-		private LinkContext _context;
+		private LinkContext? _context;
 
-		public LinkContext Context {
-			get { return _context; }
-		}
+		public LinkContext Context => _context ?? throw new InvalidOperationException ();
```
marek-safar, line 40: I don't like this patter where your turn a simple optimizable property into something that is complex check for unnecessary condition.

sbomer: Is the concern about the performance impact of the extra check? One alternative would be to return `_context!` and leave this as a nullref exception if accessed incorrectly. From an API design standpoint I'd rather throw `InvalidOperationException` if the performance is acceptable. Let me know what you prefer or if you have any other suggestions!

src/ILLink.Shared/DiagnosticString.cs
```diff
@@ -8,24 +8,24 @@
 		public DiagnosticString (DiagnosticId diagnosticId)
 		{
 			var resourceManager = SharedStrings.ResourceManager;
-			_titleFormat = resourceManager.GetString ($"{diagnosticId}Title");
-			_messageFormat = resourceManager.GetString ($"{diagnosticId}Message");
+			_titleFormat = resourceManager.GetString ($"{diagnosticId}Title") ?? string.Empty;
```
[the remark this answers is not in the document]
mateoatr: Note that not all message strings have a title counterpart, take for example `BaseRequiresMismatchMessage` (there's no BRMTitle string).

src/linker/Linker/XmlDependencyRecorder.cs
```diff
@@ -83,22 +83,28 @@ public void Dispose ()
...
-			stream.Dispose ();
+			stream?.Dispose ();
 			writer = null;
 			stream = null;
 		}
 
 		public void RecordDependency (object target, in DependencyInfo reason, bool marked)
 		{
+			if (writer == null)
+				throw new InvalidOperationException ();
```
agocke, line 94: I think `ArgumentNullException` is appropriate here.

src/linker/Linker/XmlDependencyRecorder.cs
```diff
@@ -83,22 +83,28 @@ public void Dispose ()
...
 
 			// For now, just report a dependency from source to target without noting the DependencyKind.
 			RecordDependency (reason.Source, target, marked);
 		}
 
-		public void RecordDependency (object source, object target, bool marked)
+		public void RecordDependency (object? source, object target, bool marked)
 		{
+			if (writer == null)
+				throw new InvalidOperationException ();
```
agocke, line 106: Same here.

src/linker/Linker/XmlDependencyRecorder.cs
sbomer: `writer` isn't an argument - I'd like the exception to indicate that this method was called while the object wasn't in an appropriate state and I think InvalidOperationException is right for that.

src/ILLink.Shared/DiagnosticString.cs
```diff
@@ -8,24 +8,24 @@
...
 			var resourceManager = SharedStrings.ResourceManager;
-			_titleFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Title");
-			_messageFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Message");
+			_titleFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Title") ?? string.Empty;
+			_messageFormat = resourceManager.GetString ($"{diagnosticResourceStringName}Message") ?? string.Empty;
 		}
 
 		public string GetMessage (params string[] args) =>
-			string.Format (_messageFormat, args);
+			string.Format (_messageFormat ?? string.Empty, args);
```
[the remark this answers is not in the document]
sbomer: I think it was left-over from an earlier version, thanks for catching it.

src/ILLink.Shared/DiagnosticString.cs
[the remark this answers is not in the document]
sbomer: Never mind - looks like not all diagnostics have fixed messages, because we use different strings depending on the context - see here:
https://github.com/dotnet/linker/blob/3418bcd1d864ce7f62d789c8961f9c2d750fb079/src/ILLink.Shared/MessageFormat.cs#L23-L33

PR #2306, opened 2021-10-04
Add COMAnalyzer 

mateoatr:
Rebased #2275 on top of main. ILLinkAnalyzer test checker was validating `ExpectedWarningAttribute`s for any diagnostic id -- I made changes in this PR so that we only validate diagnostic ids supported by the analyzer we are testing:
https://github.com/mateoatr/linker/blob/a540731187a7a0f1a7beb3d648d79003c589e4a6/test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs#L34-L39

### DIFF
--- a/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
@@ -30,7 +30,7 @@ namespace ILLink.RoslynAnalyzer.Tests
 
 		[Theory]
 		[MemberData (nameof (TestCaseUtils.GetTestData), parameters: nameof (Interop))]
-		public void Interop (MethodDeclarationSyntax m, List<AttributeSyntax> attrs)
+		public void Interop (string testName, MethodDeclarationSyntax m, List<AttributeSyntax> attrs)
 		{
 			RunTest<COMAnalyzer> (m, attrs, UseMSBuildProperties (MSBuildPropertyOptionNames.EnableTrimAnalyzer));
 		}
--- a/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
@@ -27,5 +27,12 @@ namespace ILLink.RoslynAnalyzer.Tests
 
 			RunTest<RequiresUnreferencedCodeAnalyzer> (m, attrs, UseMSBuildProperties (MSBuildPropertyOptionNames.EnableTrimAnalyzer));
 		}
+
+		[Theory]
+		[MemberData (nameof (TestCaseUtils.GetTestData), parameters: nameof (Interop))]
+		public void Interop (MethodDeclarationSyntax m, List<AttributeSyntax> attrs)
+		{
+			RunTest<COMAnalyzer> (m, attrs, UseMSBuildProperties (MSBuildPropertyOptionNames.EnableTrimAnalyzer));
+		}
 	}
 }
--- a/test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
@@ -25,10 +25,16 @@ namespace ILLink.RoslynAnalyzer.Tests
 
 		private readonly SyntaxNode MemberSyntax;
 
+		private readonly HashSet<string> TestingAnalyzers; 
+
 		public TestChecker (MemberDeclarationSyntax memberSyntax, (CompilationWithAnalyzers Compilation, SemanticModel SemanticModel) compilationResult)
 		{
 			Compilation = compilationResult.Compilation;
 			SemanticModel = compilationResult.SemanticModel;
+			TestingAnalyzers = new HashSet<string> ();
+			foreach (var analyzer in Compilation.Analyzers)
+				TestingAnalyzers.Add (analyzer.GetType ().Name);
+
 			DiagnosticMessages = Compilation.GetAnalyzerDiagnosticsAsync ().Result
 				.Where (d => {
 					// Filter down to diagnostics which originate from this member.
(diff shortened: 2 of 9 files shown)


test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
```diff
@@ -137,6 +151,11 @@ private bool TryValidateExpectedWarningAttribute (AttributeSyntax attribute, Lis
 			if (!expectedWarningCode.StartsWith ("IL"))
 				throw new InvalidOperationException ($"Expected warning code should start with \"IL\" prefix.");
 
+			if (!SupportedDiagnosticIds.Contains (expectedWarningCode)) {
+				unsupportedDiagnostic = true;
+				return false;
```
sbomer, line 156: I think this approach makes it too easy to miss potential coverage holes (since it'll let tests pass if we forget to implement a particular warning ID or if there are interactions between the analyzers). I wonder if eventually we'll want to run these tests with all trim-related Roslyn analyzers. For now, could we add a new enum value to ProducedBy and use that instead for the warnings not produced by the COM analyzer?

mateoatr: Good point. I do think that at some point we'll want to run tests with multiple/all Roslyn analyzers.
> For now, could we add a new enum value to ProducedBy and use that instead for the warnings not produced by the COM analyzer?

Sure thing. I think it would make sense to have each of the analyzers in that enum.

test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
```diff
@@ -30,7 +30,7 @@ public void RequiresCapability (string testName, MemberDeclarationSyntax m, List
 
 		[Theory]
 		[MemberData (nameof (TestCaseUtils.GetTestData), parameters: nameof (Interop))]
-		public void Interop (MethodDeclarationSyntax m, List<AttributeSyntax> attrs)
+		public void Interop (string testName, MethodDeclarationSyntax m, List<AttributeSyntax> attrs)
```
tlakollo, line 33: I think the testName is never used here

test/Mono.Linker.Tests.Cases/Interop/PInvoke/Warnings/ComPInvokeWarning.cs
```diff
@@ -175,7 +175,7 @@ static void Call_CanSuppressPInvokeWithRequiresUnreferencedCode ()
 		static extern void CanSuppressPInvokeWithRequiresUnreferencedCode (IFoo foo);
 
 		[ExpectedWarning ("IL2050")]
-		[ExpectedWarning ("IL2026")]
+		[ExpectedWarning ("IL2026", ProducedBy = ProducedBy.Trimmer | ProducedBy.RequiresUnreferencedCodeAnalyzer)]
 		static void Call_PInvokeWithRequiresUnreferencedCode ()
```
tlakollo, line 179: Should we have this test in another file that tests using all analyzers? since is an interaction between them

mateoatr: I think we want to move there little by little (have all/most tests in Mono.Linker.Tests be run using all analyzers or a combination of them via the `ProducedBy` property). We can run tests under `Interop` with the `Requires_` analyzers, but I'd prefer doing that in another PR.

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
```diff
@@ -67,18 +73,39 @@ bool IsExpectedDiagnostic (AttributeSyntax attribute)
...
 				var args = TestCaseUtils.GetAttributeArguments (attribute);
-				if (args.TryGetValue ("ProducedBy", out var producedBy) &&
-					producedBy is MemberAccessExpressionSyntax memberAccessExpression &&
-					memberAccessExpression.Name is IdentifierNameSyntax identifierNameSyntax &&
-					identifierNameSyntax.Identifier.ValueText == "Trimmer")
-					return false;
+				if (args.TryGetValue ("ProducedBy", out var producedBy)) {
+					var producedByValues = GetProducedByValues (producedBy);
+					// Skip if this warning is not expected to be produced by any of the analyzers that we are currently testing.
+					return !producedByValues.Contains ("Analyzer") && (producedByValues.Intersect (TestingAnalyzers) == null);
```
sbomer, line 79: The condition looks inverted - I think it is skipping `ProducedBy.Analyzer` for example. I also am not sure an empty intersection will be null.

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
```diff
@@ -67,18 +73,39 @@ bool IsExpectedDiagnostic (AttributeSyntax attribute)
...
+
 				return true;
 			case "LogContains":
 			case "UnrecognizedReflectionAccessPattern":
 				return true;
 			default:
 				return false;
 			}
+
+			static List<string> GetProducedByValues (ExpressionSyntax expression)
```
sbomer, line 90: I wonder - should we make this return a `ProducedBy` value? Yes, we would have to map it back into strings to compare with the set of active analyzers - but I would find it a little easier to reason about since it would collapse arguments like `ProducedBy.RUCAnalyzer | ProducedBy.Analyzer` into `ProducedBy.Analyzer`. Also ideally we would handle `ProducedBy.TrimmerAndAnalyzer` if passed explicitly.

test/Mono.Linker.Tests.Cases.Expectations/Assertions/ProducedBy.cs
```diff
@@ -9,7 +9,10 @@ namespace Mono.Linker.Tests.Cases.Expectations.Assertions
 	public enum ProducedBy
 	{
 		Trimmer = 1,
-		Analyzer = 2,
+		RequiresAssemblyFileAnalyzer = 2,
```
agocke, line 12: This seems like it's leaking implementation details. We shouldn't need to know which analyzer produces it, only if it's a diagnostic currently produced by any of the analyzers.

mateoatr: But then that would be a problem whenever some test suite in Mono.Linker.Tests.Cases has `ExpectedWarning` with warning codes for different analyzers, no? e.g., say you have A.cs with:

```C#
[ExpectedWarning("ILXXXX", ProducedBy = ProducedBy.Analyzer)] // This is only produced by analyzer A
[ExpectedWarning("ILYYYY", ProducedBy = ProducedBy.Analyzer)] // This is only produced by analyzer B
void Foo () { ... }
```

Since analyzer tests are run separately, when running the tests in A.cs with analyzer A, we'd fail since it wouldn't produce ILYYYY; similar when running A.cs tests with B, in which case we'd be missing ILXXXX.

mateoatr: We could also allow to pass multiple analyzers to `RunTest` in the analyzer test infra, if that makes more sense than having all analyzers in the enum.

sbomer: Yeah, it is leaking details a bit due to the impedance mismatch between the linker/analyzer test infrastructure - running with all the analyzers would be good to avoid this. But I like this approach as a way to ignore some expected warnings when testing with only one analyzer for now. Maybe it's worth a comment here to say why we have the individual enum values.

PR #2307, opened 2021-10-05
[feature/damAnalyzer] Enable a few linker DAM tests for analyzer

sbomer:
This turns on a few of the basic testcases for DAM analysis that don't require dataflow. I changed the signature format slightly because Roslyn formats methods with spaces separating parameters by default. I also changed the field signature format to use a format consistent with methods.


test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -1018,5 +1018,82 @@ private static void M2<S>()
...
+
+    [return: DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
+    private static Type M()
+    {
+        return typeof(C);
+    }
+}";
+
+			// TODO: https://github.com/dotnet/linker/issues/2308
+			// This should produce no warnings. Currently it produces the following:
```
mateoatr, line 1048: Should we have this test verify the expected result (i.e., no warnings produced) and use `[Fact(Skip = "This should...")]` instead?

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -1018,5 +1018,82 @@ private static void M2<S>()
...
+        M(typeof(C));
+    }
+
+    private static void M([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type type)
+    {
+    }
+}";
+
+			// TODO: https://github.com/dotnet/linker/issues/2308
+			// This should produce no warnings. Currently it produces the following:
```
mateoatr, line 1085: Same as above.

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -119,6 +120,13 @@ public static void GetDirectoryPaths (out string rootSourceDirectory, out string
 			testAssemblyPath = Path.GetFullPath (Path.Combine (artifactsBinDir, "ILLink.RoslynAnalyzer.Tests", configDirectoryName, tfm));
 		}
 
+		// Accepts typeof expressions, with a format specifier
+		public static string GetStringFromExpression (TypeOfExpressionSyntax expr, SemanticModel semanticModel, SymbolDisplayFormat displayFormat)
+		{
+			var typeSymbol = semanticModel.GetSymbolInfo (expr.Type).Symbol;
+			return typeSymbol?.ToDisplayString (displayFormat) ?? throw new InvalidOperationException ();
+		}
```
mateoatr, line 128: Should we add `displayFormat` as an optional parameter to the bellow method and put these two lines there?

sbomer: I made them separate because I don't think there's an obvious way to give a string representation of an expression in general. I'd rather limit the below method to expressions which are building up strings from literals, and have separate variants used to format types (or maybe other kinds in the future).

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
sbomer: I was going back and for on this - done!

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
```diff
@@ -70,22 +70,49 @@ internal static bool TryGetOverriddenMember (this ISymbol? symbol, [NotNullWhen
...
+			new SymbolDisplayFormat (
+				typeQualificationStyle: SymbolDisplayTypeQualificationStyle.NameOnly,
+				genericsOptions: SymbolDisplayGenericsOptions.IncludeTypeParameters,
+				memberOptions:
+					SymbolDisplayMemberOptions.IncludeParameters |
+					SymbolDisplayMemberOptions.IncludeExplicitInterface,
+				parameterOptions: SymbolDisplayParameterOptions.IncludeType
+			);
+
 		public static string GetDisplayName (this ISymbol symbol)
```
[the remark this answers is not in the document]
sbomer: I'd love to share more code and tests! I gave it some thought but don't see a great way to share more of the display name code (or even interfaces). Even if we could implement interfaces for foreign types we'd need to specialize the implementation for different `IMetadataTokenProvirder`s since the interface doesn't provide enough information to format a display name. Ideas welcome!

PR #2317, opened 2021-10-14
[release/6.0] Fix tlens Mono.Cecil reference

lbussell:
This change helps resolve Mono.Cecil in dotnet source-build and also makes the Mono.Cecil reference in tlens the same as the rest of the linker sub-projects.

### DIFF
--- a/src/tlens/tlens.csproj
+++ b/src/tlens/tlens.csproj
@@ -10,7 +10,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Mono.Cecil" Version="$(MonoCecilVersion)" />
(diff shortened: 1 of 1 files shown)


lbussell:
> Why for release/6.0 and not main?

We would like this change to flow into the installer/6.0.1xx branch for 6.0 GA, otherwise we will carry a [source-build patch](https://github.com/dotnet/installer/pull/12372) for this change in 6.0. Is main the better option for this? or I should also submit a separate PR for main as well?

akoeplinger:
Do we still need this in 6.0?

lbussell:
This is no longer needed. We are using this as a patch in source-build so it would cause dependency flow conflicts in dotnet/installer if we merged this now - not worth the risk.

PR #2327, opened 2021-10-20
Use all diagnostic analyzers when running analyzer's tests

### LINKED ISSUE #2315: Run trimmer tests in ILLink.RoslynAnalyzer.Tests with all analyzers
The ILLink.RoslynAnalyzer.Tests infra only supports running trimmer tests with a single analyzer at a time. This forces us to hand-pick which are the expected warnings so that we skip a warning A whenever an analyzer B which does not produce A is running. We could have a more general approach and avoid over-specifying values in the `ProducedBy` property of the `ExpectedWarningAttribute` put on tests if the analyzers tests infra had support for running a single trimmer test with all analyzers at once.

mateoatr:
Fixes #2315. See https://github.com/dotnet/linker/pull/2306#discussion_r724483985. This removes the enumerated analyzers from `ProducedBy` and sets the analyzer's tests to use all diagnostic analyzers for each run (instead of running each test with each of the analyzers, which we currently do).

### DIFF
--- a/test/ILLink.RoslynAnalyzer.Tests/TestCaseCompilation.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/TestCaseCompilation.cs
@@ -54,7 +54,6 @@ namespace ILLink.RoslynAnalyzer.Tests
 				concurrentAnalysis: true,
 				logAnalyzerExecutionTime: false);
 
-			// Aquí mero.
 			return (new CompilationWithAnalyzers (comp, SupportedDiagnosticAnalyzers, compWithAnalyzerOptions), comp.GetSemanticModel (src));
 		}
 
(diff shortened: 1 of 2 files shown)


test/ILLink.RoslynAnalyzer.Tests/TestCaseCompilation.cs
```diff
@@ -0,0 +1,100 @@
...
+				ImmutableArray<AdditionalText>.Empty,
+				new SimpleAnalyzerOptions (globalAnalyzerOptions));
+
+			var compWithAnalyzerOptions = new CompilationWithAnalyzersOptions (
+				analyzerOptions,
+				(_1, _2, _3) => { },
+				concurrentAnalysis: true,
+				logAnalyzerExecutionTime: false);
+
+			// Aquí mero.
```
tlakollo, line 57: Ahí mero indeed :D

mateoatr: Oops! Haha thanks. (Translation: "here here").

tlakollo: Right here
Right there indeed

PR #2334, opened 2021-10-26
Allow .NET style nested type name

mrvoorhe:
A request came in awhile back to allow the .NET style nested type name instead of cecil's nested type naming scheme.

This seems like a sensible request to me.  Someone may copy a fully qualified type name out of a debugger, or print it to stdout at runtime and then paste it into a linker xml file.  To find out that name doesn't work and you have to replace the +'s with /'s would be a negative experience for users.

### DIFF
--- a/src/linker/Linker.Steps/ProcessLinkerXmlBase.cs
+++ b/src/linker/Linker.Steps/ProcessLinkerXmlBase.cs
@@ -168,6 +168,10 @@ namespace Mono.Linker.Steps
 
 				TypeDefinition type = assembly.MainModule.GetType (fullname);
 
+				// Before giving up, check .NET style nested type naming
+				if (type == null)
+					type = assembly.MainModule.GetType (fullname.Replace ('+', '/'));
+
 				if (type == null && assembly.MainModule.HasExportedTypes) {
 					foreach (var exported in assembly.MainModule.ExportedTypes) {
 						if (fullname == exported.FullName) {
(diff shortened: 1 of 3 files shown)


mrvoorhe:
@marek-safar Does this PR seem OK?

src/linker/Linker.Steps/ProcessLinkerXmlBase.cs
```diff
@@ -168,6 +168,10 @@ protected virtual void ProcessTypes (AssemblyDefinition assembly, XPathNavigator
 
 				TypeDefinition type = assembly.MainModule.GetType (fullname);
 
+				// Before giving up, check .NET style nested type naming
```
marek-safar, line 171: I'm not convinced this is a useful change. Why support it only for one of the possible XML inputs ?

MichalStrehovsky: `<type fullname="Namespace.Type/Nested*" />` handled in `ProcessTypePattern` above or `<method signature="SomeType/NestedType SomeMethod()" />`.

mrvoorhe: Good point.  Didn't think about those.

It seems like the same approach could be used in those places as well.

I'm not wed to this idea.  Someone internally mentioned it to me.  I think what happened was

1) They copied a nested type name from somewhere where it was in the .net style and pasted it into a link.xml file.
2) Linker issues a warning about no type found
3) Unity doesn't report any linker warnings.  So the warning goes unnoticed.
4) The person had to troubleshoot and figure out why their preservation had no effect.  I'm sure this was time consuming and confusing.

Given that (3) makes this scenario considerably more frustrating.  And (3) is on us for poor integration.  I can see how this scenario may be less problematic to the larger .net community.

If I implemented support in the other places would you want this change?  If not.  That's OK.

mrvoorhe:
I'm going to close this PR.  It doesn't seem like this change is desired upstream and we don't feel strongly about it either.

PR #2335, opened 2021-10-26
Merge main back to release/6.0.2xx

agocke:


marek-safar:
That's not what I'm seeing in the branch. For example, you pushed non-LTS 7.0 arcade changes to LTS 6.0 branch which is one of the problems with doing such merges and breaking the support policy.

PR #2336, opened 2021-10-27
Test Restructure

tlakollo:
Add RequiresAssemblyFiles to RequiresCapability file
Add ProducedBy for LogContains and LogDoesNotContain
Remove Analyzer Generic for test case run, since now all analyzers are run


test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresCapability.cs
```diff
@@ -809,18 +897,279 @@ class AccessThroughLdToken
...
+				private string name;
+				public override string VirtualPropertyAnnotationInAccesor {
+					[ExpectedWarning ("IL2046", "DerivedClassWithoutRequires.VirtualPropertyAnnotationInAccesor.get", "BaseClassWithRequires.VirtualPropertyAnnotationInAccesor.get")]
+					[ExpectedWarning ("IL3003", "DerivedClassWithoutRequires.VirtualPropertyAnnotationInAccesor.get", "BaseClassWithRequires.VirtualPropertyAnnotationInAccesor.get", ProducedBy = ProducedBy.Analyzer)]
+					get { return name; }
+					set { name = value; }
+				}
+
+				[ExpectedWarning ("IL3003", "DerivedClassWithoutRequires.VirtualPropertyAnnotationInProperty", "BaseClassWithRequires.VirtualPropertyAnnotationInProperty", ProducedBy = ProducedBy.Analyzer)]
+				public override string VirtualPropertyAnnotationInProperty { get; set; }
```
vitek-karas, line 992: Why is this for assembly files only and not for unreferenced code as well?

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
```diff
@@ -314,13 +387,25 @@ static async IAsyncEnumerable<int> TestReflectionAccess ()
...
 			[ExpectedWarning ("IL2026", "--MethodWithRequires--", CompilerGeneratedCode = true)]
+			[ExpectedWarning ("IL3002", "--MethodWithRequires--", ProducedBy = ProducedBy.Analyzer)]
 			static async IAsyncEnumerable<int> TestLdftn ()
 			{
 				await MethodAsync ();
 				yield return 0;
 				var action = new Action (MethodWithRequires);
 			}
 
+			[ExpectedWarning ("IL2026", "Message from --InitString--", CompilerGeneratedCode = true)]
```
vitek-karas, line 398: I don't think these should have the `CompilerGeneratedCode = true` on them... that is meant for cases where the origin is from the compiler generated code (and thus it's not possible to put the attribute there). In this case the warning happens on a static field directly - so no need for this.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
```diff
@@ -1052,18 +1276,27 @@ static async Task<int> MethodAsync ()
...
 			[RequiresUnreferencedCode ("--TypeWithMethodWithRequires.MethodWithRequires--")]
+			[RequiresAssemblyFiles ("--TypeWithMethodWithRequires.MethodWithRequires--")]
 			static void MethodWithRequires ()
 			{
 			}
 		}
 
+		[RequiresUnreferencedCode ("Message from --InitString--")]
+		[RequiresAssemblyFiles ("Message from --InitString--")]
+		public static string InitString ()
```
vitek-karas, line 1295: Can you please give this a better name? Something like `MethodWithRequiresAndReturns` or similar.

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresCapability.cs
tlakollo: You cannot place RUC on a property, but you can do it in RAF

vitek-karas: Hmm that's a curious limitation... can you please create an issue somewhere to make this consistent? Unless there's a reason (and I just forgot).

tlakollo: What about events?

vitek-karas: I think it's the same deal... In theory there's no reason for these two attributes to have different behavior (at least I'm not aware of one). It's fine if they're not identical right now, but I think they should be eventually.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
tlakollo: According to linker the warning comes from the static constructor, for example:
`ILLink: Trim analysis warning IL2026: Mono.Linker.Tests.Cases.RequiresCapability.RequiresInCompilerGeneratedCode.SuppressInAsyncBody..cctor(): Using member 'Mono.Linker.Tests.Cases.RequiresCapability.RequiresInCompilerGeneratedCode.InitString()' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Message from --InitString--.` so the CompilerGeneratedCode is needed. In fact I think I forgot to add some code handling that I had before

vitek-karas: OK _ I guess we can treat static .ctors as compiler generated code. This property was initially designed for the async/iterator state machine code, but it seems to make sense in other cases as well 😄

tlakollo: Added an additional piece of code in the result checker to verify the static constructor call (was already in https://github.com/dotnet/linker/pull/2113)

tlakollo:
Marked as done several points of https://github.com/dotnet/linker/issues/2299

PR #2337, opened 2021-10-27
Improve analyzer test infra to split linker tests into separate test cases in the TestExplorer

vitek-karas:
xUnit will not split the test cases if the parameters to the Theory are complex objects. But if the parameter is a single string it will do it. So using a trick to store a string->object mapping and identify the tests by the string. When the test is executed the object representation is looked up from the string and executed.

It's not perfect, since xUnit doesn't have the same capability as NUnit to fully customize the generated test case name, so the tests show up as (for example):
`RequiresCapability(m: "RequiresCapability.TestMethod")`

But at least this allows running one test case in isolation.

As a followup, we should split the large tests cases into multiple files. `RequiresCapability` is the first obvious candidate as it has >1500 lines and currently generates 134 test cases for the analyzer. In general the tests should be refactored and split into more files:
* Reasonably small file size and complexity - this helps with debugging, both linker and analyzer will run the product on the entire test file, and so if it has many methods and so on, debugging gets rather tedious.
* Better isolation - we're already doing this by creating lot of subclasses which are used to isolate assets for each test (methods, types, ...). Unfortunately we took this too far, the `RequiresCapability` tests have 3 levels deep nesting currently.
* Easier to navigate - Should be easier to find existing tests
* More test cases in the linker tests - each file is treated as a separate test case for the linker, currently if something fails in `RequiresCapability`, it's just a single test failure - even if internally it is actually many different failures.

### DIFF
--- a/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
@@ -16,14 +16,14 @@ namespace ILLink.RoslynAnalyzer.Tests
 		[MemberData (nameof (TestCaseUtils.GetTestData), parameters: nameof (RequiresCapability))]
 		public void RequiresCapability (string m)
 		{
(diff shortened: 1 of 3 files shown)


vitek-karas:
Thanks a lot @tlakollo and @sbomer for finishing this!

PR #2343, opened 2021-10-29
Ruc on type analyzer part 1 (#2330)

tlakollo:
Add logic to check some of the Ruc on type analysis warnings
Change field and event signature to match analyzer and linker to a single signature mode
Add class as a diagnostic target
Fix issue where attributes ProducedBy.Trimmer/ProducedBy.All are skipped if they are surrounded by ProducedBy.Analyzer attributes
Bring IsConstructor and IsStaticConstructor from Roslyn source
Remove special case for printing warnings in the event instead of the caller of the event accessor

Scenarios of Ruc on type that are not addressed by this PR:
Dealing with interfaces with RUC and having a class implementing it
DAM interaction with RUC
Reflection patterns, some of the tests work because they use typeof() but in general, they are not supported


agocke:
We should do this as an auto-merge

Issue #2308, opened 2021-10-05
[DAM analyzer] Spurious warnings for typeof(Foo) arguments


sbomer:
```csharp
public static void Main() {
    M(typeof(C));
}

// warning IL2082: 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'C.M(Type)'.
// The implicit 'this' argument of method 'C.Main()' does not have matching annotations.
// The source value must declare at least the same requirements as those declared on the target location it is assigned to.
private static void M([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type type) {}
```

@mateoatr I think the typeof case here is causing the extra warning:
https://github.com/dotnet/linker/blob/83c51d44e0ec5cdf408f93bc5576cc22e9cc6ced/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs#L216

I think the intention was to support typeof(T) flowing into an annotated location where T references a generic parameter. Unfortunately it produces unexpected warnings in cases like the above. Discovered while adding shared tests as part of https://github.com/dotnet/linker/pull/2307.

sbomer:
We will want to track both `typeof(Foo)` where `Foo` is a known type, and `typeof(T)` where `T` is a generic parameter. However, since we don't have the binding logic that produces warnings for `typeof(Foo)` values yet, I think it would be acceptable not to track `typeof(Foo)` for now. In that case I think this particular issue could be fixed by only returning `typeOf.TypeOperand` when the type operand in the typeof operation is an `ITypeParameterSymbol`.

PR #2344, opened 2021-10-29
[feature/damAnalyzer] Merge changes from release/6.0.2xx

sbomer:


sbomer:
The recent changes that run tests with all analyzers enabled are causing https://github.com/dotnet/linker/issues/2308 to show up in more places here. We'll probably want to fix that rather than disabling the failing tests like I did in https://github.com/dotnet/linker/pull/2307.

tlakollo:
> The recent changes that run tests with all analyzers enabled are causing #2308 to show up in more places here. We'll probably want to fix that rather than disabling the failing tests like I did in #2307.

Just to double-check, the RAF analyzer is not running because we didn't pass the MSBuildOptions.EnableSingleFileAnalyzer yet. And the Interop files don't have any errors so I guess is just that some changes in that PR make exacerbated #2308, right?

sbomer:
That's right - sorry, I meant the changes that run tests with all _trim_ analyzers. Those changes are exacerbating https://github.com/dotnet/linker/issues/2308. For example the failures in the RequiresCapability tests are coming from the DAM analyzer which didn't used to run on that testcase.

tlakollo:
Considering that there is already a PR that has a fix for the spurious IL2082 and that other PRs will fail without the fix 
`Switch to ubuntu 20.04 agents (dotnet#2322)` I think is best to merge this PR and pollute the branch while we merge #2348

tlakollo (approved):
Looks good to me

PR #2348, opened 2021-11-01
Fixes Issue 2308

### LINKED ISSUE #2308: [DAM analyzer] Spurious warnings for typeof(Foo) arguments

LakshanF:
Fixes #2308 by trying to get the symbol for an operation type only if it is generics

### DIFF
--- a/test/Mono.Linker.Tests.Cases/DataFlow/MethodReturnParameterDataFlow.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/MethodReturnParameterDataFlow.cs
@@ -50,11 +50,6 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 		[UnrecognizedReflectionAccessPattern (typeof (MethodReturnParameterDataFlow), nameof (ReturnPublicParameterlessConstructor),
 			new Type[] { typeof (Type), typeof (Type), typeof (Type) }, returnType: typeof (Type), messageCode: "IL2068")]
 		// TODO: https://github.com/dotnet/linker/issues/2308
-		// This warning should not be produced.
-		[ExpectedWarning ("IL2083",
-			nameof (MethodReturnParameterDataFlow) + "." + nameof (ReturnPublicParameterlessConstructor) + "(Type, Type, Type)",
-			"'this'",
-			ProducedBy = ProducedBy.Analyzer)]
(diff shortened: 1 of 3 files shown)


tlakollo:
@sbomer I see that the lint and some of the other ci tasks are failing because they are trying to run in Ubuntu 16
`##[warning]An image label with the label Ubuntu16 does not exist.`
I think this was solving it https://github.com/dotnet/linker/pull/2338
But it hasn't been merged in the DAM branch, right?

test/Mono.Linker.Tests.Cases/DataFlow/MethodReturnParameterDataFlow.cs
```diff
@@ -50,11 +50,6 @@ static Type NoRequirements ()
 		[UnrecognizedReflectionAccessPattern (typeof (MethodReturnParameterDataFlow), nameof (ReturnPublicParameterlessConstructor),
 			new Type[] { typeof (Type), typeof (Type), typeof (Type) }, returnType: typeof (Type), messageCode: "IL2068")]
 		// TODO: https://github.com/dotnet/linker/issues/2308
```
sbomer, line 52: Nit: remove the TODOs here and below please

sbomer (approved):
Looks good, thank you!

PR #2351, opened 2021-11-01
[DAM Analyzer] RequiresCapability now runs the DAM analyzer

tlakollo:
Let the warnings be generated by Trimmer and analyzer
Whenever discrepancies ocurred open an issue to track and let the test pass


sbomer (approved):
Thank you!

Issue #2310, opened 2021-10-08
.net 5 blazor webassembly project trying to update to .net 6 rc 1 and AOT  


dkboon01:
I have a blazor webassemly project - the base project ABCBlazor is wasm , 2nd project is a razor component library and 3rd project is a dll that is built on .net standard 2.0.   Should I be able to convert my razor component library and my blazor wasm project to dotnet 6 and use AOT when publishing ?  Here is the error I receive.   Please help we are trying to improve on the speed of our application

ILLink(0,0): Error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues
Fatal error in IL Linker
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Mono.Linker.CustomAttributeSource.HasAny(ICustomAttributeProvider provider)
   at Mono.Linker.AnnotationStore.GetLinkerAttributes[T](IMemberDefinition member)
   at Mono.Linker.AnnotationStore.TryGetLinkerAttribute[T](IMemberDefinition member, T& attribute)
   at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
   at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
   at Mono.Linker.Steps.MarkStep.MarkField(FieldDefinition field, DependencyInfo& reason)
   at Mono.Linker.Steps.MarkStep.MarkEntireType(TypeDefinition type, DependencyInfo& reason)
   at Mono.Linker.Steps.MarkStep.MarkEntireType(TypeDefinition type, DependencyInfo& reason)
   at Mono.Linker.Steps.MarkStep.MarkEntireAssembly(AssemblyDefinition assembly)
   at Mono.Linker.Steps.MarkStep.MarkAssembly(AssemblyDefinition assembly, DependencyInfo reason)
   at Mono.Linker.Steps.MarkStep.MarkModule(ModuleDefinition module, DependencyInfo reason)
   at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
   at Mono.Linker.Steps.MarkStep.MarkGenericArguments(IGenericInstance instance)
   at Mono.Linker.Steps.MarkStep.GetOriginalType(TypeReference type, DependencyInfo reason)
[...]

Issue #2290, opened 2021-09-23
NullReferenceException from IL Linker when using Blazor WASM AoT


dsplaisted:
- Clone https://github.com/dsplaisted/MSBuildStructuredLog
- Check out commit [2970055](https://github.com/dsplaisted/MSBuildStructuredLog/commit/2970055062010ada4d440b6bdc6c8c65b6f06d56)
- `cd src\StructuredLogViewer.Web`
- `dotnet publish -p:Configuration=Release /p:RunAOTCompilation=true`

**Expected:** Successful publish

**Actual:** Null reference in IL Linker (and the build also seems to hang, rather then ending with an error):

```
ILLink : error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues [C:\git\MSBuildStructuredLog\src\StructuredLogViewer.Web\StructuredLogViewer.Web.csproj]
  Fatal error in IL Linker
  Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
     at Mono.Linker.CustomAttributeSource.HasAny(ICustomAttributeProvider provider)
     at Mono.Linker.AnnotationStore.GetLinkerAttributes[T](IMemberDefinition member)
     at Mono.Linker.AnnotationStore.TryGetLinkerAttribute[T](IMemberDefinition member, T& attribute)
     at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
     at Mono.Linker.Steps.MarkStep.MarkField(FieldDefinition field, DependencyInfo& reason)
     at Mono.Linker.Steps.MarkStep.MarkEntireType(TypeDefinition type, DependencyInfo& reason)
     at Mono.Linker.Steps.MarkStep.MarkEntireType(TypeDefinition type, DependencyInfo& reason)
     at Mono.Linker.Steps.MarkStep.MarkEntireAssembly(AssemblyDefinition assembly)
     at Mono.Linker.Steps.MarkStep.MarkAssembly(AssemblyDefinition assembly, DependencyInfo reason)
     at Mono.Linker.Steps.MarkStep.ProcessMarkedPending()
     at Mono.Linker.Steps.MarkStep.Initialize()
     at Mono.Linker.Steps.MarkStep.Process(LinkContext context)
     at Mono.Linker.Pipeline.ProcessStep(LinkContext context, IStep step)
     at Mono.Linker.Pipeline.Process(LinkContext context)
[...]

Issue #2286, opened 2021-09-20
 IL Linker has encountered an unexpected error.


textcenter:
I have a blazor project. After upgraded to version 17.0.0 Preview 4.1. It cannot be published any more. It is still working in debug because of no linker involves.
Updates wasm-tools ( dotnet workload install wasm-tools) the issue still
Cannot disable linker in project file with this
<PropertyGroup>
  <BlazorWebAssemblyEnableLinking>false</BlazorWebAssemblyEnableLinking>
</PropertyGroup>

ILLink(0,0): Error IL1012: IL Linker has encountered an unexpected error. Please report the issue at 
https://github.com/mono/linker/issues
Fatal error in IL Linker
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Mono.Linker.CustomAttributeSource.HasAny(ICustomAttributeProvider provider)
   at Mono.Linker.AnnotationStore.GetLinkerAttributes[T](IMemberDefinition member)
   at Mono.Linker.AnnotationStore.TryGetLinkerAttribute[T](IMemberDefinition member, T& attribute)
   at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
   at Mono.Linker.Steps.MarkStep.MarkGenericArguments(IGenericInstance instance)
   at Mono.Linker.Steps.MarkStep.GetOriginalType(TypeReference type, DependencyInfo reason)
   at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
   at Mono.Linker.Steps.MarkStep.MarkField(FieldDefinition field, DependencyInfo& reason)
   at Mono.Linker.Steps.MarkStep.MarkEntireType(TypeDefinition type, DependencyInfo& reason)
   at Mono.Linker.Steps.MarkStep.MarkEntireAssembly(AssemblyDefinition assembly)
   at Mono.Linker.Steps.MarkStep.MarkAssembly(AssemblyDefinition assembly, DependencyInfo reason)
   at Mono.Linker.Steps.MarkStep.MarkModule(ModuleDefinition module, DependencyInfo reason)
   at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
   at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
[...]

Issue #2274, opened 2021-09-15
Crash in linker building with .NET 6 RC1


slluis:
Building VSMac with .NET 6 RC1 I get this error:

```
ILLink : error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues [/Users/lluis/prog/work/vsmac/test/main/src/tools/vstool/vstool.csproj]
  Fatal error in IL Linker
  Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
     at Mono.Linker.CustomAttributeSource.HasAny(ICustomAttributeProvider provider)
     at Mono.Linker.AnnotationStore.GetLinkerAttributes[T](IMemberDefinition member)
     at Mono.Linker.AnnotationStore.TryGetLinkerAttribute[T](IMemberDefinition member, T& attribute)
     at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
     at Mono.Linker.Steps.MarkStep.MarkField(FieldDefinition field, DependencyInfo& reason)
     at Mono.Linker.Steps.MarkStep.MarkEntireType(TypeDefinition type, DependencyInfo& reason)
     at Mono.Linker.Steps.MarkStep.MarkEntireAssembly(AssemblyDefinition assembly)
     at Mono.Linker.Steps.MarkStep.MarkAssembly(AssemblyDefinition assembly, DependencyInfo reason)
     at Mono.Linker.Steps.MarkStep.ProcessMarkedPending()
     at Mono.Linker.Steps.MarkStep.Initialize()
     at Mono.Linker.Steps.MarkStep.Process(LinkContext context)
     at Mono.Linker.Pipeline.ProcessStep(LinkContext context, IStep step)
     at Mono.Linker.Pipeline.Process(LinkContext context)
     at Mono.Linker.Driver.Run(ILogger customLogger)
     at Mono.Linker.Driver.Main(String[] args)
  Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
[...]

PR #2354, opened 2021-11-04
Fixing a null ref exception related to function pointers and interfaces

tannergooding:
One of my projects was throwing a `NullReferenceException` in .NET 6 RC2:
```
All projects are up-to-date for restore.
TerraFX.Interop.Windows -> C:\Users\tagoo\Source\repos\terrafx.interop.windows\artifacts\bin\sources\TerraFX.Interop.Windows\Release\net6.0\TerraFX.Interop.Windows.dll
TerraFX.Samples.DirectX -> C:\Users\tagoo\Source\repos\terrafx.interop.windows\artifacts\bin\samples\TerraFX.Samples.DirectX\Release\net6.0\win-x64\TerraFX.Samples.DirectX.dll
ILLink : error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues [C:\Users\tagoo\Source\repos\terrafx.interop.windows\samples\DirectX\TerraFX.Samples.DirectX.csproj]
Fatal error in IL Linker
at Mono.Linker.TypeMapInfo.TypeMatch(TypeReference a, TypeReference b)
at Mono.Linker.TypeMapInfo.MethodMatch(MethodReference candidate, MethodReference method)
at Mono.Linker.TypeMapInfo.TryMatchMethod(TypeReference type, MethodReference method)
at Mono.Linker.TypeMapInfo.MapInterfaceMethodsInTypeHierarchy(TypeDefinition type)
at Mono.Linker.TypeMapInfo.MapType(TypeDefinition type)
at Mono.Linker.TypeMapInfo.EnsureProcessed(AssemblyDefinition assembly)
at Mono.Linker.TypeMapInfo.GetBaseMethods(MethodDefinition method)
at Mono.Linker.Steps.MarkStep.IsVirtualNeededByInstantiatedTypeDueToPreservedScope(MethodDefinition method)
at Mono.Linker.Steps.MarkStep.GetRequiredMethodsForInstantiatedType(TypeDefinition type)+MoveNext()
at Mono.Linker.Steps.MarkStep.MarkRequirementsForInstantiatedTypes(TypeDefinition type)
at Mono.Linker.Steps.MarkStep.MarkType(TypeReference reference, DependencyInfo reason, Nullable`1 origin)
at Mono.Linker.Steps.MarkStep.MarkField(FieldDefinition field, DependencyInfo& reason)
at Mono.Linker.Steps.MarkStep.MarkEntireType(TypeDefinition type, DependencyInfo& reason)
at Mono.Linker.Steps.MarkStep.MarkEntireAssembly(AssemblyDefinition assembly)
at Mono.Linker.Steps.MarkStep.MarkAssembly(AssemblyDefinition assembly, DependencyInfo reason)
at Mono.Linker.Steps.MarkStep.ProcessMarkedPending()
at Mono.Linker.Steps.MarkStep.Initialize()
at Mono.Linker.Steps.MarkStep.Process(LinkContext context)
Optimizing assemblies for size, which may change the behavior of the app. Be sure to test after publishing. See: https://aka.ms/dotnet-illink
C:\Program Files\dotnet\sdk\6.0.100-rc.2.21505.57\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.ILLink.targets(108,5): error NETSDK1144: Optimizing assemblies for size failed. Optimization can be disabled by setting the PublishTrimmed property to false. [C:\Users\tagoo\Source\repos\terrafx.interop.windows\samples\DirectX\TerraFX.Samples.DirectX.csproj]
```

This was ultimately because the types that were being matched were `FunctionPointerType` for which `.ElementType` returns `null`. This updates the `TypeMatch` to account for `FunctionPointerType` and do a comparison of calling convention, return type, and parameter types.

### DIFF
--- a/src/linker/Linker/TypeMapInfo.cs
+++ b/src/linker/Linker/TypeMapInfo.cs
@@ -369,6 +369,9 @@ namespace Mono.Linker
 			if (a is IModifierType mta)
 				return TypeMatch (mta, (IModifierType) b);
 
+			if (a is FunctionPointerType fpta)
+				return TypeMatch (fpta, (FunctionPointerType) b);
+
 			return TypeMatch (a.ElementType, b.ElementType);
 		}
 
@@ -407,6 +410,39 @@ namespace Mono.Linker
 			return true;
 		}
 
+		static bool TypeMatch (FunctionPointerType a, FunctionPointerType b)
+		{
+			if (a.HasParameters != b.HasParameters)
+				return false;
+
+			if (a.CallingConvention != b.CallingConvention)
+				return false;
+
+			// we need to track what the generic parameter represent - as we cannot allow it to
+			// differ between the return type or any parameter
+			if (a.ReturnType is not TypeReference aReturnType ||
+				b.ReturnType is not TypeReference bReturnType ||
+				!TypeMatch (aReturnType, bReturnType))
+				return false;
+
+			if (!a.HasParameters)
+				return true;
+
+			var ap = a.Parameters;
+			var bp = b.Parameters;
+			if (ap.Count != bp.Count)
+				return false;
+
+			for (int i = 0; i < ap.Count; i++) {
+				if (a.Parameters[i].ParameterType is not TypeReference aParameterType ||
+					b.Parameters[i].ParameterType is not TypeReference bParameterType ||
+					!TypeMatch (aParameterType, bParameterType))
+					return false;
+			}
+
+			return true;
+		}
+
 		static bool TypeMatch (TypeReference a, TypeReference b)
 		{
 			if (a is TypeSpecification || b is TypeSpecification) {
@@ -418,6 +454,9 @@ namespace Mono.Linker
 
 			if (a is GenericParameter genericParameterA && b is GenericParameter genericParameterB)
 				return TypeMatch (genericParameterA, genericParameterB);
+
+			if ((a is null) || (b is null))
+				return a == b;
 
 			return a.FullName == b.FullName;
 		}
--- a/src/linker/Linker/TypeMapInfo.cs
+++ b/src/linker/Linker/TypeMapInfo.cs
@@ -455,9 +455,6 @@ namespace Mono.Linker
 			if (a is GenericParameter genericParameterA && b is GenericParameter genericParameterB)
 				return TypeMatch (genericParameterA, genericParameterB);
 
-			if ((a is null) || (b is null))
-				return a == b;
-
 			return a.FullName == b.FullName;
 		}
 	}
--- a/test/Mono.Linker.Tests.Cases/FunctionPointers/CanCompileInterfaceWithFunctionPointerParameter.cs
+++ b/test/Mono.Linker.Tests.Cases/FunctionPointers/CanCompileInterfaceWithFunctionPointerParameter.cs
@@ -1,0 +1,37 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.FunctionPointers
+{
+	[SetupCompileArgument ("/unsafe")]
+	unsafe class CanCompileInterfaceWithFunctionPointerParameter
+	{
+		public static void Main()
+		{
+			new CanCompileInterfaceWithFunctionPointerParameter.B ().Method (null);
+		}
+
+		[KeptMember (".ctor()")]
+		class B : I
+		{
+			public void Unused (delegate* unmanaged<void> fnptr)
+			{
+			}
+
+			[Kept]
+			public void Method (delegate* unmanaged<void> fnptr)
+			{
+			}
+		}
+
+		interface I
+		{
+			void Unused (delegate* unmanaged<void> fnptr);
+
+			void Method (delegate* unmanaged<void> fnptr);
+		}
+	}
+}
--- a/test/Mono.Linker.Tests.Cases/FunctionPointers/CanCompileInterfaceWithFunctionPointerParameter.cs
+++ b/test/Mono.Linker.Tests.Cases/FunctionPointers/CanCompileInterfaceWithFunctionPointerParameter.cs
@@ -9,7 +9,7 @@ namespace Mono.Linker.Tests.Cases.FunctionPointers
 	[SetupCompileArgument ("/unsafe")]
 	unsafe class CanCompileInterfaceWithFunctionPointerParameter
 	{
-		public static void Main()
+		public static void Main ()
 		{
 			new CanCompileInterfaceWithFunctionPointerParameter.B ().Method (null);
 		}
--- a/test/Mono.Linker.Tests.Cases/FunctionPointers/CanCompileMethodWithFunctionPointerParameter.cs
+++ b/test/Mono.Linker.Tests.Cases/FunctionPointers/CanCompileMethodWithFunctionPointerParameter.cs
@@ -9,7 +9,7 @@ namespace Mono.Linker.Tests.Cases.FunctionPointers
 	[SetupCompileArgument ("/unsafe")]
 	unsafe class CanCompileMethodWithFunctionPointerParameter
 	{
-		public static void Main()
+		public static void Main ()
 		{
 			new CanCompileMethodWithFunctionPointerParameter.B ().Method (null);
 		}
--- a/test/Mono.Linker.Tests.Cases/FunctionPointers/CanCompileMethodWithFunctionPointerParameter.cs
+++ b/test/Mono.Linker.Tests.Cases/FunctionPointers/CanCompileMethodWithFunctionPointerParameter.cs
@@ -1,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
+
+namespace Mono.Linker.Tests.Cases.FunctionPointers
+{
+	[SetupCompileArgument ("/unsafe")]
+	unsafe class CanCompileMethodWithFunctionPointerParameter
+	{
+		public static void Main()
+		{
+			new CanCompileMethodWithFunctionPointerParameter.B ().Method (null);
+		}
+
+		[KeptMember (".ctor()")]
+		class B
+		{
+			public void Unused (delegate* unmanaged<void> fnptr)
+			{
+			}
+
+			[Kept]
+			public void Method (delegate* unmanaged<void> fnptr)
+			{
+			}
+		}
+	}
+}
--- a/test/Mono.Linker.Tests/TestCases/TestDatabase.cs
+++ b/test/Mono.Linker.Tests/TestCases/TestDatabase.cs
@@ -86,6 +86,11 @@ namespace Mono.Linker.Tests.TestCases
 			return NUnitCasesBySuiteName ("FeatureSettings");
 		}
 
+		public static IEnumerable<TestCaseData> FunctionPointersTests ()
+		{
+			return NUnitCasesBySuiteName ("FunctionPointers");
+		}
+
 		public static IEnumerable<TestCaseData> GenericsTests ()
 		{
 			return NUnitCasesBySuiteName ("Generics");
--- a/test/Mono.Linker.Tests/TestCases/TestSuites.cs
+++ b/test/Mono.Linker.Tests/TestCases/TestSuites.cs
@@ -103,6 +103,12 @@ namespace Mono.Linker.Tests.TestCases
 			Run (testCase);
 		}
 
+		[TestCaseSource (typeof (TestDatabase), nameof (TestDatabase.FunctionPointersTests))]
+		public void FunctionPointerTests (TestCase testCase)
+		{
+			Run (testCase);
+		}
+
 		[TestCaseSource (typeof (TestDatabase), nameof (TestDatabase.GenericsTests))]
 		public void GenericsTests (TestCase testCase)
 		{


src/linker/Linker/TypeMapInfo.cs
```diff
@@ -407,6 +410,39 @@ static bool TypeMatch (GenericParameter a, GenericParameter b)
 			return true;
 		}
 
+		static bool TypeMatch (FunctionPointerType a, FunctionPointerType b)
```
tannergooding, line 413: Not 100% this is covering everything, but it at least appears to work correctly with the function pointers I have declared.

src/linker/Linker/TypeMapInfo.cs
```diff
@@ -419,6 +455,9 @@ static bool TypeMatch (TypeReference a, TypeReference b)
 			if (a is GenericParameter genericParameterA && b is GenericParameter genericParameterB)
 				return TypeMatch (genericParameterA, genericParameterB);
 
+			if ((a is null) || (b is null))
```
tannergooding, line 458: This is protecting against the crashing case where either input is null.

marek-safar:
Do you have any test to show that the current logic is not good enough?

tannergooding:
@marek-safar, I pushed up two tests. `CanCompileInterfaceWithFunctionPointerParameter` fails with a `NullReferenceException` without this change.

tannergooding:
-- Given this results in a crash on .NET 6, should this be considered for servicing as well?

tannergooding:
Retargeted to release/6.0.2xx as requested by @vitek-karas offline.

rickbrew:
This crash will affect Paint.NET's use of illink once TerraFX.Interop.Windows starts using some of the new code that Tanner is putting in there, so I much appreciate this being pulled in for 6.0 servicing :)

src/linker/Linker/TypeMapInfo.cs
tannergooding: Most notably, function pointers can't be generic but they can have generic arguments.

There isn't a `context` available here, so I'm unsure if there needs to be any additional handling or logic to cover that scenario.

tannergooding: There are also `modreq` that can appear as part of the calling convention. I think `modreq` are handled further up in the `TypeMatch` logic, however.

src/linker/Linker/TypeMapInfo.cs
vitek-karas: Hmm - maybe it would be better to let it crash - at least that way we find out quickly. This way it may just not work correctly - which would be MUCH harder to figure out.

vitek-karas (commented):
Can you please run `.\lint.cmd` (or `./lint.sh`) from the repo root? It will reformat the code as appropriate.

src/linker/Linker/TypeMapInfo.cs
tannergooding: Is there a way to add a diagnostic/warning if it hits this path?

I think a crash is bad/worst case scenario, particularly in a shipping product. Having the ability to have a reasonable fallback is better, but hinders the ability to catch these scenarios as you indicated. Surfacing a warning is the next best thing so users can still see the issue without being blocked.

MichalStrehovsky: Warnings need to have error codes so that they're suppressible for people who WarnAsError + they need localizable messages. I'm not sure introducing warnings for internal errors is worth the cost of documenting and maintaining them.

There's also people who ignore warnings, come with broken app asking for support without mentioning the warning, and we're left reading from tea leaves (my experience from years in .NET Native support).

tannergooding: My own view here, as a consumer of the app, is that having absolutely no path forward because I happened to hit an unhandled path in the app is an even worse experience experience. My two options here were:
A. "don't use trimming"
B. "figure out what broke the trimming and don't do that"

I didn't get the option to live with additional bloat in my trimmed app and I didn't get any diagnostic information that I could provide in an issue. Instead, I had a stack trace and happened to be working on an open source project that would have potentially made it simple to reproduce the error. This isn't going to be the case most of the time and most users aren't going to have the interest to even debug and try to root cause the issue in the first place (let alone consider logging an issue or being aware of where to log the issue).

We're advertising trimming even more for .NET 6 and have made massive leaps and bounds in reporting useful information to the user. Having a warning that can provide additional information and having a workaround that can be used, even with a slightly degraded experience is, to me, a significantly better setup and is effectively the minimum I'd come to expect from our .NET tooling.

tannergooding: > I'm not sure introducing warnings for internal errors is worth the cost of documenting and maintaining them.

Other tools, like Roslyn, catch and surface these as a general error documenting its an internal error. They always gracefully exit where possible.

MichalStrehovsky: `ILLink : error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues` is already in the message.

We could hide the stack trace; we were hiding it on .NET Native because the PM team insisted it looks unprofessional. Now that we're "agile" and nothing comes with a spec, the PM team doesn't concern themselves with these things anymore. For the dev team the stack trace is useful for quick triage (e.g. there was a hard hitter bug #2290 in RC1 that was easy to triage thanks to the stack). If the process exits cleanly, we don't get a Watson dump.

I'm a proponent of complex systems failing as soon as they realize they are in an unexpected state. The extra null check in a place that shouldn't see a null is no different from `On Error Resume Next` in VB. Unexpected states only get worse and harder to root cause the longer they're tolerated. Crash dumps we get don't have a list of warnings that were generated as the process was trying to `Resume Next`.

jeffhandley: > My two options here were:
> A. "don't use trimming"
> B. "figure out what broke the trimming and don't do that"

@MichalStrehovsky Is this unexpected state happening in a narrow enough code path that the message could confidently indicate that trimming needs to be disabled until this issue is fixed? That way if the two options @tannergooding mentioned aren't obvious to the user, at least we're giving them that hint.

MichalStrehovsky: > Is this unexpected state happening in a narrow enough code path that the message could confidently indicate that trimming needs to be disabled until this issue is fixed?

Suggestion to disable trimming is also there (`error NETSDK1144: Optimizing assemblies for size failed. Optimization can be disabled by setting the PublishTrimmed property to false.`). Anything more specific would probably be just guesswork (the tool could report the member it was looking at when unexpected state happened, but that could be just a red herring depending on what unexpected thing actually happened).

I'm afraid the messages getting drowned out by the stack trace. I'm kind of starting to understand why the PM team insisted we don't show the stack trace in .NET Native.

tannergooding: > For the dev team the stack trace is useful for quick triage (e.g. there was a hard hitter bug #2290 in RC1 that was easy to triage thanks to the stack).

This can easily be left enabled for debug or non release/preview builds.

> Unexpected states only get worse and harder to root cause the longer they're tolerated. Crash dumps we get don't have a list of warnings that were generated as the process was trying to Resume Next.

We don't get crash dumps on Linux or MacOS either and this isn't an issue for the SDK, Compiler, VS Code, MSBuild, or any of our other tooling that are arguably much more widely used (at least today since trimming support in the SDK is "relatively new" in comparison).

Additionally, WER (which is what gives us Watsons and handles the opt-in telemetry reporting at the OS level) has explicit APIs available that can be used to create and send reports without crashing the application. 

This really feels like something that should be handled more at the SDK or general .NET tooling level via the telemetry or other helper APIs so we have a consistent experience on all platforms/operating systems and don't force crashing experiences onto the end user.

> The extra null check in a place that shouldn't see a null is no different from On Error Resume Next in VB.

This is not an unexpected null. This is something that will keep happening as the language and ecosystem evolves. There are a near infinite number of possible IL patterns and plenty of features, attributes, and other special cases that we will never handle or that will get introduced as the runtime and ecosystem keep evolving.

> Suggestion to disable trimming is also there

I think the more people start depending on trimming, the less acceptable this is going to be.

When a major app starts using trimming as a primary feature to help reduce size or otherwise optimize the app; some new feature happening to bring down trimming for the entire app isn't really a good experience or good look on us.

The worst case scenario of the `null` check here (or any error handling in these comparisons functions) is that less gets trimmed than intended. That is, IMO, a better experience than disabling trimming altogether which means absolutely nothing gets trimmed as it at least gives options to users.

Alternatives exist for error reporting, t

vitek-karas: > The worst case scenario of the null check here (or any error handling in these comparisons functions) is that less gets trimmed than intended.

That's not always true. The worse case is if trimmer doesn't keep something it should. If I read this correctly: If we get null here we will perform reference comparison, which means we make decide that something  is different even though it really is the same. This would translate into virtual method lookup not matching a method, even though it should. The end result might be that we don't mark the method in question. The app will break at runtime then.

This should definitely at the very least produce a warning.

tannergooding: > If I read this correctly: If we get null here we will perform reference comparison

Right the logic is:
* if both are null, consider them a match
* if only one is null, consider them not a match
* if both are not null, it continues using the full name comparison it was already using.

We could consider them always a match if either is null, if we think that's better. Such a scenario being a false negative would be dependent on both `TypeReference` being different enough that they are equal despite one returning `null` and not the other, which doesn't quite seem right from what I recall of how they work.

MichalStrehovsky: > > For the dev team the stack trace is useful for quick triage (e.g. there was a hard hitter bug #2290 in RC1 that was easy to triage thanks to the stack).

> This can easily be left enabled for debug or non release/preview builds

It would also downgrade the experience around root causing #2274, #2282, #2286, #2290, #2310 to "please collect all the files on the input so that we can repro this bug in house under debugger" because we don't ship non-release builds (this was already in an RC). The stack trace easily saved an hour of dev time in the past month only. It is a compromise one way or the other.

> This is not an unexpected null. This is something that will keep happening as the language and ecosystem evolves.

Illinker needs to exit unsuccessfully if it doesn't understand the input. The optimizations in illinker assume whole program view - if it doesn't understand the whole program, it's not safe to remove things from it. If we end up in a situation where a trimmed program starts behaving weirdly after trimming because illinker was fine not knowing what's going on, we'll see bugs getting opened in the runtime repo before a type system dev root causes them to be an illinker bug. The code is currently used to decide what to keep, but maybe in the future illinker starts doing devirtualization - would we still be fine with `TypeMatch` for `delegate*<void>` and `delegate*<int>` to return "yeah, totally"? At minimum we would have to rename this to `TypeMightMatch` so that it's more obvious at the callsite what this does, but even then I'm not sure the risk of bad bugs is worth it.

tannergooding: > It would also downgrade the experience around root causing #2274, #2282, #2286, #2290, #2310 to "please collect all the files on the input so that we can repro this bug in house under debugger" because we don't ship non-release builds (this was already in an RC). The stack trace easily saved an hour of dev time in the past month only. It is a compromise one way or the other.

So roughly the same experience as the rest of the tools we ship in the .NET SDK? This isn't something unique to the linker.

> The optimizations in illinker assume whole program view - if it doesn't understand the whole program, it's not safe to remove things from it.

Why not just assume that thing can't be trimmed if it can't be understood. Having slightly too much is better than forcing you to have everything. I agree that having too little is worse than both, but we don't have to be in that scenario. Most of the time, the thing will be limited to one aspect and a warning at least makes it visible "we couldn't trim this because we couldn't understand it, please log an issue here".

tannergooding: It's not clear to me why the linker needs to be different from the rest of the tooling the SDK ships, in terms of diagnostics reported, how users provide crash dumps or repros, etc.

It's likewise not clear to me why we have to force trimming to be entirely disabled because one aspect or feature isn't recognized. We should be able to correctly handle "fully understood" vs "partially understood" and "partially understood" vs "partially understood" and have it do the right thing (whether that's assume not trimmable or assume can't be a match, etc)

MichalStrehovsky: > So roughly the same experience as the rest of the tools we ship in the .NET SDK? This isn't something unique to the linker.

Right, but illinker already has a devloop experience around that that is a lot more efficient for the dev team, so the illinker team would be giving up that. It's harder to give up something one has than something one never had, especially if it's this easy to quantify the cost of not having it (just see how easy it was to troubleshoot the 5 bug reports - even the initial one was easy because of the stack trace). I'm not really on the illink dev team officially, so it's not my call to make but the advantage is obvious. If you filed the thing you wrote on Teams that prompted this fix as a drive-by issue and never responded to any questions after, we would probably still be able to come up with a repro case and a fix.

> Why not just assume that thing can't be trimmed if it can't be understood.

It might not even be possible to reconstruct the appropriate metadata with Cecil without having some knowledge about it. We (as the runtime team) have been doing breaking changes to the metadata format without revving the version field with the assumption that it's "close enough and tools might handle it", but none of that was ever tested with the tools. Changes to the file formats need to be understood by illinker. Not even that - changes to _runtime_ need to be understood by illinker. E.g. `IDynamicIntefaceCastable` needed linker changes not to strip default implementations of things that (absent `IDynamicInterfaceCastable`) provably weren't used. The newly added `CreateSpan` API will need illinker changes so that the underlying RVA static fields are properly aligned when writing the trimmed output. S.R.Metadata is not able to encode byref to a byref and will need a public API change to emit it; who knows what Cecil does with that. Etc. I don't see much value in trying to futureproof for something that likely will need special casing anyway at the risk of the futureproofing code kicking in for straight up bugs.

tannergooding: > It might not even be possible to reconstruct the appropriate metadata with Cecil without having some knowledge about it.

Perhaps there needs to be a way for consumers to be able to have a faster inner loop around the `illinker` tooling and getting a fixed version? Right now, I can't even override the inbox linker with a private copy that has the fix, so my repo is "dead in the water", as it were, until the next SDK servicing release.

If I had caught this just after 6.0 shipped or just after the last servicing release, I'd have to wait around a month for the next fix and that's my main concern here.
- That is, my main concern is I found a bug and now I have no real way to workaround the issue so my app is "regressed" in terms of size and startup perf. Its not a huge deal right now, but it very well could be at another time or for another person using the tooling.

One example, of what I mean, is that Roslyn for example defaults to the inbox version, but I can trivially reference https://www.nuget.org/packages/Microsoft.Net.Compilers/4.0.0-6.final for example and get a newer compiler version or even a nightly build for prototyping; which unblocks these kinds of issues, even temporarily, until the proper fix becomes available.

vitek-karas: > Right now, I can't even override the inbox linker with a private copy


You can set a property `ILLinkTasksAssembly` to point to a private build of illink (and the related task) on your build/publish and it will use that. We don't have a convenient NuGet to do this for you though.

sbomer: You _should_ be able to add a packageref to `Microsoft.NET.ILLink.Tasks` to override the inbox version. We don't ship this package on nuget.org (it's on the dotnet feeds), and we don't advertise or test this scenario, but it's set up to work this way.

This logic doesn't look like it covers the case where:
- one is a non-null TypeSpecification, and the other is null (I think this would give a nullref exception on `GetType`).

In any case I agree we should let it fail with an exception since that's the approach we have been taking in this codebase.

@tannergooding my read of your feedback is that we should avoid blocking our customers if at all possible in case of a linker crash - whether that means:
1. shipping a fix OOB
2. providing enough diagnostics to let customers understand the issue and work around it in their code
3. an option to tell the linker to do a best-effort recovery attempt (accepting the risk that it might generate bad code), possibly with extra diagnostics

We have the option to do 1, and the stacktrace is close to the best we can do for 2 given that linker crashes are unknown unknowns until we uncover them. I understand the frustration of hitting a linker crash and being blocked by it - do you still think something like 3 would be valuable (with a big disclaimer of course)? If so I'd appreciate if you could file an issue to track the suggestion. It would probably impact more of the linker than just the `TypeMatch` logic.

sbomer (commented):
Thanks a lot for the contribution!

src/linker/Linker/TypeMapInfo.cs
tannergooding: I've removed the changes I added here and kept this just as checking for and handling `FunctionPointerType`

marek-safar:
> Retargeted to release/6.0.2xx as requested 

This should go to main first and only then be backported to relevant branches (note 6.0.2 is likely months away)

tannergooding:
@marek-safar, retargeted.

Please let me know if anything else is needed here.

sbomer (approved):

marek-safar (approved):

PR #2360, opened 2021-11-09
[DAM analyzer] Add basic dataflow analysis

sbomer:
This adds basic dataflow analysis for the Roslyn analyzer. So far it only tracks DAM annotations (not known types, reflection intrinsics, and similar). It also doesn't handle exceptional control flow, but unlike the linker's analysis it does handle backwards branches.

The dataflow framework is set up to (hopefully) facilitate code sharing going forward - the core abstractions don't depend on Roslyn constructs, but there's more work that would need to be done to actually share this code with the linker.

As the analysis runs, it caches any reflection access patterns it encounters, and warnings are reported after it converges. The caching is intended to be additive - it should ensure that we increase the set of possible reflection accesses as we go, but it should never lose track of a pattern encountered in a previous iteration of the analysis.

The core logic for creating and propagating the tracked values is in DynamicallyAccessedMembersVisitor, similar to ReflectionMethodBodyScanner. There is a base class LocalDataFlowVisitor, similar to MethodBodyScanner, which knows how to propagate values across locals in Roslyn in a generic way.

The actual dataflow algorithm (ForwardDataFlowAnalysis) does the simplest possible thing in this change. The goal is to set it up in a way that the core abstractions don't need to change even when we go optimize the algorithm (and ideally for any optimizations to be generic, not tightly bound with our particular analysis). This should make it easier to reuse the code in the linker and possibly in other future analyses.

PTAL! I am particularly interested in feedback on naming of some of the wrapper classes which exist only to implement interfaces or to provide generic instantiations.

@agocke I know you were interested in seeing a spec for the dataflow analysis - I still plan on doing that but wanted to get this change out for now.


vitek-karas:
First batch of comments  - still working on this

vitek-karas (commented):
Actually adding some comments

src/ILLink.Shared/Dataflow/DictionaryLattice.cs
```diff
@@ -0,0 +1,77 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+
+namespace ILLink.Shared
+{
+	public class DefaultValueDictionary<TKey, TValue> : Dictionary<TKey, TValue>,
```
vitek-karas, line 9: Might be a good idea to move this to a separate file - it's used outside of the DictionaryLattice below. And it's a more "general use case" class I think.

src/ILLink.Shared/Dataflow/DictionaryLattice.cs
```diff
@@ -0,0 +1,77 @@
...
+
+		public DefaultValueDictionary (DefaultValueDictionary<TKey, TValue> other)
+			: base (other)
+		{
+			DefaultValue = other.DefaultValue;
+		}
+
+		public TValue Get (TKey key) => TryGetValue (key, out var value) ? value : DefaultValue;
+
+		public void Set (TKey key, TValue value)
```
vitek-karas, line 26: This is tricky - if I forget and call `Add` or use the indexer, it will actually insert the default value. Functionality-vise it's still going to work just fine, but there will effectively be a "memory leak" (well sort of).

I wonder if it would be better to simply wrap the dictionary and have only a limitted APIs available on the class. It's unlikely we will need it to act as a true dictionary anyway.

src/ILLink.Shared/Dataflow/SingleValue.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace ILLink.Shared
+{
+	// Similar to ValueNode, and a candidate for future code sharing.
+	// This is a sum type over the various kinds of values we track:
+	// - dynamicallyaccessedmembertypes-annotated locations (types or strings)
+	// - known typeof values and similar
+	// - known strings
+	// - known integers
+
+	// The implementation is taken from the generated code for records.
```
agocke, line 15: I'm pretty sure we can just add a definition for System.Runtime.CompilerServices.IsExternalInit into the project, and records will work fine.

src/ILLink.Shared/Dataflow/ILattice.cs
```diff
@@ -0,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace ILLink.Shared
+{
+	// The interface constraint on TValue also ensures that trying to instantiate
+	// ILattice over a nullable type will produce a warning or error.
+	public interface ILattice<TValue> where TValue : IEquatable<TValue>
```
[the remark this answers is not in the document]
agocke: You could also just call this LocalDataFlow. I think we'll eventually want to share the implementation of tracking of values for both the analyzer and the linker.

src/ILLink.Shared/Dataflow/DictionaryLattice.cs
sbomer: Good point, I made it a wrapper struct instead.

src/ILLink.Shared/Dataflow/HashSetLattice.cs
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		public override int GetHashCode ()
+		{
+			if (Values == null)
+				return typeof (HashSetWrapper<TValue>).GetHashCode ();
+
+			int hashCode = 0;
+			foreach (var item in Values) {
+				hashCode *= -1521134295;
```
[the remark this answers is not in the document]
sbomer: I at least added a helper that hides these constants. But the helper isn't used from linker code yet.

src/ILLink.Shared/Dataflow/HashSetLattice.cs
```diff
@@ -0,0 +1,94 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ILLink.Shared
+{
+	public struct HashSetWrapper<TValue> : IEquatable<HashSetWrapper<TValue>>, IEnumerable<TValue>
```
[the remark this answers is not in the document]
sbomer: I like that suggestion, done!

src/ILLink.Shared/Dataflow/ITransfer.cs
```diff
@@ -0,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace ILLink.Shared
+{
+	// TValue is a class because the typical implementation is a
```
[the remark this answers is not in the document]
sbomer: Added some comments - please let me know if you have suggestions how they could be improved.

src/ILLink.RoslynAnalyzer/LocalDataFlow/LocalDataFlowVisitor.cs
```diff
@@ -0,0 +1,186 @@
...
+	public abstract class LocalDataFlowVisitor<TValue, TValueLattice> : OperationVisitor<LocalState<TValue>, TValue>,
+		ITransfer<BlockWrapper, LocalState<TValue>, LocalStateLattice<TValue, TValueLattice>>
+		where TValue : IEquatable<TValue>
+		where TValueLattice : ILattice<TValue>
+	{
+		protected readonly LocalStateLattice<TValue, TValueLattice> LocalStateLattice;
+
+		protected readonly OperationBlockAnalysisContext Context;
+
+		protected TValue TopValue => LocalStateLattice.Lattice.ValueLattice.Top;
```
agocke, line 27: Why not fold the LocalStateLattice into this type?

src/ILLink.Shared/Dataflow/ILattice.cs
```diff
@@ -0,0 +1,39 @@
...
+	// so the lattice instance provides the Top value.
+	public interface ILattice<TValue> where TValue : IEquatable<TValue>
+	{
+		// We require that the lattice has a "Top" or maximum element.
+		// Top is >= a for every element a of the lattice.
+		// Top is an identity for Meet: Meet(a, Top) = a.
+
+		// The typical use in a dataflow analysis is for Top to represent the "unknown" initial state
+		// with the least possible information about the analysis.
+		public TValue Top { get; }
```
agocke, line 32: FWIW, if we don't anticipate writing more than one data flow analysis, we could use descriptive, DAM-specific names

src/ILLink.Shared/Dataflow/ITransfer.cs
```diff
@@ -0,0 +1,32 @@
...
+	// "simulates". It isn't constrained by the interface, but is typically a basic block,
+	// where the transfer functions are defined in terms of transfer functions for individual
+	// operations in the block.
+
+	// TLattice isn't typically used in the implementation except to provide the "Top" value.
+	// This expresses the conceptual constraint that the transferred values are part of a lattice.
+	public interface ITransfer<TOperation, TValue, TLattice>
+		where TValue : IEquatable<TValue>
+		where TLattice : ILattice<TValue>
+	{
```
agocke, line 26: I don't think this abstraction is right -- a lattice isn't used by the transfer function, instead the mathematical structure is that

1. A data-flow analysis is a program for calculating the set of possible values
2. It requires a domain, which consists of a set of values being tracked, a control flow graph made of basic blocks, a transfer function between those blocks, a defined top state, and a joining function for control flow edges.

I think this means that the right abstraction combines all these things into one. They wouldn't be seperable:

```C#
interface IDataFlowAnalysis<TValueSet>
{
    public Run(ControlFlowGraph graph)
    {
        ...
    }

    public TValue Transfer (TValue);
    public TValue Top { get; }
    public TValue Meet (TValue, TValue);
}
```

In other words, a lattice is a structure that a data-flow analysis _is_, not something that other things _have_

agocke: Ah, I see -- you're abstracting away local variable analysis as well, so you need to need two lattice implementations: one for the variable state, and one for the global state.

I'll see if I can come up with a shorter way to express this.

agocke: Hm, I think the biggest simplification would be getting rid of support for general data flow and just doing DAM, but I'm not sure if there was another data flow analysis we need to support.

The other simplification could be creating an `IDataFlowAnalysis : ILattice` and adding `Transfer` to `IDataFlowAnalysis`. Having those split out doesn't seem to providing a lot of generality.

src/ILLink.RoslynAnalyzer/LocalDataFlow/LocalDataFlowVisitor.cs
```diff
@@ -0,0 +1,186 @@
...
+		public abstract void HandleArgument (TValue argument, IArgumentOperation operation);
+
+		// This takes an IOperation rather than an IReturnOperation because the return value
+		// may (must?) come from BranchValue of an operation whose FallThroughSuccessor is the exit block.
+		public abstract void HandleReturnValue (TValue returnValue, IOperation operation);
+
+#pragma warning disable CS8765
+		// Override with a non-nullable return value to prevent some warnings.
+		// This is safe because the interface constraint on TValue ensures that it's not a nullable type.
+		public override TValue Visit (IOperation operation, LocalState<TValue> argument) => operation.Accept (this, argument)!;
```
agocke, line 86: Are you sure the warning here is about `TValue` not `operation`?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersAnalysis.cs
```diff
@@ -0,0 +1,85 @@
...
+		public BlockWrapper Entry => new BlockWrapper (ControlFlowGraph.Blocks[0]);
+
+		public IEnumerable<BlockWrapper> GetPredecessors (BlockWrapper block)
+		{
+			foreach (var predecessor in block.Block.Predecessors)
+				yield return new (predecessor.Source);
+		}
+	}
+
+	public class DynamicallyAccessedMembersAnalysis
```
vitek-karas, line 50: TrimAnalysis maybe?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersAnalysis.cs
```diff
@@ -0,0 +1,85 @@
...
+				foreach (var block in ControlFlowGraph.Blocks)
+					yield return new BlockWrapper (block);
+			}
+		}
+
+		public BlockWrapper Entry => new BlockWrapper (ControlFlowGraph.Blocks[0]);
+
+		public IEnumerable<BlockWrapper> GetPredecessors (BlockWrapper block)
+		{
+			foreach (var predecessor in block.Block.Predecessors)
```
vitek-karas, line 45: Why not `return block.Predecessors;`?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersAnalysis.cs
```diff
@@ -0,0 +1,85 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using ILLink.Shared;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.FlowAnalysis;
+
+namespace ILLink.RoslynAnalyzer
```
vitek-karas, line 10: Namespaces should match folders.

Also - can we rename this to TrimAnalysis... instead of DynamicallyAccessedMembers...

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersSymbol.cs
```diff
@@ -0,0 +1,82 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
+using ILLink.Shared;
+using Microsoft.CodeAnalysis;
+
+namespace ILLink.RoslynAnalyzer
+{
+	public class DynamicallyAccessedMembersSymbol : ValueWithDynamicallyAccessedMembers
```
vitek-karas, line 13: I would probably call this something like `SymbolValue` or something similarly simple.
Yes - it can support DAM, but that almost an after thought.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersSymbol.cs
```diff
@@ -0,0 +1,82 @@
...
+				EqualityComparer<ISymbol>.Default.Equals (Source, other.Source) &&
+				EqualityComparer<bool>.Default.Equals (IsMethodReturn, other.IsMethodReturn));
+		}
+
+		public override int GetHashCode ()
+		{
+			return HashUtils.Combine (EqualityContract, SymbolEqualityComparer.Default.GetHashCode (Source), IsMethodReturn);
+		}
+
+#if DEBUG
```
vitek-karas, line 52: Why debug only?
I mean I understand this is for debugging purposes, but I don't see almost any value in leaving it out of Release builds.

vitek-karas:
Some more comments

vitek-karas (commented):
Some more comments - still working through this

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersAnalysis.cs
```diff
@@ -0,0 +1,85 @@
...
+{
```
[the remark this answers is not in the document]
sbomer: Done!

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersSymbol.cs
```diff
@@ -0,0 +1,82 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
+using ILLink.Shared;
+using Microsoft.CodeAnalysis;
+
+namespace ILLink.RoslynAnalyzer
+{
+	public class DynamicallyAccessedMembersSymbol : DynamicallyAccessedMembersValue
```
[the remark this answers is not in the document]
sbomer: Just in that it provides a value (possibly None) for the DAM annotation. But I've renamed it to SymbolValue per @vitek-karas' feedback.

src/ILLink.Shared/DynamicallyAccessedMembers/ValueWithDynamicallyAccessedMembers.cs
```diff
@@ -0,0 +1,12 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace ILLink.Shared
```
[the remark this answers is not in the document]
sbomer: Renamed the folder to TrimAnalysis, and added nested namespaces.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersAnalysis.cs
sbomer: I went with TrimDataFlowAnalysis since it's only one part of the trim analysis.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersSymbol.cs
sbomer: I like SymbolValue - changed.

src/ILLink.Shared/Dataflow/ITransfer.cs
sbomer: Yes, I'd like to keep the lattice abstraction separate because it allows defining the dictionary lattices in terms of another lattices - I personally think it makes the algebraic structure clearer. One simplification that would be really nice is to combine TValue and TValueLattice, by having an interface like ILatticeValue where Top is a static virtual interface method. But AFAIK we can't do that with netstandard2.0.

I don't think we need to support another analysis yet. If we don't think there's value in abstracting this out I can make everything DAM-specific, but one of my original design goals was to make this a reusable framework (or at least for the core pieces to be a reasonable starting point for such a framework). I also think using these abstractions forces thinking in more mathematical terms which I believe is going to help with correctness - but that might just be the way I think. :)

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersAnalysis.cs
sbomer: Thanks, I had changed this in a previous version and missed this. Fixed

src/ILLink.Shared/Dataflow/ForwardDataFlowAnalysis.cs
```diff
@@ -0,0 +1,53 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace ILLink.Shared
+{
+	public abstract class ForwardDataFlowAnalysis<TValue, TLattice, TBlock, TControlFlowGraph, TTransfer>
```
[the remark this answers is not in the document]
sbomer: Sorry about that - added a comment.

src/ILLink.RoslynAnalyzer/LocalDataFlow/LocalDataFlowVisitor.cs
sbomer: Great catch, thank you! PTAL and see if this looks right now. Generics with nullable are confusing... 😖

src/ILLink.Shared/Dataflow/SingleValue.cs
sbomer: Neat, thanks! I changed this to be a trivial record :)

src/ILLink.Shared/Dataflow/ILattice.cs
sbomer: Addressed in https://github.com/dotnet/linker/pull/2360#discussion_r748501628

src/ILLink.RoslynAnalyzer/LocalDataFlow/LocalDataFlowVisitor.cs
sbomer: Addressed in https://github.com/dotnet/linker/pull/2360#discussion_r748501628

src/ILLink.Shared/Dataflow/ITransfer.cs
vitek-karas: I personally don't mind this abstraction. Precisely for the point that it nicely maps to all of the theoretical work in this space.
I wonder if we could use a similar approach (haven't really thought it through) for constant propagation and branch removal.... in the linker/native AOT.

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -0,0 +1,185 @@
...
+					// This should be used as the location of the warning. This is important to provide the right
+					// warning location and because warnings are disambiguated based on the operation.
+					var parentSyntax = branchValueOperation.Syntax.Parent;
+					if (parentSyntax == null)
+						throw new InvalidOperationException ();
+
+					var parentOperation = Context.Compilation.GetSemanticModel (branchValueOperation.Syntax.SyntaxTree).GetOperation (parentSyntax);
+
+					if (parentOperation is IThrowOperation)
+						throw new NotImplementedException ();
```
vitek-karas, line 64: Should we track a TODO for this somewhere?
I don't think we have tests like this in the linker tests.

vitek-karas: Or maybe track the TODO by adding a linker test for this - eventually we will pick it up in analyzer and find this hole :-)

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -0,0 +1,185 @@
...
+						throw new InvalidOperationException ();
+
+					HandleReturnValue (branchValue, returnOperation);
+				}
+			}
+		}
+
+		public abstract void HandleAssignment (TValue source, TValue target, IOperation operation);
+
+		public abstract void HandleReceiverArgument (TValue receiver, IInvocationOperation operation);
```
vitek-karas, line 76: I assume this is some Roslyn term, but I think it might be good to have a comment about what this means.
Reading the code, this is basically a call to a method, right?

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -0,0 +1,141 @@
...
+		// - parameters
+		// - 'this' parameter (for annotated methods)
+		// - field reference
+
+		public override MultiValue VisitInvocation (IInvocationOperation operation, StateValue state)
+		{
+			// Base logic takes care of visiting arguments, etc.
+			base.VisitInvocation (operation, state);
+
+			return new MultiValue (new SymbolValue (operation.TargetMethod, isMethodReturn: true));
```
vitek-karas, line 39: Performance wise: I wonder how is this going to stack up against the current linker implementation.
In the linker we basically ignore values from "unintristing" methods (or actually anything), meaning we return "Unknown" (or "Top"). The values are really only tracked when they are interesting in some way.
But here, we're attaching a symbol to everything, and fully track everything.

vitek-karas: Maybe we should use a factory method to create the SymbolValue and if the value is not interesting in any way, just return a singleton. That would more or less map to what linker is doing currently.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -0,0 +1,141 @@
...
+			if (typeOfOperation.TypeOperand is ITypeParameterSymbol typeParameter)
+				return new MultiValue (new SymbolValue (typeParameter));
+
+			return TopValue;
+		}
+
+		// Override handlers for situations where annotated locations may be involved in reflection access:
+		// - assignments
+		// - arguments passed to method parameters
+		//   this also needs to create the annotated value for parmeters, because they are not represented
```
vitek-karas, line 87: "parameters"

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -0,0 +1,141 @@
...
+		// - instance passed as explicit or implicit receiver to a method invocation
+		//   this also needs to create the annotation for the implicit receiver parameter.
+		// - value returned from a method
+
+		public override void HandleAssignment (MultiValue source, MultiValue target, IOperation operation)
+		{
+			if (target.Equals (TopValue))
+				return;
+
+			TrimAnalysisPatterns.Add (new TrimAnalysisPattern (source, target, operation));
```
vitek-karas, line 98: Perf again here: This will track every assignment, whether it's interesting or not. Ideally we would filter this down to only the cases where there's something interesting happening.

vitek-karas (commented):
More comments... still working on it.

agocke (approved):
LGTM as first pass

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
sbomer: Yup, specifically an instance method call. I added a comment.

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
sbomer: Added a comment

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
```diff
@@ -0,0 +1,82 @@
...
+
+		public TrimDataFlowAnalysis (OperationBlockAnalysisContext context, ControlFlowGraph cfg)
+		{
+			ControlFlowGraph = new ControlFlowGraphProxy (cfg);
+			Lattice = new (new ValueSetLattice<SingleValue> ());
+			Visitor = null;
+			Context = context;
+		}
+
+		public IEnumerable<TrimAnalysisPattern> GetTrimAnalysisPatterns ()
```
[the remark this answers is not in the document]
sbomer: Called it ComputeTrimAnalysisPatterns

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
```diff
@@ -0,0 +1,82 @@
...
+			BlockProxy,
+			ControlFlowGraphProxy,
+			TrimAnalysisVisitor
+		>
+	{
+		readonly ControlFlowGraphProxy ControlFlowGraph;
+
+		readonly LocalStateLattice<ValueSet<SingleValue>, ValueSetLattice<SingleValue>> Lattice;
+
+		TrimAnalysisVisitor? Visitor;
```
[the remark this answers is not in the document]
sbomer: It was initially so that the result could be cached, but I don't think we need that - removed (now stack allocated).

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembers/DynamicallyAccessedMembersAnalysis.cs
```diff
@@ -0,0 +1,85 @@
...
+		public BlockWrapper (BasicBlock block) => Block = block;
+
+		public IEnumerable<BlockWrapper> Predecessors {
+			get {
+				foreach (var predecessor in Block.Predecessors)
+					yield return new BlockWrapper (predecessor.Source);
+			}
+		}
+
+		public bool Equals (BlockWrapper other) => Block.Equals (other.Block);
```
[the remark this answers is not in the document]
sbomer: Added a comment for this.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
sbomer: I left a comment here, to be addressed in a later change

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
sbomer: Added a comment, like above

src/ILLink.Shared/Dataflow/ITransfer.cs
sbomer: For the second suggestion (combining the flow analysis with the transfer functions under an `IDataFlowAnalysis` or similar): as discussed, we can't use default interface implementations so this would prevent the current approach where the core dataflow analysis is generic. Alternatively, it would require the visitor state to be stored together with the class implementing the dataflow analysis, but this state doesn't need to live as long as the analysis results. For now I've left them separate but it's easy enough to change if we want to choose a different tradeoff later.

sbomer:
Thanks for all the feedback!

PR #2362, opened 2021-11-11
Move analyzer tests to one test per file

agocke:
Moves away from generating Theory cases for each method in the file and
uses one Fact per test file.


test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -38,49 +40,27 @@ public async static ValueTask<ImmutableArray<MetadataReference>> GetNet6Referenc
...
+					msbuildProperties,
+					additionalSources: testDependenciesSource);
+			foreach (var testCase in BuildTestCasesForFile (tree)) {
+				testCase.Run (comp);
 			}
-			return testCases;
 		}
 
-		private static IEnumerable<TestCase> BuildTestCasesForFile (string testFile)
+		private static IEnumerable<TestCase> BuildTestCasesForFile (SyntaxTree tree)
```
sbomer, line 61: Maybe leave a note that there's still one `TestCase` struct here per member and rename this to something like `BuildMemberTestCasesForTree`?

test/ILLink.RoslynAnalyzer.Tests/LinkerTestCases.cs
```diff
@@ -1,29 +0,0 @@
...
-		[Theory]
-		[MemberData (nameof (TestCaseUtils.GetTestData), parameters: nameof (RequiresCapability))]
-		public void RequiresCapability (string m)
-		{
-			RunTest (nameof (RequiresCapability), m, UseMSBuildProperties (MSBuildPropertyOptionNames.EnableTrimAnalyzer, MSBuildPropertyOptionNames.EnableSingleFileAnalyzer));
-		}
-
-		[Theory]
-		[MemberData (nameof (TestCaseUtils.GetTestData), parameters: nameof (Interop))]
-		public void Interop (string m)
```
sbomer, line 24: I don't think we want to remove the interop tests, do we? Also these are an exception to normal directory structure - the testcases are defined in nested directories so we should make sure that still works.

vitek-karas (approved):
The Interop tests should be kept. Otherwise looks good.

PR #2363, opened 2021-11-11
[DAM analyzer] Separate known type case

sbomer:
This tracks a bit of extra information so that when we check annotations we can tell whether the source is a known type - no functional changes.

This isn't the final solution (https://github.com/dotnet/linker/pull/2360 is adding a better abstraction for this) but is meant to help unblock https://github.com/dotnet/linker/issues/2339 since @tlakollo ran into unrelated warnings when trying to enable binding logic for a known type flowing into an annotated location. The not-supported case for a known type would be the right place to call into the DAM binding logic.


tlakollo (approved):
Looks good
Thanks! :D

PR #2370, opened 2021-11-15
Add a custom flag to disable marking of copy assemblies

vitek-karas:
For https://github.com/dotnet/linker/issues/2089

Enabled by `--custom-data DisableMarkingOfCopyAssemblies=true` on the command line. Assumes that ALL assemblies on the input are in "copy" action (it doesn't validate this fact).

It disables marking basically fully - linker will go over all assemblies, and process them in "copy" mode (copy the original file over) and will call all the custom steps and so on, but it will do no marking (or very little, depends on descriptors and such which this doesn't disable).

This is intentionally non-discoverable feature, to be used only by the mono AOT toolchain.

/cc @rolfbjarne

### DIFF
--- a/src/linker/Properties/launchSettings.json
+++ b/src/linker/Properties/launchSettings.json
@@ -1,9 +1,0 @@
-{
-  "profiles": {
-    "Mono.Linker": {
-      "commandName": "Project",
-      "commandLineArgs": "@F:\\ILLinker\\repro\\SlowApp\\linker.rsp",
-      "workingDirectory": "F:\\ILLinker\\repro\\SlowApp"
-    }
-  }
-}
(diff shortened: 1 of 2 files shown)


vitek-karas:
@agocke for triage - this should go into 6.0.2xx and maybe potentially even sooner depending on mono needs.

src/linker/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
```
sbomer, line 1: Left-over file I think

sbomer (approved):
LGTM

rolfbjarne (approved):

vitek-karas:
@agocke - can you please confirm that it's OK to merge this to 6.0.2xx? (since it's kind of the first real linker fix for servicing).

agocke:
LGTM

Issue #2359, opened 2021-11-09
Linker doesn't preserve type forwarder for a nested type in a second type forwarder


vitek-karas:
Repro:

app.csproj
```XML
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net6.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <TrimMode>copyused</TrimMode>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Azure.Identity" Version="1.5.0" />
    <PackageReference Include="Azure.ResourceManager.Resources" Version="1.0.0-beta.2" />
  </ItemGroup>
</Project>
```

app.cs
```C#
using Azure.ResourceManager;

typeof(ArmClientOptions).Assembly.GetTypes();
```

Publish the app with trimming enabled and run it. It fails with `Could not load type 'Enumerator' from assembly 'Microsoft.Bcl.AsyncInterfaces, Version=1.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'`

The reason is that `Azure.ResourceManager.dll` has a type reference to a `System.Runtime.CompilerServices.ConfiguredCancelableAsyncEnumerable``1/Enumerator` in `Microsoft.Bcl.AsyncInterfaces`. This assembly has a type forwarder for this nested type which points to `netstandard`. And `netstandard` has a type forward to corelib.

After trimming only the type forwarder in `Microsoft.Bcl.AsyncInterfaces` is kept. The one in `netstandard` is removed. Note that the parent type is kept in both cases, it's only the nested type forwarder which is removed.

Also notes that this runs the linker in copyused mode.

vitek-karas:
The actual problem is that setup like this, the project trims the assemblies as a mix of "copy" and "copyused" actions. This is bound to break. "copyused" makes the assumption that all type refs are rewritten to point to implementation assembly (as it doesn't preserve type forwarders by default and it doesn't mark them). On the other hand "copy" can't rewrite the type refs.

The workaround is to set `TrimmerDefaultAction=copyused`.

marek-safar:
This looks to me like something worth servicing fix, thought?

vitek-karas:
I don't think this is worth servicing. I honestly even question if we should fix it.

[...]

PR #2372, opened 2021-11-16
Adds a test for multiple nested forwarders in a row with copy/link

vitek-karas:
The problematic scenario is if there's a "copy" assembly with a forwarder  to a "link" assembly with a forwarder. The forwarder is for a nested class.
Test for #2359.
The reason for "Random" behavior is a Cecil issue: https://github.com/jbevain/cecil/issues/812.

Also improved the test infra to print out errors from ilasm if they happen.


marek-safar (approved):

sbomer (approved):
Really interesting find!

PR #2381, opened 2021-11-17
Switch to 1ES servicing pools on release/6.0.2xx

jonfortescue:
Servicing branches need to use servicing pools so this does that.


jonfortescue:
@marek-safar I think this needs approval in order to merge

akoeplinger (approved):

PR #2390, opened 2021-11-20
[DAM analyzer] Add support for properties and enable more tests

sbomer:
I enabled more of the dataflow tests for the analyzer and fixed one common bucket of failures due to properties not being supported in the analyzer. The IOperation tree represents properties as IPropertyReferenceOperation so we need to handle properties specially (whereas the linker sees calls to the getter/setter directly).


src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -118,15 +132,22 @@ public override void HandleArgument (MultiValue argumentValue, IArgumentOperatio
...
+		// where the receiver is not null (so an instance method/property).
+		public override void HandleReceiverArgument (MultiValue receiverValue, IMethodSymbol targetMethod, IOperation operation)
+		{
+			MultiValue implicitReceiverParameter = new MultiValue (new SymbolValue (targetMethod!, isMethodReturn: false));
 
 			TrimAnalysisPatterns.Add (new TrimAnalysisPattern (
-				receieverValue,
+				receiverValue,
 				implicitReceiverParameter,
 				operation
```
vitek-karas, line 152: I assume this is temporary - it doesn't make much sense to wrap the method in a MultiValue and pass it as something called `implicitReceiverParameter`. The value of the implicit receiver parameter I think is the `this`, which doesn't "match" the method. That said the pattern should record which method is being called...

test/Mono.Linker.Tests.Cases/DataFlow/ByRefDataflow.cs
```diff
@@ -40,14 +40,18 @@ public static void Main ()
 		static Type s_typeWithPublicParameterlessConstructor;
 
 		[Kept]
-		[UnrecognizedReflectionAccessPattern (typeof (ByRefDataflow), nameof (MethodWithRefParameter), new string[] { "Type&" }, messageCode: "IL2077")]
+		// Trimmer and analyzer use different formats for ref parameters
+		[ExpectedWarning ("IL2077", nameof (ByRefDataflow) + "." + nameof (MethodWithRefParameter) + "(Type&)", ProducedBy = ProducedBy.Trimmer)]
+		[ExpectedWarning ("IL2077", nameof (ByRefDataflow) + "." + nameof (MethodWithRefParameter) + "(ref Type)", ProducedBy = ProducedBy.Analyzer)]
```
vitek-karas, line 45: Could you add an issue for this. My understand is that we tried to match the C# formatting in warnings, so if there's a discrepancy, it should be treated as a bug in the linker.

test/Mono.Linker.Tests.Cases/DataFlow/ByRefDataflow.cs
```diff
@@ -40,14 +40,18 @@ public static void Main ()
...
 		public static void PassRefToField ()
 		{
 			MethodWithRefParameter (ref s_typeWithPublicParameterlessConstructor);
```
vitek-karas, line 48: Just a general note - I know that linker has issues with ref/out parameters in some cases. We should add a work item to the analyzer about the same.
Basically it reverses data flow direction to a degree....

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/DataFlow/ByRefDataflow.cs
sbomer: I think this is the issue you're referring to, right? https://github.com/dotnet/linker/issues/2158
I'll link to it from the analyzer tracking issue.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
sbomer: I renamed it to `thisParameter` - does that make more sense? The intention is to create a SingleValue representing the this parameter of the method. It's wrapped in a MultiValue because the pattern cache stores source/target as MultiValue, even though they will only contain single values in a lot of cases.

test/Mono.Linker.Tests.Cases/DataFlow/ByRefDataflow.cs
sbomer: Filed https://github.com/dotnet/linker/issues/2406

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
vitek-karas: I think for now this is fine. Going forward I think the plan was to change these classes to match the various shared operations and thus this code will change anyway.

vitek-karas (approved):

PR #2393, opened 2021-11-23
Use non-deprecated Windows pool in Azure Pipelines

akoeplinger:
See https://github.com/dotnet/core-eng/issues/14783.

Also switch to latest macOS/Ubuntu/Windows vmImage.

### DIFF
--- a/eng/azure-pipelines.yml
+++ b/eng/azure-pipelines.yml
@@ -53,7 +53,7 @@ stages:
       - job: Windows_NT
         pool:
           ${{ if eq(variables.officialBuild, 'false') }}:
-            vmImage: windows-2019
+            vmImage: windows-latest
           ${{ if eq(variables.officialBuild, 'true') }}:
             name: NetCore1ESPool-Internal
             demands: ImageOverride -equals Build.Windows.10.Amd64.VS2019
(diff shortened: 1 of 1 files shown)


vitek-karas:
Why can't we stay on the generic macOS image? I was hoping that would not have to deal with version deprecation if it's that way.

akoeplinger:
@vitek-karas that's fine too if you want to take the risk of Azure Pipelines updating the macOS behind your back. In other repos we usually make it explicit and you're already using an explicit `ubuntu-20.04` for example :)

vitek-karas:
@akoeplinger Just a personal opinion (I'll let Sven and Marek comment on this):
* Linker doesn't have platform specific dependencies, so the Linux/macOS legs in CI are mostly to validate that it works cross-plat in general (like correct path handling and such)
* Linker is consumed by several other repos which may upgrade to later versions of OSes any time, I would much prefer we catch any issues in this repo's CI, then waiting on the integration to other repos

As such I would prefer if we ran CI on the latest version of each OS. We seem to already do so for mac, we should do so for linux as well (and windows if that's possible).

sbomer:
Thanks!

Agreed, I would prefer to be on latest unless we have a good reason not to. It looks like they changed the recommended names so I would go with `macOS-latest`, `windows-latest`, and `ubuntu-latest`.

akoeplinger:
@sbomer done :)

sbomer (approved):

vitek-karas (approved):

akoeplinger:
@sbomer would you mind porting this to any applicable branches?

vitek-karas:
feature/damAnalyzer - https://github.com/dotnet/linker/pull/2412
release/6.0.2xx - https://github.com/dotnet/linker/pull/2413

Do we need anything else?

sbomer:
Thanks @vitek-karas I think those branches are enough. Not sure we need 5.0 and it's easy enough to do later if we need to spin a build.

Issue #2339, opened 2021-10-27
[DAM analyzer] Port DAM binding logic to analyzer


sbomer:
The Roslyn analyzer needs to be able to produce warnings for reflection-access to reflection-annotated methods. Whenever we see a reflection pattern that accesses, for example, `PublicMethods` on a particular type, we need to be able to find all public methods on the type.

The binding logic from https://github.com/dotnet/linker/blob/main/src/linker/Linker.Dataflow/DynamicallyAccessedMembersBinder.cs should be ported to work with Roslyn, and made to produce these warnings. To do this, I would start by enabling a few tests from https://github.com/dotnet/linker/blob/main/test/Mono.Linker.Tests.Cases/DataFlow/AnnotatedMembersAccessedViaReflection.cs in the `feature/damAnalyzer` branch (which has early support for DynamicallyAccessedMembersAttribute).

sbomer:
The binding logic has been ported, so the remaining work is to turn it on and produce warnings for reflection access to annotated members (https://github.com/dotnet/linker/pull/2418).

tlakollo:
This should be working now, closing this issue

PR #2394, opened 2021-11-24
DynamicallyAccessedMembersBinder port to Analyzer

LakshanF:
Mostly 1:1 conversion of linker DAM to Analyzer, #2339

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
 using Microsoft.CodeAnalysis;
@@ -12,13 +13,6 @@ namespace ILLink.RoslynAnalyzer.DataFlow
 {
 	internal static class DynamicallyAccessedMembersBinder
 	{
-		// Temporary workaround - should be removed once linker can be upgraded to build against
-		// high enough version of the framework which has this enum value.
-		internal static class DynamicallyAccessedMemberTypesOverlay
-		{
-			public const DynamicallyAccessedMemberTypes Interfaces = (DynamicallyAccessedMemberTypes) 0x2000;
-		}
-
 		// Returns the members of the type bound by memberTypes. For DynamicallyAccessedMemberTypes.All, this returns all members of the type and its
 		// nested types, including interface implementations, plus the same or any base types or implemented interfaces.
 		// DynamicallyAccessedMemberTypes.PublicNestedTypes and NonPublicNestedTypes do the same for members of the selected nested types.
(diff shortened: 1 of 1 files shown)


src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,438 @@
...
+namespace ILLink.RoslynAnalyzer.DataFlow
+{
+	internal static class DynamicallyAccessedMembersBinder
+	{
+		// Temporary workaround - should be removed once linker can be upgraded to build against
+		// high enough version of the framework which has this enum value.
+		internal static class DynamicallyAccessedMemberTypesOverlay
+		{
+			public const DynamicallyAccessedMemberTypes Interfaces = (DynamicallyAccessedMemberTypes) 0x2000;
+		}
```
tlakollo, line 20: I see in the metadata that linker is already using a version with interfaces so this workaround is not necessary

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,438 @@
...
+
+				if ((bindingFlags & (BindingFlags.Public | BindingFlags.NonPublic)) == BindingFlags.NonPublic && method.DeclaredAccessibility == Accessibility.Public)
+					continue;
+
+				yield return method;
+			}
+		}
+
+		public static IEnumerable<IMethodSymbol> GetMethodsOnTypeHierarchy (this ITypeSymbol thisType, OperationAnalysisContext context, Func<IMethodSymbol, bool>? filter, BindingFlags? bindingFlags = null)
+		{
```
tlakollo, line 148: Code is complaining that we don't use the context for these methods, linker used it to calculate the base type the analyzer simply doesn't need it.

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,438 @@
...
+			//		if (!nestedType.IsNestedPublic)
+			//			continue;
+			//	}
+
+			//	if ((bindingFlags & (BindingFlags.Public | BindingFlags.NonPublic)) == BindingFlags.NonPublic) {
+			//		if (nestedType.IsNestedPublic)
+			//			continue;
+			//	}
+
+			//	yield return nestedType;
```
tlakollo, line 255: I think what we want is to use [GetTypeMembers](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.inamespaceortypesymbol.gettypemembers?view=roslyn-dotnet-4.0.1) which returns an enumeration of the type inside the symbol.

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
LakshanF: Thanks!

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
LakshanF: Thanks

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
LakshanF: Thanks!

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,435 @@
...
+						var interfaceType = iface;
+						if (interfaceType == null)
+							continue;
+						GetAllOnType (interfaceType, context, declaredOnly: false, members, types);
+					}
+				}
+			}
+
+			foreach (var member in type.GetMembers ()) {
+				if (member is IFieldSymbol f)
```
sbomer, line 422: I would replace the ifs with a switch. I think you can also use member.Kind which might (?) be a bit faster than the type check, here and elsewhere.

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,435 @@
...
+				}
+
+				if (declaredOnly)
+					yield break;
+
+				type = type.BaseType;
+			}
+		}
+
+		// Issue https://github.com/dotnet/roslyn-analyzers/issues/4845
```
sbomer, line 379: I don't think this issue applies here - I believe the warning is asking to pass a symbol equality comparer to the HashSet ctor.

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
```diff
@@ -0,0 +1,435 @@
...
+		public static void GetAllOnType (this ITypeSymbol type, OperationAnalysisContext context, bool declaredOnly, List<ISymbol> members) => GetAllOnType (type, context, declaredOnly, members, new HashSet<ITypeSymbol> ());
+#pragma warning restore RS1024
+
+		static void GetAllOnType (ITypeSymbol type, OperationAnalysisContext context, bool declaredOnly, List<ISymbol> members, HashSet<ITypeSymbol> types)
+		{
+			if (!types.Add (type))
+				return;
+
+			// @TODO
+			//if (type.HasNestedTypes) {
```
sbomer, line 392: Left-over TODO. I think it would be fine to do nested types as part of the loop over all members below - I don't think we rely on the order of the results.

sbomer (commented):
Thanks, the port is looking good! I think we need to turn on at least a few tests from AnnotatedMembersAccessedViaReflection as a sanity check. Would you be able to do the following?
- Add a new subclass of SingleValue, maybe `KnownTypeValue`, alongside `ValueWithDynamicallyAccessedMembers`. Its ctor should take an `ITypeSymbol`
- In the visitor, modify `VisitTypeOf` in the visitor to return a `KnownTypeValue`
- Modify `GetDynamicallyAccessedMembersDiagnostics` to call the binder and report warnings for members which have annotations
  - Maybe it doesn't need to be complete in this change - for example it could be enough to report warnings for members which have RUC - the goal is just to get a few of the tests running as validation.

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
LakshanF: I replaced this with a switch and changed the if patterns as suggested by the IDE code fixes. But I didn't change all the if statements primarily to make it easier to see the 1:1 port with the linker.

src/ILLink.RoslynAnalyzer/DataFlow/DynamicallyAccessedMembersBinder.cs
LakshanF: I left the suppression since `SymbolEqualityComparer `seems to apply to `ISymbol `but clarified this in the comments

LakshanF:
I opened an issue, #2410 to track the test suggestion and assigned to @tlakollo since he has already started to work on enabling tests

tlakollo (approved):
Looks good to me, thanks Lakshan :D

sbomer (approved):

PR #2395, opened 2021-11-24
Initial support for RequiresDynamicCodeAttribute analyzer

tlakollo:
Add analyzer and code fixer for the attribute
Add a placeholder for the attribute in metadata in linker and as an additional resource in the analyzer.
Add attributes in RequiresCapability for testing
Add tests for code fixer

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
+++ b/src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
@@ -59,6 +59,8 @@ namespace ILLink.RoslynAnalyzer
 		static string GetDiagnosticCategory (DiagnosticId diagnosticId)
 		{
 			switch ((int) diagnosticId) {
+			case 2117:
+				return DiagnosticCategory.AOT;
 			case > 2000 and < 3000:
 				return DiagnosticCategory.Trimming;
 
--- a/src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
+++ b/src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
@@ -59,13 +59,14 @@ namespace ILLink.RoslynAnalyzer
 		static string GetDiagnosticCategory (DiagnosticId diagnosticId)
 		{
 			switch ((int) diagnosticId) {
-			case 2117:
-				return DiagnosticCategory.AOT;
 			case > 2000 and < 3000:
 				return DiagnosticCategory.Trimming;
 
-			case >= 3000 and <= 6000:
+			case >= 3000 and < 3050:
 				return DiagnosticCategory.SingleFile;
+
+			case >= 3050 and <= 6000:
+				return DiagnosticCategory.AOT;
 
 			default:
 				break;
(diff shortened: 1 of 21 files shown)


src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
```diff
@@ -59,6 +59,8 @@ public static DiagnosticDescriptor GetDiagnosticDescriptor (DiagnosticId diagnos
 		static string GetDiagnosticCategory (DiagnosticId diagnosticId)
 		{
 			switch ((int) diagnosticId) {
+			case 2117:
```
vitek-karas, line 62: I think we should move this to the 3000+ range.
Maybe something like 3050 or so.

vitek-karas: Note that below you also have 2118 which is not even covered by this switch.

src/ILLink.Shared/SharedStrings.resx
```diff
@@ -141,12 +141,24 @@
   <data name="RequiresAssemblyFilesMessage" xml:space="preserve">
     <value>Using member '{0}' which has 'RequiresAssemblyFilesAttribute' can break functionality when embedded in a single-file app.{1}{2}</value>
   </data>
+ <data name="RequiresDynamicCodeTitle" xml:space="preserve">
+    <value>Members annotated with 'RequiresDynamicCodeAttribute' require dynamic access otherwise can break functionality when trimming application code</value>
```
vitek-karas, line 145:
```suggestion
    <value>Members annotated with 'RequiresDynamicCodeAttribute' can break functionality when compiling the application ahead of time.</value>
```

Or something like that - the message seems to be more about trimming, which is not the problem in this case. I think we will need somebody like Andy to review this message.

src/ILLink.Shared/SharedStrings.resx
```diff
@@ -141,12 +141,24 @@
...
+  </data>
+  <data name="RequiresDynamicCodeMessage" xml:space="preserve">
+    <value>Using member '{0}' which has 'RequiresDynamicCodeAttribute' can break functionality when trimming application code.{1}{2}</value>
```
vitek-karas, line 148: Same as above - this is not about trimming.

marek-safar:
Why is this analyzer here? It has nothing to do with trimming and we have all other libraries analyzers elsewhere.

tlakollo:
> Why is this analyzer here? It has nothing to do with trimming and we have all other libraries analyzers elsewhere.

From a reuse standpoint is convenient to have it here, we can reuse the RequiresAnalyzerBase and linker testing. Native AOT will have access to the package anyway

vitek-karas:
It's true that logically it doesn't belong here - just like the single-file analyzer. But technically it's MUCH easier. The `Requires*` attributes are all designed to behave the same way, they basically just produce different warning codes, otherwise their behavior is almost identical. So the implementation is like 95% shared, same goes for tests.

We could try to somehow share the source files across repos, but it would be much more complex to do that.

src/ILLink.Shared/SharedStrings.resx
MichalStrehovsky: Yes, I would like Andy to have a look at this as well so that we don't need to go back and forth about this later.

MichalStrehovsky (commented):
Thank you for looking into this!

src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
tlakollo: Changed to 3050 and 3051

agocke (approved):

tlakollo:
Opened https://github.com/dotnet/linker/issues/2421 to track the issue of where the analyzers should be located

PR #2396, opened 2021-11-25
Fixes a null-ref/NotSupportedException in warning infra

vitek-karas:
If we end up marking a type due to descriptor and the type marking will cause a warning (for example a removed attribute type does this in some cases), the warning infra would fail.

This is because it requires all warnings to have a valid origin including which assembly it came from. But descriptors didn't provide this information.

Most of the change here are to fix that lack of information. When marking through annotations, we now store the message origin for the mark operation along with the target of the marking. This is then restored on the scope stack to be used when reporting warnings caused by such marking.

This required lot of changes to pass the message origin in all places which end up marking things in annotations.

This also required some small fixes as it uncovered previously unseen issues.

Additionally, it can still happen that there's really no origin (for example copy action specified on command line will mark entire assembly - there's nothing to blame, but the command line) for a message or we can't assign an assembly (descriptor file pass on the command line, we can blame the file, but we still don't have an assembly).

So this change tweaks the warning infra to allow messages without assembly of origin.

Enabled a test whcih validated the behavior around removed attribute being marked via descriptor (this now produces warning with fully valid origin).
Added another test where the warning is due to a descriptor on command line (no associated assembly for such warning).

This fixes the linker issue found in https://github.com/dotnet/linker/issues/2358.

### DIFF
--- a/src/linker/Linker/Annotations.cs
+++ b/src/linker/Linker/Annotations.cs
@@ -49,7 +49,7 @@ namespace Mono.Linker
 		// Annotations.Mark will add unmarked items to marked_pending, to be fully marked later ("processed") by MarkStep.
 		// Items go through state changes from "unmarked" -> "pending" -> "processed". "pending" items are only tracked
 		// once, and once "processed", an item never becomes "pending" again.
-		protected readonly HashSet<IMetadataTokenProvider> marked_pending = new HashSet<IMetadataTokenProvider> ();
+		protected readonly Dictionary<IMetadataTokenProvider, MessageOrigin> marked_pending = new Dictionary<IMetadataTokenProvider, MessageOrigin> ();
 		protected readonly HashSet<IMetadataTokenProvider> processed = new HashSet<IMetadataTokenProvider> ();
 		protected readonly Dictionary<TypeDefinition, (TypePreserve preserve, bool applied)> preserved_types = new Dictionary<TypeDefinition, (TypePreserve, bool)> ();
 		protected readonly HashSet<TypeDefinition> pending_preserve = new HashSet<TypeDefinition> ();
@@ -161,15 +161,16 @@ namespace Mono.Linker
 		[Obsolete ("Mark token providers with a reason instead.")]
 		public void Mark (IMetadataTokenProvider provider)
 		{
+			// No origin provided, so use the provider itself if possible
 			if (!processed.Contains (provider))
-				marked_pending.Add (provider);
+				marked_pending.TryAdd (provider, new MessageOrigin (provider as ICustomAttributeProvider));
 		}
 
-		public void Mark (IMetadataTokenProvider provider, in DependencyInfo reason)
+		public void Mark (IMetadataTokenProvider provider, in DependencyInfo reason, in MessageOrigin origin)
 		{
 			Debug.Assert (!(reason.Kind == DependencyKind.AlreadyMarked));
 			if (!processed.Contains (provider))
-				marked_pending.Add (provider);
+				marked_pending.TryAdd (provider, origin); // It's OK if it already exists, one origin is enough to remember
 			Tracer.AddDirectDependency (provider, reason, marked: true);
 		}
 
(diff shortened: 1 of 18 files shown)


src/linker/Linker/Annotations.cs
```diff
@@ -161,15 +161,16 @@ public bool HasSubstitutedInit (TypeDefinition type)
 		[Obsolete ("Mark token providers with a reason instead.")]
 		public void Mark (IMetadataTokenProvider provider)
 		{
+			// No origin provided, so use the provider itself if possible
 			if (!processed.Contains (provider))
-				marked_pending.Add (provider);
+				marked_pending.TryAdd (provider, new MessageOrigin (provider as ICustomAttributeProvider));
```
marek-safar, line 166: Why Contains with TryAdd ?

src/linker/Linker/Annotations.cs
```diff
@@ -161,15 +161,16 @@ public bool HasSubstitutedInit (TypeDefinition type)
...
 		}
 
-		public void Mark (IMetadataTokenProvider provider, in DependencyInfo reason)
+		public void Mark (IMetadataTokenProvider provider, in DependencyInfo reason, in MessageOrigin origin)
 		{
 			Debug.Assert (!(reason.Kind == DependencyKind.AlreadyMarked));
 			if (!processed.Contains (provider))
-				marked_pending.Add (provider);
+				marked_pending.TryAdd (provider, origin); // It's OK if it already exists, one origin is enough to remember
```
marek-safar, line 173: Why Contains with TryAdd ?

src/linker/Linker/MessageOrigin.cs
```diff
@@ -39,11 +39,18 @@ public MessageOrigin (ICustomAttributeProvider? provider)
 		}
 
 		public MessageOrigin (string fileName, int sourceLine = 0, int sourceColumn = 0)
+			: this (fileName, sourceLine, sourceColumn, null)
+		{
+		}
+
+		// The assembly attribute should be specified if available as it allows assigning the diagnostic
+		// to a an assembly (we group based on assembly).
+		public MessageOrigin (string fileName, int sourceLine, int sourceColumn, AssemblyDefinition? assembly)
```
marek-safar, line 48: Can you make AssemblyDefinition non-nullable? I think in this advanced ctor it would make more sense.

src/linker/Linker/Annotations.cs
vitek-karas: The `Contains` is on a different set!

src/linker/Linker/Annotations.cs
vitek-karas: The `Contains` is on a different set!

src/linker/Linker/Annotations.cs
marek-safar: oh, missed that

src/linker/Linker/MessageOrigin.cs
vitek-karas: Unfortunately not. The case where the `null` happens here is command line specified descriptor xml. Warnings from it have file name, and line/col numbers, but no assembly. It's a valid case and the code actually handles it. The "field" test I added tests for exactly that case.

src/linker/Linker/Annotations.cs
sbomer, line 166: Why'd this change to `TryAdd` - was `Add` causing problems? If so I wonder if it might be hiding a bug.

src/linker/Linker/CustomAttributeSource.cs
```diff
@@ -40,8 +40,13 @@ public bool TryGetEmbeddedXmlInfo (ICustomAttributeProvider provider, [NotNullWh
 			var assembly = GetAssemblyFromCustomAttributeProvider (provider);
 
 			if (!_embeddedXmlInfos.TryGetValue (assembly, out xmlInfo)) {
+				// Add an empty record - this prevents reentrancy
+				// If the embedded XML itself generates warnings, trying to log such warning
+				// may ask for attributes (suppressions) and thus we could end up in this very place again
+				// So first add a dummy record and once processed we will replace it with the real data
```
sbomer, line 46: Makes sense, but this scares me 😨

sbomer (approved):
LGTM, thank you!

src/linker/Linker/Annotations.cs
vitek-karas: `Add` throws if the value is already in the dictionary (unlike `HashSet`). So the `TryAdd` is there to allow calling `Mark` on the same thing multiple times (which does happen quite a bit).
I don't think it's a bug, maybe inefficiency... hard to tell.

src/linker/Linker/CustomAttributeSource.cs
vitek-karas: Yeah - I'm not happy about it either. I did go through the code which can run here, and there's only very few places where we generate warnings, so it's not that bad. Maybe we should think about this some more how to solve - but this is hard, we want to be lazy about the suppressions (too expensive to do eagerly).

src/linker/Linker/Annotations.cs
sbomer: Sorry, I forgot that `HashSet.Add` doesn't throw. Makes sense.

PR #2397, opened 2021-11-25
Improve ValueSet implementation to avoid HashSet creation every time

vitek-karas:
The idea is simple, store only a single object field.
If the field is null, the set is empty.
If the field is a HashSet, then it contains multiple values in that hashset.
If the field is TValue, then it set contains single value - that value.

I probably went a bit overboard on optimizations here, but I wanted to try what it takes to implement a good collection class. The result should not allocate at all when "Foreached" over if it contains no or single value (by far the most common case).

~~Unfortunately it still has to box the value. Maybe we should store two fields to avoid boxing by adding a cost of a single pointer field.~~
I just realized that all our uses cases will store reference types as values (`SingleValue` is a record class, `ValueNode` in the linker currently is just a normal class). This means that when a single value is stored the ValueSet will have the same memory and GC footprint as a direct reference to the value.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
@@ -84,10 +84,7 @@ namespace ILLink.RoslynAnalyzer
 
 		static IEnumerable<Diagnostic> GetDynamicallyAccessedMembersDiagnostics (ValueSet<SingleValue> source, ValueSet<SingleValue> target, Location location)
 		{
-			if (target.Values == null)
-				yield break;
-
-			foreach (var targetValue in target.Values) {
+			foreach (var targetValue in target) {
 				foreach (var diagnostic in GetDynamicallyAccessedMembersDiagnostics (source, targetValue, location))
 					yield return diagnostic;
 			}
@@ -95,10 +92,7 @@ namespace ILLink.RoslynAnalyzer
 
 		static IEnumerable<Diagnostic> GetDynamicallyAccessedMembersDiagnostics (ValueSet<SingleValue> source, SingleValue target, Location location)
 		{
-			if (source.Values == null)
-				yield break;
-
-			foreach (var sourceValue in source.Values) {
+			foreach (var sourceValue in source) {
 				foreach (var diagnostic in GetDynamicallyAccessedMembersDiagnostics (sourceValue, target, location))
 					yield return diagnostic;
 			}
--- a/src/ILLink.Shared/DataFlow/ValueSet.cs
+++ b/src/ILLink.Shared/DataFlow/ValueSet.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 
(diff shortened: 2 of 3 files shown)


src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -84,21 +84,15 @@ static void ProcessTypeArguments (OperationAnalysisContext context, IInvocationO
 
 		static IEnumerable<Diagnostic> GetDynamicallyAccessedMembersDiagnostics (ValueSet<SingleValue> source, ValueSet<SingleValue> target, Location location)
 		{
-			if (target.Values == null)
-				yield break;
-
-			foreach (var targetValue in target.Values) {
+			foreach (var targetValue in target) {
```
Therzok, line 87: Nit: SelectMany might be more optimized than the generated state machine for foreach.

src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -9,49 +10,151 @@ namespace ILLink.Shared.DataFlow
...
+		{
+			readonly object? _value;
+			int _state;  // 0 before begining, 1 at item, 2 after end
+			readonly IEnumerator<TValue>? _enumerator;
+
+			internal Enumerator (object? values)
+			{
+				_state = 0;
+				if (values is SealedHashSet valuesSet) {
+					_enumerator = valuesSet.GetEnumerator ();
```
Therzok, line 31: Won't this box the enumerator?

vitek-karas: I didn't try to optimize the case where it falls back to the hashset. Based on measurements in illink, that case is super rare (for a simple ASP.NET app out of 8200 values inspected, only 100 has more than value in them - those would fall back to hashset)

src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -9,49 +10,151 @@ namespace ILLink.Shared.DataFlow
 	public readonly struct ValueSet<TValue> : IEquatable<ValueSet<TValue>>, IEnumerable<TValue>
 		where TValue : notnull
 	{
-		public readonly HashSet<TValue>? Values;
+		// Since we're going to do lot of type checks for this class a lot, it is much more efficient
+		// if the class is sealed (as then the runtime can do a simple method table pointer comparison)
+		class SealedHashSet : HashSet<TValue>
```
marek-safar, line 15: Why not call it EnumerableValues to make it clear what it does?
I'd also seal the class

src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -9,49 +10,151 @@ namespace ILLink.Shared.DataFlow
...
+		{
+			public SealedHashSet () { }
```
marek-safar, line 17: Why is this ctor needed?

src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -9,49 +10,151 @@ namespace ILLink.Shared.DataFlow
...
+			if (_values == null)
 				return typeof (ValueSet<TValue>).GetHashCode ();
 
-			int hashCode = 0;
-			foreach (var item in Values)
-				hashCode = HashUtils.Combine (hashCode, item);
-			return hashCode;
-		}
+			if (_values is SealedHashSet valuesSet) {
+				int hashCode = 0;
```
marek-safar, line 119: I'd move the code to SealedHashSet

src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -9,49 +13,161 @@ namespace ILLink.Shared.DataFlow
...
+					_state = 0;
+			}
+		}
 
-		public ValueSet (TValue value) => Values = new HashSet<TValue> () { value };
+		// This stores the values. By far the most common case will be either no values, or a single value.
+		// Cases where there are multiple values stored are relatively very rare.
+		//   null - no values (empty set)
+		//   TValue - single value itself
+		//   SealedHashSet typed object - multiple values, stored in the hashset
```
sbomer, line 90:
```suggestion
		//   EnumerableValues typed object - multiple values, stored in the hashset
```

src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -9,49 +13,161 @@ namespace ILLink.Shared.DataFlow
...
+					_value = default (TValue);
+				} else {
+					_enumerator = null;
+					_value = values;
+				}
+			}
+
+			// TODO: How to get this to work - without the '!' at the end this complains that default can return null.
+			// But how does this work for HashSet<T>? That will return null from Current in reality as well... 
+			// It seems that the nullability is not propagated "inside" HashSet (since that one is implemented with TValue being nullable)
```
sbomer, line 51: Maybe the difference is that for `HashSet<T>`, the `T` isn't constrained to `notnull` like it is here.

src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -9,49 +13,161 @@ namespace ILLink.Shared.DataFlow
...
+			readonly object? _value;
+			int _state;  // 0 before begining, 1 at item, 2 after end
+			readonly IEnumerator<TValue>? _enumerator;
+
+			internal Enumerator (object? values)
+			{
+				_state = 0;
+				if (values is EnumerableValues valuesSet) {
+					_enumerator = valuesSet.GetEnumerator ();
+					_value = default (TValue);
```
sbomer, line 42: Should this be `null`? Alternatively, would it be possible to store the `_enumerator` in the same `object?` field?

sbomer (approved):
Very nice, thank you!

src/ILLink.Shared/DataFlow/ValueSet.cs
vitek-karas: Yes - that is the difference. Hashset implements this without the `!` obviously. But if I use Hashset with a T which is constrained, like this:
```C#
class NonNullHashSet<T> : HashSet<T> where T : notnull { }
```

There's no warning generated by the compiler - even though the implementation is identical. (I know that the compiler can't really warn here as it doesn't see INTO the hashset).

@agocke - maybe you know about this problem some more?

I don't really mind the `!` mainly because `IEnumerator.Current` is defined as returning `default` if the enumerator is not positioned on a valid value (before start, or after end), so the functionality is correct. I'm just curious how this is supposed to work.

src/ILLink.Shared/DataFlow/ValueSet.cs
vitek-karas: It should be null - you're right.

We could store the enumerator in the same field, but it would require more casting, which is slower. In the `ValueSet` the memory footprint is important as we will store these values for a relatively long time and so they add up. On the other hand the `Enumerator` class should be VERY short lived. Basically always only stack allocated, so trading memory for CPU here feels like the better option.

src/ILLink.Shared/DataFlow/ValueSet.cs
sbomer: It looks like the compiler does [warn](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK7wCYgNQB8CwAUAMQyoA25Ew5ApgAS0zV1EDC9AxvQLz0y0A7vQDiAHjYA+egAoAlADoRtBPIDcRIgAEATPQ4BvAL7a94gCrSDRerfpaAzPXOiV83tIy0AZhAoINQiMgA) when `T` is unconstrained, in case it's instantiated with a non-nullable reference type.

And the implementation of `HashSet<T>` uses `!` I think for the same reason you did here:
https://github.com/dotnet/runtime/blob/85642f8147d9e2ba860c21baea25cbb4f12e665e/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs#L1541.

vitek-karas: I could swear I looked through the HashSet code and tried something similar with the T. Anyway...
Thanks a lot!

PR #2401, opened 2021-11-27
[feature/damAnalyzer] Change linker dataflow to use ValueSet/MultiValue

vitek-karas:
Currently linker uses ValueNode to represent both single and multi-values. The multi-value is stored as a special value MergePoint. This means that basically every "Reader" should call UniqueValues which enumerates all values in the value. Otherwise it may only react to single-values and not function correctly for multi-values.

Also in order to share more code with the analyzer, we need to switch to the new model, where multi-values is stored in an explicit ValueSet.

This change replaces all the places where multiple values can be stored with MultiValue (=ValueSet<ValueNode>). Then it removes the MergePoint values and the entire infrastructure around it.
Other notable changes:
- Added implicit conversion from single TValue to ValueSet<TValue> to make it much easier to work with
- Changed how ArrayValue is created to use factory methods. Previously ArrayValue could store a multi-value as its size. This meant that it was necessary to enumerate UniqueValues on ArrayValue to get actual sizes (results in multiple ArrayValues). This change introduces a factory method which return a MultiValue which may contain multiple ArrayValues if there are multiple potential sizes. A bit more memory hungry in some cases, but makes the "reader" simpler and less error prone.
- Added IsEmpty on ValueSet - as it's relatively common to check for empty ValueSet
- This actually fixes a bug https://github.com/dotnet/linker/issues/2025 because it now correctly tracks empty MultiValue
- This change finally gets rid of all possible places where ValueNode could be null before - null was used as basically Unknown value, but it sometimes caused nullrefs in weird places. This can't happen anymore, since MultiValue is a struct and thus is always - non-null.

Note: Currently this makes the data flow in linker relatively inefficient as ValueSet always creates a HashSet internally. https://github.com/dotnet/linker/pull/2397 should fix that and store single-values as direct references without a HashSet, basically going back to what linker did before (the only difference is that ValueNode reference is now wrapped in a ValueSet struct).

### DIFF
--- a/src/ILLink.Shared/DataFlow/ValueSet.cs
+++ b/src/ILLink.Shared/DataFlow/ValueSet.cs
@@ -93,7 +93,7 @@ namespace ILLink.Shared.DataFlow
 
 		ValueSet (EnumerableValues values) => _values = values;
 
-		public static implicit operator ValueSet<TValue>(TValue value) => new (value);
+		public static implicit operator ValueSet<TValue> (TValue value) => new (value);
 
 		public override bool Equals (object? obj) => obj is ValueSet<TValue> other && Equals (other);
 
@@ -162,7 +162,7 @@ namespace ILLink.Shared.DataFlow
 			return new ValueSet<TValue> (values);
 		}
 
-		public bool IsEmpty () => Values == null || Values.Count == 0;
+		public bool IsEmpty () => _values == null;
 
 		public override string ToString ()
 		{
--- a/src/ILLink.Shared/DataFlow/ValueSet.cs
+++ b/src/ILLink.Shared/DataFlow/ValueSet.cs
@@ -93,6 +93,8 @@ namespace ILLink.Shared.DataFlow
 
 		ValueSet (EnumerableValues values) => _values = values;
 
+		public static implicit operator ValueSet<TValue>(TValue value) => new (value);
+
 		public override bool Equals (object? obj) => obj is ValueSet<TValue> other && Equals (other);
 
 		public bool Equals (ValueSet<TValue> other)
@@ -159,6 +161,8 @@ namespace ILLink.Shared.DataFlow
 			values.UnionWith (right);
 			return new ValueSet<TValue> (values);
 		}
+
+		public bool IsEmpty () => Values == null || Values.Count == 0;
 
 		public override string ToString ()
 		{
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -918,7 +918,7 @@ namespace Mono.Linker.Dataflow
 							if (value is SystemTypeValue systemTypeValue) {
 								foreach (var stringParam in methodParams[2]) {
 									if (stringParam is KnownStringValue stringValue) {
-										BindingFlags bindingFlags = methodParams[0].AsSingleValue()?.Kind == ValueNodeKind.Null ? BindingFlags.Static : BindingFlags.Default;
+										BindingFlags bindingFlags = methodParams[0].AsSingleValue ()?.Kind == ValueNodeKind.Null ? BindingFlags.Static : BindingFlags.Default;
 										if (fieldOrPropertyInstrinsic == IntrinsicId.Expression_Property) {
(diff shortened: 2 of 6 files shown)


src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -428,20 +184,29 @@ public static bool DetectCycle (this ValueNode? node, HashSet<ValueNode> seenNod
 			return foundCycle;
 		}
 
-		public static ValueNode.UniqueValueCollection UniqueValues (this ValueNode? node)
+		public static int? AsConstInt (this ValueNode? node)
```
sbomer, line 187: Can this be changed to take a non-nullable `ValueNode` now?

src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -93,6 +93,8 @@ public void Reset ()
 
 		ValueSet (EnumerableValues values) => _values = values;
 
+		public static implicit operator ValueSet<TValue> (TValue value) => new (value);
```
sbomer, line 96: How bad is it if we don't have the implicit conversion? Generally I'd prefer to be explicit, but if this is used in a lot of places it makes sense.

sbomer (approved):
This is great, I love how this avoids nullable `ValueNode` in a lot of places. Sorry it took me a while to do the review.

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -332,20 +113,6 @@ public void Dispose ()
 	/// </summary>
 	public abstract class LeafValueNode : ValueNode
```
marek-safar, line 114: Is the type needed?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -6,18 +6,22 @@
...
 
+using MultiValue = ILLink.Shared.DataFlow.ValueSet<Mono.Linker.Dataflow.ValueNode>;
+
 namespace Mono.Linker.Dataflow
 {
 	class ReflectionMethodBodyScanner : MethodBodyScanner
 	{
 		readonly MarkStep _markStep;
 		readonly MarkScopeStack _scopeStack;
+		static ValueSetLattice<ValueNode> MultiValueLattice => default;
```
marek-safar, line 24: what about making same field in base type protected?

src/ILLink.Shared/DataFlow/ValueSet.cs
vitek-karas: It was at least 20 places before I gave up... so probably another 20 after that point :-)

I thought about it and it's actually not surprising. It's a discrepancy between production and consumption.

On the production side a `MultiValue` is always created with just a single value in it - intrinsics, all the visitors/scanners, ... this has to be like that since it needs to create a specific value type (Field access, method parameter, ...). (The only exception is arrays, but those are weird). The only way (except array) we create true multiple-value `MultiValue` is through `Meet`.

On the consumption side everything needs to be able to handle multiple values, so it has to take input as `MultiValue`. Lot of the common helpers are like that. But lot of times these helpers are also called directly from the creation path, which only has single-value at hand. All these places require the conversion. Technically this could be solved by having two overloads for each helper... but that's just messy.

vitek-karas: Out of curiosity, if I comment out the implicit conversion I get 57 build errors :wink:

src/linker/Linker.Dataflow/ValueNode.cs
```diff
@@ -1216,29 +747,38 @@ protected override string NodeToString ()
...
+		{
+			MultiValue result = MultiValueLattice.Top;
+			foreach (var sizeValue in size) {
+				result = MultiValueLattice.Meet (result, new MultiValue (new ArrayValue (sizeValue, elementType)));
+			}
+
+			return result;
+		}
+
+		public static MultiValue Create (ValueNode size, TypeReference elementType)
```
marek-safar, line 762: As this is only helper would be better if this took int as size?

src/linker/Linker.Dataflow/ValueNode.cs
vitek-karas: Not right now, but I intentionally left the type design same as before to avoid too much churn. I will have a separate PR to rework the value type system.

PR #2400, opened 2021-11-27
Fix NullReferenceException in Analyzer

tlakollo:
Fix NullReferenceException in Analyzer by checking the containing type is not null before using it
Fixes #2398

PR #2408, opened 2021-11-30
Fix Bug 2379

jtschuster:
For mismatch requires, also checks if parent has the requires attribute instead of just the current symbol.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/linker?pullRequest=2408) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/linker?pullRequest=2408) before we can accept your contribution.<br/><br/>:x: jtschuster [sign now](https://cla.dotnetfoundation.org/dotnet/linker?pullRequest=2408)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/linker?pullRequest=2408) it.</sub>

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -335,7 +335,13 @@ private void ReportMismatchInAttributesDiagnostic (SymbolAnalysisContext symbolA
 				message));
 		}
 
-		private bool HasMismatchingAttributes (ISymbol member1, ISymbol member2) => member1.HasAttribute (RequiresAttributeName) ^ member2.HasAttribute (RequiresAttributeName);
+		// TODO: Should consider when a symbol is a method of a class that has the attribute
+		private bool HasMismatchingAttributes (ISymbol member1, ISymbol member2)
+		{
+			bool member1HasAttribute = IsMemberInRequiresScope (member1);
```
agocke, line 341: It looks like this should be `TargetHasRequiresAttribute`, not `IsMemberInRequiresScope`, but it looks like `TargetHasRequiresAttribute` may need more work to handle associated symbols (properties)

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -371,7 +383,7 @@ protected bool TargetHasRequiresAttribute (ISymbol member, [NotNullWhen (returnV
 			}
 
 			// Also check the containing type
-			if (member.IsStatic || member.IsConstructor ()) {
+			if (inheritRucForInstanceMethods || member.IsStatic || member.IsConstructor ()) {
```
agocke, line 386: Rather than use a flag, I think we can keep this method the same, call it, and then check this extra info in the caller

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -360,6 +367,11 @@ protected bool IsMemberInRequiresScope (ISymbol containingSymbol)
 		/// True if the target of a call is considered to be annotated with the Requires... attribute
 		/// </summary>
 		protected bool TargetHasRequiresAttribute (ISymbol member, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttribute)
```
agocke, line 369: I'd change the name here to indicate this one is for overriding. Maybe something like `TargetHasRequiresAttributeForOverriding`

jtschuster: Doesn't the `protected` vs `private` already indicate which should be overridden?

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
sbomer: Why do you think it should be `TargetHasRequires`? In the linker we check the equivalent of `IsMemberInRequiresScope`.
```csharp
class Base {
  [RUC]
  virtual void M() {} // TargetHasRequires = true, IsInRequiresScope = true
}

[RUC]
class Derived : Base {
  override void M() {} // TargetHasRequires = false, IsInRequiresScope = false
}
```

I agree that `IsMemberInRequiresScope` needs to be updated to handle properties.

However this does raise an interesting question - what happens in the following case?
```csharp
[RUC]
class Base {
  virtual void M() {} // TargetHasRequires = false, IsInRequiresScope = true
}

class Derived : Base {
  [RUC]
  override void M() {} // TargetHasRequires = true, IsInRequiresScope = true
}
```

I think this shouldn't warn about a mismatch, similar to above. The justification is that we don't need to warn for calls to `Base.M` even though it might require unreferenced code, because constructing the `Base` type produces a warning already. But now you can do `new Derived().M()`. In IL this becomes a call to `Base.M`, which doesn't warn, but at runtime it calls the derived method. Does the analyzer produce any warnings for this? In the linker it's ok since the derived ctor will produce a warning (IIRC) about the call to the base ctor (which has TargetHasRequires = true). But in the analyzer there's no location for the compiler-generated derived ctor.

Do we need to ensure that compiler-generated derived ctors warn if the base class has RUC or are we already handling this somehow?

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
agocke: Sorry, I didn't mean which method should be overridden, I meant that this method should be used for determining mismatches when overriding. But given Sven's feedback that this is done differently in the linker, maybe the right answer is to go back to the old API and make it behave as this method does.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
agocke: I didn't realize `IsMemberInRequiresScope` is what's used in the linker. My mistake. In that case, we should use that instead.

sbomer (commented):
Looking good - to nail down the expected behavior, would you be able to add some tests for the cases where either the base or virtual has the property annotated, and the other has the accessor annotated? And also for the case where one of the virtuals has both the property and the accessor annotated?

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
sbomer: Per discussion I think the concern about compiler-generated derived ctors will be addressed by https://github.com/dotnet/linker/issues/2416.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -335,7 +335,13 @@ private void ReportMismatchInAttributesDiagnostic (SymbolAnalysisContext symbolA
 				message));
 		}
 
-		private bool HasMismatchingAttributes (ISymbol member1, ISymbol member2) => member1.HasAttribute (RequiresAttributeName) ^ member2.HasAttribute (RequiresAttributeName);
+		// TODO: Should consider when a symbol is a method of a class that has the attribute
```
sbomer, line 338: I think this TODO is addressed

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -355,6 +361,18 @@ protected bool IsMemberInRequiresScope (ISymbol containingSymbol)
 			return false;
 		}
 
+		/// <summary>
+		/// True if member of a call is considered to be annotated with the Requires... attribute. Primarily for property accessor declarations.
+		/// </summary>
```
sbomer, line 366: Maybe note that this method is like `IsMemberInRequiresScope`, except that it intentionally doesn't check the property annotations because the intended behavior is to require the base/override annotations to be on the same member.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -355,6 +361,18 @@ protected bool IsMemberInRequiresScope (ISymbol containingSymbol)
...
+		/// <param name="containingSymbol">
+		///	Symbol that is either an overriding member or an overriden/virtual member
+		/// </param>
+		protected bool IsOverrideMemberInRequiresScope (ISymbol containingSymbol)
+		{
+			return containingSymbol.HasAttribute (RequiresAttributeName)
+					|| containingSymbol.ContainingType.HasAttribute (RequiresAttributeName);
```
sbomer, line 373: Does this need a null check similar to the one added in https://github.com/dotnet/linker/pull/2400? /cc @tlakollo

sbomer (approved):
A few more comments, otherwise LGTM. Thanks!

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
tlakollo: This needs the fix from #2400, it should be on main already

sbomer:
One more thing @tlakollo pointed out - would you mind retargeting this PR to the 6.0.2xx branch? The current guidance is to put analyzer-only changes into that branch, which will ship with the 6.0.2 SDK release. That branch automatically gets merged to main so it will end up in .NET 7 as well. @vitek-karas please let us know if you think otherwise.

tlakollo:
Retargeting the PR is not enough, since the branch you created is from Main now GitHub wants to merge all the changes that are in Main into 6.0.2xx 
I'm not sure what's the easiest way to solve this, sounds like resetting to 6.0.2xx and then applying the changes is the way to go.

vitek-karas:
I'm fine taking this to 6.0.2xx. Will need to go through triage though.

.gitignore
```diff
@@ -10,6 +10,7 @@
 *.user
 *.userprefs
 .vs/
+.vscode/
```
vitek-karas, line 13: Can you please separate these changes?

I'm not saying we don't want them, but we should not "merge" two unrelated things. Especially in servicing branches.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresCapability.cs
```diff
@@ -1453,7 +1573,7 @@ public static int Method ()
 				}
 
 				// Bug https://github.com/dotnet/linker/issues/2379
-				[ExpectedWarning ("IL2046", ProducedBy = ProducedBy.Analyzer)]
+				//[ExpectedWarning ("IL2046", ProducedBy = ProducedBy.Analyzer)]
```
vitek-karas, line 1576: If the warning is no longer expected, please delete this line and also the comment with the link to the bug (similarly above).

vitek-karas (approved):
Looks good.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresCapability.cs
```diff
@@ -990,6 +998,65 @@ public override void VirtualMethod ()
 
 				[ExpectedWarning ("IL3003", "DerivedClassWithoutRequires.VirtualPropertyAnnotationInProperty", "BaseClassWithRequires.VirtualPropertyAnnotationInProperty", ProducedBy = ProducedBy.Analyzer)]
 				public override string VirtualPropertyAnnotationInProperty { get; set; }
+
+				[ExpectedWarning ("IL3003", "DerivedClassWithoutRequires.VirtualPropertyAnnotationInPropertyAndAccessor", "BaseClassWithRequires.VirtualPropertyAnnotationInPropertyAndAccessor", ProducedBy = ProducedBy.Analyzer)]
+
```
tlakollo, line 1003: nit: there is an extra line between the attribute and the accesor

tlakollo (approved):

PR #2418, opened 2021-12-03
Produce diagnostics for KnownValueTypes that have annotations in its members

### LINKED ISSUE #2339: [DAM analyzer] Port DAM binding logic to analyzer

tlakollo:
Add KnownTypeValue as a subclass of ValueWithDynamicallyAccessedMembers
Add IL2110 and IL2111 to shared Diagnostic ids
Add test in analyzer for now since intrinsics are still not handled
Delete the use of context in the binder since the context is never used
Produce diagnostics for KnownValueTypes that have annotations in its members

Fixes https://github.com/dotnet/linker/issues/2410
Fixes #2339


sbomer (commented):
Looking good! Would it be reasonable to turn on the test `AnnotatedMembersAccessedViaReflection` and use `ProducedBy` to skip the ones that rely on intrinsics? It would be good to get the extra coverage.

src/ILLink.RoslynAnalyzer/TrimAnalysis/KnownValueType.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
+using ILLink.Shared;
+using ILLink.Shared.TrimAnalysis;
+using Microsoft.CodeAnalysis;
+
+namespace ILLink.RoslynAnalyzer.TrimAnalysis
+{
+	public record KnownValueType : ValueWithDynamicallyAccessedMembers
```
sbomer, line 12: This can now use `SystemTypeValue`

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -153,6 +178,56 @@ static DiagnosticId GetDiagnosticId (SymbolValue source, SymbolValue target)
...
+				member.GetDisplayName (),
+				message,
+				url);
+		}
+
+		static IEnumerable<Diagnostic> VerifyMethodSymbolForDiagnostic (IMethodSymbol methodSymbol, Location location)
+		{
+			foreach (var parameter in methodSymbol.Parameters) {
+				if (parameter.GetDynamicallyAccessedMemberTypes () != DynamicallyAccessedMemberTypes.None)
+					yield return Diagnostic.Create (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection), location, methodSymbol.GetDisplayName ());
```
sbomer, line 197: I think we should avoid yielding multiple diagnostics - there should be just one diagnostic for the method if any of its parameters or the return type on a virtual have annotations.

tlakollo: I created a piece of code that I actually didnt like much which does yield return and then yield break to keep the IEnumerator and use common code. Maybe I should divide the Property that yields an IEnumerator from the method and field analysis which should return only one diagnostic. Let me know what you think

tlakollo:
I will change this to draft, I realized that there is several tests having problems with interfaces and implementations. Need to spend more time on it.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresAttributeMismatch.cs
```diff
@@ -21,18 +21,30 @@ namespace Mono.Linker.Tests.Cases.RequiresCapability
...
-		[ExpectedWarning ("IL2026", "VirtualPropertyAnnotationInPropertyAndAccessor.get", ProducedBy = ProducedBy.Trimmer)]
-		[ExpectedWarning ("IL2026", "VirtualPropertyAnnotationInPropertyAndAccessor.get", ProducedBy = ProducedBy.Trimmer)]
-		[ExpectedWarning ("IL2026", "PropertyAnnotationInPropertyAndAccessor.set", ProducedBy = ProducedBy.Trimmer)]
+		[ExpectedWarning ("IL2026", "BaseClassWithRequires.VirtualPropertyAnnotationInAccesor.get")]
+		[ExpectedWarning ("IL2026", "BaseClassWithRequires.VirtualPropertyAnnotationInAccesor.get")]
+		[ExpectedWarning ("IL2026", "BaseClassWithRequires.VirtualPropertyAnnotationInAccesor.get")]
+		[ExpectedWarning ("IL2026", "DerivedClassWithRequires.VirtualPropertyAnnotationInAccesor.get", ProducedBy = ProducedBy.Analyzer)]
+		[ExpectedWarning ("IL2026", "DerivedClassWithAllWarnings.VirtualPropertyAnnotationInAccesor.set", ProducedBy = ProducedBy.Analyzer)]
+		[ExpectedWarning ("IL2026", "DerivedClassWithAllWarnings.VirtualPropertyAnnotationInProperty.get", ProducedBy = ProducedBy.Analyzer)]
+		[ExpectedWarning ("IL2026", "DerivedClassWithAllWarnings.VirtualPropertyAnnotationInProperty.set", ProducedBy = ProducedBy.Analyzer)]
```
vitek-karas, line 30: Do we have a bug for this discrepancy?

test/Mono.Linker.Tests.Cases/DataFlow/AnnotatedMembersAccessedViaReflection.cs
```diff
@@ -643,11 +652,11 @@ struct ValueWithAnnotatedField
 				public Type _typeField;
 			}
 
-			[ExpectedWarning ("IL2110", nameof (ValueWithAnnotatedField._typeField))]
+			[ExpectedWarning ("IL2110", nameof (ValueWithAnnotatedField._typeField), ProducedBy = ProducedBy.Trimmer)]
 			[DllImport ("nonexistent")]
 			static extern ValueWithAnnotatedField GetValueWithAnnotatedField ();
 
-			[ExpectedWarning ("IL2110", nameof (ValueWithAnnotatedField._typeField))]
+			[ExpectedWarning ("IL2110", nameof (ValueWithAnnotatedField._typeField), ProducedBy = ProducedBy.Trimmer)]
```
vitek-karas, line 659: Ideally we should add a link to a bug - it will be easier to find.

test/Mono.Linker.Tests.Cases/DataFlow/AnnotatedMembersAccessedViaReflection.cs
```diff
@@ -264,13 +264,22 @@ public static void Test ()
 		class AnnotatedMethodReturnValue
 		{
 			[return: DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicMethods)]
-			public static Type StaticMethodWithAnnotatedReturnValue () => null;
+			public static Type StaticMethodWithAnnotatedReturnValue ()
+			{
+				return null;
+			}
```
vitek-karas, line 270: I don't really mind, but why did you change this?

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -160,17 +160,12 @@ private static void M(Type type)
...
 			// (17,9): warning IL2070: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethod(String)'.
 			// The parameter 'type' of method 'C.M(Type)' does not have matching annotations.
 			// The source value must declare at least the same requirements as those declared on the target location it is assigned to.
-			return VerifyDynamicallyAccessedMembersAnalyzer (TargetMethodWithAnnotations,
-				VerifyCS.Diagnostic (DiagnosticId.DynamicallyAccessedMembersMismatchParameterTargetsThisParameter)
-				.WithSpan (17, 9, 17, 30)
-				.WithArguments ("System.Type.GetMethod(String)",
-					"type",
-					"C.M(Type)",
-					"'DynamicallyAccessedMemberTypes.PublicMethods'"));
```
vitek-karas, line 173: Could you maybe just comment out the code - we should add it later once the intrinsics are implemented, right?

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -38,6 +38,24 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
...
+				return true;
+
+			case IntrinsicId.Type_GetConstructor:
+				return true;
+
+			case IntrinsicId.Type_GetEvent:
+				return true;
+
+			case IntrinsicId.Activator_CreateInstance_Type:
+				return true;
```
vitek-karas, line 57: Please add a comment why you're doing this.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresAttributeMismatch.cs
tlakollo: https://github.com/dotnet/linker/issues/2533
I will add links to issues to several places

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionMethodBodyScanner.cs
```diff
@@ -0,0 +1,114 @@
...
+			var message = RequiresUnreferencedCodeUtils.GetMessageFromAttribute (requiresAttributeData);
+			var url = RequiresAnalyzerBase.GetUrlFromAttribute (requiresAttributeData);
+			diagnosticContext.ReportDiagnostic (DiagnosticId.RequiresUnreferencedCode, member.GetDisplayName (), message, url);
+		}
+
+		static void MarkMethod (DiagnosticContext diagnosticContext, IMethodSymbol methodSymbol)
+		{
+			if (methodSymbol.TargetHasRequiresUnreferencedCodeAttribute (out var requiresAttributeData)
+						&& RequiresUnreferencedCodeUtils.VerifyRequiresUnreferencedCodeAttributeArguments (requiresAttributeData)) {
+				ReportRequiresUnreferencedCodeDiagnostic (diagnosticContext, requiresAttributeData, methodSymbol);
```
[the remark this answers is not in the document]
tlakollo: No, because property is not allowed to have RUC, only DAM is allowed

test/Mono.Linker.Tests.Cases/DataFlow/AnnotatedMembersAccessedViaReflection.cs
tlakollo: At some point the analyzer was failing on arrow expressions, now its fixed so Im getting back the arrows

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionMethodBodyScanner.cs
```diff
@@ -0,0 +1,114 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+using ILLink.RoslynAnalyzer.DataFlow;
+using ILLink.Shared;
+using ILLink.Shared.TrimAnalysis;
+using Microsoft.CodeAnalysis;
+
+namespace ILLink.RoslynAnalyzer.TrimAnalysis
+{
+	class ReflectionMethodBodyScanner
```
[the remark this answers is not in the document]
tlakollo: Changed to diagnosticSolver, let me know what you think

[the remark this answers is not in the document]
tlakollo: Offline discussion  - changed the name to ReflectionAccessAnalyzer

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionMethodBodyScanner.cs
```diff
@@ -0,0 +1,114 @@
...
+			if (methodSymbol.IsVirtual && methodSymbol.GetDynamicallyAccessedMemberTypesOnReturnType () != DynamicallyAccessedMemberTypes.None)
+				diagnosticContext.ReportDiagnostic (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, methodSymbol.GetDisplayName ());
+			else if (methodSymbol.IsVirtual && methodSymbol.MethodKind is MethodKind.PropertyGet &&
+				(methodSymbol.GetDynamicallyAccessedMemberTypes () != DynamicallyAccessedMemberTypes.None ||
+				methodSymbol.GetDynamicallyAccessedMemberTypesOnAssociatedSymbol () != DynamicallyAccessedMemberTypes.None)) {
+				diagnosticContext.ReportDiagnostic (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, methodSymbol.GetDisplayName ());
+			} else if (methodSymbol.MethodKind is MethodKind.PropertySet &&
+				  (methodSymbol.GetDynamicallyAccessedMemberTypesOnReturnType () != DynamicallyAccessedMemberTypes.None ||
+				  methodSymbol.GetDynamicallyAccessedMemberTypesOnAssociatedSymbol () != DynamicallyAccessedMemberTypes.None)) {
+				diagnosticContext.ReportDiagnostic (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, methodSymbol.GetDisplayName ());
```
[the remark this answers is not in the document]
tlakollo: I think I need to spend more time on this one, and probably add more tests.

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionMethodBodyScanner.cs
```diff
@@ -0,0 +1,114 @@
...
+					MarkProperty (diagnosticContext, property);
+					break;
+				/* Skip Type and InterfaceImplementation marking since doesnt seem relevant for diagnostic generation
+				case ITypeSymbol nestedType:
+					MarkType (diagnosticContext, nestedType);
+					break;
+				case InterfaceImplementation interfaceImplementation:
+					MarkInterfaceImplementation (analysisContext, interfaceImplementation, dependencyKind);
+					break;
+				*/
```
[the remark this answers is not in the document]
tlakollo: So I think is kind of handled, each member looks for RUC on the member and on the type based on the conditions (being static or constructor and the ContainingType has RUC), so at least type-level RUC seems fine to me. Maybe once more intrinsics and we enable more tests like [ObjectGetType](https://github.com/dotnet/linker/blob/main/test/Mono.Linker.Tests.Cases/Reflection/ObjectGetType.cs) we can have enable the DAM on type level

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionAccessAnalyzer.cs
```diff
@@ -0,0 +1,102 @@
...
+				case IMethodSymbol method:
+					GetDiagnosticsForMethod (diagnosticContext, method);
+					break;
+				case IFieldSymbol field:
+					GetDiagnosticsForField (diagnosticContext, field);
+					break;
+				case IPropertySymbol property:
+					GetDiagnosticsForProperty (diagnosticContext, property);
+					break;
+				/* Skip Type and InterfaceImplementation marking since doesnt seem relevant for diagnostic generation
```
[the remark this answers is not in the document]
vitek-karas: I think https://github.com/dotnet/linker/blob/3f704bb49eb2fb2847ceaecfbabe8d5f3e0c86c9/test/Mono.Linker.Tests.Cases/Reflection/TypeHierarchyReflectionWarnings.cs#L267 will cover this once we get there.

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionAccessAnalyzer.cs
```diff
@@ -0,0 +1,102 @@
...
+				diagnosticContext.ReportDiagnostic (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, methodSymbol.GetDisplayName ());
+			else if (methodSymbol.IsVirtual && methodSymbol.MethodKind is MethodKind.PropertyGet &&
+				(methodSymbol.GetDynamicallyAccessedMemberTypes () != DynamicallyAccessedMemberTypes.None ||
+				methodSymbol.GetDynamicallyAccessedMemberTypesOnAssociatedSymbol () != DynamicallyAccessedMemberTypes.None)) {
+				diagnosticContext.ReportDiagnostic (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, methodSymbol.GetDisplayName ());
+			} else if (methodSymbol.MethodKind is MethodKind.PropertySet &&
+				  (methodSymbol.GetDynamicallyAccessedMemberTypesOnReturnType () != DynamicallyAccessedMemberTypes.None ||
+				  methodSymbol.GetDynamicallyAccessedMemberTypesOnAssociatedSymbol () != DynamicallyAccessedMemberTypes.None)) {
+				diagnosticContext.ReportDiagnostic (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, methodSymbol.GetDisplayName ());
+			} else {
```
vitek-karas, line 65: I'm still not sure about this one.

```C#
class TestType
{
    [DAM(Methods)]
    public Type MyType { get; }

    public Type MyTypeWithGetter { [return: DAM(Methods)] get { return null; } }
}

// This should produce 2 warnings
// but the code above will very likely only produce 1 - the first one, not for `MyTypeWithGetter`
typeof (TestType).RequiresAll ();
```

The reason is that on a getter we check if the method itself has DAM, but on a setter we check if the return value has DAM. Return value on setter is void, so that is super weird as well.

tlakollo:
Added a bug for most of the properties inconsistencies https://github.com/dotnet/linker/issues/2575
The problem is naming so I'm not sure if it's linker logic or test infra. The system to present warnings in those cases is wrong in many ways. For example, makes a non virtual getter to warn

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionAccessAnalyzer.cs
tlakollo: Added test for these cases

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionMethodBodyScanner.cs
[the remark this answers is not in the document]
tlakollo: End up using the FlowAnnotations helpers by Vitek suggestion. Thanks Vitek for the review :D

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -160,17 +160,19 @@ private static void M(Type type)
...
+			return VerifyDynamicallyAccessedMembersAnalyzer (TargetMethodWithAnnotations
+							/*,
+							VerifyCS.Diagnostic (DiagnosticId.DynamicallyAccessedMembersMismatchParameterTargetsThisParameter)
+							.WithSpan (17, 9, 17, 30)
+							.WithArguments ("System.Type.GetMethod(String)",
+								"type",
+								"C.M(Type)",
+								"'DynamicallyAccessedMemberTypes.PublicMethods'")*/);
```
tlakollo, line 175: Not sure why lint decided on this format but is on purpose

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionAccessAnalyzer.cs
```diff
@@ -44,29 +44,23 @@ static void ReportRequiresUnreferencedCodeDiagnostic (in DiagnosticContext diagn
...
-				(methodSymbol.GetDynamicallyAccessedMemberTypes () != DynamicallyAccessedMemberTypes.None ||
-				methodSymbol.GetDynamicallyAccessedMemberTypesOnAssociatedSymbol () != DynamicallyAccessedMemberTypes.None)) {
-				diagnosticContext.ReportDiagnostic (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, methodSymbol.GetDisplayName ());
-			} else if (methodSymbol.MethodKind is MethodKind.PropertySet &&
-				  (methodSymbol.GetDynamicallyAccessedMemberTypesOnReturnType () != DynamicallyAccessedMemberTypes.None ||
-				  methodSymbol.GetDynamicallyAccessedMemberTypesOnAssociatedSymbol () != DynamicallyAccessedMemberTypes.None)) {
-				diagnosticContext.ReportDiagnostic (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, methodSymbol.GetDisplayName ());
-			} else {
+			if (methodSymbol.GetDynamicallyAccessedMemberTypes () != DynamicallyAccessedMemberTypes.None)
+				diagnosticContext.AddDiagnostic (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, methodSymbol.GetDisplayName ());
```
vitek-karas, line 56: Ideally this should be only done on instance method (non-static) and only when the declaring type derives from `System.Type`. The latter part is tracked by either https://github.com/dotnet/linker/issues/2409 or a line item in https://github.com/dotnet/linker/issues/2273.

So probably OK to leave it as is for now - as we will likely handle this as a separate part of the analyzer. It might produce noise if there's such a case in the code though, but not a big deal.

src/ILLink.Shared/TrimAnalysis/RequireDynamicallyAccessedMembersAction.cs
```diff
@@ -22,7 +22,7 @@ public void Invoke (in DiagnosticContext diagnosticContext, in MultiValue value,
 					var availableMemberTypes = valueWithDynamicallyAccessedMembers.DynamicallyAccessedMemberTypes;
 					if (!Annotations.SourceHasRequiredAnnotations (availableMemberTypes, targetValue.DynamicallyAccessedMemberTypes, out var missingMemberTypes)) {
 						(var diagnosticId, var diagnosticArguments) = Annotations.GetDiagnosticForAnnotationMismatch (valueWithDynamicallyAccessedMembers, targetValue, missingMemberTypes);
-						diagnosticContext.AddDiagnostic (diagnosticId, diagnosticArguments);
+						_diagnosticContext.AddDiagnostic (diagnosticId, diagnosticArguments);
```
vitek-karas, line 25: Declare the `_diagnosticContext` field on the shared struct (so in this file) and as readonly. That will "force" the analyzer/linker to initialize it in the .ctors. Cleaner than relying on the fields existence from the "outside".

test/Mono.Linker.Tests.Cases/DataFlow/AnnotatedMembersAccessedViaReflection.cs
```diff
@@ -470,10 +521,26 @@ static void DynamicallyAccessedMembersSuppressedByRUC ()
...
+			[ExpectedWarning ("IL2111", nameof (PropertyWithAnnotationOnMembers) + ".set")]
+			[ExpectedWarning ("IL2111", nameof (PropertyWithAnnotationOnMembers) + ".set")]
+			[ExpectedWarning ("IL2111", nameof (PropertyWithAnnotationOnMembers) + ".set", ProducedBy = ProducedBy.Trimmer)]
+			[ExpectedWarning ("IL2111", nameof (PropertyWithAnnotationOnMembers) + ".set", ProducedBy = ProducedBy.Trimmer)]
+			[ExpectedWarning ("IL2111", nameof (PropertyWithAnnotationOnMembers) + ".get", ProducedBy = ProducedBy.Trimmer)]
+			[ExpectedWarning ("IL2111", nameof (PropertyWithAnnotationOnMembers) + ".get", ProducedBy = ProducedBy.Trimmer)]
+			[ExpectedWarning ("IL2111", nameof (VirtualPropertyWithAnnotationOnMembers) + ".get", ProducedBy = ProducedBy.Analyzer)]
+			[ExpectedWarning ("IL2111", nameof (VirtualPropertyWithAnnotationOnMembers) + ".get", ProducedBy = ProducedBy.Analyzer)]
+			[ExpectedWarning ("IL2111", nameof (VirtualPropertyWithAnnotationOnMembers) + ".set", ProducedBy = ProducedBy.Analyzer)]
+			[ExpectedWarning ("IL2111", nameof (VirtualPropertyWithAnnotationOnMembers) + ".set", ProducedBy = ProducedBy.Analyzer)]
```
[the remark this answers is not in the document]
vitek-karas: I pushed a change which fixes it for the property test - maybe we should do the same for the other tests in this file as well.

The fix is more of a workaround, I simply renamed the properties so that the substrings don't match.
As far as I can tell the correct warnings are generated. And the linker and analyzer behavior matches exactly now (nice).

That said there are still duplicate warnings, but that's a different matter.

src/ILLink.RoslynAnalyzer/TrimAnalysis/RequireDynamicallyAccessedMembersAction.cs
```diff
@@ -7,9 +7,9 @@
 
 namespace ILLink.Shared.TrimAnalysis
 {
+#pragma warning disable CS0282 // There is no defined ordering between fields in multiple declarations of partial struct
```
vitek-karas, line 10: Cleaner way is to add `[StructLayout (LayoutKind.Auto)]` on the struct - it will shut down the warning and also tells the runtime it's free to optimize the layout if it helps.

vitek-karas: Also please add onto the shared source.

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionAccessAnalyzer.cs
```diff
@@ -0,0 +1,96 @@
...
+			if (eventSymbol.RemoveMethod is not null)
+				GetDiagnosticsForMethod (diagnosticContext, eventSymbol.RemoveMethod);
+			if (eventSymbol.RaiseMethod is not null)
+				GetDiagnosticsForMethod (diagnosticContext, eventSymbol.RaiseMethod);
+		}
+
+		static void GetDiagnosticsForField (in DiagnosticContext diagnosticContext, IFieldSymbol fieldSymbol)
+		{
+			if (fieldSymbol.TryGetRequiresUnreferencedCodeAttribute (out var requiresAttributeData))
+				ReportRequiresUnreferencedCodeDiagnostic (diagnosticContext, requiresAttributeData, fieldSymbol);
```
vitek-karas, line 90: I don't think we allows RUC on a field, right?

vitek-karas (approved):
Looks good - just a few minor things

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionAccessAnalyzer.cs
tlakollo: No, we don't - deleted the code

src/ILLink.RoslynAnalyzer/TrimAnalysis/RequireDynamicallyAccessedMembersAction.cs
```diff
@@ -2,12 +2,13 @@
...
 
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
 using ILLink.RoslynAnalyzer.TrimAnalysis;
 using ILLink.Shared.TypeSystemProxy;
 
 namespace ILLink.Shared.TrimAnalysis
 {
-#pragma warning disable CS0282 // There is no defined ordering between fields in multiple declarations of partial struct
+	[StructLayout (LayoutKind.Auto)]
```
vitek-karas, line 11: Nit: put this onto the shared source file - it will apply to both analyzer and linker.

src/ILLink.RoslynAnalyzer/TrimAnalysis/ReflectionAccessAnalyzer.cs
tlakollo: Apparently, we are when RequiresUnreferencedCode is on a class, I think we mark access to fields as dangerous because they can call the static constructor. Getting back the code

vitek-karas (approved):
:shipit:

PR #2423, opened 2021-12-05
Fix Bug 2378: COM analyzer warnings are not suppressed by RequiresUnreferencedCode

### LINKED ISSUE #2378: COM analyzer warnings are not suppressed by RequiresUnreferencedCode
See ComPInvokeWarnings.cs:

```C#

[RequiresUnreferencedCode ("test")]
// Bug
[ExpectedWarning ("IL2050", ProducedBy = ProducedBy.Analyzer)]
static void Call_CanSuppressWithRequiresUnreferencedCode ()
{
	CanSuppressWithRequiresUnreferencedCode (null);
}

[DllImport ("Foo")]
static extern void CanSuppressWithRequiresUnreferencedCode (IFoo foo);
```

jtschuster:
Fixes #2378


src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
```diff
@@ -39,6 +40,13 @@ public override void Initialize (AnalysisContext context)
 					if (!targetMethod.HasAttribute (DllImportAttribute))
 						return;
 
+					if (operationContext.ContainingSymbol is ISymbol containingSymbol) {
+						if (containingSymbol.HasAttribute (RequiresUnreferencedCodeAttribute)
+							|| containingSymbol.ContainingType.HasAttribute (RequiresUnreferencedCodeAttribute)) {
+							return;
```
vitek-karas, line 46: The "RUC on type" only applies to the type itself, it does not propagate into nested types. So similar to https://github.com/dotnet/linker/pull/2400/files we need to only check the `ContainingType` if the symbol is not a type.

Ideally we would share this logic between these places, as it should always be the same.

jtschuster: I see, I've updated the code. Should this functionality be moved to the ISymbol extensions to share it?

Tagging @tlakollo for input.

tlakollo: I don't have a strong opinion for having some methods in the ISymbol extensions file, but given that is a copy of Roslyn's ISymbolExtensions with some modifications and 'Requires' is very linker focused thing, plus the fact that the desired behavior will be in the future to have the implementation shared between linker and analyzer I think it will be better to have a separate file.

vitek-karas: Now that Tlaka's change is in the main, I think we should move this common logic into some helper somewhere and use it from both places.

tlakollo (approved):
Looks good to me

sbomer:
Would you be able to change this to target 6.0.2xx per https://github.com/dotnet/linker/pull/2408#issuecomment-990245195? Thanks!

vitek-karas:
I agree this should go to 6.0.2xx

I typically do something like this for retargeting:
* In your local branch which targets `main`, make sure you have everything comitted
* Squash the commits into just one - and copy the commit hash somewhere
* `git fetch upstream`
* `git reset --hard upstream/release/6.0.2xx`
* (You may need to do `git submodule update` here)
* `git cherry-pick <commit hash from above>`
* Test everything
* `git push --force`

And now go to the PR and check it to use the release/6.0.2xx as its base

vitek-karas:
Note that we will need to get this through triage before merging...

jtschuster:
Changes here are included in PR https://github.com/dotnet/linker/pull/2408, so closing this.

PR #2426, opened 2021-12-07
Simplify unknown binding flags handling and fix param names

vitek-karas:
The helpers to translate binding flags to annotation already handle unsupported/unknown binding flags, so there's no need for the intrinsics to handle that yet again. Just simplifies the code.

Also fixed places where we lookup method parameters. If the lookup is done on a MethodReference the parameters typically don't have names. So instead do it on the MethodDefinition where they do have names and thus produce much better diagnostic messages.


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1086,26 +1086,24 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 							_ => null,
 						};
 
+						// Otherwise fall back to the bitfield requirements
+						var requiredMemberTypes = GetDynamicallyAccessedMemberTypesFromBindingFlagsForConstructors (bindingFlags);
+						// We can scope down the public constructors requirement if we know the number of parameters is 0
+						if (requiredMemberTypes == DynamicallyAccessedMemberTypes.PublicConstructors && ctorParameterCount == 0)
+							requiredMemberTypes = DynamicallyAccessedMemberTypes.PublicParameterlessConstructor;
```
tlakollo, line 1093: We calculate always the requiredMemberTypes but, for most operations, we end up not using it and we just use the binding flags. I don't see the advantage of having it up here.

tlakollo (approved):
Looks good to me

PR #2430, opened 2021-12-08
Changes LinkerAttributesInformation._linkerAttributes to a List instead of a Dictionary

jtschuster:
Fix for issue https://github.com/dotnet/linker/issues/2387 to use a list instead of a dictionary for LinkerAttributesInformation._linkerAttributes since the size of the collection is usually very small.

### DIFF
--- a/src/linker/Linker/LinkerAttributesInformation.cs
+++ b/src/linker/Linker/LinkerAttributesInformation.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+//using System.ComponentModel;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
@@ -12,9 +13,11 @@ namespace Mono.Linker
 {
 	readonly struct LinkerAttributesInformation
 	{
-		readonly Dictionary<Type, List<Attribute>>? _linkerAttributes;
+		readonly List<Tuple<Type, List<Attribute>>>? _linkerAttributes;
 
-		private LinkerAttributesInformation (Dictionary<Type, List<Attribute>>? cache)
+		private static Predicate<Tuple<Type, List<Attribute>>> finder (Type matchedType) { return x => x.Item1 == matchedType; }
+
+		private LinkerAttributesInformation (List<Tuple<Type, List<Attribute>>>? cache)
 		{
 			this._linkerAttributes = cache;
 		}
@@ -23,7 +26,7 @@ namespace Mono.Linker
 		{
 			Debug.Assert (context.CustomAttributes.HasAny (provider));
 
-			Dictionary<Type, List<Attribute>>? cache = null;
+			List<Tuple<Type, List<Attribute>>>? cache = null;
 
 			foreach (var customAttribute in context.CustomAttributes.GetCustomAttributes (provider)) {
 				var attributeType = customAttribute.AttributeType;
@@ -54,12 +57,17 @@ namespace Mono.Linker
 					continue;
 
 				if (cache == null)
-					cache = new Dictionary<Type, List<Attribute>> ();
+					cache = new List<Tuple<Type, List<Attribute>>> ();
 
 				Type attributeValueType = attributeValue.GetType ();
-				if (!cache.TryGetValue (attributeValueType, out var attributeList)) {
+
+				var found = cache.Find (finder (attributeValueType));
+				List<Attribute> attributeList;
+				if (found == null || found == default) {
 					attributeList = new List<Attribute> ();
-					cache.Add (attributeValueType, attributeList);
+					cache.Add (Tuple.Create (attributeValueType, attributeList));
+				} else {
+					attributeList = found.Item2;
 				}
 
 				attributeList.Add (attributeValue);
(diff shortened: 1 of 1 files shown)


src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+//using System.ComponentModel;
```
tlakollo, line 6: Nit: Extra library

tlakollo (approved):
Looks good to me, I will add Vitek so he can check if he's satisfied with the solution

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -70,13 +78,19 @@ public static LinkerAttributesInformation Create (LinkContext context, ICustomAt
...
 		{
-			if (_linkerAttributes == null || !_linkerAttributes.TryGetValue (typeof (T), out var attributeList))
+			List<Attribute> attributeList;
+			if (_linkerAttributes == null)
+				return Enumerable.Empty<T> ();
+
+			var found = _linkerAttributes.Find (finder (typeof (T)));
+			if (found == null || found == default)
 				return Enumerable.Empty<T> ();
+			attributeList = found.Item2;
```
marek-safar, line 93: It would be better to introduce `TryGetAttribute` method and use it from both places where you added the lookup logic

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -12,9 +13,11 @@ namespace Mono.Linker
 {
 	readonly struct LinkerAttributesInformation
 	{
-		readonly Dictionary<Type, List<Attribute>>? _linkerAttributes;
+		readonly List<Tuple<Type, List<Attribute>>>? _linkerAttributes;
```
vitek-karas, line 16: Nit: I find it a bit easier to read to use the named values in this case, so for example:

```suggestion
		readonly List<(Type Type, List<Attribute> Attributes)>? _linkerAttributes;
```

That way when you're referencing the pieces you can use descriptive names. So instead of `x.Item1` it would read `x.Type` below.

It does make the type a bit longer though, but personally I still prefer the descriptive names.

Note that you only need to specify the names in the declaration, no need to do so in the "new" call (at least I think).

src/linker/Linker/LinkerAttributesInformation.cs
vitek-karas: +1

In fact I would rewrite the finding to loop over the list instead of using `Find`. The call `Find (finder (type))` must create a closure -> object allocation. For each lookup - that is pretty expensive.

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -54,12 +57,17 @@ public static LinkerAttributesInformation Create (LinkContext context, ICustomAt
...
 
 				Type attributeValueType = attributeValue.GetType ();
-				if (!cache.TryGetValue (attributeValueType, out var attributeList)) {
+
+				var found = cache.Find (finder (attributeValueType));
+				List<Attribute> attributeList;
+				if (found == null || found == default) {
 					attributeList = new List<Attribute> ();
-					cache.Add (attributeValueType, attributeList);
+					cache.Add (Tuple.Create (attributeValueType, attributeList));
```
vitek-karas, line 68: With the changed type declaration above, this can be simply (and yes, this does unfortunately require double parenthesis):
```suggestion
					cache.Add ((attributeValueType, attributeList));
```

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -6,24 +6,38 @@
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Security.Cryptography.X509Certificates;
```
vitek-karas, line 10: This looks weird - should not be needed - please remove any unused "using" statements.

vitek-karas (approved):

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -12,18 +12,30 @@ namespace Mono.Linker
...
 
-		private LinkerAttributesInformation (Dictionary<Type, List<Attribute>>? cache)
+		private LinkerAttributesInformation (List<(Type Type, List<Attribute> Attributes)>? cache)
 		{
 			this._linkerAttributes = cache;
 		}
 
+		private static bool TryFindAttributeList (List<(Type Type, List<Attribute> Attributes)> list, Type type, [NotNullWhen (returnValue: true)] out List<Attribute>? foundAttributes)
+		{
+			foundAttributes = null;
```
marek-safar, line 24: it should be set only before `return false`

src/linker/Linker/LinkerAttributesInformation.cs
```diff
@@ -12,18 +12,30 @@ namespace Mono.Linker
 {
 	readonly struct LinkerAttributesInformation
 	{
-		readonly Dictionary<Type, List<Attribute>>? _linkerAttributes;
+		readonly List<(Type Type, List<Attribute> Attributes)>? _linkerAttributes;
```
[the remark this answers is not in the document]
jtschuster: It doesn't look like C# lets you use an alias for a named pair. We could make a private struct to name it nicely but that might be overkill.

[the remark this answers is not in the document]
marek-safar: Yeah, it looks like C# compiler bug as it does not support tuple syntax here.

PR #2440, opened 2021-12-13
Fix Issue 2416: RUC Analyzer doesn't emit IL2109

jtschuster:
Fixes https://github.com/dotnet/linker/issues/2416 by emitting IL2109 (Derived class without RUC derives from base class with RUC). Also modifies the path of nested types in warnings to use dots as type delimiters instead of a slash. (e.g. "namespace1.namespace2.base.nested.nested2" instead of "namespace1.namespace2.base/nested1/nested2")

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
@@ -6,8 +6,6 @@ using System.Collections.Immutable;
 using ILLink.Shared;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
-using Microsoft.CodeAnalysis.CSharp;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace ILLink.RoslynAnalyzer
 {
@@ -25,7 +23,7 @@ namespace ILLink.RoslynAnalyzer
 		static readonly DiagnosticDescriptor s_makeGenericTypeRule = DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.MakeGenericType);
 		static readonly DiagnosticDescriptor s_makeGenericMethodRule = DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.MakeGenericMethod);
 
-		static readonly DiagnosticDescriptor s_typeDerivesFromRucClassRule = DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DerivedFromBaseWithRequiresUnreferencedCode);
+		static readonly DiagnosticDescriptor s_typeDerivesFromRucClassRule = DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DerivedClassRequiresUnreferencedCodeAttributeMismatch);
 
 		static readonly Action<OperationAnalysisContext> s_dynamicTypeInvocation = operationContext => {
 			if (FindContainingSymbol (operationContext, DiagnosticTargets.All) is ISymbol containingSymbol &&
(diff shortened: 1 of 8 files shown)


jtschuster:
I duplicated the "ExpectedWarning" attributes to show the difference in the warnings produced by the linker and the analyzer, but those could be coalesced into one if the difference between the dot and slash don't matter.

jtschuster:
Can't seem to request reviews, so tagging @sbomer @tlakollo @vitek-karas

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -31,9 +33,32 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
...
+		[System.Diagnostics.CodeAnalysis.SuppressMessage ("MicrosoftCodeAnalysisPerformance", "RS1008:Avoid storing per-compilation data into the fields of a diagnostic analyzer", Justification = "Temporarily stored as a local variable in a lambda, not as a field")]
+		readonly Action<SymbolAnalysisContext> s_typeDerivesFromRucBase = symbolAnalysisContext => {
+			if (symbolAnalysisContext.Symbol is INamedTypeSymbol typeSymbol && !typeSymbol.HasAttribute (RequiresUnreferencedCodeAttribute)) {
+				if (typeSymbol.BaseType is INamedTypeSymbol baseType && baseType.HasAttribute (RequiresUnreferencedCodeAttribute)) {
+					if (baseType.TryGetAttribute (RequiresUnreferencedCodeAttribute, out var requiresUnreferencedCodeAttribute)) {
+						//string message = MessageFormat.FormatRequiresAttributeMismatch (typeSymbol.HasAttribute (RequiresUnreferencedCodeAttribute), false, RequiresUnreferencedCodeAttribute, typeSymbol.GetDisplayName (), baseType.GetDisplayName ());
+						//symbolAnalysisContext.ReportDiagnostic (Diagnostic.Create (
+						//	s_typeDerivesFromRucClassRule,
+						//	typeSymbol.Locations[0],
+						//	message));
```
tlakollo, line 45: Left over comments

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresCapability.cs
```diff
@@ -1228,6 +1228,8 @@ class RequiresOnMethod
 			}
 
 			[ExpectedWarning ("IL2109", "RequiresOnClass/DerivedWithoutRequires", "RequiresOnClass.ClassWithRequires", "--ClassWithRequires--", ProducedBy = ProducedBy.Trimmer)]
+			[ExpectedWarning ("IL2109", "RequiresOnClass.DerivedWithoutRequires", "RequiresOnClass.ClassWithRequires", "--ClassWithRequires--", ProducedBy = ProducedBy.Analyzer)]
```
tlakollo, line 1231: If the warning is produced by both the Trimmer and the Analyzer you can get rid of specifying the ProducedBy which by default is ProducedBy.TrimmerAndAnalyzer
```suggestion
			[ExpectedWarning ("IL2109", "RequiresOnClass.DerivedWithoutRequires", "RequiresOnClass.ClassWithRequires", "--ClassWithRequires--")]
```

tlakollo: Also as a nit some of the changes add an additional new line

src/ILLink.Shared/SharedStrings.resx
```diff
@@ -192,4 +192,8 @@
   <data name="CorrectnessOfCOMCannotBeGuaranteedMessage" xml:space="preserve">
     <value>P/invoke method '{0}' declares a parameter with COM marshalling. Correctness of COM interop cannot be guaranteed after trimming. Interfaces and interface members might be removed.</value>
   </data>
+  <data name="DerivedClassRequiresUnreferencedCodeAttributeMismatchMessage" xml:space="preserve">
+    <value>{0}: Type '{0}' derives from '{1}' which has 'RequiresUnreferencedCodeAttribute'.  --{2}--.</value>
+    <comment>IL2109 for RequiresUnreferenedCode only</comment>
+  </data>
```
tlakollo, line 198: Im looking at linker code and there is already a SharedString for this warning see [MarkStep](https://github.com/dotnet/linker/blob/main/src/linker/Linker.Steps/MarkStep.cs#L1920-L1924) we should be able to reuse that

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresCapability.cs
tlakollo: I just read the comment on why they are duplicated, seems to me like we should open something on linker to track this discrepancy, it doesn't make a lot of sense that we first print "/" and then "." for the same message.

jtschuster: Made a new issue for it: https://github.com/dotnet/linker/issues/2441

It looks like FullName and GetDisplayName() are the source of the discrepancy

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -32,8 +34,28 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
 				operationContext.Operation.Syntax.GetLocation ()));
 		};
 
+		private Action<SymbolAnalysisContext> typeDerivesFromRucBase {
+			get {
+				return symbolAnalysisContext => {
+					if (symbolAnalysisContext.Symbol is INamedTypeSymbol typeSymbol && !typeSymbol.HasAttribute (RequiresUnreferencedCodeAttribute)) {
+						if (typeSymbol.BaseType is INamedTypeSymbol baseType && baseType.HasAttribute (RequiresUnreferencedCodeAttribute)) {
+							if (baseType.TryGetAttribute (RequiresUnreferencedCodeAttribute, out var requiresUnreferencedCodeAttribute)) {
```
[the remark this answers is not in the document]
tlakollo: Also not sure if we need the check 'is INamedTypeSymbol' for the symbolAnalysisContext.Symbol since you register the SymbolAnalysisContext per SymbolKind.NamedType it should always be a INamedTypeSymbol. For BaseType is fine since is doing the null check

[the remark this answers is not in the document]
jtschuster: It looks like we'd still need to cast the Symbol to an `INamedTypeSymbol` to access the `BaseType` property. I'm not exactly sure, but I think a cast would do the same runtime checks as the `symbolAnalysisContext.Symbol is INamedTypeSymbol typeSymbol` and I personally prefer that notation. Let me know if I'm wrong about that, though.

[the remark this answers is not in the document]
tlakollo: You are right in the Analyzer base we cast the symbol disregard my comment about that

vitek-karas (approved):
Looks good to me - please also wait for Tlaka's sign off.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -32,8 +34,26 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
 				operationContext.Operation.Syntax.GetLocation ()));
 		};
 
+		private Action<SymbolAnalysisContext> typeDerivesFromRucBase {
+			get {
+				return symbolAnalysisContext => {
+					if (symbolAnalysisContext.Symbol is INamedTypeSymbol typeSymbol && !typeSymbol.HasAttribute (RequiresUnreferencedCodeAttribute)
+						&& typeSymbol.BaseType is INamedTypeSymbol baseType && baseType.HasAttribute (RequiresUnreferencedCodeAttribute)
+						&& baseType.TryGetAttribute (RequiresUnreferencedCodeAttribute, out var requiresUnreferencedCodeAttribute)) {
```
tlakollo, line 42: Nit: I don't think we should call HasAttribute and then TryGetAttribute we can just call TryGetAttribute and if false just bail out

tlakollo (approved):

PR #2445, opened 2021-12-14
Add TargetHasRequiresAttribute to RequiresISymbolExtensions file

tlakollo:
Add TargetHasRequiresAttribute to RequiresISymbolExtensions file
Add a RequiresUnreferencedCodeUtils file so other analyzers dont have to pass the RUC name everytime they use the RequiresISymbolExtensions
Refactor the RequiresAnalyzerBase to not have to verify the arguments while getting the attribute since the specific details on how to validate the attribute arguments are specific of the attribute implementation
Change COMAnalyzer to use the RequiresUnreferencedCodeUtils instead of the RequiresISymbolExtensions

This PR will help the DynamicallyAccessedMembersAnalyzer to produce RUC warnings

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
@@ -74,8 +74,6 @@ namespace ILLink.RoslynAnalyzer
 		private static bool TryGetRequiresAttribute (ISymbol member, string requiresAttribute, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData)
 		{
 			requiresAttributeData = null;
-			if (member == null)
-				return false;
 
 			if (!member.TryGetAttribute (requiresAttribute, out var attribute))
 				return false;
--- a/src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
@@ -71,7 +71,7 @@ namespace ILLink.RoslynAnalyzer
 		/// <param name="member">Symbol of the member to search attribute.</param>
 		/// <param name="requiresAttribute">Output variable in case of matching Requires attribute.</param>
 		/// <returns>True if the member contains a Requires attribute; otherwise, returns false.</returns>
-		private static bool TryGetRequiresAttribute (ISymbol? member, string requiresAttribute, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData)
+		private static bool TryGetRequiresAttribute (ISymbol member, string requiresAttribute, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData)
 		{
 			requiresAttributeData = null;
 			if (member == null)
--- a/src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
@@ -1,12 +1,33 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics.CodeAnalysis;
 using Microsoft.CodeAnalysis;
 
 namespace ILLink.RoslynAnalyzer
 {
 	public static class RequiresISymbolExtensions
 	{
+		// TODO: Consider sharing with linker DoesMethodRequireUnreferencedCode method
+		/// <summary>
+		/// True if the target of a call is considered to be annotated with the Requires... attribute
+		/// </summary>
+		public static bool TargetHasRequiresAttribute (this ISymbol member, string requiresAttribute, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData)
+		{
+			requiresAttributeData = null;
+			if (member.IsStaticConstructor ())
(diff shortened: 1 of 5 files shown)


src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
```diff
@@ -43,5 +64,24 @@ private static bool IsInRequiresScope (this ISymbol member, string requiresAttri
...
 			return false;
 		}
+
+		/// <summary>
+		/// This method determines if the member has a Requires attribute and returns it in the variable requiresAttribute.
+		/// </summary>
+		/// <param name="member">Symbol of the member to search attribute.</param>
+		/// <param name="requiresAttribute">Output variable in case of matching Requires attribute.</param>
+		/// <returns>True if the member contains a Requires attribute; otherwise, returns false.</returns>
+		private static bool TryGetRequiresAttribute (ISymbol? member, string requiresAttribute, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData)
```
vitek-karas, line 74: Just curious - adding the nullcheck for `member` here is different from the existing code. I assume it's intentional and attempts to make the code more robust, right?

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeUtils.cs
```diff
@@ -0,0 +1,41 @@
...
+		// TODO: Consider sharing with linker DoesMethodRequireUnreferencedCode method
+		/// <summary>
+		/// True if the target of a call is considered to be annotated with the RequiresUnreferencedCode attribute
+		/// </summary>
+		public static bool TargetHasRequiresUnreferencedCodeAttribute (this ISymbol member, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData)
+		{
+			if (member.TargetHasRequiresAttribute (RequiresUnreferencedCodeAttribute, out requiresAttributeData))
+				return true;
+			return false;
+		}
```
vitek-karas, line 22:
```suggestion
		public static bool TargetHasRequiresUnreferencedCodeAttribute (this ISymbol member, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData)
		=> member.TargetHasRequiresAttribute (RequiresUnreferencedCodeAttribute, out requiresAttributeData);
```

vitek-karas (approved):

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
jtschuster: Also curious if we need this. I'm not sure we ever need the added ability to be able to call this on a nullable `ISymbol?` variable, it looks like we only call this from places where `member` is not null.

vitek-karas: I think I agree - the nullable annotations on all of the callsites seem to suggest that the input can never be null.

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
```diff
@@ -43,5 +64,24 @@ private static bool IsInRequiresScope (this ISymbol member, string requiresAttri
...
+		/// This method determines if the member has a Requires attribute and returns it in the variable requiresAttribute.
+		/// </summary>
+		/// <param name="member">Symbol of the member to search attribute.</param>
+		/// <param name="requiresAttribute">Output variable in case of matching Requires attribute.</param>
+		/// <returns>True if the member contains a Requires attribute; otherwise, returns false.</returns>
+		private static bool TryGetRequiresAttribute (ISymbol member, string requiresAttribute, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData)
+		{
+			requiresAttributeData = null;
+			if (member == null)
+				return false;
```
vitek-karas, line 78: This should not be necessary, since member is declared as non-nullable.

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
```diff
@@ -43,5 +64,22 @@ private static bool IsInRequiresScope (this ISymbol member, string requiresAttri
...
 			return false;
 		}
+
+		/// <summary>
+		/// This method determines if the member has a Requires attribute and returns it in the variable requiresAttribute.
+		/// </summary>
+		/// <param name="member">Symbol of the member to search attribute.</param>
+		/// <param name="requiresAttribute">Output variable in case of matching Requires attribute.</param>
+		/// <returns>True if the member contains a Requires attribute; otherwise, returns false.</returns>
+		private static bool TryGetRequiresAttribute (ISymbol member, string requiresAttribute, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData)
```
jtschuster, line 74: I think it would be the same here to put:
```C# 
private static bool TryGetRequiresAttribute (ISymbol member, string requiresAttribute, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttributeData) => member.TryGetAttribute(requiresAttribute, requiresAttributeData);
```
or since it's private, we could even just replace the references to `TryGetRequiresAttribute` with `TryGetAttribute`.

tlakollo: I enabled the automerge and got merged but I will create another PR for this

PR #2447, opened 2021-12-15
[feature/damAnalyzer] Remove "reflection access pattern" logic from linker

vitek-karas:
Removes the `IReflectionPatternRecorder` and all related classes. This interface has been introduced at the very beginning of the data flow work in the linker. Back then the purpose of it was to separate the marking logic from diagnostic reporting logic. We thought we would be able to report nice virtual callstacks of warnings and so on. These ideas were superseded by the current design of forcing explicit annotations on method boundaries and thus avoiding the need to report warnings correlated across multiple methods.

As such the only consumer of this interface is a "logger" which directly outputs any unrecognized patterns as warnings. The product makes no use of the "Recognized pattern" event at all.

The only remaining logic is the ability to suppress diagnostic generation within the data flow logic (typically due to `RequiresUnreferencedCode` attribute on the method). For now I kept this by adding an `AnalysisContext` helper struct to pass around the message origin and diagnostics settings and logging warnings through it. We will likely need an abstraction for diagnostics to be able to share code with the analyzer anyway, so this is probably a step in the right direction.

This is almost entirely a mechanical change, in general simplifying the code.


vitek-karas:
I strongly suggest hiding whitespace different to review the `ReflectionMethodBodyScanner.cs` changes. This change removes a try/finally around the entire body, effectively changing whitespaces on every line. Combined with actual code changes on those same lines, this makes the diff with whitespaces unusable (the entire method is basically marked as changed).

Hide whitespace diffs for the diff to make sense:
image

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1824,7 +1751,7 @@ bool IsComInterop (IMarshalInfoProvider marshalInfoProvider, TypeReference param
 			return false;
 		}
 
-		bool AnalyzeGenericInstantiationTypeArray (in MultiValue arrayParam, ref ReflectionPatternContext reflectionContext, MethodReference calledMethod, IList<GenericParameter> genericParameters)
+		bool AnalyzeGenericInstantiationTypeArray (in AnalysisContext context, in MultiValue arrayParam, MethodReference calledMethod, IList<GenericParameter> genericParameters)
```
tlakollo: Nit: Maybe we should keep naming it analysisContext to not confuse it with the linker context

tlakollo (approved):
I'm not 100% familiar with the calls in which we create a new AnalysisContext and I remember there was a discussion on why they needed an immutable reference for the analysis context which I didn't fully understand but in general, the changes look fine.

vitek-karas:
Before this change the context struct was actually mutable, since we used state in it to track the completeness of all the checks. That's why it was passed around as `ref` all the time (stack allocated, but not copied). The new struct is `readonly` and thus immutable and so we avoid the `ref` everywhere. Other than that it should not be really different.

PR #2449, opened 2021-12-15
Remove cecil FullName property from diagnostic messages

jtschuster:
Diagnostic messages would occasionally use the Cecil `FullName` property which uses '/' as a delimiter for nested classes. The standard is to use '.' instead. This fix uses `GetDisplayName()` instead to accomplish this.

Fixes https://github.com/dotnet/linker/issues/2441

### DIFF
--- a/src/linker/Linker/MessageContainer.cs
+++ b/src/linker/Linker/MessageContainer.cs
@@ -256,10 +256,8 @@ namespace Mono.Linker
 			if (Origin?.Provider != null) {
 				if (Origin?.Provider is MethodDefinition method)
 					sb.Append (method.GetDisplayName ());
-				else if (Origin?.Provider is TypeDefinition type)
+				else if (Origin?.Provider is MemberReference type)
 					sb.Append (type.GetDisplayName ());
-				else if (Origin?.Provider is IMemberDefinition member)
-					sb.Append (member.FullName);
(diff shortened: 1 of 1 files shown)


src/linker/Linker/MessageContainer.cs
```diff
@@ -256,10 +256,8 @@ public string ToMSBuildString ()
 			if (Origin?.Provider != null) {
 				if (Origin?.Provider is MethodDefinition method)
 					sb.Append (method.GetDisplayName ());
-				else if (Origin?.Provider is TypeDefinition type)
+				else if (Origin?.Provider is MemberReference type)
```
vitek-karas, line 259: While I think this is in general correct, it feels dangerous. If there can be an instance of `IMemberDefinition` which is not `MemberReference` we would now throw `NotSupportedException`. Which would be a regression. Quickly looking through Cecil, this should not be the case, but I don't know for sure.
Would be safer to keep the `IMemberDefinition` case as a fallback.

src/linker/Linker/MessageContainer.cs
```diff
@@ -256,10 +256,10 @@ public string ToMSBuildString ()
...
```
tlakollo, line 259: nit: probably change the name from type a memberReference could be several things

vitek-karas (approved):

tlakollo (approved):

PR #2455, opened 2021-12-16
Analyzer warns when Requires... Attribute is on a static constructor

### LINKED ISSUE #2347: Analyzer allows RUC/RAF in static constructor
Trimmer generates warning IL2116 when RUC is placed in a static constructor but analyzer doesn't do anything in those cases, moreover, the behavior for RAF is undefined since it doesn't support RAF on type

### LINKED ISSUE #2446: ExpectedWarning doesn't check for warnings on constructors
ExpectedWarning attribute will not check for warnings when on a static constructor. For example, the following test will pass:
RequiresCapability.cs:
```C#
class ClassWithConstructors
{
	[ExpectedWarning ("WarningThatNeverComes", "LoremIpsum")]
	static ClassWithConstructors()
	{
	}

	[ExpectedWarning("NoWarningWillCome", "asdfjkl;")]
	public ClassWithConstructors() { }
}
```

jtschuster:
Adds warnings in the analyzer for IL2116 (RUC not allowed on static ctor) and adds IL2117 (RequiresDynamicCode not allowed on static ctor) and IL2118 (RequiresAssemblyFiles not allowed on static ctor). 

Fixes ExpectedWarning not having effects on constructors by adding a `Visit` override in the compilation tree walker in the test infra. 

Adds case in GetDisplayName to differentiate .cctor's and .ctor's.

Fixes #2446
Fixes #2347

### Customer Impact
Analyzer correctly reports warnings due to `RequiresUnreferencedCode`, `RequiresDynamicCode` and `RequiresAssemblyFiles` on static constructors. For `RequriesUnreferencedCode` the warnings would be reported by trimmer during publish. But for `RequiresDynamicCode` and `RequiresAssemblyFiles` nothing would report these warnings.

### Testing
Added targeted tests.

### Risk
Low - added a check for new case (static .ctor) and the new diagnostics in the analyzer.


tlakollo:
There are some ranges for warnings described in [DiagnosticDescriptors](https://github.com/dotnet/linker/blob/main/src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs#L59-L76) this along with currently registered warnings in [DiagnosticIds](https://github.com/dotnet/linker/blob/main/src/ILLink.Shared/DiagnosticId.cs) means that the warning produced by RequiresDynamicCode should be IL3052 and by RequiresAssemblyFiles should be IL3004

src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
```diff
@@ -16,6 +16,12 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(MicrosoftCodeAnalysisVersion)" PrivateAssets="all" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="Mono.Cecil">
+      <HintPath>..\..\artifacts\bin\Mono.Cecil\Debug\netstandard2.0\Mono.Cecil.dll</HintPath>
```
tlakollo, line 21: We shouldnt have Mono.Cecil in here

src/ILLink.RoslynAnalyzer/IMethodSymbolExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+		private static void PrependGenericParameters(ImmutableArray<ITypeParameterSymbol> genericParameters, System.Text.StringBuilder sb)
+		{
+			sb.Insert (0, '>').Insert (0, genericParameters[genericParameters.Length - 1]);
+			for (int i = genericParameters.Length - 2; i >= 0; i--)
+				sb.Insert (0, ',').Insert (0, genericParameters[i]);
+
+			sb.Insert (0, '<');
+		}
+
+		public static string GetDisplayName (this IMethodSymbol method)
```
tlakollo, line 38: The GetDisplayName method was made to match C# naming which is what Roslyn uses and most people are used to reading instead of IL namings. Therefore if there is a discrepancy between the way it's represented by the analyzer and linker is likely a fix in the GetDisplayName in the linker not an implementation on the analyzer

tlakollo: Worst case I think you can append directly the string ".cctor" at the end when displaying the diagnostic

vitek-karas: I would expect most cases to already be covered by Roslyn. I could understand that we need to do something special for property getter/setter (and events), and possibly even .cctor, but other than that I would expect to simply call Roslyn to do this for us.

jtschuster: For now, would it be best to use the ISymbolExtensions.GetDisplayName and have separate ExpectedWarning's for the analyzer and trimmer, and then later change the trimmer's GetDisplayName to match the Analyzer/Rozlyn?

jtschuster: I added a check in the GetDisplayName in the linker that exists for ctors, but was missing cctors. Now they should look the same.

tlakollo: I think I was wrong is not a fix on the linker side, because if you fix in the linker then constructor and static constructor are represented equally, there is no way to differentiate them. So in the cases like the BeforeInitField test, I think things can get confusing. Same for RequiresInCompilerGeneratedCode tests, they get more confusing although you don't see the changes because we don't check for the caller name only messages in the attribute.
But yeah I think the best place to fix it is ISymbolExtensions.GetDisplayName an if statement before printing the method name check if the method symbol is a static constructor

vitek-karas: .cctors are weird - so we might need to add a special case unfortunately.

jtschuster: I recently pushed a couple commits with different formats for the diagnostic messages. [The first one](https://github.com/dotnet/linker/pull/2455/commits/da05d3f6d36b153e20196a264daeb35c53c85fd4) includes the `static` modifier (e.g. `static Mono.Linker.Tests.Cases.RequiresCapability.RequiresCapability.StaticCtor.StaticCtor()`). [The second](https://github.com/dotnet/linker/pull/2455/commits/04270d67e306947a3d9a2aa82d65bf0f6325840b) uses the `..cctor()` notation (e.g. `Mono.Linker.Tests.Cases.RequiresCapability.RequiresCapability.StaticCtor..cctor()`).

I think the static modifier notation would be best since it still differentiates the constructors and is more familiar to most people.

tlakollo: I just created something in SharpLab to see how a warning would be printed
```
public class Foo
{
    static Foo (int a) {}
}
```
This generates `error CS0132: 'Foo.Foo(int)': a static constructor must be parameterless` also inserting the [Obsolete] attribute will generate a message `[deprecated] Foo.Foo()` so the compiler warnings at this moment don't show the modifiers nor have a clear distinction between static and instance constructors, so whatever decision we make is going to deviate from whatever the compiler normally prints.

I think is key for the trimmer to have some differentiation but I don't have a preference for which one to choose
- The first option I have the concern that not a lot of people will even see the static modifier, it's at the beginning of the string and no other member will print it. Although I agree, I like more the readability of it.
- The second option is a weird notation, but static constructors are weird, we don't even allow them to annotate them. Static constructors are handled in a special way. So the fact that the name is weird might even be beneficial.

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
```diff
@@ -72,6 +71,12 @@ public override void VisitClassDeclaration (ClassDeclarationSyntax node)
 			CheckMember (node);
 		}
 
+		public override void VisitConstructorDeclaration (ConstructorDeclarationSyntax node)
```
jtschuster, line 74: Fix for #2446

src/ILLink.Shared/SharedStrings.resx
```diff
@@ -189,6 +189,17 @@
...
     <value>The use of 'RequiresUnreferencedCodeAttribute' on static constructors is disallowed since is a method not callable by the user, is only called by the runtime. Placing the attribute directly on the static constructor will have no effect, instead use 'RequiresUnreferencedCodeAttribute' on the type which will handle warning and silencing from the static constructor.</value>
   </data>
+  <data name="RequiresDynamicCodeOnStaticConstructorMessage" xml:space="preserve">
+    <value>'RequiresDynamicCodeAttribute' cannot be placed directly on static constructor '{0}'.</value>
+  </data>
+  <data name="RequiresDynamicCodeOnStaticConstructorTitle" xml:space="preserve">
+    <value>The use of 'RequiresDynamicCodeAttribute' on static constructors is disallowed since is a method not callable by the user, is only called by the runtime. Placing the attribute directly on the static constructor will have no effect, instead use 'RequiresUnreferencedCodeAttribute' on the type which will handle warning and silencing from the static constructor.</value>
+  </data><data name="RequiresAssemblyFilesOnStaticConstructorMessage" xml:space="preserve">
```
tlakollo, line 197: Nit: add an end of line to separate the <data>

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
tlakollo: Can you add a test for this? Something that Inside the static constructor calls a method annotated with RUC and just verify that we can use the ExpectedWarning attribute

jtschuster: Good thing you mentioned this, I think I found a bug that the linker doesn't check in constructors for calls to methods with RUC. The test class for this is `WarningsInCtor` in RequiresCapability.cs. It calls a method annotated with RUC inside the constructors, but the linker doesn't seem to be producing a warning. I filed https://github.com/dotnet/linker/issues/2484 to track the issue.

src/ILLink.Shared/DiagnosticId.cs
```diff
@@ -11,11 +11,14 @@ public enum DiagnosticId
 		RequiresOnBaseClass = 2109,
 		RequiresUnreferencedCodeOnStaticConstructor = 2116,
 
+
```
tlakollo, line 14: Nit: extra line

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
tlakollo: I think its because nothing references that code in the main method, since it's not referenced Linker will get rid of that piece of code

tlakollo (approved):
Other than lint, looks good to me. Thanks :D

tlakollo:
Does another person want to sign off on these changes so they can be merged?

vitek-karas (approved):

src/ILLink.RoslynAnalyzer/RequiresDynamicCodeAnalyzer.cs
```diff
@@ -11,14 +11,16 @@ namespace ILLink.RoslynAnalyzer
 	[DiagnosticAnalyzer (LanguageNames.CSharp)]
 	public sealed class RequiresDynamicCodeAnalyzer : RequiresAnalyzerBase
 	{
+
```
tlakollo, line 14: nit: extra line

PR #2461, opened 2021-12-20
GetDisplayName uses 'ref Type' instead of 'Type&'

jtschuster:
Modifies the behavior of GetDisplayName in the Linker to use the 'ref Type' format instead of the 'Type&' that is default in Cecil. Fixes https://github.com/dotnet/linker/issues/2406


tlakollo (approved):
Looks good to me, but we will need to follow up with this change in case merged to see nothing breaks in the SDK and runtime. Maybe someone suppresses something and now it will start failing

marek-safar:
What is the motivation for this change?

jtschuster:
> What is the motivation for this change?

@marek-safar The original issue was filed in response to [this discussion](https://github.com/dotnet/linker/pull/2390#:~:text=Could%20you%20add%20an%20issue%20for%20this.%20My%20understand%20is%20that%20we%20tried%20to%20match%20the%20C%23%20formatting%20in%20warnings%2C%20so%20if%20there%27s%20a%20discrepancy%2C%20it%20should%20be%20treated%20as%20a%20bug%20in%20the%20linker.). I think in general we're trying to have our notation match C# warning notation.

MichalStrehovsky:
> > What is the motivation for this change?
> 
> @marek-safar The original issue was filed in response to [this discussion](https://github.com/dotnet/linker/pull/2390#:~:text=Could%20you%20add%20an%20issue%20for%20this.%20My%20understand%20is%20that%20we%20tried%20to%20match%20the%20C%23%20formatting%20in%20warnings%2C%20so%20if%20there%27s%20a%20discrepancy%2C%20it%20should%20be%20treated%20as%20a%20bug%20in%20the%20linker.). I think in general we're trying to have our notation match C# warning notation.

I'm not a big fan of doing this. `&` is universally correct because it matches what the reflection stack would write for byref parameters.

This change makes the warning confusing for a C# developer if they actually had `in`, `out`, or `ref readonly` instead of `ref` in the signature. To be not confusing, we would also have to consider the difference from `in` and `out`. But do note that `in` and `ref readonly` is pretty recent, C# might add more keywords in the future, and we'll need to play catch up or keep confusing users.

VB and F# devs will have no clue what `ref`/`in`/`out` mean in either case.

jtschuster:
> > > What is the motivation for this change?
> > 
> > 
> > @marek-safar The original issue was filed in response to [this discussion](https://github.com/dotnet/linker/pull/2390#:~:text=Could%20you%20add%20an%20issue%20for%20this.%20My%20understand%20is%20that%20we%20tried%20to%20match%20the%20C%23%20formatting%20in%20warnings%2C%20so%20if%20there%27s%20a%20discrepancy%2C%20it%20should%20be%20treated%20as%20a%20bug%20in%20the%20linker.). I think in general we're trying to have our notation match C# warning notation.
> 
> I'm not a big fan of doing this. `&` is universally correct because it matches what the reflection stack would write for byref parameters.
> 
> This change makes the warning confusing for a C# developer if they actually had `in`, `out`, or `ref readonly` instead of `ref` in the signature. To be not confusing, we would also have to consider the difference from `in` and `out`. But do note that `in` and `ref readonly` is pretty recent, C# might add more keywords in the future, and we'll need to play catch up or keep confusing users.
> 
> VB and F# devs will have no clue what `ref`/`in`/`out` mean in either case.

That's a really good point, if the linker deals with IL that could be from any .NET language, should we ever be using C# notation in linker diagnostics? And if we should use IL notation, do we want C# analyzers producing slightly different diagnostics that are C# specific and more understandable? I've been getting the impression we want the linker and the analyzers to produce identical diagnostics and use C#/Roslyn notation, but maybe we should rethink that.
@tlakollo @vitek-karas @sbomer @agocke thoughts?

marek-safar:
> I'm not a big fan of doing this. & is universally correct because it matches what the reflection stack would write for byref parameters.

I agree with Michal, ILLinker operates at IL/metadata level and the closest match to that is reflection. I think all error messages that come from illinker (not analyzers) should use reflection notation (or something very close to it).

We could explore the option of customizing the output based on the compiler used to produce the assembly but I'm really not sure it's worth it and it still has shortcomings because there are many different compilers (e.g. iron python).

Another option would be the explicit `--diagnostic csharp` option but that for me would make sense if we enhanced C# compiler to produce more metadadata for illinker to easily map any metadata to C# syntax.

sbomer:
@jtschuster thanks for kicking off a good discussion with this change. I made https://github.com/dotnet/linker/issues/2406 a tracking issue for changing the linker warnings to use reflection notation.

PR #2469, opened 2021-12-29
Add license header to files that have no header

tlakollo:
This PR adds the following message as header
```C#
// Licensed to the .NET Foundation under one or more agreements.
```
for files that dont contain any license information

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -101,8 +101,8 @@ dotnet_diagnostic.IDE0060.severity = warning
 dotnet_diagnostic.IDE0065.severity = warning
 
 # IDE0073: File header
-dotnet_diagnostic.IDE0073.severity = suggestion
-file_header_template = Licensed to the .NET Foundation under one or more agreements.\nThe .NET Foundation licenses this file to you under the MIT license.
+dotnet_diagnostic.IDE0073.severity = warning
+file_header_template = Copyright (c) .NET Foundation and contributors. All rights reserved.\nLicensed under the MIT license. See LICENSE file in the project root for full license information.
 
 # IDE0074: Use compound assignment
 dotnet_diagnostic.IDE0074.severity = warning
@@ -147,8 +147,13 @@ dotnet_diagnostic.CA1822.severity = none
 dotnet_diagnostic.IDE0060.severity = none
 
 [test/**/*.cs]
-
 dotnet_diagnostic.IDE0060.severity = none
+
+[test/Mono.Linker.Tests/TestCases/Dependencies/WarningSuppressionExpectations*.cs]
+dotnet_diagnostic.IDE0073.severity = none
+
+[external**]
+dotnet_diagnostic.IDE0073.severity = none
 
 [external**]
 dotnet_analyzer_diagnostic.severity = none
--- a/THIRD-PARTY-NOTICES.TXT
+++ b/THIRD-PARTY-NOTICES.TXT
@@ -1,0 +1,75 @@
+.NET Linker uses third-party libraries or other resources that may be
+distributed under licenses different than the .NET Linker software.
+
+In the event that we accidentally failed to list a required notice, please
+bring it to our attention. Post an issue or email us:
+
+           dotnet@microsoft.com
+
+The attached notices are provided for information only.
+
+License notice for Xamarin and Novell
+-------------------------------------
+
+Copyright (c) 2015 Xamarin, Inc (http://www.xamarin.com)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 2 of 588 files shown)


tlakollo:
The runtime repository has license headers for tests and files. Currently, the linker repo does not have a license displayed in all files, with most of testing files not having any header. Some concern brought by @MichalStrehovsky is that some of these files may precede the acquisition made by Microsoft and in that case is not clear if we can copyright them. 
Having said that, the files under some other license haven't been changed and this PR only adds to file with no license explicitly called

marek-safar:
I'd not be too concerned with copyright for test files, especially if they are small.

@mrvoorhe could you comment with explicit approval too?

vitek-karas:
I'm perfectly OK adding the license header to everything - makes it simple and consistent.
That said, maybe we should change the editor config to make this into a warning:
https://github.com/dotnet/linker/blob/1a6468ff722c8d362f37638989ec01ab9975ac28/.editorconfig#L103-L105

In that case linter would catch missing headers and fail the CI (and running it locally should actually add the header there for you).

tlakollo:
> I'm perfectly OK adding the license header to everything - makes it simple and consistent. 

What about files that currently have a different license header for example [AssemblyResolver.cs](https://github.com/dotnet/linker/blob/main/src/linker/Linker/AssemblyResolver.cs) which has a Novell license?

vitek-karas:
I guess we will need to involve somebody who understands licensing. The LICENSE.txt in the root doesn't mention Novell at all, it's purely a .NET Foundation license. I assume that we would need to either change the license for the file or add something to the main license.

MichalStrehovsky:
I would not touch the files that have existing license headers - see https://github.com/dotnet/runtime/issues/63146#issuecomment-1001681013.

I don't know if we have existing guidance from CELA on source files that are old, but don't have any licensing header whatsoever.

It should be fine to add the header to any files added in the past ~5 years (and it is our policy to have those headers).

tlakollo:
Following CELA guidance we can:
- Add Novel to our notice file
- Add the DFN license information at the start of the file once Microsoft had made changes to the file

tlakollo:
Files that still have previous license headers and haven't been modified are:
- linker\src\linker\Linker\TypePreserve.cs
- linker\src\linker\Linker\Tracer.cs
- linker\src\linker\Linker\MethodAction.cs
- linker\src\linker\Linker.Steps\RegenerateGuidStep.cs

Files from project:
- linker\src\analyzer\analyzer.csproj

vitek-karas:
If these files are not modified, then we should keep them as-is.
For the rest I think this makes sense. But please get a review from @richlander at least.

richlander:
> The LICENSE.txt in the root doesn't mention Novell at all

That's what the 3PN is for.

Here are some good rules:

- Do: Add .NET Foundation copyright notice to all files, including those with no notice.
- Do: Add .NET Foundation notice to the top even if there are other notices. That's just a sanity win for consistency for our project.
- Do: Add .NET Foundation notice for files we never modified. It makes it clear that these files are covered by our MIT license and our 3PN. Also, it just helps with our sanity. Let's avoid special rules where we can. Also, if someone complains about this practice, we can address it at the time. We can beg for forgiveness based on a legit intent to improve clarity.
- Do not: remove/modify the Novell notice (ever).
- Do: If something feels odd of wrong, ask.

src/linker/Linker/IReflectionPatternRecorder.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
sbomer, line 1: Looks like an older file got left over.

src/linker/Linker/LoggingReflectionPatternRecorder.cs
```diff
@@ -26,24 +26,29 @@
...
-
+using ILLink.Shared;
 using Mono.Cecil;
+using Mono.Cecil.Cil;
 
 namespace Mono.Linker
 {
-
-	public interface IXApiVisitor
+	class LoggingReflectionPatternRecorder : IReflectionPatternRecorder
```
sbomer, line 35: Unrelated changes

test/Mono.Linker.Tests.Cases.Expectations/Assertions/VerifyAllReflectionAccessPatternsAreHandledAttribute.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
sbomer, line 1: Unrelated file

test/Mono.Linker.Tests/TestCasesRunner/TestReflectionPatternRecorder.cs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
```
sbomer, line 1: Unrelated file

.editorconfig
```diff
@@ -150,6 +150,10 @@ dotnet_diagnostic.IDE0060.severity = none
 
 dotnet_diagnostic.IDE0060.severity = none
 
+[test/Mono.Linker.Tests/TestCases/Dependencies/WarningSuppressionExpectations*.cs]
```
sbomer, line 153: Is there a reason to exclude just these files, or is it left over?

sbomer (commented):
Thanks! It's probably worth re-doing the changes on top of the latest commit from a fresh state to make sure there are no unrelated changes introduced by accident.

.editorconfig
tlakollo: These files are used as 'expectations' of the generated suppressions files by the trimmer. The trimmer produces suppressions files but doesn't include the header on them, therefore adding the header to this expectation files will make a test to fail.

tlakollo:
@richlander I think this PR is ready for review, PTAL

tlakollo:
Every commit now is potentially a merge conflict, so it will be nice to merge this as soon as possible to prevent unnecessary rework

richlander (approved):

PR #2472, opened 2021-12-31
Adds Error Ids into ILLink.Shared project

tlakollo:


marek-safar:
What's the purpose of this change?

tlakollo:
My goal is to have all the ids registered, only divided for simplicity on review. Having all the ids registered will give the following advantages:
- Warning/error printing is consistent inside linker
- If any tool named Analyzer, NativeAOT, or other attempts to reuse the message they just have to create their own. See for example the [diff](https://github.com/dotnet/runtime/commit/a8e198c3306ef67d7c1ae1088d818064ede56179#diff-a69e39fd41b5056fbdde00840af5e23730a9499fdb74070691e1773075b33bc6R18) NativeAOT reuses some diagnostics that linker still doesn't transfer to the DiagnosticId/SharedStrings files.
- Eventually can get a translation for the messages since a resx file is used

I don't see a reason to hold up with adding more values to the enum, but I could be wrong

vitek-karas:
The simplest way to describe this change:
Move some of the diagnostic messages (and associated warning codes) into the shared project, so that the same enum/strings can be used by both the linker and the analyzer (and NativeAOT in the future).

vitek-karas (approved):

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -235,7 +236,8 @@ CustomAttributeArgument[] ReadCustomAttributeArguments (XPathNavigator nav, Type
 			case MetadataType.Object:
 				var argumentIterator = nav.SelectChildren ("argument", string.Empty);
 				if (argumentIterator?.MoveNext () != true) {
-					_context.LogError ($"Custom attribute argument for 'System.Object' requires nested 'argument' node.", 1043);
+					_context.LogError (new DiagnosticString (DiagnosticId.CustomAttributeArgumentForTypeRequiresNestedNode).GetMessage ("System.Object", "argument"),
```
marek-safar, line 239: This looks quite ugly, can you add overload which can simplify the code to something like

```c#
_context.LogError (DiagnosticId.CustomAttributeArgumentForTypeRequiresNestedNode, "System.Object", "argument");
```

tlakollo: I couldn't implement exactly this behavior but I implemented something very similar. PTAL

src/linker/Linker/LinkContext.cs
```diff
@@ -613,6 +614,20 @@ public void LogError (string text, int code, string subcategory = MessageSubCate
...
 
+		/// <summary>
+		/// Display an error message to the end user.
+		/// </summary>
+		/// <param name="id">Unique error ID. Please see https://github.com/dotnet/linker/blob/main/docs/error-codes.md and https://github.com/dotnet/linker/blob/main/src/ILLink.Shared/DiagnosticId.cs for the list of errors and possibly add a new one</param>
+		/// <param name="subcategory">Optionally, further categorize this error</param>
+		/// <param name="origin">Filename, line, and column where the error was found</param>
+		/// <param name="args">Additional arguments to form a humanly readable message describing the warning</param>
+		/// <returns>New MessageContainer of 'Error' category</returns>
+		public void LogError (DiagnosticId id, string subcategory = MessageSubCategory.None, MessageOrigin? origin = null, params string[] args)
```
vitek-karas, line 625: I think this is probably good enough as a first try. I think we could go one step further:
* Make the origin parameter non-optional (just nullable). The fact that we don't specify origin in lot of callsites is effectively a bug - it makes the error much harder to fix.
* subcategory - I don't know if we even need it - at least in the "Simple" overload. As far as I can tell we almost never use it. So we could remove it from this overload completely.

And so we could change this to look like:
`LogError(MessageOrigin?, DiagnosticId, params string[] args)`

That way the callsites would not need to allocate string array manually. It does reorder the parameters, but personally I actually like the fact that it puts the origin first - it mimics how the error is presented to the user: "location: Id - message".

tlakollo: I can implement those changes in this PR, the only question I have is: what about the other overloads for Logging errors? should we reorder the parameters too?

marek-safar: I'd not change other overloads for now, only ones that take DiagnosticId

vitek-karas: I agree with Marek - base don a quick scan almost all callsites will go through the new overload, so the existing overloads can be considered "expert level" and I would not change them for now. Eventually we could even remove it in favor of the low-level approach of creating the `MessageContainer` manually at the callsite.

I assume we'll make similar changes to `LogWarning` in the other PR, right?

tlakollo: > I assume we'll make similar changes to `LogWarning` in the other PR, right?

Yes, except that I'm not sure what's going to happen with the subcategory for LogWarning since there is needed

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -235,7 +236,7 @@ CustomAttributeArgument[] ReadCustomAttributeArguments (XPathNavigator nav, Type
 			case MetadataType.Object:
 				var argumentIterator = nav.SelectChildren ("argument", string.Empty);
 				if (argumentIterator?.MoveNext () != true) {
-					_context.LogError ($"Custom attribute argument for 'System.Object' requires nested 'argument' node.", 1043);
+					_context.LogError (null, DiagnosticId.CustomAttributeArgumentForTypeRequiresNestedNode, "System.Object", "argument");
```
marek-safar, line 239: Is there any usage where `null` is not used as the first argument?

vitek-karas: Yes - there are some - ideally all should be non-null, but that's probably for another change.

src/linker/Linker/LinkContext.cs
vitek-karas: In that case I would put is as a second parameter (origin, subcategory, id, args)...

In fact - can we deduce the subcategory from the id? I think we should be able to, right?

src/linker/Linker.Steps/LinkAttributesParser.cs
vitek-karas: For example the one right below.

src/linker/Linker/LinkContext.cs
tlakollo: We could create ranges and assign them a subcategory, but that would be flaky because we can add new diagnostics that don't fit the range.  Either way, there is an ilOffset optional argument that cannot be deduced in any way and is almost always null. Considering that optional arguments must appear after required arguments and that params need to be the last argument there will always be problems. I don't see how to not have to use the named arguments on this one.

vitek-karas: Since ilOffset is not of type `string` adding another overload (instead of optional parameter) should work - the overload resolution should correctly resolve it (since the types are not ambiguous).

PR #2475, opened 2022-01-03
Add warning Ids part 1

tlakollo:
Add warning Ids 2001-2050 into DiagnosticId and SharedString files


src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -228,15 +229,15 @@ TypeAnnotations BuildTypeAnnotations (TypeDefinition type)
...
 							_context.LogWarning (
-								$"The 'DynamicallyAccessedMembersAttribute' is not allowed on methods. It is allowed on method return value or method parameters though.",
-								2041, method, subcategory: MessageSubCategory.TrimAnalysis);
+								new DiagnosticString (DiagnosticId.DynamicallyAccessedMembersIsNotAllowedOnMethods).GetMessage (),
+								(int) DiagnosticId.DynamicallyAccessedMembersIsNotAllowedOnMethods, method, subcategory: MessageSubCategory.TrimAnalysis);
 						}
 					} else {
 						offset = 0;
 						if (methodMemberTypes != DynamicallyAccessedMemberTypes.None) {
 							_context.LogWarning (
```
vitek-karas, line 238: I wonder if we should add an overload to LogWarning which takes the enum instead of int - to avoid the casts in all the callsites.

vitek-karas (approved):
Looks good

src/linker/Linker/LinkContext.cs
```diff
@@ -583,6 +602,22 @@ public void LogWarning (string text, int code, IMemberDefinition origin, int? il
...
+		/// Display a warning message to the end user.
+		/// This API is used for warnings defined in the linker, not by custom steps. Warning
+		/// versions are inferred from the code, and every warning that we define is versioned.
+		/// </summary>
+		/// <param name="text">Humanly readable message describing the warning</param>
+		/// <param name="code">Unique warning ID. Please see https://github.com/dotnet/linker/blob/main/docs/error-codes.md for the list of warnings and possibly add a new one</param>
+		/// <param name="origin">Type or member where the warning is coming from</param>
+		/// <param name="subcategory">Optionally, further categorize this warning</param>
+		/// <returns>New MessageContainer of 'Warning' category</returns>
+		public void LogWarning (string text, DiagnosticId code, IMemberDefinition origin, int? ilOffset = null, string subcategory = MessageSubCategory.None)
```
marek-safar, line 615: Why do you need a string `text` if all the time you do
`new DiagnosticString (code).GetMessage (some other values)` ?

tlakollo: The return value of `new DiagnosticString (code).GetMessage (args)` is a string. GetMessage method takes the DiagnosticString name and looks in SharedStrings.resx file for an entry with `name+Message` then substitutes the args into the SharedStrings.resx value and returns the formed string.

vitek-karas: I think the idea is to have an overload which takes `DiagnosticId` and "params" and internally it does the `new DiagnosticString(code).GetMessage(params)` for you. The point is that almost all call sites in the linker now have the `DiagnosticString(code).GetMessage(params)` code, so to remove that duplication, make it part of the `LogWarning`. Ideally the call site should only pass in things in their simplest form and as few as viable. Currently that's not the case.

src/linker/Linker.Dataflow/ReflectionPatternContext.cs
```diff
@@ -95,6 +96,19 @@ public void RecordUnrecognizedPattern (int messageCode, string message)
...
+		{
+#if DEBUG
+			if (!_patternAnalysisAttempted)
+				throw new InvalidOperationException ($"Internal error: To correctly report all patterns, when starting to analyze a pattern the AnalyzingPattern must be called first. {Source} -> {MemberWithRequirements}");
+
+			_patternReported = true;
+#endif
+
+			if (ReportingEnabled)
+				_context.ReflectionPatternRecorder.UnrecognizedReflectionAccessPattern (Origin, Source, Instruction, MemberWithRequirements, message, (int) messageCode);
```
vitek-karas, line 109: Why not just call the other overload by casting the messageCode? This feels like unnecessary duplication.

tlakollo:
I will merge to continue with the second part of Ids, if any other feedback is requested I will add it to the new PR

PR #2476, opened 2022-01-03
[release/6.0.2xx] Propagate Interfaces annotation through Type.BaseType

vitek-karas:
The Interfaces annotation makes sure the type has all interfaces, meaning `GetInterfaces` reflection call will work on it. That means that all interfaces of the base type are also preserved, so the `Interfaces` annotation should be propagated to the base type as well.

Fixes https://github.com/dotnet/linker/issues/2473.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -1294,6 +1294,9 @@ namespace Mono.Linker.Dataflow
 
 									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicProperties))
 										propagatedMemberTypes |= DynamicallyAccessedMemberTypes.PublicProperties;
+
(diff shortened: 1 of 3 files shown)


MichalStrehovsky (approved):

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1294,6 +1294,9 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 
 									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.PublicProperties))
 										propagatedMemberTypes |= DynamicallyAccessedMemberTypes.PublicProperties;
+
+									if (dynamicallyAccessedMemberNode.DynamicallyAccessedMemberTypes.HasFlag (DynamicallyAccessedMemberTypes.Interfaces))
+										propagatedMemberTypes |= DynamicallyAccessedMemberTypes.Interfaces;
```
agocke, line 1299: Consider creating a mask here and &'ing, instead of checking each flag individually.

agocke: Also, whether inclusion of exclusion is preferred, i.e. if we forget to update this, would we prefer to propagate more or less?

agocke (approved):

Issue #2364, opened 2021-11-11
[DAM analyzer] Report DAM warnings for generics in all cases


sbomer:
This is a tracking issue for support for annotated generics in the analyzer. The DAM analyzer only validates annotations for generics at callsites to generic methods:
https://github.com/dotnet/linker/blob/ff065a359c767514c6fc1595de3c48b5162ca570/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs#L139

This needs to be extended to cover all type references which may reference generic instantiations, including:
- base/interface types of a type declaration
- variable declarations
- field/property/event declarations
- parameter/return types of methods

This is probably best done using the symbol-based callbacks (`RegisterSymbolAction`) rather than the IOperation-based ones.

I would start by enabling some of the tests in GenericParameterDataFlow (or temporarily copying them to DynamicallyAccessedMembersAnalyzerTests.cs if that's easier). For example `DerivedTypeWithOpenGenericOnBase` looks like a good starting point.

sbomer:
Per @agocke's suggestion it might be better to do this using the syntax APIs which would likely cover all of the above cases (whereas using the symbol APIs would only get you callbacks for the symbol _declarations_ and would require a separate approach for handling references to substituted generics inside method bodies).

sbomer:
https://github.com/dotnet/linker/issues/2574 tracks the remaining work to get validation of array types passed as generic parameters. I think all of the other cases are now covered by https://github.com/dotnet/linker/pull/2478 so I will close this issue.

PR #2478, opened 2022-01-04
generic type annotation

LakshanF:
Initial effort for #2364. 

Using Syntax API and refactored `GenericParameterDataFlow `test into 2 tests

Switched to registering `RegisterSyntaxNodeAction `to process `INamedTypeSymbol ` and `IMethodSymbol` for processing generic parameter warnings. Also refactored the `GenericParameterDataFlow ` by keeping test scenarios that handle `MakeGenericXXX `(that is awaiting Intrinsics) in the original test and the rest of test scenarios to `GenericParameterDataFlowWithoutMakeGeneric `and enabled that test in `DataFlowTests`. 

Some comments on exceptions,

- No warning (x2) generated for constructors for generic derived class (does generate for the derived class). These warnings are excluded from Analyzer since Linker can see the base ctor call in the IL.

>   [`DerivedTypeWithOpenGenericOnBase`](https://github.com/dotnet/linker/blob/b7e5e76209349a30888a8d948183137836805565/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlowWithoutMakeGeneric.cs#L177) does not generate a warning 2091.

> [`PartialyInstantiatedMethods `](https://github.com/dotnet/linker/blob/b7e5e76209349a30888a8d948183137836805565/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlowWithoutMakeGeneric.cs#L394)- does not generate a warning 2091.

- Warning message content (x3) was slightly different between the linker and analyzer and was aligned

> [`InnerTypeWithNoAddedGenerics's `](https://github.com/dotnet/linker/blob/b7e5e76209349a30888a8d948183137836805565/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlowWithoutMakeGeneric.cs#L394)

> [`StaticPartialInstantiationUnrecognized `](https://github.com/dotnet/linker/blob/b7e5e76209349a30888a8d948183137836805565/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlowWithoutMakeGeneric.cs#L394)

> Analyzer warning is on [`PublicMethodsProperty `](https://github.com/dotnet/linker/blob/b7e5e76209349a30888a8d948183137836805565/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlowWithoutMakeGeneric.cs#L335) instead of the getter and setter

- Constraint warning (x3) generated were specially handled similarly to the linker but further down the stack,

> Warning 2091 generated for [`RequiresParameterlessConstructor `](https://github.com/dotnet/linker/blob/b7e5e76209349a30888a8d948183137836805565/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlowWithoutMakeGeneric.cs#L687-L699) and are suppressed for following tests: `TestNewConstraintSatisfiesParameterlessConstructor`, `TestStructConstraintSatisfiesParameterlessConstructor`, `TestUnmanagedConstraintSatisfiesParameterlessConstructor`


src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -58,6 +58,37 @@ public override void Initialize (AnalysisContext context)
...
+			for (int i = 0; i < typeParams.Length; i++) {
+				var sourceValue = new SymbolValue (typeArgs[i]);
+				var targetValue = new SymbolValue (typeParams[i]);
+				var damtOnTarget = targetValue.DynamicallyAccessedMemberTypes;
+				var damtOnSource = sourceValue.DynamicallyAccessedMemberTypes;
+				if (Annotations.SourceHasRequiredAnnotations (damtOnSource, damtOnTarget, out var missingAnnotations))
+					continue;
+				var diag = DiagnosticId.DynamicallyAccessedMembersMismatchTypeArgumentTargetsGenericParameter;
+				var diagArgs = GetDiagnosticArguments (sourceValue, targetValue, missingAnnotations);
+				context.ReportDiagnostic (Diagnostic.Create (DiagnosticDescriptors.GetDiagnosticDescriptor (diag), Location.None, diagArgs));
```
vitek-karas, line 90: Ideally this should go through DynamicallyAccessedMembersAnalyzer.GetDynamicallyAccessedMembersDiagnostics - or we need to refactor that out and call it from both places.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -58,6 +58,37 @@ public override void Initialize (AnalysisContext context)
...
+			}, SymbolKind.NamedType);
+		}
+
+		static void ProcessGenericTypes (SymbolAnalysisContext context)
+		{
+			// Check the Symbol to see the generics relationship, specifically around DAMT
+			// We will initially explore if an unbounded generic type has mismatched annotations with its base-class,
+			// which is an open generic type. i.e. Bar<T> : Foo<[DAMT] T>
+
+			INamedTypeSymbol derivedType = (INamedTypeSymbol) context.Symbol;
```
tlakollo, line 73: This should check for RUC first, in which case is not worth analyzing because we cannot produce the warning

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
LakshanF: Thanks!

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -58,6 +59,35 @@ public override void Initialize (AnalysisContext context)
...
+			//Check if generic parameter of derived and generic argument of the base have mismatched annotation
+
+			var typeParams = baseType!.TypeParameters;
+			var typeArgs = baseType.TypeArguments;
+
+			for (int i = 0; i < typeParams.Length; i++) {
+				var sourceValue = new SymbolValue (typeArgs[i]);
+				var targetValue = new SymbolValue (typeParams[i]);
+				// if a source is known, we dont need to warn
+				if (sourceValue.Source.Kind == SymbolKind.NamedType)
```
vitek-karas, line 86: We should still call the method below. Right now it won't do anything, but going forward it will "mark" the type. Which could produce additional diagnostics (if it has RUC methods on it for example).

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
LakshanF: Thanks!

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -59,6 +61,30 @@ public override void Initialize (AnalysisContext context)
...
+
+		private void ProcessSyntaxNode (SyntaxNodeAnalysisContext context)
+		{
+			if (context.SemanticModel.GetTypeInfo (context.Node).ConvertedType is not INamedTypeSymbol type)
+				return;
+
+			// The same INamedTypeSymbol can be called back resulting in duplicate warnings
+			// One way to filter is via the Node's Parent
+			if (context.Node.Parent is not ArgumentSyntax)
+				return;
```
vitek-karas, line 77: This feels dangerous - in that if there's a place where a named type can be used and it's not an argument syntax we would miss it - and potentially not report a warning. I'm definitely not an expert on this though.
@agocke - what would be the right way to avoid duplicate warnings while still not missing anything.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -59,6 +61,30 @@ public override void Initialize (AnalysisContext context)
 				var invocationOperation = (IInvocationOperation) context.Operation;
 				ProcessInvocationOperation (context, invocationOperation);
 			}, OperationKind.Invocation);
+			context.RegisterSyntaxNodeAction (context => {
+				ProcessSyntaxNode (context);
+			}, SyntaxKind.GenericName);
+		}
+
+		private void ProcessSyntaxNode (SyntaxNodeAnalysisContext context)
```
agocke, line 69: This is a bit vague, what exactly does this method process?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
agocke: I'm not sure I understand the duplication. Is it something like

```C#
var bad1 = new G<A>();
var bad2 = new G<A>();
```

where `G` is `G<[DAM]T>` and `A` doesn't have the requirement? In that case I would expect a warning in every place that the substitution occurs.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -59,6 +61,30 @@ public override void Initialize (AnalysisContext context)
...
+			// The same INamedTypeSymbol can be called back resulting in duplicate warnings
+			// One way to filter is via the Node's Parent
+			if (context.Node.Parent is not ArgumentSyntax)
+				return;
+
+			var typeParams = type!.TypeParameters;
+			var typeArgs = type.TypeArguments;
+
+			for (int i = 0; i < typeParams.Length; i++) {
+				var sourceValue = GetTypeValueNodeFromGenericArgument (typeArgs[i]);
```
agocke, line 83: What if the number of type arguments is less than the number of type parameters or vice versa?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -59,6 +61,30 @@ public override void Initialize (AnalysisContext context)
...
+		{
+			if (context.SemanticModel.GetTypeInfo (context.Node).ConvertedType is not INamedTypeSymbol type)
+				return;
+
+			// The same INamedTypeSymbol can be called back resulting in duplicate warnings
+			// One way to filter is via the Node's Parent
+			if (context.Node.Parent is not ArgumentSyntax)
+				return;
+
+			var typeParams = type!.TypeParameters;
```
agocke, line 79: Is `type` nullable here? Doesn't seem like it should be

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
LakshanF: This was due to the `nameof `identifier that is handled now

test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlow1.cs
```diff
@@ -0,0 +1,708 @@
+﻿using System;
```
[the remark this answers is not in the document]
LakshanF: Took a stab at naming

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
LakshanF: Fixed

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
LakshanF: Added an assert

test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlowWithoutMakeGeneric.cs
```diff
@@ -360,9 +345,8 @@ class TypeGenericRequirementsOnMembers<[DynamicallyAccessedMembers (DynamicallyA
 			public void PublicMethodsMethodParameter (TypeRequiresPublicMethods<TOuter> param) { }
 
 			public TypeRequiresPublicFields<TOuter> PublicFieldsMethodReturnValue () { return null; }
-			[ExpectedWarning ("IL2091", nameof (TypeRequiresPublicMethods<TOuter>))] // Return value
-			[ExpectedWarning ("IL2091", nameof (TypeRequiresPublicMethods<TOuter>))] // Compiler generated local variable
-			public TypeRequiresPublicMethods<TOuter> PublicMethodsMethodReturnValue () { return null; }
+			//[ExpectedWarning ("IL2091", nameof (TypeRequiresPublicMethods<TOuter>))]
+			//public TypeRequiresPublicMethods<TOuter> PublicMethodsMethodReturnValue () { return null; }
```
vitek-karas, line 349: We should either delete this or fix it (whichever makes sense).

LakshanF: Fixed

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -53,33 +55,60 @@ public override void Initialize (AnalysisContext context)
...
+				typeParams = type.TypeParameters;
+				typeArgs = type.TypeArguments;
+			}
 
-		static void ProcessTypeArguments (OperationAnalysisContext context, IInvocationOperation invocationOperation)
-		{
-			var targetMethod = invocationOperation.TargetMethod;
-			if (targetMethod.HasAttribute (RequiresUnreferencedCodeAnalyzer.RequiresUnreferencedCodeAttribute))
-				return;
+			if (context.SemanticModel.GetSymbolInfo (context.Node, context.CancellationToken).Symbol is IMethodSymbol targetMethod) {
```
vitek-karas, line 86: Should this be `else if` ?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -53,33 +55,60 @@ public override void Initialize (AnalysisContext context)
...
+				// Check if the generic parameters have any constraints
+				var node = context.Node;
+				while (node != null) {
+					if (node is MethodDeclarationSyntax parentMethod) {
+						if (parentMethod.ConstraintClauses.Count > 0) {
+							// This is safe, the linker would have marked the default .ctor already
+							return;
```
vitek-karas, line 93: I don't understand this:
* If there are constraints on the generic method, those should be effectively converted into annotations (or at least that's basically how linker treats this). For example `void Method<T>() where T : new {}` basically means that T should behave as if it has `DynamicallyAccessedMemberTypes.PublicParameterlessConstructors`. That said, Roslyn would have already checked this and failed if the instantiation doesn't match this, so we don't need to do anything here. On the other hand the T could have additional annotations of its own, in which case we still need to check the instantiation.
* Why is this done only for methods and not types?

LakshanF: - Made the constraints on type parameters more granular by looking only for new(), struct and unmanaged.
- I couldn't find a test scenario in the GenericParameterDF tests but will look closely when enabling more tests. Roslyn seems to be keeping the constraint information on the Method, which required walking up the tree to get and is only handling the generic methods for now

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -53,33 +56,65 @@ public override void Initialize (AnalysisContext context)
...
-		}
-
-		static void ProcessTypeArguments (OperationAnalysisContext context, IInvocationOperation invocationOperation)
-		{
-			var targetMethod = invocationOperation.TargetMethod;
-			if (targetMethod.HasAttribute (RequiresUnreferencedCodeAnalyzer.RequiresUnreferencedCodeAttribute))
-				return;
+			ImmutableArray<ITypeParameterSymbol> typeParams = default;
+			ImmutableArray<ITypeSymbol> typeArgs = default;
+			if (context.SemanticModel.GetTypeInfo (context.Node).ConvertedType is INamedTypeSymbol type) {
```
sbomer, line 72: Why is the `ConvertedType` piece necessary?

If there is for example an implicit conversion from `GenericType<Foo>` to `OtherType<Bar>`, I think we want to check `GenericType<Foo>` here (but I might be wrong about what `ConvertedType` is - @agocke does that sound right to you?). We should have a test where the `ConvertedType` is not null to make sure we do the right thing.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -53,33 +56,65 @@ public override void Initialize (AnalysisContext context)
...
+					if (node is InvocationExpressionSyntax invocationExpression && invocationExpression.Expression is IdentifierNameSyntax ident1) {
+						if (ident1.Identifier.ValueText.Equals ("nameof"))
+							return;
+					}
+					node = node.Parent;
+				}
+				typeParams = type.TypeParameters;
+				typeArgs = type.TypeArguments;
+			} else if (context.SemanticModel.GetSymbolInfo (context.Node, context.CancellationToken).Symbol is IMethodSymbol targetMethod) {
+				// Check if the generic parameters have any constraints on this method
```
sbomer, line 86: The type arguments of this generic method could be generic parameters of the containing method, or of the containing type - this doesn't handle the type case.

I think instead of walking over the declaring member's method/type syntax to find constraints, you should be able to get the `targetMethod.TypeArguments`. I think those can be either an `INamedTypeSymbol` or an `ITypeParameterSymbol` (or an array type, but maybe we can leave that for a separate change). `ITypeParameterSymbol` can be that of the method or the type, and it has info about the constraints.

test/ILLink.RoslynAnalyzer.Tests/DataFlowTests.cs
```diff
@@ -73,10 +73,16 @@ public Task FieldDataFlow ()
 			return RunTest (nameof (FieldDataFlow));
 		}
 
+		[Fact]
+		public Task GenericParameterDataFlowWithoutMakeGeneric ()
```
sbomer, line 77: nit: I would just call this `GenericParameterDataFlow`, and the other one something like `MakeGenericDataFlow`

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -53,33 +56,65 @@ public override void Initialize (AnalysisContext context)
...
-				var sourceValue = GetTypeValueNodeFromGenericArgument (targetMethod.TypeArguments[i]);
-				var targetValue = new GenericParameterValue (targetMethod.TypeParameters[i]);
-				foreach (var diagnostic in GetDynamicallyAccessedMembersDiagnostics (sourceValue, targetValue, invocationOperation.Syntax.GetLocation ()))
-					context.ReportDiagnostic (diagnostic);
+			if (typeParams != null) {
+				Debug.Assert (typeParams.Length == typeArgs.Length);
+				for (int i = 0; i < typeParams.Length; i++) {
+					var sourceValue = GetTypeValueNodeFromGenericArgument (typeArgs[i]);
+					var targetValue = new GenericParameterValue (typeParams[i]);
+					foreach (var diagnostic in GetDynamicallyAccessedMembersDiagnostics (sourceValue, targetValue, context.Node.GetLocation ()))
```
sbomer, line 115: I think maybe we should use the location of the type parameter instead of the whole generic type name - I think the more specific locations are a better UX.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -53,33 +56,65 @@ public override void Initialize (AnalysisContext context)
...
+			context.RegisterSyntaxNodeAction (context => {
+				ProcessGenericParameters (context);
+			}, SyntaxKind.GenericName);
 		}
 
-		static void ProcessInvocationOperation (OperationAnalysisContext context, IInvocationOperation invocationOperation)
+		static void ProcessGenericParameters (SyntaxNodeAnalysisContext context)
 		{
-			if (context.ContainingSymbol.HasAttribute (RequiresUnreferencedCodeAnalyzer.RequiresUnreferencedCodeAttribute))
+			if (context.ContainingSymbol is not null && context.ContainingSymbol.HasAttribute (RequiresUnreferencedCodeAnalyzer.RequiresUnreferencedCodeAttribute)) {
```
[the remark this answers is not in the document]
LakshanF: Good catch! Using `context.Node` instead of the `ContainingSymbol` to check for `IsInRequiresUnreferencedCodeAttributeScope` seems to work. Working on testing this fully and adding a test case.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
LakshanF: That was a copy and paste error, I don't need the implicit conversion, can do with just `Type` instead.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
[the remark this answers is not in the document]
LakshanF: Chatted offline with @sbomer and continued using `ContainingSymbol` but excluding `INamedTypeSymbol`

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
LakshanF: Moved constraints outside of this to [GenericParameterValue ](https://github.com/dotnet/linker/blob/8485bfcfbc55ab2b7007e3d39903e502fee82727/src/ILLink.RoslynAnalyzer/TrimAnalysis/GenericParameterValue.cs#L22)as @vitek-karas suggested

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
LakshanF: Moved constraints outside of this to [GenericParameterValue ](https://github.com/dotnet/linker/blob/8485bfcfbc55ab2b7007e3d39903e502fee82727/src/ILLink.RoslynAnalyzer/TrimAnalysis/GenericParameterValue.cs#L22)as @vitek-karas suggested

vitek-karas (approved):
LGTM - but please wait for Sven or Andy to take another look as well.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -53,33 +55,48 @@ public override void Initialize (AnalysisContext context)
...
+			ImmutableArray<ITypeSymbol> typeArgs = default;
+			if (context.SemanticModel.GetTypeInfo (context.Node).Type is INamedTypeSymbol type) {
+				// INamedTypeSymbol inside nameof, commonly used in [ExpectedWarning] can generate diagnostics
+				// Walking the node heirarchy to check if INamedTypeSymbol is inside a nameof
+				var parentNode = context.Node;
+				while (parentNode != null) {
+					if (parentNode is InvocationExpressionSyntax invocationExpression && invocationExpression.Expression is IdentifierNameSyntax ident1) {
+						if (ident1.Identifier.ValueText.Equals ("nameof"))
+							return;
+					}
```
tlakollo, line 81: I think we should do something like
```C#
if (parentNode is InvocationExpressionSyntax invocationExpression && invocationExpression.Expression is IdentifierNameSyntax ident1) {
    if (ident1.Identifier.ValueText.Equals ("nameof"))
        return;
    break;
}					
```
In order to stop on the first InvocationExpression, it either was 'nameof' or it wasnt and we need to continue. Current code has the risk of continue computing and find an unrelated 'nameof' that disables the analysis

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
sbomer: As discussed this doesn't look straightforward so I am ok with reporting the full generic name as the location for now. I filed this issue to track the suggestion: https://github.com/dotnet/linker/issues/2569

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -53,33 +55,48 @@ public override void Initialize (AnalysisContext context)
...
+			context.RegisterSyntaxNodeAction (context => {
+				ProcessGenericParameters (context);
+			}, SyntaxKind.GenericName);
 		}
 
-		static void ProcessInvocationOperation (OperationAnalysisContext context, IInvocationOperation invocationOperation)
+		static void ProcessGenericParameters (SyntaxNodeAnalysisContext context)
 		{
-			if (context.ContainingSymbol.IsInRequiresUnreferencedCodeAttributeScope ())
+			// We dont need to check for RUC for INamedTypeSymbol here
```
sbomer, line 65: I think this warrants a more detailed comment - maybe something along these lines:

RUC on the containing symbol normally silences warnings, but when examining a generic base type, the containing symbol is the declared derived type. RUC on the derived type does not silence warnings about base type arguments.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
sbomer: Personally I would like to produce warnings even for generic types inside `nameof` - even though this might be noisier than the linker. I think it makes sense to think of the generic annotations as an extension of the type system, and to warn for any offending reference an instantiated type. It's a little painful for the linker repo but I don't know if it will be common in the wild.  I think this is similar to local variable declarations where (IIRC) the linker would not warn but the analyzer would. @vitek-karas do you have an opinion on this?

In any case, I think either approach is fine for now since it's easy enough to change. For a non-attribute testcase I think you could do something like this:

```csharp
class Outer<TOuter> {
  static void Method() {
    string typeName = nameof (TypeWithAnnotatedGenericParameter<TOuter>);
  }
}
```

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -53,33 +55,48 @@ public override void Initialize (AnalysisContext context)
...
+				&& context.ContainingSymbol is not INamedTypeSymbol
+				&& context.ContainingSymbol.IsInRequiresUnreferencedCodeAttributeScope ())
 				return;
 
-			ProcessTypeArguments (context, invocationOperation);
-		}
+			ImmutableArray<ITypeParameterSymbol> typeParams = default;
+			ImmutableArray<ITypeSymbol> typeArgs = default;
+			if (context.SemanticModel.GetTypeInfo (context.Node).Type is INamedTypeSymbol type) {
+				// INamedTypeSymbol inside nameof, commonly used in [ExpectedWarning] can generate diagnostics
```
sbomer, line 74: nit: I would leave a comment about the expected behavior for `nameof` without mentioning `ExpectedWarning` - we should design the behavior for the customer code, and `ExpectedWarning` is just a detail of our codebase.

sbomer (approved):

vitek-karas:
@sbomer 
> Personally I would like to produce warnings even for generic types inside nameof - even though this might be noisier than the linker. I think it makes sense to think of the generic annotations as an extension of the type system, and to warn for any offending reference an instantiated type. It's a little painful for the linker repo but I don't know if it will be common in the wild. I think this is similar to local variable declarations where (IIRC) the linker would not warn but the analyzer would. @vitek-karas do you have an opinion on this?

The testing problem can be solved - we could modify our validator to ignore warnings from within test attributes (assuming there's a reasonable way to detect them).
So in that sense I don't mind. 

I just checked and Roslyn does raise warnings for generic constraints even inside nameof:
```
Console.WriteLine(nameof(WithNewConstraint<WithoutNew>)); // CS0310 ....
```

So it would be good to follow suite I guess.

In general I still think that analyzer should not report warnings which the linker doesn't - mainly because if there's a warning which we think the analyzer SHOULD report and the linker doesn't -> it's almost certainly a bug in the linker.
This is a specific case though, since it's at the level the linker doesn't have visibility into - and it might help developers. So in this case I think it would be good, but only in this case :wink:.

agocke:
I'd be fine generating a warning here, just because it seems like it might be simpler to implement and understand.

LakshanF:
I agree this is a grey area but fall on the side of (slightly) preferring not to warn. My reasoning is that this warning will cause developer confusion (its not intuitive to anyone that a compiler enhancement used to generate a string will have trimming implications). After the developer understands the warning, I believe it will almost always result in no action (as in use the literal string for example) circumventing the good programming practice `nameof `was meant to add.

But I'm ok with the majority view and generating the warning.

agocke:
To be clear, I'm fine either way 😄

PR #2479, opened 2022-01-05
ExpectedWarning silences only one warning per attribute

tlakollo:
Previously the ExpectedWarning attribute in linker will silence multiple warnings if they matched the requirements. This allowed duplicated warnings to be generated without noticing. Analyzer silences only one warning per attribute showing some bugs in linker. This PR changes the ExpectedWarning behavior in linker to only silence one warning per attribute.

### DIFF
--- a/test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
+++ b/test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
@@ -725,30 +725,38 @@ namespace Mono.Linker.Tests.TestCasesRunner
 
 							int expectedWarningCodeNumber = int.Parse (expectedWarningCode.Substring (2));
 							string expectedOrigin = null;
+							bool expectedWarningFound = false;
 
-							var matchedMessages = loggedMessages.Where (mc => {
-								if (mc.Category != MessageCategory.Warning || mc.Code != expectedWarningCodeNumber)
-									return false;
+							foreach (var loggedMessage in loggedMessages) {
 
-								foreach (var expectedMessage in expectedMessageContains)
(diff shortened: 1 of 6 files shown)


vitek-karas (approved):

test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -725,30 +725,38 @@ void VerifyLoggedMessages (AssemblyDefinition original, LinkerTestLogger logger,
 
 							int expectedWarningCodeNumber = int.Parse (expectedWarningCode.Substring (2));
 							string expectedOrigin = null;
+							bool expectedWarningFound = false;
 
-							var matchedMessages = loggedMessages.Where (mc => {
-								if (mc.Category != MessageCategory.Warning || mc.Code != expectedWarningCodeNumber)
-									return false;
+							foreach (var loggedMessage in loggedMessages) {
```
sbomer, line 730: Personally I'd prefer to keep matchedMessages around, since it makes it slightly easier to debug when an attribute matches multiple warnings. Below we could either assert that only one message is matched, or remove just the first match. This also avoids having to repeat `expectedWarningFound = true; loggedMessages.Remove (loggedMessage); break;` in the code below.

tlakollo: I guess what I didn't like about matchedMessages is that we loop over all messages (except the ones are being removed), for all attributes. If we already found a match and we know that we can remove only one logged message, why not stop there. But yeah it was convenient that matched messages allow you to run the whole logic and you could review if it matched something or not.

PR #2481, opened 2022-01-06
[DAM analyzer] Support exceptional control flow

sbomer:
This includes a change to make the analyzer tests fail when an exception is thrown.

This models exceptional control flow by maintaining an exception state for try/catch regions, which is updated to include all possible states encountered within that region. This exceptional state is propagated from try regions to the beginning of catch regions, and from try or catch regions to the beginning of finally regions.

Finally regions are represented as additional information for each predecessor edge in the control flow graph, following the data model given by Roslyn's CFG. Each edge includes a list of finally regions through which control flows as part of the edge before reaching its destination. For now, this approach inefficiently propagates the old finally state along these chains of finally regions when it encounters them. This has the downside that long finally regions will cause re-analysis until each finally block sees a new state.

The finally state is also unified along all such paths, so the warning state at the exit of a finally is a conservative approximation that accounts for all normal control flow leading to the finally region.

Additionally, the finally regions are analyzed separately for the case when an exception causes a finally block to be reached. Inside the finally region, this will produce a superset of the warnings for normal control flow inside the same region. However this exceptional finally state does not flow out of the finally, to avoid spurious warnings in code that is unreachable when executing finally blocks due to exceptions.

This doesn't yet include any handling for filter regions. I'll work on that next, possibly in a follow-up change.


vitek-karas (approved):
Looks good!

PR #2491, opened 2022-01-08
[DAM analyzer] Always recurse into child operations in visitor

sbomer:
Thanks @vitek-karas for pointing this out!

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
@@ -17,9 +17,10 @@ namespace ILLink.RoslynAnalyzer.DataFlow
 	// - field
 	// - parameter
 	// - method return
-	public abstract class LocalDataFlowVisitor<TValue, TValueLattice> : OperationVisitor<LocalState<TValue>, TValue>,
+	public abstract class LocalDataFlowVisitor<TValue, TValueLattice> : OperationWalker<LocalState<TValue>, TValue>,
 		ITransfer<BlockProxy, LocalState<TValue>, LocalStateLattice<TValue, TValueLattice>>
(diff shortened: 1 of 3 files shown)


src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -90,17 +92,6 @@ public void Transfer (BlockProxy block, LocalState<TValue> state)
 		// may (must?) come from BranchValue of an operation whose FallThroughSuccessor is the exit block.
 		public abstract void HandleReturnValue (TValue returnValue, IOperation operation);
 
-		// Override with a non-nullable return value to prevent some warnings.
-		// The interface constraint on TValue ensures that it's not a nullable type, so this is safe as long
-		// as no overrides return default(TValue).
-		public override TValue Visit (IOperation? operation, LocalState<TValue> argument) => operation != null ? operation.Accept (this, argument)! : TopValue;
```
sbomer, line 96: Ideally I would override this to return TopValue if the base class returned default, but this would make the assumption that default(TValue) isn't a valid return value. Instead I am making the assumption that default(TValue) is the top value.

vitek-karas (approved):

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
vitek-karas: I ran into something similar as well - would it make sense to change the design a little and basically require the Top to be default(TValue)? It would mean we don't need to pass around the lattice in some cases :-)

sbomer: Yes, at this point I think so. I actually had a version that did that at some point, but I didn't want to constrain the top value that way.

PR #2498, opened 2022-01-10
First intrinsics

vitek-karas:
Refactor trim operation visitor to have a single HandleMethodCall

Implement first simple intrinsics handling - only if it has no diagnostics and no "marking".
Added a targetted test for GetTypeInfo so that I can easily test the new shared intrinsic.


vitek-karas:
/cc @sbomer, @MichalStrehovsky, @agocke , @tlakollo 

First attempt and sharing at least one intrinsic. I'm mostly looking for early feedback on the approach to things like:
- HandleMethodCall
- MethodProxy and its implementations
- Passing around arguments for method calls

src/ILLink.Shared/TypeSystemProxy/MethodProxy.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace ILLink.Shared.TypeSystemProxy
+{
+	internal readonly partial record struct MethodProxy : IMemberProxy
```
MichalStrehovsky, line 6: Is the `record` buying us anything here?

src/ILLink.Shared/TypeSystemProxy/IMemberProxy.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace ILLink.Shared.TypeSystemProxy
+{
+	interface IMemberProxy
```
agocke, line 6: You could also use a struct here with implicit conversions from methods

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -197,19 +200,20 @@ public static IMethodSymbol GetPropertyMethod (IPropertyReferenceOperation opera
 
 		public override TValue VisitPropertyReference (IPropertyReferenceOperation operation, LocalState<TValue> state)
 		{
-			if (operation.Instance != null) {
-				var instanceValue = Visit (operation.Instance, state);
-				HandleReceiverArgument (instanceValue, GetPropertyMethod (operation), operation);
+			var propertyMethod = GetPropertyMethod (operation);
```
agocke, line 203: I know you're just copying this over but fyi there's an IOpertionExtensions.GetValueUsageInfo that provides all this information and much more

agocke:
Generally lgtm. We could probably make the intinsic parsing code a little faster/cleaner by writing a data structure for signatures and a comparison helper

src/ILLink.Shared/TypeSystemProxy/MethodProxy.cs
vitek-karas: Nothing - will remove

src/ILLink.RoslynAnalyzer/TrimAnalysis/MethodProxy.cs
```diff
@@ -0,0 +1,70 @@
...
+		internal partial bool HasGenericParameters ()
+			=> Method.IsGenericMethod;
+
+		internal partial bool HasGenericParametersCount (int genericParameterCount)
+			=> Method.TypeParameters.Length == genericParameterCount;
+
+		internal partial bool IsStatic ()
+			=> Method.IsStatic;
+
+		bool IsTypeOf (ITypeSymbol type, string namespaceName, string typeName)
```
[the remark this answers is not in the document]
vitek-karas: It's not - I simply didn't know about the `HasName` - I searched, but could not find anything.

src/linker/Linker.Dataflow/MethodProxy.cs
```diff
@@ -0,0 +1,40 @@
...
+			=> Method.HasParameters;
+
+		internal partial bool HasParametersCount (int parameterCount)
+			=> Method.Parameters.Count == parameterCount;
+
+		internal partial bool HasParameterOfType (int parameterIndex, string namespaceName, string typeName)
+			=> Method.HasParameterOfType (parameterIndex, namespaceName, typeName);
+
+		internal partial bool HasGenericParameters ()
+			=> Method.HasGenericParameters;
```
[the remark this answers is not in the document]
vitek-karas: It is faster in Cecil (in some cases Cecil you can avoid allocating empty collection instance - it internally stores `null` until you access the collection - so asking if it's null is more efficient).

src/ILLink.Shared/TypeSystemProxy/IMemberProxy.cs
vitek-karas: Yes - I'll leave it as is for now - and see how it goes. Currently it's actually not needed for anything, but it will be soon.

src/ILLink.RoslynAnalyzer/TrimAnalysis/MethodProxy.cs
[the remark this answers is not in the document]
vitek-karas: I'm going to change this - unfortunately there's a mismatch between Roslyn/Linker here (one takes the full type name, the other namespace and type name separately), but that's a minor thing.

vitek-karas:
> Generally lgtm. We could probably make the intinsic parsing code a little faster/cleaner by writing a data structure for signatures and a comparison helper

Agreed - but I won't try to tackle this in my change yet. There's a bigger issue that we don't detect the intrinsics in a precise way - https://github.com/dotnet/linker/issues/1623. What you're suggesting is basically a fix for that issue.

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -111,12 +112,12 @@ public override TValue VisitSimpleAssignment (ISimpleAssignmentOperation operati
 				// Doesn't get called for assignments to locals, which are handled above.
 				HandleAssignment (value, targetValue, operation);
 				break;
-			case IPropertyReferenceOperation propertyRef:
-				// A property assignment is really a call to the property setter.
-				var setMethod = propertyRef.Property.SetMethod;
-				Debug.Assert (setMethod != null);
-				HandlePropertySetterArgument (value, setMethod!, operation);
-				break;
+			case IPropertyReferenceOperation propertyRef: {
```
tlakollo, line 115: Nit: I don't think these braces are providing any value

vitek-karas: True - I'll remove it.

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
vitek-karas: I changed it, could you please review this bit if it makes sense?

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -174,19 +176,19 @@ public static IMethodSymbol GetPropertyMethod (IPropertyReferenceOperation opera
 
 		public override TValue VisitPropertyReference (IPropertyReferenceOperation operation, LocalDataFlowState<TValue, TValueLattice> state)
 		{
-			if (operation.Instance != null) {
-				var instanceValue = Visit (operation.Instance, state);
-				HandleReceiverArgument (instanceValue, GetPropertyMethod (operation), operation);
+			if (operation.GetValueUsageInfo (operation.Property).HasFlag (ValueUsageInfo.Read)) {
```
agocke, line 179: This looks right to me. The tricky case is something like

```C#
class C
{
    public int P { get; set; } = 0;
}
var c = new C();
c.P += 1; // both a read and a write in the same operation
```

vitek-karas: So Assignment operation is not the only place where values are assigned to property/field/variable?

@sbomer - I guess we have a lot more work to make the data flow work reliably. I guess our saving grace is that we're currently only tracking Type/string values for which operations like `+=` don't really make sense.

sbomer: One thing we probably need to do is go down the list of IOperations looking for any cases that aren't handled by `VisitSimpleAssignment` and `VisitFlowCapture`. I haven't encountered any missing cases in the tests we've enabled so far but there are almost certainly some.

sbomer: Filed https://github.com/dotnet/linker/issues/2507 to track this

agocke:
Roslyn calls this concept "WellKnownTypes" and "WellKnownMembers"

vitek-karas:
Thanks @agocke. I was actually thinking about another direction, but this works for the intrinsics themselves. My point was more about the fact that we want to share code which is not just about intrinsics - for example we want to share code which will handle the "annotation value" vs "annotated target" validation/marking. That has basically nothing to do with intrinsics...

sbomer (approved):
Looks good!

PR #2499, opened 2022-01-11
Add WarningIds part 2

tlakollo:
Add DiagnosticIds IL2050 to IL2116 in DiagnosticId and SharedStrings files

Notice that the following ids are missing from the list since they are no longer produced https://github.com/dotnet/linker/issues/2474


vitek-karas:
I would be careful just removing any note of the old warning codes. Ideally we should not reuse warning codes, even if they're not produced anymore. As such, I would add them to the enum and mark them as unused - maybe `_unused_WarningCodeName` or something like that. To make it explicit that these should not be used again for anything.

vitek-karas (approved):
Looks good - the merge with damAnalyzer will be a pain, but there's no good way to avoid that.

PR #2500, opened 2022-01-13
Create a schema for the LinkAttribute XML files

jtschuster:
Adds xml schema for `ILLink.LinkAttributes.xml`, and a more general schema that encompasses all of `ILLink.LinkAttributes.xml`, `ILLink.Descriptors.xml`, and `ILLink.Substittutions.xml`. In test xml files there are relative paths to the relevant schema to enable linting.

Also fixes where the `preserve` attribute was sometimes "none" and sometimes "nothing".


test/Mono.Linker.Tests.Cases/Attributes.Debugger/DebuggerAttributesRemoved.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8" ?>
-<linker>
+<linker xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://www.w3schools.com ../../../src/ILLink.Shared/ILLink.LinkAttributes.xsd">
```
tlakollo, line 2: nit: I'm not experienced on XML schemas, but having a link to w3schools seems like something we don't want

src/ILLink.Shared/ILLink.LinkAttributes.xsd
```diff
@@ -0,0 +1,161 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<xsd:schema attributeFormDefault="unqualified"
+		elementFormDefault="unqualified"
+		xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+
+	<xsd:complexType name="typechild">
```
vitek-karas, line 6: NIt: `typechild` doesn't sound right as a name - I would probably call it `membertype` or something like that. "member" in the type system means basically everything (type, method, field, proprety, ...). "child" has no meaning in type system.

src/ILLink.Shared/ILLink.LinkAttributes.xsd
```diff
@@ -0,0 +1,161 @@
...
+							<xsd:element ref="method" />
+							<xsd:element ref="field" />
+							<xsd:element ref="property" />
+							<xsd:element ref="attribute"></xsd:element>
+						</xsd:choice>
+						<xsd:attribute name="name" type="xsd:string" />
+						<xsd:attribute name="feature" />
+						<xsd:attribute name="featurevalue" type="xsd:boolean" />
+						<xsd:attribute name="featuredefault" type="xsd:boolean" />
+					</xsd:complexType>
```
vitek-karas, line 142: I think this should be recursive - technically it's allowed to have nested types within nested types.

src/ILLink.Shared/ILLink.xsd
```diff
@@ -0,0 +1,193 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
vitek-karas, line 1: I would rename the file to `ILLink.Descriptor.xsd`

vitek-karas (approved):
Looks good - just couple of minor things

src/ILLink.Shared/ILLink.xsd
jtschuster: I made this schema broad enough to cover `ILLink.Substitutions.xml` as well as `ILLink.Descriptor.xml` and `ILLink.LinkAttributes.xml`, so I don't want to name it just for descriptors, but if we don't like `ILLink.xsd`, maybe something like `ILLink.General.xsd`?

src/ILLink.Tasks/CreateRuntimeRootDescriptorFile.cs
```diff
@@ -356,7 +356,7 @@ static void AddXmlTypeNode (XmlDocument doc, XmlNode assemblyNode, string typeNa
 					typeNode.Attributes.Append (preserve);
 				} else if ((members.fields == null) && (members.methods == null)) {
 					XmlAttribute preserve = doc.CreateAttribute ("preserve");
-					preserve.Value = "nothing";
+					preserve.Value = "none";
```
tlakollo, line 359: It's correct to have only one between the two, but just be aware that seems like runtime is using the 'nothing' notation since is what it was documented. See [file](https://github.com/dotnet/runtime/blob/main/src/mono/System.Private.CoreLib/src/ILLink/ILLink.Descriptors.xml#L59).
The way I see it we either agreed on nothing or have to change other files outside the linker repo who may have used 'nothing' already

tlakollo: Seems like the code is looking for Nothing and if other value is passed then it defaults to Nothing
https://github.com/dotnet/linker/blob/main/src/linker/Linker.Steps/DescriptorMarker.cs#L137-L146

src/ILLink.Shared/ILLink.xsd
```diff
@@ -0,0 +1,193 @@
...
+			<xsd:attribute name="feature" />
+			<xsd:attribute name="featurevalue" type="xsd:boolean" />
+			<xsd:attribute name="featuredefault" type="xsd:boolean" />
+		</xsd:complexType>
+	</xsd:element>
+	
+	<xsd:element name="resource">
+		<xsd:complexType>
+			<xsd:attribute name="name" use="required" />
+			<xsd:attribute name="action" use="required" />
```
MichalStrehovsky, line 82: Maybe add a type that restricts this to `remove`?

src/ILLink.Shared/ILLink.xsd
```diff
@@ -0,0 +1,193 @@
...
+			<xsd:choice minOccurs="0" maxOccurs="unbounded">
+				<xsd:element ref="event" />
+				<xsd:element ref="method" />
+				<xsd:element ref="attribute" />
+				<xsd:element ref="field" />
+				<xsd:element ref="property" />
+				<xsd:element ref="type" />
+			</xsd:choice>
+			<xsd:attribute name="fullname" type="xsd:string" />
+			<xsd:attribute name="name" type="xsd:string" />
```
MichalStrehovsky, line 176: Do we support `name` on `type`?

src/ILLink.Shared/ILLink.xsd
```diff
@@ -0,0 +1,193 @@
...
+			<xsd:attribute name="required" type="xsd:boolean" />
+			<xsd:attribute name="preserve" type="preservevalues" />
+			<xsd:attribute name="feature" type="xsd:string" />
+			<xsd:attribute name="featurevalue" type="xsd:boolean" />
```
MichalStrehovsky, line 180: Add `featuredefault`?

src/ILLink.Shared/ILLink.xsd
jtschuster: This one I started as a form of documentation before I realized how different each xml file is. I think that's a good point, this schema is a little too general to provide any value. I'll remove this file and only do the more specific schema.

src/ILLink.Shared/ILLink.xsd
jtschuster: The documentation has `name` supported on nested `type`s, but not on top level.

vitek-karas: What does the code do?

jtschuster: Looks like it also allows name on nested types but not on top level.

jtschuster: `protected virtual void ProcessTypes (AssemblyDefinition assembly, XPathNavigator nav, bool warnOnUnresolvedTypes)` in ProcessLinkerXmlBase.cs gets the `fullname` attribute, but in LinkAttributesParser.cs it will look for name, and doesn't look for `fullname`

PR #2501, opened 2022-01-13
Ensure trim analysis patterns are additive

sbomer:
During the analysis, we may pass over an operation multiple times with different dataflow states. As we go, we cache the dataflow values that are later used to produce warnings. This change makes the logic more robust by ensuring we never lose cached values.

As part of this I discovered (via some test failures) that caching these by operation isn't enough - for example, in `return Foo();` the `Foo()` may produce warnings both for a mismatch on the implicit receiver (`this`), and for the return value of `Foo()`. To fix this, I think it's enough to cache them by operation and whether it is a returned value.


vitek-karas (approved):
I think for now this is good - but we will need to replace this by storing the entire "method call". But that's to come later.

PR #2516, opened 2022-01-18
Enable multiple arguments for RemoveAttributeInstancesAttribute ctor

jtschuster:
Generates a constructor for each number of ctor arguments found in the
LinkAttributes.xml file, and updates some of the logic around that.

Previously we would create a single `RemoveAttributeInstancesAttribute` with one ctor of type Object. This meant the xml could only have a single argument for that attribute, and thus made it harder to be specific with which attributes to remove. This generates ctors with an arbitrary number of arguments depending on the `ILLink.LinkAttributes.xml`. 

At the moment, the arguments to the `RemoveAttributeInstancesAttribute` in xml must be "boxed" in an Object type because that is how we got around having a single constructor for many argument types previously. Now that we generate arbitrary ctors for the attribute, it should be possible to allow the arguments in the xml to be unboxed without the Object wrapper to simplify things, and can probably be done without breaking "boxed" implementations of the xml. Curious to hear thoughts on that and if it's worth the effort and potential for breaking.

Ex. Boxed argument to remove `[TestConditionalRemove(100, ...)]`
```xml
<type fullname="Mono.Linker.Tests.Cases.LinkAttributes.TestConditionalRemoveAttribute">
  <attribute internal="RemoveAttributeInstances">
    <argument type="System.Object">
      <argument type="System.Int32">100</argument>
    </argument>
  </attribute>
</type>
```
vs. Unboxed
```xml
<type fullname="Mono.Linker.Tests.Cases.LinkAttributes.TestConditionalRemoveAttribute">
  <attribute internal="RemoveAttributeInstances">
    <argument type="System.Int32">100</argument>
  </attribute>
</type>
```


src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -91,10 +92,17 @@ static string FormatCustomAttribute (CustomAttribute ca)
...
 		{
-			if (_context.MarkedKnownMembers.RemoveAttributeInstancesAttributeDefinition != null)
-				return _context.MarkedKnownMembers.RemoveAttributeInstancesAttributeDefinition;
+			TypeDefinition? td = null;
+
+			if (_context.MarkedKnownMembers.RemoveAttributeInstancesAttributeDefinition is TypeDefinition knownTypeDef) {
+				if (knownTypeDef.GetConstructorsOnType (_ => true).Any (m => m.Parameters.Count == ctorArgumentCount)) {
+					return knownTypeDef;
+				} else {
+					td = knownTypeDef;
```
tlakollo, line 103: Here the type was recognized but a constructor with the number of parameters wasn't found, I think this is because you can delete attributes that match things like the first argument even if the constructor accepts 3 arguments. For example the test
```XML
<type fullname="Mono.Linker.Tests.Cases.LinkAttributes.TestConditionalRemoveAttribute">
	<attribute internal="RemoveAttributeInstances">
		<argument type="System.Object">
			<argument>remove1</argument>
		</argument>
	</attribute>
</type>
```
```C#
[TestConditionalRemove ("remove1", '1', 3)]  //This gets removed because the first arg matches
static void Removed_2 ()
{
}
```
The problem here is that I think there is nothing preventing from creating a malformed version of the type constructor, something that doesn't exist, for example:
```XML
<type fullname="Mono.Linker.Tests.Cases.LinkAttributes.TestConditionalRemoveAttribute">
	<attribute internal="RemoveAttributeInstances">
		<argument type="System.Object">
			<argument>arg1</argument>
		</argument>
		<argument type="System.Object">
			<argument>arg2</argument>
		</argument>
		<argument type="System.Object">
			<argument>arg3</argument>
		</argument>
		<argument type="System.Object">
			<argument>arg4</argument>
		</argument>
		<argument type="System.Object">
			<argument>arg5</argument>
		</argument>
	</attribute>
</type>

File: src/linker/Linker/LinkerAttributesInformation.cs [hunk 2]
```diff
@@ -124,18 +124,9 @@ public static LinkerAttributesInformation Create (LinkContext context, ICustomAt
 			return null;
 		}
 
-		static RemoveAttributeInstancesAttribute? BuildRemoveAttributeInstancesAttribute (LinkContext context, TypeDefinition attributeContext, CustomAttribute ca)
+		static RemoveAttributeInstancesAttribute? BuildRemoveAttributeInstancesAttribute (LinkContext _, TypeDefinition __, CustomAttribute ca)
```
tlakollo, line 127: Instead of calling the method you could just do `new RemoveAttributeInstancesAttribute (ca.ConstructorArguments);` and delete the method, also flag the diagnostic as _unused_ in the [DiagnosticId.cs](https://github.com/dotnet/linker/blob/main/src/ILLink.Shared/DiagnosticId.cs) file.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -36,6 +36,7 @@ public override void Initialize (AnalysisContext context)
 			context.EnableConcurrentExecution ();
 			context.ConfigureGeneratedCodeAnalysis (GeneratedCodeAnalysisFlags.ReportDiagnostics);
 			context.RegisterCompilationStartAction (context => {
+
```
tlakollo, line 39: nit: extra line

src/linker/Linker.Steps/LinkAttributesParser.cs
jtschuster: Here, `knownTypeDef` is the type definition of `RemoveAttributeInstancesAttribute`, not the `TestConditionalRemoveAttribute` type. From my understanding, later in the linker it checks the arguments passed to the constructor of each instance of `TestConditionalRemoveAttribute` (the ones placed on the methods), against the arguments to the constructor of `RemoveAttributeInstancesAttribute` (which is placed on the `TestConditionalRemoveAttribute` definition by the xml). So we are generating a new ctor for that `RemoveAttributeInstancesAttribute` type to enable constructors of that to enable the xml to specify multiple arguments to match, not a ctor for `TestConditionalRemoveAttribute`.

So yes, it enables the xml with 5 arguments, but it's not generating a new ctor for the `TestConditionalRemoveAttribute` that shouldn't be there, and if the xml has more arguments than any of the `TestConditionalRemoveAttribute` ctors, then none are removed.

src/linker/Linker/LinkerAttributesInformation.cs
tlakollo: My bad there was more than one instance of the diagnostic creation. Sorry

docs/data-formats.md
```diff
@@ -544,31 +579,38 @@ instead of specifying `fullname` in the attribute as described in the following
...
         </argument>
       </attribute>
     </type>
   </assembly>
 </linker>
 ```
 
+This can also be done to match multiple arguments. In the example below, all instances with
+`RequiresUnreferencedCodeAttribute` where the first argument is of type `System.String` and the 
+
```
vitek-karas, line 606: This doesn't seem to belong here.

src/linker/Linker/RemoveAttributeInstancesAttribute.cs
```diff
@@ -14,14 +16,14 @@ namespace Mono.Linker
...
-		}
-
-		public RemoveAttributeInstancesAttribute (CustomAttributeArgument value1)
-		{
-			Arguments = new[] { value1 };
+			// Arguments were previously boxed in Object, but don't need to be
+			Arguments = values.Select ((arg) => {
+				Debug.Assert (arg.Value is CustomAttributeArgument);
+				if (arg.Value is CustomAttributeArgument caa) return caa;
+				return arg;
```
vitek-karas, line 25: I don't understand this - the assert says that `arg.Value` would always be `CustomAttributeArgument` - but then we check for it again?
Is this to prevent runtime crashes if something goes wrong?

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -91,10 +92,17 @@ static string FormatCustomAttribute (CustomAttribute ca)
...
 
-		TypeDefinition? GenerateRemoveAttributeInstancesAttribute ()
+		TypeDefinition? GenerateRemoveAttributeInstancesAttribute (int ctorArgumentCount)
 		{
-			if (_context.MarkedKnownMembers.RemoveAttributeInstancesAttributeDefinition != null)
-				return _context.MarkedKnownMembers.RemoveAttributeInstancesAttributeDefinition;
+			TypeDefinition? td = null;
+
+			if (_context.MarkedKnownMembers.RemoveAttributeInstancesAttributeDefinition is TypeDefinition knownTypeDef) {
+				if (knownTypeDef.GetConstructorsOnType (_ => true).Any (m => m.Parameters.Count == ctorArgumentCount)) {
```
vitek-karas, line 100: Why not apply the filter in the GetConstructorsOnType ?

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -117,16 +125,22 @@ static string FormatCustomAttribute (CustomAttribute ca)
 			//		public RemoveAttributeInstancesAttribute (object value1) {}
 			// }
```
vitek-karas, line 126: I think this is worth updating the comment and adding details to it. Specifically:
* We generate new constructors for the syntetic `RemoveAttributeInstancesAttribute` type (in the Cecil space, with no namespace) based on the number of arguments in the XML
* Each such .ctor is defined as `.ctor(object, object, ....)` - all arguments are boxed then in the XML by wrapping them with "object" argument.
* The internal linker representation which is a type called `Mono.Linker.RemoveAttributeInstancesAttribute` has only one `.ctor` which takes a collection of arguments. The linker attribute caching mechanism (link) converts the call to `.ctor` with arbitrary number of arguments to the instantiation of the internal representation.
* The boxing is removed in this conversion

src/linker/Linker.Steps/LinkAttributesParser.cs
jtschuster: That's a good point, that would be simpler and probably allocate less.

src/linker/Linker/RemoveAttributeInstancesAttribute.cs
jtschuster: Yeah, unless we do decide to support "unboxed" arguments to RemoveAttributeInstances, this doesn't make sense here.

src/linker/Linker.Steps/LinkAttributesParser.cs
tlakollo: For me, it's generating a new fake constructor with 5 parameters and adding it to _context.MarkedKnownMembers.RemoveAttributeInstancesAttributeDefinition.
There is no validation afterward, so it gets added and no warning is produced. Before it would warn saying that having more than 1 argument in the XML is not allowed, which is also not a great experience.
I do understand that TestConditionalRemoveAttribute won't have the constructor and therefore you cannot create an attribute with 5 constructors without getting a compiler error, but I think we should produce a warning for constructors that don't exist. Otherwise, we could have people thinking that what they wrote in the XML was correct and is actually doing nothing, or even worse they could write something that at some point is invalid but later on is deleting something that might not be needed to be removed.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -853,19 +853,27 @@ bool IsAttributeRemoved (CustomAttribute ca, TypeDefinition attributeType)
...
+					// RemoveAttributeInstances constructor should unbox each argument out of the Object box
+					Debug.Assert (raiVal is not CustomAttribute);
+
+					// Arguments to a custom argument may be boxed in an Object
+					if (aiVal is CustomAttributeArgument caaB) {
+						aiVal = caaB.Value;
+						aiType = caaB.Type;
+					}
+
+					if (!raiVal.Equals (aiVal) || !(raiType.FullName == aiType.FullName)) return false;
```
marek-safar, line 876: What is `raiType.FullName == aiType.FullName` buying us here?

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -117,16 +136,22 @@ static string FormatCustomAttribute (CustomAttribute ca)
...
+
+				var ctor0 = new MethodDefinition (".ctor", ctorAttributes, voidType);
+				td.Methods.Add (ctor0);
+			}
+
+			var ctorN = new MethodDefinition (".ctor", ctorAttributes, voidType);
+			for (int i = 0; i < ctorArgumentCount; i++) {
+				ctorN.Parameters.Add (new ParameterDefinition (objectType));
+			}
+			td.Methods.Add (ctorN);
```
marek-safar, line 154: This is going to add ctor with 0 parameters twice

docs/data-formats.md
```diff
@@ -2,7 +2,8 @@
 
 ## Input Data Formats
 
-ILLink uses several data formats to control or influence the trimming process. The data formats are not versioned but are backward compatible.
```
marek-safar, line 5: Please make any formating changes in separate PR

test/Mono.Linker.Tests.Cases/LinkAttributes/LinkerAttributeRemovalConditional.cs
```diff
@@ -61,6 +63,17 @@ static void Removed_2 ()
 		static void Removed_3 ()
 		{
 		}
+		[Kept]
+		[TestConditionalRemove ("99", "98")]
+		static void Removed_4 () // Removed because both 99 and 98 match
```
marek-safar, line 68: The comments seems misplaced

src/linker/Linker/RemoveAttributeInstancesAttribute.cs
```diff
@@ -14,14 +16,14 @@ namespace Mono.Linker
 		AttributeTargets.Class, Inherited = false)]
 	public sealed class RemoveAttributeInstancesAttribute : Attribute
 	{
-		public RemoveAttributeInstancesAttribute ()
+		public RemoveAttributeInstancesAttribute (System.Collections.Generic.ICollection<CustomAttributeArgument> values)
```
marek-safar, line 19: Why ICollection instead of `Collection<CustomAttributeArgument>` ?

src/linker/Linker/RemoveAttributeInstancesAttribute.cs
```diff
@@ -14,14 +16,14 @@ namespace Mono.Linker
...
 		{
-			Arguments = Array.Empty<CustomAttributeArgument> ();
-		}
-
-		public RemoveAttributeInstancesAttribute (CustomAttributeArgument value1)
-		{
-			Arguments = new[] { value1 };
+			// Arguments must be boxed in an Object CustomArgumentAttribute where the .Value is a CustomArgumentAttribute
+			Debug.Assert(values.All (a => a.Value is CustomAttributeArgument));
+			Arguments = values
```
marek-safar, line 23: Why is this needed if we control what is passed in?

marek-safar:
> Now that we generate arbitrary ctors for the attribute, it should be possible to allow the arguments in the xml to be unboxed without the Object wrapper to simplify things, and can probably be done without breaking "boxed" implementations of the xml. Curious to hear thoughts on that and if it's worth the effort and potential for breaking.

I think that's not what we want to do. The fact that the ctor needs to be generated is only "temporary" fix before someone add the exact attribute to the framework and we can more convenient way to remove attributes conditionally. It'd be beneficial if the signature of such type ctors would be same to avoid problems with 2 types doing the same thing.

src/linker/Linker.Steps/MarkStep.cs
jtschuster: I suppose that doesn't really get us anything since the first already factors in the type of the values, I'll remove that

src/linker/Linker/RemoveAttributeInstancesAttribute.cs
jtschuster: Since the list of arguments is filtered to only the boxed values to avoid crashing, I put a sanity check in there to avoid a situation where a new call to the ctor mistakenly passes in a list of "unboxed" arguments and later finds that none of them are actually stored in the property.

src/linker/Linker/RemoveAttributeInstancesAttribute.cs
jtschuster: No particular reason, would it be better to use `Collection<CustomAttributeArgument>`?

src/linker/Linker/RemoveAttributeInstancesAttribute.cs
marek-safar: I think it'd be better to show error message when someone enters wrong data instead of skipping wrong entries and possibly matching different overload

src/linker/Linker/RemoveAttributeInstancesAttribute.cs
marek-safar: Yes, less interface dispatch

src/linker/Linker/RemoveAttributeInstancesAttribute.cs
jtschuster: It looks like elsewhere we do ensure that the argument will be boxed in an `Object`. I'll leave the `Debug.Assert`, but I'll remove the `.Where` to make sure we don't silently match the wrong constructor.

jtschuster:
I updated this to allow `RemoveAttributeInstances` to match a constructor with `params`. I did this by flattening the `params` array from the attribute to be removed in MarkStep when we check if the arguments match, but I think it might make more sense to instead support arrays as an argument to `RemoveAttributeInstances`. Then, to match an attribute constructed with a `params` array, you would have to pass an array as the last argument to `RemoveAttributeInstances`.

@marek-safar @vitek-karas Do you have opinions here?

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -33,38 +33,69 @@ public void Parse (AttributeInfo xmlInfo)
...
+
+		CustomAttribute? _paramArrayCustomAttribute;
+		CustomAttribute? ParamArrayCustomAttribute {
+			get {
+				if (_paramArrayCustomAttribute is CustomAttribute ca)
+					return ca;
+				else {
+					var paramArrayAttributeType = BCL.FindPredefinedType ("System", "ParamArrayAttribute", _context);
+					if (paramArrayAttributeType == null)
+						return null;
```
vitek-karas, line 54: Why not call `ParamArrayAttributeType` here - it does the same thing?

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -90,11 +121,25 @@ static string FormatCustomAttribute (CustomAttribute ca)
 				return sb.ToString ();
 			}
 		}
-
-		TypeDefinition? GenerateRemoveAttributeInstancesAttribute ()
+		///	<summary>
+		/// Here, we generate new constructors for the synthetic RemoveAttributeInstancesAttribute type
+		///	(in the Cecil space, with no namespace) based on the number of arguments in the XML.
+		///	Each such .ctor is defined as .ctor(object, object, ....) - all arguments are boxed
+		///	in the XML by wrapping them with "object" argument.
```
vitek-karas, line 128: This seems to be out of date

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -107,26 +152,32 @@ static string FormatCustomAttribute (CustomAttribute ca)
...
-			td.Methods.Add (ctor);
+			if (td == null) {
+				td = new TypeDefinition ("", nameof (RemoveAttributeInstancesAttribute), TypeAttributes.Public);
+
+				td.BaseType = attributeType;
+			}
+
+			var ctorN = new MethodDefinition (".ctor", ctorAttributes, voidType);
+			var param = new ParameterDefinition (objectArrayType);
+			param.CustomAttributes.Add (ParamArrayCustomAttribute);
```
vitek-karas, line 178: Nit: technically there's no reason to add this attribute here - linker doesn't care, it will consume the object array directly. The attribute is really only useful for compilers when somebody tries to call the method, the compiler knows that it should pass additional parameters as an array. At runtime it's mostly ignored.

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -154,26 +205,39 @@ static string FormatCustomAttribute (CustomAttribute ca)
...
+				var parameters = method.Parameters;
+				if (args.Length != parameters.Count
+					&& !(parameters.Count > 0
+						&& method.Parameters.Last ().CustomAttributes
+							.Any (ca => _context.Resolve (ca.AttributeType) == ParamArrayAttributeType)))
 					continue;
 
 				bool match = true;
-				for (int ii = 0; match && ii != args.Length; ++ii) {
+				for (int ii = 0; match && ii < args.Length; ++ii) {
```
vitek-karas, line 220: Why do we need this code here - I thought we are always using the object[] constructor now.

vitek-karas:
I must admit I'm a bit confused. Why don't we use just one constructor the RemoveAttributeInstance - the one which takes `params object[]`? Yes - everything will be boxed, but that's a technical detail which we can easily hide. I guess I don't see the value of being able to generate different .ctor overloads... it seems to just complicate things.

src/linker/Linker.Steps/LinkAttributesParser.cs
marek-safar: Note, I'd keep the default ctor

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -843,29 +843,95 @@ bool IsAttributeRemoved (CustomAttribute ca, TypeDefinition attributeType)
...
+
+				return true;
+			}
+
+			static bool HasMatchingParamArrayArguments (CustomAttributeArgument[] removeAttrInstancesArgs, CustomAttribute attributeInstance)
+			{
+				var attributeInstanceArgs = attributeInstance.ConstructorArguments;
+				int i = 0;
+				// Check regular parameters
+				for (; i < attributeInstanceArgs.Count - 1; ++i) {
```
marek-safar, line 897: Can this be removed if only array based overload exist?

src/linker/Linker.Steps/LinkAttributesParser.cs
jtschuster: If I understand what you're pointing to, this code finds the matching constructor for any attribute added by the XML, so it needs to account for more than `RemoveAttributeInstances`

jtschuster:
After realizing that `params` arrays are just C# sugar and end up looking like a regular array in IL, I decided to pivot force users to match a param array with a regular array in the xml. I had to add functionality for building array `CustomAttributeArgument`s in the GetCustomAttributeArgument function. There I also added automatic boxing of arguments within `object[]` arrays. This way, you could write
```xml 
<attribute internal="RemoveAttributeInstances">
  <argument type="System.Object[]">
    <argument type="System.Int32">99</argument>
    <argument type="System.String">remove3</argument>
  </argument>
</attribute>
```
instead of 
```xml
<attribute internal="RemoveAttributeInstances">
  <argument type="System.Object[]">
    <argument type="System.Object">
      <argument type="System.Int32">99</argument>
    </argument>
    <argument type="System.Object">
      <argument type="System.String">remove3</argument>
    </argument>
  </argument>
</attribute>
```

Also, to make matching `CustomAttributeArgument`s easier, I created an extension method `IsEqualTo()` for the struct. It ignores any boxing in object types and checks only the value (or values for arrays) contained in the "botttom most" `CustomAttributeArgument`.

jtschuster:
> I must admit I'm a bit confused. Why don't we use just one constructor the RemoveAttributeInstance - the one which takes `params object[]`? Yes - everything will be boxed, but that's a technical detail which we can easily hide. I guess I don't see the value of being able to generate different .ctor overloads... it seems to just complicate things.

@vitek-karas I've updated `RemoveAttributeInstances` to have .ctors that either take `void`, `object` (to not break runtime), and `params object[]`. This did really simplify things, and I think this addresses your concern, but let me know if it doesn't.

src/linker/Linker/CustomAttributeSource.cs
```diff
@@ -37,7 +37,14 @@ public static AssemblyDefinition GetAssemblyFromCustomAttributeProvider (ICustom
...
 		{
-			var assembly = GetAssemblyFromCustomAttributeProvider (provider);
+			AssemblyDefinition? assembly;
+			try {
+				// Might be the internal RemoveAttributesInstancesAttribute which has no Module or Assembly and throws a NullRef
+				assembly = GetAssemblyFromCustomAttributeProvider (provider);
+			} catch (NullReferenceException) {
+				xmlInfo = null;
+				return false;
+			}
```
vitek-karas, line 47: Exceptions should not be used for unexceptional cases. If the code ran into the case where the provider is the made-up attribute, then we should either check for that specifically or make the code able to handle such case.

vitek-karas (changes requested):

src/linker/Linker/CustomAttributeSource.cs
jtschuster: I went to fix this, but I can't reproduce the exception anymore, so I removed the try/catch

vitek-karas (approved):

PR #2517, opened 2022-01-19
Create LinkAttributes.xml data model and add XmlAnalyzer

jtschuster:
Creates data structures to model the LinkAttributes.xml data in src/ILLink.Shared/XmlAttributeParser.cs. The helpers take an XDocument and return a tree of the LinkAttributes records modeling the xml data. No error catching or rule enforcement happens in creating the model, so the analyzers and linker will have to produce the errors as they are encountered when traversing the model. 

The XML analyzer currently reports just a single diagnostic for duplicate method parameters in xml. Reporting diagnostics for malformed Xml will be easy and only necessary in the XmlAnalyzer. However, sharing which attributes are injected into types presents some problems. For example, when an action for symbol with an injected attribute is called, the attribute may not have been resolved yet. My thought is that this might be possible by using async actions and awaiting a future action to resolve a symbol, but I haven't thought through the details on if that would work. Also, it's unclear if and how this data might be shared across analyzers to reduce repeated work.

Includes changes from PR https://github.com/dotnet/linker/pull/2537


src/ILLink.RoslynAnalyzer/XmlAttributeParser.cs
```diff
@@ -0,0 +1,308 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using ILLink.Shared;
+using System.Collections;
+using System.Net.NetworkInformation;
+using System.Data;
+using System.Collections.Generic;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace ILLink.RoslynAnalyzer
```
vitek-karas, line 24: We should share this code between linker and analyzer. The XML format is the same, and that's all this code cares about. The specifics of converting the results into Cecil or Roslyn representations can be built on top of this.

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Collections.Generic;
```
vitek-karas, line 5: Nit: Is this necessary?

src/ILLink.Shared/XmlAttributeParser.cs
```diff
@@ -0,0 +1,347 @@
...
+			// TODO: handle the correct exceptions correctly
+			catch (ArgumentException) {
+				// XML doesn't have a <linker> tag and is invalid. The document should have been validated before 
+				return new List<IRootNode> ();
+			}
+		}
+
+		public record AttributeNode : NodeBase
+		{
+			public string Fullname;
```
vitek-karas, line 28: Nit: Maybe should be `FullName` ?

src/ILLink.Shared/XmlAttributeParser.cs
```diff
@@ -0,0 +1,347 @@
...
+					var attr = new AttributeNode (attributeNav);
+					if (attr == null)
+						continue;
+					attributes.Add (attr);
+				}
+				Attributes = attributes;
+			}
+		}
+
+		public abstract record FeatureGatedNode : AttributeTargetNode
```
vitek-karas, line 311: I think we should change the name here. Main reason is that it's "new" - we don't use "feature gate" anywhere else. I understand that the meaning is slightly different to the widely used "feature switch", but maybe it would be better to keep the same terminology.

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
```diff
@@ -0,0 +1,57 @@
...
+using ILLink.Shared;
+using Microsoft.CodeAnalysis;using Microsoft.CodeAnalysis.Diagnostics;
+using System.IO;
+using Microsoft.CodeAnalysis.Text;
+using System.Text;
+using System.Xml.Linq;
+
+namespace ILLink.RoslynAnalyzer
+{
+	public abstract class XmlAnalyzer : DiagnosticAnalyzer
```
vitek-karas, line 15: Do we have a plan how to share this between the analyzers?

I guess maybe a better question: Why is this a separate analyzer? So that it can generate its own diagnostics for the problems found inside the XMLs?

jtschuster: Yes, I was planning to have each analyzer refer to the same "model" and use it however they need to. I was planning to use a separate XML to generate all diagnostics related to the xml, like malformed xml or not being able to resolve a type. This way we wouldn't generate the same warnings multiple times from each analyzer. Since the model would ideally be shared between the linker and analyzer, I didn't want to report diagnostics when the model is being built, so the model may have bad data that the XmlAnalyzer would report.

src/ILLink.Shared/XmlAttributeParser.cs
jtschuster: Do you mean something like `FeatureSwitch` and `FeatureSwitchedNode`?

vitek-karas: `FeatureSwitchNode` could work. Or maybe a descriptive `NodeWithFeatureSwitch` ...

Technically in the linker we don't call them "feature switches", just "features", but that is a very generic name I think. so maybe "feature switches" in good enough.

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
vitek-karas: OK- makes sense. It's going to be a bit tricky as there are grey areas of overlap - "is this problem part of the XML, part of the way we apply the XML to the program?". But that's relatively minor.

vitek-karas:
There's something weird going on - the diff for this PR includes changes which should be in main already...

src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
```diff
@@ -68,6 +68,9 @@ static string GetDiagnosticCategory (DiagnosticId diagnosticId)
 			case >= 3050 and <= 6000:
 				return DiagnosticCategory.AOT;
 
+			case 1013:
+				return DiagnosticCategory.XML;
```
tlakollo, line 72: I think the CategoryRanges were made to englobe similar diagnostics, 0 to 1000 is invalid, 1000 to 2000 is trimming errors, 2000 to 3000 is trimming warnings, etc. Xml is inherently part of the trimming, so maybe we could have a subcategory instead? (see https://github.com/dotnet/linker/blob/main/src/ILLink.Shared/DiagnosticId.cs#L194-L210 and https://github.com/dotnet/linker/blob/main/src/ILLink.Shared/MessageSubCategory.cs)

I guess the big difference would be on the suppression side, our code fixers use the diagnostic category to create the UnconditionalSuppressMessage attribute values.

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
```diff
@@ -0,0 +1,114 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+
+namespace ILLink.RoslynAnalyzer
+{
+	[DiagnosticAnalyzer (LanguageNames.CSharp)]
+	public class XmlAnalyzer : DiagnosticAnalyzer
+	{
+		private static readonly DiagnosticDescriptor s_moreThanOneValueForParameterOfMethod = DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.XmlMoreThanOneValyForParameterOfMethod);
```
tlakollo, line 24: I introduce a typo on the diagnosticId it says Valy instead of Value, maybe you can fix the diagnostic and the resources

src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
```diff
@@ -59,7 +59,7 @@ public static DiagnosticDescriptor GetDiagnosticDescriptor (DiagnosticId diagnos
 		static string GetDiagnosticCategory (DiagnosticId diagnosticId)
 		{
 			switch ((int) diagnosticId) {
-			case > 2000 and < 3000:
+			case > 1000 and < 3000:
```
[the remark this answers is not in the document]
jtschuster: In the meantime should I make 1000 - 2000 a new DiagnosticCategory for XML?

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
```diff
@@ -0,0 +1,114 @@
...
+				document = XDocument.Load (xmlStream, LoadOptions.SetLineInfo | LoadOptions.PreserveWhitespace);
+			} catch (XmlException ex) {
+				context.ReportDiagnostic (Diagnostic.Create (s_errorProcessingXmlLocation, null, ex.Message));
+				return false;
+			}
+			XmlSchemaSet schemaSet = new XmlSchemaSet ();
+			schemaSet.Add (LinkAttributesSchema);
+			bool valid = true;
+			document.Validate (schemaSet, (sender, error) => {
+				context.ReportDiagnostic (Diagnostic.Create (s_errorProcessingXmlLocation, null, "ILLinkAttributes.xml': '" + "At line " + error.Exception.LineNumber + ": " + error.Message));
```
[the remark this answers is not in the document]
jtschuster: I'm realizing now that the XML schema enforcement overlaps with a number of the trimmer warnings, so this `ReportDiagnostic` should probably be removed.
Related to that though, would it be better / possible to remove the diagnostics that can be caught by a schema validation and only include things like failing to resolve type names?

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
```diff
@@ -0,0 +1,114 @@
...
+				context.RegisterAdditionalFileAction (additionalFileContext => {
+					// Check if it's a LinkAttributes xml
+					if (_linkAttributesRegex.IsMatch (Path.GetFileName (additionalFileContext.AdditionalFile.Path))) {
+						if (additionalFileContext.AdditionalFile.GetText () is not SourceText text)
+							return;
+
+						if (!ValidateLinkAttributesXml (additionalFileContext, text))
+							return;
+
+						if (!context.TryGetValue (text, ProcessXmlProvider, out var xmlData) || xmlData is null)
```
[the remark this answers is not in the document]
jtschuster: That makes the most sense to me, but I'm not totally sure. The documentation says it "Returns true on success, false otherwise"

src/ILLink.Shared/XmlAttributeParser.cs
```diff
@@ -0,0 +1,354 @@
...
+				if (feature is null && featurevalue is null && featuredefault is null) {
+					return;
+				}
+				// if (feature is null && (featurevalue is not null || featuredefault is not null)) ;
+				// error, feature must be defined if featurevalue or featuredefault is defined
+				FeatureSwitch = new FeatureSwitch (feature, featurevalue, featuredefault);
+			}
+		}
+
+		public record FeatureSwitch
```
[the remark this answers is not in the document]
jtschuster: Won't it be over the recommended 16 bytes for a value type?

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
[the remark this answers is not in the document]
sbomer: I'm personally all for using the schema to validate the XML. It will depend on what the schema validation warnings look like - we should try not to give worse warnings than the linker (I'm guessing the schema validator will give better information if anything). We might want to map some of the validation errors to the existing error codes.

I'd suggest doing whatever is easiest for now - since the more important piece is to get the correctly-formatted XML files working. The XML files are an advanced scenario with a UX that already isn't great so improvements are welcome, but really if we cared about the UX we would be designing a different solution that doesn't use these XML files in the first place. :)

src/ILLink.Shared/XmlAttributeParser.cs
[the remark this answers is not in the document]
sbomer: Where is that recommended? I was just thinking that it has value semantics - it's just three pointers that we want to pass around together as a group. From a performance standpoint, I would guess that it's probably not worth  allocating an object on the heap to store those pointers, but I doubt this will be on the critical path for performance anyway.

src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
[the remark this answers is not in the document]
sbomer: I think they are not all XML - for example 1007 is about a well-known type failing to be resolved.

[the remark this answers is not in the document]
tlakollo: I though SuppressTrimAnalysisWarnings operated over the subcategory and not the category

src/ILLink.Shared/XmlAttributeParser.cs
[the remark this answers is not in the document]
jtschuster: It's recommended here: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct but since it's only really used when stored in other reference types and not passed/copied around the stack outside of constructors it probably wouldn't matter much.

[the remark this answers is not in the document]
sbomer: Thanks! I'll leave it up to you - I don't feel strongly about it.

src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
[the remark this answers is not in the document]
sbomer: Yeah, I think the linker calls a subcategory what the analyzer calls a category (a category in the linker is just error/warning, etc.). I think we should keep them the same.

src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
```diff
@@ -16,6 +16,10 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(MicrosoftCodeAnalysisVersion)" PrivateAssets="all" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="..\..\external\corert\**\*.cs" />
```
tlakollo, line 20: Why do we need the corert module?

test/ILLink.RoslynAnalyzer.Tests/XmlText.cs
```diff
@@ -1,12 +1,12 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.IO;
```
tlakollo, line 4: Nit: the usings should be sorted

test/ILLink.RoslynAnalyzer.Tests/TestCaseCompilation.cs
```diff
@@ -10,8 +10,8 @@
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Diagnostics;
-using Microsoft.CodeAnalysis.Text;
 using Mono.Linker.Tests.Cases.Expectations.Metadata;
+using Microsoft.CodeAnalysis.Text;
```
tlakollo, line 14: Nit: the usings should be sorted

src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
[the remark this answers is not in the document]
jtschuster: I made a `DiagnosticCategory` called TrimmingError for 1000-2000.

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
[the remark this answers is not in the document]
jtschuster: For now I'll leave it as is, and in the future I'll look into making xml schema errors a single error code in a separate PR.

src/ILLink.Shared/XmlProcessorBase.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System.Xml.XPath;
+
+namespace ILLink.Shared
+{
+	public abstract class XmlProcessorBase
```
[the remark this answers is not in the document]
jtschuster: Yes, since ProcessLinkerXml is a class and NodeBase is a record and ideally we will share the NodeBase implementation in the linker, I made it base of NodeBase.

src/ILLink.RoslynAnalyzer/ILLink.RoslynAnalyzer.csproj
jtschuster: I remember needing it for something at some point, but building without it passes, so I'll remove it.

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+			try {
+				document = XDocument.Load (xmlStream, LoadOptions.SetLineInfo | LoadOptions.PreserveWhitespace);
+			} catch (XmlException ex) {
+				context.ReportDiagnostic (Diagnostic.Create (s_errorProcessingXmlLocation, null, ex.Message));
+				return false;
+			}
+			XmlSchemaSet schemaSet = new XmlSchemaSet ();
+			schemaSet.Add (LinkAttributesSchema);
+			bool valid = true;
+			document.Validate (schemaSet, (sender, error) => { });
```
vitek-karas, line 87: If I read this correctly there will be no reported warning if the XSD validation fails - so what's the point of doing it?

src/ILLink.Shared/XmlAttributeParser.cs
```diff
@@ -0,0 +1,317 @@
...
+			{
+				FullName = GetFullName (nav);
+				Types = new List<TypeNode> ();
+				foreach (XPathNavigator typeNav in nav.SelectChildren (TypeElementName, "")) {
+					Types.Add (new TypeNode (typeNav));
+				}
+			}
+		}
+
+		public interface IRootNode
```
vitek-karas, line 235: Nit: In XML context "root" is the one element at the root of the document. If I read this correctly, this interface is not it - instead it's the "top-level" element in a way (the root is always "linker", right?).

jtschuster: Good point, I updated the name to `ITopLevelNode`

src/ILLink.Shared/XmlAttributeParser.cs
```diff
@@ -0,0 +1,317 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
sbomer, line 1: Rename this file to `LinkAttributes.cs` to match the class

src/ILLink.Shared/XmlAttributeParser.cs
```diff
@@ -0,0 +1,317 @@
...
+			public IXmlLineInfo? LineInfo;
+			protected NodeBase (XPathNavigator nav)
+			{
+				LineInfo = (nav is IXmlLineInfo lineInfo) ? lineInfo : null;
+			}
+
+			public static List<ITopLevelNode> ProcessRootNodes (XPathNavigator nav)
+			{
+				if (!nav.MoveToChild (LinkerElementName, XmlNamespace)) {
+					throw new ArgumentException ($"XML does not have <{LinkerElementName}> base tag");
```
sbomer, line 250: Generally we shouldn't throw exceptions on malformed XML. Is the idea here that the schema validation will cover it already, and we can never get here for an XML which doesn't have `<linker>` at the root? If that's the case it seems fine.

src/ILLink.Shared/XmlAttributeParser.cs
```diff
@@ -0,0 +1,317 @@
...
+				{
+					var type = argNav.GetAttribute ("type", "");
+					if (type == "") type = "System.String";
+					if (type == "System.Object") {
+						foreach (XPathNavigator innerArgNav in argNav.SelectChildren (ArgumentElementName, "")) {
+							if (ProcessArgument (innerArgNav) is AttributeArgumentNode innerArg) {
+								return new AttributeArgumentBoxNode (innerArg);
+							} else {
+								// error
+								throw new ArgumentException ("Arguments to attribute cannot be wrapped in more than one type");
```
sbomer, line 71: See my other comment about `ArgumentException` - same question here. Will the schema validation catch this? We should add tests for this to ensure it behaves as expected.

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
```diff
@@ -0,0 +1,117 @@
...
+			try {
+				document = XDocument.Load (xmlStream, LoadOptions.SetLineInfo | LoadOptions.PreserveWhitespace);
+			} catch (XmlException ex) {
+				context.ReportDiagnostic (Diagnostic.Create (s_errorProcessingXmlLocation, null, ex.Message));
+				return false;
+			}
+			XmlSchemaSet schemaSet = new XmlSchemaSet ();
+			schemaSet.Add (LinkAttributesSchema);
+			bool valid = true;
+			document.Validate (schemaSet, (sender, error) => { valid = false; });
```
sbomer, line 91: Can we get the line number and position out of the validation error, and report it as the location of the error?

src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
```diff
@@ -59,7 +59,10 @@ public static DiagnosticDescriptor GetDiagnosticDescriptor (DiagnosticId diagnos
 		static string GetDiagnosticCategory (DiagnosticId diagnosticId)
 		{
 			switch ((int) diagnosticId) {
-			case > 2000 and < 3000:
+			case >= 1000 and < 2000:
+				return DiagnosticCategory.TrimmingError;
```
sbomer, line 63: Is the category mandatory? In the linker these just get `MessageSubCategory.None`. `TrimmingError` sounds a bit off to me because it suggests that there was an error in the trimmer. If we really need to introduce a category, I think something closer to your original change, maybe `TrimXml` would be more appropriate - but it should only be applied to the specific warning codes that are about Xml.

src/ILLink.Shared/XmlAttributeParser.cs
```diff
@@ -0,0 +1,317 @@
...
+{
+	public static class LinkAttributes
+	{
+		public static List<ITopLevelNode> ProcessXml (XDocument doc)
+		{
+			try {
+				XPathNavigator nav = doc.CreateNavigator ();
+				return NodeBase.ProcessRootNodes (nav);
+			}
+			// TODO: handle the correct exceptions correctly
```
sbomer, line 20: Clarify or remove the comment

sbomer (commented):
Looking good!

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
```diff
@@ -0,0 +1,117 @@
...
+
+			context.RegisterCompilationStartAction (context => {
+				// Report Diagnostics on malformed XML with additionalFileContext and register actions to resolve names with (CompilationStartAnalysisContext) context.
+				context.RegisterAdditionalFileAction (additionalFileContext => {
+					// Check if it's a LinkAttributes xml
+					if (_linkAttributesRegex.IsMatch (Path.GetFileName (additionalFileContext.AdditionalFile.Path))) {
+						if (additionalFileContext.AdditionalFile.GetText () is not SourceText text)
+							return;
+
+						Debug.Assert (ValidateLinkAttributesXml (additionalFileContext, text));
```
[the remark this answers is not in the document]
jtschuster: I looked at the validation error messages, and a lot of them aren't super helpful, and some restrctions we want can't be enforced by any xml schema (e.g. schema can't enforce that an `<attribute>` tag needs either `fullname` or `internal` but not both). So I wanted to only enforce the validation on the xml in our tests.

This reminds me that we have tests that check the error messages on invalid xml, and if we throw at this assert those tests will fail, so maybe this should be removed altogether.

src/ILLink.RoslynAnalyzer/DiagnosticDescriptors.cs
jtschuster: I guess the category isn't mandatory, I just wanted to avoid throwing an exception so I added the category for 1XXX as an error during trimming. Maybe just `Error` makes more sense? Otherwise I could use None and make the `DiagnosticCategory` enum value an empty string.

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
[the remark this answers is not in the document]
sbomer: If the goal is just to validate that our test XML is well-formed, maybe you could move it to the test infrastructure, before the XML is passed into the compilation?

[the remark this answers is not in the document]
jtschuster: Yes, that's the plan for now.

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
jtschuster: Yes, that is possible. My concern here is we would still have to report some errors when parsing the xml, and reporting errors in two different places seems like a bad idea.

src/ILLink.Shared/XmlAttributeParser.cs
jtschuster: I updated the parsing to record diagnostics in the nodes of the tree built in ParseXml, then the analyzer emits as it traverses the tree. This way, I can remove the exceptions from ParseXml.

src/ILLink.Shared/XmlAttributeParser.cs
jtschuster: I removed exceptions from xml parsing, so I'll resolve this conversation.

test/ILLink.RoslynAnalyzer.Tests/LinkAttributesTests.cs
```diff
@@ -31,5 +33,10 @@ public Task LinkerAttributeRemovalConditional ()
 		{
 			return RunTest (nameof (LinkerAttributeRemovalConditional));
 		}
+		[Fact (Skip = "XML Analyzer not implemented")]
```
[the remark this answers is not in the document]
jtschuster: Yes, I've included the `LinkAttributesErrorCases` test for the exceptions the analyzer supports, and will add more as I add more functionality.

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
```diff
@@ -0,0 +1,117 @@
...
+
+						if (!context.TryGetValue (text, ProcessXmlProvider, out var xmlData) || xmlData is null) {
+							additionalFileContext.ReportDiagnostic (Diagnostic.Create (s_errorProcessingXmlLocation, null, additionalFileContext.AdditionalFile.Path));
+							return;
+						}
+						foreach (var root in xmlData) {
+							if (root is LinkAttributes.TypeNode typeNode) {
+								foreach (var duplicatedMethod in typeNode.Methods.GroupBy (m => m.Name).Where (m => m.Count () > 0)) {
```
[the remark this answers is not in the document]
jtschuster: Yes, thank you for catching that

src/ILLink.RoslynAnalyzer/XmlAnalyzer.cs
```diff
@@ -0,0 +1,117 @@
...
+							return;
+
+						Debug.Assert (ValidateLinkAttributesXml (additionalFileContext, text));
+
+						if (!context.TryGetValue (text, ProcessXmlProvider, out var xmlData) || xmlData is null) {
+							additionalFileContext.ReportDiagnostic (Diagnostic.Create (s_errorProcessingXmlLocation, null, additionalFileContext.AdditionalFile.Path));
+							return;
+						}
+						foreach (var root in xmlData) {
+							if (root is LinkAttributes.TypeNode typeNode) {
```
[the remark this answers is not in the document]
jtschuster: Yes, I've updated it and it should report for every method now

PR #2523, opened 2022-01-19
[DAM analyzer] Check for unexpected analyzer warnings in linker tests

sbomer:
This adds a source generator which generates analyzer xunit facts for the linker testcases. It discovers linker testcases by examining the referenced assembly, and avoids generating cases which already exist as analyzer facts (this requires tests to follow the existing convention for naming testcase files and facts). The generated sources are checked in to make changes easier to review.

The generated testcases only validate that the analyzer produces no unexpected warnings on the input (that is, these tests may pass even if the analyzer doesn't produce some of the expected warnings). Any tests which are currently skipped don't get this validation, so before closing out the feature branch we will want to replace skipped facts with facts that run in the `allowMissingWarnings` mode.

The failing tests have been disabled and are tracked in separate issues. This also contains a number of small infrastructure tweaks to match the linker's test infrastructure enough to run all of these tests.


LakshanF (approved):
I didn't quite fully understand the part of the generated analyzer test passing even if it doesn't produce expected warnings. Does not having `[ExpectedWarning]` in the linker test ensure that at least the analyzer achieves parity with the linker via `TestChecker`?

Thanks for automating this process!

test/ILLink.RoslynAnalyzer.Tests/ILLink.RoslynAnalyzer.Tests.csproj
```diff
@@ -1,17 +1,24 @@
...
   <PropertyGroup>
     <Nullable>enable</Nullable>
     <DeterministicSourcePaths>false</DeterministicSourcePaths>
+    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
+    <CompilerGeneratedFilesOutputPath>generated</CompilerGeneratedFilesOutputPath>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="$(CompilerGeneratedFilesOutputPath)" />
+  </ItemGroup>
```
vitek-karas, line 12: I don't really understand this - we ask the compiler to write the generated code to disk (I get that, at least for validation purposes and such). But then remove it from the compilation... this is probably because the source generator already added it?

Also - why do we commit the generated files into the source control, if they are always regenerated?

vitek-karas (approved):
Looks good - only one question about the compiler/source control setup.

sbomer:
> Does not having [ExpectedWarning] in the linker test ensure that at least the analyzer achieves parity with the linker via TestChecker?

Generally yes, but in this test run the generated tests (and any tests which pass `allowMissingWarnings: true`) specifically _don't_ check that all of the `ExpectedWarnings` are covered. They only check that there aren't any extra warnings. We can remove the `allowMissingWarnings` from individual tests as they reach parity with the linker.

test/ILLink.RoslynAnalyzer.Tests/ILLink.RoslynAnalyzer.Tests.csproj
sbomer: Yes, if we didn't remove them from the compilation they would be defined twice. I'm committing them to source control to make it easier to review changes, and because it sounds like the test UI may have trouble displaying generated tests otherwise. I don't have a strong opinion on it but this is per @agocke's suggestion - happy to remove them if we would rather not check in a bunch of generated sources.

agocke: For code that you're expected to interact w/ directly (e.g. tests), I'd prefer to have them checked in. VS Code can't run them individually otherwise

agocke (approved):

test/ILLink.RoslynAnalyzer.Tests/ILLink.RoslynAnalyzer.Tests.csproj
agocke: Also, to properly enable these test cases, it will be easy to simple move the file into a non-generated directory. The generator will see that the test file is already present and will stop generating it, and git will understand and track the move operation.

PR #2524, opened 2022-01-19
Share annotation validation

vitek-karas:
The new bit of shared code is the `RequireDynamicallyAccessedMembersAction` which is the core logic of how to validate annotations on source -> target assignments.

This change includes:
- Move `HandleCall` from `Intrinsics` into its own `HandleCallAction`
- Introduce the new `RequireDynamicallyAccessedMembersAction`
- Introduce `DiagnosticContext` which is currently only used by `RequireDynamicallyAccessedMembersAction` but it will be used by `HandleCallAction` as well in the future.

This doesn't really enable any new tests on the analyzer side of things because:
- In order to support `KnownStringValue` it would need support to resolve type name -> type (missing right now, another PR in the future) - there are some interesting questions how to do this in the analyzer since it doesn't have a global view.
- In order to support `UnknownValue` we need the visitors to actually report it, currently they don't - another discussion how to implement this.
- "marking" of types in the analyzer - not yet implemented

### DIFF
--- a/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
@@ -105,8 +105,8 @@ namespace ILLink.RoslynAnalyzer.TrimAnalysis
 
 		public override MultiValue HandleMethodCall (IMethodSymbol calledMethod, ValueOfOperation instance, ImmutableArray<ValueOfOperation> arguments, IOperation operation)
 		{
-			var handleCallACtion = new HandleCallAction (Context, operation);
-			if (handleCallACtion.Invoke (new MethodProxy (calledMethod), instance.Value, arguments.Select (a => a.Value).ToImmutableList (), out MultiValue methodReturnValue))
+			var handleCallAction = new HandleCallAction (Context, operation);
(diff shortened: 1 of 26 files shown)


src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -105,8 +105,8 @@ public override void HandleAssignment (MultiValue source, MultiValue target, IOp
 
 		public override MultiValue HandleMethodCall (IMethodSymbol calledMethod, ValueOfOperation instance, ImmutableArray<ValueOfOperation> arguments, IOperation operation)
 		{
-			Intrinsics intrinsics = new (Context, operation);
-			if (intrinsics.HandleMethodCall (new MethodProxy (calledMethod), instance.Value, arguments.Select (a => a.Value).ToImmutableList (), out MultiValue methodReturnValue))
+			var handleCallACtion = new HandleCallAction (Context, operation);
+			if (handleCallACtion.Invoke (new MethodProxy (calledMethod), instance.Value, arguments.Select (a => a.Value).ToImmutableList (), out MultiValue methodReturnValue))
```
sbomer, line 109: nit: handleCallAction

sbomer (approved):
It took me a second to understand why we needed a DiagnosticContext - but given the way we have the origin in an ambient context in the linker it makes sense. What's the use case for HandleCallAction to use the DiagnosticContext?

PR #2530, opened 2022-01-20
[release/6.0.2xx] Fix analyzer nullref on assembly attribute

sbomer:
The analyzer runs for property assignment operations, including those in
attributes. To check whether the property assignment should warn, we
look for RUC on the containing symbol. However, assembly-level attributes
are contained in the global namespace, which has a null containing type.

## Customer Impact

Assembly-level attribute with property assignment causes a crash in the RequiresAttribute analyzers. https://github.com/dotnet/linker/issues/2563 hits what is likely this issue.

## Testing

Unit tests added in this change. This was originally discovered as part of running the analyzer over the existing linker tests as part of https://github.com/dotnet/linker/pull/2523.

## Risk

Very low risk - the change adds an extra null check without impacting other code paths.


jtschuster (approved):
LGTM

agocke:
This will have to go through QB approval. Could you make the change in main first? I'm not sure it will meet the bar without a customer report.

vitek-karas (approved):

jeffschwMSFT (approved):
Approved.  Please take for consideration in 6.0.2xx

PR #2537, opened 2022-01-22
Enable LinkAttributes.xml to be passed to analyzers in unit tests in AdditionalFiles

jtschuster:
Creates a new implementation of the `AdditionalText` class to use when passing xml files to the analyzers and adds test file path parameters to `CreateCompilation` signatures to find the xml files. The names of the test files are found in the `[SetupCompileResourceAttribute]` and `[SetupLinkAttributesFile]` attributes, then the paths are resolved relative to the test files and added to the compilation as "ILLink.LinkAttributes.xml". This does not yet add substitution and descriptor xml files, but it should be easy to add those later.

LinkAttributesTests test cases always fail, but they all include a LinkAttributes.xml file and are good for debugging to check that the ILLink.LinkAttributes.xml file shows up in the analyzers.

### DIFF
--- a/test/ILLink.RoslynAnalyzer.Tests/TestCaseCompilation.cs
+++ b/test/ILLink.RoslynAnalyzer.Tests/TestCaseCompilation.cs
@@ -30,17 +30,15 @@ namespace ILLink.RoslynAnalyzer.Tests
 			(string, string)[]? globalAnalyzerOptions = null,
 			IEnumerable<MetadataReference>? additionalReferences = null,
 			IEnumerable<SyntaxTree>? additionalSources = null,
-			string? testCaseDirectory = null,
-			string? testCaseName = null)
-			=> CreateCompilation (CSharpSyntaxTree.ParseText (src), globalAnalyzerOptions, additionalReferences, additionalSources, testCaseDirectory, testCaseName);
+			IEnumerable<AdditionalText>? additionalFiles = null)
+			=> CreateCompilation (CSharpSyntaxTree.ParseText (src), globalAnalyzerOptions, additionalReferences, additionalSources, additionalFiles);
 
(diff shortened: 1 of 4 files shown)


vitek-karas (approved):

sbomer (commented):
Nice, thank you!

test/ILLink.RoslynAnalyzer.Tests/TestCaseCompilation.cs
```diff
@@ -43,9 +51,26 @@ internal static class TestCaseCompilation
 				syntaxTrees: sources,
 				references: (await TestCaseUtils.GetNet6References ()).Add (mdRef).AddRange (additionalReferences),
 				new CSharpCompilationOptions (OutputKind.DynamicallyLinkedLibrary));
-
+			List<AdditionalText> additionalTexts = new ();
```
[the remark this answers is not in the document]
jtschuster: Good idea, I moved this logic out to the same place as `GetTestDpendencies`

test/ILLink.RoslynAnalyzer.Tests/TestCaseUtils.cs
```diff
@@ -67,10 +72,10 @@ public static async Task RunTestFile (string suiteName, string testName, params
 
 		private static IEnumerable<string> GetTestDependencies (string rootSourceDir, SyntaxTree testSyntaxTree)
 		{
+			List<AdditionalText> additionalTexts = new ();
```
sbomer, line 75: Unused variable

sbomer (approved):

test/ILLink.RoslynAnalyzer.Tests/TestCaseCompilation.cs
```diff
@@ -5,10 +5,14 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Net.NetworkInformation;
```
tlakollo, line 8: Why this library is needed?

PR #2538, opened 2022-01-22
Title and message resources should be enforced to exist to prevent printing empty messages

tlakollo:
Currently, missing resources will default to an empty string printing an empty message for the diagnostics. Code should not try to gracefully handle errors. 

Included in this PR:
DiagnosticString should be able to locate a title and message resource strings otherwise it throws
Add title resource strings for all missing diagnostic Ids

### DIFF
--- a/src/ILLink.Shared/DiagnosticString.cs
+++ b/src/ILLink.Shared/DiagnosticString.cs
@@ -13,15 +13,15 @@ namespace ILLink.Shared
 		public DiagnosticString (DiagnosticId diagnosticId)
 		{
 			var resourceManager = SharedStrings.ResourceManager;
-			_titleFormat = resourceManager.GetString ($"{diagnosticId}Title") ?? throw new NullReferenceException($"{diagnosticId} does not have a matching resource called {diagnosticId}Title");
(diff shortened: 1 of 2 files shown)


src/ILLink.Shared/DiagnosticString.cs
```diff
@@ -11,15 +13,15 @@ namespace ILLink.Shared
 		public DiagnosticString (DiagnosticId diagnosticId)
 		{
 			var resourceManager = SharedStrings.ResourceManager;
-			_titleFormat = resourceManager.GetString ($"{diagnosticId}Title") ?? string.Empty;
-			_messageFormat = resourceManager.GetString ($"{diagnosticId}Message") ?? string.Empty;
+			_titleFormat = resourceManager.GetString ($"{diagnosticId}Title") ?? throw new NullReferenceException($"{diagnosticId} does not have a matching resource called {diagnosticId}Title");
```
vitek-karas, line 16: `NullReferenceException` sounds weird here. Maybe `ArgumentOutOfRange` or just `InvalidOperation`.... ?

vitek-karas (approved):

PR #2540, opened 2022-01-23
Add analyzer bootstrap phase

agocke:
Should run by default in CI and catch analyzer crashes. Adding this already found one -- the fix is in the second commit.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,26 +1,6 @@
 #!/usr/bin/env bash
 
-# Stop script if unbound variable found (use ${var:-} if intentional)
-set -u
-
-# Stop script if subcommand fails
-set -e
-
 source="${BASH_SOURCE[0]}"
-
-bootstrap=false
-
-args=""
-while [[ $# > 0 ]]; do
-  opt="$(echo "${1/#--/-}" | tr "[:upper:]" "[:lower:]")"
-  case "$opt" in
-    -integrationtest)
-      bootstrap=true
-      ;;
-  esac
-  args="$args $1"
-  shift
-done
 
 # resolve $SOURCE until the file is no longer a symlink
 while [[ -h $source ]]; do
--- a/build.sh
+++ b/build.sh
@@ -1,6 +1,26 @@
 #!/usr/bin/env bash
 
+# Stop script if unbound variable found (use ${var:-} if intentional)
+set -u
+
+# Stop script if subcommand fails
+set -e
+
 source="${BASH_SOURCE[0]}"
+
+bootstrap=false
+
+args=""
+while [[ $# > 0 ]]; do
+  opt="$(echo "${1/#--/-}" | tr "[:upper:]" "[:lower:]")"
+  case "$opt" in
+    -integrationtest)
+      bootstrap=true
+      ;;
+  esac
+  args="$args $1"
+  shift
+done
 
 # resolve $SOURCE until the file is no longer a symlink
 while [[ -h $source ]]; do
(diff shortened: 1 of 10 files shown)


build.sh
```diff
@@ -1,7 +1,27 @@
...
+  opt="$(echo "${1/#--/-}" | tr "[:upper:]" "[:lower:]")"
+  case "$opt" in
+    -integrationtest)
+      bootstrap=true
+      ;;
+  esac
+  args="$args $1"
+  shift
+done
+
```
vitek-karas, line 24: I'm no bash expert, but I don't see how the `bootstrap` value makes it into the rest of the build?
Maybe it's intentional that this only works on Windows? (I would very much vote against doing that, but maybe for now it's OK).

eng/Analyzers.props
```diff
@@ -4,4 +4,16 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeStyle" Version="$(MicrosoftCodeAnalysisCSharpCodeStyleVersion)" PrivateAssets="all" />
     <!-- <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.205" PrivateAssets="all" /> -->
   </ItemGroup>
+
+  <Import Project="$(BootstrapBuildPath)/Microsoft.NET.ILLink.Analyzers.props" Condition="'$(BootstrapBuildPath)' != ''" />
+
+  <!-- Don't enable for Cecil, as they can't be suppressed -->
+  <PropertyGroup Condition="'$(BootstrapBuildPath)' != '' and '$(MSBuildProjectName)' != 'Mono.Cecil.Pdb'">
```
vitek-karas, line 11: Why just `Mono.Cecil.Pdb` and not the other cecil projects (mainly the `Mono.Cecil` itself)?

eng/common/build.ps1
```diff
@@ -33,7 +33,7 @@ Param(
 # Unset 'Platform' environment variable to avoid unwanted collision in InstallDotNetCore.targets file
 # some computer has this env var defined (e.g. Some HP)
 if($env:Platform) {
-  $env:Platform=""  
+  $env:Platform=""
```
vitek-karas, line 36: I don't think we should touch the file at all - otherwise Arcade updates could cause merge conflicts.
In addition - these are just whitespace changes, so not needed.

marek-safar:
I'm not sure how much value this adds. Wouldn't be better to write integration tests with e.g. SPC where different code patterns actually exist?

eng/common/build.ps1
agocke: Yeah, looks like VS Code got a bit too aggressive with the whitespace changes.

build.sh
agocke: I can revert this change -- I copied it over from my other bootstrap change and found a simpler way, so we no longer need it.

eng/Analyzers.props
agocke: Mono.Cecil.PDB has a warning (COM) that I need to file a bug for.

We also can't disable it because it sets `<NoWarn>` without using `$(NoWarn)` inside, so it overrides the higher warnings. None of the other projects do, so we can temporarily disable warnings for them.

agocke:
@marek-safar /runtime is definitely a good test suite, but our changes take longer to flow there. Since this already found a crash, I think it's worth bringing in.

eng/Analyzers.props
vitek-karas: We could push a change to the project to correctly pick up `NoWarn` from the outside.
Once we have XML support it should be possible to get rid of the warning that way.

vitek-karas (approved):

PR #2548, opened 2022-01-25
Record call patterns

vitek-karas:
The main change is that method calls are now tracked as separate trim analysis pattern. This allows us to handle diagnostics on method calls via the `HandleMethodCallAction` (which currently doesn't do this, but in the future it will).

I didn't create abstraction over trim analysis patterns yet - as it seemed unnecessary (we only have 2 right now and this way its faster).

The main downside of this approach is that we don't produce as precise location for warnings as before. Currently any warning generated due to a method call will be reported with the location of the entire call site. Previously we pointed at certain argument if possible. Going forward this could be improved, but it requires adding some notion of location into the shared code and for the `HandleMethodCallAction` to carefully report warnings with the right location.

The other thing to improve going forward can be to somehow not do the "Merge" of the recorded patterns. This is currently relatively expensive operation. But the best way to solve this would be to reduce the number of passes of the solver - ideally for most methods it should do exactly one pass. Once we're there, the perf of the "Merge" should not matter much as it will be very rare.


src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -47,9 +47,8 @@ public override void Initialize (AnalysisContext context)
 					ControlFlowGraph cfg = context.GetControlFlowGraph (operationBlock);
 					TrimDataFlowAnalysis trimDataFlowAnalysis = new (context, cfg);
 
-					foreach (TrimAnalysisPattern trimAnalysisPattern in trimDataFlowAnalysis.ComputeTrimAnalysisPatterns ()) {
-						foreach (var diagnostic in GetDynamicallyAccessedMembersDiagnostics (trimAnalysisPattern.Source, trimAnalysisPattern.Target, trimAnalysisPattern.Operation.Syntax.GetLocation ()))
-							context.ReportDiagnostic (diagnostic);
+					foreach (var diagnostic in trimDataFlowAnalysis.ComputeTrimAnalysisPatterns ().ReportDiagnostics ()) {
```
sbomer, line 50: Maybe we can rename `ReportDiagnostics` to make this read better - how about `CollectDiagnostics` or similar? I realize it might actually report things for the linker, but it would be nice to pick a more neutral name. I would also rename `DIagnosticContext.ReportDiagnostic` to `AddDiagnostic` which can mean either report for the linker or add to a collection for the analyzer.

sbomer (commented):
I'd like it if we could continue reporting the more accurate warning location for arguments.

Not sure if this would work, but here's an idea: what if we continue to use TrimAnalysisPattern (with source/target values) for the argument values, and instead of collecting method call patterns, allow the intrinsic handling in HandleCallAction.Invoke to add TrimAnalysisPatterns?

Some of the intrinsics would just record normal patterns for the annotated arguments, but others (like GetMethod I think) would not. Instead, we can add another cache for members accessed by intrinsics when they are more specific. For example, `typeof(KnownType).GetMethod("Method")` would save the return value which is a `MultiValue` containing `SystemReflectionMethodBaseValue`. Then later when reporting diagnostics, we would iterate over these and see if any of the accessed members have annotations.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
sbomer: Which part of the linker do you mean? I don't see the changes - did you push to the PR? Thanks!

vitek-karas: There's nothing in the linker yet, but going forward the "patterns" is something we should be able to share, at which point this will be relevant.

PR #2551, opened 2022-01-26
Add a link to an issue for existing test case

vitek-karas:
"fixes" a test case in `GetInterface` which was not meant to run into the wrong linker behavior as per bug https://github.com/dotnet/linker/issues/2550


test/Mono.Linker.Tests.Cases/DataFlow/LocalDataFlow.cs
```diff
@@ -221,13 +221,14 @@ public static void TestBranchGoto ()
 			return;
 		}
 
+		// https://github.com/dotnet/linker/issues/2550
```
sbomer, line 224: Would you mind moving the call to this method in `Main` to the section with "extraneous warnings"? And the issue link could go there as well.

sbomer (approved):

PR #2553, opened 2022-01-27
Fix inconsistencies in Requires attribute checks

sbomer:
While investigating failures in https://github.com/dotnet/linker/pull/2536 I came across a bunch of places where we check for RequiresAttribute without considering RUC on type.

- DAM analyzer:
  - checking for dataflow warnings. @tlakollo fixed these in https://github.com/dotnet/linker/pull/2536/commits/8121647f09929ed49f072ff878fe2681361c1c80. This change contains a few more tests.
  - checking for generic method warnings. These tests aren't turned on yet but I added a few testcases that should be enabled as part of https://github.com/dotnet/linker/issues/2364.
- Requires analyzer:
  - checking for attributes on ctors of types substituted for generic parameters with new constraint
  - the above should be silenced if the declaring type of the caller has RequiresAttribute
  - checking for RequiresAttribute on attribute ctors should check the attribute type for RequiresAttribute
  - the above should be silenced if the declaring type of the attribute provider has RequiresAttribute (but not when the provider is itself a type) (https://github.com/dotnet/linker/issues/2529)
- RUC analyzer:
  - warnings for use of `dynamic` should be silenced in a method whose declaring type has RequiresAttribute

I added tests for all of the above. Also fixes https://github.com/dotnet/linker/issues/2529.

I noticed another inconsistency in the attribute checks that I'm not addressing here: https://github.com/dotnet/linker/issues/2554

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -183,7 +183,8 @@ namespace ILLink.RoslynAnalyzer
 							if (instanceCtor.Arity > 0)
 								continue;
 
-							if (instanceCtor.TargetHasRequiresAttribute (RequiresAttributeName, out var requiresAttribute)) {
+							if (instanceCtor.TargetHasRequiresAttribute (RequiresAttributeName, out var requiresAttribute) &&
+								VerifyAttributeArguments (requiresAttribute)) {
 								syntaxNodeAnalysisContext.ReportDiagnostic (Diagnostic.Create (RequiresDiagnosticRule,
 									syntaxNodeAnalysisContext.Node.GetLocation (),
 									containingSymbol.GetDisplayName (),
@@ -365,30 +366,6 @@ namespace ILLink.RoslynAnalyzer
 		{
 			var url = requiresAttribute?.NamedArguments.FirstOrDefault (na => na.Key == "Url").Value.Value?.ToString ();
 			return MessageFormat.FormatRequiresAttributeUrlArg (url);
-		}
-
-		/// <summary>
-		/// This method determines if the member has a Requires attribute and returns it in the variable requiresAttribute.
-		/// </summary>
-		/// <param name="member">Symbol of the member to search attribute.</param>
-		/// <param name="requiresAttribute">Output variable in case of matching Requires attribute.</param>
-		/// <returns>True if the member contains a Requires attribute; otherwise, returns false.</returns>
-		private bool TryGetRequiresAttribute (ISymbol? member, [NotNullWhen (returnValue: true)] out AttributeData? requiresAttribute)
-		{
-			requiresAttribute = null;
-			if (member == null)
-				return false;
-
-			foreach (var _attribute in member.GetAttributes ()) {
-				if (_attribute.AttributeClass is { } attrClass &&
-					attrClass.HasName (RequiresAttributeFullyQualifiedName) &&
-					VerifyAttributeArguments (_attribute)) {
-					requiresAttribute = _attribute;
-					return true;
-				}
-			}
-
-			return false;
 		}
 
 		/// <summary>
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -151,7 +151,7 @@ namespace ILLink.RoslynAnalyzer
 
 				context.RegisterSyntaxNodeAction (syntaxNodeAnalysisContext => {
 					var model = syntaxNodeAnalysisContext.SemanticModel;
-					if (syntaxNodeAnalysisContext.ContainingSymbol is not ISymbol containingSymbol || containingSymbol.HasAttribute (RequiresAttributeName))
+					if (syntaxNodeAnalysisContext.ContainingSymbol is not ISymbol containingSymbol || containingSymbol.IsInRequiresScope (RequiresAttributeName))
 						return;
 
 					GenericNameSyntax genericNameSyntaxNode = (GenericNameSyntax) syntaxNodeAnalysisContext.Node;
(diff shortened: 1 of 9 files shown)


src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
```diff
@@ -51,6 +51,11 @@ public static bool IsOverrideInRequiresScope (this ISymbol member, string requir
 
 		private static bool IsInRequiresScope (this ISymbol member, string requiresAttribute, bool checkAssociatedSymbol)
 		{
+			// Requires attribute on a type does not silence warnings that originate
+			// from the type directly.
+			if (member is ITypeSymbol)
+				return false;
+
 			if (member is ISymbol containingSymbol) {
```
tlakollo, line 59: I think this line makes no sense, we know that the member is ISymbol, because that's what we get as input so we are effectively only changing the name to containingSymbol, but is not the containingSymbol, so I guess we went off track at some point.

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
```diff
@@ -51,6 +51,11 @@ public static bool IsOverrideInRequiresScope (this ISymbol member, string requir
...
 				if (containingSymbol.HasAttribute (requiresAttribute)
 					|| (containingSymbol is not ITypeSymbol &&
```
tlakollo, line 61: Considering that containingSymbol is actually member and we already check that member is not ITypeSymbol this check can get deleted too

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
sbomer: I think it was there as a null check. I am not sure if it's necessary but I will keep it to be safe. I made the null check more explicit to avoid just renaming the variable.

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
sbomer: Good point, fixed.

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
sbomer: I think it was there as a null check. Not sure how necessary it is, but I'd rather keep it. I changed the code to make it more clear and to avoid pointlessly changing the variable name. Thanks!

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
sbomer: Good point, fixed.

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
tlakollo: But we have nullable enabled, the ISymbol cannot be null. Also Im not sure if calling it containingSymbol, the fact that this is a containingSymbol or not is completely on the caller.

sbomer: I was a little worried that we could be getting a null ISymbol from a null-oblivious API somewhere, especially since this logic has caused nullref exceptions in the past. But I looked at all the callers and I think it's ok, so I removed the null check.

For the name, I changed it to just `symbol`. `member` doesn't seem right since it can be called on types. Then again, the method doesn't make sense for all kinds of `ISymbol`, so I don't really have a strong opinion - let me know if you prefer.

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
jtschuster: Is there any enforcement that it is a member symbol? It looks like the method and callers assume it is right now, but I wonder if we should enforce that somehow.

tlakollo (approved):
Looks good to me

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
tlakollo: When I was doing the reflection call GetMembers() nested types is one of the things that returns, so I consider a type as a member (see https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo?view=net-6.0)

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -183,12 +183,13 @@ public override void Initialize (AnalysisContext context)
 							if (instanceCtor.Arity > 0)
 								continue;
 
-							if (instanceCtor.TryGetAttribute (RequiresAttributeName, out var requiresUnreferencedCodeAttribute)) {
+							if (instanceCtor.TargetHasRequiresAttribute (RequiresAttributeName, out var requiresAttribute) &&
+								VerifyAttributeArguments (requiresAttribute)) {
```
vitek-karas, line 187: In @tlakollo PR there will be a single method to do this at once: https://github.com/dotnet/linker/pull/2418/files#diff-6417c9266a49f206907dd0ece745dc0c77c5f5c9a0c92ec3bb749d4c8e2c039aR14-R15
Please make sure we end up using it here as well.

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
```diff
@@ -49,18 +49,21 @@ public static bool IsOverrideInRequiresScope (this ISymbol member, string requir
...
-						containingType.HasAttribute (requiresAttribute))) {
-					return true;
-				}
+			// Requires attribute on a type does not silence warnings that originate
+			// from the type directly. We also only check the containing type for members
+			// below, not of nested types.
+			if (symbol is ITypeSymbol)
+				return false;
+
+			if (symbol.HasAttribute (requiresAttribute)
```
vitek-karas, line 60: I'm thinking a safer thing would be to call "Validate" on the attribute as well - it basically checks that the attribute is in the form we recognize. As a safe guard against potential future changes which could for example add some kind of "feature condition" to the RUC attribute (I know we've discussed this), in which case the attribute is not always "active", so it's safer to NOT recognize it.

vitek-karas (approved):
Looks good - maybe just one nit about making it future proof by always validating the RUC attribute.

sbomer:
Thanks @vitek-karas, I will track those suggestions in https://github.com/dotnet/linker/issues/2554

src/ILLink.RoslynAnalyzer/RequiresISymbolExtensions.cs
sbomer: Changed to `member`

PR #2555, opened 2022-01-27
RequiresUnreferencedCode tests with lambdas

jtschuster:
I wasn't sure how the analyzer and linker handle method calls inside of local lambdas, so I made some tests to find out and make sure everything works, and I think there is an inconsistency between the analyzer and linker. I'm not sure why the second `// Bug` isn't warning in the trimmer when it should be a pretty simple case, but it very well could be an issue with how I set up the tests.


test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresWithLambdas.cs
```diff
@@ -0,0 +1,101 @@
...
+				return () => { RucMethod (); return; };
+			}
+
+			// Bug - Should warn in Trimmer
+			[ExpectedWarning ("IL2026", nameof (ReturnRucAction), ProducedBy = ProducedBy.Analyzer)]
+			public void InvokeReturnedRucAction ()
+			{
+				var f = ReturnRucAction ();
+				f ();
+			}
```
vitek-karas, line 86: This is really weird - I do agree that on the surface this should warn. I'll have to look into it.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresWithLambdas.cs
```diff
@@ -0,0 +1,101 @@
...
+using System.Security.Permissions;
+using System.Text;
+using System.Threading.Tasks;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Helpers;
+
+namespace Mono.Linker.Tests.Cases.RequiresCapability
+{
+	[SkipKeptItemsValidation]
+	public class RequiresWithLambdas
```
vitek-karas, line 17: Ideally all tests should have `ExpectNoWarnings` attribute.
In which case you may want to add a suppression onto the Main as it will produce lot of 2026 warnings - but that's OK and the suppression is also OK - this test doesn't verify the exact warnings for that scope.

jtschuster:
It looks like it is warning at
```C#
public Action ReturnRucAction ()
{
	return () => { RucMethod (); return; };
}
```
but the error is shown as coming from the compiler generated private method
`Mono.Linker.Tests.Cases.RequiresCapability.RequiresWithLambdas.NoRequiresOnClass.<InvokeLocalRucAction>b__2_0()`

The tests have been updated accordingly.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresWithLambdas.cs
jtschuster: I'm not sure why, but as I fixed other parts of the tests, this warning started appearing in the linker.

vitek-karas:
There's a `CompilerGeneratedCode` property on the `ExpectedWarning` attribute for this - it should make it so that the expected warning matches on all of the compiler generated methods for that method (so lambdas and local functions).

jtschuster:
The expected behavior for lambdas in the docs differs from the expected behavior in the test. Updated the PR to just include a comment to reflect that.

vitek-karas (approved):

PR #2570, opened 2022-01-31
Use internal machine pool for src build

agocke:


eng/azure-pipelines.yml
```diff
@@ -90,7 +90,10 @@ stages:
       - job: SourceBuild_Managed
         displayName: Source-Build (Managed)
         pool:
-          vmImage: ubuntu-latest
+          ${{ if eq(variables.officialBuild, 'false' )}}:
+            vmImage: ubuntu-latest
+          ${{ if eq(variables.officialBuild, 'true' )}}:
+            name: NetCore1ESPool-Internal -equals Build.Ubuntu.1804.Amd64
```
sbomer, line 96: How does the `-equals` syntax work here? Is it supposed to have a `demands` section?

vitek-karas: Based on Matt's template I think this should have:
```yaml
demands: ImageOverride -equals Build.Ubuntu.1804.Amd64
```

agocke: Ah, you're right, I misread that

vitek-karas (approved):
Looks good, now the question is: "How do we test this" ?
But I think we should merge - let it mirror and then see how it goes.

agocke:
I know of no other way

sbomer (approved):

PR #2571, opened 2022-01-31
Share more intrinsics

vitek-karas:
Shares several more intrinsics, the most notable one is `Type.GetInterface`.
Changes:
* Change TypeProxy to represent a resolved type on the linker side (no change to analysis)
  This means that it can't represent generic instantiations, pointers and so on. Which is OK for the current usage.
  Generic instantiations will need to be handled separately anyway due to analyzer
  and the way they work in the linker is not ideal either (produce duplicate warnings in places).
  Currently only Activator.CreateInstance<T> requires knowledge of generic instantiations
  and that one is going to be removed as intrinsic (already has annotation which works the same).
* Add a type system proxy for generic parameter - this helps with representing `typeof(T)` like expressions
* `RequireDynamicallyAccessedMembersAction` is now part of `HandleCallAction` - enables the more interesting intrinsics like `GetInterface`
* Simplify some values to only-shared (no partial) due to type system proxy improvements
* Enable tests for the newly shared intrinsics
* Added some small improvements to tests


src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -38,6 +45,89 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
...
+				break;
+
+			case IntrinsicId.Type_get_TypeHandle:
+				foreach (var value in instanceValue) {
+					if (value is SystemTypeValue typeValue)
+						methodReturnValue = MultiValueLattice.Meet (methodReturnValue, new RuntimeTypeHandleValue (typeValue.RepresentedType));
+					else if (value is GenericParameterValue genericParameterValue)
+						methodReturnValue = MultiValueLattice.Meet (methodReturnValue, new RuntimeTypeHandleForGenericParameterValue (genericParameterValue.GenericParameter));
+					else if (value == NullValue.Instance)
+						methodReturnValue = MultiValueLattice.Meet (methodReturnValue, value);
```
sbomer, line 77: The docs say GetTypeHandle throws if the argument is null - should this maybe not add a return value?

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -38,6 +45,89 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
...
+			case IntrinsicId.Type_get_AssemblyQualifiedName: {
+					foreach (var value in instanceValue) {
+						if (value is ValueWithDynamicallyAccessedMembers valueWithDynamicallyAccessedMembers) {
+							// Currently we don't need to track the difference between Type and String annotated values
+							// that only matters when we use them, so Type.GetType is the difference really.
+							// For diagnostics we actually don't want to track the Type.AssemblyQualifiedName
+							// as the annotation does not come from that call, but from its input.
+							methodReturnValue = MultiValueLattice.Meet (methodReturnValue, valueWithDynamicallyAccessedMembers);
+						} else if (value == NullValue.Instance) {
+							// Just propagate null - it's not perfect, but we can't just skip it, we have to produce some value to the output
```
sbomer, line 121: What's wrong with just skipping it?

sbomer (approved):
LGTM, just a question about handling intrinsics that don't return.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
vitek-karas: Currently there's a technical problem - if the only input is `null`, and we skip it, then we produce empty methodReturnValue. The code below tries to handle cases where the intrinsics doesn't do anything, and thus if the return value is empty, it will replace it with the "return value of the method". Which is effectively unknown value which could lead to warnings.

We would have to either somehow remember that we did "set" the return value, make it nullable for example, or modify the code so that intrinsics never "not return value" in which case we could remove the code below.

sbomer: I see, thanks. I would personally prefer to remember whether the return value was set (I like the nullable suggestion) but am also ok with the workaround.

vitek-karas: I tried, but it causes problems in the linker. Currently it's pretty tricky - linker is not setup to return "empty value". If that happens the `MethodBodyScanner` basically fails thinking it was an invalid IL (since it expects a real value on the stack).
So I had to keep the behavior - I did cleanup return value handling a little bit, but not functionality.

vitek-karas:
@sbomer  Could you please take another look - specifically I added a new intrinsic: https://github.com/dotnet/linker/pull/2571/commits/20826198e66c604906c21dcddea130a048e66df7

(We already had a passing test for it so I didn't want to mark it as failing in the analyzer, so I ported it over).

sbomer (approved):
LGTM!

Issue #2574, opened 2022-02-01
Support array types in generic parameter analysis


sbomer:
For example, this testcase throws a `NotImplementedException`
```csharp
		[Kept]
		static void TestArrayOnGeneric ()
		{
			_ = new RequirePublicMethodsGeneric<ArrayElementInGenericType[]> ();
		}
```

https://github.com/dotnet/linker/blob/56421a1b1c1713d9d02567f481595d6b7b5ac000/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs#L114

(It was not initially hit in the generic parameter change by @LakshanF, but the fix in https://github.com/dotnet/linker/pull/2572/commits/d9f7347856fa589d879c28722f16a1456599caf8 caused us to hit more cases like the one above).

PR #2576, opened 2022-02-02
Support array types in generic parameter

LakshanF:
Fix for issue #2574, 

We create a `SystemTypeValue `of `System.Array` for Generic arguments that are of `SymbolKind.ArrayType`

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
@@ -114,7 +114,8 @@ namespace ILLink.RoslynAnalyzer
 				// That said we only use it to perform the dynamically accessed members checks and for that purpose treating it as System.Type is perfectly valid.
 				SymbolKind.NamedType => new SystemTypeValue ((INamedTypeSymbol) type),
 				SymbolKind.ErrorType => UnknownValue.Instance,
-				// What about things like ArrayType or PointerType and so on. Linker treats these as "named types" since it can resolve them to concrete type
+				SymbolKind.ArrayType => new ArrayValue (GetTypeValueNodeFromGenericArgument (((IArrayTypeSymbol) type).ElementType)),
+				// What about things like PointerType and so on. Linker treats these as "named types" since it can resolve them to concrete type
 				_ => throw new NotImplementedException ()
 			};
 		}
--- a/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
@@ -98,7 +98,7 @@ namespace ILLink.RoslynAnalyzer
 				Debug.Assert (typeParams.Length == typeArgs.Length);
 
 				for (int i = 0; i < typeParams.Length; i++) {
-					var sourceValue = GetTypeValueNodeFromGenericArgument (typeArgs[i]);
+					var sourceValue = GetTypeValueNodeFromGenericArgument (context, typeArgs[i]);
 					var targetValue = new GenericParameterValue (typeParams[i]);
 					foreach (var diagnostic in GetDynamicallyAccessedMembersDiagnostics (sourceValue, targetValue, context.Node.GetLocation ()))
 						context.ReportDiagnostic (diagnostic);
@@ -106,7 +106,7 @@ namespace ILLink.RoslynAnalyzer
 			}
 		}
 
-		static SingleValue GetTypeValueNodeFromGenericArgument (ITypeSymbol type)
+		static SingleValue GetTypeValueNodeFromGenericArgument (SyntaxNodeAnalysisContext context, ITypeSymbol type)
 		{
 			return type.Kind switch {
 				SymbolKind.TypeParameter => new GenericParameterValue ((ITypeParameterSymbol) type),
@@ -114,7 +114,7 @@ namespace ILLink.RoslynAnalyzer
 				// That said we only use it to perform the dynamically accessed members checks and for that purpose treating it as System.Type is perfectly valid.
 				SymbolKind.NamedType => new SystemTypeValue ((INamedTypeSymbol) type),
 				SymbolKind.ErrorType => UnknownValue.Instance,
-				SymbolKind.ArrayType => new ArrayValue (GetTypeValueNodeFromGenericArgument (((IArrayTypeSymbol) type).ElementType)),
+				SymbolKind.ArrayType => new SystemTypeValue (context.Compilation.GetTypeByMetadataName ("System.Array")!),
 				// What about things like PointerType and so on. Linker treats these as "named types" since it can resolve them to concrete type
 				_ => throw new NotImplementedException ()
 			};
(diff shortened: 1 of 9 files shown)


src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -114,7 +114,8 @@ static SingleValue GetTypeValueNodeFromGenericArgument (ITypeSymbol type)
 				// That said we only use it to perform the dynamically accessed members checks and for that purpose treating it as System.Type is perfectly valid.
 				SymbolKind.NamedType => new SystemTypeValue ((INamedTypeSymbol) type),
 				SymbolKind.ErrorType => UnknownValue.Instance,
-				// What about things like ArrayType or PointerType and so on. Linker treats these as "named types" since it can resolve them to concrete type
+				SymbolKind.ArrayType => new ArrayValue (GetTypeValueNodeFromGenericArgument (((IArrayTypeSymbol) type).ElementType)),
```
sbomer, line 117: I think it doesn't make sense to call `GetTypeValueNodeFromGenericArgument` on the element type - the element type isn't the generic argument.

src/ILLink.RoslynAnalyzer/TrimAnalysis/ArrayValue.cs
```diff
@@ -9,7 +9,12 @@ namespace ILLink.Shared.TrimAnalysis
 {
 	partial record ArrayValue
 	{
-		public ArrayValue (SingleValue size) => Size = size;
+		public SingleValue ElementType;
+		public ArrayValue (SingleValue elementType)
```
sbomer, line 13: I think we could make `ElementType` an `ITypeSymbol` in the roslyn-specific code, similar to how it's a `TypeReference` in the linker code.

sbomer:
Looks like the linker will resolve array type arguments to System.Array: https://github.com/dotnet/linker/blob/3f704bb49eb2fb2847ceaecfbabe8d5f3e0c86c9/src/linker/Linker.Dataflow/MethodBodyScanner.cs#L921-L930

(I don't know if we can get there with an array type argument since I think this code might resolve array types to their element type first: https://github.com/dotnet/linker/blob/3f704bb49eb2fb2847ceaecfbabe8d5f3e0c86c9/src/linker/Linker.Steps/MarkStep.cs#L2592 - it would be good to add tests to check that the required System.Array members are actually kept.)

It probably won't make a difference for the analyzer since I don't think System.Array has or will have annotated members, but I think the correct thing to do here to model reflection is to have the array case hit this logic for System.Array:
https://github.com/dotnet/linker/blob/554a87bb283054936c939bbc9f306f9e71f7062f/src/ILLink.Shared/TrimAnalysis/RequireDynamicallyAccessedMembersAction.cs#L21-L26

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
vitek-karas: In this case it also doesn't make sense to use `ArrayValue`. The generic parameter is a type, not a value. So even the result of typeof<int[]> is not an array value, it's still a type (`SystemTypeValue`).

vitek-karas: In linker what this does (roughly) is to mark the item type of the array as a type (so nothing on it) - this operation is something I don't know yet if it needs to be handled by the analyzer (the only case where it might matter is type hierarchy marking... but I didn't look into the details on it yet). The resulting type which we then track through data flow is just the basic `System.Array`, since that's what one would actually reflect on if this were used in reflection.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -114,7 +114,8 @@ static SingleValue GetTypeValueNodeFromGenericArgument (ITypeSymbol type)
...
+				// What about things like PointerType and so on. Linker treats these as "named types" since it can resolve them to concrete type
```
vitek-karas, line 118: Since we're here, we should handle these as well - linker resolves pointers and alike to their base element type - so for example "foo *" is resolved to "foo" and then tracked like that. I don't actually know if that is correct from semantics point of view (as in, what will reflection do). So it would be good to figure out the reflection behavior and potentially file an issue against linker as well.

src/ILLink.RoslynAnalyzer/TrimAnalysis/ArrayValue.cs
vitek-karas: As noted above, for this change we don't need to handle array values yet... so I would not try to solve it right now.

src/ILLink.Shared/TrimAnalysis/RequireDynamicallyAccessedMembersAction.cs
```diff
@@ -34,6 +34,8 @@ public void Invoke (in DiagnosticContext diagnosticContext, in MultiValue value,
 					}
 				} else if (uniqueValue == NullValue.Instance) {
 					// Ignore - probably unreachable path as it would fail at runtime anyway.
+				} else if (uniqueValue is ArrayValue) {
+					// Ignore
```
vitek-karas, line 38: Arrays in this case should be handled as "unknown" or "unsupported" and fall through to the else case below. It means that either data flow screwed up, or that there something really weird going on in the code.
Technically it should not be possible to get an ArrayValue here, the source/target for this method should be statically typed as either `Sytem.Type` or `System.String` and nothing else (since we can't/don't handle annotations on anything else).

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
sbomer: I don't think pointers can be used as generic arguments, at least in C# - do we even need to handle these?

vitek-karas: That's true, at least for now - I vaguely remember that there's some discussion to allow refs there (reflection could really use this to avoid boxing I think), but I could be wrong.

src/ILLink.RoslynAnalyzer/TrimAnalysis/ArrayValue.cs
```diff
@@ -9,7 +9,12 @@ namespace ILLink.Shared.TrimAnalysis
 {
 	partial record ArrayValue
 	{
-		public ArrayValue (SingleValue size) => Size = size;
+		public SingleValue ElementType;
```
vitek-karas, line 12: Can you please revert this file? We don't really know what we're going to need for arrays yet.

vitek-karas (approved):

src/ILLink.RoslynAnalyzer/TrimAnalysis/ArrayValue.cs
LakshanF: Oops, completely forgot this, thanks!

PR #2577, opened 2022-02-02
Handle object creation as a call to .ctor

vitek-karas:
This is necessary for correct data flow processing since we need to check the arguments to ctor the same way we do any other method.
Added couple of simple tests.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using ILLink.Shared.DataFlow;
+using ILLink.Shared.TrimAnalysis;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.FlowAnalysis;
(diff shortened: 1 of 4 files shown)


src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -221,5 +201,49 @@ public override TValue VisitConversion (IConversionOperation operation, LocalDat
 			var operandValue = Visit (operation.Operand, state);
 			return operation.OperatorMethod == null ? operandValue : TopValue;
 		}
+
+		public override TValue VisitObjectCreation (IObjectCreationOperation operation, LocalDataFlowState<TValue, TValueLattice> state)
+		{
+			if (operation.Constructor == null)
```
sbomer, line 207: Just curious, do you know if we can actually get here without a ctor?

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -221,5 +201,49 @@ public override TValue VisitConversion (IConversionOperation operation, LocalDat
...
+				return TopValue;
+
+			// Passing null for instance value is a bit wrong - below we will translate it to TopValue right now.
+			// Technically this is an instance call and the instance is some valid value, we just don't know which
+			// but for example it does have a static type. For now this is OK since we don't need the information
+			// for anything yet.
+
+			// The return here is also technically problematic, the return value is an instance of a known type.
```
sbomer, line 215: nit: Would you mind moving this comment to `HandleMethodCall`? (Since this code is generic and doesn't know about DAM analysis, and also because that's where the issue would probably get fixed).

sbomer (approved):

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
vitek-karas: I didn't try - but I would expect this to happen for compiler generated .ctors - as those don't exist yet. So for example class without explicit constructor - it very likely doesn't have any method symbol for the default .ctor, but it will eventually get one.

Fortunately we don't need to check for that in data flow. RUC needs to do that though and check for RUC on type in that case.

PR #2581, opened 2022-02-02
Fix lint job in CI

vitek-karas:
It did run linter, but didn't fail if any changes were to be applied, instead it actually applied them (and then dropped). So the job would actually never fail.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -135,6 +135,9 @@ dotnet_diagnostic.RS2008.severity = none
 # Exception type is not sufficiently specific
 dotnet_diagnostic.CA2201.severity = none
 
+# xUnit1004: Test methods should not be skipped
+dotnet_diagnostic.xUnit1004.severity = none
+
 [src/linker/ref/**/*.cs]
 
 # CA1822: Mark members as static
--- a/.editorconfig
+++ b/.editorconfig
@@ -146,6 +146,10 @@ dotnet_diagnostic.CA1822.severity = none
 # IDE0060: Remove unused parameter
 dotnet_diagnostic.IDE0060.severity = none
 
+[test/**/*.cs]
+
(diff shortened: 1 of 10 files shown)


sbomer (approved):
There are still some formatting failures in test files, otherwise LGTM.

vitek-karas:
This is really weird - I ran it locally to "fix" things, and it didn't do those - but then it fails on them?

sbomer:
Weird... could it be using a different version of dotnet-format locally for some reason? I have hit problems in the past with breaking changes to dotnet-format and IIRC I had to be very careful to make sure it used the right SDK.

vitek-karas:
Honestly - the tool is... confusing.

`dotnet format` may not fix some warnings - and it says so with something like `Unable to apply fixes for <id>`. But it does not show the warnings it wasn't able to fix.
`dotnet format --verify-no-changes` on the other hand shows these warnings.

You can force `dotnet format` to show them as well with `-v d`, but that produces lot of output - mostly useless and the warnings are sort of hidden in the middle.

`dotnet format` does not return error exit code if fails to fix things.

`dotnet format` may say that it failed to fix things, while it actually did fix them (specifically IDE00005 - unused using).

So for now I did this:
* Most of the warnings were unused parameters in tests -> I disabled those in the editorconfig as this is basically expected for tests
* I fixed one warning which made sense
* I modified our `lint` scripts to run the tool twice. First to actually apply formatting and second time with the `--verify-no-change` - this has the effect that it will report unfixed warnings and it will actually fail the command if there are any.

It's not perfect, it takes a while to run (twice) and the output is still unnecessarily long, but at least it works.

vitek-karas:
I filed https://github.com/dotnet/format/issues/1508 to hopefully improve the tool

marek-safar (approved):

Issue #2583, opened 2022-02-03
[feature/damAnalyzer] Warning code differ between the linker and analyzer


LakshanF:
The following code generates the warning [IL2077](https://github.com/dotnet/linker/blob/main/docs/error-codes.md#il2077-trim-analysis-target-parameter-argument-does-not-satisfy-dynamicallyaccessedmembersattribute-in-call-to-target-method-the-field-source-field-does-not-have-matching-annotations-the-source-value-must-declare-at-least-the-same-requirements-as-those-declared-on-the-target-location-it-is-assigned-to) in the linker and [IL2067 ](https://github.com/dotnet/linker/blob/main/docs/error-codes.md#il2067-trim-analysis-target-parameter-argument-does-not-satisfy-dynamicallyaccessedmembersattribute-in-call-to-target-method-the-parameter-source-parameter-of-method-source-method-does-not-have-matching-annotations-the-source-value-must-declare-at-least-the-same-requirements-as-those-declared-on-the-target-location-it-is-assigned-to)in the analyzer.

Impacts test [SuppressWarningsInCompilerGeneratedCode ](https://github.com/dotnet/linker/blob/feature/damAnalyzer/test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInCompilerGeneratedCode.cs)

**REPRO**
[Program.txt](https://github.com/dotnet/linker/files/7998820/Program.txt)

**Linker Warning**
`...Program.cs(21,5): Trim analysis warning IL2077: Mono.Linker.Tests.Cases.Warnings.WarningSuppression.SuppressWarningsInCompilerGeneratedCode.SuppressInIteratorBody.<TestMethodParameterWithRequirements>d__1.MoveNext(): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.NonPublicMethods' in call to 'Mono.Linker.Tests.Cases.Warnings.WarningSuppression.DataFlowTypeExtensions.RequiresNonPublicMethods(Type)'. The field 'System.Type Mono.Linker.Tests.Cases.Warnings.WarningSuppression.SuppressWarningsInCompilerGeneratedCode/SuppressInIteratorBody/<TestMethodParameterWithRequirements>d__1::unknownType' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. 
`

**Analyzer Warning**
 ```
[...]

Issue #2579, opened 2022-02-02
[DAM analyzer] Add support for attribute warning suppressions


sbomer:
The analyzer needs to support suppressing warnings like the linker does. `RUC` already works, but we still need to add support for `UnconditionalSuppressMessage` and probably also adding both of these attributes via XML.

edit: `UnconditionalSuppressMessage` should already be working since roslyn supports it natively - so for non-attribute suppressions we should only need to turn on the existing tests.

@jtschuster is working on attribute XML support so you might want to sync with him @LakshanF

sbomer:
These are the failing suppressions tests I noticed. I would start with the first two since they aren't about XML suppressions:
- Warnings.WarningSuppressionTests.SuppressWarningsInMembersAndTypesUsingTarget
warning suppressions in code using Target = "..." syntax
- Warnings.WarningSuppressionTests.SuppressWarningsInCompilerGeneratedCode
extra warnings in analyzer, needs investigation
- Warnings.WarningSuppressionTests.SuppressWarningsViaXml
analyzer needs to support XML attributes

MichalStrehovsky:
> analyzer needs to support XML attributes

I know we're trying to add support for reading the XML to the analyzer, but I've internally had question marks around that for a while that I didn't voice.

I see that allowing to do suppressions by faking the attributes through XML will add extra work to this since Roslyn normally handless all the suppressions for us (we don't need to parse and bind the `Target`, etc.).

Do we _really need_ the attribute XML support? I've always looked at the attribute XML as something that we have internally, but externally it's only used to remove attribute instances and that part is not even real attribute XML, it's just a hack on the file format. I don't see a compelling external use case for attribute XML that would warrant the support cost. When this was first introduced, it was only meant for testing (#1060).

It's kind of like if Roslyn provided an XML way to add nullable attributes to existing libraries for the purposes of compiling code against those libraries. Would some people find it useful? Probably. Would it be really fragile because it doesn't version with the assembly we're attributing? Definitely. Such feature would probably not meet the bar for Roslyn. Why does it meet the bar for us?

Cc @vitek-karas @agocke

vitek-karas:
@MichalStrehovsky I think that is fair analysis. As far as I know we currently use them for two purposes:
* Suppress warnings for assemblies which we didn't annotate yet
* Attribute removal

[...]

Issue #2578, opened 2022-02-02
[DAM analyzer] don't produce warnings for unsupported intrinsics


sbomer:
As per the title - until we implement support for an intrinsic, the DAM analyzer should not produce warnings for calls to the intrinsic (even when annotated with RUC), or for values returned from the intrinsic.

tlakollo:
I open this one for the same purpose https://github.com/dotnet/linker/issues/2559

sbomer:
Thanks, I missed that one when I was looking for an existing issue. Closing as a duplicate.

sbomer:
Actually - I will keep this issue open to track "disabling" the intrinsics (making it so they don't produce warnings, as a precondition for closing out the feature branch), and we can use your issue to track enabling them in the analyzer for feature parity.

sbomer:
Fixed by https://github.com/dotnet/linker/pull/2596 and https://github.com/dotnet/linker/pull/2597

PR #2584, opened 2022-02-04
Analyzer suppression test fix

LakshanF:
Fix for #2579, Roslyn handles `UnconditionalSuppressMessage ` on its own and we are not planning to add any special handling. The PR addresses the following 3 exceptions from the test bed.

- `SuppressWarningsInMembersAndTypesUsingTarget`: This test case uses [Expression.Call ](https://docs.microsoft.com/en-us/dotnet/api/system.linq.expressions.expression.call?view=net-6.0#system-linq-expressions-expression-call(system-type-system-string-system-type()-system-linq-expressions-expression()))which has a RUC but is treated as an intrinsic by the trimmer. Ensured that the analyzer handles this method correctly
- `SuppressWarningsInCompilerGeneratedCode`: Analyzer and Linker differ in the warning code generated. Filed #2583 (and #2587 also has relevant information) for this
- `SuppressWarningsViaXml`: As https://github.com/dotnet/linker/issues/2579#issuecomment-1028549170 discussion notes, we plan on not supporting this in the analyzer, at least for the time being. The test does not run in the analyzer.


vitek-karas:
> SuppressWarningsInMembersAndTypesUsingTarget: Analyzer was generating couple of 2026 warnings that seems reasonable. Suppressed them

If we think that the warning is valid and linker doesn't produce it then it's a bug in the linker. Somebody needs to investigate why the linker doesn't warn in this case.

LakshanF:
The reason the linker doesn't warn seems to be because we have `SkipKeptItemsValidation `on the test and linker's `ResultChecker `skips validation for the extra 2026 warnings needed. I was being a little lazy trying to understand the goal of this test with this combination of warning mix but will take a closer look at the intent of the test

vitek-karas:
`SkipKeptItemsValidation` should have nothing to do with warnings - this is weird.

tlakollo:
This test is failing because it doesn't expect RUC from Expression.Call(). In runtime Expression.Call() is annotated with RUC (https://github.com/dotnet/runtime/blob/895c99cc502913fd7393b95af7cae5b9e6fbdea3/src/libraries/System.Linq.Expressions/ref/System.Linq.Expressions.cs#L439). Linker has intrinsics for handling it and that's why it doesn't warn. What I think the fix should be:
- Add Expression.Call into RequiresUnreferencedCodeAnalyzer ReportSpecialIncompatibleMembersDiagnostic list of methods, next to MakeGenericMethod and MakeGenericType.

tlakollo:
I was looking that currently the analyzer does the same with calls to `System.Type.GetType` (its annotated with RUC in runtime https://github.com/dotnet/runtime/blob/main/src/coreclr/System.Private.CoreLib/src/System/Type.CoreCLR.cs and we have an intrinsic that 'knows better' how to handle it) it might be worth it to just check all intrisics and reuse the logic from the shared project in ReportSpecialIncompatibleMembersDiagnostic.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -84,15 +85,22 @@ protected override ImmutableArray<ISymbol> GetSpecialIncompatibleMembers (Compil
 				incompatibleMembers.AddRange (methodInfoType.GetMembers ("MakeGenericMethod").OfType<IMethodSymbol> ());
 			}
 
+			var expressionCallType = compilation.GetTypeByMetadataName ("System.Linq.Expressions.Expression");
+			if (expressionCallType != null) {
+				incompatibleMembers.AddRange (expressionCallType.GetMembers ("Call").OfType<IMethodSymbol> ()
+					.Where (m => m.Parameters.Length == 4 && m.Parameters[0] is IParameterSymbol parameterSymbol
```
tlakollo, line 91: The list of incompatible member purposes is to say "RequiresUnreferencedCodeAnalyzer please don't analyze this call" then the dataflow will handle the intrinsic and decide if it needs to warn or not. You are restricting the number of parameters and type of the parameter, which means that RequiresUnreferencedCodeAnalyzer will only bypass that case and for everything else will warn. Im afraid that dataflow will do something similar and we end up generating two warnings, I think what we want is:
```diff
- incompatibleMembers.AddRange (expressionCallType.GetMembers ("Call").OfType<IMethodSymbol> ()
-					.Where (m => m.Parameters.Length == 4 && m.Parameters[0] is IParameterSymbol parameterSymbol
-					&& parameterSymbol.Type?.ContainingNamespace?.Name == "System" && parameterSymbol.Type?.Name == "Type"));
+ incompatibleMembers.AddRange (expressionCallType.GetMembers ("Call").OfType<IMethodSymbol> ());
```

LakshanF: unfortunately, there are lots of these Call methods and only 1 should be handled intrinsically. The code mimics the [linker ](https://github.com/dotnet/linker/blob/feature/damAnalyzer/src/ILLink.Shared/TrimAnalysis/Intrinsics.cs#L59-L63)version. There are some other expression members that need special handling and I'm taking a look at those as part of #2578

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -84,15 +85,22 @@ protected override ImmutableArray<ISymbol> GetSpecialIncompatibleMembers (Compil
...
 			return incompatibleMembers.ToImmutable ();
 		}
 
 		protected override bool ReportSpecialIncompatibleMembersDiagnostic (OperationAnalysisContext operationContext, ImmutableArray<ISymbol> specialIncompatibleMembers, ISymbol member)
 		{
 			if (member is IMethodSymbol method && ImmutableArrayOperations.Contains (specialIncompatibleMembers, member, SymbolEqualityComparer.Default) &&
-				(method.Name == "MakeGenericMethod" || method.Name == "MakeGenericType")) {
-				// These two RUC-annotated APIs are intrinsically handled by the trimmer, which will not produce any
-				// RUC warning related to them. For unrecognized reflection patterns realted to generic type/method
+				(method.Name == "MakeGenericMethod" || method.Name == "MakeGenericType" || method.Name == "Call")) {
```
sbomer, line 101: Why do we need to check the method name if we are already comparing with `specialIncompatibleMembers`? I think we could get rid of the name check here. cc @tlakollo

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
sbomer: I think https://github.com/dotnet/linker/issues/2578 will require a more general solution but I am fine with this change doing the minimum necessary to get the suppression tests working.

sbomer (commented):
Thanks @LakshanF!

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
LakshanF: missed this duplicate check, thanks!

PR #2592, opened 2022-02-08
Run dotnet format only once

sbomer:
@vitek-karas I saw some strange behavior in the lint script again in https://github.com/dotnet/linker/pull/2585. It was failing in PR but not locally. I tracked this down to the fact that our `global.json` doesn't specify an SDK version, so it was running locally with a version of `dotnet format` from the 7.0 SDK I have installed.

The pipeline is already passing `--verify-no-changes` so I don't think we need to run it twice.

I took the `allowPrerelease` and `rollForward` settings from dotnet/runtime to make it possible to build the repo locally even with a later SDK (even though it becomes possible then to run into this issue). I don't have a strong preference on it and would be fine with requiring the exact SDK if you prefer.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,4 +1,9 @@
 {
+  "sdk": {
+    "version": "6.0.100",
+    "allowPrerelease": true,
+    "rollForward": "major"
+  },
   "tools": {
     "dotnet": "6.0.100"
   },
--- a/lint.sh
+++ b/lint.sh
@@ -14,4 +14,3 @@ done
 
 scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
 "$scriptroot/eng/dotnet.sh" format illink.sln --exclude external $@
-"$scriptroot/eng/dotnet.sh" format illink.sln --exclude external --verify-no-changes $@


vitek-karas:
I agree that my solution is partially wrong because it runs the same command in CI twice...
And having a somewhat fixed SDK version is definitely a good idea.

That said, removing the second invocation locally removes the "fix" I tried to make. You can try this, just add an internal method with unused parameter and run lint. It should "fail" due to `IDE0060`, but with your change it behaves as before - it simply says "Unable to fix IDE0060. No associated code fix found." and succeeds the command. It doesn't show why/where the problem is.

So I think that locally we should still run the tool twice - once to apply any automatic fixes it can, and second time to report detailed warnings for the things it couldn't fix.

The CI should be fine running it once with `--verify-no-changes`.

sbomer:
Thanks @vitek-karas, I see what you mean. Personally I wouldn't mind having to run `./lint.sh --verify-no-changes` locally to get the details, but I see the problem that running it twice addresses. It's confusing either way. I'll leave it as-is and just fix the SDK version here.

vitek-karas (approved):

PR #2593, opened 2022-02-08
[DAM analyzer] Analyze exception filters

sbomer:
This adds support for exception filters in the analyzer. Now the exception state will flow from from try blocks to any catch blocks without filters, or to the filter regions of any catch blocks that do have filters. Additionally, the dataflow state flows from a filter region to the next filter or catch. We don't check whether the exception type of the filter regions match, but conservatively assume that control can flow out of any filter to the next filter or catch.

This also assumes that control leaves the filter region from its last block - I don't know if there are cases where a filter region can contain more than one basic block, so if this is possible the logic might need to be updated.

I added a few testcases with dataflow and side effects in filter regions. This fixes https://github.com/dotnet/linker/issues/2415 and allows enabling two more of the previously failing tests.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DataFlow/ControlFlowGraphProxy.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/ControlFlowGraphProxy.cs
@@ -130,10 +130,8 @@ namespace ILLink.RoslynAnalyzer.DataFlow
 			throw new InvalidOperationException ();
 		}
 
-		public bool TryGetPreviousFilter (RegionProxy catchOrFilterRegion, out RegionProxy filterRegion)
+		public IEnumerable<RegionProxy> GetPreviousFilters (RegionProxy catchOrFilterRegion)
 		{
-			filterRegion = default;
-
 			var region = catchOrFilterRegion.Region;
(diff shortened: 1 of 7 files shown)


MichalStrehovsky:
> This also assumes that control leaves the filter region from its last block - I don't know if there are cases where a filter region can contain more than one basic block, so if this is possible the logic might need to be updated.

From the IL perspective (not sure how much that is relevant to the analyzer), a ​valid filter region will have exactly one `endfilter` instruction and it will be the last instruction of the filter block.

There can be multiple basic blocks within the filter but they all have to lead to the single `endfilter`:

```csharp
Type academicalSituation = null;
try { throw new Exception(); }
catch (Exception ex) when ((ex is NullReferenceException ? (academicalSituation = typeof(object)) : (academicalSituation = typeof(string))) == null) { }
``

sbomer:
Thanks for the example @MichalStrehovsky, I added a test similar to it. It looks like Roslyn does insert a final basic block that merges the two branches and is the source for the fall-through case (where control doesn't enter the corresponding handler), so the assumption holds at least for this case.

vitek-karas (approved):
Looks good

vitek-karas (approved):

PR #2595, opened 2022-02-09
[DAM analyzer] Fix property set expression value

sbomer:
The value of a property set expression is the value, not the return value of the set method (void).


vitek-karas (approved):
Looks good.

PR #2596, opened 2022-02-09
[DAM analyzer] Disable dataflow warnings for unimplemented intrinsics

sbomer:
This skips unimplemented intrinsics in the dataflow analyzer and enables more tests. Addresses part of https://github.com/dotnet/linker/issues/2578. Now the remaining warnings in skipped tests are all about RUC which @LakshanF is working on.

PR #2597, opened 2022-02-10
No analyzer warnings for intrinsics

### LINKED ISSUE #2578: [DAM analyzer] don't produce warnings for unsupported intrinsics

LakshanF:
Fixes #2578 - temporary fix so as for the analyzer to not generate warnings until we add intrinsic support. This PR suppresses RUC and combined with #2596 should generate no warnings.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
@@ -73,16 +73,10 @@ namespace ILLink.RoslynAnalyzer
 		protected override bool IsAnalyzerEnabled (AnalyzerOptions options, Compilation compilation) =>
 			options.IsMSBuildPropertyValueTrue (MSBuildPropertyOptionNames.EnableTrimAnalyzer, compilation);
 
-		/// <summary>
-		/// Add methods that the Trimmer handles intrinsically to an exlusionary list
-		/// </summary>
-		/// <param name="compilation"></param>
-		/// <returns></returns>
-
 		protected override bool ReportSpecialIncompatibleMembersDiagnostic (OperationAnalysisContext operationContext, ImmutableArray<ISymbol> specialIncompatibleMembers, ISymbol member)
 		{
 			// Some RUC-annotated APIs are intrinsically handled by the trimmer
-			if (member is IMethodSymbol method && Intrinsics.GetIntrinsicIdForMethod (new MethodProxy (method)) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel) {
+			if (member is IMethodSymbol method && Intrinsics.GetIntrinsicIdForMethod (new MethodProxy (method)) != IntrinsicId.None) {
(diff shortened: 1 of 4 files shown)


src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -72,35 +73,16 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
...
-					&& parameterSymbol.Type?.ContainingNamespace?.Name == "System" && parameterSymbol.Type?.Name == "Type"));
-			}
-
-			return incompatibleMembers.ToImmutable ();
-		}
+		/// <summary>
+		/// Add methods that the Trimmer handles intrinsically to an exlusionary list
+		/// </summary>
+		/// <param name="compilation"></param>
+		/// <returns></returns>
```
sbomer, line 80: nit: left-over comment for the removed method

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -72,35 +73,16 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
...
 
 		protected override bool ReportSpecialIncompatibleMembersDiagnostic (OperationAnalysisContext operationContext, ImmutableArray<ISymbol> specialIncompatibleMembers, ISymbol member)
 		{
-			if (member is IMethodSymbol && ImmutableArrayOperations.Contains (specialIncompatibleMembers, member, SymbolEqualityComparer.Default)) {
-				// These RUC-annotated APIs are intrinsically handled by the trimmer, which will not produce any
-				// RUC warning related to them. For unrecognized reflection patterns related to generic type/method
-				// creation IL2055/IL2060 should be used instead.
+			// Some RUC-annotated APIs are intrinsically handled by the trimmer
+			if (member is IMethodSymbol method && Intrinsics.GetIntrinsicIdForMethod (new MethodProxy (method)) > IntrinsicId.RequiresReflectionBodyScanner_Sentinel) {
```
sbomer, line 85: Per our offline discussion with @tlakollo, instead of the sentinel I think we should check that it's not `IntrinsicId.None`. It probably makes no difference since intrinsics < the sentinel shouldn't have RUC (as far as I understand) but this seems more consistent and avoids introducing terminology from a linker-specific optimization into the analyzer code.

src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
LakshanF: Thanks for catching this!

tlakollo (approved):

PR #2599, opened 2022-02-11
Produce warnings for annotations on invalid types

tlakollo:
Produce warnings for annotations on invalid types using a symbol analyzer
Don't produce dataflow warnings for annotations on invalid types
Add extension file to ask if a type is interesting for dataflow

Fixes https://github.com/dotnet/linker/issues/2598

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
@@ -170,6 +170,9 @@ namespace ILLink.RoslynAnalyzer.DataFlow
 
 		public override TValue VisitPropertyReference (IPropertyReferenceOperation operation, LocalDataFlowState<TValue, TValueLattice> state)
 		{
+			if (!operation.Property.Type.IsTypeInterestingForDataflow ())
+				return TopValue;
+
 			if (operation.GetValueUsageInfo (operation.Property).HasFlag (ValueUsageInfo.Read)) {
 				// Accessing property for reading is really a call to the getter
 				// The setter case is handled in assignment operation since here we don't have access to the value to pass to the setter
--- a/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
@@ -170,9 +170,6 @@ namespace ILLink.RoslynAnalyzer.DataFlow
 
 		public override TValue VisitPropertyReference (IPropertyReferenceOperation operation, LocalDataFlowState<TValue, TValueLattice> state)
 		{
-			if (!operation.Property.Type.IsTypeInterestingForDataflow ())
-				return TopValue;
-
 			if (operation.GetValueUsageInfo (operation.Property).HasFlag (ValueUsageInfo.Read)) {
 				// Accessing property for reading is really a call to the getter
 				// The setter case is handled in assignment operation since here we don't have access to the value to pass to the setter
--- a/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
@@ -29,8 +29,10 @@ namespace ILLink.RoslynAnalyzer
 		{
 			var diagDescriptorsArrayBuilder = ImmutableArray.CreateBuilder<DiagnosticDescriptor> (26);
 			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.RequiresUnreferencedCode));
+			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersIsNotAllowedOnMethods));
 			AddRange (DiagnosticId.MethodParameterCannotBeStaticallyDetermined, DiagnosticId.DynamicallyAccessedMembersMismatchTypeArgumentTargetsGenericParameter);
 			AddRange (DiagnosticId.DynamicallyAccessedMembersOnFieldCanOnlyApplyToTypesOrStrings, DiagnosticId.DynamicallyAccessedMembersOnPropertyCanOnlyApplyToTypesOrStrings);
+			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersOnMethodReturnValueCanOnlyApplyToTypesOrStrings));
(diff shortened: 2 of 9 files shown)


tlakollo:
They are two tests failing, one of them at least to me seems like it should be a warning since the property type is object. But linker does dataflow analysis and ends up generating another warning. Second one seems to be related to an intrinsic, it believes is operating over a TypeHandle but we are using Type.GetTypeFromHandle() so its actually operating on a type. Will have to go deeper on those failures

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -150,5 +160,19 @@ static IEnumerable<Diagnostic> GetDynamicallyAccessedMembersDiagnostics (SingleV
 
 			return diagnosticContext.Diagnostics;
 		}
+
+		static void VerifyMemberOnlyApplyToTypesOrStrings (SymbolAnalysisContext context, ISymbol member)
+		{
+			if (member is IFieldSymbol field && field.GetDynamicallyAccessedMemberTypes () != DynamicallyAccessedMemberTypes.None && !field.Type.IsTypeInterestingForDataflow ())
```
sbomer, line 166: Maybe we should use `TryGetAttribute` here so that if you explicitly put `DAMT.None` on an unsupported type, it still produces a warning. What does the linker do in that case?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -150,5 +160,19 @@ static IEnumerable<Diagnostic> GetDynamicallyAccessedMembersDiagnostics (SingleV
...
+				context.ReportDiagnostic (Diagnostic.Create (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersOnFieldCanOnlyApplyToTypesOrStrings), member.Locations[0], member.GetDisplayName ()));
+			else if (member is IMethodSymbol method) {
+				foreach (var parameter in method.Parameters) {
+					if (parameter.GetDynamicallyAccessedMemberTypes () != DynamicallyAccessedMemberTypes.None && !parameter.Type.IsTypeInterestingForDataflow ())
+						context.ReportDiagnostic (Diagnostic.Create (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersOnMethodParameterCanOnlyApplyToTypesOrStrings), member.Locations[0], parameter.GetDisplayName (), member.GetDisplayName ()));
```
sbomer, line 171: We should also look at the method return type, and the method itself for annotated this parameters.

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -170,6 +170,9 @@ public static IMethodSymbol GetPropertyMethod (IPropertyReferenceOperation opera
 
 		public override TValue VisitPropertyReference (IPropertyReferenceOperation operation, LocalDataFlowState<TValue, TValueLattice> state)
 		{
+			if (!operation.Property.Type.IsTypeInterestingForDataflow ())
```
sbomer, line 173: We should avoid introducing DAM-specific logic into this file, which is a generic dataflow visitor. This check can be done in `TrimAnalysisVisitor` instead. For property getters, they go through `HandleMethodCall` so I think they will be handled if we add a similar check to `VisitInstanceReference` and `HandleMethodCall` (when creating the `MethodReturnValue`) instead.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
tlakollo: Linker does not warn on this case, should I proceed to warn even if this differs from linker's behavior?

sbomer: Then I would just follow the linker behavior, and maybe open an issue as a future improvement.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
tlakollo: I dont think this warning is prepared for warning on a method return type. The SharedString is:
`Parameter '{0}' of method '{1}' has 'DynamicallyAccessedMembersAttribute', but that attribute can only be applied to parameters of type 'System.Type' or 'System.String'.`

sbomer: Ah, it's a different warning code - I think IL2106 is the one for return types.

tlakollo: Added warnings for IL2106 and IL2041

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
tlakollo: Open https://github.com/dotnet/linker/issues/2603 to track this discussion

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -51,14 +51,17 @@ public override MultiValue VisitConversion (IConversionOperation operation, Stat
...
+			return paramRef.Parameter.Type.IsTypeInterestingForDataflow () ? new MethodParameterValue (paramRef.Parameter) : TopValue;
 		}
 
 		public override MultiValue VisitInstanceReference (IInstanceReferenceOperation instanceRef, StateValue state)
 		{
 			if (instanceRef.ReferenceKind != InstanceReferenceKind.ContainingTypeInstance)
 				return TopValue;
 
+			if (instanceRef.Type != null && !instanceRef.Type.IsTypeInterestingForDataflow ())
+				return TopValue;
```
vitek-karas, line 63: We should return TopValue if the instance ref doesn't have a type as well - the condition is technically the other way round "Return tracked value only if the type is interesting" - null types are by definition "not interesting".

src/ILLink.RoslynAnalyzer/ITypeSymbolExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+			IsSystemType = 0x01,
+			IsSystemReflectionIReflect = 0x02,
+		}
+
+		public static bool IsTypeInterestingForDataflow (this ITypeSymbol type)
+		{
+			if (type.SpecialType == SpecialType.System_String)
+				return true;
+
+			return IsSystemType (type) || IsSystemReflectionIReflect (type);
```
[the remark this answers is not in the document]
tlakollo: I didn't want to explore if a concurrent cache was needed, so for now I'm just refactoring the code to only call about flags once.

vitek-karas (approved):
Looks good.

PR #2600, opened 2022-02-11
[DAM analyzer] Produce unknown value for array access

sbomer:
And fix invocations of GetValueUsageInfo


vitek-karas:
@agocke - could you please review the part about `GetValueUsageInfo` - it has to do with Roslyn object model and I don't really understand the "why".

PR #2601, opened 2022-02-11
GetDisplayName uses definition's type parameter names instead of instance's

jtschuster:
Addresses https://github.com/dotnet/linker/issues/2573

Uses IMethodSymbol.OriginalDefinition to get the display name for reporting warnings in the analyzer

### DIFF
--- a/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlow.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlow.cs
@@ -766,9 +766,7 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 
 		// Warn about calls to the static methods and the ctor on the RUC type:
 		[ExpectedWarning ("IL2026", "RUCTypeRequiresPublicFields<T>.StaticMethod", "message")]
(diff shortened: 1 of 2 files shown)


test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlow.cs
```diff
@@ -766,9 +766,7 @@ static void TestNoWarningsInRUCMethod<T> ()
 
 		// Warn about calls to the static methods and the ctor on the RUC type:
 		[ExpectedWarning ("IL2026", "RUCTypeRequiresPublicFields<T>.StaticMethod", "message")]
-		[ExpectedWarning ("IL2026", "RUCTypeRequiresPublicFields<T>.StaticMethodRequiresPublicMethods<U>", "message", ProducedBy = ProducedBy.Trimmer)]
-		// https://github.com/dotnet/linker/issues/2573
-		[ExpectedWarning ("IL2026", "RUCTypeRequiresPublicFields<T>.StaticMethodRequiresPublicMethods<T>", "message", ProducedBy = ProducedBy.Analyzer)]
+		[ExpectedWarning ("IL2026", "RUCTypeRequiresPublicFields<T>.StaticMethodRequiresPublicMethods<U>", "message")]
```
sbomer, line 769: Here the generic type declaration happens to have the same name as the instantiation because both use `T` for the generic parameter name. Would you mind adding another test where the type is instantiated over something other than `T`?

sbomer (approved):
Thank you!

PR #2604, opened 2022-02-12
[DAM analyzer] Track null values

sbomer:
This tracks null values and adjusts the handling of null values in a
few intrinsics which either take structs that can never be null, or
don't return when the input is null.

MethodBodyScanner needed a small adjustment to correctly deal with the
case where an intrinsic returns TopValue.

Fixes https://github.com/dotnet/linker/issues/2528


vitek-karas (approved):
Thanks a lot for the fix.

PR #2610, opened 2022-02-15
Fix how test find the ref pack to compile test cases against

vitek-karas:
With latest VS we get a 6.0.2 runtime but only 6.0.1 ref pack. The tests so far expected to use the exact same version, but there's no hard rule that it will be the case.

Change the algorithm to search for highest available major.minor.* ref pack if the matching one doesn't exist.

### DIFF
--- a/test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompilationMetadataProvider.cs
+++ b/test/Mono.Linker.Tests/TestCasesRunner/TestCaseCompilationMetadataProvider.cs
@@ -111,7 +111,27 @@ namespace Mono.Linker.Tests.TestCasesRunner
 			string runtimeDir = Path.GetDirectoryName (typeof (object).Assembly.Location);
 			string ncaVersion = Path.GetFileName (runtimeDir);
 			var dotnetDir = Path.GetDirectoryName (Path.GetDirectoryName (Path.GetDirectoryName (runtimeDir)));
-			return Path.Combine (dotnetDir, "packs", "Microsoft.NETCore.App.Ref", ncaVersion, "ref", PathUtilities.TFMDirectoryName);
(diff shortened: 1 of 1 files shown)


vitek-karas:
Without this fix it's tedious to run tests on Windows. It requires using the `.\eng\dotnet.ps1 test ...` command to force the usage of the repo-local SDK/runtime (which is 6.0.0 so doesn't have this problem). Running tests from VS would require a custom environment to force VS to use the repo-local runtime as well.

vitek-karas:
I'm going to merge this to unblock local development. Please leave any feedback you might have and I'll fix it in a subsequent PR.

marek-safar:
This is surprising. Are you sure the ref pack can lag and is it only at the minor version?

vitek-karas:
I don't see a reason why they would need to be in sync. If the runtime ships a patch for functionality, but no impact on public API, then there's no reason to release a new ref pack. I coded it for patch version only because that feels very safe (the public API surface should barely change in patch releases, and it's good enough for our tests). Technically it could be made for both minor and patch, but then there's higher risk of discrepancies in the API surfaces.

tlakollo:
@jtschuster was having trouble with this, we though it was a bad install or similar. Thanks for fixing it! Changes look good to me :D

marek-safar:
> I don't see a reason why they would need to be in sync.

In theory yes but there is usually some ref pack change as well like https://github.com/dotnet/runtime/pull/62859

vitek-karas:
I'm trying to ask some other people about this - apparently it happened to me and Jakson through VS update, but it didn't happen to Andy who installed it manually - so there's something fishy.

I still think it's a good idea to keep this code in the linker as it makes the tests robust to these kind of changes - and I don't see anything wrong with it in the first place.

vitek-karas:
Thanks @MichalStrehovsky for the link - my problem is kind of the opposite (I'm missing the new ref pack, but have the old one), but I guess it's the same underlying problem.

PR #2616, opened 2022-02-15
Suppressing analyzer for compiler generated methods

### LINKED ISSUE #2587: [DAM analyzer] don't produce dataflow warnings for compiler-generated code
This issue tracks temporarily disabling dataflow warnings in the analyzer for compiler-generated code, until the linker better understands compiler-generated code.

Because the linker sees compiler-generated code for lambdas/iterators/async/local functions, it produces different dataflow warnings than those naturally produced by the analyzer. In particular, parameters and locals which are lifted to fields of state machines will show up as IL2077, whereas the linker will produce IL2066 for parameters.

Until the linker is fixed to produce the correct warnings for compiler-generated code, we would like the analyzer not to produce any warnings for these patterns. (This seems better than the alternative of producing slightly different warning codes for the same pattern, which would create usability issues - people would add annotations according to the analyzer warnings, but would get a completely different set when actually linking).

LakshanF:
Fixes #2587. This fix is temporary until we fix the linker.

- #2561 impacts the local functions and lambda data flow analysis
- The fix is sub-optimal in going through all the `IOperations `in a method to detect candidates for compiler generated code (other than async)
- Roslyn should expose `IsIterator `as a public API, which is a feature [request ](https://github.com/dotnet/roslyn/issues/20179)currently.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
@@ -16,6 +16,7 @@ using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.FlowAnalysis;
+using Microsoft.CodeAnalysis.Operations;
 
 namespace ILLink.RoslynAnalyzer
 {
@@ -64,6 +65,32 @@ namespace ILLink.RoslynAnalyzer
 				context.RegisterOperationBlockAction (context => {
 					if (context.OwningSymbol.IsInRequiresUnreferencedCodeAttributeScope ())
 						return;
+
+					// See https://github.com/dotnet/linker/issues/2587
+					// Need to punt on handling compiler generated methods until the linker is fixed
+					// async is handled here and the rest are handled just below
+					// iterators could be handled here once https://github.com/dotnet/roslyn/issues/20179 is fixed
+					if (context.OwningSymbol is IMethodSymbol methodSymbol && methodSymbol.IsAsync) {
+						return;
+					}
+
+					// Sub optimal way to handle analyzer not to generate warnings until the linker is fixed
+					// Iterators, local functions and lambdas are handled 
+					bool doNotHandleThisMethod = false;
+					foreach (IOperation blockOperation in context.OperationBlocks) {
+						if (blockOperation is IBlockOperation blocks) {
+							foreach (IOperation operation in blocks.Operations) {
+								if (operation.Kind == OperationKind.AnonymousFunction ||
+								operation.Kind == OperationKind.LocalFunction ||
+								operation.Kind == OperationKind.YieldBreak ||
+								operation.Kind == OperationKind.YieldReturn)
+									doNotHandleThisMethod = true;
+							}
+						}
+					}
+					if (doNotHandleThisMethod)
+						return;
+
 
 					foreach (var operationBlock in context.OperationBlocks) {
 						ControlFlowGraph cfg = context.GetControlFlowGraph (operationBlock);
--- a/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
@@ -76,7 +76,6 @@ namespace ILLink.RoslynAnalyzer
 
 					// Sub optimal way to handle analyzer not to generate warnings until the linker is fixed
 					// Iterators, local functions and lambdas are handled 
-					bool doNotHandleThisMethod = false;
 					foreach (IOperation blockOperation in context.OperationBlocks) {
 						if (blockOperation is IBlockOperation blocks) {
 							foreach (IOperation operation in blocks.Operations) {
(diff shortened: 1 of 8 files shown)


test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInCompilerGeneratedCode.cs
```diff
@@ -45,7 +45,8 @@ static IEnumerable<int> TestReflectionAccessRUCMethod ()
 			static IEnumerable<int> TestLdftnOnRUCMethod ()
 			{
 				yield return 0;
-				var _ = new Action (RequiresUnreferencedCodeMethod);
+				var a = new Action (RequiresUnreferencedCodeMethod);
```
tlakollo, line 48: Maybe we can create an additional test with the discard, analyzer handles it in a particular way. It would be interesting to have a test case with a discard for later

test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInCompilerGeneratedCode.cs
```diff
@@ -302,7 +304,7 @@ static void TestSuppressionOnLocalFunction ()
 				LocalFunction ();
 
 				[UnconditionalSuppressMessage ("Test", "IL2026")] // This supresses the RequiresUnreferencedCodeMethod
-				void LocalFunction (Type unknownType = null)
+				void LocalFunction (/**Type unknownType = null**/)
```
tlakollo, line 307: Nit: is this comment needed?

test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInCompilerGeneratedCode.cs
```diff
@@ -344,55 +348,62 @@ class SuppressInLambda
 		{
 			// Suppression currently doesn't propagate to local functions
 
-			[ExpectedWarning ("IL2026", CompilerGeneratedCode = true)]
+			[ExpectedWarning ("IL2026", CompilerGeneratedCode = true, ProducedBy = ProducedBy.Trimmer)]
```
tlakollo, line 351: I'm puzzled by this one, here the warning should be produced by the RequiresAnalyzerBase/RequiresUnreferencedCodeAnalyzer not having to do anything with the change in the dataflow analyzer. I'm not certain why this stopped warning. Seems more related to the discard operation change than the dataflow change.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -65,6 +66,32 @@ public override void Initialize (AnalysisContext context)
...
+					// Iterators, local functions and lambdas are handled 
+					bool doNotHandleThisMethod = false;
+					foreach (IOperation blockOperation in context.OperationBlocks) {
+						if (blockOperation is IBlockOperation blocks) {
+							foreach (IOperation operation in blocks.Operations) {
+								if (operation.Kind == OperationKind.AnonymousFunction ||
+								operation.Kind == OperationKind.LocalFunction ||
+								operation.Kind == OperationKind.YieldBreak ||
+								operation.Kind == OperationKind.YieldReturn)
+									doNotHandleThisMethod = true;
```
vitek-karas, line 87: Why not just `return`?

test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInCompilerGeneratedCode.cs
```diff
@@ -344,55 +348,62 @@ class SuppressInLambda
...
 			[UnconditionalSuppressMessage ("Test", "IL2026")]
 			static void TestCallRUCMethod ()
 			{
-				Action _ = () => RequiresUnreferencedCodeMethod ();
+				Action a = () => RequiresUnreferencedCodeMethod ();
+				a ();
```
vitek-karas, line 356: Why do we need this change?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
LakshanF: Oops

test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInCompilerGeneratedCode.cs
LakshanF: The analyzer doesn't produce a warning since there is an `UnconditionalSuppressMessage `attribute on that method

test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInCompilerGeneratedCode.cs
LakshanF: I used this in testing to see if there is a difference when the action is invoked (there wasn't) and forgot to remove it

test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInCompilerGeneratedCode.cs
LakshanF: I forgot to revert testing changes

test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInCompilerGeneratedCode.cs
tlakollo: Then the question reversed, dataflow will not do anything with the call AFAIK. The dataflow only produces IL2026 when using the binder and this is not the case, why it was warning before? Do we have a problem with the testing infra?

PR #2617, opened 2022-02-16
Comment out extra warnings coming from DAM Analyzer

jtschuster:
Comments out all warnings related to the DAM Analyzer that the analyzer produced but the linker does not. @LakshanF is working on disabling the analyzer in these scenarios and this should pass tests once merged with his changes.

I left alone places where the linker and analyzer produce the same warnings with different formatting, or where the analyzer warns on the associated symbol and the linker warns on the getter/setter.

- [x] TestFromTypeGetTypeOverConstant: this one is interesting Type.GetType() intrinsic should be disabled and therefore not sure why it warns
  - Resolved by https://github.com/dotnet/linker/pull/2627
- [x] AssemblyQualifiedNameDataflow/ComplexTypeHanlding 
  - Resolved by https://github.com/dotnet/linker/pull/2600 once is merged
- [x] ExceptionalDataFlow differences are not going to be addressed. The analyzer does it right and we will leave it like that till the linker catches up
  - By Design for now
- [x] Backward branches analysis differences are also not going to be addressed. The analyzer does it right and we will leave it like that
  - By Design for now
  - Related linker bug https://github.com/dotnet/linker/issues/2624
- [x] We might have to do some work into the backing field for properties. It has not been handled
  - Mostly By Design (forever) - updated comments in tests with https://github.com/dotnet/linker/pull/2630
  - One missing bit in analyzer - https://github.com/dotnet/linker/issues/2628
- [x] Not supporting ExpectedWarning in top-level statements in the analyzer infra is not going to be addressed for now, so it's ok to leave the expected warnings
  - By Design for now


tlakollo:
So from this list, I see:
- TestFromTypeGetTypeOverConstant: this one is interesting Type.GetType() intrinsic should be disabled and therefore not sure why it warns
- AssemblyQualifiedNameDataflow/ComplexTypeHanlding tests should be fixed with https://github.com/dotnet/linker/pull/2600 once is merged
- ExceptionalDataFlow differences are not going to be addressed. The analyzer does it right and we will leave it like that till the linker catches up
- Backward branches analysis differences are also not going to be addressed. The analyzer does it right and we will leave it like that
- We might have to do some work into the backing field for properties. It has not been handled
- Not supporting ExpectedWarning in top-level statements in the analyzer infra is not going to be addressed for now, so it's ok to leave the expected warnings

Actually, the things that Lakshan is fixing is not covered in the [ExpectedWarning] scenario, the testing file had suppressions and therefore the fix was to comment on the suppression
```
// [UnconditionalSuppressMessage ("Analyzer only sees the method parameter from the source file", "IL2067")]
[UnconditionalSuppressMessage ("Linker sees the compiler generated property", "IL2077")]
``` 
See: https://github.com/dotnet/linker/pull/2616

jtschuster:
I don't see anything unaddressed either. Closing this PR.

PR #2621, opened 2022-02-16
Update docs code for IL2078

jtschuster:
Docs for this error code looked wrong, but I might be misunderstanding.


jtschuster:
Should we even keep documentation for error codes or linker usage in this repo anymore, or just link to the [dotnet/docs](https://github.com/dotnet/docs) repo? I feel like we should have a single source of information or else they might diverge.

vitek-karas (approved):

vitek-karas:
If all of this information is already up on the docs - then I agree, we should only maintain that one.

PR #2623, opened 2022-02-16
Add check for pointer types in IsComInterop

jtschuster:
Fixes issue https://github.com/dotnet/linker/issues/2622. Pointer type `ISymbol`s return null for ContainingNamespace, so it would throw when it came across a pointer type. This adds a check in `IsCOMInterop` to check if the sybol is an `IPointerTypeSymbol` and returns true if it is.


test/ILLink.RoslynAnalyzer.Tests/ComAnalyzerTests.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    static void Main()
+    {
+        unsafe
+        {
+            WhateverName(null);
+        }
+    }
+}";
+			return VerifyComAnalyzer (TargetParameterWithAnnotations,
```
tlakollo, line 43: Im wondering if it would be better to have it as a shared test with linker. Maybe inside [test](https://github.com/dotnet/linker/blob/a372828948529c86f3ec02db9ddd9ce7ce1d8b06/test/Mono.Linker.Tests.Cases/Interop/PInvoke/Warnings/ComPInvokeWarning.cs)?

vitek-karas: Yes - pleas add this a shared test.

vitek-karas (approved):
I like it.

tlakollo (approved):

PR #2627, opened 2022-02-17
Update tests to match analyzer behavior around intrinsics

vitek-karas:
Analyzer intentionally "ignores" intrinsics which linker recognizes, but analyzer doesn't yet. Meaning it will not produce any warnings due to them.

Some tests didn't reflect this change correctly.


tlakollo (approved):
Thanks! I didn't realize these warnings were not even produced anymore. Definitely, allowMissingWarnings has some caveats.

jtschuster (approved):

PR #2629, opened 2022-02-17
Add test which accessed property backing field via reflection

vitek-karas:
Test for https://github.com/dotnet/linker/issues/2628

### DIFF
--- a/test/Mono.Linker.Tests.Cases/DataFlow/AnnotatedMembersAccessedViaReflection.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/AnnotatedMembersAccessedViaReflection.cs
@@ -563,6 +563,13 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 				typeof (AnnotatedProperty).RequiresAll ();
 			}
 
+			// Analyzer doesn't produce this warning https://github.com/dotnet/linker/issues/2628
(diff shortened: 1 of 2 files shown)


jtschuster (approved):
Just curious, why use `derivedInstance.GetType ()` instead of `typeof (DerivedWithAnnotation)` in TypeHierarchyReflectionWarnings.cs?

tlakollo:
> Just curious, why use `derivedInstance.GetType ()` instead of `typeof (DerivedWithAnnotation)` in TypeHierarchyReflectionWarnings.cs?

typeof() is an operator to obtain a type known at compile-time, the behavior is consistent from the standpoint of static analysis so the linker relies on the exact knowledge of the type. And we ignore the fact that the `type` can be annotated for this case.
Object.GetType() gets the exact runtime type of the current instance. The behavior can vary and therefore we have different ways to handle it. For example, if the type is sealed, Object.GetType() will behave like typeof() ignoring the annotations on the type if any. The difference is that in certain scenarios we would have to fulfill the annotation on the type.
It's not a simple scenario, but at least is somewhat documented in the [code](https://github.com/dotnet/linker/blob/main/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs#L967-L1027) and a little bit less covered in this [doc](https://github.com/dotnet/linker/blob/main/docs/design/DAM-on-type-and-RUC-interaction.md)

tlakollo (approved):

PR #2631, opened 2022-02-18
Fix merging null and non-null stack slots

### LINKED ISSUE #2624: Merging values on branch ends drops unknowns - underreports warnings
For example:

```csharp
		public static Type TestBackwardEdgeWithLdElem (Type[] types = null)
		{
			Type resultType = null; // returning null is fine, it fulfills any annotation
			foreach (var type in types) {  // this is effectively ldelem, which should produce "unknown" value
				resultType = type;   // assigns "unknown" to the result
			}

			return resultType; // IL2063: "unknown" doesn't have requires annotations
		}
```

Internally the "unknown" value loaded from the array is represented as `StackSlot` with `null` Value. The null literal is represented as `NullValue`. What happens inside the loop is that we're storing a second value for a local, so we merge the "unknown" and `NullValue`. But because the "unknown" is represented as `null`, the merge ignores it, and we end up with just the `NullValue`, which won't produce a warning.

MichalStrehovsky:
Make StackSlot.Value non-nullable. The fix falls out of that.

Fixes #2624.

### DIFF
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -15,14 +15,14 @@ namespace Mono.Linker.Dataflow
 	/// </summary>
 	readonly struct StackSlot
 	{
-		public ValueNode? Value { get; }
+		public ValueNode Value { get; }
 
 		/// <summary>
 		/// True if the value is on the stack as a byref
 		/// </summary>
 		public bool IsByRef { get; }
 
-		public StackSlot (ValueNode? value, bool isByRef = false)
+		public StackSlot (ValueNode value, bool isByRef = false)
 		{
 			Value = value;
 			IsByRef = isByRef;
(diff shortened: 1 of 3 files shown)


src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -49,14 +49,14 @@ private void CheckForInvalidStack (Stack<StackSlot> stack, int depthRequired, Me
...
-												   // Analysis of this method will be incorrect.
+					stack.Push (new StackSlot (UnknownValue.Instance)); // Push dummy values to avoid crashes.
+												                        // Analysis of this method will be incorrect.
 			}
 		}
 
 		private static void PushUnknown (Stack<StackSlot> stack)
 		{
-			stack.Push (new StackSlot ());
+			stack.Push (new StackSlot (UnknownValue.Instance));
```
MichalStrehovsky, line 59: One would think that the nullable analysis helped me to find these spots, but today I learned that:

```csharp
Foo f = new();
f.Obj.ToString();

struct Foo
{
    public object Obj { get; }
    public Foo(object obj) => Obj = obj;
}
```

will crash with a NullReferenceException, `#nullable enabled` be damned.

vitek-karas: Yeah - nullable on structs is... a lie :wink:. The default .ctor on struct is just evil - in many ways.

vitek-karas: We could declare a parameterless .ctor on the struct and make it throw. It doesn't catch all cases (the default initialization will not invoke it), but it would catch all the `new StackSlot()` calls.

vitek-karas (approved):
Thanks a lot @MichalStrehovsky

vitek-karas:
Actually we should probably close this. We're about to merge the damAnalyzer branch where this already works. As I was refactoring linker I "accidentaly" fixed this - added the default .ctor to StackValue and it initialized with UnknownValue.Instance. And I modified the .ctors to not allow null values (just like you do here).

https://github.com/dotnet/linker/blob/f678247e45c54b82045c64861967b77b18f6d684/src/linker/Linker.Dataflow/MethodBodyScanner.cs#L20-L46

MichalStrehovsky:
> Actually we should probably close this. We're about to merge the damAnalyzer branch where this already works

Makes sense! Learning more about nullable analysis was still a valuable exercise for me.

PR #2635, opened 2022-02-19
[release 6.0.2xx] Port Fix: Add check for pointer types in IsComInterop (#2623)

### LINKED ISSUE #2622: 6.0.200 Regression: COMAnalyzer chokes on PInvoke with pointer argument
My project uses a lot of CsWin32 automatically generated PInvokes.
Since updating from 6.0.102 to 6.0.200 I get a lot of 

```
CSC : warning AD0001: Analyzer 'ILLink.RoslynAnalyzer.COMAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
```

I distilled to the following minimal program that reproduces the warning:

```csharp
using System.Runtime.InteropServices;

static class Program
{
    // Since 6.0.200:
    // CSC : warning AD0001: Analyzer 'ILLink.RoslynAnalyzer.COMAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
    //
    // Only when the trim analyzer is enabled (e.g. with <PublishTrimmed>true</PublishTrimmed>)
    [DllImport("Unimportant")]
    static extern unsafe void WhateverName(void* pointerArg);

    static void Main()
    {
        unsafe
        {
            // Make sure the problematic function is used and not optimized away.
            // NOTE: This is not the NullReferenceException of the warning, that is thrown by the analyzer itself.
            WhateverName(null);
        }
    }
}
```

tlakollo:
Analyzing the ContainingNamespace for a pointer type returns null, operating with this null value causes a crash in the COMAnalyzer analyzer. 

## Customer Impact

An analyzer crash will generate a compiler warning. Depending on user settings this could make the build to fail if warn as error is enabled.

Link to customer issue: https://github.com/dotnet/linker/issues/2622

## Testing

Unit tests were added to this change.

## Risk

Very low risk - the change adds an extra check for pointer type before operating with the value, this makes the method return before impacting other code paths.

Fixes #2622

### DIFF
--- a/src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
+++ b/src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
@@ -81,6 +81,9 @@ namespace ILLink.RoslynAnalyzer
 				if (symbol is IParameterSymbol parameterSymbol)
 					typeSymbol = parameterSymbol.Type;
 
+				if (typeSymbol is IPointerTypeSymbol)
+					return false;
+
 				if (typeSymbol == null)
 					return false;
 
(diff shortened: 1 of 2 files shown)


jeffschwMSFT (approved):
Approved.  We should take for consideration in 6.0.2xx

agocke (approved):
LGTM

PR #2639, opened 2022-02-22
Share intrinsic handling of GetMember and similar APIs

vitek-karas:
Other than sharing more code and adapting it so that it works on both linker and analyzer, this change brings simple analysis of constant integer values in the analyzer. This is necessary to make most reflection API calls recognize binding flags. For example `GetMethods(BindingFlags.Public | BindingFlags.Static)`. So this change adds analysis of constant values (as recognized by the Roslyn's operation tree) and the OR binary operator for integers and enums.

Added some new tests for the binding flags handling.

Reenabled some disabled tests for analyzer.
Moved the main affected tests from the generated source files to the hardcoded one and force them to exact match of warnings for both linker and analyzer.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.RoslynAnalyzer/TrimAnalysis/HandleCallAction.cs
@@ -1,8 +1,11 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
 using ILLink.RoslynAnalyzer;
+using ILLink.RoslynAnalyzer.DataFlow;
 using ILLink.RoslynAnalyzer.TrimAnalysis;
 using ILLink.Shared.TypeSystemProxy;
 using Microsoft.CodeAnalysis;
@@ -16,13 +19,15 @@ namespace ILLink.Shared.TrimAnalysis
 
 		readonly ISymbol _owningSymbol;
 		readonly IOperation _operation;
+		readonly ReflectionAccessAnalyzer _reflectionAccessAnalyzer;
 
 		public HandleCallAction (in DiagnosticContext diagnosticContext, ISymbol owningSymbol, IOperation operation)
 		{
 			_owningSymbol = owningSymbol;
 			_operation = operation;
 			_diagnosticContext = diagnosticContext;
-			_requireDynamicallyAccessedMembersAction = new (diagnosticContext, new ReflectionAccessAnalyzer ());
+			_reflectionAccessAnalyzer = new ReflectionAccessAnalyzer ();
+			_requireDynamicallyAccessedMembersAction = new (diagnosticContext, _reflectionAccessAnalyzer);
 		}
 
 		// TODO: This is relatively expensive on the analyzer since it doesn't cache the annotation information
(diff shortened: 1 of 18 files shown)


src/ILLink.RoslynAnalyzer/TrimAnalysis/HandleCallAction.cs
```diff
@@ -41,9 +46,39 @@ public HandleCallAction (in DiagnosticContext diagnosticContext, ISymbol owningS
 		private partial MethodThisParameterValue GetMethodThisParameterValue (MethodProxy method, DynamicallyAccessedMemberTypes dynamicallyAccessedMemberTypes)
 			=> new (method.Method, dynamicallyAccessedMemberTypes);
 
+		private partial MethodParameterValue GetMethodParameterValue (MethodProxy method, int parameterIndex, DynamicallyAccessedMemberTypes dynamicallyAccessedMemberTypes)
+			=> new (method.Method.Parameters[parameterIndex], dynamicallyAccessedMemberTypes);
+
+		private partial IEnumerable<SystemReflectionMethodBaseValue> GetMethodsOnTypeHierarchy (TypeProxy type, string name, BindingFlags? bindingFlags)
+		{
+			foreach (var method in type.Type.GetMethodsOnTypeHierarchy (m => m.Name == name, bindingFlags))
+				yield return new SystemReflectionMethodBaseValue (new MethodProxy (method));
```
agocke, line 55: I think this is going to cause a quadratic number of calls in the depth of the tree. Since iterators are lazy, we'll walk all the way down the tree to fetch the first element, then walk n-1 levels to fetch the second, and so on.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -88,6 +104,35 @@ public override MultiValue VisitLiteral (ILiteralOperation literalOperation, Sta
 			return literalOperation.ConstantValue.Value == null ? NullValue.Instance : TopValue;
 		}
 
+		public override MultiValue VisitBinaryOperator (IBinaryOperation operation, StateValue argument)
```
agocke, line 107: I tried this locally and it looks like `operation` has a constant value for many of the tests. Is there a specific case I'm missing? And is it worth checking the ConstantValue anyway for the easy cases?

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -88,6 +104,35 @@ public override MultiValue VisitLiteral (ILiteralOperation literalOperation, Sta
 			return literalOperation.ConstantValue.Value == null ? NullValue.Instance : TopValue;
```
agocke, line 104: Why don't we want to use a ConstIntValue for cases where the literal is an integer?

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
vitek-karas: You're right - I was looking at the `ArgumentOperation` which is the parent of this and that one doesn't have a constant value. I'll improve this to use the Roslyns const evaluator when available.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
vitek-karas: I guess I just didn't get that far. When I do the constant checks, this should sort of take care of itself, but I'll add it here as well.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
agocke: You could maybe put it in the top-level `Visit(IOperation)` helper, unless there's some reason we would want to skip that.

vitek-karas: I'm kind of split on that - it's semi-dangerous. It would mean that the actual specific operation visitor method would not have final say in what the return value is. Maybe not a big deal... I'll probably do it.

src/ILLink.RoslynAnalyzer/TrimAnalysis/HandleCallAction.cs
vitek-karas: Sorry I don't understand. The GetMethodsOnTypeHierarchy helper we're calling here already uses iterator, so this is basically just a pass through...

I guess I could rewrite both to use ImmutableArray...

I must admit I don't understand what the problem is.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
vitek-karas: I went with that approach - Visit(IOperation) now handles constants (if the operation specific visit doesn't provide a value). It solves lot of easy cases and works generally for things like null and strings as well.
I kept the BinaryOperation visitor as there are cases where it still works (if one or both of the operands are local variables for example - that requires data flow to figure out values, so constant detection itself doesn't do it) and also as a sample what we would need to add for other operators if necessary.

agocke (approved):

sbomer (approved):
LGTM, thank you!

PR #2641, opened 2022-02-22
Enable latest analysis

marek-safar:


test/ILLink.RoslynAnalyzer.Tests.Generator/TestCaseGenerator.cs
```diff
@@ -178,7 +178,7 @@ public void Initialize (GeneratorInitializationContext context)
 
 	class ExistingTestCaseDiscoverer : ISyntaxContextReceiver
 	{
-		public readonly TestCases ExistingTestCases = new ();
+		public readonly TestCases ExistingTestCases;
```
vitek-karas, line 181: This probably a known issues between the analyzers - the fix is to make it into a property, so something like:
```C#
public TestCases ExistingTestCases { get; } = new ();
```

marek-safar: `TestCases` is a struct so `new ()` has no effect or am I missing something?

marek-safar: I put pragma there instead

vitek-karas (approved):

PR #2643, opened 2022-02-22
[release/6.0.2xx] Fix invalid cast of type parameter

sbomer:
Brings the fix @agocke made as part of https://github.com/dotnet/linker/pull/2540 to 6.0.2. Fixes https://github.com/dotnet/linker/issues/2642.

## Customer Impact

Updating VS from 17 to 17.1 brings in a 6.0.2 SDK which causes the analyzer to crash on a project that previously worked, producing a warning. The warning also shows up in CI when using the 6.0.2 SDK, and fails the build with warnaserror. See details in https://github.com/dotnet/linker/issues/2642.

## Testing

Unit tests confirmed the issue and the fix. This was originally discovered as part of running the analyzer over itself and the linker in https://github.com/dotnet/linker/pull/2523.

## Risk

Very low risk - this turns a cast to a type check, and skips the case which used to throw an InvalidCastException.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -172,10 +172,10 @@ namespace ILLink.RoslynAnalyzer
 					for (int i = 0; i < typeParams.Length; i++) {
 						var typeParam = typeParams[i];
 						var typeArg = typeArgs[i];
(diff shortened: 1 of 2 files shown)


vitek-karas (approved):

jeffschwMSFT (approved):
Approved.  We should take for consideration in 6.0.2xx

agocke (approved):

PR #2644, opened 2022-02-23
Remove error-codes.md and add script to update dotnet/docs

jtschuster:
Removes error-codes.md and adds ILDocs.ps1 to automate updating docs when warning messages change. ILDocs.ps1 uses the added messages in SharedStrings.resx which have replaced the "{X}" with what the inserted string represents ('field' or 'method', etc).


marek-safar:
really, why? Any documentation that is generated is no documentation.

vitek-karas:
Would it make sense to use this tool mainly as a way to check that there are no differences between docs and the linker repo? And it would be up to the person fixing those differences to either use it or not to make actual changes.

The whys for this:
* `main` branch is not always in a "Shipping" state - it should be, but it can't be relied on (at least yet) - running this would pollute the docs with things which may change in the near future.
* As @marek-safar pointed out, it doesn't actually provide any documentation. While I agree with the search engine argument, I think forcing manual work around this process should act as a forcing function to actually write real docs about the error code.

I could imagine we make this tool part of CI in some kind of "warning" mode (not fail the CI, but warn about missing docs) - but then it would probably require ability to baseline, which is more work... so probably not worth it.

jtschuster:
@vitek-karas 
> Would it make sense to use this tool mainly as a way to check that there are no differences between docs and the linker repo? And it would be up to the person fixing those differences to either use it or not to make actual changes.

That's what I was thinking, I see this script only being used manually to check if there are differences and only update the docs when there is a change that gets pushed to the SDK. I don't think this would make much sense in to use in CI. 

If we don't think it's worth it to have this and prefer always making manual changes to docs, I'll remove it from the PR.

vitek-karas:
I perfectly OK keeping the tool in the repo ... I just wanted to make sure we're not trying to shift the responsibility of updating docs to a tool :wink:.

marek-safar:
@jtschuster the script seems useful, do you want to port it to runtime repo?

PR #2645, opened 2022-02-23
Remove fixing-warnings.md since it is duplicated in dotnet/docs

jtschuster:


vitek-karas (approved):
I think it might be a good idea to add links to the official docs to the readme.md in the docs directory - so that if somebody find their way to the repo, they can find the official docs easily.

PR #2646, opened 2022-02-23
Fix generated test sources

sbomer:
And actually exclude them from the compilation.


vitek-karas (approved):

sbomer:
One of the linker tests failed like this:
```
Failed TestGetDisplayName(MethodWithPartiallyInstantiatedNestedGenericTypeArguments,DisplayNameAttribute) [60 ms]
  Error Message:
     Expected string length 232 but was 218. Strings differ at index 164.
  Expected: "...enericClassMultipleParameters<MethodT,String>.NestedGeneri..."
  But was:  "...enericClassMultipleParameters<Int32,MethodV>.NestedGeneric..."
```

This is the second time I've seen this in CI, maybe there's a race condition somewhere in that test.

PR #2649, opened 2022-02-24
nameof calls should not warn on analyzer

### LINKED ISSUE #2647: RUC analyzer seems to react to nameof references
```C#
[RUC("Don't")]
class TestMe
{
    public static int StaticField;
}

void TestMethod()
{
    Console.WriteLine(nameof(TestMe.StaticField));  // IL2026 Don't
}
```

Since linker doesn't produce these warnings, the analyzer probably shouldn't either.

tlakollo:
Add a condition to bypass `nameof` usage since it will not invoke the member and instead will be translated to load a string
Add tests
Fix ExpectedWarnings from disabled intrinsics that were producing warnings from accessing fields in a `nameof` call

Fixes #2647


vitek-karas (approved):
Thanks a lot for the fix.

vitek-karas:
I'm going to merge this as it unblocks some of the work on the intrinsics (or makes it easier)

PR #2650, opened 2022-02-24
Rename AOT property to match msbuild camel case

tlakollo:


sbomer (commented):
What's the reasoning behind the change? We seem to use NativeAOT elsewhere. I personally think EnableAOTAnalyzer reads better and I'd prefer to keep the acronym capitalized.

tlakollo:
I think we should follow a convention if there is any. Although they are examples of using the properties in both ways saw a couple of ARM properties. I think the inclination is to use the camel case in msbuild. They are several examples when using things like ASP.NET, HTTP, SDK, URL, etc.
https://github.com/dotnet/sdk/blob/main/eng/Versions.props#L136-L144
https://github.com/dotnet/sdk/blob/main/src/WebSdk/Package.props#L27-L28

agocke:
It does seem like the MSBuild convention is CamelCase. I'd recommend sticking with MSBuild convention unless we have some other convention to follow

sbomer (approved):
Thanks for the rationale, makes sense!

PR #2652, opened 2022-02-24
Fix behavior of intrinsics with empty inputs

sbomer:
The latest analyzer is producing warnings in runtime for this pattern:
https://github.com/dotnet/runtime/blob/959a13fa0d1822dc4c7bfc99a9d5e7a0c50d6306/src/libraries/System.Data.Common/src/System/Data/Common/DbProviderFactories.cs#L129

This fixes that by ensuring that `AssemblyQualifiedName` and other similar intrinsics pass through a `TopValue` input instead of falling back on the annotations when there is no input. (The analyzer intentionally returns `TopValue` for unimplemented intrinsics, and the idea is that these should not produce any warnings.)

### DIFF
--- a/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
@@ -72,6 +72,11 @@ namespace ILLink.Shared.TrimAnalysis
 				break;
 
 			case IntrinsicId.Type_get_TypeHandle:
+				if (instanceValue.IsEmpty ()) {
+					returnValue = MultiValueLattice.Top;
+					break;
+				}
+
 				foreach (var value in instanceValue) {
 					if (value is SystemTypeValue typeValue)
 						AddReturnValue (new RuntimeTypeHandleValue (typeValue.RepresentedType));
@@ -90,6 +95,11 @@ namespace ILLink.Shared.TrimAnalysis
 			// GetInterface (String, bool)
 			//
 			case IntrinsicId.Type_GetInterface: {
+					if (instanceValue.IsEmpty ()) {
+						returnValue = MultiValueLattice.Top;
+						break;
+					}
+
 					var targetValue = GetMethodThisParameterValue (calledMethod, DynamicallyAccessedMemberTypesOverlay.Interfaces);
 					foreach (var value in instanceValue) {
 						// For now no support for marking a single interface by name. We would have to correctly support
@@ -115,6 +125,11 @@ namespace ILLink.Shared.TrimAnalysis
 			// AssemblyQualifiedName
 			//
 			case IntrinsicId.Type_get_AssemblyQualifiedName: {
+					if (instanceValue.IsEmpty ()) {
+						returnValue = MultiValueLattice.Top;
+						break;
+					}
+
 					foreach (var value in instanceValue) {
 						if (value is ValueWithDynamicallyAccessedMembers valueWithDynamicallyAccessedMembers) {
 							// Currently we don't need to track the difference between Type and String annotated values
@@ -159,9 +174,10 @@ namespace ILLink.Shared.TrimAnalysis
 				return true;
 			}
 
-			// For some intrinsics we just use the return value from the annotations.
-			if (returnValue == null)
-				returnValue = calledMethod.ReturnsVoid () ? MultiValueLattice.Top : GetMethodReturnValue (calledMethod, returnValueDynamicallyAccessedMemberTypes);
+			if (returnValue == null && !calledMethod.ReturnsVoid ())
+				throw new InvalidOperationException ("No return value set for intrinsic");
+
+			returnValue ??= MultiValueLattice.Top;
 
 			// Validate that the return value has the correct annotations as per the method return value annotations
 			if (returnValueDynamicallyAccessedMemberTypes != 0) {
--- a/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
@@ -183,9 +183,7 @@ namespace ILLink.Shared.TrimAnalysis
 				return true;
 			}
 
-			if (returnValue == null && !calledMethod.ReturnsVoid ())
-				throw new InvalidOperationException ("No return value set for intrinsic");
-
+			Debug.Assert (returnValue != null || calledMethod.ReturnsVoid (), "Intrinsics must set a return value.");
 			returnValue ??= MultiValueLattice.Top;
 
 			// Validate that the return value has the correct annotations as per the method return value annotations
--- a/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
@@ -95,28 +95,37 @@ namespace ILLink.Shared.TrimAnalysis
 			// GetInterface (String, bool)
 			//
 			case IntrinsicId.Type_GetInterface: {
-					if (instanceValue.IsEmpty ()) {
+					if (instanceValue.IsEmpty () || argumentValues[0].IsEmpty ()) {
 						returnValue = MultiValueLattice.Top;
 						break;
 					}
 
 					var targetValue = GetMethodThisParameterValue (calledMethod, DynamicallyAccessedMemberTypesOverlay.Interfaces);
 					foreach (var value in instanceValue) {
-						// For now no support for marking a single interface by name. We would have to correctly support
-						// mangled names for generics to do that correctly. Simply mark all interfaces on the type for now.
+						foreach (var interfaceName in argumentValues[0]) {
+							if (interfaceName == NullValue.Instance) {
+								// Throws on null string, so no return value.
+								returnValue ??= MultiValueLattice.Top;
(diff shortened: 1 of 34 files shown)


sbomer:
Thanks for the suggestion - I added tests for null/no-value flowing into intrinsics where it made sense and it caught a few more bugs, including a crash in the linker. :) I left in some warnings that we're currently producing in the linker, so please let me know if you think any of these should be changed:
- Type.GetType: https://github.com/dotnet/linker/pull/2652/files#diff-59053285ad8572c6eb9752b25058bbaa6f0302b8a57d185c5c8dba14f2b533e6R74-R89
- Object.GetType: https://github.com/dotnet/linker/pull/2652/files#diff-255083891e7d372b532e9304e8f77091939a71fe80aba241ed4fef2ff7f5c1c9R1473-R1500
- Activator.CreateInstanceFrom: https://github.com/dotnet/linker/pull/2652/files#diff-2b6f8f8ac5ca78ddecaeff24b91204739a2c5a3c659706ba02319e73a0a4e911R421-R427

I also didn't add tests for AppDomain.CreateInstance* which currently only has basic testing: https://github.com/dotnet/linker/blob/45b481f3504fa3c7142ed22f37d2813b0c7df43f/test/Mono.Linker.Tests.Cases/Reflection/ActivatorCreateInstance.cs#L383-L384

sbomer:
> I wonder if there's a way to somehow make this a bit more general solution.

Not really a general solution, but I did at least add an exception here to force all of the shared intrinsics to set a return value (when they have return types): https://github.com/dotnet/linker/pull/2652/files#diff-4d295034befb7f16882c6686e3bc9dcc12ce427af6af6cef53ebc0ca3a7562baR186-R187

test/Mono.Linker.Tests.Cases/Reflection/ObjectGetType.cs
```diff
@@ -1458,5 +1461,44 @@ public static void Test ()
...
+			[Kept]
+			static TestType GetInstance () => null;
+
+			[Kept]
+			[ExpectedWarning ("IL2072", nameof (DataFlowTypeExtensions.RequiresAll) + "(Type)", nameof (Object.GetType) + "()")]
+			public static void Test ()
+			{
+				TestType noValue = GetInstance ();
+				// Even though this throws at runtime, we warn about the return value of GetType
+				noValue.GetType ().RequiresAll ();
```
vitek-karas, line 1500: This one I think is correct - we don't really track values across method calls (I mean in theory this could be handled through const propagation, but that's a different problem). So the fact that we warn here is a good thing I think.

test/Mono.Linker.Tests.Cases/Reflection/ActivatorCreateInstance.cs
```diff
@@ -366,6 +408,56 @@ public WithAssemblyPathParameterless (int i, object o)
...
+			WithAssemblyPathAndNoValueTypeName ();
+		}
+
+		[Kept]
+		// Technically this shouldn't warn because CreateInstanceFrom throws if the assembly file path is null.
+		// However, our implementation is the same for CreateInstance and CreateInstanceFrom.
+		[ExpectedWarning ("IL2032", nameof (Activator) + "." + nameof (Activator.CreateInstance), "assemblyFile")]
+		private static void WithNullAssemblyPath ()
+		{
+			Activator.CreateInstanceFrom (null, "Mono.Linker.Tests.Cases.Reflection.ActivatorCreateInstance+WithAssemblyPathParameterless");
```
vitek-karas, line 426: Not ideal, but I don't think this is that important (honestly I've never seen a call to this method in real app, we have these intrinsics for historical reasons mainly).

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -159,9 +183,10 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
 				return true;
 			}
 
-			// For some intrinsics we just use the return value from the annotations.
-			if (returnValue == null)
-				returnValue = calledMethod.ReturnsVoid () ? MultiValueLattice.Top : GetMethodReturnValue (calledMethod, returnValueDynamicallyAccessedMemberTypes);
+			if (returnValue == null && !calledMethod.ReturnsVoid ())
+				throw new InvalidOperationException ("No return value set for intrinsic");
```
vitek-karas, line 187: Given how many times we've been struggling with this (and thus the implied lack of test coverage) - I'm a bit nervous making it throw here (if linker throws, there's basically no workaround for trimming the app). Maybe we should assert and fallback to returning the annotated value - that should in theory lead to warnings which people report to us as noise (and we still catch the most glaring holes in our tests if we hit them).

On a related note (as this has been a recurring topic recently) - maybe we should use Debug builds of linker in our repos (runtime, MAUI, ...) when they run their Debug CI legs - that way we get better test coverage of the linker without the need to break app trimming with exceptions (we could basically use asserts more frequently without losing that much data).

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -1310,7 +1310,7 @@ bool AnalyzeGenericInstantiationTypeArray (in AnalysisContext analysisContext, i
 
 				bool allIndicesKnown = true;
 				for (int i = 0; i < size.Value; i++) {
-					if (!array.TryGetValueByIndex (i, out MultiValue value) || value.IsEmpty () || value.AsSingleValue () is UnknownValue) {
+					if (!array.TryGetValueByIndex (i, out MultiValue value) || value.AsSingleValue () is UnknownValue) {
```
vitek-karas, line 1313: good catch!

test/Mono.Linker.Tests.Cases/DataFlow/GetTypeDataFlow.cs
```diff
@@ -69,6 +71,22 @@ static void TestConstructors ()
 			type.RequiresNone ();
 		}
 
+		[ExpectedWarning ("IL2072", nameof (DataFlowTypeExtensions.RequiresAll) + "(Type)", nameof (Type.GetType) + "(String)")]
+		static void TestNull ()
+		{
+			// Warns about the return value of GetType, even though this throws at runtime.
+			Type.GetType (null).RequiresAll ();
```
vitek-karas, line 78: Can you please file an issue about this? I don't think it's super important to fix, but we should track it as it's inconsistent right now, some intrinsics handle this without warnings, and some don't.

test/Mono.Linker.Tests.Cases/DataFlow/TypeBaseTypeDataFlow.cs
```diff
@@ -211,6 +212,15 @@ static void TestNull ()
 			type.BaseType.RequiresPublicMethods ();
 		}
 
+		[ExpectedWarning ("IL2072", nameof (DataFlowTypeExtensions) + "." + nameof (DataFlowTypeExtensions.RequiresPublicMethods))]
+		static void TestNoValue ()
+		{
+			Type t = null;
+			Type noValue = Type.GetTypeFromHandle (t.TypeHandle);
+			// Warns about the base type even though the above throws an exception at runtime.
+			noValue.BaseType.RequiresPublicMethods ();
```
vitek-karas, line 221: +1 on filing an issue for this.

test/Mono.Linker.Tests.Cases/Reflection/ActivatorCreateInstance.cs
vitek-karas: We could add this to the issue - but the priority of this one is even lower.

test/Mono.Linker.Tests.Cases/Reflection/ObjectGetType.cs
```diff
@@ -1458,5 +1461,44 @@ public static void Test ()
...
+			{
+			}
+
+			[Kept]
+			[ExpectedWarning ("IL2072", nameof (DataFlowTypeExtensions.RequiresAll) + "(Type)", nameof (Object.GetType) + "()")]
+			public static void Test ()
+			{
+				TestType nullInstance = null;
+				// Even though this throws at runtime, we warn about the return value of GetType
+				nullInstance.GetType ().RequiresAll ();
```
vitek-karas, line 1479: +1 on the issue

vitek-karas (approved):
Thanks a lot for all of the new tests!

test/Mono.Linker.Tests.Cases/Reflection/ObjectGetType.cs
sbomer: Thanks for catching this, the test wasn't exactly what I wanted. I will change it to use the same pattern for getting an empty value (of type Type) as in the other tests.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
sbomer: I left this as before, because some of the new intrinsics rely on the fall-back. Even though they don't return types, they currently get are tracked as ValueWithDynamicallyAccessedMembers, and then cause warnings in some cases, like MakeGenericMethod. I think these should be changed to track UnknownValue.

PR #2656, opened 2022-02-25
Warn on DAM mismatch between overrides

jtschuster:
Fixes https://github.com/dotnet/linker/issues/2512

Adds checks to ensure that method overrides have the same DAM annotations as the virtual methods / interfaces. Warning on the implicit 'this' meant I had to add annotations to the test class that inherited from `Type` and update the expected warnings related to that class.

I found that the when there is a mismatch between DAM on a property and on the accessor, the linker will warn on the accessor, but use that DAM annotation instead of the property's. I think it would make more sense to warn on the property, so the analyzer warns there, and I filed https://github.com/dotnet/linker/issues/2654 to track the issue with the linker.

Also created a helper to find all the interface members that a type implements, which also takes into account any interfaces an interface recursively implements, and updated the RequiresAnalyzer to use that helper.

Finally, updated the checks for `LogContains` and `LogDoesNotContain` in the analyzer tests to support Regex matches and deduplicated some code there.


vitek-karas:
> I think it would make more sense to warn on the property, so the analyzer warns there, and I filed https://github.com/dotnet/linker/issues/2654 to track the issue with the linker.

I agree that it would make more sense, the question is if it's worth the change. If there's an existing suppression, fixing this will break that suppression. I'm not sure the change is important enough to risk a break like that.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -208,5 +227,130 @@ static void VerifyMemberOnlyApplyToTypesOrStrings (SymbolAnalysisContext context
...
+					context.ReportDiagnostic (Diagnostic.Create (
+						DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersMismatchOnGenericParameterBetweenOverrides),
+						methodTypeParam.Locations[0],
+						methodTypeParam.GetDisplayName (), method.GetDisplayName (), overriddenTypeParam.GetDisplayName (), overriddenMethod.GetDisplayName ()));
+				}
+			}
+
+			var basestTypeSymbol = context.Symbol.ContainingType;
+			while (basestTypeSymbol?.BaseType != null) {
+				if (basestTypeSymbol.Name == "Type" && basestTypeSymbol.ContainingNamespace.Name == "System") {
```
vitek-karas, line 281: linker doesn't do the check for `System.Type` in this case, it simply compares the annotations on the two methods.
The analyzer would produce warning `DynamicallyAccessedMembersIsNotAllowedOnMethods` if the annotation is applied to a method which is on non-interesting type.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -208,5 +227,130 @@ static void VerifyMemberOnlyApplyToTypesOrStrings (SymbolAnalysisContext context
...
+				var overriddenTypeParam = overriddenMethod.TypeParameters[i];
+				var methodParamHasDam = methodTypeParam.TryGetAttribute (DynamicallyAccessedMembersAttribute, out var methodTypeParamAttribute);
+				var overriddenParamHasDam = overriddenTypeParam.TryGetAttribute (DynamicallyAccessedMembersAttribute, out var overriddenTypeParamAttribute);
+				if ((methodParamHasDam && overriddenParamHasDam && !DamAttributesMatch (methodTypeParamAttribute!, overriddenTypeParamAttribute!))
+					|| (methodParamHasDam ^ overriddenParamHasDam)) {
+					context.ReportDiagnostic (Diagnostic.Create (
+						DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersMismatchOnGenericParameterBetweenOverrides),
+						methodTypeParam.Locations[0],
+						methodTypeParam.GetDisplayName (), method.GetDisplayName (), overriddenTypeParam.GetDisplayName (), overriddenMethod.GetDisplayName ()));
+				}
```
vitek-karas, line 276: Would it be possible to refactor this to share the code between the four places? The condition and warnings are nearly identical.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -208,5 +227,130 @@ static void VerifyMemberOnlyApplyToTypesOrStrings (SymbolAnalysisContext context
...
+					&& interfaceMember is IMethodSymbol interfaceMethod)
+					VerifyMethodDamAttributesMatch (implementationMethod, interfaceMethod, context);
+			}
+		}
+
+		static void VerifyDamOnPropertyAndAccessorMatch (IMethodSymbol methodSymbol, SymbolAnalysisContext context)
+		{
+			if (!(methodSymbol.AssociatedSymbol?.TryGetAttribute (DynamicallyAccessedMembersAttribute, out var propertyAttribute) == true)
+				|| (propertyAttribute.ConstructorArguments.Length == 1 && propertyAttribute.ConstructorArguments[0].Value is (int)DynamicallyAccessedMemberTypes.None))
+				return;
```
vitek-karas, line 309: This will probably run for event accessors as well (`add`, `remove`, ...) - probably not a problem, but maybe to keep the analyzer more strict, we should check that the method in question is getter/setter first.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -208,5 +227,130 @@ static void VerifyMemberOnlyApplyToTypesOrStrings (SymbolAnalysisContext context
...
+
+			if (methodSymbol.MethodKind == MethodKind.PropertyGet
+				&& methodSymbol.TryGetReturnTypeAttribute (DynamicallyAccessedMembersAttribute, out var getValueAttribute)
+				&& !(getValueAttribute.ConstructorArguments.Length == 1 && getValueAttribute.ConstructorArguments[0].Value is (int)DynamicallyAccessedMemberTypes.None)) {
```
vitek-karas, line 313: Why not use the `GetDynamicallyAcessedMemberTypes` extension method which does both the "get attribute" and the extraction of the value (in this case it would be `GetDynamicallyAccessedMemberTypesOnReturnValue`).

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -208,5 +227,130 @@ static void VerifyMemberOnlyApplyToTypesOrStrings (SymbolAnalysisContext context
...
+
+		// DAM flows from property to ".get" return value
+		static bool TryGetReturnDamAttribute (IMethodSymbol methodSymbol, [NotNullWhen (true)] out AttributeData? attribute)
+		{
+			// Accessor is used when both are present. Warn on conflicts with VerifyDamOnPropertyAndAccessorMatch
+			return methodSymbol.TryGetReturnTypeAttribute (DynamicallyAccessedMembersAttribute, out attribute)
+				? true
+				: methodSymbol.MethodKind == MethodKind.PropertyGet
+					&& methodSymbol.AssociatedSymbol!.TryGetAttribute (DynamicallyAccessedMembersAttribute, out attribute);
+		}
```
vitek-karas, line 344: This should already be handled by `FlowAnnotations.GetMethodReturnValueAnnotation` - and the one for parameters is done in `FlowAnnotations.GetMethodReturnValueAnnotation`. They also extract the annotation value from the attribute and so on.

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -514,6 +514,7 @@ static string GetSystemTypeBase ()
 
 using System.Globalization;
 using System.Reflection;
+using System.Diagnostics.CodeAnalysis;
```
vitek-karas, line 517: This should not be needed...

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -537,6 +538,8 @@ public class TestSystemTypeBase : Type
 
 		public override string Name => throw new NotImplementedException ();
 
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors
+			| DynamicallyAccessedMemberTypes.NonPublicConstructors)]
```
vitek-karas, line 542: Indentation is off - you added one with spaces, while the rest of the code here uses tabs (even in the string literals).

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
```diff
@@ -271,43 +272,68 @@ bool Matches (Diagnostic diagnostic)
...
 		private bool TryValidateLogContainsAttribute (AttributeSyntax attribute, List<Diagnostic> diagnostics, out int? matchIndex, out string? missingDiagnosticMessage)
 		{
-			missingDiagnosticMessage = null;
-			matchIndex = null;
+			if (!LogContains(attribute, diagnostics, out matchIndex, out string text)) {
+				missingDiagnosticMessage = $"Could not find text:\n{text}\nIn diagnostics:\n{string.Join (Environment.NewLine, _diagnostics)}";
+				return false;
+			}
+			else {
+				missingDiagnosticMessage = null;
```
vitek-karas, line 280: Unnecessary - it's already assigned null

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
jtschuster: I tried to, and for me the slight differences in each case are just enough so that anything I tried ended up adding complexity without reducing repetition.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
jtschuster: I didn't know that existed, thank you!

test/ILLink.RoslynAnalyzer.Tests/TestChecker.cs
jtschuster: I removed the initialization above, the build fails without this assignment.

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
jtschuster: This is needed for the `DynamicallyAccessedMembersAttribute`s below in the `TestSystemTypeBase` class in the string.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -163,6 +175,13 @@ static void ProcessGenericParameters (SyntaxNodeAnalysisContext context)
 			}
 		}
 
+		static bool DamAttributesMatch (AttributeData attribute1, AttributeData attribute2)
+		{
+			return attribute1!.ConstructorArguments.All (methodArg =>
```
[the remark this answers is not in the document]
jtschuster: I ended up removing this method in the end.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -209,5 +222,73 @@ static void VerifyMemberOnlyApplyToTypesOrStrings (SymbolAnalysisContext context
...
+		{
+			if (context.Symbol.TryGetOverriddenMember (out var overriddenSymbol) && overriddenSymbol is IMethodSymbol overriddenMethod
+				&& context.Symbol is IMethodSymbol method) {
+				VerifyDamOnMethodsMatch (method, overriddenMethod, context);
+			}
+		}
+
+		static void VerifyDamOnMethodsMatch (IMethodSymbol method, IMethodSymbol overriddenMethod, SymbolAnalysisContext context)
+		{
+			if (FlowAnnotations.GetMethodReturnValueAnnotation (method)  != FlowAnnotations.GetMethodReturnValueAnnotation (overriddenMethod))
```
vitek-karas, line 236: Nit unnecessary space before the `!=`

src/ILLink.RoslynAnalyzer/INamedTypeSymbolExtensions.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.CodeAnalysis;
+using System.Collections.Generic;
```
vitek-karas, line 6: Sort usings

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
```diff
@@ -1,6 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System;
+using System.Diagnostics;
```
vitek-karas, line 5: Do we need these - there are no other changes in the file....

test/Mono.Linker.Tests.Cases/DataFlow/PropertyDataFlow.cs
```diff
@@ -292,6 +292,8 @@ public void TestPropertyWithExistingAttributes ()
 			[CompilerGenerated]
 			Type PropertyWithExistingAttributes_Field;
 
+			[ExpectedWarning ("IL2043", "PropertyWithExistingAttributes", "PropertyWithExistingAttributes.get", ProducedBy = ProducedBy.Analyzer)]
+			[ExpectedWarning ("IL2043", "PropertyWithExistingAttributes", "PropertyWithExistingAttributes.set", ProducedBy = ProducedBy.Analyzer)]
 			[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicConstructors)]
 			Type PropertyWithExistingAttributes {
 				// Analyzer doesn't try to detect backing fields of properties: https://github.com/dotnet/linker/issues/2273
```
vitek-karas, line 299: This comment is actually wrong - the IL2043 is not about backing fields - the difference is now solely about https://github.com/dotnet/linker/issues/2654. Could you please update the comments?

test/Mono.Linker.Tests.Cases/DataFlow/PropertyDataFlow.cs
```diff
@@ -328,6 +330,8 @@ public void TestPropertyWithConflictingAttributes ()
 			[CompilerGenerated]
 			Type PropertyWithConflictingAttributes_Field;
 
+			[ExpectedWarning ("IL2043", "PropertyWithConflictingAttributes", "PropertyWithConflictingAttributes.get", ProducedBy = ProducedBy.Analyzer)]
+			[ExpectedWarning ("IL2043", "PropertyWithConflictingAttributes", "PropertyWithConflictingAttributes.set", ProducedBy = ProducedBy.Analyzer)]
 			[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicConstructors)]
 			Type PropertyWithConflictingAttributes {
 				// Analyzer doesn't try to detect backing fields of properties: https://github.com/dotnet/linker/issues/2273
```
vitek-karas, line 337: Same as above - this is about https://github.com/dotnet/linker/issues/2654

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -209,5 +222,73 @@ static void VerifyMemberOnlyApplyToTypesOrStrings (SymbolAnalysisContext context
...
+						DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersMismatchOnGenericParameterBetweenOverrides),
+						method.TypeParameters[i].Locations[0],
+						method.TypeParameters[i].GetDisplayName (), method.GetDisplayName (),
+						overriddenMethod.TypeParameters[i].GetDisplayName (), overriddenMethod.GetDisplayName ()));
+
+			if (!method.IsStatic && method.GetDynamicallyAccessedMemberTypes () != overriddenMethod.GetDynamicallyAccessedMemberTypes ())
+				context.ReportDiagnostic (Diagnostic.Create (
+					DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersMismatchOnImplicitThisBetweenOverrides),
+					method.Locations[0],
+					method.GetDisplayName (), overriddenMethod.GetDisplayName ()));
```
vitek-karas, line 260: Could you please file an issue about the fact that linker doesn't validate that method overrides have the same annotation on the method itself? And also add tests for this case - it's not really covered anywhere as far as I can tell.

vitek-karas (approved):
Some minor comments, but looks good otherwise.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
jtschuster: I might be misunderstanding what you mean but it looks like we do have a test for this [here](https://github.com/dotnet/linker/blob/26e0c5cdc29a5a4f541100148b5cb7a814773110/test/Mono.Linker.Tests.Cases/DataFlow/VirtualMethodHierarchyDataflowAnnotationValidation.cs#L606-L633), and the linker reports this warning [here](https://github.com/dotnet/linker/blob/26e0c5cdc29a5a4f541100148b5cb7a814773110/src/linker/Linker.Dataflow/FlowAnnotations.cs#L540-L542).

vitek-karas: Sorry - you're right. I looked through the callers of `LogValidationWarning` and could not find the place where we would check on the method itself. But it's "hidden" in the `ValidateMethodParametersHaveNoAnnotations` - this method considers `this` as a parameter and that will validate the method-level annotations.

Please ignore my comment.

vitek-karas (approved):

src/ILLink.RoslynAnalyzer/INamedTypeSymbolExtensions.cs
```diff
@@ -28,5 +29,31 @@ internal static bool HasName (this INamedTypeSymbol type, string typeName)
...
+					yield return pair;
+				}
+			}
+		}
+
+		private static IEnumerable<(ISymbol InterfaceMember, ISymbol ImplementationMember)> GetMatchingMembers (INamedTypeSymbol implementationSymbol, INamedTypeSymbol interfaceSymbol)
+		{
+			var members = interfaceSymbol.GetMembers ();
+			foreach (var member in members) {
+				if (member is ISymbol interfaceMember
```
sbomer, line 47: nit: could we replace the `ISymbol` check by doing `foreach (ISymbol interfaceMember in interfaceSymbol.GetMembers ())`?

sbomer (approved):
LGTM, only a few nits. Thank you!

PR #2675, opened 2022-03-04
Propagate DynamicallyAccessedMemberTypes through Nullable<T> to T

jtschuster:
This PR enables annotations placed on Nullable<T> to propagate through to T in the linker. This does not work in the analyzer yet and will fail in CI until I update the analyzer to work with the changes.

This creates a new InstrinsicId value for `Nullable.GetUnderlyingType` it creates a new struct `NullableSystemTypeValue` that inherits from `SystemTypeValue`, and tracks the `MultiValue` of the wrapped type. It also creates `NullableRuntimeTypeHandleValue` which works similarly. In MethodBodyScanner, when previously a `RuntimeTypeHandleValue` was created in `ScanLdtoken`, it now will create a `NullableRuntimeTypeHandleValue` when the type of the handle is ``Nullable`1``.  Then when a `typeof` is called on that handle, it converts a `NullableRuntimeTypeHandleValue` to the `NullableSystemTypeValue`. Finally, if `Nullable.GetUnderlyingType` is called on that `NullableSystemTypeValue`, it propagates the annotations of each `ValueWithDynamicallyAccessedMembers` to the return value, or propagates the `SystemTypeValue` to the return value.

I think this covers all the areas where nullables need to be handled, but I might be missing something.

### DIFF
--- a/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
@@ -64,7 +64,9 @@ namespace ILLink.Shared.TrimAnalysis
 				}
 
 				foreach (var value in argumentValues[0]) {
-					if (value is RuntimeTypeHandleValue typeHandle)
+					if (value is RuntimeNullableTypeHandleValue runtimeNullableTypeHandleValue)
+						AddReturnValue (new SystemNullableTypeValue (runtimeNullableTypeHandleValue.RepresentedType, runtimeNullableTypeHandleValue.UnderlyingTypeValue));
+					else if (value is RuntimeTypeHandleValue typeHandle)
 						AddReturnValue (new SystemTypeValue (typeHandle.RepresentedType));
 					else if (value is RuntimeTypeHandleForGenericParameterValue typeHandleForGenericParameter)
 						AddReturnValue (GetGenericParameterValue (typeHandleForGenericParameter.GenericParameter));
@@ -541,7 +543,16 @@ namespace ILLink.Shared.TrimAnalysis
 				break;
 
 			case IntrinsicId.Nullable_GetUnderlyingType:
-				AddReturnValue (argumentValues[0]);
+				var values = new List<SingleValue>();
+				foreach (var singlevalue in argumentValues[0].AsEnumerable ()) {
+					if (singlevalue is ValueWithDynamicallyAccessedMembers damValue) {
+						values.Add(GetMethodReturnValue(calledMethod, damValue.DynamicallyAccessedMemberTypes));
+					}
+					else if (singlevalue is SystemNullableTypeValue systemNullableTypeValue) {
+						values.AddRange (systemNullableTypeValue.UnderlyingTypeValue);
+					}
+				}
+				returnValue = new MultiValue (values);
 				break;
 
 			case IntrinsicId.None:
@@ -566,7 +577,7 @@ namespace ILLink.Shared.TrimAnalysis
 					if (uniqueValue is ValueWithDynamicallyAccessedMembers methodReturnValueWithMemberTypes) {
 						if (!methodReturnValueWithMemberTypes.DynamicallyAccessedMemberTypes.HasFlag (returnValueDynamicallyAccessedMemberTypes))
 							throw new InvalidOperationException ($"Internal linker error: in {GetContainingSymbolDisplayName ()} processing call to {calledMethod.GetDisplayName ()} returned value which is not correctly annotated with the expected dynamic member access kinds.");
-					} else if (uniqueValue is SystemTypeValue) {
+					} else if (uniqueValue is SystemTypeValue systemTypeValue) {
 						// SystemTypeValue can fullfill any requirement, so it's always valid
 						// The requirements will be applied at the point where it's consumed (passed as a method parameter, set as field value, returned from the method)
 					} else if (uniqueValue == NullValue.Instance) {
--- a/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
@@ -64,8 +64,8 @@ namespace ILLink.Shared.TrimAnalysis
 				}
 
 				foreach (var value in argumentValues[0]) {
-					if (value is RuntimeNullableTypeHandleValue runtimeNullableTypeHandleValue)
-						AddReturnValue (new SystemNullableTypeValue (runtimeNullableTypeHandleValue.RepresentedType, runtimeNullableTypeHandleValue.UnderlyingTypeValue));
+					if (value is NullableRuntimeTypeHandleValue runtimeNullableTypeHandleValue)
+						AddReturnValue (new NullableSystemTypeValue (runtimeNullableTypeHandleValue.RepresentedType, runtimeNullableTypeHandleValue.UnderlyingTypeValue));
 					else if (value is RuntimeTypeHandleValue typeHandle)
 						AddReturnValue (new SystemTypeValue (typeHandle.RepresentedType));
 					else if (value is RuntimeTypeHandleForGenericParameterValue typeHandleForGenericParameter)
@@ -543,12 +543,11 @@ namespace ILLink.Shared.TrimAnalysis
 				break;
 
 			case IntrinsicId.Nullable_GetUnderlyingType:
-				var values = new List<SingleValue>();
+				var values = new List<SingleValue> ();
 				foreach (var singlevalue in argumentValues[0].AsEnumerable ()) {
 					if (singlevalue is ValueWithDynamicallyAccessedMembers damValue) {
-						values.Add(GetMethodReturnValue(calledMethod, damValue.DynamicallyAccessedMemberTypes));
-					}
-					else if (singlevalue is SystemNullableTypeValue systemNullableTypeValue) {
+						values.Add (GetMethodReturnValue (calledMethod, damValue.DynamicallyAccessedMemberTypes));
+					} else if (singlevalue is NullableSystemTypeValue systemNullableTypeValue) {
 						values.AddRange (systemNullableTypeValue.UnderlyingTypeValue);
 					}
 				}
--- a/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
@@ -577,7 +577,7 @@ namespace ILLink.Shared.TrimAnalysis
 					if (uniqueValue is ValueWithDynamicallyAccessedMembers methodReturnValueWithMemberTypes) {
 						if (!methodReturnValueWithMemberTypes.DynamicallyAccessedMemberTypes.HasFlag (returnValueDynamicallyAccessedMemberTypes))
 							throw new InvalidOperationException ($"Internal linker error: in {GetContainingSymbolDisplayName ()} processing call to {calledMethod.GetDisplayName ()} returned value which is not correctly annotated with the expected dynamic member access kinds.");
-					} else if (uniqueValue is SystemTypeValue systemTypeValue) {
+					} else if (uniqueValue is SystemTypeValue) {
 						// SystemTypeValue can fullfill any requirement, so it's always valid
 						// The requirements will be applied at the point where it's consumed (passed as a method parameter, set as field value, returned from the method)
 					} else if (uniqueValue == NullValue.Instance) {
--- a/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
@@ -540,6 +540,10 @@ namespace ILLink.Shared.TrimAnalysis
 				}
 				break;
 
+			case IntrinsicId.Nullable_GetUnderlyingType:
+				AddReturnValue (argumentValues[0]);
+				break;
+
 			case IntrinsicId.None:
 				methodReturnValue = MultiValueLattice.Top;
 				return false;
--- a/src/ILLink.Shared/TrimAnalysis/RuntimeTypeHandleValue.cs
+++ b/src/ILLink.Shared/TrimAnalysis/RuntimeTypeHandleValue.cs
@@ -3,18 +3,27 @@
 
 using ILLink.Shared.DataFlow;
 using ILLink.Shared.TypeSystemProxy;
+using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.SingleValue>;
 
 namespace ILLink.Shared.TrimAnalysis
 {
 	/// <summary>
 	/// This is the System.RuntimeTypeHandle equivalent to a <see cref="SystemTypeValue"/> node.
 	/// </summary>
-	sealed record RuntimeTypeHandleValue : SingleValue
+	record RuntimeTypeHandleValue : SingleValue
 	{
 		public RuntimeTypeHandleValue (in TypeProxy representedType) => RepresentedType = representedType;
 
 		public readonly TypeProxy RepresentedType;
 
 		public override string ToString () => this.ValueToString (RepresentedType);
+	}
+
+	sealed record RuntimeNullableTypeHandleValue : RuntimeTypeHandleValue
+	{
+		public RuntimeNullableTypeHandleValue (in TypeProxy representedType, in MultiValue underlyingTypeValue) : base(representedType)
+			=> UnderlyingTypeValue = underlyingTypeValue;
+
+		public readonly MultiValue UnderlyingTypeValue;
 	}
 }
--- a/src/ILLink.Shared/TrimAnalysis/RuntimeTypeHandleValue.cs
+++ b/src/ILLink.Shared/TrimAnalysis/RuntimeTypeHandleValue.cs
@@ -19,9 +19,9 @@ namespace ILLink.Shared.TrimAnalysis
 		public override string ToString () => this.ValueToString (RepresentedType);
 	}
 
-	sealed record RuntimeNullableTypeHandleValue : RuntimeTypeHandleValue
+	sealed record NullableRuntimeTypeHandleValue : RuntimeTypeHandleValue
 	{
-		public RuntimeNullableTypeHandleValue (in TypeProxy representedType, in MultiValue underlyingTypeValue) : base(representedType)
+		public NullableRuntimeTypeHandleValue (in TypeProxy representedType, in MultiValue underlyingTypeValue) : base (representedType)
 			=> UnderlyingTypeValue = underlyingTypeValue;
 
 		public readonly MultiValue UnderlyingTypeValue;
--- a/src/ILLink.Shared/TrimAnalysis/SystemTypeValue.cs
+++ b/src/ILLink.Shared/TrimAnalysis/SystemTypeValue.cs
@@ -15,15 +15,15 @@ namespace ILLink.Shared.TrimAnalysis
 	record SystemTypeValue : SingleValue
 	{
 		public SystemTypeValue (in TypeProxy representedType) => RepresentedType = representedType;
-		
+
 		public readonly TypeProxy RepresentedType;
 
 		public override string ToString () => this.ValueToString (RepresentedType);
 	}
 
-	sealed record SystemNullableTypeValue : SystemTypeValue
+	sealed record NullableSystemTypeValue : SystemTypeValue
 	{
-		public SystemNullableTypeValue (in TypeProxy representedType, in MultiValue underlyingTypeValue) : base(representedType)
+		public NullableSystemTypeValue (in TypeProxy representedType, in MultiValue underlyingTypeValue) : base (representedType)
 			=> UnderlyingTypeValue = underlyingTypeValue;
 
 		public readonly MultiValue UnderlyingTypeValue;
--- a/src/ILLink.Shared/TrimAnalysis/SystemTypeValue.cs
+++ b/src/ILLink.Shared/TrimAnalysis/SystemTypeValue.cs
@@ -1,20 +1,31 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using ILLink.Shared.DataFlow;
 using ILLink.Shared.TypeSystemProxy;
+using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.SingleValue>;
 
 namespace ILLink.Shared.TrimAnalysis
 {
 	/// <summary>
 	/// This is a known System.Type value. TypeRepresented is the 'value' of the System.Type.
 	/// </summary>
-	sealed record SystemTypeValue : SingleValue
+	record SystemTypeValue : SingleValue
 	{
 		public SystemTypeValue (in TypeProxy representedType) => RepresentedType = representedType;
-
+		
 		public readonly TypeProxy RepresentedType;
 
 		public override string ToString () => this.ValueToString (RepresentedType);
+	}
+
+	sealed record SystemNullableTypeValue : SystemTypeValue
+	{
+		public SystemNullableTypeValue (in TypeProxy representedType, in MultiValue underlyingTypeValue) : base(representedType)
+			=> UnderlyingTypeValue = underlyingTypeValue;
+
+		public readonly MultiValue UnderlyingTypeValue;
 	}
 }
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using ILLink.Shared.DataFlow;
 using ILLink.Shared.TrimAnalysis;
+using ILLink.Shared.TypeSystemProxy;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using Mono.Collections.Generic;
@@ -733,10 +734,19 @@ namespace Mono.Linker.Dataflow
 
 			if (operation.Operand is TypeReference typeReference) {
 				var resolvedReference = ResolveToTypeDefinition (typeReference);
+
 				if (resolvedReference != null) {
-					StackSlot slot = new StackSlot (new RuntimeTypeHandleValue (resolvedReference));
-					currentStack.Push (slot);
-					return;
+					if (resolvedReference.Name == "Nullable`1" && resolvedReference.Namespace == "System" 
+						&& typeReference is IGenericInstance instance && ResolveToTypeDefinition (instance.GenericArguments![0]) is TypeDefinition underlyingType) {
+						StackSlot slot1 = new StackSlot (new RuntimeNullableTypeHandleValue (TypeProxy.NullableTypeProxy (resolvedReference, underlyingType), new SystemTypeValue (underlyingType)));
+						currentStack.Push (slot1);
+						return;
+					}
+					else {
+						StackSlot slot = new StackSlot (new RuntimeTypeHandleValue (resolvedReference));
+						currentStack.Push (slot);
+						return;
+					}
 				}
 			} else if (operation.Operand is MethodReference methodReference) {
 				var resolvedMethod = _context.TryResolve (methodReference);
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -734,15 +734,13 @@ namespace Mono.Linker.Dataflow
 
 			if (operation.Operand is TypeReference typeReference) {
 				var resolvedReference = ResolveToTypeDefinition (typeReference);
-
 				if (resolvedReference != null) {
-					if (resolvedReference.Name == "Nullable`1" && resolvedReference.Namespace == "System" 
+					if (resolvedReference.Name == "Nullable`1" && resolvedReference.Namespace == "System"
 						&& typeReference is IGenericInstance instance && ResolveToTypeDefinition (instance.GenericArguments![0]) is TypeDefinition underlyingType) {
-						StackSlot slot1 = new StackSlot (new RuntimeNullableTypeHandleValue (new TypeProxy(resolvedReference), new SystemTypeValue (underlyingType)));
+						StackSlot slot1 = new StackSlot (new NullableRuntimeTypeHandleValue (new TypeProxy (resolvedReference), new SystemTypeValue (underlyingType)));
 						currentStack.Push (slot1);
 						return;
-					}
-					else {
+					} else {
 						StackSlot slot = new StackSlot (new RuntimeTypeHandleValue (resolvedReference));
 						currentStack.Push (slot);
 						return;
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -738,7 +738,7 @@ namespace Mono.Linker.Dataflow
 				if (resolvedReference != null) {
 					if (resolvedReference.Name == "Nullable`1" && resolvedReference.Namespace == "System" 
 						&& typeReference is IGenericInstance instance && ResolveToTypeDefinition (instance.GenericArguments![0]) is TypeDefinition underlyingType) {
-						StackSlot slot1 = new StackSlot (new RuntimeNullableTypeHandleValue (TypeProxy.NullableTypeProxy (resolvedReference, underlyingType), new SystemTypeValue (underlyingType)));
+						StackSlot slot1 = new StackSlot (new RuntimeNullableTypeHandleValue (new TypeProxy(resolvedReference), new SystemTypeValue (underlyingType)));
 						currentStack.Push (slot1);
 						return;
 					}
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -288,6 +288,7 @@ namespace Mono.Linker.Dataflow
 			case IntrinsicId.Type_GetMember:
 			case IntrinsicId.Type_GetMethod:
 			case IntrinsicId.Type_GetNestedType:
+			case IntrinsicId.Nullable_GetUnderlyingType:
 			case IntrinsicId.Expression_Property when calledMethod.HasParameterOfType (1, "System.Reflection.MethodInfo"):
 			case var fieldOrPropertyInstrinsic when fieldOrPropertyInstrinsic == IntrinsicId.Expression_Field || fieldOrPropertyInstrinsic == IntrinsicId.Expression_Property: {
 					var instanceValue = MultiValueLattice.Top;
@@ -806,10 +807,6 @@ namespace Mono.Linker.Dataflow
 					// MakeGenericMethod doesn't change the identity of the MethodBase we're tracking so propagate to the return value
 					methodReturnValue = methodParams[0];
 				}
-				break;
-
-			case IntrinsicId.Nullable_GetUnderlyingType:
-				methodReturnValue = methodParams[0];
 				break;
 
 			default:
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -808,6 +808,10 @@ namespace Mono.Linker.Dataflow
 				}
 				break;
 
+			case IntrinsicId.Nullable_GetUnderlyingType:
+				methodReturnValue = methodParams[0];
+				break;
+
 			default:
 
 				if (calledMethodDefinition.IsPInvokeImpl) {
--- a/test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
@@ -58,8 +58,8 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 
 		}
 
-		[ExpectedWarning("IL2026", "message")]
-		static void RequireMethodWithRUC()
+		[ExpectedWarning ("IL2026", "message")]
+		static void RequireMethodWithRUC ()
 		{
 			var T = typeof (Nullable<TestStructWithRucMethod>);
 			var uT = Nullable.GetUnderlyingType (T);
@@ -77,20 +77,20 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 
 		static void NullableOfAnnotatedGenericParameter<[DAM (DAMT.PublicProperties)] T> () where T : struct
 		{
-			RequireProperties (Nullable.GetUnderlyingType(typeof(Nullable<T>)));
+			RequireProperties (Nullable.GetUnderlyingType (typeof (Nullable<T>)));
 		}
 
 		[ExpectedWarning ("IL2072")]
 		static void UnderlyingTypeOfUnannotatedGenericParameterRequiresProperties<TNullable> ()
 		{
-			RequireProperties (Nullable.GetUnderlyingType(typeof(TNullable)));
+			RequireProperties (Nullable.GetUnderlyingType (typeof (TNullable)));
 		}
 
 		static void RequiresPropertiesFromInstance<[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicProperties)] T> (T instance)
 		{
 		}
 
-		[return: DynamicallyAccessedMembers(DAMT.PublicProperties)]
+		[return: DynamicallyAccessedMembers (DAMT.PublicProperties)]
 		static Type ReturnUnderlyingTypeThatRequiresProperties<[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.PublicProperties)] T> (T instance)
 		{
 			Type type = Nullable.GetUnderlyingType (typeof (T)) ?? typeof (T); // RuntimeHandleForGenericParameterValue for typeof(T) does not carry annotations through -- JK looks like it did once I get to the GetUnderlyingType
--- a/test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
@@ -4,6 +4,9 @@
 using System;
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 using System.Diagnostics.CodeAnalysis;
+using Mono.Linker.Tests.Cases.Expectations.Helpers;
+using DAM = System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute;
+using DAMT = System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes;
 
 namespace Mono.Linker.Tests.Cases.DataFlow
 {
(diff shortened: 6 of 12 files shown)


test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
```diff
@@ -0,0 +1,101 @@
...
+		{
+			var T = typeof (Nullable<TestStructWithRucMethod>);
+			var uT = Nullable.GetUnderlyingType (T);
+			uT.RequiresAll ();
+		}
+
+		static void RequireProperties ([DAM (DAMT.PublicProperties)] Type t) { }
+
+		static void RequireAll ([DAM (DAMT.All)] Type t) { }
+
```
vitek-karas, line 72: There are extension method on `Type` defined in the tests in `DataFlowTypeExtensions`

test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
```diff
@@ -0,0 +1,101 @@
...
+		static void UnderlyingTypeOfAnnotatedParameterRequiresAll ([DAM (DAMT.All)] Type t)
+		{
+			RequireAll (Nullable.GetUnderlyingType (t));
+		}
+
+		static void NullableOfAnnotatedGenericParameter<[DAM (DAMT.PublicProperties)] T> () where T : struct
+		{
+			RequireProperties (Nullable.GetUnderlyingType (typeof (Nullable<T>)));
```
vitek-karas, line 80: I'm surprised this doesn't warn... - hmmm

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -540,6 +542,18 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
...
+			case IntrinsicId.Nullable_GetUnderlyingType:
+				var values = new List<SingleValue> ();
+				foreach (var singlevalue in argumentValues[0].AsEnumerable ()) {
+					if (singlevalue is ValueWithDynamicallyAccessedMembers damValue) {
+						values.Add (GetMethodReturnValue (calledMethod, damValue.DynamicallyAccessedMemberTypes));
+					} else if (singlevalue is NullableSystemTypeValue systemNullableTypeValue) {
+						values.AddRange (systemNullableTypeValue.UnderlyingTypeValue);
+					}
+				}
+				returnValue = new MultiValue (values);
```
vitek-karas, line 554: Nit: This conversion should be implicit (no need for `new MultiValue`)

vitek-karas (commented):
Other than the comments (main worry there is handling of generic arguments), this is missing marking changes.

We need to add a test which validates that code like
```C#
[StructLayout(LayoutKind.Auto)]  // This is important since without this linker will not trim fields, but add a test for methods as well just in case
struct TestS
{
    [Kept]  // <--- this is the important bit. There's no reference to Field, it should be preserved because of DAM
    public int Field;
}

void Test()
{
    typeof(Nullable<StructS>).RequirePublicFields(); // Applies DAM PublicFields onto the type, this should propagate to the StructS.
}
```

Note that "marking" is not just a linker concept - in analyzer it means check all members of the "marked" type that they're not RUC (for example). So add tests which check for these warnings as well - those will be valid for both linker and analyzer.

But we do need the "Keep" tests for linker as well.

test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
jtschuster: Thanks for pointing this out, it looks like in this case the Nullable<T> is still represented by SystemTypeValue, not NullableSystemTypeValue, so that needs to be fixed.

vitek-karas: Can you please also add a negative test - that if the inner T is not annotated this generates a warning?

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -170,7 +170,13 @@ static SingleValue GetTypeValueNodeFromGenericArgument (SyntaxNodeAnalysisContex
 				SymbolKind.TypeParameter => new GenericParameterValue ((ITypeParameterSymbol) type),
 				// Technically this should be a new value node type as it's not a System.Type instance representation, but just the generic parameter
 				// That said we only use it to perform the dynamically accessed members checks and for that purpose treating it as System.Type is perfectly valid.
-				SymbolKind.NamedType => new SystemTypeValue (new TypeProxy ((INamedTypeSymbol) type)),
+				SymbolKind.NamedType 
+					=> ((INamedTypeSymbol)type).NullableAnnotation switch {
```
vitek-karas, line 174: I find this pretty weird - so roslyn doesn't translate nullable value types to true `Nullable<T>` type symbols?
If not, then this will need to somehow differentiate ref types and not do this for those.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -82,15 +86,18 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
...
-					if (value is SystemTypeValue typeValue)
-						AddReturnValue (new RuntimeTypeHandleValue (typeValue.RepresentedType));
-					else if (value is GenericParameterValue genericParameterValue)
-						AddReturnValue (new RuntimeTypeHandleForGenericParameterValue (genericParameterValue.GenericParameter));
-					else if (value == NullValue.Instance) {
-						// Throws if the input is null, so no return value.
-						returnValue ??= MultiValueLattice.Top;
-					} else
-						AddReturnValue (GetMethodReturnValue (calledMethod, returnValueDynamicallyAccessedMemberTypes));
+					if (value != NullValue.Instance)
```
vitek-karas, line 89: You're missing an else branch which would populate the return value to Top (the this worked before).

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -545,11 +552,13 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
 			case IntrinsicId.Nullable_GetUnderlyingType:
 				var values = new List<SingleValue> ();
 				foreach (var singlevalue in argumentValues[0].AsEnumerable ()) {
-					if (singlevalue is ValueWithDynamicallyAccessedMembers damValue) {
-						values.Add (GetMethodReturnValue (calledMethod, damValue.DynamicallyAccessedMemberTypes));
-					} else if (singlevalue is NullableSystemTypeValue systemNullableTypeValue) {
-						values.AddRange (systemNullableTypeValue.UnderlyingTypeValue);
-					}
+					values.Add (singlevalue switch {
+						SystemTypeValue systemType => systemType,
```
vitek-karas, line 556: Maybe add an assert (if possible) that the type is not Nullable<T> - since below we rely on that.

src/ILLink.Shared/TrimAnalysis/SystemTypeValue.cs
```diff
@@ -19,11 +21,33 @@ record SystemTypeValue : SingleValue
...
+			UnderlyingTypeValue = underlyingType;
+			NullableType = nullableType;
+		}
+		public readonly TypeProxy NullableType;
 
-		public readonly MultiValue UnderlyingTypeValue;
+		public readonly TypeProxy UnderlyingTypeValue;
+	}
+
+	sealed record NullableValueWithDynamicallyAccessedMembers : ValueWithDynamicallyAccessedMembers
```
vitek-karas, line 37: The only reason we need this value type is that applying annotation to it needs to apply the annotation to both the Nullable<T> and the underlying type (otherwise treating this as the underlying value would work, since we transparently return the same value when asked for underlying type).
Could you please add a comment what is the purpose of this value node?

vitek-karas: Similar reason applies to the `NullableSystemTypeValue`.

src/ILLink.Shared/TrimAnalysis/RuntimeTypeHandleValue.cs
```diff
@@ -1,16 +1,18 @@
...
 
 namespace ILLink.Shared.TrimAnalysis
 {
 	/// <summary>
 	/// This is the System.RuntimeTypeHandle equivalent to a <see cref="SystemTypeValue"/> node.
 	/// </summary>
-	record RuntimeTypeHandleValue : SingleValue
+	sealed record RuntimeTypeHandleValue : SingleValue
 	{
 		public RuntimeTypeHandleValue (in TypeProxy representedType) => RepresentedType = representedType;
```
vitek-karas, line 17: Maybe add an assert that the type is not `Nullable<T>`

src/ILLink.Shared/TrimAnalysis/RequireDynamicallyAccessedMembersAction.cs
```diff
@@ -26,6 +26,9 @@ public void Invoke (in MultiValue value, ValueWithDynamicallyAccessedMembers tar
 					&& genericParam.HasDefaultConstructorConstraint ()) {
 					// We allow a new() constraint on a generic parameter to satisfy DynamicallyAccessedMemberTypes.PublicParameterlessConstructor
 				} else if (uniqueValue is ValueWithDynamicallyAccessedMembers valueWithDynamicallyAccessedMembers) {
+					if (uniqueValue is NullableValueWithDynamicallyAccessedMembers nullableValue) {
+						MarkTypeForDynamicallyAccessedMembers (nullableValue.NullableType, nullableValue.DynamicallyAccessedMemberTypes);
```
vitek-karas, line 30: I think this is missing the marking of `Nullable<T>` itself.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -327,6 +327,7 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 			// Type MakeGenericType (params Type[] typeArguments)
 			//
 			case IntrinsicId.Type_MakeGenericType: {
+					// We don't yet handle the very unlikely case where you can create a nullable type with typeof(Nullable<>).MakeGenericType(T)
```
vitek-karas, line 330: This is actually not that unlikely, but it's fine to not handle it right now.

src/ILLink.Shared/TrimAnalysis/RequireDynamicallyAccessedMembersAction.cs
jtschuster: `nullableValue.NullableType` represents the `Nullable<>`, and since `NullableValueWithDynamicallyAccessedMembers` has an unknown underlying type, we wouldn't want to mark anything else, would we?

vitek-karas: Sorry - I'm blind. Please ignore.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
jtschuster: Thanks for pointing that out, I hit some issues because of that.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
jtschuster: It looks like that `NullableAnnotation` isn't reliable for our purposes, so I had to change it to check the name and namespace instead.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
vitek-karas: You're right, if the input is a list then there's no implicit conversion.

src/ILLink.Shared/TrimAnalysis/RuntimeTypeHandleValue.cs
jtschuster: When I added an assert I ran into cases where ``System.Nullable`1`` wouldn't have a type parameter, so it wouldn't make a `NullableValue` and would instantiate a SystemTypeValue or RuntimeTypeHandleValue. To me a SystemTypeValue makes sense since there is nothing extra to track in those situations, but would it make more sense to have the `UnderlyingType` of `NullableValue`'s be optional?

src/ILLink.Shared/TrimAnalysis/SystemTypeValue.cs
jtschuster: Yes, I added comments for those definitions.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -734,9 +734,16 @@ void ScanLdtoken (Instruction operation, Stack<StackSlot> currentStack)
 			if (operation.Operand is TypeReference typeReference) {
 				var resolvedReference = ResolveToTypeDefinition (typeReference);
 				if (resolvedReference != null) {
-					StackSlot slot = new StackSlot (new RuntimeTypeHandleValue (resolvedReference));
-					currentStack.Push (slot);
-					return;
+					if (resolvedReference.Name == "Nullable`1" && resolvedReference.Namespace == "System"
+						&& typeReference is IGenericInstance instance && ResolveToTypeDefinition (instance.GenericArguments![0]) is TypeDefinition underlyingType) {
```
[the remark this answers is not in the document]
jtschuster: I added a test for this and it's passing now.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
jtschuster: Updated the comment accordingly.

test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
```diff
@@ -0,0 +1,289 @@
...
+		static void TestGetUnderlyingTypeOnClasses ()
+		{
+			UnderlyingTypeOfAnnotatedParameterRequiresPublicProperties (typeof (TestClass));
+			UnderlyingTypeOfAnnotatedGenericParameterRequiresPublicProperties<TestClass> ();
+			UnderlyingTypeOfUnannotatedParameterRequiresPublicProperties (typeof (TestClass));
+			UnderlyingTypeOfUnannotatedGenericParameterRequiresProperties<TestClass> ();
+		}
+
+		[Kept]
+		static void TestGetUnderlyingTypeOnStructs ()
```
vitek-karas, line 143: I don't see the value of having both class and struct tests for this. Basically what this does is validating that applying PublicProperties annotation to a class or struct works correctly. While a valid test, it has nothing to do with nullable. The fact that the method it calls does some nullable magic inside, is not seen at this callsite.

test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
```diff
@@ -0,0 +1,289 @@
...
+			UnderlyingTypeOfUnannotatedGenericParameterRequiresProperties<TestStruct> ();
+		}
+
+		[Kept]
+		static void TestGetUnderlyingTypeOnNullableStructs ()
+		{
+			UnderlyingTypeOfAnnotatedParameterRequiresPublicProperties (typeof (Nullable<TestStruct>));
+			UnderlyingTypeOfAnnotatedGenericParameterRequiresPublicProperties<Nullable<TestStruct>> ();
+			UnderlyingTypeOfUnannotatedParameterRequiresPublicProperties (typeof (Nullable<TestStruct>));
+			UnderlyingTypeOfUnannotatedGenericParameterRequiresProperties<Nullable<TestStruct>> ();
```
vitek-karas, line 157: While valid, the test is corrupted by the fact that we apply the annotation to `TestStruct` directly above. What we want to test is that passing `Nulable<TestStruct>` to annotated value applies the annotation on the `TestStruct` - but to do that we need to make sure that `TestStruct` is not marked by anything anywhere else in the test.

test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
```diff
@@ -0,0 +1,289 @@
...
+		}
+
+		[Kept]
+		static void TestGetUnderlyingTypeOfCreatedNullableOnStructs ()
+		{
+			NullableOfAnnotatedParameterRequiresPublicProperties (typeof (TestStruct));
+			NullableOfUnannotatedGenericParameterRequiresPublicProperties<TestStruct> ();
+			NullableOfUnannotatedParameterRequiresPublicProperties (typeof (TestStruct));
+			NullableOfUnannotatedGenericParameterRequiresPublicProperties<TestStruct> ();
```
vitek-karas, line 166: Same as above - we just need to call these methods to "Trigger" their analysis - for that we should use some random `TestType` (note that lot of the data flow tests have an empty `TestType` exactly for this reason, just to call a method).

test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
```diff
@@ -0,0 +1,289 @@
...
+		}
+
+		[Kept]
+		// Bug - GetUnderlyingType intrinsic handling does not yet have special handling of Nullables
+		//[ExpectedWarning("IL2067")]
+		static void NullableOfUnannotatedParameterRequiresPublicProperties (Type t)
```
vitek-karas, line 172: Exactly the case I used as my counter example above... I think we need to fix this.

test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
vitek-karas: You actually have tests like this below... I guess my only complaint is then the use of both TestStruct and TestClass - I don't see the need for that - they will do the exact same thing.

test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
```diff
@@ -0,0 +1,289 @@
...
+			TestGetUnderlyingTypeOnStructs ();
+			TestGetUnderlyingTypeOnNullableStructs ();
+			TestGetUnderlyingTypeOfCreatedNullableOnStructs ();
+		}
+
+		[Kept]
+		[ExpectedWarning ("IL2026", "message")]
+		static void RequireMethodWithRUC ()
+		{
+			var T = typeof (Nullable<TestStructWithRucMethod>);
```
vitek-karas, line 83: To test the analyzer we should also have a test which does just `typeof(Nullable<TestStructWithRucMethod>).RequiresAll()` - and that should also produce IL2026.

test/Mono.Linker.Tests.Cases/DataFlow/NullableAnnotations.cs
```diff
@@ -0,0 +1,289 @@
...
+			[Kept]
+			[KeptBackingField]
+			public static string StaticString { [Kept]get; [Kept]set; }
+
+			public void Method () { }
+		}
+
+		[Kept]
+		public static void Main ()
+		{
```
vitek-karas, line 59: Basically all of the parameters or generic parameters here are not actually testing anything - so we should use one empty TestType for all of it. Currently we use various types from below which may (or may not) corrupt the test by applying marking on things we didn't wanted.

vitek-karas (changes requested):
I think we need to fix the MakeGenericType as it currently creates an analysis hole which could break apps.

src/ILLink.Shared/TrimAnalysis/NullableRuntimeTypeHandleWithGenericParameterValue.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace ILLink.Shared.TrimAnalysis
+{
+	/// <summary>
+	/// This represents a type handle of a Nullable<T> where T is an unknown value with DynamicallyAccessedMembers annotations. 
+	/// It is necessary to track the underlying type to ensure DynamicallyAccessedMembers annotations on the underlying type match the target parameters where the Nullable is used.
+	/// </summary>
+	sealed record NullableRuntimeTypeWithDamHandleValue : SingleValue
+	{
+		public NullableRuntimeTypeWithDamHandleValue (in TypeProxy nullableType, in SingleValue underlyingTypeValue)
```
jtschuster, line 16: `UnderlyingTypeValue needs to be SingleValue here because in MethodBodyScanner, a RuntimeTypeHandleForGenericParameterValue is passed as the underlying type.

jtschuster:
To support `typeof(Nullable<>).MakeGenericType(T)`, it was necessary to add an ArrayCreationOperation visitor since the arguments to MakeGenericType is a params array. This also meant `ArrayValue.TryGetValueAtIndex` had to be implemented in the analyzer. I still need to make a few more changes to clean up the code.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -64,12 +64,20 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
...
-					if (value is RuntimeTypeHandleValue typeHandle)
-						AddReturnValue (new SystemTypeValue (typeHandle.RepresentedType));
-					else if (value is RuntimeTypeHandleForGenericParameterValue typeHandleForGenericParameter)
-						AddReturnValue (GetGenericParameterValue (typeHandleForGenericParameter.GenericParameter));
-					else
-						AddReturnValue (GetMethodReturnValue (calledMethod, returnValueDynamicallyAccessedMemberTypes));
+					AddReturnValue (value switch {
+						NullableRuntimeSystemTypeHandleValue nullableSystemType
+							=> new NullableSystemTypeValue (nullableSystemType.NullableType, nullableSystemType.UnderlyingTypeValue),
+						NullableRuntimeTypeWithDamHandleValue nullableDamType when nullableDamType.UnderlyingTypeValue is RuntimeTypeHandleForGenericParameterValue underlyingGenericParameter
```
sbomer, line 70: Does this need a corresponding case in `Type_get_TypeHandle` that turns a `NullableValueWithDynamicallyAccessedMembers` of `GenericParameterValue` back into a `NullableRuntimeTypeWithDamHandleValue` of `RuntimeTypeHandleForGenericParameterValue`? I am wondering why there's an asymmetry between the intrinsics that go from TypeHandle -> Type vs Type -> TypeHandle.

src/ILLink.Shared/DataFlow/IDeepCopyValue.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace ILLink.Shared.DataFlow
+{
+	// Adds ability to deep copy a value
+	public interface IDeepCopyValue<TSingleValue>
```
sbomer, line 7: I think ideally this would be required by every type used as a value in a lattice, but we could do that in a separate change.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -540,6 +553,50 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
...
+				}
+				break;
+
+			case IntrinsicId.Type_MakeGenericType:
+				// Contains part of the analysis done in the Linker, but is not complete
+				// instanceValue here is like argumentValues[0] in linker
+				foreach (var value in instanceValue) {
+					// Nullables without a type argument are considered SystemTypeValues
+					if (value is SystemTypeValue typeValue
+							&& typeValue.RepresentedType.IsTypeOf ("System", "Nullable`1")) {
```
sbomer, line 574: nit: to reduce some of the nesting here, could you use a style that inverts the condition? Like this:
```csharp
if (!condition)
    continue;
foreach (var argumentValue in argumentValues[0]) {
// ...
}
```

And similar below, inside the for loop.

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -194,7 +199,12 @@ public override TValue VisitArrayElementReference (IArrayElementReferenceOperati
 			if (operation.GetValueUsageInfo (Context.OwningSymbol).HasFlag (ValueUsageInfo.Read)) {
 				// Accessing an array element for reading is a call to the indexer
 				// or a plain array access. Just handle plain array access for now.
-				return HandleArrayElementAccess (operation.ArrayReference);
+
+				// Only handle simple index access
+				if (operation.Indices.Length != 1)
+					return TopValue;
```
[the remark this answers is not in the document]
vitek-karas: I think linker in this case ends up with `UnknownValue` - but I also think it's OK to keep the analyzer with `TopValue` - for these cases where we "didn't implement it yet", I'm perfectly OK with analyzer not producing warnings and leaving it up to the linker to do so... or not. It's the "Safer" option :-).
+1 on adding a test.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -162,11 +188,39 @@ public override void HandleAssignment (MultiValue source, MultiValue target, IOp
...
+		{
+			int? index = indexValue.AsConstInt ();
+			foreach (var arraySingleValue in arrayValue) {
+				if (arraySingleValue is ArrayValue arr) {
+					if (index == null) {
+						// Reset the array to all unknowns - since we don't know which index is being assigned
+						arr.IndexValues.Clear ();
+					} else {
+						if (arr.IndexValues.TryGetValue (index.Value, out var existingValue)) {
+							arr.IndexValues[index.Value] = _multiValueLattice.Meet (existingValue, valueToWrite);
```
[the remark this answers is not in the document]
vitek-karas: Hmm - in the analyzer you're probably right - it should overwrite. Linker is different since it doesn't have the solver yet, so it tracks basic blocks - but it does overwrite withing the same basic block.

src/ILLink.Shared/DataFlow/IDeepCopyValue.cs
vitek-karas: Yeah - I think it makes sense to require it everywhere - this change does require it on SingleValue. Other values - I would also prefer a different PR>

src/ILLink.Shared/TrimAnalysis/NullableRuntimeSystemTypeHandleValue.cs
```diff
@@ -0,0 +1,30 @@
...
+using ILLink.Shared.DataFlow;
+using ILLink.Shared.TypeSystemProxy;
+
+namespace ILLink.Shared.TrimAnalysis
+{
+	/// <summary>
+	/// This represents a type handle Nullable<T> where T is a known SystemTypeValue.
+	/// It is necessary to track the underlying type to propagate DynamicallyAccessedMembers annotations to the underlying type when applied to a Nullable.
+	/// </summary>
+	sealed record NullableRuntimeSystemTypeHandleValue : SingleValue
```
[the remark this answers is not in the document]
vitek-karas: Same question about `SystemTypeValue` and `RuntimeTypeHandleValue` - Linker currently tries to conform to static types - meaning that `SystemTypeValue` has static type of `Type`, while `RuntimeTypeHandleValue` is `RuntimeTypeHandle`. Otherwise they're identical.
I guess we could merge them.... this was carried over from .NET Native - and it's true that linker still holds onto the static typic pretty much always.
I'm fine keeping this as is - if we don't like it, let's make it a different change which applies to non-Nullable as well.

src/ILLink.Shared/TrimAnalysis/NullableRuntimeTypeHandleWithGenericParameterValue.cs
```diff
@@ -0,0 +1,30 @@
...
+using ILLink.Shared.DataFlow;
+using ILLink.Shared.TypeSystemProxy;
+
+namespace ILLink.Shared.TrimAnalysis
+{
+	/// <summary>
+	/// This represents a type handle of a Nullable<T> where T is an unknown value with DynamicallyAccessedMembers annotations. 
+	/// It is necessary to track the underlying type to ensure DynamicallyAccessedMembers annotations on the underlying type match the target parameters where the Nullable is used.
+	/// </summary>
+	sealed record NullableRuntimeTypeWithDamHandleValue : SingleValue
```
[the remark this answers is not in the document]
vitek-karas: Same as above - Type versus RuntimeTypeHandle. I would prefer we leave it as-is for now - and potentially change it later on for everything.

src/ILLink.Shared/TrimAnalysis/NullableRuntimeSystemTypeHandleValue.cs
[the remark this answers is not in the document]
jtschuster: This class is to `NullableSystemTypeValue` as `RuntimeTypeHandleValue` is to the `SystemTypeValue` (Nullable<T> vs typeof(Nullable<T>)). My understanding is that theoretically yes these could be merged, but it might be helpful to keep them separate to more accurately model the program being analyzed.

src/ILLink.Shared/TrimAnalysis/NullableSystemTypeValue.cs
```diff
@@ -0,0 +1,30 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics;
+using ILLink.Shared.DataFlow;
+using ILLink.Shared.TypeSystemProxy;
+
+namespace ILLink.Shared.TrimAnalysis
+{
+	/// <summary>
+	/// This represents a Nullable<T> where T is a known SystemTypeValue.
```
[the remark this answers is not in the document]
jtschuster: I think I considered that, but Vitek had a reason to leave it as `SingleValue`. I can't remember exactly why, but having it different makes sure that we handle it separately from `SystemTypeValue` instead of accidentally matching `NullableSystemTypeValue` to the `SystemTypeValue` case.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
jtschuster: Yes, there was a hole here, thank you!

src/ILLink.Shared/TrimAnalysis/NullableRuntimeTypeHandleWithGenericParameterValue.cs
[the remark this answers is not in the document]
jtschuster: When scanning IL, if we come across a generic parameter as the type parameter for Nullable (i.e. the T of Nullable<T> is the type param for the closing method), there's no way yet to get the DAM annotations on the generic parameter in the MethodBodyScanner (as far as I can tell), so we have to wrap the `GenericParameterProxy` in `RuntimeTypeHandleForGenericParameterValue`, which is just a `SingleValue`, not a `ValueWithDynamicallyAccessedMembers`.

src/ILLink.Shared/TrimAnalysis/NullableSystemTypeValue.cs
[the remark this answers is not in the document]
vitek-karas: I think it's OK to change the underlying type to `SystemTypeValue` if it's always that.
The `NullableSystemTypeValue` must be separate from `SystemTypeValue` - it does act very similarly, but it adds the "this is Nullable flag" to the type.

src/ILLink.Shared/TrimAnalysis/NullableRuntimeTypeHandleWithGenericParameterValue.cs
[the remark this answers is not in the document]
vitek-karas: We could get the annotation for a generic parameter at any time, that's not the problem. The difference is that we don't treat `RuntimeTypeHandle` values as annotatable. The reason we have a separate class structure is already described above.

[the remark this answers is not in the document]
sbomer: Thanks! So if I follow, we have:
- NullableSystemTypeValue: a Nullable<KnownType>, represented as TypeProxy
- NullableRuntimeSystemTypeHandleValue: type handle corresponding to the above, same representation

- NullableValueWithDynamicallyAccessedMembers: a Nullable<T> where T is a generic parameter, or an unknown type passed as an argument to MakeGenericType, represented as ValueWithDynamicallyAccessedMembers
- NullableRuntimeTypeWithDamHandleValue: type handle corresponding to the above, represented as a SingleValue
  - It sounds like we could change the underlying type to a ValueWithDynamicallyAccessedMembers to match, if we wrapped the generic parameter in MethodBodyScanner in a `GenericParameterValue`

Could we maybe rename these a bit to make that more clear? I would suggest:
- NullableSystemTypeValue
- RuntimeTypeHandleForNullableSystemTypeValue:
- NullableValueWithDynamicallyAccessedMembers:
- RuntimeTypeHandleForNullableValueWithDynamicallyAccessedMembers:

Which also matches the naming of RuntimeTypeHandleForGenericParameterValue. Yes, the names are long and ugly, but I think it makes the meaning more clear (I'm open to better suggestions of course!). Maybe we can abbreviate DynamicallyAccessedMembers to DAM if it helps but I would be consistent about it.

src/ILLink.Shared/TrimAnalysis/NullableRuntimeTypeHandleWithGenericParameterValue.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
sbomer, line 1: Rename file to match the class name

src/ILLink.Shared/TrimAnalysis/NullableRuntimeTypeHandleWithGenericParameterValue.cs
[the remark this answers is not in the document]
vitek-karas: Yes - that's right. And I do like the suggestions for naming changes!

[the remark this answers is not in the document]
sbomer: Sounds like we would get along :)

src/ILLink.RoslynAnalyzer/TrimAnalysis/SingleValueExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+							new GenericParameterValue ((ITypeParameterSymbol) underlyingType)),
+					// typeof(Nullable<>) 
+					TypeKind.Error => new SystemTypeValue (new TypeProxy (type)),
+					_ => new NullableSystemTypeValue (new TypeProxy (type), new TypeProxy (underlyingType!)),
+				};
+			}
+			return type.Kind switch {
+				SymbolKind.TypeParameter => new GenericParameterValue ((ITypeParameterSymbol) type),
+				SymbolKind.NamedType => new SystemTypeValue (new TypeProxy (type)),
+				SymbolKind.ArrayType => new SystemTypeValue (new TypeProxy (type.BaseType!)),
```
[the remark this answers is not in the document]
jtschuster: Yes, it is. I'll add a comment to clarify.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -574,31 +574,32 @@ GenericParameterValue genericParam
...
+							case SystemTypeValue systemTypeValue:
+								AddReturnValue (new NullableSystemTypeValue (typeValue.RepresentedType, systemTypeValue.RepresentedType));
+								break;
+							case ValueWithDynamicallyAccessedMembers damValue:
+								AddReturnValue (new NullableValueWithDynamicallyAccessedMembers (typeValue.RepresentedType, damValue));
+								break;
+							// Nullable values and array values cannot be used as generic arguments to nullables, so we don't need to worry about anything else here
+							default:
+								AddReturnValue (MultiValueLattice.Top);
+								break;
```
vitek-karas, line 598: I think we need to be a bit more careful. We should special case array and nullable and add Top to the return, but everything else should probably add the "Return value of MakeGenericType" with no annotations. Basically we can't skip over unknown values:
```C#
void Test(object[] maybetypes)
{
    Type[] types = new Type[] { maybetypes[0] as Type };  // Roundabout way to get UnknownValue - it is getting tricky to do that reliably
    Type nullable = typeof (Nullable<>).MakeGenericType(types);
    nullable.GetProperties();  // Must WARN
}
```

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -823,26 +823,27 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
-				}
+			maybeMethodReturnValue ??= GetReturnTypeWithoutModifiers (calledMethod.ReturnType).MetadataType == MetadataType.Void ? 
+				MultiValueLattice.Top : GetMethodReturnValue (calledMethodDefinition, returnValueDynamicallyAccessedMemberTypes);
+
+
+			if (maybeMethodReturnValue is not null) {
+				methodReturnValue = (MultiValue) maybeMethodReturnValue;
+			} else {
+				methodReturnValue = GetReturnTypeWithoutModifiers (calledMethod.ReturnType).MetadataType == MetadataType.Void ? 
+					MultiValueLattice.Top : GetMethodReturnValue (calledMethodDefinition, returnValueDynamicallyAccessedMemberTypes);
```
vitek-karas, line 845: This is weird - the above will make sure that `maybeMethodReturnValue` is never null (and it does it the exact same way)...

test/Mono.Linker.Tests.Cases/DataFlow/ArrayDataFlow.cs
```diff
@@ -205,6 +207,246 @@ static void TestArrayResetAfterAssignment ()
...
+				// should be treated as unknown value.
+				TestArrayResetStoreUnknownIndex ();
+				TestArrayResetGetElementOnByRefArray ();
+				TestArrayResetAfterCall ();
+				TestArrayResetAfterAssignment ();
+			}
+
+			[ExpectedWarning ("IL2062", nameof (DataFlowTypeExtensions.RequiresPublicMethods), ProducedBy = ProducedBy.Trimmer)]
+			// Bug
+			[ExpectedWarning ("IL2062", nameof (DataFlowTypeExtensions.RequiresAll), ProducedBy = ProducedBy.Trimmer)]
```
vitek-karas, line 240: Why bug? I think the intent currently is to simply not track multi dimensional arrays at all...

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
jtschuster: Thanks for pointing this out, this needs to be changed

test/Mono.Linker.Tests.Cases/DataFlow/ArrayDataFlow.cs
vitek-karas: I would not say never... but it's definitely VERY low on the list... I mean all we care about are `Type` values really - why would you put one into a multi dimensional array (I know, famous last words).
Not counting the fact that I haven't seen multi dimensional array in customer code for a long time...

src/ILLink.Shared/TrimAnalysis/NullableSystemTypeValue.cs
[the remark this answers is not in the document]
jtschuster: Now I remember we also wanted to keep SystemTypeValue sealed to make type checking faster. I think it might be best to leave it as inheriting from SingleValue.

vitek-karas (approved):
Looks good. Let's wait on Sven for another opinion.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -540,6 +557,64 @@ public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnl
...
+							// Don't warn on Nullable<T> - it will throw instead
+							case NullableValueWithDynamicallyAccessedMembers:
+								AddReturnValue (MultiValueLattice.Top);
+								break;
+							// Generic Parameters and method parameters with annotations
+							case ValueWithDynamicallyAccessedMembers damValue:
+								AddReturnValue (new NullableValueWithDynamicallyAccessedMembers (typeValue.RepresentedType, damValue));
+								break;
+							// Don't warn on Nullable<T>, it will throw instead
+							case NullableSystemTypeValue:
```
sbomer, line 604: nit: maybe put this and NullableValueWithDynamicallyAccessedMembers next to each other and have one fall through - since they aren't tracked for the same reason.

src/ILLink.Shared/TrimAnalysis/NullableSystemTypeValue.cs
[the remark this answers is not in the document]
sbomer: Sorry - what I meant was, can we make the underlyingType a `SystemTypeValue` (not that `NullableSystemTypeValue` should itself be a `SystemTypeValue`)? I think this is more consistent with how the underlying type of a `NullableValueWithDynamicallyAccessedMembers` is also a dataflow value, rather than a `TypeProxy`.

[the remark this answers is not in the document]
sbomer: Same question for RuntimeTypeHandleForNullableSystemTypeValue.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -145,16 +145,22 @@ MultiValue GetTypeValueNodeFromGenericArgument (TypeReference genericArgument)
...
+			} else if (ResolveToTypeDefinition (genericArgument) is TypeDefinition genericArgumentType) {
+				if (genericArgumentType.IsTypeOf ("System", "Nullable`1")) {
+					var innerGenericArgument = (genericArgument as IGenericInstance)?.GenericArguments.FirstOrDefault ();
+					switch (innerGenericArgument) {
+					case GenericParameter gp:
+						return new NullableValueWithDynamicallyAccessedMembers (genericArgumentType,
+							new GenericParameterValue (gp, _context.Annotations.FlowAnnotations.GetGenericParameterAnnotation (gp)));
+
+					case TypeReference underlyingType:
+						return new NullableSystemTypeValue (genericArgumentType, ResolveToTypeDefinition (underlyingType)!);
```
sbomer, line 157: Probably uncommon, but I think it's possible for `ResolveToTypeDefinition` to fail - maybe this should return an `UnknownValue` if it resolves to null?

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -349,18 +359,53 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+												break;
+											// Don't warn on Nullable<T> - it will throw instead
+											case NullableValueWithDynamicallyAccessedMembers:
+												AddReturnValue (MultiValueLattice.Top);
+												break;
+											// Generic Parameters and method parameters with annotations
+											case ValueWithDynamicallyAccessedMembers damValue:
+												AddReturnValue (new NullableValueWithDynamicallyAccessedMembers (typeValue.RepresentedType, damValue));
+												break;
+											// Don't warn on Nullable<T>, it will throw instead
```
sbomer, line 383: Same nit here, I would combine these two cases.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -822,6 +866,11 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 			}
 
 			return true;
+
+			void AddReturnValue (MultiValue value)
```
sbomer, line 870: Thanks for making this more consistent with the shared code!

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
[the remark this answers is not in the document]
sbomer: Would you mind opening an issue to track making the analyzer consistent with the linker on this?

sbomer (approved):
A few more comments, overall LGTM. Thanks a lot!

jtschuster:
> @sbomer 
> Would you mind opening an issue to track making the analyzer consistent with the linker on this?

Unless I'm misunderstanding what you're talking about, it looks like https://github.com/dotnet/linker/issues/2680 tracks the overall discrepancy between the analyzer and linker for arrays. I'll make a note about multidimensional arrays though.

PR #2681, opened 2022-03-11
Share Type.Base intrinsic

vitek-karas:


src/ILLink.RoslynAnalyzer/TrimAnalysis/HandleCallAction.cs
```diff
@@ -62,6 +62,17 @@ public HandleCallAction (in DiagnosticContext diagnosticContext, ISymbol owningS
 				yield return new SystemTypeValue (new TypeProxy (nestedType));
 		}
 
+		private partial bool TryGetBaseType (TypeProxy type, out TypeProxy baseType)
```
sbomer, line 65: I'm not sure what's best practice for value types, but I wonder if it's worth making the out param `[NotNullWhen(true)] out TypeProxy?` so that we get analysis warnings if someone tries to access the value if it returns false?

sbomer (approved):
Thank you!

PR #2682, opened 2022-03-11
Use attributes on lambdas in tests

sbomer:
This updates a few of the compiler-generated code tests to put the `ExpectedWarning` attribute on lambdas and local functions when possible, removing the need for `CompilerGeneratedCode = true` in these places.

Unfortunately the formatting became weird - dotnet format seems to require the lambda to not be indented following the attribute, and I couldn't find a good workaround so I'm leaving it as-is. (Filed https://github.com/dotnet/format/issues/1548) edit: Ok, I changed it to at least unindent the attributes as well to match the lambda expression. Otherwise multiple attributes on lambdas have inconsistent indentation.


vitek-karas (approved):
:+1:

Issue #2686, opened 2022-03-15
COMAnalyzer NullReferenceException in Maui Essentials


eerhardt:
I'm getting the following error when setting `IsTrimmable=true` for the Maui Essentials project and building with `msbuild Essentials.csproj -p:Configuration=Release -t:Rebuild /bl`:

```
    Errors
        CSC error AD0001: Analyzer 'ILLink.RoslynAnalyzer.COMAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'. [C:\git\maui\src\Essentials\src\Essentials.csproj]
    Exception occurred with following context:
    Compilation: Microsoft.Maui.Essentials
    IOperation: Invocation
    SyntaxTree: C:\git\maui\src\Essentials\src\DeviceDisplay\DeviceDisplay.uwp.cs
    SyntaxNode: EnumDisplaySettings(mi.Value.DeviceNameToLPTStr ... [InvocationExpressionSyntax]@[2465..2533) (101,3)-(101,71)
    System.NullReferenceException: Object reference not set to an instance of an object.
    at ILLink.RoslynAnalyzer.COMAnalyzer.<Initialize>g__IsComInterop|6_1(ISymbol symbol)
    at ILLink.RoslynAnalyzer.COMAnalyzer.<>c.<Initialize>b__6_2(OperationAnalysisContext operationContext)
    at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c.<ExecuteOperationAction>b__63_0(ValueTuple`2 data)
    at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info)
    -----
    Suppress the following diagnostics to disable this analyzer: IL2050
    CompilerServer: server - server processed compilation - 1477aa4a-10c1-4109-b699-b092d24b4231
```

This looks like https://github.com/dotnet/linker/issues/2622, but I believe it is a different issue because this method isn't using a pointer, and because the version of the ILLink.RoslynAnalyzer assembly I'm using is from https://github.com/dotnet/linker/commits/e9cfb5413a6a7a7b5bfc3b9a73671be2b18642cf, which has the fix for #2622.

### To Repro

1. git clone https://github.com/dotnet/maui
[...]

PR #2688, opened 2022-03-15
Fix null refs in byte array COM interop

jtschuster:
Roslyn API doesn't use nullable types and will return null on properties and fields that aren't annotated with the `?`. The linker assumes that fields will not be null, and those issues tend to appear most frequently with COM interop when a type doesn't have a namespace.

This fixes the null reference in a minimal repro, but there are likely more holes to address https://github.com/dotnet/linker/issues/2687.

This should fix https://github.com/dotnet/linker/issues/2686.

I wasn't sure where the test case best fits in, so it's in a separate file in the `Interop` test cases folder. Let me know if there's a better place for it.


test/Mono.Linker.Tests.Cases/Interop/ByteArrayCom.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+
+namespace Mono.Linker.Tests.Cases.Interop
+{
+	[SkipKeptItemsValidation]
+	class ByteArrayCom
```
vitek-karas, line 15: ExpectNoWarnings

src/ILLink.RoslynAnalyzer/COMAnalyzer.cs
```diff
@@ -88,11 +88,11 @@ static bool IsComInterop (ISymbol symbol)
 				if (typeSymbol == null)
 					return false;
 
-				if (typeSymbol.ContainingNamespace.Name == "System" && typeSymbol.Name == "Array") {
+				if (typeSymbol.ContainingNamespace?.Name == "System" && typeSymbol.Name == "Array") {
```
marek-safar, line 91: Don't we have enabled nullable to catch such errors?

vitek-karas: See https://github.com/dotnet/linker/issues/2687.

jtschuster:
The issues were fixed by https://github.com/dotnet/linker/pull/2692, this now just adds tests.

tlakollo (approved):

vitek-karas (approved):

eerhardt:
For my education, how does adding new tests fix the original issue? Where is the source code change that fixed #2686?

akoeplinger:
@eerhardt see https://github.com/dotnet/linker/pull/2688#issuecomment-1072774547

PR #2689, opened 2022-03-16
Lambda and local function suppressions

sbomer:
This adds support for `RequiresUnreferencedCode` and `UnconditionalSuppressMessage` on lambdas and local functions, relying on heuristics and knowledge of compiler implementation details to detect lambdas and local functions.

This approach scans the code for IL references to lambdas and local functions, which has some limitations.
- Unused local functions aren't referenced by the containing method, so warnings from these will not be suppressed by suppressions on the containing method. Lambdas don't seem to have this problem, because they contain a reference to the generated method as part of the delegate conversion.
- The IL doesn't in general contain enough information to determine the nesting of the scopes of lambdas and local functions, so we make no attempt to do this. We only try to determine to which user method a lambda or local function belongs. So suppressions on a lambda or local function will not silence warnings from a nested lambda or local function in the same scope.

This also adds warnings for reflection access to compiler-generated state machine members, and to lambdas or local functions. For these, the analyzer makes no attempt to determine what the actual IL corresponding to the user code will be, so it produces fewer warnings (see tests for some examples). The linker will warn for what is actually in IL.

One way to work around the first limitation would be to rely on the compiler-generated names to determine the name of the user method, but this is fragile due to the issue @vitek-karas [pointed out](https://github.com/dotnet/linker/pull/2689#pullrequestreview-911356059).

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -566,13 +566,20 @@ namespace Mono.Linker.Steps
 				// We rely on the callers of compiler-generated methods being marked first... which might not always be true,
 				// if they are marked for a reason other than a discovered call to them.
 				// Might need to postpone processing of such methods. But for now just check the assumption.
-				Debug.Assert (reason.Kind is DependencyKind.DirectCall or DependencyKind.VirtualCall or DependencyKind.Newobj);
-				Debug.Assert (Annotations.IsMarked ((IMetadataTokenProvider) reason.Source));
-				// We should have tracked a user method for this compiler-generated method.
-				Debug.Assert (Context.CompilerGeneratedState._compilerGeneratedMethodToUserCodeMethod.TryGetValue (method, out _));
 
-				MethodDefinition caller = (MethodDefinition) reason.Source;
-				Context.CompilerGeneratedState.TrackCallToLambdaOrLocalFunction (caller, method);
+				// Forget the reason assert. There are too many.
+				// Debug.Assert (reason.Kind is DependencyKind.DirectCall or DependencyKind.VirtualCall or DependencyKind.Newobj or DependencyKind.Ldftn
+				// // Could be a directcall to a generic LocalFunction... :/
+				// 	or DependencyKind.ElementMethod
+				// 	// Or could by dynamically accessed... hopefully not the first time?
+				// 	or DependencyKind.DynamicallyAccessedMember);
+				// We might mark a generic instantiation, which gets resolved to the definition, and the source
+				// is the generic instance. Which doesn't satisfy the above.
+				if (reason.Kind is not (DependencyKind.ElementMethod or DependencyKind.MethodOnGenericInstance)) {
+					Debug.Assert (Annotations.IsMarked ( (IMetadataTokenProvider) reason.Source));
+					MethodDefinition caller = (MethodDefinition) reason.Source;
+					Context.CompilerGeneratedState.TrackCallToLambdaOrLocalFunction (caller, method);
+				}
 			}
 			_methods.Enqueue ((method, reason, scope));
 		}
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -559,6 +559,21 @@ namespace Mono.Linker.Steps
 
 		protected virtual void EnqueueMethod (MethodDefinition method, in DependencyInfo reason, in MarkScopeStack.Scope scope)
 		{
+			if (GeneratedNames.IsGeneratedMemberName (method.Name)) {
+				// We only expect a generated method to represent a lambda or local function.
+				// TODO: what about implicit Main? Async main?
+				Debug.Assert (GeneratedNames.TryParseLambdaMethodName (method.Name, out _) || GeneratedNames.TryParseLocalFunctionMethodName (method.Name, out _, out _));
+				// We rely on the callers of compiler-generated methods being marked first... which might not always be true,
+				// if they are marked for a reason other than a discovered call to them.
+				// Might need to postpone processing of such methods. But for now just check the assumption.
+				Debug.Assert (reason.Kind is DependencyKind.DirectCall or DependencyKind.VirtualCall or DependencyKind.Newobj);
+				Debug.Assert (Annotations.IsMarked ((IMetadataTokenProvider) reason.Source));
+				// We should have tracked a user method for this compiler-generated method.
+				Debug.Assert (Context.CompilerGeneratedState._compilerGeneratedMethodToUserCodeMethod.TryGetValue (method, out _));
+
+				MethodDefinition caller = (MethodDefinition) reason.Source;
+				Context.CompilerGeneratedState.TrackCallToLambdaOrLocalFunction (caller, method);
+			}
 			_methods.Enqueue ((method, reason, scope));
 		}
 
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2903,7 +2903,7 @@ namespace Mono.Linker.Steps
 
 			MethodDefinition? owningMethod;
 			while (Context.CompilerGeneratedState.TryGetOwningMethodForCompilerGeneratedMember (member, out owningMethod)) {
-				Debug.Assert (owningMethod != originMember);
+				Debug.Assert (owningMethod != member);
 				if (Annotations.IsMethodInRequiresUnreferencedCodeScope (owningMethod))
 					return true;
 				member = owningMethod;
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -2901,13 +2901,15 @@ namespace Mono.Linker.Steps
 			if (originMember is not IMemberDefinition member)
 				return false;
 
-			MethodDefinition? userDefinedMethod = Context.CompilerGeneratedState.GetUserDefinedMethodForCompilerGeneratedMember (member);
-			if (userDefinedMethod == null)
-				return false;
+			MethodDefinition? owningMethod;
+			while (Context.CompilerGeneratedState.TryGetOwningMethodForCompilerGeneratedMember (member, out owningMethod)) {
+				Debug.Assert (owningMethod != originMember);
+				if (Annotations.IsMethodInRequiresUnreferencedCodeScope (owningMethod))
+					return true;
+				member = owningMethod;
+			}
 
-			Debug.Assert (userDefinedMethod != originMember);
-
-			return Annotations.IsMethodInRequiresUnreferencedCodeScope (userDefinedMethod);
+			return false;
 		}
 
 		internal void CheckAndReportRequiresUnreferencedCode (MethodDefinition method)
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -559,28 +559,6 @@ namespace Mono.Linker.Steps
 
 		protected virtual void EnqueueMethod (MethodDefinition method, in DependencyInfo reason, in MarkScopeStack.Scope scope)
 		{
-			if (GeneratedNames.IsGeneratedMemberName (method.Name)) {
-				// We only expect a generated method to represent a lambda or local function.
-				// TODO: what about implicit Main? Async main?
-				Debug.Assert (GeneratedNames.TryParseLambdaMethodName (method.Name, out _) || GeneratedNames.TryParseLocalFunctionMethodName (method.Name, out _, out _));
-				// We rely on the callers of compiler-generated methods being marked first... which might not always be true,
-				// if they are marked for a reason other than a discovered call to them.
-				// Might need to postpone processing of such methods. But for now just check the assumption.
-
-				// Forget the reason assert. There are too many.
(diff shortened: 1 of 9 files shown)


src/linker/Linker/CompilerGeneratedNames.cs
```diff
@@ -0,0 +1,70 @@
...
+		{
+			if (!IsGeneratedMemberName (className))
+				return false;
+
+			// This is true for static lambdas (which are emitted into a class like <>c)
+			// and for instance lambdas (which are emitted into a class like <>c__DisplayClass1_0)
+			return className.StartsWith ("<>c");
+		}
+
+		// Lambda methods have generated names like "<UserMethod>c__0_1" where "UserMethod" is the name
```
vitek-karas, line 22: Did we consider if this approach is robust enough?
I know we have to rely on detection of compiler generated names, but I honestly didn't expect to parse the name to get information from it this way. The current approach can't differentiate overloads, so what might happen is this:
```C#
[UnconditionalSuppressMessage("IL2026")]
void TestMethod()
{
}

void TestMethod(string p)
{
    CallLocal ();

    void CallLocal()
    {
        MethodWithRUC();  // No warning - since we would apply the suppression from TestMethod() onto local function from TestMethod(string)
    }
}

[RequiresUnreferencedCode("Don't call me")]
void MethodWithRUC() { }
```

I thought we would have to scan IL of all methods on the containing type to find the method which refers to the lambda/local in question and associate them that way. It is potentially relatively expensive, but then again it should only happen on types which use lambdas/local functions which should be a relative minority.

sbomer: This was just a quick and dirty way to hook it up to test that this would correctly identify the lambdas and local functions. Thanks for the example - you are right that it will probably need to actually look at IL references rather than the name - I'll make sure that case is tested.

vitek-karas: I was just browsing the repo... only noticed this is WIP after I posted the comment.

src/linker/Linker/Annotations.cs
```diff
@@ -606,6 +606,13 @@ internal bool DoesMethodRequireUnreferencedCode (MethodDefinition method, [NotNu
 				TryGetLinkerAttribute (method.DeclaringType, out attribute))
 				return true;
 
+			MethodDefinition? owningMethod;
+			while (context.CompilerGeneratedState.TryGetOwningMethodForCompilerGeneratedMember (method, out owningMethod)) {
+				if (DoesMethodRequireUnreferencedCode (owningMethod, out attribute))
+					return true;
+				method = owningMethod;
+			}
```
jtschuster, line 614: Aren't the while loop and recursion redundant here? Why don't we only include the recursion:
```C#
if (DoesMethodRequireUnreferencedCode (owningMethod, out attribute))
	return true;
```

or make the whole function a do...while loop like:
```C#
do 
{
	if (method.IsStaticConstructor ()) {
		attribute = null;
		return false;
	}
	if (TryGetLinkerAttribute (method, out attribute))
		return true;
	if ((method.IsStatic || method.IsConstructor) && method.DeclaringType is not null &&
		TryGetLinkerAttribute (method.DeclaringType, out attribute))
		return true;
} while (context.CompilerGeneratedState.TryGetOwningMethodForCompilerGeneratedMember (method, out method))
```

src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -64,27 +139,43 @@ void PopulateCacheForType (TypeDefinition type)
...
-				return null;
+			TypeDefinition sourceType = (sourceMember as TypeDefinition) ?? sourceMember.DeclaringType;
+
+			if (_compilerGeneratedTypeToUserCodeMethod.TryGetValue (sourceType, out owningMethod))
+				return true;
+
+			if (sourceType.DeclaringType == null)
+				return false;
+
+			var typeToCache = HasRoslynCompilerGeneratedName (sourceType) ? sourceType.DeclaringType : sourceType;
```
jtschuster, line 165: If we need to check two levels deep in `PopulateCacheForType.ProcessMethod`, should this check if `sourceType.DeclaringType` is a compiler generated name and go up one more level if so?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -55,13 +55,15 @@ public bool IsSuppressed (int id, MessageOrigin warningOrigin, out SuppressMessa
...
-				return false;
-
-			Debug.Assert (userDefinedMethod != provider);
+			MethodDefinition? owningMethod;
+			while (_context.CompilerGeneratedState.TryGetOwningMethodForCompilerGeneratedMember (member, out owningMethod)) {
+				Debug.Assert (owningMethod != provider);
+				if (IsSuppressed (id, owningMethod, out info))
+					return true;
+				member = owningMethod;
+			}
```
jtschuster, line 64: Same thing here as above, are the while loop and recursion redundant here?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2901,13 +2901,15 @@ internal bool ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode ()
...
-				return false;
-
-			Debug.Assert (userDefinedMethod != originMember);
+			MethodDefinition? owningMethod;
+			while (Context.CompilerGeneratedState.TryGetOwningMethodForCompilerGeneratedMember (member, out owningMethod)) {
+				Debug.Assert (owningMethod != originMember);
+				if (Annotations.IsMethodInRequiresUnreferencedCodeScope (owningMethod))
+					return true;
+				member = owningMethod;
+			}
```
jtschuster, line 2910: Is there a reason we don't want this while loop in `IsMethodInRequiresUnreferencedCodeScope`?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
sbomer: This one isn't actually recursive, it's calling a different method with the same name.

src/linker/Linker/CompilerGeneratedState.cs
sbomer: I think you are right, thanks! I couldn't find a good way to write a test for this because all the cases I can think of happen to cache the correct type before we get here. So I adjusted the logic and added an assert, hopefully this is good enough.

This also caused me to realize I had removed the check that only built a cache for compiler-generated types (it was actually building a cache for all types... oops).

src/linker/Linker/Annotations.cs
sbomer: Good catch, the recursion was doing unnecessary work. I took your suggestion and removed the recursion here and for the corresponding code in the analyzer.

src/linker/Linker.Steps/MarkStep.cs
sbomer: I considered it, but  think eventually we might want to pull this out so that it applies not only to methods but also to fields and other members - so I left it here for now, and when I add dataflow for nested functions I'll see what makes more sense.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -55,13 +55,15 @@ public bool IsSuppressed (int id, MessageOrigin warningOrigin, out SuppressMessa
...
 				return false;
 
-			MethodDefinition? userDefinedMethod = _context.CompilerGeneratedState.GetUserDefinedMethodForCompilerGeneratedMember (member);
-			if (userDefinedMethod == null)
-				return false;
-
-			Debug.Assert (userDefinedMethod != provider);
+			MethodDefinition? owningMethod;
+			while (_context.CompilerGeneratedState.TryGetOwningMethodForCompilerGeneratedMember (member, out owningMethod)) {
+				Debug.Assert (owningMethod != provider);
```
vitek-karas, line 60: Maybe this should be `owningMethod != member`, but it probably doesn't matter much.

src/linker/Linker/CallGraph.cs
```diff
@@ -0,0 +1,43 @@
...
+		}
+
+		public IEnumerable<MethodDefinition> GetReachableMethods (MethodDefinition start)
+		{
+			Queue<MethodDefinition> queue = new ();
+			HashSet<MethodDefinition> visited = new ();
+			visited.Add (start);
+			queue.Enqueue (start);
+			while (queue.TryDequeue (out MethodDefinition? method)) {
+				if (!callGraph.TryGetValue (method, out HashSet<MethodDefinition>? neighbors))
```
vitek-karas, line 31: Nit: I find `neighbors` confusing - maybe `callees` or `children`... for some reason I think about the callgraph as a tree mostly. Nobig deal.

src/linker/Linker/CompilerGeneratedNames.cs
```diff
@@ -0,0 +1,55 @@
...
+
+			// This is true for static lambdas (which are emitted into a class like <>c)
+			// and for instance lambdas (which are emitted into a class like <>c__DisplayClass1_0)
+			return className.StartsWith ("<>c");
+		}
+
+		internal static bool IsLambdaOrLocalFunction (string methodName) => IsLambdaMethod (methodName) || IsLocalFunction (methodName);
+
+		// Lambda methods have generated names like "<UserMethod>c__0_1" where "UserMethod" is the name
+		// of the original user code that contains the lambda method declaration.
```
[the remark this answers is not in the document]
sbomer: Thanks, it had a typo and was outdated.

src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -12,27 +15,54 @@ public class CompilerGeneratedState
...
+				if (!CompilerGeneratedNames.IsLambdaOrLocalFunction (method.Name)) {
+					// If it's not a nested function, track as an entry point to the call graph.
+					var added = callingMethods.Add (method);
+					Debug.Assert (added);
+				}
+
+				// Discover calls to lambdas or local functions.
+				if (method.Body != null) {
+					foreach (var instruction in method.Body.Instructions) {
+						if (instruction.OpCode.OperandType != OperandType.InlineMethod)
```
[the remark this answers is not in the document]
sbomer: Yeah, it covers `ldftn` which includes for example the case where a lambda expression is assigned to a variable but is unused. I clarified the comment.

src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -43,17 +73,59 @@ void PopulateCacheForType (TypeDefinition type)
...
+			// Also scan compiler-generated state machine methods (in case they have calls to nested functions),
+			// and nested functions inside compiler-generated closures (in case they call other nested functions).
+			foreach (var nestedType in type.NestedTypes) {
+				if (!CompilerGeneratedNames.IsGeneratedMemberName (nestedType.Name))
+					continue;
+
+				foreach (var method in nestedType.Methods)
+					ProcessMethod (method);
+
+				// State machines can be emitted into lambda display classes, so we need to go down one
```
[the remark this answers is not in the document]
sbomer: Good idea, I made it recursive instead and mentioned in a comment that we recurse to avoid depending on implementation details.

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
```diff
@@ -1121,12 +1254,98 @@ static void TestGenericTypeParameterRequirement<TUnknown> ()
...
+				var lambda =
+				[RequiresUnreferencedCode ("Suppress in body")]
+				[RequiresAssemblyFiles ("Suppress in body")]
+				[RequiresDynamicCode ("Suppress in body")]
+				() => {
+					// The linker doesn't have enough information to associate the RUC on lambda
+					// with this nested lambda.
+					var nestedLambda =
+					[ExpectedWarning ("IL2026", ProducedBy = ProducedBy.Trimmer)]
+					() => MethodWithRequires ();
```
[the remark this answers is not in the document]
sbomer: The issue is that with local functions the IL reference (if there is one at all) doesn't tell us where it is defined. So `M1` and `M2` produce effectively the same [IL](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYAs6AsnABQCU+hxjUZA8gK4AuApgE631JGQ4gEkAdmN60A3A2GM58os3TjJfOgUFLiAX0WN92okYXHSFGPy1KV7bhoE7REqTVnnDB4t+Vk1bk7ypnrIukA=):
```csharp
public class C {
    void M1() {
        void Outer() {
            Inner();
            
            void Inner() {
            }
        }
    }
    
    void M2() {
        void Outer() {
            Inner();
        }
        
        void Inner() {
        }
    }
}
```

It might be possible to do better for cases where they capture enclosing state - there we could potentially infer the nesting from what got captured - but this would mean more reliance on compiler implementation details.

This is only really an issue for local functions though. You're right that we could do better for lambdas since the IL will always have an actual reference to the generated lambda method at the point where it is defined (I'll update the comment to reflect this). It's debatable whether we should implement better understanding of nested lambdas, or leave lambdas with the same limitation for consistency with local functions, what do you think?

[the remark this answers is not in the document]
vitek-karas: I didn't realize that with local functions you can have them "cross scope" like this. You're right, this is probably not easily solvable. I think in that case, we just need to make sure that we don't omit warnings which should be there. I don't think we will with the current implementation (in this PR), but maybe give that a thought. It's better if we produce warnings which should not be there, than the other way round.

PR #2691, opened 2022-03-17
Remove SuppressionContextMember

sbomer:
Instead of computing the suppression context member when we push to the scope stack, it is now computed on demand when we need to know whether a warning is suppressed. The suppression context should be entirely determined by the static scopes, so there isn't any need to track it dynamically.

This will make it easier to add support for suppressions on local functions and lambdas, where we need to walk up multiple enclosing lambdas or local functions, each of which could be a suppression context.


vitek-karas (approved):
For some reason I'm nervous about this change... but I can't find anything wrong with it.
Looks good.

PR #2692, opened 2022-03-17
Create shared WellKnownTypes enum

jtschuster:
Create an enum in the shared project for well known types to simplify type checking, avoid using string comparisons when possible, and avoid dereferencing null properties in Roslyn APIs.


tlakollo (approved):
Looks good to me

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
```diff
@@ -95,7 +95,7 @@ public static string GetDisplayName (this ISymbol symbol)
 			var sb = new StringBuilder ();
 			switch (symbol) {
 			case IFieldSymbol fieldSymbol:
-				sb.Append (fieldSymbol.ContainingSymbol.ToDisplayString (ILLinkTypeDisplayFormat));
+				sb.Append (fieldSymbol.ContainingSymbol?.ToDisplayString (ILLinkTypeDisplayFormat));
```
sbomer, line 98: I'd prefer not to add `?` unless we actually have some kind of reasonable handling for the null case, and this would produce a weird-looking string if the containing symbol is null. Do we know of a way that a field symbol can have a null ContainingSymbol?

I'd be fine with just leaving out the `Append (".")` in that case. Similar for the cases below.

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -358,6 +359,12 @@ public static bool IsTypeOf<T> (this TypeReference tr)
 			return tr.Name == type.Name && tr.Namespace == tr.Namespace;
 		}
 
+		public static bool IsTypeOf (this TypeReference tr, WellKnownType type)
+		{
+			var (Namespace, Name) = type.GetNamespaceAndName ();
```
sbomer, line 364: nit: use lower-case variable names.

sbomer (approved):

src/ILLink.RoslynAnalyzer/ISymbolExtensions.cs
jtschuster: Good point, most of those cases we can assume the ContainingSymbol/ContainingType will not be null.

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -358,6 +359,12 @@ public static bool IsTypeOf<T> (this TypeReference tr)
 			return tr.Name == type.Name && tr.Namespace == tr.Namespace;
 		}
 
+		public static bool IsTypeOf (this TypeReference tr, WellKnownType type)
+		{
+			var (@namespace, name) = type.GetNamespaceAndName ();
+			return tr.IsTypeOf (@namespace, name);
```
vitek-karas, line 365: Cecil has a similar concept to SpecialType - it's called MetadataType and through it one can detect primitive types (and couple of other cases) faster than string compares.

For example https://github.com/dotnet/linker/blob/ae1f280e2f58d0007a4a0acbb254b06e17da9b98/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs#L505 does a boolean check.

vitek-karas (approved):
Looks good - I'd like to employ the MetadataType as well, but that can be done in a subsequent change.

src/linker/Linker/TypeReferenceExtensions.cs
jtschuster: Good to know, I'll add that in a new PR. Thanks!

PR #2659, opened 2022-02-25
Remove special logic for TypeConverterAttribute

jtschuster:
Fixes https://github.com/dotnet/linker/issues/1942

We no longer need special logic for `TypeConverterAttribute`s. This removes that code, and adds a test to reproduce the issue found in https://github.com/dotnet/linker/issues/1942.

PR #2695, opened 2022-03-18
[release/6.0.2xx] Remove special logic for TypeConverterAttribute (#2659)

agocke:
(cherry picked from commit b10e1bc3da22862b4ddedc2b50639dee9d8dacfd)

# Summary

The linker has had a special case for the TypeConverterAttribute for a while. In .NET 6, the logic was properly annotated, and the special casing can actually produce spurious warnings. This change removes the special casing.

# Impact

Consumers like MAUI use TypeConverterAttribute and are seeing lots of spurious warnings, which is harming their ability to make things trimmable for .NET 6.

# Risk

Low. Removes special casing.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -819,7 +819,6 @@ namespace Mono.Linker.Steps
 						continue;
 
 					MarkCustomAttribute (ca, reason);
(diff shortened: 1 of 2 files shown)


jeffschwMSFT (approved):
Approved.  We should take for consideration in 6.0.2xx.

PR #2699, opened 2022-03-21
Use WellKnownType in more places and add mapping to MetadataType

jtschuster:
Use of WellKnownType enum where possible.

Add mapping from WellKnownType to Cecil's MetadataType for quicker checking in IsTypeOf.

Add wrapper method for BCL.FindPredfinedType that takes a WellKnownType.


src/ILLink.RoslynAnalyzer/ITypeSymbolExtensions.cs
```diff
@@ -61,9 +61,11 @@ public static bool IsTypeOf (this ITypeSymbol symbol, string @namespace, string
 		public static bool IsTypeOf (this ITypeSymbol symbol, WellKnownType wellKnownType)
 		{
 			if (wellKnownType.TryGetSpecialType (out var specialType)) {
+				// Nullables are SpecialType.None, but the originalDefinition is SpecialType.Nullable_T
+				var symbolSpecialType = symbol.SpecialType == SpecialType.None ? symbol.OriginalDefinition.SpecialType : symbol.SpecialType;
```
agocke, line 65: Substituted types are never special types, so rather than try to fix this up, I would make this a part of the API, meaning that the API expects to always receive an original definition. You could assert this by saying

```
Debug.Assert(symbol.Equals(symbol.OriginalDefinition, SymbolEqualityComparer.Default));
```

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -361,6 +361,11 @@ public static bool IsTypeOf<T> (this TypeReference tr)
 
 		public static bool IsTypeOf (this TypeReference tr, WellKnownType type)
 		{
+			// TypeReferences of System.Array doesn't have metadata type of MetadataType.Array
+			if (type.TryGetMetadataType (out var metadataType)) {
+				Debug.Assert (tr.IsTypeOf (type.GetNamespace (), type.GetName ()) == (metadataType == tr.MetadataType));
+				return metadataType == tr.MetadataType;
+			}
 			var (@namespace, name) = type.GetNamespaceAndName ();
 			return tr.IsTypeOf (@namespace, name);
```
marek-safar, line 370: Can this be a simple switch instead of this please?

src/ILLink.RoslynAnalyzer/ITypeSymbolExtensions.cs
jtschuster: If we always want the SpecialType of the OriginalDefinition, would it make more sense to do something like the following? This way the caller doesn't need to remember to use the OriginalDefinition.
```C#
var symbolSpecialType = symbol.OriginalDefinition.SpecialType;
```

src/linker/Linker/TypeReferenceExtensions.cs
jtschuster: I'm not sure I understand what part here can be turned into a switch. Can you explain what you mean?

src/ILLink.RoslynAnalyzer/ITypeSymbolExtensions.cs
agocke: That would always work in getting the well-known type, but I'm not sure if it would create a trap. That is, at the call-site, does the code then use the substituted type or the original definition? If the code requires the original definition to work properly, it's probably best to just require its use, and catch any logic errors early.

src/ILLink.Shared/TypeSystemProxy/WellKnownType.cs
```diff
@@ -28,10 +31,32 @@ public static (string Namespace, string Name) GetNamespaceAndName (this WellKnow
...
+		{
+			return @namespace switch {
+				"System" => name switch {
+					"String" => WellKnownType.System_String,
+					"Nullable`1" => WellKnownType.System_Nullable_T,
+					"Type" => WellKnownType.System_Type,
+					"Array" => WellKnownType.System_Array,
+					"Attribute" => WellKnownType.System_Attribute,
+					"Object" => WellKnownType.System_Object,
+					_ => null
```
tlakollo, line 52: I think this is missing the new values you added:
- System_NotSupportedException 
- System_Runtime_CompilerServices_DisablePrivateReflectionAttribute 
- System_Void

tlakollo: Nevermind, I saw that is not used by TryGetWellKnownType and only used by FindPredefinedType

tlakollo (approved):
Looks good to me

vitek-karas (approved):

src/linker/Linker/TypeReferenceExtensions.cs
```diff
@@ -364,8 +364,17 @@ public static bool IsTypeOf<T> (this TypeReference tr)
...
-			var (@namespace, name) = type.GetNamespaceAndName ();
-			return tr.IsTypeOf (@namespace, name);
+			return tr.TryGetWellKnownType () == type;
+		}
+
+		public static WellKnownType? TryGetWellKnownType (this TypeReference tr)
+		{
+			return tr.MetadataType switch {
+				MetadataType.String => WellKnownType.System_String,
+				MetadataType.Object => WellKnownType.System_Object,
```
marek-safar, line 374: MetadataType.Void could be included too

marek-safar (approved):

PR #2700, opened 2022-03-21
Inline simple wrapper functions in TypeProxy

jtschuster:
To limit the overhead of the wrapper functions in TypeProxy, we could inline the methods that redirect to a method on the `Type` field.


sbomer (commented):
In general I'd prefer to withhold perf tweaks like this until we collect some data and can see that this kind of change actually makes a difference. For this change, I'm not familiar with the JIT but I would kind of expect these methods to be inlined already, without the `AggressiveInlining`. That's what it looks like to me here: https://godbolt.org/z/x56xP3zWT

marek-safar:
I agree with Sven. I don't see reasons why this should be hand optimized.

PR #2701, opened 2022-03-21
Avoid running GetDisplayNameTests in parallel

### LINKED ISSUE #2693: TestGetDisplayName is flaky
TestGetDisplayName on Linux sometimes fails randomly, then will pass on a retry. It seems to be with formatting generic parameters.

tlakollo:
There seems to be a race condition on Cecil while trying to get GenericParameters, if a multithreaded test asks simultaneously about GenericParameters on the same type Cecil answers could vary creating test failures. 
Fixes #2693

### DIFF
--- a/test/Mono.Linker.Tests/Tests/GetDisplayNameTests.cs
+++ b/test/Mono.Linker.Tests/Tests/GetDisplayNameTests.cs
@@ -8,6 +8,7 @@ using NUnit.Framework;
 
 namespace Mono.Linker.Tests
 {
+	[NonParallelizable]
 	[TestFixture]
 	public class GetDisplayNameTests
 	{


sbomer (approved):
Thank you!

marek-safar:
Do you know which Cecil piece is causing this?

tlakollo:
Given that the problematic code in linker is https://github.com/dotnet/linker/blob/main/src/linker/Linker/TypeReferenceExtensions.cs#L43-L58
We could assume that is likely caused because the count of `genericParametersCount` or `declaringTypeGenericParametersCount` was incorrect. Causing the code to go to the else statement and print the incorrect type.
That would make https://github.com/mono/cecil/blob/main/Mono.Cecil/IGenericParameterProvider.cs#L46-L56 in Cecil the possible cause. Given that is likely that the problem will go away while debugging the bug would need to be found only by code inspection. Spending a few min looking at the code there, it seems to be like it should run fine, so it might be something subtle.

PR #2704, opened 2022-03-23
Update array data flow test

jtschuster:
It looks like there was a mistake on the tests for array tracking. Passing an array element as a ref parameter should reset that element, but not others. Previously it expected that it would not reset the known index.


sbomer (approved):
Nice catch!

PR #2705, opened 2022-03-23
Don't let RUC on cctor silence warnings

sbomer:
We don't allow RUC on static constructors, and putting one there produces a warning. The current behavior happens to silence warnings from the cctor but it seems more consistent for it not to have any effect on the warnings since it isn't supported.


tlakollo (approved):
Looks good to me, thanks :D

PR #2707, opened 2022-03-25
Move parameter annotation checks to shared code for analyzer

jtschuster:
Moves the logic for checking argument values against the annotations of the corresponding parameters to the shared code. The linker does a little bit more handling on non-intrinsics, so the code will still only be reached by the analyzer.

I am working on a separate PR to fix where the analyzer sees a foreach loop variable as the result of IEnumerable.Current.get. I'm not yet sure if it would be best to make a new intrinsic for IEnumerable.Current.get, or make a new Visit method on the OperationWalker for foreach loops. Let me know if one seems clearly better than the other.


src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -803,8 +814,12 @@ internal static bool BindingFlagsAreUnsupported (BindingFlags? bindingFlags)
 
 		private partial GenericParameterValue GetGenericParameterValue (GenericParameterProxy genericParameter);
 
+		private partial DynamicallyAccessedMemberTypes GetMethodThisParameterAnnotation (MethodProxy method);
```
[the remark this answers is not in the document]
jtschuster: Good point, I added a `GetMethodThisParameterValue (MethodProxy)` and used that instead of the `MethodThisParameterValue` ctor

vitek-karas (approved):

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -666,8 +666,16 @@ GenericParameterValue genericParam
...
+				// Verify the argument values match the annotations on the parameter definition
+				if (requiresDataFlowAnalysis) {
+					if (!calledMethod.IsStatic ()) {
+						_requireDynamicallyAccessedMembersAction.Invoke (instanceValue, GetMethodThisParameterValue (calledMethod));
+					}
+					for (int argumentIndex = 0; argumentIndex < argumentValues.Count; argumentIndex++) {
+						_requireDynamicallyAccessedMembersAction.Invoke (argumentValues[argumentIndex], GetMethodParameterValue (calledMethod, argumentIndex));
+					}
+				}
+				break;
```
sbomer, line 678: Now that we always return `true`, we can stock handling the `IntrinsicId.None` case when this is called from the analyzer in `TrimAnalysisVisitor`.

sbomer: I would also prefer if we could use the new shared logic for `IntrinsicId.None` on the linker side. The linker has some extra handling in the `default` case that we would need to separate out, but if we are sharing this code we should remove the corresponding code from linker/analyzer if possible.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -803,10 +811,20 @@ internal static bool BindingFlagsAreUnsupported (BindingFlags? bindingFlags)
 
 		private partial GenericParameterValue GetGenericParameterValue (GenericParameterProxy genericParameter);
 
+		private partial DynamicallyAccessedMemberTypes GetMethodThisParameterAnnotation (MethodProxy method);
+
 		private partial MethodThisParameterValue GetMethodThisParameterValue (MethodProxy method, DynamicallyAccessedMemberTypes dynamicallyAccessedMemberTypes);
 
+		private MethodThisParameterValue GetMethodThisParameterValue (MethodProxy method)
+			=> GetMethodThisParameterValue (method, GetMethodThisParameterAnnotation (method));
```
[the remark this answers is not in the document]
sbomer: Agreed, if `GetMethodThisParameterAnnotation` is unused from shared code except to implement a different shared helper, I'd prefer the declaration of `GetMethodThisParameterAnnotation` not to be shared.

sbomer (commented):
> I'm not yet sure if it would be best to make a new intrinsic for IEnumerable.Current.get, or make a new Visit method on the OperationWalker for foreach loops.

Making it a Visit method is likely to be difficult because the loop will be represented in the CFG as basic blocks with a branch to the beginning - it probably won't show up as a loop operation in the visitor.

Not sure exactly how it might look, but I think the fix would be similar to an intrinsic (it would special-case Current.get), but I wouldn't actually treat it as an intrinsic in code since it's an analyzer implementation detail.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
vitek-karas: BTW: The method still returns `false` in some cases, like for example for `Array.Empty<T>` since we don't have a good way (yet) to handle it in shared code (we don't have a way to handle generic instantiations to be more precise).

src/linker/Linker.Dataflow/MethodProxy.cs
```diff
@@ -32,7 +32,18 @@ namespace ILLink.Shared.TypeSystemProxy
 
 		internal partial bool IsStatic () => Method.IsStatic;
 
-		internal partial bool ReturnsVoid () => Method.ReturnType.MetadataType == MetadataType.Void;
+		/// <summary>
+		/// Uses the logic from <see cref="MethodBodyScanner.GetReturnTypeWithoutModifiers(TypeReference)"/> to get the
+		/// return type and determine if it is void
```
vitek-karas, line 37: Why not call that method directly?
We could either make it internal on the `MethodBodyScanner` or we could move it and make it a` TypeReference` extension.

test/Mono.Linker.Tests.Cases/DataFlow/TypeBaseTypeDataFlow.cs
```diff
@@ -74,7 +74,7 @@ static void TestAllPropagated ([DynamicallyAccessedMembers (DynamicallyAccessedM
 			// so the analyzer doesn't handle this case at all - meaning it sees it as empty value.
 			// Unlike linker which sees an unknown value and thus warns that it doesn't fulfill the All annotation.
 			// It's OK for the analyzer to be more forgiving in this case, due to the very special circumstances.
```
sbomer, line 76: nit: update the comment since analyzer now behaves the same

sbomer (approved):
LGTM, thank you!

PR #2708, opened 2022-03-25
[release/6.0.x] Update cecil to the same version as main as of 3/25/2022

vitek-karas:
## Customer Impact

Fixes https://github.com/dotnet/linker/issues/2203

The version of cecil in the 6.0.x SDK does not properly update the hashes of binaries when they are modified, resulting in colliding PDB hashes, and inability to fetch symbols for binaries. This affects many binaries in 6.0.3 and makes them undebugable in VS using the standard symbol servers.

## Testing

Unit tests. The same fix has also been in the linker/main for 2 months now and it shipped in 7.0 preview without any issues.

## Risk

Low-medium. Cecil has had very few modifications since 6.0, mostly just this fix. This was a targeted fix, but it does involve a change to Cecil. No reports of problems with 7.0 preview provides some confidence.


agocke (approved):

akoeplinger:
@vitek-karas this will break the assembly deduplication on Android, let's talk about what  options we have tomorrow.

PR #2711, opened 2022-03-28
Don't remove unused interfaces in library mode

jtschuster:
@vitek-karas's changes in the linker appears to fix the library interface trimming issue without any additions. I've just added another tests with behavior closer to the issue.

After linking, `IPAddressInformationCollection` retains the `ICollection` interface and the properties and methods:
image

Fixes https://github.com/dotnet/linker/issues/2238


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -581,11 +581,27 @@ void ProcessMarkedTypesWithInterfaces ()
 			foreach ((var type, var scope) in typesWithInterfaces) {
 				// Exception, types that have not been flagged as instantiated yet.  These types may not need their interfaces even if the
 				// interface type is marked
-				if (!Annotations.IsInstantiated (type) && !Annotations.IsRelevantToVariantCasting (type))
+				// UnusedInterfaces optimization is turned off mark all interface implementations
+				bool unusedInterfacesOptimizationEnabled = Context.IsOptimizationEnabled (CodeOptimizations.UnusedInterfaces, type);
+				if (!Annotations.IsInstantiated (type) && !Annotations.IsRelevantToVariantCasting (type) &&
+					unusedInterfacesOptimizationEnabled)
```
sbomer, line 587: It looks like this is marking the `.interfaceimpl` unconditionally. Is that what we want? It seems like it could result in keeping the impl for private interfaces, which isn't strictly necessary for library mode - it should be ok to remove impls of unused private interfaces.

If we want to keep all impls for simplicity, it's worth clarifying whether the intention is to keep all interface methods as well. My gut reaction is that it would be confusing to keep all impls but still allow trimming unused interface methods.

vitek-karas: You're right - this didn't consider private interfaces.

vitek-karas:
We might need to modify the behavior of this change. I think the best description of the desired behavior for library mode is:
* Public types must keep all public interfaces even if they're not instantiated. Not doing this means that we remove parts of the public API effectively. Library mode needs to guarantee that it's not going to change the public API of the assembly in any way.

The change doesn't try to implement exactly that, instead it tries to define and implement what is the correct behavior if the `UnusedInterfaces` optimization is disabled. It's questionable if these two should mean the same thing. If they're not the same we would have either add a new optimization or modify how rooting is done in library mode.

What library mode already does:
* All public types are marked
  * Note that this includes public interface types
* All public methods on these types are marked
  * Note that this means that public interfaces are effectively entirely preserved

What we could do is change the linker such that if `UnusedInterfaces` optimization is disabled we will on all marked types keep all interface impls for marked interfaces - always (so regardless if the type is instantiated or not).

This would mean that since we already mark all public types and interfaces, we will preserve all public interfaces on all public types. It also means that the public interfaces will keep all their methods and thus we should keep all the method impls of such interfaces.

The remaining question is what is the desired behavior of `UnusedInterfaces=false` for a type which implements private interface and the interface is not used anywhere. Based on the name of the optimization, I would prefer we simply keep all interfaces of a marked type (public or private). Such behavior would not be necessary for library mode though - it would be effectively overmarking slightly. The question is if we should then rename the optimization to better fit the library mode requirements or "relax" library mode trimming (or add a new optimization which fits the library mode).

Personally I would keep this simple and "relax" the library mode trimming - the size impact of this should be relatively small, but we need to measure for sure. But my preference is almost purely based on the simplicity of the solution - with the assumption that size impact is negligible.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -581,11 +581,27 @@ void ProcessMarkedTypesWithInterfaces ()
...
 					continue;
 
-				using (ScopeStack.PushScope (scope))
+				using (ScopeStack.PushScope (scope)) {
 					MarkInterfaceImplementations (type);
+
+					if (unusedInterfacesOptimizationEnabled)
+						continue;
+
+					// If the optimization is disabled, make sure to mark all methods which implement interfaces
```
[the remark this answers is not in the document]
vitek-karas: You might be right - I must admit that I still don't fully grasp the exact order in which things happen in this case.
I'm pretty sure the comment was there because that was what I wanted to do - but then I realized that I don't need that actually since ProcessOverride will do it for me. So I simplified the code. It's possible it can be entirely removed.

What the code does currently is that it will mark methods which are overrides/impls of methods from non-link assemblies (in library mode that's basically anything outside of the current assembly). I'm not sure if we can remove that fully - it depends if in library mode we mark all of the non-link assemblies, I'm not sure we do.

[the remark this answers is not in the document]
sbomer: I'm not 100% sure what we currently do for "library mode" since we never really came up with a design for it, but I think you're right that we don't mark assemblies other than the library. That's also what would make sense to me.

So I guess `ProcessOverride` won't be called for the case of a non-virtual method implementing an interface method defined in a referenced assembly - then I can see why we would need this logic. I wonder if there's a way we could simplify this, since it's very similar to `IsVirtualNeededByTypeDueToPreservedScope`. It feels like we want to do substantially the same thing for virtual method overrides and for interface method implementations.

Maybe that could be generalized to `IsMethodNeededByTypeDueToPreservedScope` so that it works for non-virtual methods, and methods implementing interfaces.

vitek-karas:
Discussed this with Sven offline: we would like to keep this as simple as possible - so we should try to go with the "keep all interfaces on marked types" solution (basically fix the behavior of `UnusedInterfaces=false`). Measure the size impact on framework and then decide if it's good enough.

test/Mono.Linker.Tests.Cases/Libraries/RootLibrary.cs
```diff
@@ -8,6 +12,7 @@ namespace Mono.Linker.Tests.Cases.Libraries
 {
 	[SetupLinkerArgument ("-a", "test.exe", "library")]
 	[SetupLinkerArgument ("--enable-opt", "ipconstprop")]
+	[SetupLinkerArgument ("--skip-unresolved")]
```
marek-safar, line 15: Could you come up with a test that does not require this extra setting?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2309,7 +2325,7 @@ bool IsVirtualNeededByTypeDueToPreservedScope (MethodDefinition method)
 			return false;
 		}
 
-		bool IsVirtualNeededByInstantiatedTypeDueToPreservedScope (MethodDefinition method)
+		bool IsMethodNeededByTypeDueToPreservedScope (MethodDefinition method)
 		{
 			if (!method.IsVirtual)
```
vitek-karas, line 2330: This should be something like `if (!(method.IsVirtual || method.IsStatic))` - it needs to let static methods through as some may have a "base method" in which case we may need to mark them.

jtschuster:
Finally got managed to get sizes comparing this branch vs main on runtime that I am fairly confident in.

| path | FixLibraryInterfaces Size | main Size | difference |
| --- | --- | --- | --- |
| artifacts/bin | 4000565435 | 4000553307 | +12128 | 
| artifacts/packages/Release | 860664283 | 860665167 | -884 |

I have no idea why the packages would be smaller after the fix, and I can drill down into this further, but just not 100% sure which outputs are trimmed or not.

vitek-karas:
The `testhost` is the interesting measurement (or the nupkg but that's compressed) - so the impact is minimal (+2K). Can you please also measure this on Windows build? (I assume this is a Release build, right?)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1931,7 +1922,7 @@ internal void MarkStaticConstructorVisibleToReflection (TypeDefinition type, in
 			MarkGenericParameterProvider (type);
 
 			// There are a number of markings we can defer until later when we know it's possible a reference type could be instantiated
-			// For example, if no instance of a type exist, then we don't need to mark the interfaces on that type
+			// For example, if no instance of a type exist, then we don't need to mark the interfaces on that type -- Note this is not true for static interfaces
```
vitek-karas, line 1925: While the added comment is correct, I think it makes this whole section confusing. Maybe you should add that the `MarkRequiermentsForInstantiatedTypes` also works for static interfaces - and maybe we should rename that method then.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2325,9 +2321,13 @@ bool IsVirtualNeededByTypeDueToPreservedScope (MethodDefinition method)
...
 
-		bool IsMethodNeededByTypeDueToPreservedScope (MethodDefinition method)
+		/// <summary>
+		/// Returns true if any of the base methods of <paramref name="method" /> is defined in an assembly that is not trimmed (i.e. action!=trim)
+		/// </summary>
+		/// <remarks>This is very similar to <see cref="IsVirtualNeededByTypeDueToPreservedScope(MethodDefinition)"/>, except this also checks if the base method is an interface.</remarks>
+		bool IsVirtualNeededByInstantiatedTypeDueToPreservedScope (MethodDefinition method)
 		{
-			if (!method.IsVirtual)
+			if (!(method.IsVirtual || method.IsStatic))
```
vitek-karas, line 2330: Maybe add a comment that this handles static interface methods (it's not obvious why statics are interesting here otherwise).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2325,9 +2321,13 @@ bool IsVirtualNeededByTypeDueToPreservedScope (MethodDefinition method)
 			return false;
 		}
 
-		bool IsMethodNeededByTypeDueToPreservedScope (MethodDefinition method)
+		/// <summary>
+		/// Returns true if any of the base methods of <paramref name="method" /> is defined in an assembly that is not trimmed (i.e. action!=trim)
+		/// </summary>
+		/// <remarks>This is very similar to <see cref="IsVirtualNeededByTypeDueToPreservedScope(MethodDefinition)"/>, except this also checks if the base method is an interface.</remarks>
+		bool IsVirtualNeededByInstantiatedTypeDueToPreservedScope (MethodDefinition method)
```
vitek-karas, line 2328: The new name is confusing - since this method explicitly handles interface methods as well (not just virtuals). I personally don't mind the previous method, it's not super descriptive, but at least it's not wrong.

vitek-karas: Actually with the removal of the call to this method from the delayed processing, this method should not handle static interface methods. Now it's only called when the type is instantiated - and in that case only virtual methods are affected. static interface methods must be marked regardless of instantiation.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2294,6 +2286,10 @@ void MarkGenericParameter (GenericParameter parameter)
 			}
 		}
 
+		/// <summary>
+		/// Returns true if any of the base methods of the <paramref name="method"/> passed is in an assembly that is not trimmed (i.e. action != trim)
+		/// </summary>
+		/// <remarks>This ignores any base methods defined in interfaces. To also check methods defined in interfaces</remarks>
 		bool IsVirtualNeededByTypeDueToPreservedScope (MethodDefinition method)
 		{
 			if (!method.IsVirtual)
```
vitek-karas, line 2295: I think with the current changes this will have to handle static interface methods as well.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: So the new name seems correct, but the method should not deal with statics at all.

jtschuster:
> Can you please also measure this on Windows build? (I assume this is a Release build, right?)

Yes, this is a release build. Here's the results for windows for the artifacts/bin/teshost folder (using the same runtime commit as before: 6ffb7f0)

| FixLibraryInterface branch | main branch | difference |
| -- | -- | -- |
| 248062280 | 248061288 | +992 |

jtschuster:
I added some more comments on the methods with my current understanding, please correct me if they're wrong.

vitek-karas:
Validated this change against the static interface issue found in runtime with Numerics. 
https://github.com/dotnet/runtime/pull/67453

With this change the static interface methods are correctly preserved.

vitek-karas (approved):

PR #2713, opened 2022-03-29
Break up ValueNode.cs so each Value has it's own file

jtschuster:
Copy and pasted each of the class definitions from ValueNode.cs to their own file to make navigation easier.


vitek-karas (approved):

sbomer (commented):
One more thing I noticed: would you mind moving ValueNodeList and ValueBasicBlockPair out of the shared namespace? They should just live in `Mono.Linker.Dataflow` I think.

PR #2716, opened 2022-04-01
Fix headers in new files

sbomer:
Fixes new headers which were causing the build to fail after https://github.com/dotnet/linker/commit/4f0d349e43a71d44ef71339293701fcfc2da997e.


jtschuster (approved):

jtschuster:
Thanks for catching this! Is this check not run on all PRs before they merge?

tlakollo:
> Thanks for catching this! Is this check not run on all PRs before they merge?

It does but your CI run, ran before the change was merged. Unless running a second time it wouldn't be possible to detect. Sorry for not informing you before

tlakollo (approved):

sbomer:
This can happen because PRs run against a merge commit that's created against main at the time that the PR is made. If main is later updated, the PRs don't get re-run on a new merge commit. Sometimes that kind of situation is caught by merge conflicts, but sometimes (like here) there's no visible merge conflict.

PR #2722, opened 2022-04-04
Fix local function called from method and state machine

sbomer:
Fixes https://github.com/dotnet/linker/issues/2723

I wasn't able to reproduce the issue by building dotnet/runtime for wasm locally (not sure what I'm doing wrong), but the stacktrace included enough info to see the problem. I was able to observe the same failure mode with these new tests. The issue was that a local function might be called from both the user method, and a state machine local function - and this case wasn't correctly handled in the call graph discovery. See commit comments for more details.

I also made it a warning (instead of throwing an exception) if we for some reason discover multiple user methods for a local function even with the fix - just in case. This is similar to what we're doing for state machines.


src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -92,6 +104,7 @@ void ProcessMethod (MethodDefinition method)
 						Debug.Assert (stateMachineType.DeclaringType == type ||
 							(CompilerGeneratedNames.IsGeneratedMemberName (stateMachineType.DeclaringType.Name) &&
 							 stateMachineType.DeclaringType.DeclaringType == type));
+						callGraph.TrackCall (method, stateMachineType);
 						if (!_compilerGeneratedTypeToUserCodeMethod.TryAdd (stateMachineType, method)) {
 							var alreadyAssociatedMethod = _compilerGeneratedTypeToUserCodeMethod[stateMachineType];
 							_context.LogWarning (new MessageOrigin (method), DiagnosticId.MethodsAreAssociatedWithStateMachine, method.GetDisplayName (), alreadyAssociatedMethod.GetDisplayName (), stateMachineType.GetDisplayName ());
```
vitek-karas, line 110: I see that we already have a problematic warning like the one you're adding. Maybe just file an issue that these warnings are not actionable - especially if we don't have a way to actually hit it.

src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -129,9 +142,24 @@ void ProcessMethod (MethodDefinition method)
...
+						Debug.Assert (CompilerGeneratedNames.IsLambdaOrLocalFunction (nestedFunction.Name));
+						// Nested functions get suppressions from the user method only.
+						if (!_compilerGeneratedMethodToUserCodeMethod.TryAdd (nestedFunction, userDefinedMethod)) {
+							var alreadyAssociatedMethod = _compilerGeneratedMethodToUserCodeMethod[nestedFunction];
+							_context.LogWarning (new MessageOrigin (userDefinedMethod), DiagnosticId.MethodsAreAssociatedWithUserMethod, userDefinedMethod.GetDisplayName (), alreadyAssociatedMethod.GetDisplayName (), nestedFunction.GetDisplayName ());
+						}
+						break;
+					case TypeDefinition stateMachineType:
+						// State machine types get suppressions from the state machine method,
+						// already tracked above.
```
vitek-karas, line 157:
```suggestion
						// Types in the call graph are always state machine types
						// For those all their methods are also part of the same call graph (the type is considered to be "transparent" and we track references/calls through it)
						// So we can rely on the graph to report all of the state machine type's methods to be reported
						// as part of the "reachable" members anywhere the state machine type itself is used.
```

It took me a long time to realize what this does and what the comment means. Something more descriptive in the comment would help a lot I think.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
```diff
@@ -1567,6 +1603,54 @@ static IEnumerable<int> TestDynamicallyAccessedMethodViaGenericMethodParameterIn
...
+				IEnumerable<int> IteratorLocalFunction ()
+				{
+					yield return 0;
+					LocalFunction ();
+					MethodWithRequires ();
+					yield return 1;
+
+					// Trimmer doesn't try to associate LocalFunction with IteratorLocalFunction
+					[ExpectedWarning ("IL2026", "--MethodWithRequires--", ProducedBy = ProducedBy.Trimmer)]
+					void LocalFunction () => MethodWithRequires ();
```
vitek-karas, line 1625: Should we create an issue for this? More as a documentation.
In theory this should be fixable, I would expect the local function to be generated on the state machine type, right?

vitek-karas (approved):
Looks good.

docs/error-codes.md
```diff
@@ -1827,6 +1827,37 @@ void TestMethod()
   }
   ```
 
+#### `IL2117`: Trim analysis: Methods 'method1' and 'method2' are both associated with state machine type 'type'. This is currently unsupported and may lead to incorrectly reported warnings.
+
```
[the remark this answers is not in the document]
sbomer: I don't expect to hit it, I just made the unexpected case more like the warning we would give if we saw multiple methods associated with the same state machine type, instead of throwing an exception. I'm personally fine with throwing an exception instead - should we take the same approach for IL2107 or is there a reason to take a different approach?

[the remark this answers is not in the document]
sbomer: I think your other comment https://github.com/dotnet/linker/pull/2722#discussion_r842688650 answers it - I'll file an issue.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
sbomer: No, the local function actually gets generated outside the state machine type so I think it's similar to the case of non-state-machine nested functions (this one https://github.com/dotnet/linker/pull/2689#discussion_r838799632). And this seems to be the case even if it captures state from the state machine - all of the captured state is hoisted into a separate type next to the state machine. Example: https://sharplab.io/#v2:CYLg1APgAgTAjFADAAinALAbgLACg0Bsq6yAsgBQCUyA3nsg8vY2gMwA8AlgHYAuAfGSq1mjRqLHIevKcgC8yRDlySWJADIAxZMLmDOYMMtUM0cVAHZFxxgF88toA===

src/linker/Linker/CompilerGeneratedState.cs
sbomer: https://github.com/dotnet/linker/issues/2726

src/linker/Linker/CompilerGeneratedState.cs
sbomer: I adjusted the comment slightly, hopefully it's more clear now. Thanks!

docs/error-codes.md
```diff
@@ -1827,6 +1827,37 @@ void TestMethod()
...
+- Trimmer currently can't correctly handle if the same compiler generated lambda or local function is associated with two different methods.
+
+  Only a meta-sample:
+
+  ```C#
+  [RequiresUnreferencedCode ("")] // This should suppress all warnings from the method
```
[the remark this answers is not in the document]
sbomer: That's right - I added a note.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
sbomer: I filed https://github.com/dotnet/linker/issues/2727, please feel free to add to it if I am missing anything specific to state machines.

agocke:
Yeah, I don't think the C# compiler should ever reference a local function or lambda from two methods, so if it did, that would be surprising.

Of course, the one exception would be two local functions referencing another local function in the same method. At the IL level, that first local function would appear to be referenced by two other methods, but those methods are themselves local functions.

sbomer:
> two local functions referencing another local function in the same method

Yup, that case was handled in the original change, just not when one of those local functions was a state machine - the implementation didn't identify MoveNext methods as part of the state machine local function.

PR #2725, opened 2022-04-05
Share GetConstructor intrinsic a lot of small fixes

vitek-karas:
The main change is sharing of the GetConstructor intrinsic and related tests.
To make those tests work this also implements Type.EmptyTypes and String.Empty and Array.Empty intrinsics.
These are all treated as special cases:
* Type.EmptyTypes and String.Empty are actually field accesses, so we don't have infra to share these yet, no need to add one for now
* Array.Empty is a generic method where we need to know the generic instantiation. The current sharing doesn't propagate this information
  so this forced the implementation to NOT be shared but instead done as a fallback which the shared one doesn't handle.

The rest of the changes are all about fixing how the shared intrinsic handling propagates return value from intrinsics.
Lot of small subtle bugs around null or emtpy inputs and forgetting to add return value and so on. The main changes are in GetMethod and GetNestedType intrinsics which had lot of issues.

Added tests for all of the corner cases.
NestedTypesUsedViaReflection test has been reworked as previously it didn't work correctly. Lot of the [Kept] attributes were fulfilled by some of the negative tests and thus the positive tests didn't have coverage (for example calling GetNestedType(unknownName) will mark all nested types on the type, so a subsequent call to GetNestedType("knownname") will not have an observable effect).

### DIFF
--- a/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
@@ -406,6 +406,7 @@ namespace ILLink.Shared.TrimAnalysis
 								} else {
 									// Otherwise fall back to the bitfield requirements
 									_requireDynamicallyAccessedMembersAction.Invoke (value, targetValue);
+									AddReturnValue (GetMethodReturnValue (calledMethod, returnValueDynamicallyAccessedMemberTypes));
 								}
 							}
 						} else if (value is NullValue) {
@@ -414,6 +415,7 @@ namespace ILLink.Shared.TrimAnalysis
 						} else {
 							// Otherwise fall back to the bitfield requirements
 							_requireDynamicallyAccessedMembersAction.Invoke (value, targetValue);
+							AddReturnValue (GetMethodReturnValue (calledMethod, returnValueDynamicallyAccessedMemberTypes));
 						}
 					}
 				}
@@ -441,6 +443,7 @@ namespace ILLink.Shared.TrimAnalysis
 						if (value is SystemTypeValue systemTypeValue) {
 							foreach (var stringParam in argumentValues[0]) {
 								if (stringParam is KnownStringValue stringValue && !BindingFlagsAreUnsupported (bindingFlags)) {
+									AddReturnValue (MultiValueLattice.Top);
 									foreach (var nestedTypeValue in GetNestedTypesOnType (systemTypeValue.RepresentedType, stringValue.Contents, bindingFlags)) {
 										MarkType (nestedTypeValue.RepresentedType);
 										AddReturnValue (nestedTypeValue);
@@ -531,6 +534,7 @@ namespace ILLink.Shared.TrimAnalysis
 									AddReturnValue (MultiValueLattice.Top);
 								} else {
 									_requireDynamicallyAccessedMembersAction.Invoke (value, targetValue);
+									AddReturnValue (GetMethodReturnValue (calledMethod, returnValueDynamicallyAccessedMemberTypes));
 								}
 							}
 						} else if (value is NullValue) {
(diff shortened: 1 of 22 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -302,14 +302,15 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
+			case IntrinsicId.Type_get_BaseType:
+			case IntrinsicId.Type_GetConstructor: {
 					var instanceValue = MultiValueLattice.Top;
 					IReadOnlyList<MultiValue> parameterValues = methodParams;
 					if (calledMethodDefinition.HasImplicitThis ()) {
 						instanceValue = methodParams[0];
 						parameterValues = parameterValues.Skip (1).ToImmutableList ();
 					}
-					return handleCallAction.Invoke (calledMethodDefinition, instanceValue, parameterValues, out methodReturnValue);
+					return handleCallAction.Invoke (calledMethodDefinition, instanceValue, parameterValues, out methodReturnValue, out _);
```
sbomer, line 313: Would it make sense to always call `handleCallAction.Invoke`, and then handle any non-shared intrinsics, to avoid duplicating the cases above?

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -665,6 +706,51 @@ GenericParameterValue genericParam
...
+						1 => (argumentValues[0].AsSingleValue () as ArrayValue)?.Size.AsConstInt (),
+						2 => (argumentValues[1].AsSingleValue () as ArrayValue)?.Size.AsConstInt (),
+						4 => (argumentValues[2].AsSingleValue () as ArrayValue)?.Size.AsConstInt (),
+						5 => (argumentValues[3].AsSingleValue () as ArrayValue)?.Size.AsConstInt (),
+						_ => null,
+					};
+
+					// Go over all types we've seen
+					foreach (var value in instanceValue) {
+						if (value is SystemTypeValue systemTypeValue && !BindingFlagsAreUnsupported (bindingFlags)) {
```
sbomer, line 733: Does this also need handling for null/empty inputs?

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -357,6 +363,11 @@ GenericParameterValue genericParam
 			// GetMethod (string, int, BindingFlags, Binder?, CallingConventions, Type[], ParameterModifier[]?)
 			//
 			case IntrinsicId.Type_GetMethod: {
+					if (instanceValue.IsEmpty () || argumentValues[0].IsEmpty ()) {
```
sbomer, line 366: Thanks a lot for the null/empty handling cleanup in this PR!

sbomer (approved):

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
vitek-karas: It does make sense - but I didn't want to do it in this change. That should happen when we have shared all of the intrinsics and those which can't be shared should all switch to the same behavior as the Array.Empty does in this change.
Similarly to MakeGenericMethod - it's not finished yet, and it doesn't handle empty input well yet - again, once all of that is shared we will fix it up.

The change was already getting out of hand as is, I didn't want to make it even bigger.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
vitek-karas: Not really - we could add it, but it's not necessary. We need special handling only for intrinsics which set return value, those which don't can have empty foreach loops, it doesn't cause a problem.

Not sure what's better - be consistent with unnecessary lines of code, or keep it shorter.

sbomer: Oh right, the difference is that it doesn't have a return value. My preference is consistency, but either way is fine.

PR #2728, opened 2022-04-05
Don't throw on write to get-only property

sbomer:
This fixes one of NREs encountered in https://github.com/dotnet/linker/issues/2718, where the analyzer failed to analyze an assignment to a get-only property. The linker will warn about the compiler-generated backing field in this case.

This fix just skips this case in the analyzer. It would be more correct to produce a warning because this represents a legitimate dataflow issue. The problem is that the analyzer has no insight into the compiler-generated backing field, as far as I know.

One option would be to produce a warning with the same code as the linker, but adjust the message so that it mentions the property instead of the backing field - it would require some changesto the DiagnosticId infrastructure since we would need slightly different messages for all of the warnings where a field assignment in the linker is seen as a property assignment in the analyzer. Or maybe we could ignore the discrepancy and just emit a warning with the same text, pretending that the property name is the name of the backing field. @vitek-karas curious about your opinion on this.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
@@ -103,7 +103,14 @@ namespace ILLink.RoslynAnalyzer.DataFlow
 				break;
 			case IPropertyReferenceOperation propertyRef:
 				// A property assignment is really a call to the property setter.
-				var setMethod = propertyRef.Property.SetMethod!;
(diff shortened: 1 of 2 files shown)


vitek-karas:
I'm fine with this change as-is to unblock usages of the analyzer.
But I would prefer we eventually fix this for real - Honestly I don't mind too much if the warning message mentions backing field - or we can reword the message such that it can be used for both cases. The linker one is already somewhat confusing anyway, since backing fields have weird names and normal users will not know what they are. So simplifying the message would probably be an improvement either way.

vitek-karas (approved):

PR #2732, opened 2022-04-07
Avoid analyzing cref for generic parameters

tlakollo:
Add check for cref in generic parameter dataflow to avoid failing with a nullreference exception
Refactor code to reuse the while loop for both INamedType and Method symbols
Add tests

Fixes https://github.com/dotnet/linker/issues/2724


LakshanF (approved):
Wow, didn't know that the analyzer needs to handle cref comments as well!

tlakollo:
I forgot to add a reference to the issue that Sven found (https://github.com/dotnet/linker/issues/2724) added to the description

vitek-karas (approved):

PR #2742, opened 2022-04-13
Condition out the BannedApiAnalyzers usage in source-build

MichaelSimons:
Fixes https://github.com/dotnet/linker/issues/2615

This will need to be backported to 6.0 as it is a source-build regression in the sense that it introduces a prebuilt.


sbomer (approved):
LGTM, thanks!

PR #2744, opened 2022-04-15
Handle multiple values for an array node

tlakollo:
Handle multiple values for an array node
Add test
Fixes https://github.com/dotnet/linker/issues/2736

### DIFF
--- a/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
@@ -195,16 +195,15 @@ namespace ILLink.RoslynAnalyzer.TrimAnalysis
 
 		public override MultiValue HandleArrayElementRead (MultiValue arrayValue, MultiValue indexValue, IOperation operation)
 		{
-			if (arrayValue.AsSingleValue () is not ArrayValue arr)
-				return UnknownValue.Instance;
-
+			MultiValue result = TopValue;
 			if (indexValue.AsConstInt () is not int index)
 				return UnknownValue.Instance;
 
-			if (arr.TryGetValueByIndex (index, out var elementValue))
-				return elementValue;
-
-			return UnknownValue.Instance;
+			foreach (var value in arrayValue) {
+				if (value is ArrayValue arr && arr.TryGetValueByIndex (index, out var elementValue))
+					result = result.Equals (TopValue) ? elementValue : _multiValueLattice.Meet (result, elementValue);
+			}
+			return result.Equals (TopValue) ? UnknownValue.Instance : result;
 		}
 
 		public override void HandleArrayElementWrite (MultiValue arrayValue, MultiValue indexValue, MultiValue valueToWrite, IOperation operation)
--- a/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
@@ -195,13 +195,13 @@ namespace ILLink.RoslynAnalyzer.TrimAnalysis
 
 		public override MultiValue HandleArrayElementRead (MultiValue arrayValue, MultiValue indexValue, IOperation operation)
 		{
-			MultiValue result = TopValue;
 			if (indexValue.AsConstInt () is not int index)
 				return UnknownValue.Instance;
 
+			MultiValue result = TopValue;
 			foreach (var value in arrayValue) {
 				if (value is ArrayValue arr && arr.TryGetValueByIndex (index, out var elementValue))
-					result = result.Equals (TopValue) ? elementValue : _multiValueLattice.Meet (result, elementValue);
+					result = _multiValueLattice.Meet (result, elementValue);
 			}
 			return result.Equals (TopValue) ? UnknownValue.Instance : result;
 		}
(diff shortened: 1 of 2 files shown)


tlakollo:
I found out that the trimmer is not merging correctly the two values either, is only using the PublicMethods value and disregarding the PublicFields one, therefore only producing a single warning instead of two.

sbomer:
I think the linker is just eliminating a branch because it understands `true` as a constant in the if statement - if you change it to something like `string.Empty.Length == 0` I think it will merge the values.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -195,16 +195,15 @@ public override void HandleAssignment (MultiValue source, MultiValue target, IOp
...
-
+			MultiValue result = TopValue;
 			if (indexValue.AsConstInt () is not int index)
 				return UnknownValue.Instance;
 
-			if (arr.TryGetValueByIndex (index, out var elementValue))
-				return elementValue;
-
-			return UnknownValue.Instance;
+			foreach (var value in arrayValue) {
```
sbomer, line 202: Why do we need to look over multiple `arrayValue`s here but not in `ScanLdelem`? Is it because the analyzer has made a copy of the `ArrayValue` and merged them into something that has multiple arrays (instead of one array with some merged value at the known index?

I think we should aim to have the linker and analyzer use essentially the same logic here. I would be ok with copying the `ArrayValue` and tracking merged values as multiple arrays since it seems like that is already what we do in the analyzer.

One alternative would be to maintain a notion of object identity, and adjust the meet operation so that meeting two array `ArrayValues` which represent the same object in memory, but with different elements, produces a single `ArrayValue` whose elements are the merged values. But I don't think it's worth it at the moment.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -195,16 +195,15 @@ public override void HandleAssignment (MultiValue source, MultiValue target, IOp
...
+				if (value is ArrayValue arr && arr.TryGetValueByIndex (index, out var elementValue))
+					result = result.Equals (TopValue) ? elementValue : _multiValueLattice.Meet (result, elementValue);
```
sbomer, line 204:
```suggestion
					result = _multiValueLattice.Meet (result, elementValue);
```

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -195,16 +195,15 @@ public override void HandleAssignment (MultiValue source, MultiValue target, IOp
 
 		public override MultiValue HandleArrayElementRead (MultiValue arrayValue, MultiValue indexValue, IOperation operation)
 		{
-			if (arrayValue.AsSingleValue () is not ArrayValue arr)
-				return UnknownValue.Instance;
-
+			MultiValue result = TopValue;
```
sbomer, line 198: nit: move this to after the `indexValue.AsConstInt ()` check

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
tlakollo: I looked at the linker code and we are executing `ScanStelem` which is serving the purpose of going through the array and merging the elements together. I don't think I see a lot of difference with the `HandleArrayElementRead`. If something analyzer code might be more efficient because it only asks once for the index being null

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -195,16 +195,15 @@ public override void HandleAssignment (MultiValue source, MultiValue target, IOp
...
 				return UnknownValue.Instance;
 
-			if (arr.TryGetValueByIndex (index, out var elementValue))
-				return elementValue;
-
-			return UnknownValue.Instance;
+			MultiValue result = TopValue;
+			foreach (var value in arrayValue) {
+				if (value is ArrayValue arr && arr.TryGetValueByIndex (index, out var elementValue))
+					result = _multiValueLattice.Meet (result, elementValue);
```
sbomer, line 204: I think this needs a case which returns `UnknownValue` if any of the values isn't an `ArrayValue` or doesn't have an element at this index.

test/Mono.Linker.Tests.Cases/DataFlow/ArrayDataFlow.cs
```diff
@@ -148,6 +168,19 @@ static void TestGetElementAtUnknownIndex (int i = 0)
 			arr[i].RequiresPublicFields ();
 		}
 
+		// Trimmer code doesnt handle locals from different branches separetely, therefore merges incorrectly GetMethods with Unknown producing both warnings
```
sbomer, line 171: Would you mind filing an issue about this?

sbomer (approved):
Thanks!

test/Mono.Linker.Tests.Cases/DataFlow/ArrayDataFlow.cs
tlakollo: Opened #2746

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -195,16 +195,17 @@ public override void HandleAssignment (MultiValue source, MultiValue target, IOp
...
+				else
+					return UnknownValue.Instance;
+			}
+			return result.Equals (TopValue) ? UnknownValue.Instance : result;
```
vitek-karas, line 208: Nit: In linker (And everywhere else really) we simply call `IsEmpty` as it reads better.

PR #2747, opened 2022-04-19
Remove AnalysisContext and ScopeStack from method body scanner

sbomer:
**Motivation:**

For compiler-generated code dataflow, we may need to analyze a compiler-generated method multiple times as part of the closure of methods reached from user code. This is necessary because as we analyze this closure of methods we might find new captured state that flows into a previously analyzed method in the same closure.

To avoid spurious warnings, we need a deduplication mechanism similar to what we do in the analyzer. The eventual goal is for the linker to have something similar to the analyzer's `TrimAnalysisPatternStore` that deduplicates warnings produced from multiple scans of the same method body.

We also might need the ability to run the method body scanner without producing any warnings, when analyzing compiler-generated methods reachable by reflection. (The linker will warn about reflection over compiler-generated code, and won't produce the full set of diagnostics for these methods, but still needs to scan them to ensure dependencies are kept).

The first step is to remove some levels of indirection around the message origins, to make it easier to cache messages by origin in the future.

**This change:**

This is an initial factoring of ReflectionMethodBodyScanner that just gets rid of AnalysisContext and ScopeStack usages in that class. No functional changes are intended.

Some of the ScopeStack logic is a bit subtle: the shared global ScopeStack meant that it was difficult to tell which scope might be used when reporting a warning - it was path-dependent. I pushed all of the `ScopeStack.CurrentScope.Origin` calls up into callsites outside of ReflectionMethodBodyScanner, so that the scanner logic could avoid concerning itself with pushing and popping scopes. Instead, it passes message origins along explicitly when required.


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -257,19 +251,16 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
...
-
 			DynamicallyAccessedMemberTypes returnValueDynamicallyAccessedMemberTypes = 0;
 
 			bool requiresDataFlowAnalysis = _context.Annotations.FlowAnnotations.RequiresDataFlowAnalysis (calledMethodDefinition);
 			returnValueDynamicallyAccessedMemberTypes = requiresDataFlowAnalysis ?
 				_context.Annotations.FlowAnnotations.GetReturnParameterAnnotation (calledMethodDefinition) : 0;
 
-			var handleCallAction = new HandleCallAction (_context, this, analysisContext, callingMethodDefinition);
+			_origin = _origin.WithInstructionOffset (operation.Offset);
+			var origin = _origin;
```
tlakollo, line 261: Is there any reason why we copy the value here?, eventually there are calls using sometimes origin and two lines later _origin

sbomer: No good reason, I think it was left over from the way I mechanically did some of the changes. Fixed, thank you!

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -212,17 +208,16 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
...
+				// Report warnings on access to annotated members, with the annotated type as the origin.
+				ApplyDynamicallyAccessedMembersToType (ref reflectionMethodBodyScanner, origin, type, annotation);
 				_typesInDynamicallyAccessedMembersHierarchy[type] = (annotation, true);
 			}
 
 			return applied;
 		}
 
-		void ApplyDynamicallyAccessedMembersToType (ref ReflectionMethodBodyScanner reflectionMethodBodyScanner, in ReflectionMethodBodyScanner.AnalysisContext analysisContext, TypeDefinition type, DynamicallyAccessedMemberTypes annotation)
+		void ApplyDynamicallyAccessedMembersToType (ref ReflectionMethodBodyScanner reflectionMethodBodyScanner, MessageOrigin origin, TypeDefinition type, DynamicallyAccessedMemberTypes annotation)
```
tlakollo, line 220: nit: We were passing the analysisContext with the in parameter modifier, I think we still want that

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -979,111 +983,114 @@ void ProcessCreateInstanceByName (in AnalysisContext analysisContext, MethodDefi
...
-		void RequireDynamicallyAccessedMembers (in AnalysisContext analysisContext, in MultiValue value, ValueWithDynamicallyAccessedMembers targetValue)
+		void RequireDynamicallyAccessedMembers (in MessageOrigin origin, bool diagnosticsEnabled, in MultiValue value, ValueWithDynamicallyAccessedMembers targetValue)
 		{
-			var requireDynamicallyAccessedMembersAction = new RequireDynamicallyAccessedMembersAction (_context, this, analysisContext);
+			var requireDynamicallyAccessedMembersAction = new RequireDynamicallyAccessedMembersAction (_context, this, origin, diagnosticsEnabled);
 			requireDynamicallyAccessedMembersAction.Invoke (value, targetValue);
 		}
 
-		internal void MarkTypeForDynamicallyAccessedMembers (in AnalysisContext analysisContext, TypeDefinition typeDefinition, DynamicallyAccessedMemberTypes requiredMemberTypes, DependencyKind dependencyKind, bool declaredOnly = false)
+		internal void MarkTypeForDynamicallyAccessedMembers (MessageOrigin origin, TypeDefinition typeDefinition, DynamicallyAccessedMemberTypes requiredMemberTypes, DependencyKind dependencyKind, bool declaredOnly = false)
```
tlakollo, line 1005: nit: Another missing 'in' parameter modifier

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1625,7 +1625,7 @@ static bool IsDeclaredWithinType (IMemberDefinition member, TypeDefinition type)
 			var memberScope = reportOnMember ? ScopeStack.PushScope (new MessageOrigin (member)) : null;
 
 			try {
-				var origin = ScopeStack.CurrentScope.Origin;
+				origin = reportOnMember ? ScopeStack.CurrentScope.Origin : origin;
```
tlakollo, line 1628: Nit: I think it might be better to just leave: `origin = ScopeStack.CurrentScope.Origin;`

At the beginning of the method, `origin` is equal to `ScopeStack.CurrentScope.Origin`, then we attempt to push into the stack depending on `reportOnMember`. Either way `ScopeStack.CurrentScope.Origin` contains the right value, either contains origin or contains the new MessageOrigin.

sbomer: `origin` might not be equal to `ScopeStack.CurrentScope.Origin` if the caller didn't push the scope to the scope stack - so I only wanted this to look at the scope stack if we know we pushed to it.

sbomer: While looking at it I realized this scope stack usage was no longer necessary, so I simplified the logic some. Thanks!

tlakollo: What I'm saying is that is safe to always assign `origin` to `ScopeStack.CurrentScope.Origin` because that value is either `origin` or `new MessageOrigin (member)`

tlakollo: New changes look good to me, thanks! :D

PR #2750, opened 2022-04-19
Add support for Debug output in Analyzer

tlakollo:
When debugging applications via de console is very useful to have a Console print about the different values being used. In the case of debugging via the Test Explorer in Visual Studio, the Console output is not available to the user.

This PR adds the functionality to be able to choose to print to the Debug console that is available in Visual Studio while running a test via the Test Explorer
Adds enum to choose between Console and Debug output

### DIFF
--- a/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
+++ b/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
@@ -54,7 +54,7 @@ namespace ILLink.RoslynAnalyzer.TrimAnalysis
 		// Set this to true to print out the dataflow states encountered during the analysis.
 		readonly bool showStates = false;
 
-		static readonly TracingType tracingMechanism = TracingType.Console;
+		static readonly TracingType tracingMechanism = Debugger.IsAttached ? TracingType.Debug : TracingType.Console;
 #pragma warning restore CA1805 // Do not initialize unnecessarily
 		ControlFlowGraphProxy cfg;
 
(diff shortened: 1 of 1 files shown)


src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
```diff
@@ -80,20 +89,40 @@ public override void TraceVisitBlock (BlockProxy block)
 			if (!trace)
 				return;
 
-			Console.Write ("block " + block.Block.Ordinal + ": ");
+			TraceWrite ("block " + block.Block.Ordinal + ": ", tracingMechanism);
```
sbomer, line 92: I think we might as well remove the `tracingMechanism` parameter and just check the field inside `TraceWriteLine`.

sbomer:
I checked and it seems like the Console output is visible for failed testcases on macOS (when running with `dotnet test`), but not windows. :( I wonder if there's a way to make this more consistent.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
```diff
@@ -80,20 +89,40 @@ public override void TraceVisitBlock (BlockProxy block)
...
+				Console.WriteLine (tracingInfo);
+			else if (tracingMechanism == TracingType.Debug)
+				Debug.WriteLine (tracingInfo);
+			else
+				throw new NotImplementedException (message: "invalid TracingType is being used");
+		}
+
+		private static void TraceWrite (string tracingInfo)
+		{
+			if (tracingMechanism == TracingType.Console)
```
sbomer, line 120: nit: might be more appropriate as a switch

sbomer (approved):
Thanks!

tlakollo:
> I checked and it seems like the Console output is visible for failed test cases on macOS (when running with `dotnet test`), but not windows. :( I wonder if there's a way to make this more consistent.

I found several issues open because of VS behavior and the inability to see the console output in tests. Apparently, NUnit might show a link with the console output in the test explorer but Xunit will not. Most of the provided answers are just "use Debug.Write/Debug.WriteLine"

PR #2754, opened 2022-04-20
Share method handle related intrinsics

vitek-karas:
Moving some simple method handle related intrinsics and unifying the values related to it.


test/Mono.Linker.Tests.Cases/DataFlow/MakeGenericDataFlow.cs
```diff
@@ -684,6 +686,12 @@ static void TestWithUnmanagedConstraint ()
 			{
 				var t = new T ();
 			}
+
+			static void TestGetMethodFromHandle (Type unknownType = null)
+			{
+				MethodInfo m = (MethodInfo) MethodInfo.GetMethodFromHandle (typeof (MakeGenericMethod).GetMethod (nameof (GenericWithNoRequirements)).MethodHandle);
+				m.MakeGenericMethod (unknownType);
```
sbomer, line 693: Just making sure I understand how the test works - is the idea that if we didn't support GetMethodFromHandle, the call to `MakeGenericMethod` would warn about an unknown method?

Might be worth adding a test that produces some warnings (where the generic parameters have requirements).

test/ILLink.RoslynAnalyzer.Tests/generated/ILLink.RoslynAnalyzer.Tests.Generator/ILLink.RoslynAnalyzer.Tests.TestCaseGenerator/UnreachableBlockTests.g.cs
```diff
@@ -37,6 +37,12 @@ public Task InstanceMethodSubstitutions ()
 			return RunTest (allowMissingWarnings: true);
 		}
 
+		[Fact]
+		public Task MethodArgumentPropagation ()
```
sbomer, line 41: I think @tlakollo already added these

sbomer (approved):
LGTM, thank you!

test/Mono.Linker.Tests.Cases/DataFlow/MakeGenericDataFlow.cs
vitek-karas: Yes - if the method is not know MakeGenericMethod should always warn... so the fact that it doesn't warn means we were able to correctly flow the MethodInfo.
I'll add a negative test as well.

PR #2756, opened 2022-04-22
Fix handling of special intrinsics in analyzer

vitek-karas:
Some intrinsics require generic instantiation handling. Currently the shared code doesn't have the abitlity to do this, so we leave it as a special case and analyzer/linker have special code to handle it.
In the analyzer this code was missing from the diagnsotic reporting part and thus the intrinsic was handled as a normal method call - which caused problems with TypeDelegator.

Added tests for TypeDelegator which cover these issues.

This is blocking linker->runtime flow: https://github.com/dotnet/runtime/pull/68386

### DIFF
--- a/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisMethodCallPattern.cs
+++ b/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisMethodCallPattern.cs
@@ -44,20 +44,11 @@ namespace ILLink.RoslynAnalyzer.TrimAnalysis
 			DiagnosticContext diagnosticContext = new (Operation.Syntax.GetLocation ());
 			HandleCallAction handleCallAction = new (diagnosticContext, OwningSymbol, Operation);
 			if (!handleCallAction.Invoke (new MethodProxy (CalledMethod), Instance, Arguments, out _, out _)) {
-				// If the intrinsic handling didn't work we have to:
-				//   Handle the instance value
-				//   Handle argument passing
-				//   Construct the return value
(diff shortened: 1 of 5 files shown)


src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisMethodCallPattern.cs
```diff
@@ -44,20 +44,11 @@ public IEnumerable<Diagnostic> CollectDiagnostics ()
...
-				// Note: this is temporary as eventually the handling of all method calls should be done in the shared code (not just intrinsics)
-				if (!CalledMethod.IsStatic) {
-					var requireDynamicallyAccessedMembersAction = new RequireDynamicallyAccessedMembersAction (diagnosticContext, new ReflectionAccessAnalyzer ());
-					requireDynamicallyAccessedMembersAction.Invoke (Instance, new MethodThisParameterValue (CalledMethod));
-				}
-
-				for (int argumentIndex = 0; argumentIndex < Arguments.Length; argumentIndex++) {
-					var requireDynamicallyAccessedMembersAction = new RequireDynamicallyAccessedMembersAction (diagnosticContext, new ReflectionAccessAnalyzer ());
-					requireDynamicallyAccessedMembersAction.Invoke (Arguments[argumentIndex], new MethodParameterValue (CalledMethod.Parameters[argumentIndex]));
-				}
```
sbomer, line 60: Should we also remove the `default` case from `TrimAnalysisVisitor.HandleMethodCall` (make it throw) since `None` is already handled by the shared code?

sbomer (approved):
LGTM, one tangentially related suggestion.

sbomer:
LGTM, thank you!

PR #2757, opened 2022-04-22
Move shared dataflow value creation helpers out of HandleCallAction

sbomer:
This builds on the work @jtschuster did in https://github.com/dotnet/linker/commit/c2fff337c62780b2c84d5fda975a4e001bb6fe8b to share code for the annotation checks.

I noticed that ReflectionMethodBodyScanner and HandleCallAction started accumulating these helpers for creating dataflow values from methods and I thought it would make sense for that to be handled as a separate concern, so I moved them out to a separate class and adjusted some of the callsites in ReflectionMethodBodyScanner to use the shared version.

This was motivated in part by some of the factoring I am doing as a follow-up to https://github.com/dotnet/linker/commit/137b7b994e410843b783cb5b04c92a6ca693b2e8, but isn't directly related. (While working on the follow-up, I came across a point where I needed to create a dataflow value in shared code from outside HandleCallAction, and there wasn't a very clean way to do this).


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -906,7 +889,7 @@ bool AnalyzeGenericInstantiationTypeArray (in MessageOrigin origin, bool diagnos
 						// https://github.com/dotnet/linker/issues/2428
 						// We need to report the target as "this" - as that was the previous behavior
 						// but with the annotation from the generic parameter.
-						var targetValue = GetMethodParameterValue (calledMethod, 0, _context.Annotations.FlowAnnotations.GetGenericParameterAnnotation (genericParameters[i]));
+						var targetValue = GetMethodParameterValueInternal (calledMethod, 0, _context.Annotations.FlowAnnotations.GetGenericParameterAnnotation (genericParameters[i]));
```
tlakollo, line 892: Why do we call the internal method here?

tlakollo (approved):
In general looks good to me

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
sbomer: I interpreted it as "internal to ReflectionMethodBodyScanner" but maybe that was meant more as a helper for the `GetMethodParameterValue` overload that I removed. I renamed it to just be an overload of `GetMethodParameterValue`.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -21,15 +21,16 @@ partial struct HandleCallAction
...
 
 		public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnlyList<MultiValue> argumentValues, out MultiValue methodReturnValue, out IntrinsicId intrinsicId)
 		{
 			MultiValue? returnValue = null;
 
 			bool requiresDataFlowAnalysis = MethodRequiresDataFlowAnalysis (calledMethod);
-			DynamicallyAccessedMemberTypes returnValueDynamicallyAccessedMemberTypes = requiresDataFlowAnalysis ?
-				GetReturnValueAnnotation (calledMethod) : 0;
+			var annotatedMethodReturnValue = _annotationContext.GetMethodReturnValue (calledMethod);
+			Debug.Assert (requiresDataFlowAnalysis || annotatedMethodReturnValue.DynamicallyAccessedMemberTypes == DynamicallyAccessedMemberTypes.None);
```
jtschuster, line 33: Nit: we could inline `MethodRequiresDataFlowAnalysis(calledMethod)` and remove `requiresDataFlowAnalysis` since it's only used in one place outside of the assert.

jtschuster (approved):
Would it make more sense to put all of this into `FlowAnnotations` and make that a partial class? It looks like there are already a few duplicate methods in FlowAnnotations.cs on the linker side. LGTM otherwise.

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
sbomer: I can't inline it exactly because it's called from the shared code - but I moved it to `AnnotationContext` (now `FlowAnnotations` in response to your other suggestion.)

sbomer:
I was debating whether to use `FlowAnnotations` for this - if you had the same thought  it's probably worth it. I took the suggestion, Thanks!

vitek-karas (commented):
Sorry for the late review - just one perf related comment.

PR #2761, opened 2022-04-25
Factor ReflectionMethodBodyScanner to be more similar to analyzer

sbomer:
Follow-up to https://github.com/dotnet/linker/pull/2747. This is a factoring change only and doesn't introduce functional changes.

This does two things to move the method body scanner to be more similar to the trim analysis visitor:
- Uses `DiagnosticContext` to record diagnostics. Right now, this still happens inline as we run the scanner, but the idea will be to move this out to a driver that runs the analysis, and then records diagnostics from the recorded patterns (like how it is done in the analyzer)
- Moves reflection-marking helpers out of `ReflectionMethodBodyScanner` to a separate class. This prevents us from having to pass around the scanner to other helpers, and makes it more obvious which parts of the code to scanning only vs other things. The new `ReflectionMarker` is conceptually similar to `ReflectionAccessAnalyzer` in the Roslyn analyzer and could be a candidate for future code sharing.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMarker.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMarker.cs
@@ -51,12 +51,12 @@ namespace Mono.Linker.Dataflow
 			// Only difference here is whether we treat arrays specially. Maybe we should do that more frequently elsewhere too.
 			// _reflectionMarker.MarkType (origin, );
 			if (!_context.TypeNameResolver.TryResolveTypeName (typeName, origin.Provider, out TypeReference? typeRef, out AssemblyDefinition? typeAssembly)
-				|| ReflectionMethodBodyScanner.ResolveToTypeDefinition (_context, typeRef) is not TypeDefinition foundType) {
+				|| typeRef.ResolveToTypeDefinition (_context) is not TypeDefinition foundType) {
 				type = default;
 				return false;
 
 			}
-			
+
 			_markStep.MarkTypeVisibleToReflection (typeRef, foundType, new DependencyInfo (DependencyKind.AccessedViaReflection, origin.Provider), origin);
 			_context.MarkingHelpers.MarkMatchingExportedType (foundType, typeAssembly, new DependencyInfo (DependencyKind.DynamicallyAccessedMember, foundType), origin);
 
--- a/src/linker/Linker.Dataflow/ReflectionMarker.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMarker.cs
@@ -46,10 +46,27 @@ namespace Mono.Linker.Dataflow
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -695,8 +699,9 @@ bool IsComInterop (IMarshalInfoProvider marshalInfoProvider, TypeReference param
 			return false;
 		}
 
-		void ProcessCreateInstanceByName (in MessageOrigin origin, bool diagnosticsEnabled, MethodDefinition calledMethod, ValueNodeList methodParams)
+		void ProcessCreateInstanceByName (in MessageOrigin origin, in DiagnosticContext diagnosticContext, MethodDefinition calledMethod, ValueNodeList methodParams)
```
tlakollo, line 702: Why this case is special and has both `origin` and `diagnosticContext` and not just `diagnosticContext`?

src/linker/Linker.Dataflow/ReflectionMarker.cs
```diff
@@ -0,0 +1,126 @@
...
+					MarkInterfaceImplementation (origin, interfaceImplementation, dependencyKind);
+					break;
+				}
+			}
+		}
+
+		internal bool TryResolveTypeNameAndMark (string typeName, MessageOrigin origin, [NotNullWhen (true)] out TypeDefinition? type)
+		{
+			// Only difference here is whether we treat arrays specially. Maybe we should do that more frequently elsewhere too.
+			// _reflectionMarker.MarkType (origin, );
```
tlakollo, line 52: nit: this comment seems kind of unfinished

src/linker/Linker.Dataflow/ReflectionMarker.cs
```diff
@@ -0,0 +1,126 @@
...
+			if (!_context.TypeNameResolver.TryResolveTypeName (typeName, origin.Provider, out TypeReference? typeRef, out AssemblyDefinition? typeAssembly)
+				|| typeRef.ResolveToTypeDefinition (_context) is not TypeDefinition foundType) {
+				type = default;
+				return false;
+
```
tlakollo, line 57: nit: extra line

tlakollo (approved):

src/linker/Linker.Dataflow/ReflectionMarker.cs
sbomer: Thanks, this was accidentally left over.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
sbomer: No good reason, I meant to remove this assert but forgot about it. Good catch!

vitek-karas (approved):

vitek-karas:
I'm going to merge this - working on more sharing and this would come really handy.

PR #2766, opened 2022-04-26
Change requires on type checking for mismatch

### LINKED ISSUE #2763: Analyzer: RUC/RDC on type triggers override mismatch warnings
```C#
public class Base
{
    public virtual void Method() {}
}

[RequiresUnreferencedCode("RUC")]
[RequiresDynamicCode("RDC")]
public class Derived : Base
{
    // IL2046 - Analyzer only
    // IL3051
    public override void Method() {}
}
```

RUC on type should mean:
* RUC on all statics and .ctors
* Suppress trim analysis warnings in all method bodies

But the analyzer in this case treats it as if it has RUC on virtual methods as well.
The IL2046 is expected if the derived method as RUC on it, but the class as a whole should not trigger this warning since it doesn't imply any changes to instance methods as seen from outside (calling instance methods on RUC annotated type does not produce warnings, only .ctors do).

This prevents usage of the RUC/RDC on type feature in runtime repo. There are quite a few examples in System.Text.Json specifically - for example: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonValueNotTrimmable.cs. Instead of having a RUC on .ctor and suppression on virtual method, this should be done with RUC on type alone.

tlakollo:
Fixes #2763

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -357,10 +357,7 @@ namespace ILLink.RoslynAnalyzer
 			bool member2CreatesRequirement = member2.TargetHasRequiresAttribute (RequiresAttributeName, out _);
 			bool member1FulfillsRequirement = member1.IsOverrideInRequiresScope (RequiresAttributeName);
 			bool member2FulfillsRequirement = member2.IsOverrideInRequiresScope (RequiresAttributeName);
-			if ((member1CreatesRequirement && !member2FulfillsRequirement) || (member2CreatesRequirement && !member1FulfillsRequirement))
-				return true;
-			else
-				return false;
+			return (member1CreatesRequirement && !member2FulfillsRequirement) || (member2CreatesRequirement && !member1FulfillsRequirement);
 		}
 
 		protected abstract string GetMessageFromAttribute (AttributeData requiresAttribute);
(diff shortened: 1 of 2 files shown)


src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -353,9 +353,14 @@ private void ReportMismatchInAttributesDiagnostic (SymbolAnalysisContext symbolA
...
 		private bool HasMismatchingAttributes (ISymbol member1, ISymbol member2)
 		{
-			bool member1HasAttribute = member1.IsOverrideInRequiresScope (RequiresAttributeName);
-			bool member2HasAttribute = member2.IsOverrideInRequiresScope (RequiresAttributeName);
-			return member1HasAttribute ^ member2HasAttribute;
+			bool member1CreatesRequirement = member1.TargetHasRequiresAttribute (RequiresAttributeName, out _);
+			bool member2CreatesRequirement = member2.TargetHasRequiresAttribute (RequiresAttributeName, out _);
+			bool member1FulfillsRequirement = member1.IsOverrideInRequiresScope (RequiresAttributeName);
+			bool member2FulfillsRequirement = member2.IsOverrideInRequiresScope (RequiresAttributeName);
+			if ((member1CreatesRequirement && !member2FulfillsRequirement) || (member2CreatesRequirement && !member1FulfillsRequirement))
```
sbomer, line 360: nit: instead of an if/else this could just return the boolean expression.

sbomer (approved):
LGTM, thanks!

jtschuster (approved):

PR #2771, opened 2022-04-28
Don't remove MethodImpl if overridden method is not in a link assembly

jtschuster:
Before removing overrides, we should check if the overridden method is
in an ignore scope.

In the linker we consider types and methods in a `skip` assembly as "removed", so when the linker would check if an overridden method is removed, it would see that it is considered removed, even though it should be considered kept. The workaround is to use the IgnoreScope helper to determine if the interface method is in a `link` assembly and removed, and only remove the MethodImpl if both are true.

Now, whether `ShouldRemove` should return true or false for things in a skip assembly, I'm not sure. It depends on whether it does / should mean "Actively removed from the assembly" or "not in the final output".

This PR fixes the issues found in https://github.com/dotnet/runtime/pull/68612.

### DIFF
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -482,10 +482,10 @@ namespace Mono.Linker.Steps
 		/// <summary>
 		/// Returns true if the assembly of the <paramref name="scope"></paramref> is not set to link (i.e. action=copy is set for that assembly)
 		/// </summary>
-	    private bool IgnoreScope (IMetadataScope scope)
+		private bool IgnoreScope (IMetadataScope scope)
 		{
-		   AssemblyDefinition? assembly = Context.Resolve (scope);
-		   return assembly != null && Annotations.GetAction (assembly) != AssemblyAction.Link;
+			AssemblyDefinition? assembly = Context.Resolve (scope);
+			return assembly != null && Annotations.GetAction (assembly) != AssemblyAction.Link;
 		}
 
 		void SweepDebugInfo (Collection<MethodDefinition> methods)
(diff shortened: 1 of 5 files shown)


src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -472,12 +472,22 @@ protected virtual void SweepMethods (Collection<MethodDefinition> methods)
 		void SweepOverrides (MethodDefinition method)
 		{
 			for (int i = 0; i < method.Overrides.Count;) {
-				if (Context.Resolve (method.Overrides[i]) is MethodDefinition ov && ShouldRemove (ov))
+				if (Context.Resolve (method.Overrides[i]) is MethodDefinition ov && ShouldRemove (ov) && !IgnoreScope (ov.DeclaringType.Scope))
```
agocke, line 475: Can we avoid checking each overridden method individually by sweeping the overrides themselves, instead of sweeping the overrides from the base type?

agocke:
Definitely think we need a test here. @sbomer any idea on what might be happening in the runtime repo that isn't the same in our testing system?

vitek-karas (approved):

agocke:
Ah, so the thing I was missing here was

> [SetupLinkerAction ("skip", "skiplibrary")]

which turns the S.P.C reference to "skip"

tannergooding:
About how long does it normally take this to flow to `dotnet/runtime`? No particular rush atm, just wanting to make sure it stays on my own radar.

tannergooding:
Looks like https://github.com/dotnet/runtime/pull/68650 has it and I just need to wait on that to merge.

vitek-karas:
Typically less then a day... if everything is green - which it hasn't been (as we improve the analyzer we find either issues in the analyzer or in the setup in runtime repo)

PR #2772, opened 2022-04-29
Don't throw on InstanceReference assignment

sbomer:
Fixes https://github.com/dotnet/linker/issues/2770

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
@@ -176,6 +176,10 @@ namespace ILLink.RoslynAnalyzer.DataFlow
 				// This can happen for a field assignment in an attribute instance.
 				// TODO: validate against the field attributes.
 				break;
+			case IInstanceReferenceOperation:
+				// Assignment to 'this' is not tracked currently.
+				// Not relevant for trimming dataflow.
+				break;
 			default:
 				// NoneOperation represents operations which are unimplemented by Roslyn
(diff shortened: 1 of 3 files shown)


agocke (approved):
LGTM, thanks!

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -176,6 +176,10 @@ public override TValue VisitSimpleAssignment (ISimpleAssignmentOperation operati
 				// This can happen for a field assignment in an attribute instance.
 				// TODO: validate against the field attributes.
 				break;
+			case IInstanceReferenceOperation:
+				// Assignment to 'this' is not tracked currently.
+				// Not relevant for trimming dataflow.
```
vitek-karas, line 181: Technically if the method itself is annotated with DAM - that annotation actually applies to `this` and so if the method assigned to this, the source should have matching annotations.

That said - what is the code which would even do this in the first place?
And we only allow annotating methods on corelib Type related types, so it doesn't matter if we don't do anything.

test/Mono.Linker.Tests.Cases/DataFlow/MethodThisDataFlow.cs
```diff
@@ -148,6 +156,24 @@ public static void StaticMethod ()
...
+
+		struct StructType
+		{
+			int f;
+			public StructType (int f) => this.f = f;
+
+			public void AssignToThis ()
+			{
+				// Not relevant for dataflow, but this should not crash the analyzer.
+				this = new StructType ();
```
vitek-karas, line 168: Wow - this is... so weird...

PR #2774, opened 2022-05-02
Add ability for the analyzer to recognize const fields

vitek-karas:
This means the value of the fields is tracked as a const value instead of a field reference.
This is to support some additional code patterns as well as align the behavior with the linker. Compiler will inline the const fields into the IL effectively removing the field reference in these cases and linker only sees the constant. So ideally the analyzer should have a similar behavior.

This is currently blocking runtime integration: https://github.com/dotnet/runtime/pull/68650

### DIFF
--- a/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
@@ -55,19 +55,8 @@ namespace ILLink.RoslynAnalyzer.TrimAnalysis
 			// If the return value is empty (TopValue basically) and the Operation tree
 			// reports it as having a constant value, use that as it will automatically cover
 			// cases we don't need/want to handle.
-			if (operation != null && returnValue.IsEmpty () && operation.ConstantValue.HasValue) {
-				object? constantValue = operation.ConstantValue.Value;
-				if (constantValue == null)
-					return NullValue.Instance;
-				else if (operation.Type?.SpecialType == SpecialType.System_String && constantValue is string stringConstantValue)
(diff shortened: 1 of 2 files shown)


src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs
```diff
@@ -297,5 +289,31 @@ public override void HandleReturnValue (MultiValue returnValue, IOperation opera
...
+					return true;
+				} else if (operation.Type?.TypeKind == TypeKind.Enum && constantValue is int enumConstantValue) {
+					constValue = new ConstIntValue (enumConstantValue);
+					return true;
+				} else if (operation.Type?.SpecialType == SpecialType.System_Int32 && constantValue is int intConstantValue) {
+					constValue = new ConstIntValue (intConstantValue);
+					return true;
+				} else if (operation.Type?.SpecialType == SpecialType.System_Boolean && constantValue is bool boolConstantValue) {
+					constValue = new ConstIntValue (boolConstantValue ? 1 : 0);
+					return true;
```
stephentoub, line 311: Why only these types rather than any primitive that can be const?

vitek-karas: Because only these are recognized by the rest of trimming analysis. And also because we only have value system representation for these (for example `ConstInstValue`... we would need to introduce similar types for other values like `DateTime`, `double`, ...)

vitek-karas: I'll look into this in a followup change

PR #2777, opened 2022-05-02
Don't depend on marking order in DAM type hierarchy

sbomer:
This will allow `ApplyDynamicallyAccessedMembersToTypeHierarchy` to be called from the scanner when it sees `object.GetType` to get annotations on the type, without requiring that the type has been marked yet. (It will be marked during the final pass after the flow analysis has converged.)

### DIFF
--- a/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
+++ b/src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
@@ -47,14 +47,13 @@ namespace Mono.Linker.Dataflow
 
 		public (DynamicallyAccessedMemberTypes annotation, bool applied) ProcessMarkedTypeForDynamicallyAccessedMembersHierarchy (TypeDefinition type)
 		{
+			// We'll use the cache also as a way to detect and avoid recursion for interfaces and annotated base types
 			if (_typesInDynamicallyAccessedMembersHierarchy.TryGetValue (type, out var existingValue))
 				return existingValue;
 
 			DynamicallyAccessedMemberTypes annotation = _context.Annotations.FlowAnnotations.GetTypeAnnotation (type);
 			bool apply = false;
 
-			// We'll use the cache also as a way to detect and avoid recursion
-			// There's no possiblity to have recursion among base types, so only do this for interfaces
 			if (type.IsInterface)
 				_typesInDynamicallyAccessedMembersHierarchy.Add (type, (annotation, false));
 


src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -47,26 +47,20 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
...
 		{
-			// Non-interfaces must be marked already
-			Debug.Assert (type.IsInterface || _context.Annotations.IsMarked (type));
+			if (_typesInDynamicallyAccessedMembersHierarchy.TryGetValue (type, out var existingValue))
+				return existingValue;
 
 			DynamicallyAccessedMemberTypes annotation = _context.Annotations.FlowAnnotations.GetTypeAnnotation (type);
 			bool apply = false;
 
 			// We'll use the cache also as a way to detect and avoid recursion
```
vitek-karas, line 56: This comment should move above the very first thing in the method (which is what sort of checks for recursion). It will need to be updated as it now acts on all types, not just interfaces.

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
```diff
@@ -210,14 +201,14 @@ public DynamicallyAccessedMembersTypeHierarchy (LinkContext context, MarkStep ma
...
-				ApplyDynamicallyAccessedMembersToType (ref reflectionMarker, origin, type, annotation);
+				ApplyDynamicallyAccessedMembersToType (reflectionMarker, origin, type, annotation);
 				_typesInDynamicallyAccessedMembersHierarchy[type] = (annotation, true);
 			}
 
 			return applied;
 		}
 
-		void ApplyDynamicallyAccessedMembersToType (ref ReflectionMarker reflectionMarker, in MessageOrigin origin, TypeDefinition type, DynamicallyAccessedMemberTypes annotation)
+		void ApplyDynamicallyAccessedMembersToType (in ReflectionMarker reflectionMarker, in MessageOrigin origin, TypeDefinition type, DynamicallyAccessedMemberTypes annotation)
```
vitek-karas, line 211: Is it correct that we should use the RequireDynamicallyAccessedMembersAction.MarkTypeForDynamicallyAccessedMembers instead of the ReflectionMarker once we share this code with the analyzer?
(Maybe you thought about this... if not, no matter, somebody will look into it once we implement sharing of this).

vitek-karas (approved):
I "think" this is correct - this code still hurts my brain every time I look at it :-)

src/linker/Linker.Dataflow/DynamicallyAccessedMembersTypeHierarchy.cs
vitek-karas: I actually didn't even think about that. I was thinking what it would take to share this code with the analyzer, which doesn't have a `ReflectionMarker`.

PR #2779, opened 2022-05-02
Add ValueHolder to represent places where a value can be stored and referenced

jtschuster:
Related to https://github.com/dotnet/linker/pull/2769

In order to fully support `ref` parameters and locals, we need to be able to (1) know what kind of variable the value comes from (parameter, local, or array), and (2) mutate the value in place (or share to all other references that the value has changed). The current dataflow tracking doesn't make a real distinction between a value and an entity that can hold a value, and I think that it might be a good idea to fully separate those concepts into something like Values and ValueHolders.

This PR is a prototype for what the ValueHolder might look like in the Linker, and some of the changes necessary in MethodBodyScanner. There would be much more changes necessary to integrate ValueHolder completely (I'm expecting ValueHolder to replace MultiValue in most places), and it won't pass CI, I just wanted to get feedback and alternative ideas early on, and discuss if this is a direction we should go in.

Alternatively, we could use a struct that has a reference to a MultiValue, and a ValueHolderKind as a special case for references. This may be a simpler change, but it seems like it would involve lots of special casing and doesn't separate the concept of Values and ThingsThatCanHoldValues.

### DIFF
--- a/src/ILLink.Shared/TrimAnalysis/ValueHolder.cs
+++ b/src/ILLink.Shared/TrimAnalysis/ValueHolder.cs
@@ -1,0 +1,60 @@
+﻿using System.Diagnostics;
+using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.SingleValue>;
+
+namespace ILLink.Shared.TrimAnalysis
+{
+	internal partial class ValueHolder : IValueHolder
+    {
+		public MultiValue Value { get; private set; }
+		internal ValueHolderKind Kind { get; init; }
+		internal ValueHolder (MultiValue value, ValueHolderKind kind)
+		{
+			Value = value;
+			Kind = kind;
+		}
+		internal partial class ValueHolderReference<TKey> : IValueHolder
+		{
+			readonly ValueHolder Holder;
+			public readonly TKey Key;
+			internal ValueHolderReference (ValueHolder slot, TKey key)
+			{
+				Holder = slot;
+				Key = key;
+			}
+			public  MultiValue Value => Holder.Value;
+			internal ValueHolderKind Kind => Holder.Kind;
+			internal void SetValue (MultiValue value)
+			{
+				Debug.Assert (IsMutable);
+				Holder.Value = value;
+			}
+			public bool IsMutable { get => IsMutableKind (Kind); }
+		}
+
+
+		public static bool IsMutableKind (ValueHolderKind kind)
+		{
+			return kind switch {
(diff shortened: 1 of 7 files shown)


src/ILLink.Shared/TrimAnalysis/ValueHolder.cs
```diff
@@ -0,0 +1,61 @@
...
+	internal partial class ValueHolder : IValueHolder
+    {
+		public MultiValue Value { get; private set; }
+		internal ValueHolderKind Kind { get; init; }
+		internal ValueHolder (MultiValue value, ValueHolderKind kind)
+		{
+			Value = value;
+			Kind = kind;
+		}
+		internal partial class ValueHolderReference<TKey> : IValueHolder
```
vitek-karas, line 15: I don't really understand what is this type for.
I think I understand the idea behind the ValueHolder itself, but not really this reference type.
The linker already has a notion of a reference - that's basically what the ByRef field on stack slot is for. Why do we need a similar concept again? Unless you want to replace one with the other...

jtschuster:
After giving it more thought, I think I have a better idea of what might make sense here.

ValueHolder is just a way to wrap the MultiValue and allow us to change the value in a variable through a reference. These would be used in local variable stores, `ArrayValue`s, and anything else that can't be annotated and we need to track it ourself. These shouldn't need to have a ValueHolderKind or any additional info except a MultiValue. 

ValueHolder.Reference will point to a ValueHolder and should have a ValueHolderKind or some info about what kind of variable it points to. It should be a SingleValue, since that can be assigned to a local or passed as a parameter. This would require some care to make sure when assigning `A = B` when both are references to make sure A doesn't point to B, but instead to the ValueHolder that B points to.

jtschuster:
> I am wondering if some of this bookkeeping would be easier if instead of adding a ValueHolder abstraction, we made the ref-value type hold a LocalVariableDefinition (or ParameterDefinition, etc), similar to [CapturedReferenceValue](https://github.com/dotnet/linker/blob/main/src/ILLink.RoslynAnalyzer/DataFlow/CapturedReferenceValue.cs#L10) in the analyzer. Then there's no need to do the manual bookkeeping because we can just look up the local when we need it and update it in the dictionary.

Definitely agree on this point, Vitek and I spoke yesterday and came to the same conclusion - sorry for not giving an update here. I am working on the changes to accommodate that.

jtschuster:
Most of this shouldn't be needed anymore as per our discussions. Closing this PR and putting any new changes in https://github.com/dotnet/linker/pull/2769.

PR #2780, opened 2022-05-02
Fix warning about duplicate attributes

sbomer:
Duplicate attributes were producing warnings on every callsite in the linker to `GetLinkerAttributes`, which makes the number of warnings depend on the number of callsites in user code.

Deduplicate these by producing the warning only once, when building the linker attributes cache.

Fixes one of the remaining issues encountered while adding trim analysis patterns in the linker.

### DIFF
--- a/src/linker/Linker/Annotations.cs
+++ b/src/linker/Linker/Annotations.cs
@@ -582,7 +582,8 @@ namespace Mono.Linker
 		public bool TryGetLinkerAttribute<T> (IMemberDefinition member, [NotNullWhen (returnValue: true)] out T? attribute) where T : Attribute
 		{
 			var attributes = GetLinkerAttributes<T> (member);
-			attribute = attributes.FirstOrDefault ();
+			// This should only be called for attribute types which don't allow multiple attributes.
+			attribute = attributes.SingleOrDefault ();
 			return attribute != null;
 		}
 
--- a/src/linker/Linker/LinkerAttributesInformation.cs
+++ b/src/linker/Linker/LinkerAttributesInformation.cs
@@ -60,6 +60,7 @@ namespace Mono.Linker
 						continue;
 
 					attributeValue = new RemoveAttributeInstancesAttribute (customAttribute.ConstructorArguments);
+					allowMultiple = true;
 					break;
 				default:
 					continue;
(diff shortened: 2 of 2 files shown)


tlakollo (approved):
LGTM

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresViaXml.cs
```diff
@@ -25,15 +25,17 @@ class RequiresViaXml
 		// The second attribute is added through link attribute XML
 		[RequiresUnreferencedCode ("Message for --MethodWithDuplicateRequiresAttribute--")]
 		[ExpectedWarning ("IL2027", "RequiresUnreferencedCodeAttribute", nameof (MethodWithDuplicateRequiresAttribute), ProducedBy = ProducedBy.Trimmer)]
-		[ExpectedWarning ("IL2027", "RequiresUnreferencedCodeAttribute", nameof (MethodWithDuplicateRequiresAttribute), ProducedBy = ProducedBy.Trimmer)]
 		static void MethodWithDuplicateRequiresAttribute ()
 		{
 		}
 
+		[ExpectedWarning ("IL2026", "--MethodWithDuplicateRequiresAttribute--")]
 		[ExpectedWarning ("IL2026", "--MethodWithDuplicateRequiresAttribute--")]
```
vitek-karas, line 33: Could you modify the RUC message so that it's clear that one of the warnings comes from the attribute in code and the other from the attribute in XML?

src/linker/Linker/Annotations.cs
```diff
@@ -583,10 +582,6 @@ public bool SetPreservedStaticCtor (TypeDefinition type)
 		public bool TryGetLinkerAttribute<T> (IMemberDefinition member, [NotNullWhen (returnValue: true)] out T? attribute) where T : Attribute
 		{
 			var attributes = GetLinkerAttributes<T> (member);
-			if (attributes.Count () > 1) {
```
vitek-karas, line 586: We should probably at least assert that there's only one such attribute. For example DynamicaDependency can still occur multiple times, but this method doesn't handle it correctly - basically it should never be called for attributes which allow multiple instances per member.

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresViaXml.cs
vitek-karas: OK - got it

PR #2782, opened 2022-05-03
Record trim analysis patterns in scanner

sbomer:
This allows running the scanner first with warnings disabled, then "replaying" the collected patterns to produce warnings. This will enable running the scanner multiple times for a compiler-generated method.

It basically works the same as the trim patterns in the analyzer. The main difference is that it also can enable
or disable marking - so during the scan, it doesn't do marking or report warnings. Then once the scan is complete,
it marks and produces warnings for the recorder call/assignment patterns. The exception is for the DAM type hierarchy
logic which will mark and warn during the scan, because it has its own cache that prevents doing any extra work the
second time around. (See https://github.com/dotnet/linker/pull/2777#discussion_r864004887 for more context.)

This includes a few small changes to pass around `DiagnosticContext` to a few places that can produce warnings during the scan, such as resolving type names.

### DIFF
--- a/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
@@ -255,6 +255,7 @@ namespace Mono.Linker.Dataflow
 			}
 
 			TrimAnalysisPatterns.Add (new TrimAnalysisMethodCallPattern (
+				operation,
 				calledMethod,
 				instanceValue,
 				arguments,
@@ -263,6 +264,7 @@ namespace Mono.Linker.Dataflow
 
 			var diagnosticContext = new DiagnosticContext (_origin, diagnosticsEnabled: false, _context);
 			return HandleCall (
+				operation,
 				calledMethod,
 				instanceValue,
 				arguments,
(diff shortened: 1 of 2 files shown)


src/linker/Linker.Dataflow/TrimAnalysisPatternStore.cs
```diff
@@ -0,0 +1,45 @@
...
+		}
+
+		public void Add (TrimAnalysisAssignmentPattern pattern, bool isReturnValue)
+		{
+			// In the linker, each pattern should have a unique origin (which has ILOffset)
+			// but we don't track the correct ILOffset for return instructions.
+			// https://github.com/dotnet/linker/issues/2778
+			// For now, work around it with a separate bit.
+			AssignmentPatterns.Add ((pattern.Origin, isReturnValue), pattern);
+		}
```
vitek-karas, line 29: We should be able to deduce the "returnvalue" bit from the assignment, as the target should be MethodReturnValue.
Also in the linker the target can only be a single value 

No need for the additional parameter.

vitek-karas (commented):
Did you measure perf impact? Since we're now running some of the code twice.
It's not necessarily to gate this change, but to have a feeling how bad it's going to be and if we should do something about it.

jtschuster (approved):
LGTM

PR #2783, opened 2022-05-05
Fix NullReferenceException when sweeping unused static interface

vitek-karas:
The case here is if the static interface itself is actually used, but the method on it is not and the implementation of that method is directly referecned. In that case we should remove the iface method, but keep everything else.
This can cause (depends on order) a NRE in the sweep step since the removed iface method is still in the list of overrides for the implementation method.

This is blocking https://github.com/dotnet/runtime/pull/67771.

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/InterfaceVariants.cs
+++ b/test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/InterfaceVariants.cs
@@ -95,7 +95,7 @@ namespace Mono.Linker.Tests.Cases.Inheritance.Interfaces
 			}
 
 			[Kept]
-			[KeptInterface(typeof(IStaticInterfaceMethodUnused))]
+			[KeptInterface (typeof (IStaticInterfaceMethodUnused))]
 			internal class InterfaceMethodUnused : IStaticInterfaceMethodUnused, IStaticInterfaceUnused
 			{
 				public static void InterfaceUsedMethodNot () { }
@@ -104,7 +104,7 @@ namespace Mono.Linker.Tests.Cases.Inheritance.Interfaces
 			}
 
 			[Kept]
-			public static void Test()
+			public static void Test ()
 			{
 				InterfaceMethodUsedThroughImplementation.InterfaceUsedMethodNot ();
 				InterfaceMethodUsedThroughImplementation.InterfaceAndMethodNoUsed ();


vitek-karas:
There's some additional cleanup we should do, but I don't want to block the fix on it so that we can unblock the runtime repo.

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -472,7 +472,13 @@ protected virtual void SweepMethods (Collection<MethodDefinition> methods)
 		void SweepOverrides (MethodDefinition method)
 		{
 			for (int i = 0; i < method.Overrides.Count;) {
```
agocke, line 474: Is this overrides of the target method? Or methods the target overrides?

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -472,7 +472,13 @@ protected virtual void SweepMethods (Collection<MethodDefinition> methods)
...
-				if (Context.Resolve (method.Overrides[i]) is MethodDefinition ov && ShouldRemove (ov) && !IgnoreScope (ov.DeclaringType.Scope))
+				// We can't rely on the context resolution cache anymore, since it may remember methods which are already removed
+				// So call the direct Resolve here and avoid the cache.
+				// It can still happen that the Resolve will return a removed method (this happens if the Overrides collection stores actual
+				// MethodDefinition and not MethodReference, in which case the Resolve is just "return this" and doesn't check the existence of the method
+				// in any way). In such case the Declaring type of the method is null. And since it has been removed the override pointing
+				// to it should be removed as well.
```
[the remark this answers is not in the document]
jtschuster: Also confused on why the ShouldRemove doesn't catch any of this. Also, would the DeclaringType null check catch the issues with the Resolve cache returning removed methods?

src/linker/Linker.Steps/SweepStep.cs
jtschuster: This is the methods the target overrides.

src/linker/Linker.Steps/SweepStep.cs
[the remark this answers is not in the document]
vitek-karas: The `ShouldRemove` will return `true` but the condition is AND - so all of the things in it must be true. Which means even if `ShouldRemove` returns `true` we need to check the scope - and in order to do that, we need to get the declaring type - which can be null.

The null check would technically catch the wrong resolve as well, but it felt cleaner to do the resolve correctly. If you look at this file near the end there's a similar problem with type refs and resolve - it has a big comment as well (and needs a cleanup as the bug mentioned there is effectively fixed and so on).

src/linker/Linker.Steps/SweepStep.cs
vitek-karas: To show it on an example:
* `method` is typically `ImplementationClass.StaticInterfaceMethod`
* `override` is typically `IStaticInterface.StaticInterfaceMethod`.

PR #2786, opened 2022-05-09
Track attributes through Roslyn-generated types

agocke:
Type parameters for Roslyn-generated types are all alpha renames, so
there should be a unique original type parameter to map back to, which
can be used to get the original user attributes applied to the type
parameter. This change tries to build a map back from types to
constructor calls, which should map to the original type parameter annotations.

Fixes https://github.com/dotnet/linker/issues/2785

### DIFF
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -392,7 +392,7 @@ namespace ILLink.Shared.TrimAnalysis
 			if (!(CompilerGeneratedNames.IsStateMachineType (typeDef.Name) || CompilerGeneratedNames.IsLambdaDisplayClass (typeDef.Name))) {
 				return null;
 			}
-			var attrs = _context.CompilerGeneratedState.TryGetGeneratedTypeAttributes (typeDef);
+			var attrs = _context.CompilerGeneratedState.GetGeneratedTypeAttributes (typeDef);
 			Debug.Assert (attrs is null || attrs.Count == typeDef.GenericParameters.Count);
 			return attrs;
 		}
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -372,7 +372,7 @@ namespace ILLink.Shared.TrimAnalysis
 
 			DynamicallyAccessedMemberTypes[]? typeGenericParameterAnnotations = null;
 			if (type.HasGenericParameters) {
-				var attrs = TryGetGeneratedTypeAttributes(type);
+				var attrs = TryGetGeneratedTypeAttributes (type);
 				for (int genericParameterIndex = 0; genericParameterIndex < type.GenericParameters.Count; genericParameterIndex++) {
 					var provider = attrs?[genericParameterIndex] ?? type.GenericParameters[genericParameterIndex];
 					var annotation = GetMemberTypesForDynamicallyAccessedMembersAttribute (type, providerIfNotMember: provider);
@@ -389,11 +389,11 @@ namespace ILLink.Shared.TrimAnalysis
 
 		private IReadOnlyList<ICustomAttributeProvider>? TryGetGeneratedTypeAttributes (TypeDefinition typeDef)
 		{
-			if (!(CompilerGeneratedNames.IsStateMachineType(typeDef.Name) || CompilerGeneratedNames.IsLambdaDisplayClass(typeDef.Name))) {
+			if (!(CompilerGeneratedNames.IsStateMachineType (typeDef.Name) || CompilerGeneratedNames.IsLambdaDisplayClass (typeDef.Name))) {
 				return null;
 			}
 			var attrs = _context.CompilerGeneratedState.TryGetGeneratedTypeAttributes (typeDef);
-			Debug.Assert(attrs is null || attrs.Count == typeDef.GenericParameters.Count);
+			Debug.Assert (attrs is null || attrs.Count == typeDef.GenericParameters.Count);
 			return attrs;
 		}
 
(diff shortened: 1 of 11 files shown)


src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -165,6 +185,77 @@ void ProcessMethod (MethodDefinition method)
...
+
+			void MapGeneratedTypeTypeParameters (TypeDefinition stateMachineType)
+			{
+				var typeInfo = _generatedTypeToTypeArgumentInfo[stateMachineType];
+				if (typeInfo.OriginalAttributes is not null) {
+					return;
+				}
+				var method = typeInfo.CreatingMethod;
+				if (method.Body is { } body) {
+					var typeArgs = new ICustomAttributeProvider[stateMachineType.GenericParameters.Count];
```
vitek-karas, line 208: Nit: Move this after the ScanForInit and if - since it only needs to be allocated when we're actually going to use it.

src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -165,6 +185,77 @@ void ProcessMethod (MethodDefinition method)
...
+			}
+
+			GenericInstanceType? ScanForInit (TypeDefinition stateMachineType, MethodBody body)
+			{
+				foreach (var instr in body.Instructions) {
+					switch (instr.OpCode.Code) {
+					case Code.Initobj:
+					case Code.Newobj:
+						if (instr.Operand is MethodReference { DeclaringType: GenericInstanceType typeRef }
+							&& stateMachineType.MetadataToken == _context.TryResolve (typeRef)?.MetadataToken) {
```
vitek-karas, line 251: I think you can compare the TypeDefinition by reference - I think Cecil guarantees uniqueness. Note that we don't use special comparers when using TypeDefinition as a key in a Dictionary.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -386,6 +387,16 @@ TypeAnnotations BuildTypeAnnotations (TypeDefinition type)
 			return new TypeAnnotations (type, typeAnnotation, annotatedMethods.ToArray (), annotatedFields.ToArray (), typeGenericParameterAnnotations);
 		}
 
+		private IReadOnlyList<ICustomAttributeProvider>? TryGetGeneratedTypeAttributes (TypeDefinition typeDef)
```
vitek-karas, line 390: Nit: The way I know the rules if methods starts with "TryGet", it should return a bool and probably use out parameter to return the actual value. On the other hand "Get" method can return null if it makes sense as a missing value.

test/Mono.Linker.Tests.Cases/DataFlow/CompilerGeneratedStateMachine.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.Threading.Tasks;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+
+namespace Mono.Linker.Tests.Cases.DataFlow
+{
+	[ExpectedNoWarnings]
+	[SkipKeptItemsValidation]
+	class CompilerGeneratedStateMachine
+	{
+		public static void Main ()
```
vitek-karas, line 16: Nit: I would add couple more negative tests (which produce diagnostics) - especially in analyzers it's important since we give up on analysis if we don't understand something and won't produce any diagnostics. So the fact that a success test passed doesn't necessarily mean the analyzer worked.
Linker is the other way round, so success tests are much more "reliable".

vitek-karas (approved):

src/linker/Linker.Dataflow/FlowAnnotations.cs
agocke: Yeah, I admit that this is a personal thing -- I prefix with `Try` any time there's a potential for "no result." Happy to change it if that doesn't seem helpful to you, though.

vitek-karas: If the "no result" is an important bit I would go with TryGet returning a boolean - but again it's a subjective thing. That said - as is it would not pass FX API review ;-)

src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -75,6 +84,13 @@ void ProcessMethod (MethodDefinition method)
 						if (lambdaOrLocalFunction == null)
 							continue;
 
+						if (lambdaOrLocalFunction.IsConstructor && CompilerGeneratedNames.IsLambdaDisplayClass (lambdaOrLocalFunction.DeclaringType.Name)) {
```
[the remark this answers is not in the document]
agocke: I think there has to be a call to a constructor (or at least initobj) if the display class is generic, even if it's a struct. Otherwise, there's no substitution. I'll try to construct a test case.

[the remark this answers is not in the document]
agocke: Turns out, this is impossible in the current scope because there are no struct display classes for iterators or async methods.

src/linker/Linker/CompilerGeneratedState.cs
```diff
@@ -105,11 +121,15 @@ void ProcessMethod (MethodDefinition method)
...
+						// Initially fill the dictionary with all null type args. After we have fully filled out
+						// user methods, we'll fill in the type args
 						if (!_compilerGeneratedTypeToUserCodeMethod.TryAdd (stateMachineType, method)) {
 							var alreadyAssociatedMethod = _compilerGeneratedTypeToUserCodeMethod[stateMachineType];
 							_context.LogWarning (new MessageOrigin (method), DiagnosticId.MethodsAreAssociatedWithStateMachine, method.GetDisplayName (), alreadyAssociatedMethod.GetDisplayName (), stateMachineType.GetDisplayName ());
 						}
-
+						// Already warned above if multiple methods map to the same type
+						// Fill in null for argument providers now, the real providers will be filled in later
+						_ = _generatedTypeToTypeArgumentInfo.TryAdd (stateMachineType, new TypeArgumentInfo (method, null));
```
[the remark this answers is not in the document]
agocke: Happy to look into this but I think it's mostly equivalent to what I have now. I'm going to check this in to make Preview 5, but we can fix it up if necessary.

agocke:
Tested this change locally against dotnet/runtime and it didn't seem to have any problems.

PR #2788, opened 2022-05-09
Match analyzer and linker requires methods

tlakollo:
Match analyzer and linker requires methods to be more similar according to their behavior
Sync diagnostic Id in NativeAOT with the information in linker

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -183,7 +183,7 @@ namespace ILLink.RoslynAnalyzer
 							if (instanceCtor.Arity > 0)
 								continue;
 
-							if (instanceCtor.DoesMemberRequires (RequiresAttributeName, out var requiresAttribute) &&
+							if (instanceCtor.DoesMemberRequire (RequiresAttributeName, out var requiresAttribute) &&
 								VerifyAttributeArguments (requiresAttribute)) {
 								syntaxNodeAnalysisContext.ReportDiagnostic (Diagnostic.Create (RequiresDiagnosticRule,
 									syntaxNodeAnalysisContext.Node.GetLocation (),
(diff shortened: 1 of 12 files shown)


src/ILLink.Shared/DiagnosticId.cs
```diff
@@ -187,7 +187,11 @@ public enum DiagnosticId
 		// Dynamic code diagnostic ids.
 		RequiresDynamicCode = 3050,
 		RequiresDynamicCodeAttributeMismatch = 3051,
-		RequiresDynamicCodeOnStaticConstructor = 3052
+		RequiresDynamicCodeOnStaticConstructor = 3052,
+		COMInteropNotSupportedInFullAOT = 3052,
```
tlakollo, line 191: Notice that there is a collision on the code IL3052, I don't have a strong preference on which one should remain IL3052 and which one should change, but one should be changed.

vitek-karas: Did we ship one of them in .NET 6? If only one, then we should keep that one as is and change the new one.
If not... I guess just pick one...

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -183,7 +183,7 @@ public override void Initialize (AnalysisContext context)
 							if (instanceCtor.Arity > 0)
 								continue;
 
-							if (instanceCtor.TargetHasRequiresAttribute (RequiresAttributeName, out var requiresAttribute) &&
+							if (instanceCtor.DoesMemberRequires (RequiresAttributeName, out var requiresAttribute) &&
```
vitek-karas, line 186: Nit - not for this PR: The grammar of the new name is really weird - "Does member requires" feels wrong, I think it should read "Does member require <something>". I know it's how we have it everywhere, so not for this change, but maybe something to consider one day :-)

vitek-karas (approved):

src/ILLink.Shared/DiagnosticId.cs
tlakollo: COMInteropNotSupportedInFullAOT in NativeAOT seems to exist even before runtimelab project existed, so probably I will leave that as 3052 even if we haven't shipped NativeAOT yet

vitek-karas (approved):

PR #2791, opened 2022-05-11
Re-add static interface trimming with more testing

jtschuster:
Revert the commit that reverted static interface trimming and add more comprehensive tests.

I also found that sometimes `method.Overrides[i].Resolve()` would sometimes return null if it had been removed earlier, in which case we want to remove the override. The code has been updated to account for this case.

This is missing library tests, but I wanted to make sure that I don't miss any cases before I move to that. Please let me know if there are any other cases that needs to be tested.


src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -467,6 +469,32 @@ protected virtual void SweepMethods (Collection<MethodDefinition> methods)
...
+		{
+			for (int i = 0; i < method.Overrides.Count;) {
+				// We can't rely on the context resolution cache anymore, since it may remember methods which are already removed
+				// So call the direct Resolve here and avoid the cache.
+				// It can still happen that the Resolve will return a removed method (this happens if the Overrides collection stores actual
+				// MethodDefinition and not MethodReference, in which case the Resolve is just "return this" and doesn't check the existence of the method
+				// in any way). In such case the Declaring type of the method is null. And since it has been removed the override pointing
+				// to it should be removed as well.
+				// Resolve() may also return null if the method has been removed, in which case the override should be removed from the list
+				if (method.Overrides[i].Resolve () is not MethodDefinition ov || ShouldRemove (ov) && (ov.DeclaringType == null || !IgnoreScope (ov.DeclaringType.Scope)))
```
vitek-karas, line 482: Nit: Could you also explain in the comment about the ShouldRemove and the IgnoreScope?
Basically I think we should list all reasons a method override should be removed in an 'easy to understand list' - basically this but expanded:
* Doesn't resolve (couple of ways this can occur)
* Not marked (ShouldRemove), but only if coming from trimmed scope

I know that ideally the code should speak for itself, but this has proven to be really tricky, so having a good completely comment won't hurt.

test/Mono.Linker.Tests/TestCasesRunner/AssemblyChecker.cs
```diff
@@ -212,6 +213,35 @@ void VerifyInterfaces (TypeDefinition src, TypeDefinition linked)
...
+				.Where (ca => ca.AttributeType.Name == nameof (RemovedOverrideAttribute))
+				.Select (ca => (ca.ConstructorArguments[0].Value as TypeReference).FullName));
+			foreach (var expectedRemovedBaseType in expectedBaseTypesNotOverridden) {
+				Assert.IsTrue (originalBaseTypesOverridden.Contains (expectedRemovedBaseType),
+					$"Method {linked.FullName} was expected to remove override {expectedRemovedBaseType}::{linked.Name}, " +
+					$"but it wasn't in the unlinked assembly");
+				Assert.IsFalse (linkedBaseTypesOverridden.Contains (expectedRemovedBaseType),
+					$"Method {linked.FullName} was expected to not override {expectedRemovedBaseType}::{linked.Name}");
+			}
+		}
```
vitek-karas, line 243: We should also validate that all overrides on the linked method point to existing methods. For all methods (so no Kept attribute required). This is one of the bugs we ran into and it's hard to test directly (depends on ordering), so as a defense we should validate that we will never "corrupt" the IL that way (ideally Cecil would do this when writing the IL out, but apparently it doesn't).

src/linker/Linker.Steps/SweepStep.cs
jtschuster: I've updated the comment here, let me know if it's clear and understandable.

test/ILLink.RoslynAnalyzer.Tests/generated/ILLink.RoslynAnalyzer.Tests.Generator/ILLink.RoslynAnalyzer.Tests.TestCaseGenerator/Inheritance.Interfaces.StaticInterfaceMethodsTests.g.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+	public sealed partial class StaticInterfaceMethodsTests : LinkerTestBase
+	{
+
+		protected override string TestSuiteName => "Inheritance.Interfaces.StaticInterfaceMethods";
+
+		[Fact]
+		public Task StaticAbstractInterfaceMethods ()
+		{
+			return RunTest (allowMissingWarnings: true);
```
agocke, line 15: Do these need to be moved into the source to remove the `allowMissingWarnings` spec? I presume that's not what we want in this case.

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/StaticInterfaceMethods/StaticAbstractInterfaceMethodsLibrary.cs
```diff
@@ -0,0 +1,897 @@
...
+			public class UnusedIUsedThroughConstrainedTypeMethods : IUsedThroughConstrainedType, IUsedThroughConstrainedTypeInternal
+			{
+				[Kept]
+				[KeptOverride (typeof (IUsedThroughConstrainedType))]
+				[RemovedOverride (typeof (IUsedThroughConstrainedTypeInternal))]
+				public static int UsedThroughConstrainedType () => 0;
+			}
+
+			// Should this be kept?
+			private class UnusedIUsedThroughConstrainedTypeMethodsPrivate : IUsedThroughConstrainedType, IUsedThroughConstrainedTypeInternal
```
agocke, line 71: I couldn't find any references to it, so it seems like it should be removed. Why would it be kept?

agocke (commented):
LGTM aside from a few small questions

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -467,6 +469,38 @@ protected virtual void SweepMethods (Collection<MethodDefinition> methods)
...
+				//		This can happen for a couple of reasons, but it indicates the method isn't in the final assembly.
+				//		Resolve also may return a removed value if method.Overrides[i] is a MethodDefinition. In this case, Resolve short circuits and returns `this`.
+				//	OR
+				//	ov.DeclaringType is null
+				//		ov.DeclaringType may be null if Resolve short circuited and returned a removed method. In this case, we want to remove the override.
+				//	OR
+				//	ov is in a `link` scope and is unmarked
+				//		ShouldRemove returns true if the method is unmarked, but we also We need to make sure the override is in a link scope.
+				//		Only things in a link scope are marked, so ShouldRemove is only valid for items in a `link` scope.
+				if (method.Overrides[i].Resolve () is not MethodDefinition ov || ov.DeclaringType is null || (IsLinkScope (ov.DeclaringType.Scope) && ShouldRemove (ov)))
```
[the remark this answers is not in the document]
vitek-karas: That is correct - if the method in the overrides is actually a `MethoDefinition` instance, than `Resolve` doesn't do anything (`return this;`) and it will work regardless if the method is "attached" or not. If it's removed it would have `DeclaringType == null` though - so we have to check that.
Unlike the case where the method in the overrides is actually a true `MethodReference` (not a def) in which case `Resolve` would return `null` if the method has been removed.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3017,8 +3021,18 @@ protected virtual void ProcessMethod (MethodDefinition method, in DependencyInfo
...
+			// Mark overridden methods except for static interface methods
 			if (method.HasOverrides) {
 				foreach (MethodReference ov in method.Overrides) {
+					// Method implementing a static interface method will have an override to it - note nonstatic methods usually don't unless they're explicit.
+					// Calling the implementation method directly has no impact on the interface, and as such it should not mark the interface or its method.
+					// Only if the interface method is referenced, then all the methods which implemented must be kept, but not the other way round.
+					if (Context.Resolve (ov)?.IsStatic == true
+						&& Context.Resolve (ov.DeclaringType)?.IsInterface == true
+						// Public methods can be called directly on the type. If it non-public it is an explicit implementation and if the method is marked we need to mark the override.
+						&& method.IsPublic) {
```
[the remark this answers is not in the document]
vitek-karas: I would also prefer we remove the IsPublic check. What if it's internal and there InternalsVisibleTo applied - then it is effectively public. Visibility should have basically no effect here. The explicit implementation methods should be never accessed directly as C# doesn't let you do that - so the only way is via reflection, and in that case I think it's OK to overmark, definitely not something I would optimize for.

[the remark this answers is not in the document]
vitek-karas: Technically speaking we should be able to avoid marking overrides for interface methods completely (not just for static interfaces) - but that's a bigger/unrelated change. (And it is unlikely to actually do much since non-static interfaces do not use overrides in C#, so the cases where this would happen are very limitted in real world).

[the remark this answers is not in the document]
jtschuster: From what I can tell, a method implementing an interface method must be public to implicitly implement the method or private for an explicit implementation, so I don't think we'd run into the internal with InternalsVisibleTo issue.

I think I could avoid of complication by looking from base -> override instead of override -> base like I do here. I didn't see the `AnnotationStore.GetOverrides` which does the inverse of Cecil's `.Overrides`. If I check if a method is a static interface method and kept, I can mark the methods in `GetOverrides` as well as the interface implementation.

> I would have roughly expected that logic elsewhere, in ProcessMarkedTypesWithInterfaces
I agree, that would make sense to me, but at the moment that method returns early if the method isn't instantiated (or relevant to variant casting). If we move the logic for static interface methods, we'd have to iterate through each method to check if any override a static interface method before we can say we don't need an interface implementation. I think it could be worth it to keep the related logic in the same place though.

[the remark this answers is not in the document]
jtschuster: I think here we can keep marking interface implementations for explicitly implemented non-static interface methods. I moved the logic to mark static interface method overrides from base -> implementation to ProcessOverrides. One issue with moving all marking of interface implementations to ProcessMarkedTypesWithInterfaces is that this test: https://github.com/dotnet/linker/blob/main/test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/OnReferenceType/NoKeptCtorButInterfaceNeeded/NestedInterfacesWithExplicit1.cs#L48
I think this makes the most sense to me, but I can rework if we think another way would be better.

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/StaticInterfaceMethods/StaticAbstractInterfaceMethodsLibrary.cs
jtschuster: If I remember my thought process, I was wondering if the static interface method being kept would cause this to be kept. I don't think that makes sense and I agree it should be removed though.

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/InterfaceVariants.cs
```diff
@@ -71,41 +71,32 @@ internal class ImplementsUnusedStaticInterface
...
 
 				[Kept]
+				[RemovedOverride (typeof (IStaticInterfaceUnused))]
 				public static void InterfaceAndMethodNoUsed () { }
 			}
 
 			[Kept]
-			[KeptInterface (typeof (IStaticInterfaceMethodUnused))]
-			[KeptInterface (typeof (IStaticInterfaceUnused))]
+			[KeptInterface (typeof (IStaticInterfaceMethodUnused))] // Bug
```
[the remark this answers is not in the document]
jtschuster: Changed how that method is kept and added a comment to explain it.

jtschuster:
I've updated the marking of static interface methods to be like the virtual method marking, where we store each static interface method that is marked, then mark the overriding methods later. The interface implementations are then marked in ProcessOverride.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -726,6 +752,11 @@ void ProcessOverride (OverrideInformation overrideInformation)
 				MarkMethod (method, new DependencyInfo (DependencyKind.Override, @base), ScopeStack.CurrentScope.Origin);
 			}
 
+			// We need to mark the interface implementation for static interface methods
+			// Explicit interface method implementations already mark the interface implementation in ProcessMethod
+			if (overrideInformation.IsStaticInterfaceMethodPair)
```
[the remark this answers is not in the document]
jtschuster: Thanks for pointing this out! There was a bug here that needed fixing. I moved the interfaceImpl marking to ProcessStaticInterfaceMethods instead.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -738,6 +742,10 @@ void ProcessOverride (OverrideInformation overrideInformation)
 			if (IsInterfaceOverrideThatDoesNotNeedMarked (overrideInformation, isInstantiated))
 				return;
 
+			// Interface static veitual methods will be abstract and will also by pass this check to get marked
```
vitek-karas, line 745:
```suggestion
			// Interface static virtual methods will be abstract and will also bypass this check to get marked
```

src/linker/Linker/OverrideInformation.cs
```diff
@@ -10,17 +10,22 @@ namespace Mono.Linker
...
-			Base = @base;
-			Override = @override;
+			_pair = new OverridePair (@base, @override);
 			MatchingInterfaceImplementation = matchingInterfaceImplementation;
 			this.resolver = resolver;
 		}
 
-		public MethodDefinition Base { get; }
-		public MethodDefinition Override { get; }
+		public record struct OverridePair (MethodDefinition Base, MethodDefinition Override)
```
vitek-karas, line 22: Can it also be readonly ?

vitek-karas (approved):

PR #2792, opened 2022-05-12
Scan compiler-generated methods as a group

sbomer:
~Still a WIP. Needs more tests, optimization, and cleanup, but functionally is mostly working. Don't look too closely at the code yet, but I wanted to get feedback on the design which I've tried to describe here @vitek-karas @agocke:~

This scans compiler-generated methods together as a group when marking the corresponding user code.

There are two scans that we do:
1. An initial scan to determine whether we need to run the dataflow scanner. It is also what marks static dependencies of the method IL.
2. The full dataflow scan (only if the initial scan says we need to). This produces warnings and marks reflection dependencies.

Both scans are now done for the group of compiler-generated methods when marking the user code. For now, we only do the dataflow scan once per method, but in a later change, we will need to allow re-scanning compiler-generated callees (as part of the full scan for the corresponding user code) to properly track captured state.

When compiler-generated code is accessed via reflection, we now don't do a dataflow scan because we don't have the context that might be captured from user code. This also means that reflection-dependencies of the compiler-generated code aren't kept, and dataflow warnings aren't produced, unless the code is reached through the corresponding user method. To guard against this, there are new warnings on reflection access to compiler-generated code. (edit: The reflection access warnings are now only for compiler-generated code which would normally require the reflection method body scanner. This is a heuristic meant to conservatively approximate "this compiler-generated code would produce dataflow warnings if invoked by reflection with an unknown context".)

~Currently, the initial scan (1.) may be done multiple times for a compiler-generated method, once when marking the corresponding user code, and again when marking the compiler-generated code. It only produces warnings when marking for the user code. We can't just fold the compiler-generated code marking into the user code marking, because the compiler-generated code might independently be accessed via reflection, and we still want to keep IL dependencies in that case. I initially hesitated to introduce another cache since I assumed the initial scan was supposed to be inexpensive, but now I am thinking it would make sense to cache even if only to simplify the code. Curious if you have an opinion @vitek-karas.~

edit: I added a cache to ensure that the initial scan 1. is still only done once per method (including for compiler-generated methods). Warnings produced as a result of this initial scan will be produced whether the code is accessed by reflection or via a call from user code.

### DIFF
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -45,6 +45,7 @@ namespace Mono.Linker.Dataflow
 	{
 		protected readonly LinkContext _context;
 		protected static ValueSetLattice<SingleValue> MultiValueLattice => default;
+		protected static ValueSetLattice<MethodProxy> MethodLattice => default;
 
 		protected MethodBodyScanner (LinkContext context)
 		{
@@ -225,7 +226,7 @@ namespace Mono.Linker.Dataflow
 		// reachable from it.
 		public virtual void InterproceduralScan (MethodBody methodBody)
 		{
-			var methodsInGroup = new ValueSet<MethodDefinition> (methodBody.Method);
+			var methodsInGroup = new ValueSet<MethodProxy> (methodBody.Method);
 
 			// Optimization to prevent multiple scans of a method.
 			// Eventually we will need to allow re-scanning in some cases, for example
@@ -268,8 +269,9 @@ namespace Mono.Linker.Dataflow
 			bool TryGetNextMethodToScan ([NotNullWhen (true)] out MethodDefinition? method)
 			{
 				foreach (var candidate in methodsInGroup) {
-					if (!scannedMethods.Contains (candidate) && candidate.HasBody) {
-						method = candidate;
+					var candidateMethod = candidate.Method;
+					if (!scannedMethods.Contains (candidateMethod) && candidateMethod.HasBody) {
+						method = candidateMethod;
 						return true;
 					}
 				}
@@ -278,7 +280,7 @@ namespace Mono.Linker.Dataflow
 			}
 		}
 
-		void TrackNestedFunctionReference (MethodReference referencedMethod, ref ValueSet<MethodDefinition> methodsInGroup)
+		void TrackNestedFunctionReference (MethodReference referencedMethod, ref ValueSet<MethodProxy> methodsInGroup)
 		{
 			if (_context.TryResolve (referencedMethod) is not MethodDefinition method)
 				return;
@@ -286,13 +288,10 @@ namespace Mono.Linker.Dataflow
 			if (!CompilerGeneratedNames.IsLambdaOrLocalFunction (method.Name))
 				return;
 
-			// Work around the fact that we can't mutate the ValueSet in-place because it's a readonly struct.
-			var methods = new HashSet<MethodDefinition> (methodsInGroup);
-			methods.Add (method);
-			methodsInGroup = new ValueSet<MethodDefinition> (methods);
+			methodsInGroup = MethodLattice.Meet (methodsInGroup, new (method));
 		}
 
-		protected virtual void Scan (MethodBody methodBody, ref ValueSet<MethodDefinition> methodsInGroup)
+		protected virtual void Scan (MethodBody methodBody, ref ValueSet<MethodProxy> methodsInGroup)
 		{
 			MethodDefinition thisMethod = methodBody.Method;
 
--- a/src/linker/Linker.Dataflow/MethodProxy.cs
+++ b/src/linker/Linker.Dataflow/MethodProxy.cs
@@ -1,13 +1,14 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Immutable;
 using Mono.Cecil;
 using Mono.Linker;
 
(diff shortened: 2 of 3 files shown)


src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -207,7 +208,71 @@ public int MoveNext (Instruction op)
...
+				scannedMethods.Add (methodToScan);
+			}
+
+			if (_context.CompilerGeneratedState.TryGetCompilerGeneratedCalleesForUserMethod (methodBody.Method, out List<IMemberDefinition>? compilerGeneratedCallees)) {
+				// should be the same set.
+				var calleeMethods = compilerGeneratedCallees.OfType<MethodDefinition> ();
+				Debug.Assert (calleeMethods.Count() == scannedMethods.Count - 1);
+				foreach (var method in calleeMethods)
+					Debug.Assert (scannedMethods.Contains (method));
+			}
```
vitek-karas, line 257: Should this whole thing be under `#ifdef DEBUG`?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2195,7 +2202,7 @@ protected int MarkNamedMethod (TypeDefinition type, string method_name, in Depen
 				if (method.Name != method_name)
 					continue;
 
-				MarkMethod (method, reason, ScopeStack.CurrentScope.Origin);
+				MarkMethod (method, reason, ScopeStack.CurrentScope.Origin, enableReflectionPatternReporting: true);
```
vitek-karas, line 2205: There are a lot of places where we pass in `true` - I would probably make that a default parameter value. I know it's a bit error prone (easy to forget about), but I would probably still prefer it over the much longer text everywhere else. Thoughts?

vitek-karas:
> I initially hesitated to introduce another cache since I assumed the initial scan was supposed to be inexpensive, but now I am thinking it would make sense to cache even if only to simplify the code.

I'm probably slow today - can you please expand what kind of cache you're talking about and what would it simplify?

Re reflection access to compiler generated methods:
I'm sort of split on this.
On one hand, if we generate a warning then it sort of doesn't matter what we do with the rest, so it would be OK to not mark the method in that case.
On the other hand, this would go against how linker operates today (we mark even though we're not sure about it in all other cases, basically we overmark members, we don't overmark types). But more importantly this would make it effectively impossible to handle some cases, see example below.

I think with this we're going to potentially have a problem:
```C#
class TestType
{
    public void MethodWithLocalFunction() { Local(); void Local() {} }
    private void InitMethod() {}
}

void InitializeEverything([DAM(All)] Type type)
{
    // Get all private methods which have a name starting with `Init`
    // Call them
    ....

    // There will be no warnings in this method as it's perfectly correct.
}

void Test()
{
    // This will produce a warning since we're accessing compiler generated code (the local function) potentially via reflection
    DoMyPrivateReflectionStuff(typeof(TestType));
}
```

The `InitializeEverything` method knows that it actually won't do anything bad with compiler generated code (since those methods have unspeakable names and it won't match them), but there's no way to tell that to the caller. And the caller is the one being blamed for this.

One potential "fix" for this could be:
If a compiler generate method is marked via reflection - play dumb and simply mark it the way we do today. It may produce warnings which would not happen when it's accessed normally from the user code, but so be it.
The idea behind this is:
* I fear it's going to be relatively common to apply All or similar "bad" annotations to "random" types. It's already a problem with RUC methods, but there's nothing we can do about that.
* Compiler generated code is relatively common and will probably expand in the future - so the new warning would happen more frequently.
* It's pretty rare for compiler generated code to do something with reflection - most such code will NOT do that. So if we would detect that case - we would not need to warn.
* Even today without cross-method analysis, we need to guarantee that dangerous code produces warnings - so if a compiler generated code does something weird, we should warn already. So the above should not break the basic "0 warnings => works" premise, but should be able to avoid new warnings in lot of cases.

vitek-karas:
I agree that the warnings could be wrong - as in more than should be there, but that's only for the case where it's actually using some reflection, which I think is rare. I think we should prioritize the experience for the much more common case where there's no reflection inside the compiler generated code.

vitek-karas:
Technically possible (but probably a bit complex): If we detect reflection access to compiler generated code, run the analysis on the compiler generated code but "suppress" warnings, only capture a single boolean if there is a warning or not. And if there's none -> no warning at all. If there is something, issue a warning similar to what you added here - on the callsite.

Didn't look into the implementation if this is reasonable. It might be - the reflection access is granular -> we will be accessing things one item at a time (one field, one method, ...). We can re-run the body scanner for each method in this case with a custom diagnostic context. It will not be perfect, but should be good enough - there will likely still be cases where marking the method in question will eventually produce warnings somewhere deeper. We can always try to improve, even though it's somewhat of a breaking change (we would be moving the place where a warning comes from, and also changing the warning code).

MichalStrehovsky:
> Was this an argument against producing these warnings or a general observation? I think the same would be true of DAM-annotated methods (we would need to warn on any methods in a fully-kept assembly that have DAM annotations). I'm not sure how big of a problem that is, but it seems like the compiler-generated code warnings are no worse off.

Just a general observation. You're right that it affects user code as well.

> I like the idea to treat descriptors/arguments as keeping things for reflection. If someone is going out of their way to preserve things we should assume the worst. This would also encourage people to turn on more aggressive trimming (since they might otherwise get more warnings from unused parts of the rooted assemblies).

We should guide people towards:
* If you `[DAM]` anything in the assembly, you should also set `IsTrimmable=true`
* Never set `TrimMode` to anything but `link`

Then it should be less of a nuisance.

vitek-karas (approved):
Mostly cosmetic comments - looks good, great work.

PR #2793, opened 2022-05-12
Override sanity check

jtschuster:
This adds back support for checking the presence / removal of `.override`s / `MethodImpl`'s on methods, and adds a sanity check to make sure that `MethodImpl`s point to an interface that the type implements or a base type that is inherited from.

There is currently a case in the implementation of static interface removal where we remove an interface implementation, but leave a method with an override pointing to a method on that interface, and this check will catch that.


vitek-karas (approved):
Looks good - I'm surprised (in a good way) that this doesn't break any tests :-)

PR #2794, opened 2022-05-13
Retarget product binaries to .NET 6

vitek-karas:
So that they can run in runtime repo which has .NET 7 Preview 3 still.

We need .NET 7 Preview 4 in tests as they rely on some new attributes in the framework only available in that release. And it would be difficult to use different SDK for tests and product.


marek-safar:
Could you keep using a property everywhere and only change the places to use a new name?

marek-safar (approved):

PR #2795, opened 2022-05-13
Don't warn on enum switch if named values are exhaustive

jtschuster:
In some cases, it is helpful to only cover named values of enums in a switch without a catch-all at the bottom. Since the warning code is different for named and unnamed variants we can disable the unnamed variants warning. This way, if all named variants are accounted for without the catch-all and a new variant is added it will warn in that location rather than only showing up at runtime.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -116,6 +116,9 @@ dotnet_diagnostic.IDE0082.severity = warning
 # IDE0110: Remove unnecessary discard
 dotnet_diagnostic.IDE0110.severity = warning
 
+# CS8524: Switch doesn't handle unnamed enum values
+dotnet_diagnostic.CS8524.severity = none
+
 ## CA analyzer rules
 dotnet_analyzer_diagnostic.category-performance.severity = warning
 dotnet_analyzer_diagnostic.category-maintainability.severity = warning
--- a/.editorconfig
+++ b/.editorconfig
@@ -116,9 +116,6 @@ dotnet_diagnostic.IDE0082.severity = warning
 # IDE0110: Remove unnecessary discard
 dotnet_diagnostic.IDE0110.severity = warning
 
-# CS8524: Switch doesn't handle unnamed enum values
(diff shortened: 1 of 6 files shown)


.editorconfig
```diff
@@ -116,6 +116,9 @@ dotnet_diagnostic.IDE0082.severity = warning
 # IDE0110: Remove unnecessary discard
 dotnet_diagnostic.IDE0110.severity = warning
 
+# CS8524: Switch doesn't handle unnamed enum values
+dotnet_diagnostic.CS8524.severity = none
```
marek-safar, line 120: I don't like disabling this globally. If there is a reason why the warning should be suppressed please use #pragma

marek-safar (changes requested):

agocke:
I agree with @marek-safar on this one. I think the real value from completeness checking is actually the warning: if the switch is not complete, a warning occurs. In this case, no warning would occur even if you're missing a legitimate case.

I think this should be fixed at the C# level (and I have a proposal for it somewhere), but in the meantime I would live with the warning. I would silence it in code by using a helper like,

```C#
var x = e switch
{
   E.Item1 => ...,
   E.Item2 => ...
   _ => throw ExceptionUtilities.UnexpectedValue(e)
};
```

agocke:
OK, I misunderstood -- turns out Roslyn generates two warnings here, one for switches which are incomplete due to not covering all integer enum cases, and one for everything else. So when we're silencing this warning, we're just silencing the enum-cast case.

I would be fine taking this change, but I also think there would be a good improvement here if we could use an analyzer to tage enums as not-convertible and suppress only in that case. I've made an analyzer to do this [https://www.nuget.org/packages/StaticCS/](https://www.nuget.org/packages/StaticCS/), I just need to figure out if we can reference it in the build.

agocke (approved):
This should be all fixed now. No need to disable the warning everywhere, but no warnings on "Closed" enums

PR #2796, opened 2022-05-13
Split up ReflectionMethodBodyScanner

sbomer:
This separates the pieces which require the dataflow analysis from those that don't.
Also includes a few other small cleanup changes (no functional changes).
Please let me know if you have better ideas for a name than `FlowInsensitiveReflectionScanner`.

This addresses the feedback in https://github.com/dotnet/linker/pull/2782#discussion_r864587118

### DIFF
--- a/src/linker/Linker.Dataflow/AttributeDataFlow.cs
+++ b/src/linker/Linker.Dataflow/AttributeDataFlow.cs
@@ -3,9 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
-using ILLink.Shared.DataFlow;
 using ILLink.Shared.TrimAnalysis;
 using Mono.Cecil;
 using Mono.Linker.Steps;
@@ -13,13 +11,13 @@ using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.Single
 
 namespace Mono.Linker.Dataflow
 {
-	public class FlowInsensitiveReflectionScanner
+	public readonly struct AttributeDataFlow
 	{
 		readonly LinkContext _context;
 		readonly MarkStep _markStep;
 		readonly MessageOrigin _origin;
 
-		public FlowInsensitiveReflectionScanner (LinkContext context, MarkStep markStep, in MessageOrigin origin)
+		public AttributeDataFlow (LinkContext context, MarkStep markStep, in MessageOrigin origin)
 		{
 			_context = context;
 			_markStep = markStep;
@@ -31,7 +29,7 @@ namespace Mono.Linker.Dataflow
 			for (int i = 0; i < method.Parameters.Count; i++) {
 				var parameterValue = _context.Annotations.FlowAnnotations.GetMethodParameterValue (method, i);
 				if (parameterValue.DynamicallyAccessedMemberTypes != DynamicallyAccessedMemberTypes.None) {
-					MultiValue value = GetValueNodeForCustomAttributeArgument (arguments[i]);
+					MultiValue value = GetValueForCustomAttributeArgument (arguments[i]);
 					var diagnosticContext = new DiagnosticContext (_origin, diagnosticsEnabled: true, _context);
 					RequireDynamicallyAccessedMembers (diagnosticContext, value, parameterValue);
 				}
@@ -40,7 +38,7 @@ namespace Mono.Linker.Dataflow
 
 		public void ProcessAttributeDataflow (FieldDefinition field, CustomAttributeArgument value)
 		{
-			MultiValue valueNode = GetValueNodeForCustomAttributeArgument (value);
+			MultiValue valueNode = GetValueForCustomAttributeArgument (value);
 			var fieldValueCandidate = _context.Annotations.FlowAnnotations.GetFieldValue (field);
 			if (fieldValueCandidate is not ValueWithDynamicallyAccessedMembers fieldValue)
 				return;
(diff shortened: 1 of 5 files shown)


src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -670,29 +672,76 @@ public FieldAnnotation (FieldDefinition field, DynamicallyAccessedMemberTypes an
...
+			case "Empty" when field.DeclaringType.IsTypeOf (WellKnownType.System_String): {
+					return new KnownStringValue (string.Empty);
+				}
+
+			default: {
+					DynamicallyAccessedMemberTypes memberTypes = GetFieldAnnotation (field);
+					return new FieldValue (field.FieldType.ResolveToTypeDefinition (_context), field, memberTypes);
+				}
+			}
+		}
```
tlakollo, line 717: Nit: Maybe use the new switch statement?
```C#
internal SingleValue GetFieldValue(FieldDefinition field) =>
field.Name switch
{
	"EmptyTypes" when field.DeclarinType.IsTypeOf(WellKnownType.System_Type) => ArrayValue.Create(0, field.DeclaringType),
	"Empty" when field.DeclaringType.IsTypeOf(WellKnownType.System_String) => new KnownStringValue(string.Empty),
	_ => new FieldValue(field.FieldType.ResolveToTypeDefinition(_context), field, GetFieldAnnotation(field))

};
```

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -670,29 +672,76 @@ public FieldAnnotation (FieldDefinition field, DynamicallyAccessedMemberTypes an
...
 
 		internal partial MethodParameterValue GetMethodParameterValue (MethodProxy method, int parameterIndex, DynamicallyAccessedMemberTypes dynamicallyAccessedMemberTypes)
 			=> new (method.Method.Parameters[parameterIndex].ParameterType.ResolveToTypeDefinition (_context), method.Method, parameterIndex, dynamicallyAccessedMemberTypes);
 
 		internal partial MethodParameterValue GetMethodParameterValue (MethodProxy method, int parameterIndex)
-			=> GetMethodParameterValue (method, parameterIndex, _context.Annotations.FlowAnnotations.GetParameterAnnotation (method.Method, parameterIndex + (method.IsStatic () ? 0 : 1)));
+			=> GetMethodParameterValue (method, parameterIndex, GetParameterAnnotation (method.Method, parameterIndex + (method.IsStatic () ? 0 : 1)));
+
+		// Linker-specific dataflow value creation. Eventually more of these should be shared.
+
```
tlakollo, line 701: Nit: extra line

tlakollo (approved):
In general looks good to me, I agree that FlowInsensitiveReflectionScanner is a weird name but so far I haven't come up with something different. Will reply if something comes to mind

src/linker/Linker/Annotations.cs
```diff
@@ -620,6 +620,33 @@ internal bool DoesMemberRequireUnreferencedCode (IMemberDefinition member, [NotN
 			};
 		}
 
+		internal bool ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode (ICustomAttributeProvider? originMember)
+		{
+			// Check if the current scope method has RequiresUnreferencedCode on it
+			// since that attribute automatically suppresses all trim analysis warnings.
+			// Check both the immediate origin method as well as suppression context method
+			// since that will be different for compiler generated code.
+			if (originMember == null)
```
marek-safar, line 629: This check is redundant

src/linker/Linker/Annotations.cs
```diff
@@ -620,6 +620,33 @@ internal bool DoesMemberRequireUnreferencedCode (IMemberDefinition member, [NotN
...
+			if (originMember is MethodDefinition &&
+				IsInRequiresUnreferencedCodeScope ((MethodDefinition) originMember))
+				return true;
+
+			if (originMember is not IMemberDefinition member)
+				return false;
+
+			MethodDefinition? owningMethod;
+			while (context.CompilerGeneratedState.TryGetOwningMethodForCompilerGeneratedMember (member, out owningMethod)) {
+				Debug.Assert (owningMethod != member);
```
marek-safar, line 641: What is this assertion protecting against?

src/linker/Linker.Dataflow/FlowInsensitiveReflectionScanner.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Diagnostics.CodeAnalysis;
+using ILLink.Shared.DataFlow;
+using ILLink.Shared.TrimAnalysis;
+using Mono.Cecil;
+using Mono.Linker.Steps;
+using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.SingleValue>;
+
+namespace Mono.Linker.Dataflow
+{
+	public class FlowInsensitiveReflectionScanner
```
marek-safar, line 16: Very odd name. If the type does not need DataFlow why is it in the namespace?

marek-safar, line 16: Why is it not readonly struct?

src/linker/Linker.Dataflow/FlowInsensitiveReflectionScanner.cs
```diff
@@ -0,0 +1,90 @@
...
+				else
+					value = new SystemTypeValue (referencedType);
+			} else if (argument.Type.MetadataType == MetadataType.String) {
+				value = new KnownStringValue ((string) argument.Value);
+			} else {
+				// We shouldn't have gotten a non-null annotation for this from GetParameterAnnotation
+				throw new InvalidOperationException ();
+			}
+
+			Debug.Assert (value != null);
```
marek-safar, line 68: What is this assert for ?

src/linker/Linker.Dataflow/FlowInsensitiveReflectionScanner.cs
```diff
@@ -0,0 +1,90 @@
...
+			if (fieldValueCandidate is not ValueWithDynamicallyAccessedMembers fieldValue)
+				return;
+
+			var diagnosticContext = new DiagnosticContext (_origin, diagnosticsEnabled: true, _context);
+			RequireDynamicallyAccessedMembers (diagnosticContext, valueNode, fieldValue);
+		}
+
+		MultiValue GetValueNodeForCustomAttributeArgument (CustomAttributeArgument argument)
+		{
+			SingleValue value;
```
marek-safar, line 54: I'd simply return the instances and avoid extra local variable

src/linker/Linker.Dataflow/FlowInsensitiveReflectionScanner.cs
```diff
@@ -0,0 +1,90 @@
...
+			} else {
+				// We shouldn't have gotten a non-null annotation for this from GetParameterAnnotation
+				throw new InvalidOperationException ();
+			}
+
+			Debug.Assert (value != null);
+			return value;
+		}
+
+		public void ProcessGenericArgumentDataFlow (GenericParameter genericParameter, TypeReference genericArgument)
```
marek-safar, line 72: Isn't this odd name ...ArgumentDataFlow inside FlowInsensitive.. type ?

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -670,29 +672,76 @@ public FieldAnnotation (FieldDefinition field, DynamicallyAccessedMemberTypes an
...
+				}
+
+			default: {
+					DynamicallyAccessedMemberTypes memberTypes = GetFieldAnnotation (field);
+					return new FieldValue (field.FieldType.ResolveToTypeDefinition (_context), field, memberTypes);
+				}
+			}
+		}
+
+		internal SingleValue GetTypeValueNodeFromGenericArgument (TypeReference genericArgument)
```
vitek-karas, line 719: Nit: Can you please rename this - there are `ValueNode` things anymore - so probably just `GetTypeValueFromGenericArgument`.

src/linker/Linker.Dataflow/FlowInsensitiveReflectionScanner.cs
vitek-karas: I actually don't mind this being in the DataFlow - it is data flow, just not inside a method body.
Maybe we should split this further:
* `AttributeDataFlow` - handle all of the attribute stuff - we could even move some of the code from MarkStep to here possibly (NativeAOT has lot more of the code in these methods as well).
* `GenericArgumentDataFlow` - just the generic stuff - it's going to be small, but so what.
* And maybe we should then rename the `ReflectionMethodBodyScanner` to `MethodBodyDataFlow` - or something like that. But I'm not sure about this - both names have downsides as well.

vitek-karas (approved):

src/linker/Linker.Dataflow/AttributeDataFlow.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Mono.Linker.Dataflow
+{
+	public readonly struct AttributeDataFlow
+	{
+		readonly LinkContext _context;
+		readonly MarkStep _markStep;
+		readonly MessageOrigin _origin;
+
+		public AttributeDataFlow (LinkContext context, MarkStep markStep, in MessageOrigin origin)
```
jtschuster, line 20: We only ever use one instance method once before the struct goes out of scope. I know since it's a struct on the stack, it shouldn't allocate so it isn't less performant than static methods, but is there a reason we follow this pattern instead of making the methods static?

jtschuster (approved):

PR #2804, opened 2022-05-23
Better tests for attribute dataflow

vitek-karas:
Use RUC warnings to validate correct marking caused by attribute data flow.
The analyzer side of these tests doesn't work yet as they use string->type marking and the analyzer is missing a type name parser for now.


marek-safar (approved):

tlakollo (approved):
Aside from linting looks good

PR #2807, opened 2022-05-26
Warn on unhandled store in reference case instead of throwing

jtschuster:
This change should address the issues in https://github.com/dotnet/runtime/pull/69710.

Previously, if we found a value that we didn't know how to handle when storing a value to a reference, we would throw and fail. This change is to warn in that situation instead.

### DIFF
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using ILLink.Shared;
 using ILLink.Shared.DataFlow;
@@ -845,7 +844,7 @@ namespace Mono.Linker.Dataflow
 					HandleStoreMethodReturnValue (method, methodReturnValue, operation, source);
 					break;
 				case UnknownValue:
-					// These cases should only be refs to array elements.
+				// These cases should only be refs to array elements.
 				default:
 					// Should Add a new "Reference cannot be statically determined" warning code
 					_context.LogWarning (new MessageOrigin (method, operation.Offset), DiagnosticId.ReferenceStoredToCannotBeStaticallyDetermined, method.GetDisplayName ());
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -848,7 +848,7 @@ namespace Mono.Linker.Dataflow
 					// These cases should only be refs to array elements.
 				default:
 					// Should Add a new "Reference cannot be statically determined" warning code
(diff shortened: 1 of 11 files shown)


vitek-karas:
Alternatively we could only warn if the write is to a `System.Type` or `System.String` based type/pointer. After all that's all we care about - there's a method `FlowAnnotations.IsTypeInterestingForDataflow` which already does basically this - we would just need to resolve the pointer type to its base type and ask that method.

In general I agree that we don't care about unsafe code too much, but this doesn't seem to hurt too much and it adds a layer of protection (I'm specifically thinking about runtime/aspnet or other highly optimized libraries which are more likely to use unsafe code). It would definitely not cover everything (you could get around this by casting pointers), but better than nothing.

I do agree that we should not warn on the case in the runtime.

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -843,8 +843,17 @@ protected void StoreInReference (MultiValue target, MultiValue source, MethodDef
 					// Ref returns don't have special ReferenceValue values, so assume if the target here is a MethodReturnValue then it must be a ref return value
 					HandleStoreMethodReturnValue (method, methodReturnValue, operation, source);
 					break;
+				case IValueWithStaticType typedValue:
+					if (typedValue.StaticType is not TypeReference typeRef
+						|| ResolveToTypeDefinition (typeRef) is not TypeDefinition typeOfTarget)
+						break;
+
+					if (typeOfTarget.IsTypeOf(WellKnownType.System_String) || typeOfTarget.IsTypeOf (WellKnownType.System_Type))
```
vitek-karas, line 851: Please make the method on FlowAnnotations internal and call it from here - we don't want to duplicate the same logic.

vitek-karas:
Change the checks to only throw if we see a truly unexpected combination

I scoped the exception down to the case of a typed value which is of interesting type. Currently we don't think this can every happen and it's basically the only case where there's a potential for analysis hole.

I removed the new warning as there's nothing to warn about really. Originally we had the default case warn, but I don't think that makes sense currently.

I also added a test case which actually reproes the failure from the runtime - not exactly (it seems the compiler in runtime is slightly different or something else is going on and the exact same code from runtime produces different IL in the linker tests), but modified so that it does hit the exception in unchanged code.

I also validated that with this change the linker doesn't fail on runtime anymore.

vitek-karas:
@marek-safar We discussed this fix and we think we know what we want - I made the necessary changes. I'll wait on @jtschuster and @sbomer to take another look, but I think we should be able to merge this today.

src/linker/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
+  "profiles": {
+    "Mono.Linker": {
+      "commandName": "Project",
+      "commandLineArgs": "@linker.rsp",
+      "workingDirectory": "F:\\ILLinker\\fx\\c1\\repro"
```
jtschuster, line 6: Nit: I don't think we want this file.

jtschuster (commented):
GitHub won't let me approve my own PR, but it looks great, thanks for finishing this!

PR #2825, opened 2022-06-07
Don't catch all exceptions; propagate unexpected errors to trigger Watson dump

jtschuster:
This allows us to collect a dump when we encounter a failure that we don't expect and handle with our own exception.


src/linker/Linker/Driver.cs
```diff
@@ -789,14 +790,18 @@ public int Run (ILogger? customLogger = null, bool throwOnFatalLinkerException =
 				if (throwOnFatalLinkerException)
 					throw;
 				return lex.MessageContainer.Code ?? 1;
+			}
+			catch (ArgumentException) {
```
marek-safar, line 794: What is special on ArgumentException ?

src/linker/Linker/Driver.cs
```diff
@@ -789,14 +790,18 @@ public int Run (ILogger? customLogger = null, bool throwOnFatalLinkerException =
...
+				Context.LogError (null, DiagnosticId.LinkerUnexpectedError);
+				throw;
 			} catch (ResolutionException e) {
 				Context.LogError (null, DiagnosticId.FailedToResolveMetadataElement, e.Message);
-			} catch (Exception) {
+			} catch (Exception e) {
 				// Unhandled exceptions are usually linker bugs. Ask the user to report it.
 				Context.LogError (null, DiagnosticId.LinkerUnexpectedError);
+				Environment.FailFast ("Unexpected Error", e);
```
marek-safar, line 802: Are we sure we want to take down the whole process?

vitek-karas: The goal of this change is to make it possible to diagnose hard to repro failures (like the ones we see in runtime CI currently). What this is trying to do is to crash the process in such a way that dump will be captured... and we can then investigate the dump.

We shortly discussed the possibility to use non fatal Watson dump on Windows, but for now decided that the hard crash would be better (it works everywhere, not just Windows).

marek-safar: Right just ensure that all the products that integrate linker (VS, VS4M, VSCode, etc.) via msbuild are prepared for that.

vitek-karas: All of those should do it through msbuild - and in that case linker should run as an external process. If it crashes msbuild will produce an error - it doesn't look exactly pretty, but that's basically the tradeoff here - between pretty failures and diagnosable ones.

MichalStrehovsky: Why do we need to catch these exceptions at all?

If we don't have a catch anywhere, it will just go into the runtime-default unhandled exception paths that also trigger Watson.

For example here we could do the equivalent of:

```csharp
try
{
    ...
}
catch when (PrintMessageToContactLinkerDevsAndReturnFalse()) { /* unreachable */ }
```

The difference is basically that if you have FailFast, the stack trace that the runtime prints to stderr is the one of the FailFast. If you don't catch the exception, the stack trace printed to stderr is the one for when the exception was thrown.

test/Mono.Linker.Tests/TestCasesRunner/TestRunner.cs
```diff
@@ -109,7 +109,7 @@ private LinkedTestCaseResult Link (TestCase testCase, TestCaseSandbox sandbox, M
 			AddLinkOptions (sandbox, compilationResult, builder, metadataProvider);
 
 			LinkerTestLogger logger = new LinkerTestLogger ();
-			linker.Link (builder.ToArgs (), linkerCustomizations, logger);
+			linker.Link (builder.ToArgs (), linkerCustomizations, logger, testCase.Name == "CustomStepWithWarnings");
```
jtschuster, line 112: The test "InvalidWarningCodeThrows" expects an ArgumentException to be thrown and fails if the process crashes. This flag makes sure we don't fail that test.

src/linker/Linker/Driver.cs
jtschuster: Sorry for not marking this as still in progress, after discussing with Sven and Andy I realized we want to fail in the exception filter like Michal suggested to get the Watson dumps. I've updated the code to do that.

src/linker/Linker/Driver.cs
```diff
@@ -780,29 +781,41 @@ public int Run (ILogger? customLogger = null, bool throwOnFatalLinkerException =
...
+				Context.Tracer.Finish ();
+			}
+
+			return Context.ErrorsCount > 0 ? 1 : 0;
+		}
+
+		/// <summary>
+		/// Handles exceptions in the linker. Prints error messages, then calls Environment.FailFast to create a dump with the full call stack.
+		/// </summary>
+		bool FailCatch (Exception e)
```
vitek-karas, line 798: Nit: I would call this something like `LogFatalFailure` or something like that. To note that it's solely about logging more info.

src/linker/Linker/Driver.cs
```diff
@@ -780,29 +781,41 @@ public int Run (ILogger? customLogger = null, bool throwOnFatalLinkerException =
...
-				// The stack trace will go to stderr, and the MSBuild task will surface it with High importance.
-				throw;
-			} finally {
-				Context.FlushCachedWarnings ();
-				Context.Tracer.Finish ();
+				break;
 			}
-
-			return Context.ErrorsCount > 0 ? 1 : 0;
+			Context.FlushCachedWarnings ();
```
vitek-karas, line 816: I'm a bit nervous about this. Ideally we would do very little in this method since we can't really trust data structures anymore. I don't see why we would need to print out all the warnings we got so far on a fatal crash. I know that there's some chance that it might help, but in the worst case if we get a dump the warnings will be in the dump.

Same goes about the tracer - the tracer produces the large XML files which are rarely used and only to figure out "why was this kept". Not having that info complete when the linker hard crashes is absolutely fine.

Definitely worth a comment that we intentionally don't do it.

src/linker/Linker/Driver.cs
MichalStrehovsky: My question was more about "why do we catch the exception in the first place when we don't handle it?". If we catch unexpected things, we need to add extra code to do the FailFast - but we could just be more mindful of what we catch instead.

Like in this specific spot I'm commenting on: we catch only to print something into the logger (and then we retrhow previously, or FailFast with this change). But if we instead print into the logger from a filter clause (a filter that returns `false` to not actually trigger a catch) we don't need to worry about rethrow or FailFast - the exceptions will fly past this spot.

src/linker/Linker/Driver.cs
```diff
@@ -780,29 +781,41 @@ public int Run (ILogger? customLogger = null, bool throwOnFatalLinkerException =
 
 			try {
 				p.Process (Context);
-			} catch (LinkerFatalErrorException lex) {
+			} catch (Exception e) when (FailCatch (e)) {
+				// Proces will be terminated in exception filter.
+				throw e;
```
agocke, line 786:
```suggestion
				throw;
```

src/linker/Linker/Driver.cs
jtschuster: We'll probably want to move them out of the `finally` block too then, right?

jtschuster: Looking closer I realize messages and errors are not cached, so we would be able to log some extra information without flushing.

src/linker/Linker/Driver.cs
jtschuster: This was thrown when a custom step tries to use an invalid error code. It's no longer relevant to the PR though.

src/linker/Linker/Driver.cs
```diff
@@ -780,29 +780,37 @@ public int Run (ILogger? customLogger = null, bool throwOnFatalLinkerException =
...
+		/// <summary>
+		/// This method is called in the exception filter for unexpected exceptions.
+		/// Prints error messages and returns false to avoid catching in the exception filter.
+		/// </summary>
+		bool LogFatalError (Exception e)
+		{
+			switch (e) {
+			case LinkerFatalErrorException lex:
 				Context.LogMessage (lex.MessageContainer);
-				Console.Error.WriteLine (lex.ToString ());
```
agocke, line 785: Why delete this line? Is it just a duplicate for some reason?

agocke (approved):
LGTM

src/linker/Linker/Driver.cs
jtschuster: My understanding is that when the exception causes the program to crash, it will print out all the same information as lex.ToString(), so it isn't necessary to do it in the exception filter too.

PR #2833, opened 2022-06-10
Fix analyzer issues related to visiting assignment targets

sbomer:
Fixes more issues discovered in https://github.com/dotnet/linker/issues/2828. Best reviewed commit-by-commit.

This contains fixes for three issues:
- Assignment to events (see the first commit).
- Nullref in assignment to captured property reference, when the property doesn't have a getter.
- [This](https://github.com/dotnet/runtime/blob/fae7ee8e7e3aa7f86836318a10ed676641e813ad/src/libraries/System.Private.CoreLib/src/System/Type.cs#L621) assignment to `Type.EmptyTypes`. It was hitting this case in `VisitFieldReference`:
  https://github.com/dotnet/linker/blob/618eb724e7320fd2e4a2b6e8f7e0a22195a9e633/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisVisitor.cs#L123-L124
  and returning an ArrayValue for the LHS of the assignment, causing a crash later:
  https://github.com/dotnet/linker/blob/618eb724e7320fd2e4a2b6e8f7e0a22195a9e633/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisAssignmentPattern.cs#L45-L46

It also contains a testcase for https://github.com/dotnet/linker/issues/2832 which I noticed while investigating this.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
@@ -217,7 +217,10 @@ namespace ILLink.RoslynAnalyzer.DataFlow
 		public override TValue VisitFlowCaptureReference (IFlowCaptureReferenceOperation operation, LocalDataFlowState<TValue, TValueLattice> state)
 		{
 			if (!operation.GetValueUsageInfo (Context.OwningSymbol).HasFlag (ValueUsageInfo.Read)) {
-				// Debug.Assert (IsLValueFlowCapture (operation.Id)); // https://github.com/dotnet/roslyn/issues/60757
+				// There are known cases where this assert doesn't hold, because LValueFlowCaptureProvider
+				// produces the wrong result in some cases for flow captures with IsInitialization = true.
+				// https://github.com/dotnet/linker/issues/2749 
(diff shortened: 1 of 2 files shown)


src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -227,7 +230,10 @@ public override TValue VisitFlowCaptureReference (IFlowCaptureReferenceOperation
 		// is like a local reference.
 		public override TValue VisitFlowCapture (IFlowCaptureOperation operation, LocalDataFlowState<TValue, TValueLattice> state)
 		{
-			TValue value = Visit (operation.Value, state);
+			// If the captured value is a property reference, we can't easily tell inside of
+			// VisitPropertyReference whether it is accessed for reads or writes.
+			// https://github.com/dotnet/roslyn/issues/25057
+			// Avoid visiting the captured value unless it is an RValue.
```
vitek-karas, line 236: Just curious - won't this cause similar problems for properties like we have with not seeing the "instance" operation for field access? Basically a property variant of https://github.com/dotnet/linker/issues/2832

PR #2836, opened 2022-06-13
Mark all type forwarders used during string->Type resolution

vitek-karas:
Previously we only marked the type forwarder for the outer-most type (if any), but the same problem can happen with type names in generic arguments. If they are forwarder we need to mark those type forwarders as well.

This change records all type resolutions during the string->Type resolution and then makes sure that all type forwarders used get marked.

Renamed a method on HandleCallAction to make it more descriptive of what it actually does. Also unifies all string->Type resolution into the ReflectionMarker.

Enabled test which has been already added for the bug and added another variation (CreateInstance).

Improved ResultsChecker to include assembly names in validation messages.

Fixes https://github.com/dotnet/linker/issues/1536


jtschuster (approved):
LGTM, thank you!

PR #2841, opened 2022-06-15
Revert "Re-add static interface trimming with more testing (#2791)"

jtschuster:
This reverts commit ff646e0ba4f5b919af854ec73dcee6c4903c3def.

The latest linker update in https://github.com/dotnet/runtime/pull/70614 with the static interface trimming has more failures than the previous and they look similar to the previous failures static interface trimming has caused. I think we should revert static interface trimming in the meantime.


jtschuster:
> Are you able to repro this locally? We should make sure we can run apicompat the same way as it's run in ci so that we can validate the fix in advance next time.

I haven't been able to repro yet. Should we consider making a pipeline for apicompat in this repo to test riskier linker changes?

PR #2842, opened 2022-06-16
Add compiler-generated code dataflow analysis

sbomer:
~Still a work in progress, but feel free to take an early look!~

edit: After some discussion I changed the approach. See old notes below for the earlier approach.

This treats fields of display classes and state machines as hoisted locals. We track all assignments to hoisted locals within a method group (the set of compiler-generated code reachable from a given user method). The analysis is technically flow-insensitive, because it assumes that any assignment to a hoisted local can reach any read of the same local. This will produce extra warnings in some cases, but it closes the holes in the original approach:
- State will "flow" out of nested functions. That is, writes (to hoisted locals) within nested functions will reach reads in the enclosing user method
- Lambdas are analyzed at the point of delegate conversion, but with all possible states of captured variables. So effectively, writes after the lambda declaration will reach reads within the lambda.

Previously, hoisted locals were treated as unannotated, so they would produce dataflow warnings if they reached an annotated location. Now that we analyze hoisted locals, cases where the value satisfies requirements at the point of consumption won't warn. This means that accessing these fields (representing hoisted locals) via reflection is problematic, since it could mutate the values of these fields and invalidate the correctness analysis. For this reason we now warn on reflection access to compiler-generated fields.

To prevent noise, we only warn for reflection access to compiler-generated fields that represent types which may be annotated - so `Type`, `string`, etc. - but not `int`. This is technically a hole because ints also participate in dataflow analysis but per discussion with @vitek-karas I think we agree that this is an ok tradeoff.

<hr />
Old notes:

This flows state through hoisted locals in state machine methods, and into nested functions (lambdas/local functions). Some caveats to be aware of:
- Lambdas are analyzed with the state of captured variables at the point of delegate conversion, not at the point of invocation (which would require tracking delegates as dataflow values). This is an analysis hole in some cases. It follows the model for nullable analysis, which has an analysis hole in this case for lambdas but not for local functions: https://sharplab.io/#v2:EYLgtghglgdgNAFxFANnAJiA1AHwAIBMAjALABQAxDAK4ooTAoCmABEzA8+XgMwuEsAwiwDe5FhP588AFhYBZABQBKUeMkbBAfhYBjFgF4WMJgHchKgNzqNEvEQIsUhlisMA+NWVs/+RAJyKugB0ACoA9gDKCABOsADmKsrW3r4Avim+LDa2siwAMuG6ECh4AKwIUOEwbgae9oEhEdFxMInKyTkaXZL6RjR0mb4oVj0ShcWlFVU1nakSaeSLZEA= 
- This doesn't track state flowing _out_ of nested functions. Again, similar to nullable: https://sharplab.io/#v2:EYLgtghglgdgNAFxFANnAJiA1AHwAIBMAjALABQAxDAK4ooTAoCmABEzA8+XgMwuEsAwiwDe5FhP588AFhYBZABQBKUeMkbBAfhYBjFgF4WMJgHchKgNzqNEvEQIsUhlisMA+NWVs+9LmnTW3r4AvkG+LDa2siwAMgD2uhAoeACsCFDxMG4GnvpGAShBURooViWSCUkp6ZnZyuG+9gCciroAdAAq8QDKCABOsADmKg1RIeQTZEA=.

@vitek-karas @agocke @MichalStrehovsky I'm curious to hear your opinion on whether we should try to plug those holes, or if this approach is good enough.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -1634,8 +1634,8 @@ namespace Mono.Linker.Steps
 			// When marking override methods via DynamicallyAccessedMembers, we should only issue a warning for the base method.
 			bool skipWarningsForOverride = member is MethodDefinition m && m.IsVirtual && Annotations.GetBaseMethods (m) != null;
 
-			bool warnsOnRUC = Annotations.DoesMemberRequireUnreferencedCode (member, out RequiresUnreferencedCodeAttribute? requiresUnreferencedCodeAttribute);
-			if (warnsOnRUC && !skipWarningsForOverride) {
+			bool isReflectionAccessCoveredByRUC = Annotations.DoesMemberRequireUnreferencedCode (member, out RequiresUnreferencedCodeAttribute? requiresUnreferencedCodeAttribute);
+			if (isReflectionAccessCoveredByRUC && !skipWarningsForOverride) {
 				var id = reportOnMember ? DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesMemberWithRequiresUnreferencedCode : DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesMemberOnBaseWithRequiresUnreferencedCode;
 				Context.LogWarning (origin, id, type.GetDisplayName (),
 					((MemberReference) member).GetDisplayName (), // The cast is valid since it has to be a method or field
@@ -1643,21 +1643,25 @@ namespace Mono.Linker.Steps
 					MessageFormat.FormatRequiresAttributeMessageArg (requiresUnreferencedCodeAttribute!.Url));
 			}
 
-			bool warnsOnReflection = Annotations.FlowAnnotations.ShouldWarnWhenAccessedForReflection (member);
-			if (warnsOnReflection && !skipWarningsForOverride) {
+			bool isReflectionAccessCoveredByDAM = Annotations.FlowAnnotations.ShouldWarnWhenAccessedForReflection (member);
+			if (isReflectionAccessCoveredByDAM && !skipWarningsForOverride) {
 				var id = reportOnMember ? DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesMemberWithDynamicallyAccessedMembers : DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesMemberOnBaseWithDynamicallyAccessedMembers;
 				Context.LogWarning (origin, id, type.GetDisplayName (), ((MemberReference) member).GetDisplayName ());
 			}
 
-			// Warn on reflection access to compiler-generated methods. This must happen even for virtual overrides,
-			// because it should include the virtual MoveNext method of state machines.
-			if (member is MethodDefinition method && ShouldWarnForReflectionAccessToCompilerGeneratedCode (method, warnsOnRUC, warnsOnReflection)) {
+			// Warn on reflection access to compiler-generated methods, if the method isn't already unsafe to access via reflection
+			// due to annotations. For the annotation-based warnings, we skip virtual overrides since those will produce warnings on
+			// the base, but for unannotated compiler-generated methods this is not the case, so we must produce these warnings even
+			// for virtual overrides. This ensures that we include the unannotated MoveNext state machine method. Lambdas and local
+			// functions should never be virtual overrides in the first place.
+			bool isCoveredByAnnotations = isReflectionAccessCoveredByRUC || isReflectionAccessCoveredByDAM;
+			if (member is MethodDefinition method && ShouldWarnForReflectionAccessToCompilerGeneratedCode (method, isCoveredByAnnotations)) {
 				var id = reportOnMember ? DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesCompilerGeneratedMember : DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesCompilerGeneratedMemberOnBase;
 				Context.LogWarning (origin, id, type.GetDisplayName (), method.GetDisplayName ());
 			}
 
 			// Warn on reflection access to compiler-generated fields.
-			if (member is FieldDefinition field && ShouldWarnForReflectionAccessToCompilerGeneratedCode (field, warnsOnRUC, warnsOnReflection)) {
+			if (member is FieldDefinition field && ShouldWarnForReflectionAccessToCompilerGeneratedCode (field, isCoveredByAnnotations)) {
 				var id = reportOnMember ? DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesCompilerGeneratedMember : DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesCompilerGeneratedMemberOnBase;
 				Context.LogWarning (origin, id, type.GetDisplayName (), field.GetDisplayName ());
 			}
@@ -1724,10 +1728,10 @@ namespace Mono.Linker.Steps
 			}
 		}
 
-		bool ShouldWarnForReflectionAccessToCompilerGeneratedCode (FieldDefinition field, bool warnsOnRUC, bool warnsOnReflection)
+		bool ShouldWarnForReflectionAccessToCompilerGeneratedCode (FieldDefinition field, bool isCoveredByAnnotations)
 		{
 			// No need to warn if it's already covered by the Requires attribute or explicit annotations on the field.
-			if (warnsOnRUC || warnsOnReflection)
+			if (isCoveredByAnnotations)
 				return false;
 
 			if (!CompilerGeneratedState.IsNestedFunctionOrStateMachineMember (field))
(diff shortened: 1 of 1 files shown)


src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -258,47 +293,53 @@ public virtual void InterproceduralScan (MethodBody methodBody)
...
 
-		protected virtual void Scan (MethodBody methodBody, ref ValueSet<MethodProxy> methodsInGroup)
+		protected virtual void Scan (MethodBody methodBody, ref InterproceduralState interproceduralState)
 		{
 			MethodDefinition thisMethod = methodBody.Method;
 
 			LocalVariableStore locals = new (methodBody.Variables.Count);
+			// Initialize the hoisted locals state on entry to this method, from the interprocedural state.
+			// Use -1 as the basic block index for any values we track on entry to the method.
+			var defaultValue = new ValueBasicBlockPair (UnknownValue.Instance, -1);
```
vitek-karas, line 337: This is only used right below - can you please inline it? (It's confusing - "default value" can have many meanings)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1735,6 +1717,62 @@ void MarkField (FieldDefinition field, in DependencyInfo reason, in MessageOrigi
 			}
 		}
 
+		bool ShouldWarnForReflectionAccessToCompilerGeneratedCode (FieldDefinition field)
+		{
+			if (!CompilerGeneratedState.IsNestedFunctionOrStateMachineMember (field))
+				return false;
+
+			// No need to warn if it's already covered by the Requires attribute or explicit annotations on the field.
+			if (Annotations.DoesFieldRequireUnreferencedCode (field, out _) || Annotations.FlowAnnotations.ShouldWarnWhenAccessedForReflection (field))
```
jtschuster, line 1726: Should the second case be `!Annotations.FlowAnnotations.ShouldWarnWhenAccessedForReflection`? Or do we skip because it will already warn elsewhere?

test/Mono.Linker.Tests.Cases/DataFlow/CompilerGeneratedCodeDataflow.cs
```diff
@@ -0,0 +1,540 @@
...
+				LocalFunction ();
+
+				[ExpectedWarning ("IL2072", nameof (GetWithPublicMethods), nameof (DataFlowTypeExtensions.RequiresAll),
+					ProducedBy = ProducedBy.Trimmer)]
+				[ExpectedWarning ("IL2072", nameof (GetWithPublicFields), nameof (DataFlowTypeExtensions.RequiresAll),
+					ProducedBy = ProducedBy.Trimmer)]
+				void LocalFunction () => t.RequiresAll ();
+			}
+
+			static void ReadCapturedVariableInMultipleBranches (bool b = false)
```
jtschuster, line 219: Could you add a test with something like

```C#
static void ReadCapturedVariableInMultipleBranches (bool b = false)
{
    Type t;
    if (b) {
        t = GetWithPublicMethods ();
        TRequiresMethods ();
    } else {
        t = GetWithPublicFields ();
        TRequiresFields ();
    }

    void TRequiresFields () => t.RequiresPublicFields ();
    void TRequiresMethods () => t.RequiresPublicMethods ();
}
```

My guess is that this shouldn't have to warn.

jtschuster (approved):
LGTM, Thank you!

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -226,29 +241,40 @@ static void ValidateNoReferenceToReference (LocalVariableStore locals, MethodDef
...
-			HashSet<MethodDefinition> scannedMethods = new HashSet<MethodDefinition> ();
+			var methodGroupLattice = new ValueSetLattice<MethodProxy> ();
+			// Note that the default value of a hoisted local will be MultiValueLattice.Top, not UnknownValue.Instance.
+			// This ensures that there are no warnings for the "unassigned state" of a parameter.
+			// Definite assignment should ensure that there is no way for this to be an analysis hole.
+			// When getting a hoisted local value, we also assert that the value is not Top.
+			var hoistedLocalsLattice = new DictionaryLattice<HoistedLocalKey, MultiValue, ValueSetLattice<SingleValue>> (
+				MultiValueLattice
+			);
+			var interproceduralStateLattice = new InterproceduralStateLattice (methodGroupLattice, hoistedLocalsLattice);
```
jtschuster, line 252: It looks like `methodGroupLattice` and `hoistedLocalsLattice` are only used in the `InterproceduralStateLattice` ctor, maybe we should inline them to make it clear they're not used anywhere else.

vitek-karas:
Regarding the potential holes with reflection access to closure fields:
* I'm mostly worried about noise - we've seen cases where `All` is applied to a type, if this type uses state machines/closures anywhere all of the fields on those subtypes will be marked with "Reflection access" and can generate warnings. This is often counter intuitive and hard to diagnose where the warning came from (and why), and also is tricky to fix - suppressions are problematic (especially since they would typically apply to entire types) and real fix is basically impossible.
* With that said - reducing noise from this should be a relatively high priority. Personally I would not mind if we leave some holes behind in favor of reducing noise - ideally with a plan how to "plug" them.
* So the `int` hole - I'm perfectly OK with that.
* We need to test this, we might need to ignore `string` as well (even though it can carry annotations) - again to reduce noise.
* The reason I'm OKish leaving holes in this - It is VERY unlikely the reflection will be actually exercised at runtime and even less so to modify the field's value. And honestly, if somebody is doing private reflection on compiler generated code with modifications... I'm fine if that's the one case where we break them.
* Going forward if this becomes a sore point we can do a lot better by:
  * Add the ability to track access to values (fields, parameters) during data flow analysis - basically being able to determine which values where used to make marking/warning decisions. This is relatively easy, just needs a good C# model to make it easy to write and reliable.
  * Once we have that we would use that to determine which of the hoisted fields are actually interesting for data flow - not just a heuristic based on their type. And then we would warn on reflection access only to those fields. Such warnings would greatly reduce the noise ratio (but still relatively high - as noted above I doubt we will see code which actually wants to reflect over the compiler generated stuff).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1735,6 +1717,62 @@ void MarkField (FieldDefinition field, in DependencyInfo reason, in MessageOrigi
...
+
+		void ProcessAnalysisAnnotationsForField (FieldDefinition field, DependencyKind dependencyKind, in MessageOrigin origin)
+		{
+			if (origin.Provider != ScopeStack.CurrentScope.Origin.Provider) {
+				Debug.Assert (dependencyKind == DependencyKind.DynamicallyAccessedMemberOnType ||
+					(origin.Provider is MethodDefinition originMethod && CompilerGeneratedState.IsNestedFunctionOrStateMachineMember (originMethod)));
+			}
+
+			if (dependencyKind == DependencyKind.DynamicallyAccessedMemberOnType) {
+				ReportWarningsForTypeHierarchyReflectionAccess (field, origin);
```
vitek-karas, line 1746: Does this mean that we will get more noise when using type annotations?
This code path doesn't use the heuristic to determine which fields to warn about - and it can still effectively apply `All` to a type (for example).

vitek-karas: Perfectly OK to try to create an issue/postpone - since I think we have the problem already.

vitek-karas: Well - please add a test for this at least.

src/linker/Linker.Steps/MarkStep.cs
vitek-karas, line 1726: Maybe we should restructure the code below - basically do the "Reflection access on compiler generated code" last (currently we do it "first"). And then remember if we already produced a warning due to RUC or annotation for the field and then simply avoid the compiler generated check completely.

The current implementation requires the two methods to be in sync, which is not robust. And it also performs the same check multiple times.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1628,32 +1628,39 @@ static bool IsDeclaredWithinType (IMemberDefinition member, TypeDefinition type)
...
 
-			if (Annotations.FlowAnnotations.ShouldWarnWhenAccessedForReflection (member)) {
+			bool warnsOnReflection = Annotations.FlowAnnotations.ShouldWarnWhenAccessedForReflection (member);
+			if (warnsOnReflection && !skipWarningsForOverride) {
 				var id = reportOnMember ? DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesMemberWithDynamicallyAccessedMembers : DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesMemberOnBaseWithDynamicallyAccessedMembers;
 				Context.LogWarning (origin, id, type.GetDisplayName (), ((MemberReference) member).GetDisplayName ());
 			}
+
+			// Warn on reflection access to compiler-generated methods. This must happen even for virtual overrides,
+			// because it should include the virtual MoveNext method of state machines.
```
vitek-karas, line 1653: This is worth a larger comment I think - basically explain that MoveNext is virtual, but in this case we're not generating warning because of any annotation (RUC or DAM), instead we generate a warning because the analysis can't handle direct access to the method. And so in such case the normal virtual hierarchy rules don't aply.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1628,32 +1628,39 @@ static bool IsDeclaredWithinType (IMemberDefinition member, TypeDefinition type)
...
+			if (member is MethodDefinition method && ShouldWarnForReflectionAccessToCompilerGeneratedCode (method, warnsOnRUC, warnsOnReflection)) {
+				var id = reportOnMember ? DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesCompilerGeneratedMember : DiagnosticId.DynamicallyAccessedMembersOnTypeReferencesCompilerGeneratedMemberOnBase;
+				Context.LogWarning (origin, id, type.GetDisplayName (), method.GetDisplayName ());
+			}
```
vitek-karas, line 1657: This would warn even if we already warned due to annotations above, right?
Didn't we say we don't want that (technically it's going to look almost like a duplicate warning, right?)
This can happen for example for local functions - which can have RUC/DAM on them, but they're still considered compiler generated code.

vitek-karas, line 1657: This needs another comment which basically says that the `skipWarningsForOverride=true` can only happen on MoveNext like methods in compiler generated code, that it will never happen for local functions and alike. Meaning that it's never going to happen that we decided to not warn above and would also skip warning here.
Reason: Technically the logic here is wrong, the `warnsOnReflection` is `true` regardless if we produced warning above or not (due to override auto-suppression) and that will suppress producing the compiler-generated warning here. But it will never actually matter. So I'm fine keeping the code as-is, but I think this needs good explanation.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2960,41 +2969,37 @@ void ProcessAnalysisAnnotationsForMethod (MethodDefinition method, DependencyKin
...
 
-				Context.LogWarning (origin, DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection, method.GetDisplayName ());
+			// Warn about reflection access to compiler-generated code. This must happen even for virtual overrides,
+			// because it should include the virtual MoveNext method of state machines.
+			switch (dependencyKind) {
+			case DependencyKind.AccessedViaReflection:
+			case DependencyKind.DynamicallyAccessedMember:
+				if (ShouldWarnForReflectionAccessToCompilerGeneratedCode (method, warnsOnRUC, warnsOnReflection))
+					Context.LogWarning (origin, DiagnosticId.CompilerGeneratedMemberAccessedViaReflection, method.GetDisplayName ());
+				break;
```
vitek-karas, line 3002: Same comment as above about the tricky interactions between override auto-suppression and the compiler-generated code. Probably worth the same comment (or reference one from the other).

vitek-karas (approved):
Can't wait for all of this light up :-)

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Looks good - thanks!

PR #2848, opened 2022-06-17
Fix package props for older SDKs

sbomer:
This should keep the existing package override logic working, since
the current SDK will only import the in-box ILLink.Tasks.props if
this is not already set.

We could probably remove this once runtime is updated to use an SDK
with https://github.com/dotnet/sdk/pull/25993. Or, we might want to
keep it as long as any SDKs using the old override logic are still in support.
This would let them reference newer ILLink.Tasks packages as long as
the task surface doesn't contain breaking changes (like new required parameters).


agocke (approved):
LGTM, nice catch

PR #2855, opened 2022-06-20
Display init properties as ".init" in GetDisplayName

jtschuster:
Shows `init` properties as `Property.init` instead of `Property.set` in GetDisplayName and adds an `IsInit` extension to MethodDefinition.

Fixes https://github.com/dotnet/linker/issues/2816, but may want to wait until NativeAOT has the same changes to merge.


marek-safar:
Closing, please port to runtime if you think it's still worth it

PR #2856, opened 2022-06-22
Add more trimming options to TrimMode and change defaults

agocke:
The new default for trimming is equivalent to TrimmerDefaultAction=link.
Since we're trying to stay away from the term link and TrimmerDefaultAction
isn't a very clear name, this change repurposes TrimMode.

The two new TrimModes are 'full' and 'partial'. The new default is equivalent
to 'TrimMode=full'. These new modes are also recognized for earlier TFMs, but
the defaults are only changed for net7.0+

### DIFF
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -192,6 +192,13 @@ namespace ILLink.Tasks
 		bool? _removeSymbols;
 
 		/// <summary>
+		/// The version of the link task. Used to determine the meaning of certain properties.
+		/// Roughly maps to the TargetFrameworkVersion.
+		/// </summary>
+		[Required]
+		public int LinkVersion { get; set; }
+
+		/// <summary>
 		///   Sets the default action for trimmable assemblies.
 		///   Maps to '--trim-mode'
 		/// </summary>
@@ -323,10 +330,10 @@ namespace ILLink.Tasks
 
 				args.Append ("-reference ").AppendLine (Quote (assemblyPath));
 
-				string trimMode = assembly.GetMetadata ("TrimMode");
-				if (!String.IsNullOrEmpty (trimMode)) {
+				string assemblyTrimMode = assembly.GetMetadata ("TrimMode");
+				if (!String.IsNullOrEmpty (assemblyTrimMode)) {
 					args.Append ("--action ");
-					args.Append (trimMode);
+					args.Append (assemblyTrimMode);
 					args.Append (' ').AppendLine (Quote (assemblyName));
 				}
 
@@ -441,8 +448,44 @@ namespace ILLink.Tasks
 			if (_removeSymbols == false)
 				args.AppendLine ("-b");
 
+			string trimMode;
 			if (TrimMode != null)
-				args.Append ("--trim-mode ").AppendLine (TrimMode);
+			{
+				trimMode = TrimMode switch
+				{
+					"full" => "link",
+					"partial" => "link",
+					var x => x
+				};
+			}
+			else
+			{
+				trimMode = LinkVersion switch
+				{
+					< 6 => "copyused",
+					_ => "link"
+				};
+			}
+			args.Append ("--trim-mode ").AppendLine (trimMode);
+
+			string defaultAction;
+			if (DefaultAction != null && LinkVersion < 7)
+			{
+				defaultAction = DefaultAction;
+			}
+			else
+			{
+				defaultAction = (LinkVersion, TrimMode) switch
+				{
+					(_, "full") => "link",
+					(_, "partial") => "copy",
+					(< 6, _) => trimMode, // Use resolved trim mode, not original
+					(6, _) => "copy",
+					_ => "link",
+				};
+			}
+			args.Append("--action ").AppendLine (defaultAction);
+
 
 			if (DefaultAction != null)
 				args.Append ("--action ").AppendLine (DefaultAction);
--- a/src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
+++ b/src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
@@ -51,8 +51,8 @@ Copyright (c) .NET Foundation. All rights reserved.
   <PropertyGroup Condition="'$(SuppressTrimAnalysisWarnings)' == '' And '$(PublishTrimmed)' == 'true' And '$(EnableTrimAnalyzer)' != 'true'">
     <!-- Trim analysis warnings are suppressed for .NET < 6. -->
     <SuppressTrimAnalysisWarnings Condition="$([MSBuild]::VersionLessThan('$(TargetFrameworkVersion)', '6.0'))">true</SuppressTrimAnalysisWarnings>
-    <!-- Suppress for WPF/WinForms (unless linking everything) -->
-    <SuppressTrimAnalysisWarnings Condition="'$(TrimmerDefaultAction)' != 'link' And ('$(UseWpf)' == 'true' Or '$(UseWindowsForms)' == 'true')">true</SuppressTrimAnalysisWarnings>
+    <!-- Suppress for WPF/WinForms -->
+    <SuppressTrimAnalysisWarnings Condition="'$(UseWpf)' == 'true' Or '$(UseWindowsForms)' == 'true'">true</SuppressTrimAnalysisWarnings>
     <!-- Otherwise, for .NET 6+, warnings are on by default -->
     <SuppressTrimAnalysisWarnings Condition="'$(SuppressTrimAnalysisWarnings)' == ''">false</SuppressTrimAnalysisWarnings>
   </PropertyGroup>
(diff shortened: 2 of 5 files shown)


agocke:
I pushed a lot of stuff in this change into the task for two reasons: 1) it was getting really complicated to do this in MSBuild properties, and 2) it was much easier to test.

@MichalStrehovsky this will probably look different for NativeAOT, but I haven't looked at the code yet to see how.

src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
```diff
@@ -212,6 +221,10 @@ Copyright (c) .NET Foundation. All rights reserved.
          potentially problematic when warnings are suppressed. -->
     <NETSdkInformation Condition="'$(PublishTrimmed)' == 'true' And '$(SuppressTrimAnalysisWarnings)' == 'true'" ResourceName="ILLinkOptimizedAssemblies" />
 
+    <!-- In .NET 7, TrimmerDefaultAction is deprecated. TrimMode can be used for the supported configurations. -->
+    <Warning Condition="'$(TrimmerDefaultAction)' != '' And $([MSBuild]::VersionGreaterThan('$(TargetFrameworkVersion)', '6.0'))"
+             Text="Property 'TrimmerDefaultAction' is deprecated in .NET 7." />
```
MichalStrehovsky, line 226:
```suggestion
             Text="Property 'TrimmerDefaultAction' is deprecated in .NET 7 and will be ignored." />
```

Could even go further and be a bit more descriptive on what to do.

MichalStrehovsky (approved):
Looks good to me but I would like Sven to see this too.

Happy to see the TrimMode/TrimmerDefaultAction confusion going away. I kept telling people to forget TrimMode exists. It's now getting a new/better life.

I assume we'll make sure ASP.NET doesn't get insta-broken by this by adding a `<TrimMode>partial</>` somewhere in their SDK.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -191,6 +191,13 @@ public class ILLink : ToolTask
 		public bool RemoveSymbols { set => _removeSymbols = value; }
 		bool? _removeSymbols;
 
+		/// <summary>
+		/// The version of the link task. Used to determine the meaning of certain properties.
+		/// Roughly maps to the TargetFrameworkVersion.
+		/// </summary>
+		[Required]
+		public int LinkVersion { get; set; }
```
vitek-karas, line 199: Works - but I would prefer if we simply passed something like TargetFrameworkVersion - could be the same numerical value, but make it clear that it's the TFM version - I don't want to introduce yet another version number to think about.

src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
```diff
@@ -51,8 +51,8 @@ Copyright (c) .NET Foundation. All rights reserved.
   <PropertyGroup Condition="'$(SuppressTrimAnalysisWarnings)' == '' And '$(PublishTrimmed)' == 'true' And '$(EnableTrimAnalyzer)' != 'true'">
     <!-- Trim analysis warnings are suppressed for .NET < 6. -->
     <SuppressTrimAnalysisWarnings Condition="$([MSBuild]::VersionLessThan('$(TargetFrameworkVersion)', '6.0'))">true</SuppressTrimAnalysisWarnings>
-    <!-- Suppress for WPF/WinForms (unless linking everything) -->
-    <SuppressTrimAnalysisWarnings Condition="'$(TrimmerDefaultAction)' != 'link' And ('$(UseWpf)' == 'true' Or '$(UseWindowsForms)' == 'true')">true</SuppressTrimAnalysisWarnings>
+    <!-- Suppress for WPF/WinForms -->
+    <SuppressTrimAnalysisWarnings Condition="'$(UseWpf)' == 'true' Or '$(UseWindowsForms)' == 'true'">true</SuppressTrimAnalysisWarnings>
```
vitek-karas, line 55: Why? If I trim WinForms app and ask for TrimMode=full, I should get all the warnings. Similarly I would expect this behavior in all other places, basically if I say TrimMode=full - I should get warnings by default (since I asked for the "hard " mode).

src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
vitek-karas, line 226: Please make the warning actionable - maybe just add "Please use TrimMode instead" or something like that.

src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
```diff
@@ -225,16 +238,9 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-      <TrimMode Condition=" '$(TrimMode)' == '' ">copyused</TrimMode>
-      <!-- Action is the same regardless of whether the assembly has an IsTrimmable attribute. (The attribute didn't exist until .NET 6.) -->
-      <TrimmerDefaultAction>$(TrimMode)</TrimmerDefaultAction>
-    </PropertyGroup>
     <PropertyGroup>
-      <TrimMode Condition=" '$(TrimMode)' == '' ">link</TrimMode>
-      <!-- For .NET 6+, assemblies without IsTrimmable attribute get the "copy" action. -->
-      <TrimmerDefaultAction Condition=" '$(TrimmerDefaultAction)' == '' ">copy</TrimmerDefaultAction>
+      <!-- Set TrimmerDefaultAction for compat -->
+      <_TrimmerDefaultAction>$(TrimmerDefaultAction)</_TrimmerDefaultAction>
```
vitek-karas, line 243: What's the reason to introduce a private version of this property - we don't seem to use it for anything really.

akoeplinger:
This will also need a change in xamarin-android since they set the TrimmerDefaultAction: https://github.com/xamarin/xamarin-android/blob/2ca2a103d4e473eeeceecd7689961038bb374cdb/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets#L67

/cc @jonathanpeppers

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -441,8 +448,35 @@ protected override string GenerateResponseFileCommands ()
...
 				args.AppendLine ("-b");
 
-			if (TrimMode != null)
-				args.Append ("--trim-mode ").AppendLine (TrimMode);
+			string trimMode;
+			if (TrimMode != null) {
+				trimMode = TrimMode switch {
+					"full" => "link",
+					"partial" => "link",
+					var x => x
```
marek-safar, line 456: Should this warn instead of passing an empty string to linker and fail there?

src/ILLink.Tasks/LinkTask.cs
agocke: I tried to make that work, but that means we have to do all of the mapping logic in MSBuild properties. That seems much harder to get right, and also very hard to test

src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
agocke: It's currently blocked off completely, which is why we hide the warnings at all. I think we should either show the warnings or disallow trimming.

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -323,10 +330,10 @@ protected override string GenerateResponseFileCommands ()
 
 				args.Append ("-reference ").AppendLine (Quote (assemblyPath));
 
-				string trimMode = assembly.GetMetadata ("TrimMode");
-				if (!String.IsNullOrEmpty (trimMode)) {
+				string assemblyTrimMode = assembly.GetMetadata ("TrimMode");
```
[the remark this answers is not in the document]
agocke: Right now I think it will just error in the linker, which seems like an OK solution for now

agocke:
In the process of switching over to setting defaults in the targets, but fixing things up in the tasks (with the intent of eventually passing things all the way through to the linker). I still have to set up SDK testing for this.

src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
vitek-karas: OK - if we rely on the error for this, then I agree. We should at least add a comment here (why it's OK to suppress). Ideally if the error is disabled we would reenable the warnings (this should really only affect analyzer anyway, the error will trigger before we get to run the linker anyway)

src/ILLink.Tasks/LinkTask.cs
```diff
@@ -323,10 +341,20 @@ protected override string GenerateResponseFileCommands ()
...
+				string isTrimmable = assembly.GetMetadata ("IsTrimmable");
+				if (string.IsNullOrEmpty(assemblyTrimMode)) {
+					if (isTrimmable.Equals("true", StringComparison.OrdinalIgnoreCase)) {
+						// isTrimmable ~= true
+						assemblyTrimMode = trimMode;
+					} else if (isTrimmable.Equals("false", StringComparison.OrdinalIgnoreCase)) {
+						// isTrimmable ~= false
+						assemblyTrimMode = defaultAction;
```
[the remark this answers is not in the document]
MichalStrehovsky: Should we instead deprecate setting the IsTrimmable metadata at MSBuild level?

We have the `TrimmableAssembly` and `TrimmerRootAssembly` ItemGroups. Do we still need the mechanism where one has to create a custom target and then massage `ManagedAssemblyToLink` ItemGroup into a shape the user wants?

src/ILLink.Tasks/LinkTask.cs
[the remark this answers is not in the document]
agocke: I think this is resolved

agocke:
Yup that’s right

src/ILLink.Tasks/LinkTask.cs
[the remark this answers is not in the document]
agocke: I took your suggestion and set `false` to mean `copy`. I think that's the most predictable behavior. I agree that we don't necessarily need to document or recommend this to people.

PR #2858, opened 2022-06-22
Keep unresolved interface impl

jtschuster:
In library mode with `--ckip-unresolved true`, we don't mark interfaceImpl's on public types by default if the interface is not resolved. This change makes sure we mark the interfaceImpl regardless if the `unusedinterfaces` optimization is off.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -3526,7 +3526,7 @@ namespace Mono.Linker.Steps
 			if (!Context.IsOptimizationEnabled (CodeOptimizations.UnusedInterfaces, type))
 				return true;
 
-			if (Context.Resolve(iface.InterfaceType) is not TypeDefinition resolvedInterfaceType)
+			if (Context.Resolve (iface.InterfaceType) is not TypeDefinition resolvedInterfaceType)
 				return false;
 
 			if (Annotations.IsMarked (resolvedInterfaceType))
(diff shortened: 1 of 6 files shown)


vitek-karas (approved):
Just curious - where did we hit this in the runtime repo?
(Does it mean that setup to run linker in runtime is missing assemblies?)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3523,22 +3518,26 @@ protected virtual void MarkInstruction (Instruction instruction, MethodDefinitio
...
 		}
 
-		protected virtual bool ShouldMarkInterfaceImplementation (TypeDefinition type, InterfaceImplementation iface, TypeDefinition resolvedInterfaceType)
+		protected virtual bool ShouldMarkInterfaceImplementation (TypeDefinition type, InterfaceImplementation iface)
 		{
 			if (Annotations.IsMarked (iface))
 				return false;
 
-			if (Annotations.IsMarked (resolvedInterfaceType))
+			if (!Context.IsOptimizationEnabled (CodeOptimizations.UnusedInterfaces, type))
```
marek-safar, line 3526: Should not this also check for skip-unresolved when someone runs with `skip-unresolved false` and `UnusedInterfaces false` it prints the correct error?

jtschuster:
> Just curious - where did we hit this in the runtime repo?
> 
> (Does it mean that setup to run linker in runtime is missing assemblies?)

I think we never actually hit this in runtime - we do provide all the needed assemblies when we run linker in the build. I hit this when I was trying to repro an issue from runtime.

PR #2861, opened 2022-06-23
Update ref package used by analyzer tests to Preview 5

vitek-karas:
This brings in some of the AOT annotations in the framework and thus requires suppressions in some tests.
(For example MakeGenericType now produces IL3050 in the analyzer).


LakshanF (approved):
LGTM

PR #2869, opened 2022-06-28
Fix assignment to captured invalid operation

sbomer:
Fixes https://github.com/dotnet/linker/issues/2847

I added a few basic tests for sources that have invalid syntax (to generate IInvalidOperation). Only the `AssignmentTargetIsCapturedInvalidOperation` testcase was actually hitting the failure in `CapturedReferenceValue` but I think it's good to have a little more coverage.


agocke (approved):
LGTM

vitek-karas (approved):

PR #2875, opened 2022-07-01
Adds a test with two annotated out params for the same method.

vitek-karas:
Test for https://github.com/dotnet/linker/issues/2874.


vitek-karas:
@jtschuster could you please take a look - it's just a test, no functional changes.

jtschuster (approved):

PR #2886, opened 2022-07-11
Preserve all interfaces in library mode

jtschuster:
Annotate all interfaces as fully preserved when an assembly in in library mode. This will keep all interfaces and all interface methods in library mode regardless of visibility or use.

Fixes https://github.com/dotnet/linker/issues/2881

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Libraries/RootLibrary.cs
+++ b/test/Mono.Linker.Tests.Cases/Libraries/RootLibrary.cs
@@ -214,11 +214,13 @@ namespace Mono.Linker.Tests.Cases.Libraries
 			[Kept]
 			public void InternalInterfaceMethod () { }
 
+			[Kept]
 			void IInternalInterface.ExplicitImplementationInternalInterfaceMethod () { }
 
 			[Kept]
 			public static void InternalStaticInterfaceMethod () { }
 
+			[Kept]
 			static void IInternalStaticInterface.ExplicitImplementationInternalStaticInterfaceMethod () { }
 
 			[Kept]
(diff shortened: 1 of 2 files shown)


test/Mono.Linker.Tests.Cases/Libraries/RootLibrary.cs
```diff
@@ -214,11 +214,13 @@ public class UninstantiatedPublicClassWithInterface :
 			[Kept]
 			public void InternalInterfaceMethod () { }
 
+			[Kept]
```
vitek-karas, line 217: We're probably missing a test here (from before).
Before this change something like:
```C#
internal interface IFace
{
    void Method();
}

internal class TestType : IFace
{
    IFace.Method() { }
}

void Test()
{
    new TestType();
    // IFace is never mentioned anywhere
}
```

In this case we should have removed IFace as a whole. Now we should keep it.
I was expecting to see a test which would fail on this (or rather that you had to change it, but I don't see that anywhere).

jtschuster: If I understand correctly, `IInternalInterface` was essentially the same situation, but we never got rid of it because with the `UnusedInterfaces` optimization off in library mode we had to keep the interface and the interface implementation.

vitek-karas: Ah - so it's the combination of two things which basically makes this untestable... OK I guess it makes sense.

vitek-karas (approved):
Looks good - but @sbomer should take a look since he knows this better :wink:

MichalStrehovsky:
> Fixes https://github.com/dotnet/linker/issues/2881

To fix that issue, it would be sufficient to limit this to instance virtual methods (or even instance abstract, since COM doesn't work with default interface methods). But "all interface methods" will work too.

jtschuster:
For reference, this change adds 17280 bytes to the shared framework after trimming on Linux-x64-Release 

81065987 on main
81083267 with the change

PR #2889, opened 2022-07-12
Intentionally skip switch instructions during branch removal as the code doesn't handle it yet (throws NIE).

vitek-karas:
This is more of a "workaround" to https://github.com/dotnet/linker/issues/2872

I created https://github.com/dotnet/linker/issues/2888 to track the work to add full support for switch instruction in branch removal.


src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
```diff
@@ -892,7 +896,7 @@ bool RemoveConditions ()
 							}
 
 							// Common pattern generated by C# compiler in debug mode
-							if (i > 3 && GetConstantValue (FoldedInstructions[i - 3], out operand) && operand is int opint2 && IsPairedStlocLdloc (FoldedInstructions[i - 2], FoldedInstructions[i - 1])) {
+							if (i >= 3 && GetConstantValue (FoldedInstructions[i - 3], out operand) && operand is int opint2 && IsPairedStlocLdloc (FoldedInstructions[i - 2], FoldedInstructions[i - 1])) {
```
vitek-karas, line 899: This fixes the case where the switch is the first thing in a method body - in which case the logic here would not kick in (now switch is disabled above, but if it's not it would not be recognized due to this).

agocke: Sorry, I'm having trouble visualizing the kind of code this is matching. Could you write a comment with an example?

vitek-karas: Exactly - for some reason Roslyn always put the stloc/ldloc pair even in optimized build - in debug build there are actually two pairs like that one after another. The code would not handle the debug build at all right now, but that's for when we actually support switch.

PR #2890, opened 2022-07-12
DynamicallyAccessedMembers CodeFix

iserrato:
- Reorganized the DAM CodeFix to not rely on 'BaseAttributeCodeFixProvider`, as the DAM CodeFix requires different functionality for placing attributes on parameters, returns, etc.
- Create test cases for supported Diagnostic IDs


src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+			var diagDescriptorsArrayBuilder = ImmutableHashSet.CreateBuilder<string> ();
+			diagDescriptorsArrayBuilder.Add (DiagnosticId.DynamicallyAccessedMembersOnMethodReturnValueCanOnlyApplyToTypesOrStrings.AsString ());
+			diagDescriptorsArrayBuilder.Add (DiagnosticId.DynamicallyAccessedMembersMismatchOnMethodReturnValueBetweenOverrides.AsString ());
+			return diagDescriptorsArrayBuilder.ToImmutable ();
+		}
+
+		protected static SyntaxNode[] GetAttributeArguments (ISymbol attributableSymbol, ISymbol targetSymbol, SyntaxGenerator syntaxGenerator, Diagnostic diagnostic)
+		{
+			var symbolDisplayName = targetSymbol.GetDisplayName ();
+			if (string.IsNullOrEmpty (symbolDisplayName) || HasPublicAccessibility (attributableSymbol))
```
agocke, line 45: `IsNullOrEmpty` is usually wrong -- the source should probably pick either null or empty to return if it wants a sentinel value.

Also, the call to `HasPublicAccessibility` looks wrong here. I think that makes sense for RUC but not for DAM.

src/ILLink.RoslynAnalyzer/DynamicallyAccessedMembersAnalyzer.cs
```diff
@@ -34,15 +35,15 @@ static ImmutableArray<DiagnosticDescriptor> GetSupportedDiagnostics ()
 			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersOnMethodReturnValueCanOnlyApplyToTypesOrStrings));
 			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersFieldAccessedViaReflection));
 			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection));
-			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.UnrecognizedTypeInRuntimeHelpersRunClassConstructor));
```
agocke, line 37: I think these changes can be reverted

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,6 +21,122 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
+			
+			public class Foo 
+			{
+			}
+			
+			class C
+			{
+				public static void Main()
+				{
+					M(typeof(Foo));
```
agocke, line 64: You could probably simplify this test by using `typeof(C)` and deleting the `Foo` type

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,6 +21,122 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
+			var test = @"
+using System;
+using System.Diagnostics.CodeAnalysis;
+public class Foo
+{
+}
+
+class C
+{
+    private static Type f = typeof(Foo);
```
agocke, line 109: Let's also add a test for a public field to ensure the accessibility doesn't affect anything

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+					document, attributableNode, attributeArguments, attributeSymbol, ReturnAttribute, ct),
+				equivalenceKey: codeFixTitle), diagnostic);
+		}
+
+		private static async Task<Document> AddAttributeAsync (
+			Document document,
+			SyntaxNode targetNode,
+			SyntaxNode[] attributeArguments,
+			ITypeSymbol attributeSymbol,
+			bool AddAsReturnAttribute,
```
agocke, line 107: Variable casing looks wrong

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+			var document = context.Document;
+			var root = await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false);
+			var diagnostic = context.Diagnostics.First ();
+			var props = diagnostic.Properties;
+			var model = await document.GetSemanticModelAsync (context.CancellationToken).ConfigureAwait (false);
+			SyntaxNode diagnosticNode = root!.FindNode (diagnostic.Location.SourceSpan);
+			var attributableSymbol = (diagnosticNode is InvocationExpressionSyntax invocationExpression
+					&& invocationExpression.Expression is MemberAccessExpressionSyntax simpleMember
+					&& simpleMember.Expression is IdentifierNameSyntax name) ? model.GetSymbolInfo (name).Symbol : null;
+
```
agocke, line 78:
```suggestion
			var attributableSymbol = (diagnosticNode is InvocationExpressionSyntax { 
			    Expression: MemberAccessExpressionSyntax { Expression: IdentifierNameSyntax name } }) 
					? model.GetSymbolInfo (name).Symbol 
					: null;

```

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,137 @@
...
+		public sealed override ImmutableArray<string> FixableDiagnosticIds => SupportedDiagnostics.Select (dd => dd.Id).ToImmutableArray ();
+
+		private protected static LocalizableString CodeFixTitle => new LocalizableResourceString (nameof (Resources.DynamicallyAccessedMembersCodeFixTitle), Resources.ResourceManager, typeof (Resources));
+
+		private protected static string FullyQualifiedAttributeName => DynamicallyAccessedMembersAnalyzer.FullyQualifiedDynamicallyAccessedMembersAttribute;
+
+		public sealed override Task RegisterCodeFixesAsync (CodeFixContext context) => BaseRegisterCodeFixesAsync (context);
+
+		static ImmutableHashSet<string> AttriubuteOnReturn ()
```
agocke, line 34:
```suggestion
		static readonly ImmutableArray<string> AttributesOnReturn =
```

Since this doesn't change I'd recommend using a field to build it once. Also, hash sets don't provide a perf benefit until you hit around 8 items.

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+
+namespace ILLink.CodeFix
+{
+	public class DAMCodeFixProvider : Microsoft.CodeAnalysis.CodeFixes.CodeFixProvider
+	{
+		public static ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => DynamicallyAccessedMembersAnalyzer.GetSupportedDiagnostics ();
+
+		public sealed override ImmutableArray<string> FixableDiagnosticIds => SupportedDiagnostics.Select (dd => dd.Id).ToImmutableArray ();
+
+		private protected static LocalizableString CodeFixTitle => new LocalizableResourceString (nameof (Resources.DynamicallyAccessedMembersCodeFixTitle), Resources.ResourceManager, typeof (Resources));
```
tlakollo, line 28: This line adds a new resource but I don't see the resource file being changed. So I'm guessing it will be missing

src/ILLink.CodeFix/DAMCodeFixProvider.cs
iserrato: The `GetDisplayName` function looks like it should never return null, so maybe we just check for empty?

`HasPublicAccessibility` was left over from `BaseAttributeCodeFixProvider`, and it's been used in each of the code fixers - let me know if it is relevant, otherwise I'll just remove it

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,6 +21,122 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
+			}
+			test.FixedState.ExpectedDiagnostics.AddRange (fixedExpected);
+			return test.RunAsync ();
+		}
+
+		[Fact]
+		public async Task DiagID_2070 ()
+		{
+			var test = @"
+			using System;
```
tlakollo, line 53: Nit: Just to keep consistency this test has additional spaces at the beginning of each line while the others don't

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,6 +21,122 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
+{
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]
+    private static Type f = typeof(Foo);
+
+    public static void Main()
+    {
+        f.GetMethod(""Bar"");
+	}
+}";
+			// baselineExpected: Array.Empty<DiagnosticResult> (), fixedExpected: Array.Empty<DiagnosticResult> ());
```
tlakollo, line 133: Nit: This comment seems like a leftover since you do have baselineExpected

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
```diff
@@ -278,7 +278,7 @@ public class C
     [RequiresUnreferencedCodeAttribute(""message"")]
     public int M1() => 0;
 
-    int M2 => M1();
+    int M2 => {|#0:M1()|};
```
tlakollo, line 281: Just to educate myself, could you elaborate on these changes? Im not sure what is different here and why the location is 0. Is it the fact that is a lambda?

src/ILLink.CodeFix/DAMCodeFixProvider.cs
agocke: Yeah I would just remove anything that was from the `Base` that doesn't make sense here.

agocke: Also, is there any place that GetDisplayName can return empty string?

We should either have a test case for that, or remove it if it's impossible

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Simplification;
+
+namespace ILLink.CodeFix
+{
+	public class DAMCodeFixProvider : Microsoft.CodeAnalysis.CodeFixes.CodeFixProvider
+	{
+		public static ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => DynamicallyAccessedMembersAnalyzer.GetSupportedDiagnostics ();
+
+		public sealed override ImmutableArray<string> FixableDiagnosticIds => SupportedDiagnostics.Select (dd => dd.Id).ToImmutableArray ();
```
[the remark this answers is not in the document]
iserrato: sure thing

src/ILLink.CodeFix/DAMCodeFixProvider.cs
iserrato: I'm honestly not sure what this line does - what is the resource file?

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+
+		private protected static LocalizableString CodeFixTitle => new LocalizableResourceString (nameof (Resources.DynamicallyAccessedMembersCodeFixTitle), Resources.ResourceManager, typeof (Resources));
+
+		private protected static string FullyQualifiedAttributeName => DynamicallyAccessedMembersAnalyzer.FullyQualifiedDynamicallyAccessedMembersAttribute;
+
+		public sealed override Task RegisterCodeFixesAsync (CodeFixContext context) => BaseRegisterCodeFixesAsync (context);
+
+		static readonly ImmutableArray<string> AttributesOnReturn =
```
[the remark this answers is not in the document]
iserrato: yes, it looks like the `RegisterCodeFixesAsync` is unused, but `AttributesOnReturn` is

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+			var attributableSymbol = (diagnosticNode is InvocationExpressionSyntax { 
+			    Expression: MemberAccessExpressionSyntax { Expression: IdentifierNameSyntax name } }) 
+					? model.GetSymbolInfo (name).Symbol 
+					: null;
+
+			if (attributableSymbol is null) return;
+
+			var attributableNodeList = attributableSymbol.DeclaringSyntaxReferences;
+
+			var attributableNode = (attributableNodeList.Length == 1) ? attributableNodeList[0].GetSyntax () : null;
```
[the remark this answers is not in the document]
iserrato: sure!

src/ILLink.CodeFix/DAMCodeFixProvider.cs
tlakollo: You have had some experience with them, we use resource files for the diagnostic messages (see the [SharedStrings.resx](https://github.com/dotnet/linker/blob/main/src/ILLink.Shared/SharedStrings.resx) file). We use them as a container for strings, that then we can use Localization to translate to different languages. In this case, in order to retrieve the string you need to add an entry to [Resources.resx](https://github.com/dotnet/linker/blob/main/src/ILLink.CodeFix/Resources.resx), which is the place where we place the other CodeFixer tittles.

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+
+			if (attributableNode is null) return;
+			var targetSymbol = model!.GetSymbolInfo (diagnosticNode).Symbol!;
```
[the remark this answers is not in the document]
iserrato: I've renamed it to diagnosticSymbol - does that make more sense?

src/ILLink.CodeFix/DAMCodeFixProvider.cs
tlakollo: By the way is going to be interesting going forward, because unlike RequiresAttribute the DynamicallyAccessedMembersAttribute CodeFixer will place attributes in things that are not "the parent" so the message needs to be different accordingly. I would expect a matrix for how to generate the fix using the code fixer and decide the resource message to use in that same matrix, similar to what happens for diagnostic generation. For now, I guess is fine to have only one message.

iserrato: Got it, thanks! Yes, the DynamicallyAccessedMembers gets added to all sorts of locations which makes this code fixer a little more complicated to implement.

src/ILLink.CodeFix/DAMCodeFixProvider.cs
iserrato: It was pointed out to me that symbolDisplayName is unused, so we don't need this check at all. I've taken it out

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,138 @@
...
+			var attributeSymbol = model!.Compilation.GetTypeByMetadataName (FullyQualifiedAttributeName)!;
+			var attributeArguments = GetAttributeArguments (attributableSymbol, targetSymbol, SyntaxGenerator.GetGenerator (document), diagnostic);
+			var codeFixTitle = CodeFixTitle.ToString ();
+			var returnAttributes = AttriubuteOnReturn ();
+			if (returnAttributes.Contains (attributeSymbol.ToString ())) {
```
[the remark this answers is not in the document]
iserrato: This had faulty logic - I've updated it now so it's if the `returnAttributes` contains the `diagnostic.Id`, because certain diagnostics always have attributes on the return type (is my understanding). This isn't implemented yet, but would be necessary to have when we do implement fixers for these DiagnosticIds. Is it best practice to leave it in or should I take it out and add it back later?

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
iserrato: that worked!

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,6 +21,122 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
+			
+			class C
+			{
+				public static void Main()
+				{
+					M(typeof(Foo));
+				}
+                static void M([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type T)
+				{
+                    T.GetMethods();
```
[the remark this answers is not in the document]
iserrato: I added a test, but using BindingFlags requires System.Reflection, and now there is an additional error surrounding the `DynamicallyAccessedMembersviaReflection` on the fixed version. The original `DynamicallyAccessedMembersMismatchParameterTargetsThisParameter` is addressed, so is this fine? 

Also, it looks like the CodeFix's current logic is to apply both `PublicMethods` and 'NonPublicMethods` flags, and I'm having a difficult time realizing why it's adding both every time

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
iserrato: sure thing!

test/ILLink.RoslynAnalyzer.Tests/RequiresUnreferencedCodeAnalyzerTests.cs
iserrato: Realizing I probably shouldn't have checked these in - the gist of it is that instead of using spans, you can specify locations using this syntax. In this case, the 0 indicates that it's location 0, which you can then reference in the VerifyCS line below. There's nothing actually different, it's just a readability change. At the time, I didn't realize that the Span arguments were automatically generated, so we can leave it the way it is for now. I'll revert the changes

iserrato: @jkoritzinsky would know more about this if you have further questions

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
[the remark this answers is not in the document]
tlakollo: Answering about the current logic -> what the syntax generator is actually doing is reading the annotations that we have on the runtime for the `GetMethods(BindingFlags)` method
```C#
targetSymbol.GetAttributes ().First (attr => attr.AttributeClass?.ToDisplayString () == DynamicallyAccessedMembersAnalyzer.FullyQualifiedDynamicallyAccessedMembersAttribute).ConstructorArguments[0])
```
In this case, the target symbol is `System.Type.GetMethods(System.Reflection.BindingFlags)` we directly search for an attribute that's DAM, and get the arguments in the constructor.
The annotation in `GetMethods()` and `GetMethods(BindingFlags)` is different (see [TypeExtensions.cs](https://github.com/dotnet/runtime/blob/215b39abf947da7a40b0cb137eab4bceb24ad3e3/src/libraries/System.Reflection.TypeExtensions/src/System/Reflection/TypeExtensions.cs#L214-L229)). The annotation in Runtime assumes that the BindingFlags are unknown so it keeps both Public and NonPublic. Maybe we could do better and actually analyze what are the actual requirements, not sure if it's in the scope of this PR to do that tough.

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+
+		protected static SyntaxNode[] GetAttributeArguments (ISymbol targetSymbol, SyntaxGenerator syntaxGenerator, Diagnostic diagnostic)
+		{
+			// if (diagnostic.Id == DiagnosticId.DynamicallyAccessedMembersFieldAccessedViaReflection.AsString ()) {
+			// 	return new[] { syntaxGenerator.AttributeArgument (syntaxGenerator.BitwiseOrExpression (syntaxGenerator.DottedName ("System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.NonPublicFields"), syntaxGenerator.DottedName ("System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicFields"))) };
+			// } else if (diagnostic.Id == DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection.AsString ()) {
+			// 	return new[] { syntaxGenerator.AttributeArgument (syntaxGenerator.BitwiseOrExpression (syntaxGenerator.DottedName ("System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.NonPublicMethods"), syntaxGenerator.DottedName ("System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicMethods"))) };
+			// } else
```
tlakollo, line 51: Nit: I wouldn't leave the commented code, once we are ready to add the support we add the code meanwhile it shouldn't be there

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,6 +21,568 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
+			if (numberOfIterations != null) {
+				test.NumberOfIncrementalIterations = numberOfIterations;
+				test.NumberOfFixAllIterations = numberOfIterations;
+			}
+			test.FixedState.ExpectedDiagnostics.AddRange (fixedExpected);
+			return test.RunAsync ();
+		}
+#region CodeFixTests
+		[Fact]
+		public async Task DiagID_2070 ()
```
tlakollo, line 50: Nit: the name of the tests is kind of not consistent (DiagID_2070 vs DAM2092 vs MismatchParameter2069). You also need to consider that since these are [Facts] they will appear in the TestExplorer as individual entries, and having a descriptive name does give a reward in developer experience. For example, I would list in the name that these tests are about code fixing stuff. Since now DynamicallyAccessedMembersAnalyzerTests has both, analysis-only tests and code fix tests.

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,6 +21,568 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
+using System.Reflection;
+using System.Diagnostics.CodeAnalysis;
+
+class C
+{
+	public static void Main()
+	{
+		M(typeof(C));
+	}
+	static void M([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicMethods)] Type T)
```
tlakollo, line 118: PR is failing because the code fixer adds `DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods`

src/ILLink.CodeFix/DAMCodeFixProvider.cs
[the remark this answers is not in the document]
iserrato: I'll take it out, and we can add it back in if we decide to support the return diagnostics later on

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -39,20 +48,17 @@ static ImmutableHashSet<string> AttriubuteOnReturn ()
...
+			// 	return new[] { syntaxGenerator.AttributeArgument (syntaxGenerator.BitwiseOrExpression (syntaxGenerator.DottedName ("System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.NonPublicFields"), syntaxGenerator.DottedName ("System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicFields"))) };
+			// } else if (diagnostic.Id == DiagnosticId.DynamicallyAccessedMembersMethodAccessedViaReflection.AsString ()) {
+			// 	return new[] { syntaxGenerator.AttributeArgument (syntaxGenerator.BitwiseOrExpression (syntaxGenerator.DottedName ("System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.NonPublicMethods"), syntaxGenerator.DottedName ("System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicMethods"))) };
+			// } else 
+			if (diagnostic.Id == DiagnosticId.DynamicallyAccessedMembersMismatchParameterTargetsThisParameter.AsString () || diagnostic.Id == DiagnosticId.DynamicallyAccessedMembersMismatchFieldTargetsThisParameter.AsString () || diagnostic.Id == DiagnosticId.DynamicallyAccessedMembersMismatchParameterTargetsMethodReturnType.AsString ()) {
```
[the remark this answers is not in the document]
iserrato: yes, but it's a little funky - the diagnostic.Id is formatted as IL####, and we only want the numbers. I've found it works if I take a substring of the last four numbers, do you know if there's a better way to fix this?

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
iserrato: I've standardized the names. but do you have any suggestions on how to make them specific? I'm not sure that the diagnostic ID number is very important, should I just include something more like `CodeFix_MismatchParamTargetsParam_NonPublicMethods` if I'm referring to ID 2070 and looking at a case that uses `NonPublicMethods`?

tlakollo: My preference here is that is better to give as many clues as possible, even adding the id is fine since they don't tend to change. Something like `CodeFix_IL2070_MismatchParamTargetsParam_NonPublicMethods`

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,10 +22,618 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
+
+    public static void Main()
+    {
+        f.GetMethod("Bar");
+	}
+}
+""";
+			await VerifyDynamicallyAccessedMembersCodeFix (test, fixtest, new[] {
+				// /0/Test0.cs(14,9): warning IL2080: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethod(String)'. The field 'C.f' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
+                VerifyCS.Diagnostic(DiagnosticId.DynamicallyAccessedMembersMismatchFieldTargetsThisParameter).WithSpan(13, 9, 13, 27).WithArguments("System.Type.GetMethod(String)", "C.f", "'DynamicallyAccessedMemberTypes.PublicMethods'")},
```
tlakollo, line 178: nit: Indentation is off, should match the comment

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,10 +22,618 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
```
tlakollo, line 222: Nit: indentation is off

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -381,14 +948,15 @@ public static void Main()
 		[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type type)
 	{
 	}
-}";
+}
+""";
 
 			// (15,9): warning IL2077: 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'C.NeedsPublicMethods(Type)'.
```
tlakollo, line 954: Nit: now line number is 14

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -420,22 +988,23 @@ private static Type M()
 	{
         return f;
 	}
-}";
+}
+""";
 
 			// (21,9): warning IL2078: 'C.M()' method return value does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' requirements.
```
tlakollo, line 994: Nit: now line number is 20

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -757,19 +1330,20 @@ private void M1()
         {
         }
     }
-}";
+}
+""";
 
 			return VerifyDynamicallyAccessedMembersAnalyzer (string.Concat (GetSystemTypeBase (), ConversionOperation),
```
tlakollo, line 1336: nit: missing warning message

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -482,17 +1052,18 @@ class C
...
     {
-        f.GetMethod(""Bar"");
+        f.GetMethod("Bar");
 	}
-}";
+}
+""";
 			// The warning will be generated once dataflow is able to handle GetMethod intrinsic
 
 			// (14,9): warning IL2080: 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethod(String)'.
```
tlakollo, line 1061: Nit: line is now 13

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -797,18 +1371,19 @@ private void M1()
         {
         }
     }
-}";
+}
+""";
 
 			return VerifyDynamicallyAccessedMembersAnalyzer (string.Concat (GetSystemTypeBase (), AnnotatedConversionOperation),
```
tlakollo, line 1377: nit: missing warning message

tlakollo (approved):
Minor things, I think in general looks good

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -187,17 +749,18 @@ public static void Main()
 
     private static void M(Type type)
     {
-        type.GetMethod(""Bar"");
+        type.GetMethod("Bar");
 	}
```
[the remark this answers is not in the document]
iserrato: realizing that this is because some braces are tabbed and others are spaced - I'll go through and fix this now

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		public override async Task RegisterCodeFixesAsync (CodeFixContext context)
+		{
+			var document = context.Document;
+			var root = await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false);
+			var diagnostic = context.Diagnostics.First ();
+			var props = diagnostic.Properties;
+			var model = await document.GetSemanticModelAsync (context.CancellationToken).ConfigureAwait (false);
+			SyntaxNode diagnosticNode = root!.FindNode (diagnostic.Location.SourceSpan);
```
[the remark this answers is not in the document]
iserrato: Should I add a test case associated with this fix?

src/ILLink.CodeFix/DAMCodeFixProvider.cs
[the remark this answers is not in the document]
iserrato: By implementing the change from  https://github.com/dotnet/linker/pull/2900/files, the targetSymbol is now created on a different line - should I still put a note in? 

`if (model.GetSymbolInfo (diagnosticNode).Symbol is not { } targetSymbol)
				return;`

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,114 @@
...
+namespace ILLink.CodeFix
+{
+	public class DAMCodeFixProvider : Microsoft.CodeAnalysis.CodeFixes.CodeFixProvider
+	{
+		public static ImmutableArray<DiagnosticDescriptor> GetSupportedDiagnostics ()
+		{
+			var diagDescriptorsArrayBuilder = ImmutableArray.CreateBuilder<DiagnosticDescriptor> ();
+			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersMismatchParameterTargetsThisParameter));
+			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersMismatchFieldTargetsThisParameter));
+			diagDescriptorsArrayBuilder.Add (DiagnosticDescriptors.GetDiagnosticDescriptor (DiagnosticId.DynamicallyAccessedMembersMismatchParameterTargetsMethodReturnType));
```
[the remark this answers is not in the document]
iserrato: sure thing!

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,10 +22,632 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
...
+			test.FixedState.ExpectedDiagnostics.AddRange (fixedExpected);
+			return test.RunAsync ();
+		}
+
+		#region CodeFixTests
+		[Fact]
+		public async Task CodeFix_IL2070_MismatchParamTargetsThisParam_PublicMethods ()
+		{
+			var test = $$"""
+using System;
```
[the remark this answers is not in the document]
iserrato: Do you mean indenting the whole string?

src/ILLink.CodeFix/DAMCodeFixProvider.cs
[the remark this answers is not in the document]
agocke: Yup, that's exactly what my change fixes. I would agree -- creating a test case, confirming it throws an exception, then applying the fix and confirming it no longer does so is a good idea.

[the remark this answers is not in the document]
agocke: Ah, actually it looks like `diagnosticNode` is actually properly checked below, s oit won't throw an exception. Probably just won't work (the type will be `ArgumentSyntax` instead of `InvocationExpressionSyntax`).

test/ILLink.RoslynAnalyzer.Tests/DynamicallyAccessedMembersAnalyzerTests.cs
```diff
@@ -19,10 +22,632 @@ static Task VerifyDynamicallyAccessedMembersAnalyzer (string source, params Diag
 				expected: expected);
 		}
 
+		static Task VerifyDynamicallyAccessedMembersCodeFix (
```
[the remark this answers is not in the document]
iserrato: sure thing!

src/ILLink.RoslynAnalyzer/TrimAnalysis/DiagnosticContext.cs
```diff
@@ -27,5 +27,34 @@ public DiagnosticContext (Location location)
...
 			Diagnostics.Add (Diagnostic.Create (DiagnosticDescriptors.GetDiagnosticDescriptor (id), Location, args));
 		}
+
+		public partial void AddDiagnostic (DiagnosticId id, ValueWithDynamicallyAccessedMembers sourceAttribute, params string[] args)
+		{
+			if (Location == null) {
+				return;
+			}
+
+			switch (sourceAttribute) {
```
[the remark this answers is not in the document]
iserrato: Makes sense, I'll take them out!

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -0,0 +1,134 @@
...
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.Simplification;
+
+namespace ILLink.CodeFix
+{
+	public class DAMCodeFixProvider : Microsoft.CodeAnalysis.CodeFixes.CodeFixProvider
```
Youssef1313, line 22: I believe this won't work in Visual Studio. You need to have `[ExportCodefixProvider(LanguageNames.CSharp), Shared]` on the class

PR #2891, opened 2022-07-13
Redundant warning suppressions detection

jkurdek:
Problem description and potential solution overview.

@vitek-karas

### DIFF
--- a/docs/design/redundant-suppressions.md
+++ b/docs/design/redundant-suppressions.md
@@ -2,77 +2,78 @@
 
 Dynamic reflection patterns pose a serious challenge to the linker trimming capabilities. The tool is able to infer simple reflection patterns; but there still cases which the tool will not be able to reason about. When the linker fails to recognize a certain pattern, a warning appears informing the user that the trimming process may break the functionality of the app.
 
-There are cases where the developer is confident about the safety of a given pattern, but the linker is unable to reason about it and still produces a warning. The developer may use warning suppression to silence the warning. An example of such pattern may be listing all properties of an object using reflection mechanism which may be useful for diagnostic purposes.
+There are cases where the developer is confident about the safety of a given pattern, but the linker is unable to reason about it and still produces a warning. The developer may use warning suppression to silence the warning. An example of such pattern may be using methods from a private class using reflection.
 ```csharp
-    [UnconditionalSuppressMessage("trim", "IL2075", Justification = "It's OK to print only the properties which were actually used.")]
-    void PrintProperties(object instance)
+    [DynamicDependency(DynamicallyAccessedMemberTypes.PublicMethods, typeof(NameProvider))]
+    public void Test()
     {
-        foreach (var p in instance.GetType().GetProperties())
-        {
-            PrintPropertyValue(p, instance);
-        }
+        PrintName(typeof(NameProvider));
     }
 
+    [UnconditionalSuppressMessage("trim", "IL2070", Justification = "DynamicDependency attribute will instruct the linker to keep the public methods on NameProvider.")]
+    public void PrintName(Type type)
+    {
+        string name = (string)type.GetMethod("GetName")!.Invoke(null, null)!;
+        Console.WriteLine(name);
+    }
+
+    private class NameProvider
+    {
+        public static string GetName() => "NiceName";
+        public static string suffix = "no really";
+    }
 ```
 
 ## Redundant warnings
-The warning suppression could present a challenge to the software development lifecycle. Let us again consider the above example of listing all properties on an object instance. We can rewrite the code in such a way, that the linker is able to statically reason about it. Then, the warning is no longer issued and the suppression becomes redundant. We should remove it.
+The warning suppression could present a challenge to the software development lifecycle. Let us again consider the above example of accessing methods of a private class. We can rewrite the code in such a way, that the linker is able to reason about it. Then, the warning is no longer issued and the suppression becomes redundant. We should remove it.
 
 ```csharp
-    [UnconditionalSuppressMessage("trim", "IL2075", Justification = "It's OK to print only the properties which were actually used.")] // This should be removed
-    void PrintProperties<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)] T>(T instance)
+    [UnconditionalSuppressMessage("trim", "IL2070", Justification = "DynamicDependency attribute will instruct the linker to keep the public methods on NameProvider.")] // This should be removed
+    public void PrintName([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type type)
     {
-        foreach (var p in typeof(T).GetProperties())
-        {
-            PrintPropertyValue(p, instance);
-        }
+        string name = (string)type.GetMethod("GetName")!.Invoke(null, null)!;
+        Console.WriteLine(name);
     }
-
 ```
 
 If we keep the warning suppression on this trimmer-compatible code, we will end up with a potentially dangerous case. Should we later add some trimmer-incompatible code within the scope of the suppression which triggers the suppressed warning, we will not be informed about it during the trimming process. That is, the warning issued by the linker will be silenced by the suppression we left over and it will not be displayed. This may result in a scenario in which the trimming completes with no warnings, yet errors occur at runtime. 
 
-This can be illustrated with the following example. Let us extend the above example to print properties of previously inspected properties. 
+This can be illustrated with the following example. Let us extend the above code to also print the value of `suffix` field.
 
 ```csharp
-    [UnconditionalSuppressMessage("trim", "IL2075", Justification = "It's OK to print only the properties which were actually used.")]
-    void PrintProperties<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)] T>(T instance)
+    [UnconditionalSuppressMessage("trim", "IL2070", Justification = "DynamicDependency attribute will instruct the linker to keep the public methods on NameProvider.")]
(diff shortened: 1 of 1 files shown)


vitek-karas:
@agocke @sbomer @tlakollo @eerhardt @MichalStrehovsky 
If you could take a quick look - mainly that the problem/approach makes sense to you...

docs/design/redundant-suppressions.md
```diff
@@ -0,0 +1,93 @@
...
+
+Alternatively, we could make the analyzer do the unused warning suppressions detection. The analyzer operates on a single assembly level but it has a full view of the processed assembly, as opposed to the linker which only sees the code which is actually used within the assembly. Making the solution part of the analyzer, would allow us then to detect the unused warning suppression on assembly level with higher precision. Also, an advantage of such solution would be a shorter feedback loop; we would learn about the redundant suppressions way before we run the publish command. The drawback of this approach is the added complexity. We would not be able to reuse the existing components. What is more, the analyzer has a different view of the code than the linker. We may not be able to identify the same set of trimmer-incompatible patterns using the analyzer as we do using the linker.
```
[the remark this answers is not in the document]
tlakollo: I would be worried that the analyzer can recommend removing a suppression just because doesn't have the full picture. Non-actionable code and code that either way is going to be removed seems less worrying to me. 
On the linker side of things, I would be worried if it's implementing the linker tooling is going to recommend removing a suppression because of something we annotated using a feature that the analyzer knows best (like backtracking).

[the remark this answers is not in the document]
tlakollo: Since we are already talking about things that are blurry in the document here:
What's the scope of this feature? my guess is that it should be restricted to a certain amount of warnings and attributes. For example, if it's implemented in the linker we wouldn't be able to detect a redundant UnconditionalSuppressMessage originated by RequiresDynamicCode or RequiresAssemblyFiles, or what is worse it will be flagged always as redundant just because Linker doesn't analyze it. 
Also, is this feature going to be supported by NativeAOT?

docs/design/redundant-suppressions.md
```diff
@@ -0,0 +1,93 @@
...
+
+The warning should be reported in the output of the command.
+
+```
+Trim analysis warning IL2021: Program.PrintProperties<T>(T): Unused UnconditionalSuppressMessageAttribute suppressing the IL2075 warning found. Consider removing the unused warning suppression.
+```
+
+## Other solutions
+
+The proposed solution operates by extending the functionality of the linker tool. On one hand, this allows for reusing the existing components leading to a simple implementation, on the other hand this may lead to potential problems. The linker sees only a part of code which is actually used, that means that the solution would not be able to identify the warning suppressions on the code which is trimmed away. Also, the linker may visit parts of the code which were not authored by the developer (e.g. used libraries) and look for the redundant suppressions there. This may cause the output warnings to be noisy and not useful to the developer. Moreover, the dependencies identified by the linker may be different depending on the environment the tool is run in. Hence, the proposed solution may report different redundant suppressions on different environments.
```
MichalStrehovsky, line 91: Illink also does a bunch of optimizations that may render a suppression that was necessary at source compilation time look unnecessary. E.g. if constant propagation figures out a codepath is unreachable and suddenly the suppression looks unnecessary. but it's just unnecessary for that specific deployment mode.

The analyzer has a better view of the code from this perspective.

docs/design/redundant-suppressions.md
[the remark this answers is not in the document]
jkurdek: Added a note that only suppressions produced by the linker will be checked.

docs/design/redundant-suppressions.md
```diff
@@ -0,0 +1,94 @@
+# Redundant Warning Suppression Detection
+
+Dynamic reflection patterns pose a serious challenge to the linker trimming capabilities. The tool is able to infer simple reflection patterns; but there still cases which the tool will not be able to reason about. When the linker fails to recognize a certain pattern, a warning appears informing the user that the trimming process may break the functionality of the app.
```
vitek-karas, line 3:
```suggestion
Dynamic reflection patterns pose a serious challenge to the linker trimming capabilities. The tool is able to infer simple reflection patterns; but there are still cases which the tool will not be able to reason about. When the linker fails to recognize a certain pattern, a warning appears informing the user that the trimming process may break the functionality of the app.
```

docs/design/redundant-suppressions.md
```diff
@@ -0,0 +1,94 @@
...
+
+There are cases where the developer is confident about the safety of a given pattern, but the linker is unable to reason about it and still produces a warning. The developer may use warning suppression to silence the warning. An example of such pattern may be using methods from a private class using reflection.
```
vitek-karas, line 5:
```suggestion
There are cases where the developer is confident about the safety of a given pattern, but the linker is unable to reason about it and still produces a warning. The developer may use warning suppression to silence the warning. An example of such pattern may be using methods from a class using reflection.
```

docs/design/redundant-suppressions.md
```diff
@@ -0,0 +1,94 @@
...
+        public static string suffix = "no really";
+    }
+```
+
+## Redundant warnings
+The warning suppression could present a challenge to the software development lifecycle. Let us again consider the above example of accessing methods of a private class. We can rewrite the code in such a way, that the linker is able to reason about it. Then, the warning is no longer issued and the suppression becomes redundant. We should remove it.
+
+```csharp
+    [UnconditionalSuppressMessage("trim", "IL2070", Justification = "DynamicDependency attribute will instruct the linker to keep the public methods on NameProvider.")] // This should be removed
+    public void PrintName([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type type)
```
vitek-karas, line 32:
```suggestion
    // Now the DynamicDependencyAttribute can be removed as well as the suppression here
    [UnconditionalSuppressMessage("trim", "IL2070", Justification = "DynamicDependency attribute will instruct the linker to keep the public methods on NameProvider.")]
    public void PrintName([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type type)
```

docs/design/redundant-suppressions.md
```diff
@@ -0,0 +1,94 @@
...
+
+This can be illustrated with the following example. Let us extend the above code to also print the value of `suffix` field.
+
+```csharp
+    [UnconditionalSuppressMessage("trim", "IL2070", Justification = "DynamicDependency attribute will instruct the linker to keep the public methods on NameProvider.")]
+    public void PrintName([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)] Type type)
+    {
+        string name = (string)type.GetMethod("GetName")!.Invoke(null, null)!;
+        string suffix = (string)type.GetField("suffix")!.GetValue(null)!; // This is will issue a warning (only public methods are guaranteed to be kept).
```
vitek-karas, line 48:
```suggestion
        string suffix = (string)type.GetField("suffix")!.GetValue(null)!; // IL2070 - only public methods are guaranteed to be kept
```

docs/design/redundant-suppressions.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+## Detecting redundant warning suppressions
+
+In order to avoid the above scenario, we would like to have an option to detect and report the warning suppressions which are not tied to any warnings caused by trim-incompatible patterns.
+
+This may be achieved by extending the linker tool functionality to check which suppression do in fact suppress warnings and reporting those which do not. 
+
+The said functionality can be implemented as an optional feature, triggered by passing a `
+--check-suppressions` command line argument to the linker. Running the tool with this option will report all of the warning suppressions which do not suppress any warnings.
```
vitek-karas, line 64: I would remove this - or maybe just say "The way to turn this on is TBD" or something like that.
Mainly because it's actually unlikely we will do it as a command line switch after the discussion.

docs/design/redundant-suppressions.md
```diff
@@ -0,0 +1,93 @@
...
+    {
+        foreach (var p in typeof(T).GetProperties())
+        {
+            PrintPropertyValue(p, instance);
+        }
+    }
+
+```
+
+In order to detect the warning suppression not tied to any trimmer-incompatible pattern we should run the `dotnet publish` command and pass the `--check-suppressions` option.
```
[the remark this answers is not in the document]
vitek-karas: For now we're going with producing a new warning - in which case I agree with @sbomer that using either NoWarn or warning levels would be a great way to do this.

I think we should leave out the exact mechanics our of this document - we'll figure them out later on. It obviously has to be a toggle - and if it's on by default is a discussion for a time when we have much better understanding of the impact/UX of this.

docs/design/redundant-suppressions.md
```diff
@@ -0,0 +1,94 @@
...
+    {
+        string name = (string)type.GetMethod("GetName")!.Invoke(null, null)!;
+        Console.WriteLine(name);
+    }
+```
+
+In order to detect the warning suppression not tied to any trimmer-incompatible pattern we should run the `dotnet publish` command and pass the `--check-suppressions` option.
+```shell
+  dotnet publish -r win-x64 -p:PublishTrimmed=True -p:_ExtraTrimmerArgs="--check-suppressions"
+```
```
vitek-karas, line 82: I would probably remove this - just say, running the linker with the detection enabled will produce a new warning.

docs/design/redundant-suppressions.md
jkurdek: done

docs/design/redundant-suppressions.md
jkurdek: done

vitek-karas (approved):
Nice writeup!

PR #2892, opened 2022-07-13
Analyze nested functions in Roslyn analyzer

sbomer:
This is the Roslyn analyzer counterpart to https://github.com/dotnet/linker/pull/2842.

It adds detection of captured locals, and conservatively analyzes them with all values that it encounters in the method.
The analyzer makes no attempt to warn on reflection access to compiler-generated fields, because the compilation strategy for these is not known by the analyzer.

Unlike the linker, this doesn't do conservative analysis for hoisted locals in _state machine_ methods. The analyzer already understands control flow within state machines the same way it does for normal methods, so we just continue using that behavior here. We could consider improving the linker's analysis to match, but it would require a deeper understanding of the state machine methods.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
+++ b/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
@@ -32,12 +32,7 @@ namespace ILLink.RoslynAnalyzer.TrimAnalysis
 			ControlFlowGraph methodCFG,
 			ImmutableDictionary<CaptureId, FlowCaptureKind> lValueFlowCaptures,
 			InterproceduralState<MultiValue, ValueSetLattice<SingleValue>> interproceduralState)
-		 => new (Lattice,
-				method,
-				methodCFG,
-				lValueFlowCaptures,
-				TrimAnalysisPatterns,
-				interproceduralState);
+		 => new (Lattice, method, methodCFG, lValueFlowCaptures, TrimAnalysisPatterns, interproceduralState);
 
 #if DEBUG
 #pragma warning disable CA1805 // Do not initialize unnecessarily


src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowAnalysis.cs
```diff
@@ -0,0 +1,92 @@
...
+				}
+			}
+		}
+
+		void AnalyzeMethod (MethodBodyValue method, ref InterproceduralState<TValue, TLattice> interproceduralState)
+		{
+			var cfg = method.ControlFlowGraph;
+			var lValueFlowCaptures = LValueFlowCapturesProvider.CreateLValueFlowCaptures (cfg);
+			var visitor = GetVisitor (method.Method, cfg, lValueFlowCaptures, interproceduralState);
+			Fixpoint (new ControlFlowGraphProxy (cfg), Lattice, visitor);
```
vitek-karas, line 79: Not for this PR - in theory we should be able to use the solver similar to Fixpoint to drive the inter-method analysis as well instead of rolling out a private simple version here.

src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
```diff
@@ -10,40 +11,33 @@
...
-			return visitor.TrimAnalysisPatterns;
-		}
+		protected override TrimAnalysisVisitor GetVisitor (
+			IMethodSymbol method,
+			ControlFlowGraph methodCFG,
+			ImmutableDictionary<CaptureId, FlowCaptureKind> lValueFlowCaptures,
+			InterproceduralState<MultiValue, ValueSetLattice<SingleValue>> interproceduralState)
+		 => new (Lattice,
+				method,
+				methodCFG,
```
vitek-karas, line 37: Nit: weird indentation

vitek-karas (approved):
Looks good!

PR #2899, opened 2022-07-18
Map type parameters for static closures

sbomer:
Fixes https://github.com/dotnet/linker/issues/2826

This also fixes an unrelated issue I discovered where in `Release` mode, the compiler can generate struct closure environments that will never get instantiated via a ctor call. I added a testcase that runs the CompilerGeneratedTypes tests in Release mode to cover this.

### DIFF
--- a/src/linker/Linker.Dataflow/CompilerGeneratedState.cs
+++ b/src/linker/Linker.Dataflow/CompilerGeneratedState.cs
@@ -142,12 +142,12 @@ namespace Mono.Linker.Dataflow
 					foreach (var instruction in method.Body.Instructions) {
 						switch (instruction.OpCode.OperandType) {
 						case OperandType.InlineMethod: {
-								MethodDefinition? lambdaOrLocalFunction = _context.TryResolve ((MethodReference) instruction.Operand);
-								if (lambdaOrLocalFunction == null)
+								MethodDefinition? referencedMethod = _context.TryResolve ((MethodReference) instruction.Operand);
+								if (referencedMethod == null)
 									continue;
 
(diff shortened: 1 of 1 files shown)


src/linker/Linker.Dataflow/CompilerGeneratedState.cs
```diff
@@ -327,18 +346,42 @@ void MapGeneratedTypeTypeParameters (TypeDefinition generatedType)
...
+							handled = true;
 						}
 						break;
 					}
+
+					// Also look for type substitutions into generic methods
+					// (such as AsyncTaskMethodBuilder::Start<TStateMachine>).
+					if (!handled && instr.OpCode.OperandType is OperandType.InlineMethod) {
+						var mr = (MethodReference) instr.Operand;
+						if (mr is GenericInstanceMethod gim) {
```
vitek-karas, line 377:
```suggestion
						if (instr.Operand is GenericInstanceMethod gim) {
```

vitek-karas (approved):

agocke:
Sorry, not sure I'm fully following the change. How does it prevent a sample like https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCAeAKgPkQb1kSMUOIWQBZEBZACgEp9TiyBWLXAMwkQF5EGfXABMAppwCGAVwA2AF1qZ6AbmYtk7HIk5g+AxrxHjp8xSrUBfWBaA== from producing a lot of warnings or failing? Since there's two constructor calls to a single type.

agocke (approved):
Ah, it looks like my case always goes through static fields. LGTM

PR #2900, opened 2022-07-19
Fix syntax node scope and nullable warnings

agocke:
The syntax node scope should be the innermost node, otherwise
GetSymbolInfo will often return null. As an example, GetSymbolInfo will
return null for an ArgumentSyntax node, but the underlying expression
will likely resolve into a symbol.

Also adds proper null checks in case any of the providers return null.

### DIFF
--- a/src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
+++ b/src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
@@ -32,7 +32,6 @@ namespace ILLink.CodeFix
 		protected async Task BaseRegisterCodeFixesAsync (CodeFixContext context)
 		{
 			var document = context.Document;
-			var root = await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false);
 			if (await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false) is not { } root)
 				return;
 			var diagnostic = context.Diagnostics.First ();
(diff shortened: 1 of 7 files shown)


src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
```diff
@@ -33,16 +33,22 @@ protected async Task BaseRegisterCodeFixesAsync (CodeFixContext context)
 		{
 			var document = context.Document;
 			var root = await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false);
+			if (root is null)
```
tlakollo, line 36: Nit: what about doing 
```suggestion
			if (await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false) is not { } root)
```
To have the same kind of check as the others

tlakollo (approved):
Looks good to me, thanks!

src/ILLink.CodeFix/BaseAttributeCodeFixProvider.cs
```diff
@@ -33,16 +33,22 @@ protected async Task BaseRegisterCodeFixesAsync (CodeFixContext context)
 		{
 			var document = context.Document;
 			var root = await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false);
```
tlakollo, line 35: Ah sorry my recommendation didn't include line 35, now line 35 is not needed

tlakollo (approved):

PR #2905, opened 2022-07-21
Implements dead branch removal for switch opcode

### LINKED ISSUE #2888: Support removal of unreachable switch blocks
Linker currently supports removal of unreachable if blocks (branches). For example:
```C#
int AlwaysZero { get => 0; }

void Test()
{
    if (AlwaysZero == 0)
    {
        // This will be kept
    }
    else 
    {
        // This entire block will be removed as it can't be reached
    }
}
```

It doesn't support the same for blocks/branches in `switch` instructions. So for example:
```C#
int AlwaysZero { get => 0; }

void Test()
{
    switch (AlwaysZero)
    case 0:
        // This will be kept
    case 1:
    case 2:
        // This could be removed, but currently isn't
}
```

marek-safar:
Fixes #2888


marek-safar:
@vitek-karas ping

test/Mono.Linker.Tests.Cases/UnreachableBlock/ComplexConditionsOptimized.cs
```diff
@@ -37,13 +47,117 @@ public static void Test ()
...
+				"ldloc.0",
+				"ldc.i4.1",
+				"sub",
+				"brtrue il_10",
+				"call System.Void Mono.Linker.Tests.Cases.UnreachableBlock.ComplexConditionsOptimized/TestSwitch::Reached()",
+				"ret",
+				"call System.Void Mono.Linker.Tests.Cases.UnreachableBlock.ComplexConditionsOptimized/TestSwitch::Reached()",
+				"br.s il_a",
+			})]
+			public static void TestOffset ()
```
vitek-karas, line 68: Would it make sense to also add tests to the `ComplexConditions` - basically the debug version of these tests - since the IL patterns will be somewhat different in that case.

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
```diff
@@ -537,6 +537,8 @@ struct BodyReducer
 			// 
 			List<int>? conditionInstrsToRemove;
 
+			List<(int, Instruction)>? conditionInstrsToReplace;
```
vitek-karas, line 540: Add a comment describing what is this for - I know it's mentioned below, but I think it's better to have it in one place.

vitek-karas (approved):

test/Mono.Linker.Tests.Cases/UnreachableBlock/ComplexConditionsOptimized.cs
marek-safar: I did only the most straightforward pattern. There are more including pattern matching, nullable, etc that I point I think we need to invest in so will just leave that for next time.

PR #2907, opened 2022-07-21
Fix re-entrancy when marking nested functions

sbomer:
Fixes https://github.com/dotnet/linker/issues/2903

I tried to actually fix the re-entrancy instead of putting a temporary value in the cache, because this will make it easier to fix the behavior of `CheckRequiresReflectionMethodBodyScanner` to not consider type parameter annotations.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -3467,6 +3467,16 @@ namespace Mono.Linker.Steps
 		// to avoid recursion.
 		bool MarkAndCheckRequiresReflectionMethodBodyScanner (MethodBody body)
 		{
+#if DEBUG
+			if (!Annotations.IsProcessed (body.Method)) {
+				Debug.Assert (CompilerGeneratedState.IsNestedFunctionOrStateMachineMember (body.Method));
+				MethodDefinition owningMethod = body.Method;
+				while (Context.CompilerGeneratedState.TryGetOwningMethodForCompilerGeneratedMember (owningMethod, out var owner))
+					owningMethod = owner;
+				Debug.Assert (owningMethod != body.Method);
+				Debug.Assert (Annotations.IsProcessed (owningMethod));
+			}
+#endif
 			// This may get called multiple times for compiler-generated code: once for
 			// reflection access, and once as part of the interprocedural scan of the user method.
 			// This check ensures that we only do the work and produce warnings once.
@@ -3493,6 +3503,7 @@ namespace Mono.Linker.Steps
 
 			PostMarkMethodBody (body);
 
+			Debug.Assert (requiresReflectionMethodBodyScanner == CheckRequiresReflectionMethodBodyScanner (body));
 			return requiresReflectionMethodBodyScanner;
 		}
 
--- a/test/Mono.Linker.Tests.Cases/DataFlow/CompilerGeneratedCodeAccessedViaReflection.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/CompilerGeneratedCodeAccessedViaReflection.cs
@@ -527,7 +527,7 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 		{
 			static void RequiresAllOnT<[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.All)] T> () { }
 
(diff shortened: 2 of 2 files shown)


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3413,6 +3413,8 @@ protected virtual void MarkMethodBody (MethodBody body)
 				return;
 			}
 
+			// Note: we need to mark compiler-generated method bodies in case they are only accessed via
+			// reflection, in which case we will never do a full data-flow scan of the user-defined method.
```
vitek-karas, line 3417: Took me a while what this comment was talking about.
It sounds like we're marking the CG methods only if they're reflection accessed - which is not true, we mark them always, but for reflection-access-only methods we don't do dataflow.

Maybe:
"Note: we mark the method body of every method here including compiler generated methods whichever way they're accessed. But for compiler generated methods we only do data flow if they're used through their respective user method, we will skip data flow for reflection-access-only compiler generated methods below."

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3423,11 +3425,46 @@ protected virtual void MarkMethodBody (MethodBody body)
 			MarkReflectionLikeDependencies (body, requiresReflectionMethodBodyScanner);
 		}
 
+		bool CheckRequiresReflectionMethodBodyScanner (MethodBody body)
+		{
+			if (ReflectionMethodBodyScanner.RequiresReflectionMethodBodyScannerForMethodBody (Context, body.Method))
```
vitek-karas, line 3430: Could you please add a comment why we don't need a cache for this?
On the face value this is expensive - but I assume we don't care about the cache that much since it's only going to be called for reflection accessed methods (which should be rare).

vitek-karas (approved):
I actually coded the simple fix during the day, but didn't manage to send a PR for it :-). Anyway, I added some more tests - maybe you could merge the tests with this change: https://github.com/vitek-karas/linker/tree/FixSOInCompilerGeneratedCodeMarking

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3423,11 +3428,48 @@ protected virtual void MarkMethodBody (MethodBody body)
...
+					break;
+				}
+			}
+			return false;
+		}
+
+		// Keep the return value of this method in sync with that of CheckRequiresReflectionMethodBodyScanner.
+		// It computes the same value, while also marking as it goes, as an optimization.
+		// This should only be called behind a check to IsProcessed for the method or corresponding user method,
+		// to avoid recursion.
```
vitek-karas, line 3467: Can we assert this?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3423,11 +3428,48 @@ protected virtual void MarkMethodBody (MethodBody body)
...
+					if (ReflectionMethodBodyScanner.RequiresReflectionMethodBodyScannerForCallSite (Context, (MethodReference) instruction.Operand))
+						return true;
+					break;
+				}
+			}
+			return false;
+		}
+
+		// Keep the return value of this method in sync with that of CheckRequiresReflectionMethodBodyScanner.
+		// It computes the same value, while also marking as it goes, as an optimization.
```
vitek-karas, line 3465: Can we assert this? (maybe too expensive even for Debug builds...?)

test/Mono.Linker.Tests.Cases/DataFlow/CompilerGeneratedCodeAccessedViaReflection.cs
```diff
@@ -538,6 +523,116 @@ public static void Test (LocalFunctions test = null)
 			}
 		}
 
+		class SelfMarkingMethods
+		{
+			static void RequiresAllOnT<[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.All)] T> () { }
+
+			static void RequiresNonPublicMethodsMethodsOnT<[DynamicallyAccessedMembers (DynamicallyAccessedMemberTypes.NonPublicMethods)] T> () { }
```
vitek-karas, line 530: Weird name...?

src/linker/Linker.Steps/MarkStep.cs
vitek-karas: Side-effects? I don't really see anything here which should cause side effects, but I could be wrong.

vitek-karas: True - on the other hand if there's a bug in that.. we have a bigger problem :-)

PR #2908, opened 2022-07-21
Avoid dataflow analysis for generic parameters

sbomer:
Fixes https://github.com/dotnet/linker/issues/2906

I'll update this to remove `ExpectedWarning`s from the tests added in https://github.com/dotnet/linker/pull/2907 after that is merged.

### DIFF
--- a/src/linker/Linker.Dataflow/FlowAnnotations.cs
+++ b/src/linker/Linker.Dataflow/FlowAnnotations.cs
@@ -31,13 +31,13 @@ namespace ILLink.Shared.TrimAnalysis
 			GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var methodAnnotations)
 				&& (methodAnnotations.ReturnParameterAnnotation != DynamicallyAccessedMemberTypes.None || methodAnnotations.ParameterAnnotations != null);
 
-		public bool HasAnnotations (MethodDefinition method) =>
+		public bool RequiresVirtualMethodDataFlowAnalysis (MethodDefinition method) =>
 			GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out _);
 
 		public bool RequiresDataFlowAnalysis (FieldDefinition field) =>
 			GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out _);
 
-		public bool HasAnnotations (GenericParameter genericParameter) =>
+		public bool RequiresGenericArgumentDataFlowAnalysis (GenericParameter genericParameter) =>
 			GetGenericParameterAnnotation (genericParameter) != DynamicallyAccessedMemberTypes.None;
 
 		/// <summary>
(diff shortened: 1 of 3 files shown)


src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -28,12 +28,16 @@ public FlowAnnotations (LinkContext context)
 		}
 
 		public bool RequiresDataFlowAnalysis (MethodDefinition method) =>
+			GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var methodAnnotations)
+				&& (methodAnnotations.ReturnParameterAnnotation != DynamicallyAccessedMemberTypes.None || methodAnnotations.ParameterAnnotations != null);
+
+		public bool HasAnnotations (MethodDefinition method) =>
```
vitek-karas, line 34: I would prefer we don't call this `HasAnnotations` - maybe we could make this more specific:
`RequiresDataFlowAnalysisForVirtualMethods` (based on where this is used).

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -28,12 +28,16 @@ public FlowAnnotations (LinkContext context)
...
 			GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out _);
 
 		public bool RequiresDataFlowAnalysis (FieldDefinition field) =>
 			GetAnnotations (field.DeclaringType).TryGetAnnotation (field, out _);
 
-		public bool RequiresDataFlowAnalysis (GenericParameter genericParameter) =>
+		public bool HasAnnotations (GenericParameter genericParameter) =>
```
vitek-karas, line 40: Similarly - "has annotations" is not very telling in the callsite (it has a meaning in this class, but not externally). So maybe `RequiresGenericArgumentDataFlow` ?

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -28,12 +28,16 @@ public FlowAnnotations (LinkContext context)
 		}
 
 		public bool RequiresDataFlowAnalysis (MethodDefinition method) =>
+			GetAnnotations (method.DeclaringType).TryGetAnnotation (method, out var methodAnnotations)
+				&& (methodAnnotations.ReturnParameterAnnotation != DynamicallyAccessedMemberTypes.None || methodAnnotations.ParameterAnnotations != null);
```
vitek-karas, line 32: This will have to work differently in NativeAOT - where we might have to rely on dataflow to handle generic argument dataflow - it's not clear yet though.

vitek-karas (approved):

src/linker/Linker.Dataflow/FlowAnnotations.cs
vitek-karas: I was modeling it after the `GenericArgumentDataFlow`, but if we change that, then I would also change the name of the method respectively. So `RequiresAnalysis` works ... with `GenericArgumentAnalysis`? I guess that would work.
Personally I don't mind calling it `GenericArgumentDataFlow` even though it's not really data-flow. We're already using DataFlow as the term for everything "figure out reflection" related anyway.

PR #2919, opened 2022-07-26
Update to .NET 7.0 TFM

lbussell:
Linker needs a target framework of .NET 7.0 for source build to build without prebuilts. I have tested this locally and it seems to work without any major changes.


marek-safar (approved):

lbussell:
@marek-safar @sbomer, if this change is good then please merge when appropriate.

PR #2925, opened 2022-07-28
Added support for suppressions on properties

jkurdek:
Suppressions can be now placed on properties.

\cc @vitek-karas

### DIFF
--- a/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
+++ b/src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
@@ -78,13 +78,12 @@ namespace Mono.Linker
 					if (IsSuppressedOnElement (id, warningOriginMember, out info))
 						return true;
 
-					if(warningOriginMember is MethodDefinition method) {
-						if(method.TryGetProperty (out var property)) {
+					if (warningOriginMember is MethodDefinition method) {
+						if (method.TryGetProperty (out var property)) {
 							Debug.Assert (property.DeclaringType == warningOriginMember.DeclaringType);
 							warningOriginMember = property;
 							continue;
-						}
-						else if(method.TryGetEvent (out var @event)) {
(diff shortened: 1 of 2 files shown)


vitek-karas:
Could you please do it for events as well?

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -78,7 +78,12 @@ bool IsSuppressed (int id, ICustomAttributeProvider warningOrigin, out SuppressM
 					if (IsSuppressedOnElement (id, warningOriginMember, out info))
 						return true;
 
-					warningOriginMember = warningOriginMember.DeclaringType;
+					if (warningOriginMember is MethodDefinition method && method.TryGetProperty (out var property) && property != null) {
```
tlakollo, line 81: nit: the check for property not being null is unnecessary, the TryGetProperty method will only return a value if true and it won't be null.

test/Mono.Linker.Tests.Cases/Warnings/WarningSuppression/SuppressWarningsInMembersAndTypes.cs
```diff
@@ -128,4 +130,20 @@ public static void Test ()
...
+
+	class SuppressOnProperty
+	{
+		public static void Test ()
+		{
+			var test = Property;
+		}
+
+		[UnconditionalSuppressMessage ("Test", "IL2072")]
+		static int Property {
```
tlakollo, line 142: Can you add a property and an event (once events are supported too) in the SuppressWarningsInType class just to also test that we can go through the path of "I had a method, went to the property/event didn't find a suppression, so I went to the declaring type and I found it there"

vitek-karas (approved):

tlakollo (approved):

PR #2928, opened 2022-07-29
Add an interesting test case for compiler generated code marking

vitek-karas:
See the test for description of what it does. It's basically the linker repro for the case hit in https://github.com/dotnet/runtime/issues/73027#issuecomment-1199199943


vitek-karas:
Right - the behavior currently seems OK to me - we don't produce warning in either case. Basically even though the entire assembly is preserved, we treat the compiler generated code as "trimmable" anyway (we would still keep it, but we won't produce diagnostics from it). I think that's OK, less noise and I don't see any holes (we don't see any reference to it anywhere).

The internal behavior is obviously weird - it would be great if we could make it to be consistent regardless of ordering. But I could not find any observable differences yet.

PR #2929, opened 2022-07-29
Make serialization discovery opt-in

sbomer:
Fixes https://github.com/dotnet/linker/issues/2924

Serialization discovery is an undocumented quirk that exists for back-compat in xamarin scenarios. We turned it on by default in .NET 6, but would like to remove such quirks from the default options in .NET 7.

This changes the (undocumented) command-line option to be opt-in. Xamarin SDKs will be expected to pass this option by setting `<_ExtraTrimmerArgs>--enable-serialization-discovery</_ExtraTrimmerArgs>`.


vitek-karas (approved):
Assuming we will go and set this in Xamarin repos ASAP, to make sure we don't break them...

PR #2930, opened 2022-07-29
Include all warnings under suppression setting

sbomer:
`SuppressTrimAnalysisWarnings` now includes all warning codes produced by
the linker or trim analyzer. Previously, only warnings in the
"Trim analysis" category (see MessageSubCategory.cs) were suppressed by
this setting.

This makes the logic more future-proof so we don't need to keep adding
new warning codes to the list.

This does change the behavior a bit. It includes warnings that weren't previously
part of the list. Additionally, previously if you set:
```xml
<EnableTrimAnalyzer>true</EnableTrimAnalyzer>
<SuppressTrimAnalysisWarnings>true</SuppressTrimAnalysisWarnings>
```
you wouldn't see trim analysis warnings from the analyzer. With this change, you will.
Similarly, previously if you set
```xml
<ILLinkWarningLevel>5</ILLinkWarningLevel>
<SuppressTrimAnalysisWarnings>true</SuppressTrimAnalysisWarnings>
```
you wouldn't see trim analysis warnings, but after this change you will.

The SDK defaults are imported after the ILLink targets so I think no SDK
changes are needed:
https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.Analyzers.targets#L80-L81


src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
```diff
@@ -59,38 +59,8 @@ Copyright (c) .NET Foundation. All rights reserved.
...
-    <NoWarn>$(NoWarn);IL2107;IL2117</NoWarn>
-    <!-- Unannotated type derived from base type with RequiresUnreferencedCode -->
-    <NoWarn>$(NoWarn);IL2109</NoWarn>
-    <!-- Reflection access to members with DynamicallyAccessedMembers requirements -->
-    <NoWarn>$(NoWarn);IL2110;IL2111;IL2114;IL2115</NoWarn>
-    <!-- Reflection access to members with RequiresUnreferencedCode -->
-    <NoWarn>$(NoWarn);IL2112;IL2113</NoWarn>
-    <!-- Reflection access to compiler-generated code -->
-    <NoWarn>$(NoWarn);IL2118;IL2119;IL2120</NoWarn>
+    <ILLinkWarningLevel Condition="'$(ILLinkWarningLevel)' == ''">0</ILLinkWarningLevel>
```
agocke, line 62: Maybe warn or error instead of not setting it? I think these settings are in conflict and we should probably signal there's something wrong.

Also, I think SuppressTrimAnalysisWarnings should take precedence.

agocke: This might be different from EnableTrimAnalyzer, as I could see wanting to just disable the linker.

agocke: Hmm. That makes sense. Maybe warn for both?

vitek-karas: Do we have some mechanism where `WarningLevel` influences `ILLinkWarningLevel`, or are those completely separate?
(Honestly I thought we didn't have `ILLinkWarningLevel` at all - since we only have `NoWarn` and there's no `ILLinkNoWarn`).

vitek-karas: > What should be the preferred way to enable analyzer warnings but not linker warnings?

I don't think this needs to be "easy" - should be rare to want this
But I guess the `SuppressTrimAnalysisWarnings`/`EnableTrimAnalyzer` makes more sense in this case.

I think I came around to what the code does in this PR as-is. It's not perfect, but at least it makes logical sense :-)

vitek-karas (approved):

PR #2931, opened 2022-07-29
Add another local function branch removal test

sbomer:
This is like the existing test, but additionally shows a visible
difference in the warning behavior. When the warning is produced for
generic arguments rather than dataflow, it will show up even if we don't
do dataflow for the local function whose callsite was removed.

When branch removal breaks the link between the local function and
its user method, this can result in warning suppressions that don't
take effect.

We could use the desired warning behavior to decide what the fix
needs to do (whether it should do compiler-generated owning method
detection before vs after constant propagation for the method).

### DIFF
--- a/test/Mono.Linker.Tests.Cases/DataFlow/CompilerGeneratedCodeInPreservedAssemblyWithWarning.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/CompilerGeneratedCodeInPreservedAssemblyWithWarning.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


test/Mono.Linker.Tests.Cases/DataFlow/CompilerGeneratedCodeInPreservedAssemblyWithWarning.cs
```diff
@@ -0,0 +1,71 @@
...
+			// Generic argument warnings from the local function will not be suppressed
+			// by RUC on the user method.
+			[RequiresUnreferencedCode ("--" + nameof (Inner) + "." + nameof (WithLocalFunctionInner) + "--")]
+			public static void WithLocalFunctionInner ()
+			{
+				if (AlwaysFalse) {
+					LocalWithWarning<int> ();
+				}
+
+				[ExpectedWarning ("IL2091", ProducedBy = ProducedBy.Trimmer)]
```
vitek-karas, line 42: We should probably create an issue for this and link it here.

BTW: I don't think it's important to fix this... bust still.

vitek-karas (approved):

PR #2933, opened 2022-07-30
Minor refactoring to DAMCodeFixProvider

Youssef1313:
- `GetAttributeArguments` will always return the same attribute argument. So I'm returning unconditionally.
- Moved to symbol comparison instead of string comparison when comparing attribute.
- Moved `GetAttributeArguments` into `AddAttributeAsync`

FYI @iserrato


iserrato:
Adapted argument generation move to `AddAttributeAsync` into most recent version of DAM Code Fix.
https://github.com/dotnet/linker/pull/2985

iserrato (commented):
thanks for working on this! we changed the way attributes are generated, but moving the actually creation of the attribute to AddAttributeAsync is a good catch.

PR #2940, opened 2022-08-02
Redundant suppressions on events and properties

jkurdek:
Redundant suppressions detection for suppressions introduced in #2925.
\cc @vitek-karas, @tlakollo


vitek-karas (approved):
Thanks!

PR #2942, opened 2022-08-03
New constraint on a generic argument involved in MakeGenericType/Method should warn

vitek-karas:
For correctness linker must be able to validate that generic parameters with new constraint will have their default .ctor preserved. If that can't be guaranteed, it needs to warn.

Added tests for cases around this.

`Nullable<>` is the only exception to this rule, so changed the code to completely skip any validation for it.

Fixes https://github.com/dotnet/linker/issues/2934

/cc @jkurdek


jtschuster (commented):
LGTM, Thank you!

vitek-karas:
@jtschuster could you please approve the request (it's currently disallowed to approve my own requests :-()

jtschuster (approved):

PR #2945, opened 2022-08-04
Update MSBuild version to fix official build

sbomer:
I think this will fix the official build - see https://github.com/dotnet/arcade/issues/10280 for context.

### DIFF
--- a/global.json
+++ b/global.json
@@ -5,7 +5,8 @@
     "rollForward": "major"
   },
   "tools": {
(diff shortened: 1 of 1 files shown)


MattGal (approved):
LGTM, I already tried the same thing in internal https://dev.azure.com/dnceng/internal/_build/results?buildId=1923531&view=logs&j=b11b921d-8982-5bb3-754b-b114d42fd804

PR #2952, opened 2022-08-08
Test that basic ref field operations don't crash linker

jtschuster:
Adds some tests with basic usage of ref fields to ensure the linker doesn't crash or produce incorrect warnings. Ref structs can't be used as type parameters, so some things like IEnumerable<RS> can't be tested.

Please let me know if I'm missing other cases where this could cause some issues.

### DIFF
--- a/test/Mono.Linker.Tests.Cases/Basic/LinkerHandlesRefFields.cs
+++ b/test/Mono.Linker.Tests.Cases/Basic/LinkerHandlesRefFields.cs
@@ -1,0 +1,119 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+using Mono.Linker.Tests.Cases.Expectations.Assertions;
+
+namespace Mono.Linker.Tests.Cases.Basic
+{
+	[ExpectedNoWarnings]
+	[SkipKeptItemsValidation]
+	class LinkerHandlesRefFields
+	{
+		ref struct RS
+		{
+			public ref int i;
+			public ref double d;
+			public RS2 rs2;
+			public ref string str;
+			public ref object obj;
+			public ref Type t;
+			public void MoveThis()
+			{
+				this = new RS ();
+			}
+		}
+
+		ref struct RS2
+		{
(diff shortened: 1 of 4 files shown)


test/Mono.Linker.Tests.Cases/Basic/LinkerHandlesRefFields.cs
```diff
@@ -0,0 +1,119 @@
...
+			{
+				this = new RS2 ();
+			}
+		}
+
+		delegate RS2 RsParamRs2Return (RS rs);
+		delegate ref RS2 RefRSParamRefRs2Return (ref RS rs);
+		delegate ref int RsParamRefIntReturn (RS rs);
+
+		public static void Main (string[] args)
```
vitek-karas, line 50: I actually don't know if some of these are valid, but still:
- ref struct in a generic parameter
- ref return and parameter on a virtual method?
- Put annotations (DAM/RUC on type) on the ref field - don't care if it works correctly, but should not crash

jtschuster: > ref struct in a generic parameter
Ref structs will generate [CS0306](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0306) if used as a type parameter.

> ref return and parameter on a virtual method?
Added IRS interface and Base and Derived classes with virtual methods with ref structs as return, ref return, parameter, and ref parameter.

> Put annotations (DAM/RUC on type) on the ref field
RUC can't be placed on structs and ref structs can't implement interfaces, so I don't think there's a way that they could be annotated with DAM. I've added DAM on a field and DAM and RUC on a method return and parameter.

vitek-karas: Thanks a lot!

test/Mono.Linker.Tests.Cases/Basic/LinkerHandlesRefFields.cs
```diff
@@ -9,6 +9,7 @@
 using System.Text;
 using System.Threading.Tasks;
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
+using Mono.Linker.Tests.Cases.Expectations.Helpers;
 
 namespace Mono.Linker.Tests.Cases.Basic
 {
```
vitek-karas, line 15: Nit: Add a comment describing what the test is trying to validate - mainly that it's only meant to validate that linker doesn't crash or do other weird things on ref fields. And that it's not meant as full validation of all functionality around ref fields.

vitek-karas (approved):
Thanks!

PR #2954, opened 2022-08-08
Build with preview5 SDK

sbomer:
https://github.com/dotnet/linker/pull/2948 didn't fix the dependency flow, because the packages for some reason are still being produced with a preview7 in the runtimeconfig. Lacking better workarounds, this change builds with the preview5 SDK that was selected in https://github.com/dotnet/linker/pull/2884 before we decided to update to preview7 (due to a dotnet-format bug). This disables the lint job to prevent those failures. We should turn it back on once we can use preview7 (once dotnet/runtime and dotnet/sdk are on preview7).

### DIFF
--- a/eng/azure-pipelines.yml
+++ b/eng/azure-pipelines.yml
@@ -138,16 +138,16 @@ stages:
               MSBUILDENSURESTDOUTFORTASKPROCESSES: 1
             displayName: Build illink.sln $(_BuildConfig)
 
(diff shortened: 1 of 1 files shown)


vitek-karas (approved):
Does lint work locally? I assume it doesn't ...

akoeplinger:
I don't have permission.

PR #2957, opened 2022-08-09
Use warning levels

sbomer:
This moves warning codes added since .NET 6 into a new warning wave for 7. It removes the compiler-generated code warnings that are appearing when trimming a 6.0 app with the 7.0 linker - but doesn't fix the IL2070 and IL2067 warnings (existing warning codes, which are now showing up due to improvements in how we understand byrefs).

### DIFF
--- a/src/linker/Linker/LinkContext.cs
+++ b/src/linker/Linker/LinkContext.cs
@@ -566,7 +566,7 @@ namespace Mono.Linker
 		/// <returns>New MessageContainer of 'Warning' category</returns>
 		public void LogWarning (string text, int code, MessageOrigin origin, string subcategory = MessageSubCategory.None)
 		{
-			WarnVersion version = GetWarningVersion ();
+			WarnVersion version = GetWarningVersion (code);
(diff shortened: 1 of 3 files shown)


jtschuster (approved):
Small nit: it might be a good idea to add a comment that the CanSetWarningVersion tests are meant to be identical and should stay in sync if one is changed. Otherwise LGTM, thank you!

src/linker/Linker/LinkContext.cs
```diff
@@ -730,9 +730,11 @@ public bool IsSingleWarn (string assemblyName)
 			return SingleWarn.TryGetValue (assemblyName, out value) && value;
 		}
 
-		static WarnVersion GetWarningVersion ()
+		static WarnVersion GetWarningVersion (int code)
 		{
 			// This should return an increasing WarnVersion for new warning waves.
+			if (code > 2116)
```
jtschuster, line 736: Should this also include a `&& code < 3000` for single file and native aot?

PR #2960, opened 2022-08-10
Add IL Verification to tests

jtschuster:
Adds an `ILVerifier` to check that the IL produced by the linker is valid. Unsafe C# produced unverifiable code, so we skip verification when we pass that flag to the compiler. Also, there are a few warnings that are produced by valid C# with new features like static abstract interface methods and ref fields and ref returns.

In the future, it may be nice to add better error messages with the type, method name, and IL offset that produced the error, and perhaps an `[ExpectedILVerifyError]` attribute instead of filtering all of a type of error, but those are non-trivial to implement and don't occur in many tests (<10), so I haven't done that yet.


jtschuster:
This is the final piece of ref field validation left and will close https://github.com/dotnet/linker/issues/2519.

vitek-karas (approved):
Looks good.

PR #2961, opened 2022-08-10
Add tests for MakeGenericMethod for static interface methods

jtschuster:
Part of https://github.com/dotnet/linker/issues/2058


jtschuster:
Was using an old branch and had an ugly list of commits.

PR #2965, opened 2022-08-11
IL2121 warnings point to XML files correctly

### LINKED ISSUE #2923: Redundant suppressions from XML files are reported in wrong place
If we have a redundant warning suppression in XML file, the IL2121 warning will be reported on the target of the suppression. The warning should be reported on XML file, where the redundant suppression is located.

```
<method name="Test"> // Should be the warning origin for IL2121 warning.
	<attribute fullname="System.Diagnostics.CodeAnalysis.UnconditionalSuppressMessageAttribute">
		<argument>ILLink</argument>
		<argument>IL2026</argument>
	</attribute>
</method>
```

```csharp
public static void Test () //Warning origin for IL2121 warning.
{
	DoNotTriggerWarning ();
}
```

jkurdek:
Fixes #2923. Redundant suppressions from XML will be reported in XML, not target.

\cc @vitek-karas

### DIFF
--- a/src/linker/Linker.Steps/LinkAttributesParser.cs
+++ b/src/linker/Linker.Steps/LinkAttributesParser.cs
@@ -38,9 +38,10 @@ namespace Mono.Linker.Steps
 
 		static bool IsRemoveAttributeInstances (string attributeName) => attributeName == "RemoveAttributeInstances" || attributeName == "RemoveAttributeInstancesAttribute";
 
-		CustomAttribute[]? ProcessAttributes (XPathNavigator nav, ICustomAttributeProvider provider)
+		(CustomAttribute[]? customAttributes, MessageOrigin[]? origins) ProcessAttributes (XPathNavigator nav, ICustomAttributeProvider provider)
 		{
-			var builder = new ArrayBuilder<CustomAttribute> ();
+			var customAttributesBuilder = new ArrayBuilder<CustomAttribute> ();
+			var originsBuilder = new ArrayBuilder<MessageOrigin> ();
 			foreach (XPathNavigator attributeNav in nav.SelectChildren ("attribute", string.Empty)) {
 				if (!ShouldProcessElement (attributeNav))
 					continue;
@@ -74,11 +75,12 @@ namespace Mono.Linker.Steps
 				CustomAttribute? customAttribute = CreateCustomAttribute (attributeNav, attributeType);
 				if (customAttribute != null) {
 					_context.LogMessage ($"Assigning external custom attribute '{FormatCustomAttribute (customAttribute)}' instance to '{provider}'.");
-					builder.Add (customAttribute);
+					customAttributesBuilder.Add (customAttribute);
+					originsBuilder.Add (GetMessageOriginForPosition (attributeNav));
 				}
 			}
 
-			return builder.ToArray ();
+			return (customAttributesBuilder.ToArray (), originsBuilder.ToArray ());
 
 			static string FormatCustomAttribute (CustomAttribute ca)
 			{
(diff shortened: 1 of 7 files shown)


jkurdek:
@tlakollo could you please take a look?

tlakollo (approved):
I have to say that the structure feels weird, now that we are doing things like `suppressionInfo.suppression.SuppressMessageInfo.Id` feels like we need to flatten something, but at the same time couldn't come up with a better approach. Maybe get rid of suppressMessageInfo (?) and have all the fields being part of a suppression (the use flag and the origin at least). But at the same time, I know that the SuppressMessageInfo structure is made by fields in the UncondicionalSuppressMessageAttribute. Either way, the change looks fine

tlakollo (approved):
Thanks! :D

jkurdek:
@tlakollo could you merge it if looks good? (I don't have the write access 😄)

src/linker/Linker.Steps/LinkAttributesParser.cs
```diff
@@ -38,9 +38,10 @@ public void Parse (AttributeInfo xmlInfo)
 
 		static bool IsRemoveAttributeInstances (string attributeName) => attributeName == "RemoveAttributeInstances" || attributeName == "RemoveAttributeInstancesAttribute";
 
-		CustomAttribute[]? ProcessAttributes (XPathNavigator nav, ICustomAttributeProvider provider)
+		(CustomAttribute[]? customAttributes, MessageOrigin[]? origins) ProcessAttributes (XPathNavigator nav, ICustomAttributeProvider provider)
```
vitek-karas, line 41: I would prefer to do this the other way round:
```C#
(CustomAttribute, MessageOrigin)[]? ProcessAttributes(...)
```

Return a single array where each item is the attribute and its origin.

vitek-karas: This would also simplify the code which uses this.

src/linker/Linker/UnconditionalSuppressMessageAttributeState.cs
```diff
@@ -18,8 +18,18 @@ public class UnconditionalSuppressMessageAttributeState
 
 		public class Suppression
 		{
-			public SuppressMessageInfo SuppressMessageInfo { get; set; }
+			public SuppressMessageInfo SuppressMessageInfo { get; }
 			public bool Used { get; set; }
+			public CustomAttribute OriginAttribute { get; }
+			public ICustomAttributeProvider Provider { get; }
+
+			public Suppression (SuppressMessageInfo suppressMessageInfo, bool used, CustomAttribute originAttribute, ICustomAttributeProvider provider)
```
vitek-karas, line 26: The used is always set to false upon construction. You could remove the parameter from the .ctor and simply always initialize it to `false`.

vitek-karas (commented):
Some style comments, the functionality looks good.

One other comment:
* Please include roughly what the change does in the description of the PR and commit. Saying  just "fix this bug", forces people to look inside the code change when reviewing history.
* It also helps understand the change better - The description should include "why it was done this way"

src/linker/Linker.Steps/LinkAttributesParser.cs
jkurdek: Currently [AddCustomAttributes](https://github.com/dotnet/linker/blob/main/src/linker/Linker/AttributeInfo.cs#L35) accepts `CustomAttribute[]` and places it directly into `CustomAttributes` store. Should we return a single array, we would have to either split it into two to get `CustomAttribute[]` or store `(CustomAttribute, MessageOrigin)[]` instead of `CustomAttribute[]` .

PR #2966, opened 2022-08-11
Fix il corruption

sbomer:
Thanks for sharing the fix @vitek-karas. I cleaned it up a little to be more specific (existing tests which ended in removed ret instructions - for example in a removed else branch - were failing), and changed the test to check the instruction sequence.

This is intended to fix https://github.com/dotnet/runtime/issues/73474. I had trouble creating a repro of the original issue encountered on runtime bits, but the new testcase exhibits what looks like the same issue.

To me it would feel like a cleaner fix to rewrite the always-taken conditional branches to be unconditional branches, but maybe there is some reason we don't do so today. The current fix has lower risk since it is so specifically targeted.

### DIFF
--- a/src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
+++ b/src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
@@ -807,16 +807,7 @@ namespace Mono.Linker.Steps
 			}
 
 			static bool IsConditionalBranch (OpCode opCode)
-			{
-				switch (opCode.Code) {
-				case Code.Brfalse:
-				case Code.Brfalse_S:
-				case Code.Brtrue:
-				case Code.Brtrue_S:
-					return true;
-				}
-				return false;
-			}
+			=> opCode.Code is Code.Brfalse or Code.Brfalse_S or Code.Brtrue or Code.Brtrue_S;
 
 			void RemoveUnreachableInstructions (BitArray reachable)
 			{


src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
```diff
@@ -806,6 +806,18 @@ public bool ApplyTemporaryInlining (in UnreachableBlocksOptimizer optimizer)
...
+			{
+				switch (opCode.Code) {
+				case Code.Brfalse:
+				case Code.Brfalse_S:
+				case Code.Brtrue:
+				case Code.Brtrue_S:
+					return true;
+				}
+				return false;
+			}
```
adamsitnik, line 819: nit: you could use new C# syntax and make it one liner

```suggestion
			=> opCode.Code is Code.Brfalse or Code.Brfalse_S or Code.Brtrue or Code.Brtrue_S;
```

vitek-karas (approved):
The code is currently not written to handle replacing instruction sequences of various length for conditions - in this case we would have to remove the `ldc.i4.1` and `brtrue.s` to replace it with a simple `br.s`. Obviously doable, but would require much more code churn.

https://github.com/dotnet/linker/pull/2905 runs into a similar limitation. I haven't look in detail if we could use the same approach for both though. In any case I think this targeted fix is OK for 7.

PR #2972, opened 2022-08-12
Fix exception when generating warning for method definition

vitek-karas:
If we produce a warning for the method itself (no IL offset) and the method has debug symbols, it can happen that there's no sequence point for the first instruction. In that case the current code will crash because we expect to always find sequence point for offset 0.

Fix this by looking for the first sequence point instead.

Fixes https://github.com/dotnet/linker/issues/2971


vitek-karas:
Two additional notes about this:
1. We basically lack tests which would have symbols for everything and also produce warnings. I didn't try to solve this in this change.
2. This doesn't repro when linker is launched from VS. For some reason in that case the symbols are read differently and Cecil reports as method not having any sequence points. The exact same thing executed outside of VS works.

PR #2973, opened 2022-08-12
Suppress warning in instance fields 

tlakollo:
Suppress warning in instance fields if the instance constructors are annotated with Requires

Fixes https://github.com/dotnet/linker/issues/2970
Related to https://github.com/dotnet/runtime/pull/73853


jtschuster (approved):
LGTM

PR #2977, opened 2022-08-17
Add test for dotnet/runtime#73048

MichalStrehovsky:
Running dataflow analysis on the added test takes about 20 seconds. Double the number of `yield returns` and it will take several minutes.

In the runtime repo we're running into this on https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime.Serialization.Formatters/tests/BinaryFormatterTestData.cs (method `SerializableObjects()`). That one has even more `yield returns` and I lost patience waiting for it to finish.


MichalStrehovsky:
@sbomer is there something we can do to make this finish in more reasonable time?

PR #2979, opened 2022-08-17
Don't track current field of state machines

sbomer:
Fixes the performance issue tested in https://github.com/dotnet/linker/pull/2977.
The problem was that we were tracking all possible returned values throughout the method, since we were treating them the same as hoisted locals. I'm sure it's still possible to come up with testcases that show bad performance (for example assigning many arrays to a hoisted local in the same iterator), but this should at least fix the immediate issue encountered in runtime.


jtschuster (approved):
Thank you!

MichalStrehovsky:
I tried this on my repro case and now the compilation finishes in reasonable time. Thanks!

PR #2980, opened 2022-08-17
Add support for generic attributes

sbomer:
Depends on https://github.com/mono/cecil/pull/36.
Fixes https://github.com/dotnet/linker/issues/2963.


vitek-karas (approved):
Looks good - maybe create an issue to add tests which do some simple validation of data flow stuff on the generic instantiations of attributes.
I'm not worried about linker, but analyzer and AoT might have problems.

PR #2983, opened 2022-08-18
Check for value in LocalVariableStore and only validate on changes to locals

jtschuster:
With changes in behavior with warn=5, ValidateNoReferenceToReference would sometimes fail with a local variable value not being in the LocalVariableStore. This uses a TryGetValue instead of a direct access.

Also, instead of validating the local variable state on every instruction, we can limit it to places where `locals` is modified or passed outside of the function.

### DIFF
--- a/src/linker/Linker.Dataflow/MethodBodyScanner.cs
+++ b/src/linker/Linker.Dataflow/MethodBodyScanner.cs
@@ -187,7 +187,8 @@ namespace Mono.Linker.Dataflow
 				VariableDefinition localVariable = keyValuePair.Key;
 				foreach (var val in localValue) {
 					if (val is LocalVariableReferenceValue reference
-						&& locals[reference.LocalDefinition].Value.Any (v => v is ReferenceValue)) {
+					&& locals.TryGetValue (reference.LocalDefinition, out var referencedValue)
+					&& referencedValue.Value.Any (v => v is ReferenceValue)) {
 						throw new LinkerFatalErrorException (MessageContainer.CreateCustomErrorMessage (
 								$"In method {method.FullName}, local variable {localVariable.Index} references variable {reference.LocalDefinition.Index} which is a reference.",
 								(int) DiagnosticId.LinkerUnexpectedError,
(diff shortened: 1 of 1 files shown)


src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -187,7 +187,8 @@ static void ValidateNoReferenceToReference (LocalVariableStore locals, MethodDef
 				VariableDefinition localVariable = keyValuePair.Key;
 				foreach (var val in localValue) {
 					if (val is LocalVariableReferenceValue reference
-						&& locals[reference.LocalDefinition].Value.Any (v => v is ReferenceValue)) {
+					&& locals.TryGetValue (reference.LocalDefinition, out var referencedValue)
```
tlakollo, line 190: nit: indentation is off

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -187,7 +187,8 @@ static void ValidateNoReferenceToReference (LocalVariableStore locals, MethodDef
...
+					&& referencedValue.Value.Any (v => v is ReferenceValue)) {
```
tlakollo, line 191: So basically this new logic says continue, in the scenario that was failing:
V_9 was trying to reference V_2 which was not present in the locals at that moment. What happens if we continue and then V_2 was a ReferenceValue, we are long gone from the check of V_9, is that even possible?

tlakollo (approved):
Nice! thanks

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -186,12 +186,12 @@ static void ValidateNoReferenceToReference (LocalVariableStore locals, MethodDef
 				MultiValue localValue = keyValuePair.Value.Value;
 				VariableDefinition localVariable = keyValuePair.Key;
 				foreach (var val in localValue) {
-					if (val is LocalVariableReferenceValue reference
-						&& locals[reference.LocalDefinition].Value.Any (v => v is ReferenceValue)) {
-						throw new LinkerFatalErrorException (MessageContainer.CreateCustomErrorMessage (
-								$"In method {method.FullName}, local variable {localVariable.Index} references variable {reference.LocalDefinition.Index} which is a reference.",
-								(int) DiagnosticId.LinkerUnexpectedError,
-								origin: new MessageOrigin (method, ilOffset)));
+					if (val is LocalVariableReferenceValue localReference && localReference.ReferencedType.IsByReference) {
```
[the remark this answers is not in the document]
jtschuster: An `int*` will have `IsPointer` set, but not `IsByReference`, so I think it should be okay.

PR #2993, opened 2022-08-22
Use enum for method parameter indexing

jtschuster:
I've run into multiple issues when I think an `int parameterIndex` is the source-based index when it is the IL-based index (offset by 1 for `this`) or vice versa.

In order to fix this, we can use different enums for each definition of "Parameter Index" to enforce the correct usage and avoid the off by one errors. The enums rely on the underlying `int` value to provide the index number. 

This change adds `ILParameterIndex` to represent the IL-based indexing where 0 may refer to `this, and SourceParameterIndex where 0 is the first "real" parameter in C# source code. Converting from ILParameterIndex to SourceParameterIndex returns a SourceParameterKind to differentiate between `this` and other parameters and ensure that any `SourceParameterIndex` will be able to index `MethodReference.Parameters`.

 It uses these in `ParameterReferenceValue` as well as `GetMethodParameterValue()` and adds a `GetMethodThisParameterValue()`, `ThisParameterReferenceValue`, and bubbles the changes where necessary.

Most places should use SourceParameterIndex, and convert to ILParameterIndex when necessary or at the boundary of updated and old code.

I'll continue to update the code to use these enums in other PRs.


src/linker/Linker/ParameterHelpers.cs
```diff
@@ -0,0 +1,87 @@
...
+				return (ILParameterIndex) paramNum;
+			}
+			ILParameterIndex GetStargParamIndex ()
+			{
+				ParameterDefinition param = (ParameterDefinition) operation.Operand;
+				return (ILParameterIndex) param.Sequence;
+			}
+		}
+
+		public static SourceParameterIndex GetSourceParameter (MethodDefinition method, Instruction operation)
```
vitek-karas, line 61:
```suggestion
		public static SourceParameterIndex GetSourceParameterIndex (MethodDefinition method, Instruction operation)
```

src/linker/Linker/ILParameterIndex.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Mono.Linker
+{
+	/// <summary>
+	/// Represents the index of arguments passed to a function in IL. Generally prefer to use <see cref="ILLink.Shared.SourceParameterIndex"/> when possible. See also <seealso cref="Mono.Linker.ParameterHelpers"/>.
+	/// </summary>
+	///
```
vitek-karas, line 9: Please add a comment describing that you're using enums as a way to separate source/IL indexes in the C# type system (took me a while to figure that out), and that they're actually used as numbers (relying on the fact that enum is really just an Int32).
Would also be nice to describe the numbering scheme here (IL parameter indexes are >=0 and for "instance" call the `0` is `this` and others are +1 .... and then something similar for source indexes).

src/ILLink.Shared/SourceParameterIndex.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace ILLink.Shared
+{
+	/// <summary>
+	/// Used to indicate the index of the parameter in source code (i.e. the same regardless of if there is a `this` or not)
+	/// </summary>
```
vitek-karas, line 8: Same comment as for ILParameterIndex

src/ILLink.Shared/SourceParameterIndex.cs
```diff
@@ -0,0 +1,13 @@
...
+	public enum SourceParameterIndex
```
vitek-karas, line 9: I find the naming a bit weird - it's not exactly clear what "source" means here. That said - I didn't come up with a better name really. So I guess we should just add a good comment (maybe add couple of examples).

vitek-karas (approved):
I didn't look at the details, but I like the approach of using C# type system to differentiate the two cases - really nice idea!

src/linker/Linker/TypeMapInfo.cs
```diff
@@ -347,8 +347,8 @@ bool MethodMatch (MethodReference candidate, MethodReference method)
 				return false;
 
 			for (int i = 0; i < cp.Count; i++) {
-				if (candidate.GetParameterType (i, context) is not TypeReference candidateParameterType ||
-					method.GetParameterType (i, context) is not TypeReference methodParameterType ||
+				if (candidate.GetInflatedParameterType (i, context) is not TypeReference candidateParameterType ||
+					method.GetInflatedParameterType (i, context) is not TypeReference methodParameterType ||
```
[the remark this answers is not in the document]
jtschuster: This wasn't a fix, I just renamed the function.

jtschuster:
Sorry for the force push, I ran into issues with rebasing and it seemed like the cleanest option. I've updated the description to reflect the changes.

src/linker/Linker/ParameterHelpers.cs
```diff
@@ -0,0 +1,79 @@
...
+			return SourceParameterKind.Numbered;
+		}
+
+		public static ILParameterIndex ILParameterIndexFromSourceParameterIndex (MethodReference method, SourceParameterIndex sourceIndex)
+		{
+			if (method.HasImplicitThis ())
+				return (ILParameterIndex) ((int)sourceIndex + 1);
+
+			return (ILParameterIndex) (int)sourceIndex;
+		}
```
tlakollo, line 77: Nit: maybe we can simplify it to
```suggestion
		public static ILParameterIndex ILParameterIndexFromSourceParameterIndex (MethodReference method, SourceParameterIndex sourceIndex) => 
			method.HasImplicitThis () ? (ILParameterIndex) ((int)sourceIndex + 1) : (ILParameterIndex) (int)sourceIndex;
```

src/linker/Linker/ILParameterIndex.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Mono.Linker
+{
+	/// <summary>
+	/// Represents the index of arguments passed to a function in IL (i.e. (ILParameterIndexs)0 represents `this` for non-static methods.
+	/// This is used to enforce a differentiatiation between scenarios where the 0 index should be `this` and when the 0 index should be the first non-this parameter in the type system.
+	/// There are no named enum values, the underlying integer value represents the index value.
```
tlakollo, line 9:
```suggestion
	/// Represents the index of arguments passed to a function in IL (i.e. (ILParameterIndex)0 represents `this` for non-static methods.
	/// This is used to enforce a differentiation between scenarios where the 0 index should be `this` and when the 0 index should be the first non-this parameter in the type system.
	/// There are no named enum values, the underlying integer value represents the index value.
```

src/linker/BannedSymbols.txt
```diff
@@ -1,2 +1,4 @@
 T:Mono.Cecil.Cil.ILProcessor;Use LinkerILProcessor instead
 M:Mono.Cecil.TypeReference.Resolve();Use LinkContext.Resolve and LinkContext.TryResolve helpers instead
+P:Mono.Collections.Generic.Collection`1{Mono.Cecil.ParameterDefinition}.Item(System.Int32); use x
+P:Mono.Cecil.ParameterDefinitionCollection.Item(System.Int32); use x
```
tlakollo, line 4: What's this BannedSymbols file?

src/ILLink.Shared/SourceParameterIndex.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace ILLink.Shared
+{
+	/// <summary>
+	/// Used to indicate the index of the parameter in source code (i.e. the index is not offset by 1 if there is a `this` parameter)
+	/// This enum and <see cref="T:Mono.Linker.ILParameterIndex"/> is used to enforce a differentiatiation between scenarios where the 0 index should be `this` and when the 0 index should be the first non-this parameter in the type system.
+	/// `this` is the only named enum value. For all others, the underlying integer value represents the index value.
```
tlakollo, line 9:
```suggestion
	/// This enum and <see cref="T:Mono.Linker.ILParameterIndex"/> is used to enforce a differentiation between scenarios where the 0 index should be `this` and when the 0 index should be the first non-this parameter in the type system.
	/// `this` is the only named enum value. For all others, the underlying integer value represents the index value.
```

src/ILLink.Shared/SourceParameterIndex.cs
```diff
@@ -0,0 +1,21 @@
...
+	/// </summary>
+	/// <example>
+	/// 0 is the first argument passed to the function in C#,
+	/// 1 is the next, and so on. These can be used to index parameters.
+	/// -1 is a flag to indicate the `this` parameter, and when SourceParamater
```
tlakollo, line 14: Nit: this comment ends abruptly in my view
```suggestion
	/// -1 is a flag to indicate the `this` parameter, and when SourceParameter
```

tlakollo (approved):
Only nit stuff from my review

src/linker/BannedSymbols.txt
jtschuster: Ah, I was looking into using the BannedApiAnalyzer to prevent indexing `MethodReference.Parameters`, but I couldn't get it to work. These changes shouldn't be there.

PR #2994, opened 2022-08-23
CodeFixer enhancements

tlakollo:
Move usage of the semantic model outside of the RegisterCodeFixesAsync method for all CodeFixers since it slows lightbulb creation
Add GetBestTypeByMetadataName from dotnet/roslyn and use it instead of GetTypeByMetadataName
Add numberOfIterations flag to some of the tests now that they register a code fix action instead of bailing out, then the code fix action will just return the document (no action made)
Use raw string literals in all code fixers testing files (this makes some tests to require spaces instead of tabs and change line numbers)
More nit fixes to make tests more consistent

Fixes https://github.com/dotnet/linker/issues/2986
Fixes https://github.com/dotnet/linker/issues/2987

### DIFF
--- a/src/ILLink.CodeFix/DAMCodeFixProvider.cs
+++ b/src/ILLink.CodeFix/DAMCodeFixProvider.cs
@@ -90,41 +90,41 @@ namespace ILLink.CodeFix
 			return WellKnownFixAllProviders.BatchFixer;
 		}
 
-		public override Task RegisterCodeFixesAsync (CodeFixContext context)
+		public override async Task RegisterCodeFixesAsync (CodeFixContext context)
 		{
 			var document = context.Document;
 			var diagnostic = context.Diagnostics[0];
 			var codeFixTitle = CodeFixTitle.ToString ();
 
+			if (await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false) is not { } root)
+				return;
+			if (diagnostic.AdditionalLocations.Count == 0)
+				return;
+			if (root.FindNode (diagnostic.AdditionalLocations[0].SourceSpan, getInnermostNodeForTie: true) is not SyntaxNode targetNode)
+				return;
+			if (diagnostic.Properties["attributeArgument"] is not string stringArgs || stringArgs.Contains (","))
+				return;
+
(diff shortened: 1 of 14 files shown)


src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -90,52 +90,49 @@ public sealed override FixAllProvider GetFixAllProvider ()
...
-			if (await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false) is not { } root)
-				return;
 			var diagnostic = context.Diagnostics[0];
-			if (diagnostic.AdditionalLocations.Count == 0)
-				return;
-			if (root.FindNode (diagnostic.AdditionalLocations[0].SourceSpan, getInnermostNodeForTie: true) is not SyntaxNode attributableNode)
-				return;
-			// currently not supporting multiple DAM argument, hence the check for commas in the string arguments
-			if (diagnostic.Properties[DynamicallyAccessedMembersAnalyzer.attributeArgument] is not string stringArgs || stringArgs.Contains (","))
-				return;
```
jkoritzinsky, line 105: This sort of check should be done in `RegisterCodeFixesAsync` to limit the number of "no-op" code fixes that will be offered to users. We should aim to show as few no-op fix options in normal cases as possible.

src/ILLink.CodeFix/DAMCodeFixProvider.cs
```diff
@@ -90,52 +90,49 @@ public sealed override FixAllProvider GetFixAllProvider ()
 			return WellKnownFixAllProviders.BatchFixer;
 		}
 
-		public override async Task RegisterCodeFixesAsync (CodeFixContext context)
+		public override Task RegisterCodeFixesAsync (CodeFixContext context)
 		{
 			var document = context.Document;
-			if (await document.GetSyntaxRootAsync (context.CancellationToken).ConfigureAwait (false) is not { } root)
-				return;
```
jkoritzinsky, line 97: Operating on a syntax tree is okay to do in `RegisterCodeFixesAsync`. It's only getting a `SemanticModel` object that is recommended against.

PR #2999, opened 2022-08-24
Do not report redundant suppressions on trimmed members

### LINKED ISSUE #2990: NRE in CustomAttributeSource.GetAssemblyFromCustomAttributeProvider
@vitek-karas @sbomer there's an ILLink exception with this update:

```
ILLink : error IL1012: IL Trimmer has encountered an unexpected error. Please report the issue at https://github.com/dotnet/linker/issues [/__w/1/s/src/coreclr/tools/aot/ILCompiler/ILCompiler.csproj]
##[error]ILLink(0,0): error IL1012: (NETCORE_ENGINEERING_TELEMETRY=Build) IL Trimmer has encountered an unexpected error. Please report the issue at https://github.com/dotnet/linker/issues
  Fatal error in IL Linker
     at Mono.Linker.CustomAttributeSource.GetAssemblyFromCustomAttributeProvider(ICustomAttributeProvider provider)
     at Mono.Linker.CustomAttributeSource.TryGetEmbeddedXmlInfo(ICustomAttributeProvider provider, AttributeInfo& xmlInfo)
     at Mono.Linker.CustomAttributeSource.TryGetCustomAttributeOrigin(ICustomAttributeProvider provider, CustomAttribute customAttribute, MessageOrigin& origin)
     at Mono.Linker.UnconditionalSuppressMessageAttributeState.GetSuppressionOrigin(Suppression suppression)
     at Mono.Linker.Steps.CheckSuppressionsDispatcher.Process(LinkContext context)
```

_Originally posted by @akoeplinger in https://github.com/dotnet/runtime/issues/74238#issuecomment-1220711661_

jkurdek:
The code allowed for reporting redundant suppressions on members which were removed by the linker.

This was possible e.g. when we had an assembly level suppression targeting a trimmed away member and had another warning triggered in the assembly. 

Added filter checking whether the suppression's provider is marked. Should fix #2990 

The change run into problems with suppressions on properties and events (they are not marked, only their accessors are). The PR introduces workaround for this.

### DIFF
--- a/src/linker/Linker.Steps/CheckSuppressionsDispatcher.cs
+++ b/src/linker/Linker.Steps/CheckSuppressionsDispatcher.cs
@@ -33,7 +33,6 @@ namespace Mono.Linker.Steps
 				context.LogWarning (new MessageOrigin (source), DiagnosticId.RedundantSuppression, $"IL{suppression.SuppressMessageInfo.Id:0000}");
 			}
 
-
 			bool ProviderIsMarked (ICustomAttributeProvider provider)
 			{
 				if (provider is PropertyDefinition property) {
--- a/src/linker/Linker.Steps/CheckSuppressionsDispatcher.cs
+++ b/src/linker/Linker.Steps/CheckSuppressionsDispatcher.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using ILLink.Shared;
+using Mono.Cecil;
 
 namespace Mono.Linker.Steps
 {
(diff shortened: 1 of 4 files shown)


src/linker/Linker.Steps/CheckSuppressionsDispatcher.cs
```diff
@@ -23,13 +24,30 @@ public override void Process (LinkContext context)
...
+				.Where (suppression => ProviderIsMarked (suppression.Provider));
 
 			foreach (var suppression in redundantSuppressions) {
 				var source = context.Suppressions.GetSuppressionOrigin (suppression);
 
 				context.LogWarning (new MessageOrigin (source), DiagnosticId.RedundantSuppression, $"IL{suppression.SuppressMessageInfo.Id:0000}");
 			}
+
+
+			bool ProviderIsMarked (ICustomAttributeProvider provider)
```
tlakollo, line 37: I don't have a strong preference for checking the accessors vs adding the property/event into the Annotations cache when executing MarkProperty/MarkEvent but if somebody has reasons to use the latter I'm leaving this comment to flag it

tlakollo (approved):
Looks good to me :)

src/linker/Linker.Steps/CheckSuppressionsDispatcher.cs
vitek-karas: Currently we don't do this, but it's likely we'll start soon: To reduce size, if we can prove the property is not reflection accessed we would remove the property and only leave the accessors as simple methods. So there will be a distinct difference between marking property and marking its accessor (even though currently linker doesn't do this).

But currently we don't "mark" properties (no annotation), so checking the accessors is OK. Could you please add a test where there's a property (and/or event) which is targeted by a suppression and:
* It ends up being removed (no usage)
* Used in code, but not in reflection
* Use by reflection

This should make sure that we catch the problem in this code once we implement the above mentioned optimization.
There's also probably an issue for the optimization, so maybe add a comment there that we'll need to update the logic here.

vitek-karas (approved):

PR #3000, opened 2022-08-24
Suppress warnings in 6.0 framework

sbomer:
This includes a sidecar XML file which will suppress new warnings in the 6.0 framework.
Fixes https://github.com/dotnet/linker/issues/2998
Tested in https://github.com/dotnet/sdk/pull/27461

Note that a few of the new warnings for compiler-generated code are suppressed in 7.0 with attributes on local functions. In the XML I had to suppress them with an attribute on the user method (since the signature format doesn't support compiler-generated code).

Also note that this approach only includes suppressions for the core framework, not OOB assemblies. OOB assemblies or third-party nuget packages will require a slightly different approach to avoid warnings about unresolved members referenced in XML, in apps where those assemblies aren't used. There are no warnings that we need to suppress in OOB assemblies - only new warnings in already trim-incompatible code.

This also includes a fix for a varargs method issue I found while running on the 6.0 framework.


jtschuster (approved):
Seems good to me. I haven't been able to test it, but I assume you have.

tlakollo (approved):
I learned something new about varargs, thanks!

PR #3003, opened 2022-08-25
Don't suppress all warnings with SuppressTrimAnalysisWarnings

sbomer:
This replaces https://github.com/dotnet/linker/pull/2930 with a different approach. Now SuppressTrimAnalysisWarnings only suppresses those warnings defined to be part of the "trim analysis" category, using a separate command-line argument.
An exception is 5.0 apps, where the setting continues to suppress specific warnings for compatibility.

Fixes https://github.com/dotnet/linker/issues/2982


jtschuster (approved):
Just to make sure I understand this, it looks like compared to https://github.com/dotnet/linker/pull/2930, we're just moving the logic for SuppressTrimAnalysisWarnings to the linker instead of MSBuild; is that right? And I guess custom steps can opt in to suppressing their warnings if they use the "TrimAnalysis" subcategory. Looks good to me if that's the case.

tlakollo (approved):
Looks fine to me :)

PR #3011, opened 2022-08-30
Separate the `this` parameter/argument from the method parameter/argument array

jtschuster:
We use a single array to hold the `this` parameter and method parameters/arguments in a few places, and that requires us to check for `HasImplicitThis()` anytime we want to index the array. This change removes the `this` parameter from the array and adds a side channel where necessary. This aligns better with Cecil's `MethodReference.Parameters` and makes it easier to use SourceParameterIndex in more places.

In `MethodBodyScanner.HandleCall` and `MethodBodyScanner.PopCallArguments`, separate the `this` argument from the rest of the method arguments to avoid having to check method.HasImplicitThis() anytime we want to index the array. This was already done on the path to the shared HandleCallAction, so this change just moves that higher up the call chain.

In FlowAnnotations, add a `ThisParameterAnnotation` to `MethodAnnotations` to make indexing `ParameterAnnotations` line up with `MethodReference.Parameters`. A helper `ParametersHaveAnnotations()` method checks for the `ThisParameterAnnotation` and replaces the checks for a null `ParameterAnnotations`.

Use `SourceParameterIndex` in `MethodProxy.ParameterReferenceKind` and make it a partial method.

Please let me know if you have any better names for `SourceParameterIndex`, I still don't really like the name but I can't think of anything better.


MichalStrehovsky:
> I like this approach! I would make sure to get @vitek-karas's opinion to see if this will work well in NativeAot too.

NativeAOT works with method signatures the way they exist in IL - `this` is parameter 0, always.

Separating `this` out as something special is a bit confusing. If I have a method with an explicit this, I expect the annotation for the `this` parameter would be in the aptly named `MethodAnnotations.ThisParameterAnnotation`, but it won't because it's in `ParameterAnnotations[0]` instead.

I would go the other way and make a wrapper over Cecil's parameter APIs and inject the first parameter if the method is instance and doesn't have an explicit this. With the wrapper API `this` parameter would always be index 0. Illink is an IL-based tool and things fall out more naturally if you follow those rules.

vitek-karas:
I haven't looked in detail yet, but one comment:

> Illink is an IL-based tool and things fall out more naturally if you follow those rules.

The shared code also needs to handle analyzer, which is not IL-based and thus sees "this" as very special. I don't know yet if this should be an argument to make it special everywhere, or consider analyzer "weird"... just adding this as context for now.

MichalStrehovsky:
> The shared code also needs to handle analyzer, which is not IL-based and thus sees "this" as very special. I don't know yet if this should be an argument to make it special everywhere, or consider analyzer "weird"... just adding this as context for now.

The shared abstractions should be able to represent all cases that can happen. The IL-based frontend needs to deal with explicit this because C++/CLI could generate it in some obscure scenarios. The analyzer will never have to deal with them. IMHO it's better to have the shared model be IL based than source based. The source model needs to deal with obscure scenarios like "the fake `this` when calling an extension method". We should just quickly convert those to IL terms and never think about them again. There could be another abstraction on the source side that also injects a `this` whenever appropriate to make this more streamlined (and a rule that says "you never inspect method signatures through Roslyn APIs, and always use the wrapper".

jtschuster:
Closing after our discussions in favor or IL-based indexing by default.

PR #3019, opened 2022-09-01
[release/7.0] Redundant suppression fixes

sbomer:
Backport of https://github.com/dotnet/linker/pull/2965 and https://github.com/dotnet/linker/pull/2999.

https://github.com/dotnet/linker/pull/2999 needs to go into 7.0 to fix the issue encountered in https://github.com/dotnet/sdk/pull/27564.

I also cherry-picked https://github.com/dotnet/linker/pull/2965 because the fix depends on some factoring done in that change. This does include some unrelated changes with their own risk. @vitek-karas what do you think of this approach? Would you rather do something more targeted?

I confirmed that this backport fixes the issue encountered in the SDK PR.

### DIFF
--- a/src/linker/Linker.Steps/CheckSuppressionsDispatcher.cs
+++ b/src/linker/Linker.Steps/CheckSuppressionsDispatcher.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using ILLink.Shared;
+using Mono.Cecil;
 
 namespace Mono.Linker.Steps
 {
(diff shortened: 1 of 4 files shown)


vitek-karas (approved):
I think this is fine to take for 7.

Most of the large change (which is technically not needed) is just very simple code refactoring, so very little risk (it looks much worse in the diff than it actually is). The main functional change which is added is the storage of origin of XML suppressions. But those are actually very rare in the "wild". We use them a lot in the runtime repo, but user apps will basically never run into them, so the risk is lower by the fact that the code almost never runs.
It's also a simple change, it doesn't change the behavior of anything, just stores more info.

jeffschwMSFT (approved):
approved

PR #3020, opened 2022-09-01
[release/7.0] Fixes a null ref which happens when an XML comment cref contains generic instantiation

vitek-karas:
In the cref some of the symbols (for example the type argument) are missing containing symbols, which leads to null refs.

In any case, we should not perform any analysis on symbols inside crefs, only on real code.

So this modifies the analyzer to ignore any symbol inside a cref.

Adds a test to validate this.

Backport of https://github.com/dotnet/linker/pull/3015 to 70


jeffschwMSFT (approved):
approved

Nirmal4G:
Can this be backported to v6?

PR #3021, opened 2022-09-01
Add current runtime code quality analyzer settings to linker

tlakollo:
Add current runtime analyzer settings to linker .editorconfig and eng/Analyzers.props
Fix code style warnings

*Notes:* 
- This basically breaks the baroque coding guidelines from mono (https://www.mono-project.com/community/contributing/coding-guidelines/#baroque-coding) although the repo is no longer in the mono domain several of the guidelines are still applied
- This PR can cause merge conflicts for those working in the linker repo after merging

Fixes https://github.com/dotnet/linker/issues/3018

Adding as a draft to discuss if this is the right thing we want for linker repo

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -40,11 +40,1359 @@ dotnet_style_require_accessibility_modifiers = omit_if_default:suggestion
 
 ### Code Style Analyzers
 
+## BCL analyzer rules
+
+# BCL0001: Ensure minimum API surface is respected
+dotnet_diagnostic.BCL0001.severity = warning
+
+# BCL0010: AppContext default value expected to be true
+dotnet_diagnostic.BCL0010.severity = warning
+
+# BCL0011: AppContext default value defined in if statement with incorrect pattern
+dotnet_diagnostic.BCL0011.severity = warning
+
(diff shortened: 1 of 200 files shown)


.editorconfig
```diff
@@ -40,45 +40,1453 @@ dotnet_style_require_accessibility_modifiers = omit_if_default:suggestion
 
 ### Code Style Analyzers
 
+## BCL analyzer rules
```
marek-safar, line 43: Why this way? All rules are already enabled and if there are any problematic we should disable them instead of this impossible to maintain config

tlakollo: This is a copy of what runtime is doing in  https://github.com/dotnet/runtime/blob/main/eng/CodeAnalysis.src.globalconfig
Eventually, after moving the linker to runtime this editor config would disappear and we would basically use the same globalconfig + a small editor config to keep warnings for things like unused imports

marek-safar: I have mixed feeling about doing it this way as it will get out of sync. I think it'd be better to update/fix the sources as part of the dotnet/runtime migration.

PR #3025, opened 2022-09-06
Control flow graph

jkurdek:
Currently, the MethodBodyScanner dataflow analysis relies on a simplied control flow data representation. The simplification results in an innacurate tracking of per-branch values in dataflow analysis. (#2550 ). 

The solution involves creating a control flow graph data representation and using it to improve the current dataflow analysis in the Linker. 

This PR introduces a simple control flow graph implementation and data flow analysis based on it. The new approach does not support interprocedural analysis and exception handling. For that reason it exists in parallel to the old solution, which is turned for the cases not supported by the control flow graph data flow anylysis yet.

### DIFF
--- a/src/ILLink.Shared/SharedStrings.resx
+++ b/src/ILLink.Shared/SharedStrings.resx
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
+  <!--
+    Microsoft ResX Schema
+
     Version 2.0
-    
-    The primary goals of this format is to allow a simple XML format 
-    that is mostly human readable. The generation and parsing of the 
-    various data types are done through the TypeConverter classes 
+
+    The primary goals of this format is to allow a simple XML format
+    that is mostly human readable. The generation and parsing of the
+    various data types are done through the TypeConverter classes
     associated with the data types.
-    
+
     Example:
-    
+
     ... ado.net/XML headers & schema ...
     <resheader name="resmimetype">text/microsoft-resx</resheader>
     <resheader name="version">2.0</resheader>
@@ -26,36 +26,36 @@
         <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
         <comment>This is a comment</comment>
     </data>
-                
-    There are any number of "resheader" rows that contain simple 
+
+    There are any number of "resheader" rows that contain simple
     name/value pairs.
-    
-    Each data row contains a name, and value. The row also contains a 
-    type or mimetype. Type corresponds to a .NET class that support 
-    text/value conversion through the TypeConverter architecture. 
-    Classes that don't support this are serialized and stored with the 
+
+    Each data row contains a name, and value. The row also contains a
+    type or mimetype. Type corresponds to a .NET class that support
+    text/value conversion through the TypeConverter architecture.
+    Classes that don't support this are serialized and stored with the
     mimetype set.
-    
-    The mimetype is used for serialized objects, and tells the 
-    ResXResourceReader how to depersist the object. This is currently not 
-    extensible. For a given mimetype the value must be set accordingly:
-    
-    Note - application/x-microsoft.net.object.binary.base64 is the format 
-    that the ResXResourceWriter will generate, however the reader can 
+
+    The mimetype is used for serialized objects, and tells the
+    ResXResourceReader how to depersist the object. This is currently not
+    extensible. For a given mimetype the value must be set accordingly
+
+    Note - application/x-microsoft.net.object.binary.base64 is the format
+    that the ResXResourceWriter will generate, however the reader can
     read any of the formats listed below.
-    
+
     mimetype: application/x-microsoft.net.object.binary.base64
-    value   : The object must be serialized with 
+    value   : The object must be serialized with
             : System.Runtime.Serialization.Formatters.Binary.BinaryFormatter
             : and then encoded with base64 encoding.
-    
+
     mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with 
+    value   : The object must be serialized with
             : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
             : and then encoded with base64 encoding.
 
     mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array 
+    value   : The object must be serialized into a byte array
             : using a System.ComponentModel.TypeConverter
             : and then encoded with base64 encoding.
     -->


src/linker/Linker.Dataflow/ControlFlowGraph.cs
```diff
@@ -0,0 +1,199 @@
...
+		public List<Instruction> Instructions;
+
+		public BasicBlock ()
+		{
+			Instructions = new List<Instruction> ();
+		}
+
+		public bool Equals (BasicBlock other)
+		{
+			return Instructions == other.Instructions;
```
vitek-karas, line 35: This is reference equality, not content equality. Not sure what we need here, but it does feel wrong.
Meaning:
```C#
        List<int> a = new List<int>() { 1, 2 };
        List<int> b = new List<int>() { 1, 2 };
                                       
        Console.WriteLine(a == a);  // true
        Console.WriteLine(b == a);  // false
```

src/linker/Linker.Dataflow/ControlFlowGraph.cs
```diff
@@ -0,0 +1,199 @@
...
+			throw new NotImplementedException ();
+		}
+
+		private Dictionary<BasicBlock, HashSet<BasicBlock>> SplitMethodBody (MethodBody methodBody)
+		{
+			ConstructBasicBlocks (methodBody);
+			var cfg = new Dictionary<BasicBlock, HashSet<BasicBlock>> ();
+
+			first = new BasicBlock ();
+			basicBlocks.Add (firstBlockId, first);
```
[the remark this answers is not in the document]
jkurdek: Added changes to make that possible. Please take a look.

src/linker/Linker.Dataflow/ControlFlowGraph.cs
```diff
@@ -0,0 +1,199 @@
...
+			basicBlocks.Add (lastBlockId, new BasicBlock ());
+
+
+			foreach (var (_, basicBlock) in basicBlocks) {
+				cfg.Add (basicBlock, new HashSet<BasicBlock> ());
+			}
+
+			foreach (var (_, basicBlock) in basicBlocks.Where (o => o.Value.Instructions.Count != 0).OrderBy(o => o.Key)) {
```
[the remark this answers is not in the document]
jkurdek: removed

src/ILLink.Shared/DataFlow/StructStack.cs
```diff
@@ -0,0 +1,76 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace ILLink.Shared.DataFlow
+{
+	public struct StructStack<TValue> : IEquatable<StructStack<TValue>>
+		where TValue : IEquatable<TValue>
+	{
+		private Stack<TValue>? Stack;
```
vitek-karas, line 13: Naming: the current naming convention is to call private fields like `_stack`.

src/ILLink.Shared/DataFlow/StructStack.cs
```diff
@@ -0,0 +1,76 @@
...
+
+		private readonly int _count = 0;
+
+		public StructStack ()
+		{
+		}
+
+		public StructStack (StructStack<TValue> newStack)
```
vitek-karas, line 21: Weird naming - the parameter is technically the "old" stack. You can probably just call it `stack`.

src/ILLink.Shared/DataFlow/StructStack.cs
```diff
@@ -0,0 +1,76 @@
...
+		{
+			Stack = newStack.Stack == null ? null : new Stack<TValue> (newStack.Stack);
+		}
+
+		public StructStack (int count)
```
vitek-karas, line 26: Is this meant as a "capacity" - as in how much it should allocate to begin with?
If so, the parameter should be called `capacity`. Depending on where we call this from, I might just allocate the stack immediately (since we're already asking for a set of slots).

src/ILLink.Shared/DataFlow/StructStack.cs
```diff
@@ -0,0 +1,76 @@
...
+		{
+			_count = count;
+		}
+
+		public int Count => Stack == null ? 0 : Stack.Count;
+
+		public bool Equals (StructStack<TValue> other)
+		{
+			if (this.Count != other.Count)
+				return false;
```
vitek-karas, line 36: This looks like it's a real count, not capacity... but then Push doesn't update it??

src/ILLink.Shared/TrimAnalysis/StackLattice.cs
```diff
@@ -0,0 +1,53 @@
...
+
+		public StructStack<TValue> Meet (StructStack<TValue> left, StructStack<TValue> right)
+		{
+			// Meet(value, Top) = value
+			if (left.Equals (Top)) return new StructStack<TValue> (new StructStack<TValue> (right));
+
+			if (right.Equals (Top)) return new StructStack<TValue> (new StructStack<TValue> (left));
+
+			if (left.Count != right.Count) {
+				// Force stacks to be of equal size to avoid crashes.
```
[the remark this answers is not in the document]
vitek-karas: We discussed this with Jeremi, but I wasn't sure it's actually doable. The problem is that it happens during Meet. You have two blocks which produce output state which has different stack depth. They both are predecessors to some other block. We run Meet on the states of all predecessors, and the depth discrepancy will be visible inside the Meet. Without this, we can't Meet those stacks. And scanner will not know about this since it sees each block on its own and it won't know that the stack depth is different from what it should be.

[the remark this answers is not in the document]
MichalStrehovsky: Just a note that it is invalid IL if stack heights are different. III.1.7.5 in the ECMA spec prescribes the stack height and types of elements on the stack need to be the same on control merge points. We could also abort analyzing the method by throwing a catchable exception if that's more convenient. Invalid IL is an exceptional state.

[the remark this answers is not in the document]
vitek-karas: We were trying to mimic the existing behavior (which will push unknowns in merge points). For correctness we would have to generate a warning if we give up on analysis though. For obfuscated assemblies, it might be relatively common to have unbalanced stacks and thus we could produce quite a few warnings for them, which we didn't before.
I'm not opposed to doing it though - we don't support processing obfuscated assemblies anyway.

src/linker/Linker.Dataflow/Scanner.cs
```diff
@@ -0,0 +1,984 @@
...
+						var nullableType = new RuntimeTypeHandleForNullableSystemTypeValue (new TypeProxy (resolvedDefinition), new SystemTypeValue (underlyingType));
+						state.Push (new MultiValue (nullableType));
+						return;
+					default:
+						PushUnknown (state);
+						return;
+					}
+				} else {
+					var typeHandle = new RuntimeTypeHandleValue (new TypeProxy (resolvedDefinition));
+					state.Push (new MultiValue (typeHandle));
```
vitek-karas, line 531: Nit: I think the `new MultiValue` is unnecessary - it should have an implicit conversion on it.

src/linker/Linker.Dataflow/ReflectionScanner.cs
```diff
@@ -0,0 +1,452 @@
...
+using ILLink.Shared.TrimAnalysis;
+using ILLink.Shared.TypeSystemProxy;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Mono.Linker.Steps;
+using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.SingleValue>;
+
+namespace Mono.Linker.Dataflow
+{
+	sealed class ReflectionScanner : Scanner
```
vitek-karas, line 20: This will make it difficult to merge into main - we don't want to have two copies of the same logic to maintain.

Would it be possible to move everything in Scanner to MethodBodyScanner, maybe into a subclass, such that it could still call the virtual methods on MethodBodyScanner and thus use the same ReflectionMethodBodyScanner implementation. I don't think there should be any difference in ReflectionMethodBodyScanner.

src/linker/Linker.Dataflow/LocalDataFlowAnalysis.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Collections.Generic;
+using ILLink.Shared.DataFlow;
+using ILLink.Shared.TrimAnalysis;
+using Mono.Cecil;
+using Mono.Linker.Steps;
+using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.SingleValue>;
+
+namespace Mono.Linker.Dataflow
+{
+	abstract class LocalDataFlowAnalysis<TTransfer>
```
vitek-karas, line 13: I think this is unnecessary abstraction - at least for now. I would simply hardcode the ReflectionScanner into this.

src/linker/Linker.Dataflow/ScannerExtensions.cs
```diff
@@ -39,6 +39,51 @@ public static HashSet<int> ComputeBranchTargets (this MethodBody methodBody)
...
+
+				// Targets of control flow instructions are leaders
+				if (operation.OpCode.IsControlFlowInstruction ()) {
+					var jumpTargets = operation.GetJumpTargets ();
+
+					foreach (var target in jumpTargets) {
+						leaders.Add (target.Offset);
+					}
+
+					// Instructions following conditional or unconditional jumps are leaders
```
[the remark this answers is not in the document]
jkurdek: Yes, the next instruction is not reachable from it. Currently, an unreachable basic block is created.

src/ILLink.Shared/DataFlow/StructStack.cs
jkurdek: This is the actual stack count, updated during the push. Changed the naming.

src/linker/Linker.Dataflow/ReflectionScanner.cs
jkurdek: Refactored the code so that the ReflectionMethodBody scanner is not duplicated.

src/linker/Linker.Dataflow/LocalDataFlowAnalysis.cs
jkurdek: Removed the abstract class

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -16,117 +17,41 @@
 
 namespace Mono.Linker.Dataflow
 {
-	sealed class ReflectionMethodBodyScanner : MethodBodyScanner
+	class ReflectionHandler
```
vitek-karas, line 20: I think a better name would be `TrimAnalysisHandler`. Two reasons:
* "Reflection" is not precise anymore - what this code does is more than that - it handles annotations and RUC and so on. In the analyzer and shared code we refer to this functionality as "trim analysis".
* This is the linker counterpart of `TrimAnalysisVisitor` from the analyzer - they perform basically the same services.

@sbomer, @tlakollo  - opinions?

src/linker/Linker.Dataflow/Scanner.cs
```diff
@@ -0,0 +1,984 @@
...
+using ILLink.Shared.TypeSystemProxy;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Mono.Collections.Generic;
+using static Mono.Linker.ParameterHelpers;
+using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.SingleValue>;
+
+namespace Mono.Linker.Dataflow
+{
+	abstract class Scanner
```
vitek-karas, line 18: Maybe a better name would be `LocalDataFlowScanner`:
* More descriptive (it really only handle local data flow - let's not count hoisted locals)
* It is a direct counterpart of `LocalDataFlowVisitor` from the analyzer.

vitek-karas: Basically I'm advocating for moving away from `MethodBodyScanner` and `ReflectionMethodBodyScanner` and embrace the terminology we have in analyzer and shared project.

That said, maybe this is not the right PR for such a change.

@sbomer, @tlakollo for opinions.

vitek-karas: Sounds good - maybe we should do it once we switch everything to the new CFG based scanner - so that we can only have TrimAnalysisScanner / TrimAnalysisHandler.

src/linker/Linker.Dataflow/ScannerExtensions.cs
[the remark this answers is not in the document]
vitek-karas: OK - makes sense. There was a case where linker had problems which was basically when the method ended in unconditional jump (no instruction afterward), I assume this would lead to creation of an empty basic block, which is not a problem.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -451,4 +419,36 @@ private static bool ComDangerousMethod (MethodDefinition methodDefinition, LinkC
 			return comDangerousMethod;
 		}
 	}
+
+	sealed class ReflectionMethodBodyScanner : MethodBodyScanner
```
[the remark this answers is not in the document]
jkurdek: Split into multiple files

src/ILLink.Shared/TrimAnalysis/StackLattice.cs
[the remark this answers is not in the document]
jkurdek: Removed LatticeWithUnknownValue. Added a warning indicating invalid IL.

marek-safar:
@vitek-karas what should we do about this PR?

vitek-karas:
I've been slowly working on finishing it. So if it's not too much trouble, I'd like to keep it open at least as a starting point. Obviously I'll port the change to runtime and they'll never go to linker repo.

PR #3030, opened 2022-09-08
Start transition to use IL-based parameter indexing by default

jtschuster:
After discussing in https://github.com/dotnet/linker/pull/3011, it seems we decided it would be best to use IL-based parameter indexing by default. This PR adds extension methods to IMethodSymbol and MethodReference for use instead of the `.Parameters` properties. It also adds MethodReference.Parameters to BannedSymbols.txt to warn on future use. There are a few places that use it still that I didn't want to touch yet and so I just suppressed the warnings with `#pragma`s.

There also is `NonThisParameterIndex` which is like SourceParameterIndex previously. This is just meant to translate existing code, and ideally would be removed once we remove uses of non il-based parameter indexing. 

Since we still need to be able to tell if a parameter is an implicit `this`, there is an extension method `IsThisParameterIndex` to tell if the parameter at that IL index is `this` or not.

With all of this, we can completely get rid of the MethodThisParameterValue from dataflow analysis, though this means we still need to have a special case check to see if a MethodParameterValue is `this` when reporting warnings.


src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -254,7 +254,7 @@ GenericParameterValue genericParam
 			case var callType when (callType == IntrinsicId.Type_GetConstructors || callType == IntrinsicId.Type_GetMethods || callType == IntrinsicId.Type_GetFields ||
 				callType == IntrinsicId.Type_GetProperties || callType == IntrinsicId.Type_GetEvents || callType == IntrinsicId.Type_GetNestedTypes || callType == IntrinsicId.Type_GetMembers)
 				&& calledMethod.IsDeclaredOnType ("System.Type")
-				&& calledMethod.HasParameterOfType (0, "System.Reflection.BindingFlags")
+				&& calledMethod.HasParameterOfType ((NonThisParameterIndex) 0, "System.Reflection.BindingFlags")
```
vitek-karas, line 257: Not necesarilly in this PR - but eventually we should switch this to ILParameterIndex as well.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -238,46 +226,30 @@ TypeAnnotations BuildTypeAnnotations (TypeDefinition type)
...
-							}
-						} else if (methodMemberTypes != DynamicallyAccessedMemberTypes.None) {
-							_context.LogWarning (method, DiagnosticId.DynamicallyAccessedMembersIsNotAllowedOnMethods);
-						}
-					} else {
-						offset = 0;
-						if (methodMemberTypes != DynamicallyAccessedMemberTypes.None) {
-							_context.LogWarning (method, DiagnosticId.DynamicallyAccessedMembersIsNotAllowedOnMethods);
-						}
+					// Warn is there is an annotation on a method without a `this` parameter -- we won't catch it in the for loop if there's no parameters
```
vitek-karas, line 229:
```suggestion
					// Warn if there is an annotation on a method without a `this` parameter -- we won't catch it in the for loop if there's no parameters
```

src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
jtschuster: Agreed, there's still a lot more work to do, I just thought it'd be best to separate it into multiple PR's.

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -300,6 +292,7 @@ public override bool HandleCall (MethodBody callingMethodBody, MethodReference c
 						if (staticType is null) {
 							// We don't know anything about the type GetType was called on. Track this as a usual result of a method call without any annotations
 							AddReturnValue (context.Annotations.FlowAnnotations.GetMethodReturnValue (calledMethodDefinition));
+							//AddReturnValue (UnknownValue.Instance);
```
vitek-karas, line 295: ???

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
```diff
@@ -444,9 +437,11 @@ static bool IsComInterop (IMarshalInfoProvider marshalInfoProvider, TypeReferenc
 		private static bool ComDangerousMethod (MethodDefinition methodDefinition, LinkContext context)
 		{
 			bool comDangerousMethod = IsComInterop (methodDefinition.MethodReturnType, methodDefinition.ReturnType, context);
+#pragma warning disable RS0030
```
vitek-karas, line 440: Please add a comment what the warning RS0030 is... and why it's OK to violate it here

vitek-karas (approved):

src/linker/Linker.Dataflow/ReflectionMethodBodyScanner.cs
jtschuster: Sorry, forgot to remove that

jtschuster:
> I think the meaningful distinction is not between IL/source parameters (neither includes the implicit this parameter), but between IL/source arguments.

I think this PR kind of does that, where ILParameterIndex is the index of the value that is pushed on the stack when the method is called. Maybe renaming ILParameterIndex to something like ILArgumentIndex would make more sense, then NonThisParameterIndex could just be ParameterIndex since that would align with both IL metadata and C# source code?

vitek-karas:
Isn't it different inside a method body?
If you want the "this" it's always argument index 0. Regardless if the this is implied or not.

MichalStrehovsky:
I think what might help the discussion is how we want to handle the case where method has an explicit this, and someone put a `[DAM]` annotation on the explicit this parameter. Would that fall out naturally? The abstraction that we're building here should ideally make this behave identically to `[DAM]` on the method itself and an implicit `this`.

The explicit `this` parameter encoding in the metadata was actually built exactly for this purpose - sometimes compilers want to attach custom modifiers or custom attributes to `this` and explicit this allows that. We hacked around it in the DAM annotation case in C# by annotating the method itself because we don't control the C# compiler and cannot make the compiler generate explicit this with an attribute, but explicit this with an attribute is the "correct" way to annotate DAM on this parameter.

jtschuster:
> I think what might help the discussion is how we want to handle the case where method has an explicit this, and someone put a [DAM] annotation on the explicit this parameter. Would that fall out naturally?

This PR adds an extension method `MethodReference.IsThisParameter(ILParameterIndex)` that can make it fall out naturally. It checks for `HasImplicitThis()`, but that can be changed to `HasThis`.

Also, rereading the excepts from @sbomer, this wording:

>Normally, a parameter list (which always follows the calling convention) does not provide information
about the type of the this pointer, since this can be deduced from other information.

to me sounds like the metadata only leaves out the 'this' from the parameter metadata section to save space and that 'this' is still considered a "parameter". In the second excerpt:

> the presence of a this pointer affects parameter signature/argument number pairing in CIL

the wording of "parameter signature" instead of "parameter" points me in the same direction. Since Cecil is meant to just model the metadata, they also leave out the 'this' from `Parameters`, but I don't think that's the best way to think about it for our purposes. I still find it most natural to think about parameters and arguments as having the same arity and the difference just being 'parameters' is "what a method takes" and 'arguments' is "what a method is given at a call site".

@MichalStrehovsky You probably know the spec the best, do you have any thoughts on if this interpretation makes sense?

MichalStrehovsky:
> the difference just being 'parameters' is "what a method takes" and 'arguments' is "what a method is given at a call site".

That's how I think of parameters/arguments too.

> @MichalStrehovsky You probably know the spec the best, do you have any thoughts on if this interpretation makes sense?

For me, `this` a parameter - I think it falls out from "explicit this" and "implicit this" - those two say "`this` is explicitly spelled out in the signature" and "`this` is implicitly present in the signature" respectively. It's considered part of the signature in either case.

MichalStrehovsky:
I don't have a strong opinion on this - for me it just boils down to how do we match the annotations of the target of the call with the value that is in ldarg.0. Whether the API to access the annotations for `this` is different, or whether we can call into a common piece of code that knows how to map the 0 to the encoding we chose for `this` (because that place will know whether it's is an instance methods or not).

It feels easier to do all the special casing in a single place. But I'll defer the judgement to those who spend the most time in this codebase.

jtschuster:
Closing in favor of https://github.com/dotnet/linker/pull/3059.

PR #3043, opened 2022-09-19
Add event source tracing

jtschuster:
Adds EventSource events for the start and end of `Main()` and for the start and end of each call to `IStep.Process()`.

The dotnet/performance scenarios can use ETW events to give a more detailed breakdown of where time is spent. This PR creates a few simple events to be able to track the time the linker takes from process start to end more precisely without any noise from MSBuild. If we want to track more fine-grained details, it will be easy to add them in the future.

LinkerStart is emitted right at the start of Main().
LinkerStop is emitted right before returning from Main().
LinkerStepStart is emitted right before calling IStep.Process.
LinkerStrpStop is emitted right after IStep.Process returns.


vitek-karas (approved):
Looks good - thanks!

PR #3045, opened 2022-09-20
[release/7.0] Don't hardcode Microsoft.CodeAnalysis version for source-build

lbussell:
When building from source, we need to use the latest version of Microsoft.CodeAnalysis. Since source-build only supports Linux and doesn't need to worry about Visual Studio, this doesn't cause the issue described in the comment.

I also think it would be good to check if Visual Studio has already updated the Microsoft.CodeAnalysis.Csharp version or find out when it will.


jeffschwMSFT (approved):
approved

PR #3052, opened 2022-09-27
[release/7.0] Use new build images

sbomer:
Port of https://github.com/dotnet/linker/pull/3051


vitek-karas (approved):
Based on the email maybe we should wait for the next release first, to avoid unnecessary DARC flow.

PR #3059, opened 2022-10-06
Create `ParameterProxy` to wrap logic surrounding parameters and use one `ParameterIndex` enum to index

jtschuster:
Creates `ParameterProxy` to represent parameters as a pair of method + parameter index. `IParameterSymbol` and `ParameterDefinition` don't usually include the implicit 'this' parameter, so this lets us treat the implicit 'this' parameter as the same as any other parameter. This also should help put all the `HasImplicitThis() ? 1 : 0` in a limited number of places (`ParameterProxy` and the extensions methods that produce them).

`ParameterIndex` represents 'this' as the 0 index. See the code docs for more info. I found it was best to represent the 'this' as 0 and provide a property on the `ParameterProxy` for indexing int the `Parameters` collection (`int MetadataIndex`). 

"Metadata Parameter" is what I used to name the index in the `.parameters` metadata section, which corresponds to the SourceParamterIndex from before.

In the process, I found an issue in the analyzer where some properties were indexers but were not being passed the index value in `HandleCallAction`. Since we previously only iterated over the arguments, no issues arose, but after changing to iterate over the parameters in the called method signature, the issue showed up. LocalDataflowVisitor has been updated to visit the indexer argument and pass the value to `HandleCallAction`.


src/ILLink.RoslynAnalyzer/TrimAnalysis/FlowAnnotations.cs
```diff
@@ -23,31 +24,34 @@ sealed partial class FlowAnnotations
...
-			var parameterMethod = (IMethodSymbol) parameter.ContainingSymbol;
-			Debug.Assert (parameterMethod != null);
+			Debug.Assert (method != null);
 			// If there are conflicts between the setter and the property annotation,
 			// the setter annotation wins. (But DAMT.None is ignored)
-			if (parameterMethod!.MethodKind == MethodKind.PropertySet && damt == DynamicallyAccessedMemberTypes.None) {
-				var property = (IPropertySymbol) parameterMethod.AssociatedSymbol!;
+			if (method!.MethodKind == MethodKind.PropertySet && damt == DynamicallyAccessedMemberTypes.None
+				// Indexer setters (.Item[int].set) will have the indexer as the first param and the value as the last. We only want annotations applied to the value parameter.
+				&& (int) param.Index + 1 == param.Method.GetParametersCount ()) {
```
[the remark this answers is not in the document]
jtschuster: Since there could be multiple indices on the indexer, we'll need to check if `param` is the last parameter if it is an indexer. We could only enforce that check if the method is an indexer, but we'd have to cast to IPropertySymbol. Depending on how expensive a cast is, it might not gain us much.

src/ILLink.RoslynAnalyzer/IMethodSymbolExtensions.cs
```diff
@@ -0,0 +1,85 @@
...
+		{
+			return !method.IsStatic;
+		}
+
+		/// <summary>
+		/// Returns a list of the parameters pushed onto the stack before the method call (including the implicit 'this' parameter)
+		/// </summary>
+		public static List<ParameterProxy> GetParameters (this IMethodSymbol method)
+		{
+			List<ParameterProxy> parameters = new ();
```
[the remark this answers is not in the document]
jtschuster: I agree, I don't love the idea of allocating for each call. I'm not 100% sure how foreach works with IEnumerable, but we could make a struct that implements IEnumerable<ParamterProxy> and return that struct from GetParameters(), but I'm guessing it would get boxed by the foreach. We'd be allocating the IEnumerable and IEnumerator instead of a whole list which could have some gains, but we'd lose the foreach optimization for List.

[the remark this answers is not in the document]
jtschuster: Created a `struct ParameterCollection` to avoid allocating.

src/ILLink.Shared/ParameterIndex.cs
```diff
@@ -0,0 +1,64 @@
...
+		public static ParameterIndex operator - (ParameterIndex left, ParameterIndex right)
+			=> new ParameterIndex (left.Index - right.Index);
+
+		public static ParameterIndex operator + (ParameterIndex left, int right)
+			=> new ParameterIndex (left.Index + right);
+
+		public static ParameterIndex operator - (ParameterIndex left, int right)
+			=> new ParameterIndex (left.Index - right);
+
+		public static readonly ParameterIndex This = new ParameterIndex (0);
```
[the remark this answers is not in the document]
jtschuster: Good point, I've removed the field and replaced references with `(ParameterIndex) 0`.

src/ILLink.Shared/TrimAnalysis/Intrinsics.cs
```diff
@@ -160,102 +160,102 @@ public static IntrinsicId GetIntrinsicIdForMethod (MethodProxy calledMethod)
...
 					=> IntrinsicId.Type_GetMethod,
 
 				// System.Type.GetMethods (BindingFlags)
 				"GetMethods" when calledMethod.IsDeclaredOnType ("System.Type")
-					&& calledMethod.HasParameterOfType (0, "System.Reflection.BindingFlags")
-					&& calledMethod.HasParametersCount (1)
-					&& !calledMethod.IsStatic ()
+					&& calledMethod.HasImplicitThis ()
+					&& calledMethod.HasMetadataParametersCount (1)
+					&& calledMethod.HasParameterOfType ((ParameterIndex) 1, "System.Reflection.BindingFlags")
```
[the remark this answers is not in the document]
jtschuster: I think as we get to the point where each IntrinsicId corresponds to exactly one method, we'll be able to use only HasImplicitThis and HasParameterCount. I only made HasMetadataParameterCount to make it simpler to transition everything in the first PR, and then we can remove all uses of it in a separate PR with all the intrinsics specificity improvements.

For being more explicit on what the int means in the casts to ParameterIndex, I'm not sure if calling a method on an int makes it any clearer than a cast, either way you need to check the xml docs to make sure you understand what the int represents when you cast it or pass it as an argument.

src/linker/Linker.Dataflow/FlowAnnotations.cs
```diff
@@ -347,8 +313,8 @@ TypeAnnotations BuildTypeAnnotations (TypeDefinition type)
 							_context.LogWarning (setMethod, DiagnosticId.DynamicallyAccessedMembersConflictsBetweenPropertyAndAccessor, property.GetDisplayName (), setMethod.GetDisplayName ());
 						} else {
 							int offset = setMethod.HasImplicitThis () ? 1 : 0;
-							if (setMethod.Parameters.Count > 0) {
-								DynamicallyAccessedMemberTypes[] paramAnnotations = new DynamicallyAccessedMemberTypes[setMethod.Parameters.Count + offset];
+							if (setMethod.HasParameters) {
```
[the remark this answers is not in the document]
jtschuster: Added MethodReference.HasMetadataParameters() and banned HasParameters from Mono.Linker.csproj

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -199,7 +198,14 @@ public override TValue VisitSimpleAssignment (ISimpleAssignmentOperation operati
 						// For now, just don't warn. https://github.com/dotnet/linker/issues/2731
 						break;
 					}
-					HandleMethodCall (setMethod, instanceValue, ImmutableArray.Create (value), operation);
+
+					// Property may be an indexer, in which case there will be one or more index arguments followed by a value argument
```
vitek-karas, line 202: Definitely a good fix, but I would prefer to do this in a separate PR - along with some tests for this.

src/ILLink.Shared/TypeSystemProxy/ParameterProxy.cs
```diff
@@ -0,0 +1,54 @@
...
+			Method = method;
+			Index = index;
+		}
+
+		public MethodProxy Method { get; }
+
+		public ParameterIndex Index { get; }
+
+		/// <summary>
+		/// The index of the entry in the '.parameters' metadata section corresponding to thie parameter.
```
vitek-karas, line 24:
```suggestion
		/// The index of the entry in the '.parameters' metadata section corresponding to this parameter.
```

src/ILLink.Shared/TypeSystemProxy/ParameterProxy.cs
```diff
@@ -0,0 +1,54 @@
...
+						throw new InvalidOperationException ("Cannot get metadata index of the implicit 'this' parameter");
+					return (int) Index - 1;
+				}
+				return (int) Index;
+			}
+		}
+
+		// C# doesn't have partial properties, but the following should be implemented in every project
+
+		// public partial ReferenceKind RefKind { get; }
```
vitek-karas, line 41: In the other proxies I tend to change this into a Get method - it's not as nice as a property, but it's OK and the partial works on it.

src/ILLink.Shared/TypeSystemProxy/ParameterProxy.cs
```diff
@@ -0,0 +1,54 @@
...
+
+		public partial string GetDisplayName ();
+
+		public bool IsImplicitThis => Method.HasImplicitThis () && Index == (ParameterIndex) 0;
+
+		public partial bool IsTypeOf (string typeName);
+
+		public IEnumerable<string> GetDiagnosticArgumentsForAnnotationMismatch ()
```
vitek-karas, line 49: I would prefer we "hide" the fact that implicit this parameter doesn't have a "name" and instead uses the method's name in the message alone. That is the GetDisplayName would be expected to not be called on the implicit this (maybe it throws?) and here we would just use the methods name directly.

As-is both this method and the implementation of GetDisplayName have to know that implicit this returns the method's name - which is weird (if it should return something, then probably constant `this`).

src/ILLink.Shared/TypeSystemProxy/ParameterCollection.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace ILLink.Shared.TypeSystemProxy
+{
+	internal struct ParameterCollection
```
vitek-karas, line 6: Couple of questions:
* Do we need this? Why not just `IEnumerable<ParameterProxy>` - perf? - if so please add a comment
* Naming - probably better name would be `ParameterProxyCollection`. Or even better `ParameterProxyEnumerable` - or something like that - it's not a Collection really.
* I find it really weird that it doesn't implement `IEnumerable<ParameterProxy>` - even though it would never be used - since compiler will optimize and use the exact types directly, it's really weird. It would also act as a documentation what this is for.

src/ILLink.Shared/ParameterIndex.cs
```diff
@@ -0,0 +1,64 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace ILLink.Shared.TypeSystemProxy
+{
+	/// <summary>
+	/// Used to indicate the index of the parameter in the Parameters metadata section (i.e. the first parameter that is not the implicit 'this' is 0)
+	/// It is very error prone to use an int to represent the index in parameters metadata section / source code parameter index as well as for indexing into argument lists.
+	/// Instead, use this enum whenever representing an index of a parameter.
```
vitek-karas, line 11: It's struct now.

src/ILLink.Shared/TypeSystemProxy/NonThisParameterIndex.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace ILLink.Shared
+{
+	/// <summary>
```
vitek-karas, line 6: What is this for?
Looks like an empty file.

src/ILLink.Shared/TypeSystemProxy/MethodProxy.cs
```diff
@@ -13,17 +13,51 @@ namespace ILLink.Shared.TypeSystemProxy
...
 		internal partial bool IsDeclaredOnType (string fullTypeName);
-		internal partial bool HasParameters ();
+
+		/// <summary>
+		/// Returns the number of the parameters in the 'parameters' metadata section. This should map directly to the number of parameters in the C# source declaration as well.
+		/// </summary>
+		internal partial int GetMetadataParametersCount ();
+
+		/// <summary>
+		/// Returns true if the method has parameters in the 'paramters' metadata section (i.e. has parameters besides the implicit 'this' parameter)
```
vitek-karas, line 23:
```suggestion
		/// Returns true if the method has parameters in the 'parameters' metadata section (i.e. has parameters besides the implicit 'this' parameter)
```

src/ILLink.Shared/TypeSystemProxy/MethodProxy.cs
```diff
@@ -13,17 +13,51 @@ namespace ILLink.Shared.TypeSystemProxy
...
+		/// Returns a List of <see cref="ParameterProxy"/> representing the parameters the method takes, including the implicit 'this' parameters.
+		/// </summary>
+		internal partial ParameterCollection GetParameters ();
+
+		/// <summary>
+		/// Returns the ParameterProxy corresponding to the parameter at <paramref name="index"/>.
+		/// <paramref name="index"/> is the index of the parameters as they are passed to the method, with 0 being the implicit this parameter if it exists.
+		/// See <see cref="ParameterIndex"/> for more info.
+		/// </summary>
+		internal partial ParameterProxy? GetParameter (ParameterIndex index);
```
vitek-karas, line 43: Do we actually use the fact that it can return `null` if the parameter index is out of range?
I would prefer we throw in that case.

src/ILLink.RoslynAnalyzer/TrimAnalysis/MethodParameterValue.cs
```diff
@@ -1,32 +1,43 @@
...
+
+		public ParameterIndex Index => Parameter.Index;
 
 		public override SingleValue DeepCopy () => this; // This value is immutable
 
 		public override string ToString ()
-			=> this.ValueToString (ParameterSymbol, DynamicallyAccessedMemberTypes);
+			=> this.ValueToString (MethodSymbol, DynamicallyAccessedMemberTypes);
+
+		public bool IsThisParameter () => _overrideIsThis || Parameter.IsImplicitThis;
```
vitek-karas, line 41: Could we move this to the shared code? It seems to be identical.

src/linker/Linker.Dataflow/ParameterProxy.cs
```diff
@@ -0,0 +1,57 @@
...
+		public partial string GetDisplayName () => IsImplicitThis ? Method.GetDisplayName ()
+			: !string.IsNullOrEmpty (Method.Method.Parameters[MetadataIndex].Name) ? Method.Method.Parameters[MetadataIndex].Name
+			: $"#{Index}";
+#pragma warning restore RS0030 // Do not used banned APIs
+
+		public ICustomAttributeProvider GetCustomAttributeProvider ()
+		{
+			if (IsImplicitThis)
+				return Method.Method;
+#pragma warning disable RS0030 // MethodReference is banned -- this class provides wrappers to use
```
vitek-karas, line 48: Personally I would put this around the entire class - it's a proxy so it is meant to access the underlying thing directly.

src/linker/Linker.Dataflow/TrimAnalysisMethodCallPattern.cs
```diff
@@ -27,7 +27,6 @@ namespace Mono.Linker.Dataflow
 			ImmutableArray<MultiValue> arguments,
 			MessageOrigin origin)
 		{
-			Debug.Assert (origin.Provider is MethodDefinition);
```
vitek-karas, line 30: Why did you remove this?

src/ILLink.Shared/TypeSystemProxy/MethodProxy.cs
jtschuster: We use it in MethodProxy.HasParameterOfType() where we return false if the index is out of range and GetParameter, but we could have an out of range check in HasParameterOfType() and throw in GetParameter.

src/ILLink.Shared/TypeSystemProxy/ParameterCollection.cs
jtschuster: Yeah, it was for perf. I like those names better, and I'll add the IEnumerable implementation.

src/ILLink.Shared/TypeSystemProxy/ParameterProxy.cs
jtschuster: That makes sense, I'll update GetDisplayName to return "this" and used Method.GetDisplayName() for the IsImplicitThis case of GetDiagnosticArgumentsForAnnotationsMismatch()

src/linker/Linker.Dataflow/ParameterProxy.cs
jtschuster: I wanted to disable the warning on as little code as possible since there are other banned symbols, and a different banned symbol could sneak in if we disable RS0030 on the whole class. It's not pretty, but I think it might be better to limit the scope of disabled warnings.

src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
jtschuster: Moved those changes to https://github.com/dotnet/linker/pull/3081

src/linker/Linker.Dataflow/TrimAnalysisMethodCallPattern.cs
jtschuster: That was by mistake, adding it back

src/ILLink.Shared/TypeSystemProxy/MethodProxy.cs
jtschuster: I changed it to throw and added a bounds check in HasParameterOfType()

vitek-karas (approved):
Looks good! Thanks!

PR #3060, opened 2022-10-06
Remove redundant checks in IntrinsicId switch statements

jtschuster:
In ReflectionMethodBodyScanner and HandleCallAction, we do many of the same checks that already occur in `GetIntrinsicIdFromMethod`, and in a few places we have 'when' conditions that could just be regular cases. This removes the extra checks so the switch can be simpler and more optimized.

In the linker, we also get the IntrinsicId twice, so this adds an optional parameter for a known IntrinsicId.

### DIFF
--- a/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
+++ b/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
@@ -27,7 +27,7 @@ namespace ILLink.Shared.TrimAnalysis
 		readonly FlowAnnotations _annotations;
 		readonly RequireDynamicallyAccessedMembersAction _requireDynamicallyAccessedMembersAction;
 
-		public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnlyList<MultiValue> argumentValues, out MultiValue methodReturnValue, out IntrinsicId intrinsicId, IntrinsicId? knownIntrinsicId = null)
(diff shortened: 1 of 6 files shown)


src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -27,15 +27,15 @@ partial struct HandleCallAction
...
+		public bool Invoke (MethodProxy calledMethod, MultiValue instanceValue, IReadOnlyList<MultiValue> argumentValues, out MultiValue methodReturnValue, out IntrinsicId intrinsicId, IntrinsicId? knownIntrinsicId = null)
 		{
 			MultiValue? returnValue = null;
 
 			bool requiresDataFlowAnalysis = _annotations.MethodRequiresDataFlowAnalysis (calledMethod);
 			var annotatedMethodReturnValue = _annotations.GetMethodReturnValue (calledMethod);
 			Debug.Assert (requiresDataFlowAnalysis || annotatedMethodReturnValue.DynamicallyAccessedMemberTypes == DynamicallyAccessedMemberTypes.None);
 
-			intrinsicId = Intrinsics.GetIntrinsicIdForMethod (calledMethod);
+			intrinsicId = knownIntrinsicId ?? Intrinsics.GetIntrinsicIdForMethod (calledMethod);
```
vitek-karas, line 38: I would prefer we make the parameter mandatory. In linker it should be always known, in analyzer it isn't, but it's easy to call the helper.

PR #3073, opened 2022-10-18
Check for marking virtual method due to base only when state changes

jtschuster:
Instead of checking every virtual method to see if it should be kept due to a base method every iteration of the MarkStep pipeline, check each method only when its relevant state has changed. These state changes are:
- The override's declaring type is marked as instantiated
- The base method is marked
- The override's declaring type is marked as relevant to variant casting

This required some changes to `TypeMapInfo`:
- modify `GetBaseMethods` to return OverrideInformation's
- add a new method, `GetInterfaceImplementors` to get all types that implement a given interface.

I profiled the ASP.Net benchmark's build link time and got the following time in CPU ms:
27325 before the regression related to https://github.com/dotnet/linker/issues/3067
87310 in main
32011 after this change

This also changes behavior in a couple places:
* When a type is marked as instantiated. Previously, we would mark all methods that override a method from a preserved scope, which would mark methods implementing an interface even if the interface implementation isn't kept. Now, the method should only be kept if the interface is marked and the interface implementation is marked.
* IDynamicInterfaceCastable is special cased to keep the interface implementation on all types that are instantiated or relevant to variant casting, regardless of whether or not IDynamicInterfaceCastable is marked or not.

Closes https://github.com/dotnet/linker/issues/3067
Closes https://github.com/dotnet/linker/issues/3069


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2334,6 +2376,32 @@ void MarkInterfaceImplementations (TypeDefinition type)
 				if (ShouldMarkInterfaceImplementation (type, iface))
 					MarkInterfaceImplementation (iface, new MessageOrigin (type));
 			}
+
+			bool ShouldMarkInterfaceImplementation (TypeDefinition type, InterfaceImplementation iface)
```
jtschuster, line 2380: This is moved to a local method because it assumes that 'type' is marked as instantiated, so we wouldn't want to call it elsewhere.

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3688,41 +3773,19 @@ protected virtual void MarkInstruction (Instruction instruction, MethodDefinitio
...
-
-
-			return IsFullyPreserved (type);
-		}
 
 		protected internal virtual void MarkInterfaceImplementation (InterfaceImplementation iface, MessageOrigin? origin = null, DependencyInfo? reason = null)
 		{
 			if (Annotations.IsMarked (iface))
 				return;
+			Annotations.MarkProcessed (iface, reason ?? new DependencyInfo (DependencyKind.InterfaceImplementationOnType, ScopeStack.CurrentScope.Origin.Provider));
```
jtschuster, line 3781: Before moving the MarkProcessed call, the linker hit a cycle where marking iface.InterfaceType would call MarkInterfaceImplementation with the same arguments and overflow the stack

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -1904,7 +1932,7 @@ internal void MarkStaticConstructorVisibleToReflection (TypeDefinition type, in
 			if (reference == null)
 				return null;
 
-			using var localScope = origin.HasValue ? ScopeStack.PushScope (origin.Value) : null;
+			using var localScoe = origin.HasValue ? ScopeStack.PushScope (origin.Value) : null;
```
marek-safar, line 1935: typo?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2334,6 +2376,32 @@ void MarkInterfaceImplementations (TypeDefinition type)
...
+
+				if (Annotations.IsMarked (resolvedInterfaceType))
+					return true;
+
+				// It's hard to know if a com or windows runtime interface will be needed from managed code alone,
+				// so as a precaution we will mark these interfaces once the type is instantiated
+				if (resolvedInterfaceType.IsImport || resolvedInterfaceType.IsWindowsRuntime)
+					return true;
+
+				if (resolvedInterfaceType.IsTypeOf<IDynamicInterfaceCastable> ())
```
marek-safar, line 2399: Why is this needed, should this be fixed on libraries side?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -721,43 +716,76 @@ void ProcessVirtualMethod (MethodDefinition method)
...
+				Debug.Assert (!Context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, overrideInformation.Override) || overrideInformation.Base.IsAbstract);
+				MarkMethod (overrideInformation.Override, new DependencyInfo (DependencyKind.Override, overrideInformation.Base), ScopeStack.CurrentScope.Origin);
 			}
+		}
 
-			if (method.IsVirtual)
-				ProcessVirtualMethod (method);
+		void MarkMethodIfNeededByBaseMethod (MethodDefinition method)
+		{
+			if (!Annotations.IsMarked (method.DeclaringType) || Annotations.GetBaseMethods (method) is not List<OverrideInformation> bases)
```
marek-safar, line 766: Why is the `is not` check needed?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -721,43 +716,76 @@ void ProcessVirtualMethod (MethodDefinition method)
...
+
+			IEnumerable<(OverrideInformation Override, bool IsDirectBase)> GetMarkedBaseMethods (IEnumerable<OverrideInformation> overrides)
+			{
+				foreach (var ov in overrides) {
+					// Immediate base methods
+					if (Annotations.IsMarked (ov.Base) || IgnoreScope (ov.Base.DeclaringType.Scope))
+						yield return (ov, true);
+
+					// Bases of bases
+					else if (Annotations.GetBaseMethods (ov.Base) is IEnumerable<OverrideInformation> baseBases) {
```
marek-safar, line 782: Is the `is` check redundant?

src/linker/Linker.Steps/SealerStep.cs
```diff
@@ -106,7 +107,7 @@ void ProcessType (TypeDefinition type)
 				if (!type.IsSealed)
 					continue;
 
-				var bases = Annotations.GetBaseMethods (method);
+				var bases = Annotations.GetBaseMethods (method)?.Select (static x => x.Base).ToList ();
```
marek-safar, line 110: You don't need `ToList ()` here

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -721,43 +716,76 @@ void ProcessVirtualMethod (MethodDefinition method)
...
+				return false;
+			}
 
-			if (Annotations.IsProcessed (method))
-				return;
+			var isInstantiated = Annotations.IsInstantiated (overrideInformation.Override.DeclaringType);
 
-			if (Annotations.IsMarked (method))
-				return;
+			bool markForOverride = !Context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, overrideInformation.Override);
```
marek-safar, line 736: Should not this "abort" the whole logic here?

jtschuster: That's a good point, we can early return instead of or-ing markForOverride each time.

src/linker/Linker.Steps/MarkStep.cs
jtschuster: If `Annotations.GetBaseMethods` returns null we want to return early,, so I used that notation to combine the assignment and null check into one `if` statement.

src/linker/Linker.Steps/MarkStep.cs
jtschuster: The Annotations.Get[BaseMethods|Overrides|etc] can return null. I was thinking about changing the methods to return IEnumerable.Empty if not found in the cache, or adding an empty list to the dictionary, but thought there might have been a reason it was done this way before.

marek-safar: You used type check not null check. I agree it should check for null but that could be simple `x == null` or `x is not null`

src/linker/Linker.Steps/MarkStep.cs
marek-safar: Same as bellow `!= null` is better

src/linker/Linker.Steps/MarkStep.cs
marek-safar: It seems to be it'd be better to add the interface to corelib descriptors instead of hardcoding in linker

src/linker/Linker.Steps/MarkStep.cs
jtschuster: In the case below there was a type cast, but here since the return type and pattern type are the same Roslyn will turn the pattern match into a null check. Is there still a reason to do assignment + null check besides style?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -721,43 +716,79 @@ void ProcessVirtualMethod (MethodDefinition method)
...
 
-			if (Annotations.IsMarked (method))
-				return;
+			if (!Context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, overrideInformation.Override))
+				return true;
 
-			var isInstantiated = Annotations.IsInstantiated (method.DeclaringType);
+			// Methods on instantiated types that override a ov.Override from a base type (not an interface) should be marked 
+			// Interface ov.Overrides should only be marked if the interfaceImplementation is marked, which is handled below
+			if (!overrideInformation.Base.DeclaringType.IsInterface && isInstantiated)
```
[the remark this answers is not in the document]
jtschuster: Yes, thanks for pointing that out.

src/linker/Linker.Steps/MarkStep.cs
jtschuster: Good point, and it turns out IDynamicInterfaceCastable is in the corelib descriptors, we just need to make sure we pass the framework libraries as 'trim' instead of 'skip'.

src/linker/Linker.Steps/MarkStep.cs
marek-safar: I think the null check is much more readable and describes the intent better

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2003,6 +2038,16 @@ internal void MarkStaticConstructorVisibleToReflection (TypeDefinition type, in
 			if (type.IsInterface) {
 				// There's no benefit to deferring processing of an interface type until we know a type implementing that interface is marked
 				MarkRequirementsForInstantiatedTypes (type);
+				// Mark interface implementation on each implementor that requires interfaces.
+				if (Annotations.GetInterfaceImplementors (type) is List<(TypeDefinition Implementor, InterfaceImplementation InterfaceImplementation)> implementors) {
```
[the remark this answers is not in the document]
jtschuster: Thanks for pointing that out, this was part of some work to be able to remove parts of ProcessMarkedTypesWithInterfaces, but should be removed.

jtschuster:
I rebased by mistake and the git history got messed up and duplicated a bunch of commits. The only real changes should be from [2bc88c0](https://github.com/dotnet/linker/pull/3073/commits/2bc88c03286a7665f41b1cf732fe35834e2e1f3a).

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -721,43 +715,84 @@ void ProcessVirtualMethod (MethodDefinition method)
...
+				Debug.Assert (!Context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, overrideInformation.Override) || overrideInformation.Base.IsAbstract);
+				MarkMethod (overrideInformation.Override, new DependencyInfo (DependencyKind.Override, overrideInformation.Base), ScopeStack.CurrentScope.Origin);
 			}
+		}
 
-			// Interface static veitual methods will be abstract and will also by pass this check to get marked
-			if (!isInstantiated && !@base.IsAbstract && Context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, method))
+		void MarkMethodIfNeededByBaseMethod (MethodDefinition method)
+		{
+			if (!Annotations.IsMarked (method.DeclaringType))
```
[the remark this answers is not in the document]
jtschuster: Ah you're right, I was looking at the wrong thing. It does return early though when MarkAsRelevantToVariantCasting is called before MarkType in MarkInstruction and MarkTypeVisibleToReflection. Changing MarkTypeAsRelevantToVariantCasting to include a MarkType causes some test failures, so I think changing this would be best in another PR if we do want to remove it.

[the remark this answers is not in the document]
jtschuster: Since interface method handling is still separate, we can remove the call to `MarkMethodsOnTypeIfNeededByBaseMethod` from `MarkAsRelevantToVariantCasting`

PR #3075, opened 2022-10-18
Use LinkContext caching when resolving ExportedTypes

jtschuster:
When profiling ASP.Net benchmark builds, I found significant time was being spent resolving `ExportedType`
's since we don't have a cache for those like we do for TypeReference and MethodReference.

Bans MethodReference.Resolve and ExportedType.Resolve and recommend LinkContext.Resolve instead (except in SweepStep).

Adds a cache on LinkContext for resolving ExportedTypes and replaces calls to Cecil's Resolve.

Replaces some calls to MethodReference.Resolve.

Together with https://github.com/dotnet/linker/pull/3073, performance improves to better than before the virtual method regression in August.

Trimmer CPU milliseconds on ASP.Net benchmarks build:
27325 before the regression related to https://github.com/dotnet/linker/issues/3067
87310 in main
23520 after https://github.com/dotnet/linker/pull/3073 + this change

### DIFF
--- a/src/linker/Linker.Steps/SweepStep.cs
+++ b/src/linker/Linker.Steps/SweepStep.cs
@@ -69,7 +69,7 @@ namespace Mono.Linker.Steps
 				case AssemblyAction.CopyUsed:
 				case AssemblyAction.Link:
 				case AssemblyAction.Save:
-					bool changed = AssemblyReferencesCorrector.SweepAssemblyReferences (assembly);
+					bool changed = AssemblyReferencesCorrector.SweepAssemblyReferences (assembly, Context);
 					if (changed && action == AssemblyAction.CopyUsed)
 						Annotations.SetAction (assembly, AssemblyAction.Save);
 					break;
@@ -174,7 +174,7 @@ namespace Mono.Linker.Steps
 				AssemblyAction assemblyAction = AssemblyAction.Copy;
 				if (SweepTypeForwarders (assembly)) {
 					// Need to sweep references, in case sweeping type forwarders removed any
-					AssemblyReferencesCorrector.SweepAssemblyReferences (assembly);
+					AssemblyReferencesCorrector.SweepAssemblyReferences (assembly, Context);
 					assemblyAction = AssemblyAction.Save;
 				}
 
@@ -194,7 +194,7 @@ namespace Mono.Linker.Steps
 			case AssemblyAction.Save:
 				if (SweepTypeForwarders (assembly)) {
 					// Need to sweep references, in case sweeping type forwarders removed any
-					AssemblyReferencesCorrector.SweepAssemblyReferences (assembly);
+					AssemblyReferencesCorrector.SweepAssemblyReferences (assembly, Context);
 				}
 				break;
 			}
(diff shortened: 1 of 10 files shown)


src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -40,6 +40,7 @@
 
 namespace Mono.Linker.Steps
 {
+	[System.Diagnostics.CodeAnalysis.SuppressMessage ("ApiDesign", "RS0030:Do not used banned APIs", Justification = "When sweeping, we may not want to resolve to a removed member that was cached previously.")]
```
marek-safar, line 43: Please use pragma instead

src/linker/Linker/MethodReferenceExtensions.cs
```diff
@@ -10,6 +10,7 @@ namespace Mono.Linker
 {
 	public static class MethodReferenceExtensions
 	{
+		[System.Diagnostics.CodeAnalysis.SuppressMessage ("ApiDesign", "RS0030:Do not used banned APIs", Justification = "MethodReference.Resolve is banned for performance reasons. GetDisplayName should be a really cold path and shouldn't require a context to call.")]
```
marek-safar, line 13: - Please use pragma
- Would it be too ugly to pass the context to GetDisplayName?

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -570,6 +571,7 @@ sealed class AssemblyReferencesCorrector : TypeReferenceWalker
 
 			bool changedAnyScopes;
 
+
```
marek-safar, line 574: Extra change

src/linker/Linker.Steps/SweepStep.cs
```diff
@@ -627,7 +627,7 @@ protected override void ProcessTypeReference (TypeReference type)
 
 			protected override void ProcessExportedType (ExportedType exportedType)
 			{
-				TypeDefinition td = exportedType.Resolve ();
+				TypeDefinition? td = exportedType.Resolve ();
```
marek-safar, line 630: Why is it not calling TryResolve?

jtschuster: Based off the information in https://github.com/dotnet/linker/issues/2260, I thought it would be best to use Cecil's Resolve here. In sweep step the metadata graph isn't stable, so we might resolve references to definitions that don't exist anymore. We don't have great testing around this, so I wanted to err on the side of caution and use the safer option for this. It doesn't seem to be a very hot path, so I think it would be okay for now.

src/linker/Linker/LinkContext.cs
```diff
@@ -908,6 +958,11 @@ protected virtual void ReportUnresolved (TypeReference typeReference)
 			if (unresolved_reported.Add (typeReference))
 				LogError (string.Format (SharedStrings.FailedToResolveTypeElementMessage, typeReference.GetDisplayName ()), (int) DiagnosticId.FailedToResolveMetadataElement);
 		}
+
+		protected virtual void ReportUnresolved (ExportedType et)
+		{
+			LogError (string.Format (SharedStrings.FailedToResolveTypeElementMessage, et.Name), (int) DiagnosticId.FailedToResolveMetadataElement);
```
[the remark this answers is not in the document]
jtschuster: Yes, thank you!

src/linker/Linker/MethodReferenceExtensions.cs
jtschuster: It might not necessarily be ugly, but I think it would take significant effort to bubble up the context parameter to GetDisplayName on all the SingleValue types and TypeSystemProxy types, and it's really only called from diagnostics which should be a very cold path anyway.

PR #3077, opened 2022-10-19
Change analyzer versions such that the repo can be built with .NET 7 RC2 SDK

vitek-karas:
Took the versions from dotnet/runtime.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -47,19 +47,11 @@ namespace ILLink.RoslynAnalyzer
 					if (methodSymbol.IsStaticConstructor () && methodSymbol.HasAttribute (RequiresAttributeName))
 						ReportRequiresOnStaticCtorDiagnostic (symbolAnalysisContext, methodSymbol);
 					CheckMatchingAttributesInOverrides (symbolAnalysisContext, methodSymbol);
-					CheckAttributeInstantiation (symbolAnalysisContext, methodSymbol);
-					foreach (var typeParameter in methodSymbol.TypeParameters)
-						CheckAttributeInstantiation (symbolAnalysisContext, typeParameter);
-
 				}, SymbolKind.Method);
 
 				context.RegisterSymbolAction (symbolAnalysisContext => {
 					var typeSymbol = (INamedTypeSymbol) symbolAnalysisContext.Symbol;
 					CheckMatchingAttributesInInterfaces (symbolAnalysisContext, typeSymbol);
-					CheckAttributeInstantiation (symbolAnalysisContext, typeSymbol);
-					foreach (var typeParameter in typeSymbol.TypeParameters)
-						CheckAttributeInstantiation (symbolAnalysisContext, typeParameter);
-
 				}, SymbolKind.NamedType);
 
 
@@ -68,8 +60,6 @@ namespace ILLink.RoslynAnalyzer
 					if (AnalyzerDiagnosticTargets.HasFlag (DiagnosticTargets.Property)) {
 						CheckMatchingAttributesInOverrides (symbolAnalysisContext, propertySymbol);
 					}
-
-					CheckAttributeInstantiation (symbolAnalysisContext, propertySymbol);
 				}, SymbolKind.Property);
 
 				context.RegisterSymbolAction (symbolAnalysisContext => {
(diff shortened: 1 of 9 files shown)


tlakollo:
I was debugging some of the failures too for my runtime PR, after updating to MicrosoftCodeAnalysisVersion 4.5.* the roslyn analyzers will start generating duplicate warnings on attribute scenarios seems like the compilation now takes into account more stuff and creates callbacks for object creations that it didn't before. I though it was just going to be a change like removing the CheckAttributeInstantiation method but seems like it's more conversome than that

tlakollo:
Never mind, removing CheckAttributeInstantiation is the solution. I saw some tests failing after removing the method but it was because before CheckAttributeInstantiation was producing double warnings. With the current analysis, only one warning is produced (which is the right outcome).

tlakollo:
> @tlakollo would you mind porting the fixes from your branch to this PR?

Sure, working on it

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -205,21 +188,6 @@ public override void Initialize (AnalysisContext context)
 				foreach (var extraSymbolAction in ExtraSymbolActions)
 					context.RegisterSymbolAction (extraSymbolAction.Action, extraSymbolAction.SymbolKind);
 
-				void CheckAttributeInstantiation (
```
vitek-karas, line 208: Why is it OK to remove this?
Where is the place we perform this check now? (and why did it change with a new analyzer package version)

vitek-karas:
One observation on a machine with RC2 installed:
Clean clone of the repo
* When built with "build.cmd" following usage in VS will fail (exception during build)
* Building from command line with just `dotnet build illink.sln` will work and then VS also works
* Running all tests in that mode is `dotnet test illink.sln`

Once the repo is built with `build.cmd` (which will use RC1), trying to build with RC2 may fail (and in VS it does).
If the repo is built with RC2 `dotnet build illink.sln` then building with `build.cmd` may work (it did for me), but I would suspect things will be weird eventually.

So unfortunately this means that if you need to use VS - don't use `build.cmd`. And make sure you clean the clone (`git clean -xdf`) before building for the first time.

I'm guessing that this is due to some kind of breaking change between RC1 and RC2.

The problems should go away once we switch the SDK to RC2 in the global.json (but we have to wait for dotnet/runtime to do that first).

src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
tlakollo: There is new callback calls in attributes, for at least one scenario (I didn't test all scenarios) instead of going through a type argument it went through an object creation, basically when it sees the use of the attribute it created a callback for object creation and then we have a CheckMemberCall that verifies that it has requires producing the warning. My guess is that it was a limitation of the analyzer that we had to work around, just like the global attribute and at somepoint it got fixed

vitek-karas: OK - I guess if all the tests are passing, it's at least not completely broken

vitek-karas:
`lint` is even worse - so far I didn't find a way to run on the RC2 bits. It works when the repo is built with RC1.

Does anybody know where we get the `dotnet format` tool from? I can't find anything in the repo which says what version of that tool to use.

vitek-karas:
Unfortunately the RC2 dotnet format fails with:
```console
Could not load file or assembly 'Microsoft.CodeAnalysis, Version=4.5.0.0,
```

So it seems that the only way to run linter is to clean the enlistment and use the `lint.cmd` which will download RC1. And then clean it again and rebuild to run VS.

vitek-karas:
I don't get it - RC1 ships format with Microsoft.CodeAnalysis 4.3.0.0 - and it works
RC2 ships format with Microsoft.CodeAnalysis 4.4.0.0 - and it doesn't work
I give up - let's merge this - so that at least there's some way to use VS.

tlakollo:
Notice that runtime has this same behavior, you cannot run dotnet format at all. I've been changing the versions.props file to be able to run the formater for the linker to runtime move PR. In the runtime repo it fails with "Could not load file or assembly 'Microsoft.CodeAnalysis.Workspaces, Version=4.4.0.0" which Im guessing is part of the Microsoft.CodeAnalysis

PR #3085, opened 2022-10-25
Enable compiler-generated code tests for NativeAot

sbomer:
Making the test changes in the linker repo first, to facilitate code sharing. Part of https://github.com/dotnet/runtime/issues/68786.

### DIFF
--- a/test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
+++ b/test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
@@ -13,9 +13,6 @@ namespace Mono.Linker.Tests.Cases.RequiresCapability
 {
 	[SkipKeptItemsValidation]
 	[ExpectedNoWarnings]
-#if NATIVEAOT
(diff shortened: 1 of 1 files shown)


tlakollo (approved):

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresInCompilerGeneratedCode.cs
```diff
@@ -2009,8 +2111,9 @@ static void TestAll ()
 			}
 
 			// Warnings for Reflection access to methods with Requires
-			[ExpectedWarning ("IL2026", "--TestLambdaInMethodWithRequires--")]
-			[ExpectedWarning ("IL2026", "--TestLambdaWithClosureInMethodWithRequires--")]
+
+			[ExpectedWarning ("IL2026", "--TestLambdaInMethodWithRequires--", ProducedBy = ProducedBy.Trimmer | ProducedBy.NativeAot)]
```
tlakollo, line 2115: Test are complaining that this used to include analyzer and analyzer warns, so there should be no change from the original. Now it would be Trimmer|Analyzer|NativeAot which is the default

PR #3088, opened 2022-10-27
Fix branch removal in compiler generated code

vitek-karas:
Changes to processing of compiler generated methods lead to a state where we don't call constant prop and branch removal in all cases before we mark instructions of the method. This can lead to overmarking

This change fixes this by making sure that the branch removal executes on the method in all cases before we mark instructions of the method.

To make the code simple I added a hashset to prevent multiple optimization runs on the same method. Technically this can be guaranteed by the caller, but it's really complex and very fragile.

Added tests.

Note that there's still a whole in analysis of compiler generated code around state machines, see https://github.com/dotnet/linker/issues/3087


src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3543,6 +3543,10 @@ bool MarkAndCheckRequiresReflectionMethodBodyScanner (MethodBody body)
 			if (_compilerGeneratedMethodRequiresScanner.TryGetValue (body, out bool requiresReflectionMethodBodyScanner))
 				return requiresReflectionMethodBodyScanner;
 
+			// Make sure the method's body is processed, for compiler generated code we can get here before ProcessMethod is called
+			// and thus we need to make sure we operate on the optimized method body (to avoid marking code which is otherwise optimized away).
+			UnreachableBlocksOptimizer.ProcessMethod (body.Method);
```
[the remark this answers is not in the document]
vitek-karas: I thought about it, but it felt dangerous. I looked into this some more, and we have more bugs like this:

For example, `ProcessAnalysisAnnotationsForMethod` (which is called from `MarkMethod(MethodReference`, so way before any body processing is done on the method) calls `ShouldWarnForReflectionAccessToCompilerGeneratedCode` which does instructions scanning and may produce warnings if it sees references to annotated fields/methods.

If the method in question contains a feature switch condition, we won't avoid the warning, even thought we will eventually remove the code in question.

Basically this leads me to a bigger change:
* MarkStep should never access Method.Body directly - since we can't guarantee it's the right one.
* We will have to store some hashset of already processed bodies (basically what this change does), to avoid re-analyzing method bodies for const. prop and branch removal.
* Ever place where we access Body from the MarkStep needs to go through a helper which will trigger the const. prop and branch removal first.

The good news is that the UnreachableBlocksOptimizer is already written such that it can be called from pretty much anywhere and it will make sure to process things in the right order internally.

vitek-karas:
I reworked this change to guard against accesses to Body and make sure that const. prop/branch removal happens always.

This does have one negative effect: https://github.com/dotnet/linker/issues/2937 is not consistent and happens in all cases - meaning if the callsite to a local function is removed, we can't figure out to which method the local function belongs and may incorrectly report some warnings.

Personally I prefer the consistent behavior (which is not order dependent anymore) - even if it's incorrect. We can use https://github.com/dotnet/linker/issues/2937 to track ideas on how to fix the problem.

src/linker/Linker.Steps/MethodBodyInstructionsProvider.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Mono.Collections.Generic;
+
+namespace Mono.Linker.Steps
+{
+	public class MethodBodyInstructionsProvider
```
vitek-karas, line 10: Looking for naming ideas - so consider this a temporary name.

src/linker/Linker.Steps/MethodBodyInstructionsProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+	{
+		readonly UnreachableBlocksOptimizer _unreachableBlocksOptimizer;
+
+		public readonly struct ProcessedMethodBody
```
vitek-karas, line 14: This is also something I'd like to get feedback on. Basically in the MarkStep/MethodBodyScanner there's a lot of code which accesses instructions but all of it happens in code paths which are guaranteed to have the method correctly processed already. I didn't want to add bunch of warning suppressions all of the codebase as it's messy and hard to prove that it's correct. Instead I introduced this wrapper struct which represented a method which is "Safe" to access instructions on. This means the struct is now passed around quite a bit.

So mainly looking for two things:
* Opinions about this solution + alternative ideas
* Naming suggestions/code factoring - even if we keep this, maybe we should rename/move it somehow

vitek-karas:
I measured perf of this change (using the sfx project from runtime) and there's no visible regression (although locally the noise is rather high).

jtschuster (approved):
LGTM, thank you!

src/linker/Linker.Dataflow/MethodBodyScanner.cs
```diff
@@ -699,7 +702,7 @@ protected virtual void Scan (MethodBody methodBody, ref InterproceduralState int
 			}
 		}
 
-		private static void ScanExceptionInformation (Dictionary<int, Stack<StackSlot>> knownStacks, MethodBody methodBody)
+		private static void ScanExceptionInformation (Dictionary<int, Stack<StackSlot>> knownStacks, MethodBodyInstructionsProvider.ProcessedMethodBody methodBody)
```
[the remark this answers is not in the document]
vitek-karas: I don't think I did anything with static/instance methods on this class. The idea is that Scan gets the processed method from the outside and it passes it to all of the helpers here - including this one.
Sorry, I probably don't understand the question.

src/linker/Linker.Steps/MethodBodyInstructionsProvider.cs
vitek-karas: I intentionally create a "layer" between the optimizer and the users - in short the users should not care what optimizations are done on the IL.
And I also don't think the optimizer should care about how the users handle the synchronization (as in, the need to run the optimizer before they access the method's body).
So this class acts as the separation layer.

What I'll do is move the hashset into here - as again the optimizer should not really care about being reentrant.

I must admit I don't like ILProvider too much, but I guess it's better than the current name.

src/linker/Linker.Steps/MethodBodyInstructionsProvider.cs
vitek-karas: `MethodIL` is not ideal - but it's probably better than "processed". While I agree "processed" is more accurate, the consumers of this "don't care".

src/linker/Linker.Steps/MethodBodyInstructionsProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+		{
+			_unreachableBlocksOptimizer = new UnreachableBlocksOptimizer(context);
+		}
+
+		public ProcessedMethodBody GetMethodBody (MethodBody methodBody)
+			=> GetMethodBody (methodBody.Method);
+
+		public ProcessedMethodBody GetMethodBody (MethodDefinition method)
+		{
+			_unreachableBlocksOptimizer.ProcessMethod (method);
```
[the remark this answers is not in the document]
vitek-karas: As noted above, I like the separation between the optimizer and the rest of the system.

src/linker/Linker.Steps/MethodBodyInstructionsProvider.cs
```diff
@@ -0,0 +1,52 @@
...
+
+		public MethodBodyInstructionsProvider(LinkContext context)
+		{
+			_unreachableBlocksOptimizer = new UnreachableBlocksOptimizer(context);
+		}
+
+		public ProcessedMethodBody GetMethodBody (MethodBody methodBody)
+			=> GetMethodBody (methodBody.Method);
+
+		public ProcessedMethodBody GetMethodBody (MethodDefinition method)
```
marek-safar, line 45: I think it'd be much better if `GetMethodBody` existed only on LinkContext (_unreachableBlocksOptimizer would be instantiated there too). As a bonus, you could move caching from `UnreachableBlocksOptimizer` there too for consistency.

vitek-karas: That's a good idea - I moved it there.

src/linker/Linker.Steps/MethodBodyInstructionsProvider.cs
vitek-karas: With the move to `LinkContext` I also moved this. There is now `MethodIL` type in the `Mono.Linker` namespace - and the rest is on `LinkContext`.

src/linker/Linker/MethodIL.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Mono.Linker
+{
+	/// <summary>
+	/// This is a wrapper which should be used by anything accessing method's body - see LinkContext.GetMethodIL for more details.
+	/// Any accesses made throught this wrapper are considered "safe"/OK since the wrapper is only created
+	/// once all of the optimizations are applied.
+	/// </summary>
+	public readonly record struct MethodIL
+	{
+		public MethodIL () => throw new InvalidOperationException ();
```
marek-safar, line 18: Why is it needed?

src/linker/Linker/MethodIL.cs
```diff
@@ -0,0 +1,39 @@
...
+		MethodIL (MethodBody body) => this.Body = body;
+
+		public readonly MethodBody Body;
+
+		public MethodDefinition Method => Body.Method;
+
+#pragma warning disable RS0030 // Wrapper which provides safe access to the property
```
marek-safar, line 25: I'd just fold this into one pragma block

src/linker/Linker.Dataflow/CompilerGeneratedState.cs
```diff
@@ -145,7 +145,7 @@ void ProcessMethod (MethodDefinition method)
 				// Discover calls or references to lambdas or local functions. This includes
 				// calls to local functions, and lambda assignments (which use ldftn).
 				if (method.Body != null) {
-					foreach (var instruction in _context.MethodBodyInstructionsProvider.GetMethodBody (method.Body).Instructions) {
+					foreach (var instruction in _context.GetMethodIL (method.Body).Instructions) {
```
marek-safar, line 148: Wouldn't be better to replace all `GetMethodIL (method.Body)` calls with `GetMethodIL (method)`

src/linker/Linker/MethodIL.cs
vitek-karas: it's not needed now - it was there when the struct wasn't readonly in some early version...

PR #3089, opened 2022-10-27
Upgrade MSBuild ref and downgrade global.json

agocke:
The official build has an old version of MSBuild so we can't use a new version of MSBuild. However, we do need a higher version of MSBuild as our ref for the tasks, so this also upgrades the version of MSBuild.

The two changes are technically separate, even though they both involve MSBuild.


agocke:
Ah, yeah, I could do that. Right now I'm not sure it's necessary, but that will be my next solution if I run into more problems.

PR #3094, opened 2022-10-31
Check for marking virtual method due to base only when state changes

jtschuster:
Ports https://github.com/dotnet/linker/pull/3073 to the .net7 release.

Trimming speed regressed significantly from .net6 to .net7, as much as 20x slowdown in one reported case (https://github.com/dotnet/linker/issues/3083). 

Instead of checking every virtual method to see if it should be kept due to a base method every iteration of the MarkStep pipeline, check each method only when its relevant state has changed.

This also changes behavior when a type is marked as instantiated. Previously, we would mark all methods that override a method from a preserved scope (i.e. in an assembly that is not to be trimmed), which would mark methods implementing an interface even if the interface implementation isn't kept. Now, the method should only be kept if the interface is marked and the interface implementation is marked.

Co-authored-by: Sven Boemer <sbomer@gmail.com>

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -607,12 +607,12 @@ namespace Mono.Linker.Steps
 					}
 					// OverrideInformation for interfaces in PreservedScope aren't added yet
 					foreach (var method in type.Methods) {
-						var bases = Annotations.GetBaseMethods (method);
-						if (bases is null)
+						var baseOverrideInformations = Annotations.GetBaseMethods (method);
+						if (baseOverrideInformations is null)
 							continue;
-						foreach (var @base in bases) {
-							if (@base.DeclaringType.IsInterface && IgnoreScope (@base.DeclaringType.Scope))
-								_interfaceOverrides.Add ((new OverrideInformation (@base, method, Context), ScopeStack.CurrentScope));
+						foreach (var ov in baseOverrideInformations) {
+							if (ov.Base.DeclaringType is not null && ov.Base.DeclaringType.IsInterface && IgnoreScope (ov.Base.DeclaringType.Scope))
+								_interfaceOverrides.Add ((ov, ScopeStack.CurrentScope));
 						}
 					}
 				}
@@ -707,12 +707,6 @@ namespace Mono.Linker.Steps
 		{
 			Annotations.EnqueueVirtualMethod (method);
 
-			var overrides = Annotations.GetOverrides (method);
-			if (overrides != null) {
-				foreach (OverrideInformation @override in overrides)
-					ProcessOverride (@override);
-			}
-
 			var defaultImplementations = Annotations.GetDefaultInterfaceImplementations (method);
 			if (defaultImplementations != null) {
 				foreach (var defaultImplementationInfo in defaultImplementations) {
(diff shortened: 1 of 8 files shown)


vitek-karas:
Given the perf improvements shown above, I agree we should try to take this for servicing.

jtschuster:
I've verified locally that this works in the 7.0 runtime builds and runtime tests pass after using it.

jeffschwMSFT (approved):
approved.  please get a few code reviews (given the size), and we can take for consideration in 7.0.x

agocke:
@sbomer could you review this?

agocke:
@vitek-karas Could you review as well?

test/Mono.Linker.Tests.Cases/Attributes/TypeWithDynamicInterfaceCastableImplementationAttributeIsKept.cs
```diff
@@ -16,6 +16,8 @@ namespace Mono.Linker.Tests.Cases.Attributes
 	[KeptMemberInAssembly ("impl", "Mono.Linker.Tests.Cases.Attributes.Dependencies.IReferencedAssemblyImpl", "Foo()")]
 	[KeptInterfaceOnTypeInAssembly ("impl", "Mono.Linker.Tests.Cases.Attributes.Dependencies.IReferencedAssemblyImpl",
 		"interface", "Mono.Linker.Tests.Cases.Attributes.Dependencies.IReferencedAssembly")]
+	[SetupLinkerTrimMode ("link")]
+	[IgnoreDescriptors (false)]
```
vitek-karas, line 20: I would expect these to be the defaults - do we need to specify it?

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/BaseProvidesInterfaceEdgeCase/BaseProvidesInterfaceMethodCircularReference.cs
```diff
@@ -0,0 +1,38 @@
...
+		{
+			_ = new Derived1 ();
+			Foo ();
+		}
+
+		[Kept]
+		public static void Foo ()
+		{
+			((IFoo) null).Method ();
+			object x = null;
```
vitek-karas, line 35: Why do we need this line `object x = null` - it doesn't seem related in any way to the rest of the code here.

test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/BaseProvidesInterfaceEdgeCase/BaseProvidesInterfaceMethodCircularReference.cs
```diff
@@ -0,0 +1,38 @@
...
+	/// creating a psuedo-circular assembly reference (but not quite since <see cref="Base"/> doesn't implement IFoo itself).
+	/// In the linker, IsMethodNeededByInstantiatedTypeDueToPreservedScope would iterate through <see cref="Base"/>'s method's base methods,
+	/// and in the process would trigger the assembly of <see cref="IFoo"/> to be processed. Since that assembly also has <see cref="Derived2"/> that 
+	/// inherits from <see cref="Base"/> and implements <see cref="IBar"/> using <see cref="Base"/>'s methods, the linker adds 
+	/// <see cref="IBar"/>'s method as a base to <see cref="Base"/>'s method, which modifies the collection as it's being iterated, causing an exception.
+	/// </summary>
+	[SetupCompileBefore ("base.dll", new[] { "Dependencies/Base.cs" })] // Base Implements IFoo.Method (psuedo-reference to ifoo.dll)
+	[SetupCompileBefore ("ifoo.dll", new[] { "Dependencies/IFoo.cs" }, references: new[] { "base.dll" })] // Derived2 references base.dll (circular reference)
+	[SetupCompileBefore ("derived1.dll", new[] { "Dependencies/Derived1.cs" }, references: new[] { "ifoo.dll", "base.dll" })]
+	public class BaseProvidesInterfaceMethodCircularReference
```
vitek-karas, line 22: Would be good to check that we kept the Base.Method (as the actual implementation method)
And also that we removed Derived2 since it's not used by anything (even though it affects the linker's internal structures)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -721,43 +715,65 @@ void ProcessVirtualMethod (MethodDefinition method)
...
 			}
 
-			// Interface static veitual methods will be abstract and will also by pass this check to get marked
-			if (!isInstantiated && !@base.IsAbstract && Context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, method))
-				return;
+			if (!Context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, overrideInformation.Override))
+				return true;
+
+			// Methods on instantiated types that override a ov.Override from a base type (not an interface) should be marked
+			// Interface ov.Overrides should only be marked if the interfaceImplementation is marked, which is handled below
```
vitek-karas, line 737: This explanation is... weird. I honestly couldn't figure it out.

What does it mean "method which overrides something from a base type? Or maybe I'm reading it wrong.
I think what this is trying to say is something like:
"An override needs to be kept if - a) it's an override on an instantiated type (of a marked base) or b) it's an override of an abstract base (where base is marked, and thus the override must be there to produce valid IL - we also alread operate only on overrides which type is marked)

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -721,43 +715,65 @@ void ProcessVirtualMethod (MethodDefinition method)
...
+
+			return false;
+		}
 
-			// Only track instantiations if override removal is enabled and the type is instantiated.
-			// If it's disabled, all overrides are kept, so there's no instantiation site to blame.
-			if (Context.IsOptimizationEnabled (CodeOptimizations.OverrideRemoval, method) && isInstantiated) {
-				MarkMethod (method, new DependencyInfo (DependencyKind.OverrideOnInstantiatedType, method.DeclaringType), ScopeStack.CurrentScope.Origin);
+		/// <summary>
+		/// Marks the Override of <paramref name="overrideInformation"/> with the correct reason. Should be called when <see cref="ShouldMarkOverrideForBase(OverrideInformation, bool)"/> returns true.
```
vitek-karas, line 750: If it's a prereq that ShouldMarkOverrideForBase returns true, then maybe this should assert it?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -721,43 +715,65 @@ void ProcessVirtualMethod (MethodDefinition method)
...
+			Debug.Assert (Annotations.IsMarked (method.DeclaringType));
+
+			var bases = Annotations.GetBaseMethods (method);
+			if (bases is null)
+				return;
+
+			var markedBaseMethods = bases.Where (ov => Annotations.IsMarked (ov.Base) || IgnoreScope (ov.Base.DeclaringType.Scope));
+			foreach (var ov in markedBaseMethods) {
+				if (ShouldMarkOverrideForBase (ov))
+					MarkOverrideForBaseMethod (ov);
```
vitek-karas, line 775: There are only two places where we use this - and they're both in the form "if (Should ...) Mark...." - It would feel cleaner if we had just one method "MarkOverrideForBaseAsNecessary" (or something like that), which does the checks and marking in one go (would also avoid the current assumption that Mark is only called when Should returned true).

But maybe this is solved after we fix 3090

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2027,6 +2043,10 @@ internal void MarkStaticConstructorVisibleToReflection (TypeDefinition type, in
 				_typesWithInterfaces.Add ((type, ScopeStack.CurrentScope));
 
 			if (type.HasMethods) {
+				// TODO: MarkMethodIfNeededByBaseMethod should include logic for IsMethodNeededBytTypeDueToPreservedScope
```
vitek-karas, line 2046: Link to a bug - probably 3090?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -2027,6 +2043,10 @@ internal void MarkStaticConstructorVisibleToReflection (TypeDefinition type, in
...
+				foreach (var method in type.Methods) {
+					MarkMethodIfNeededByBaseMethod (method);
+				}
 				// For methods that must be preserved, blame the declaring type.
 				MarkMethodsIf (type.Methods, IsMethodNeededByTypeDueToPreservedScope, new DependencyInfo (DependencyKind.VirtualNeededDueToPreservedScope, type), ScopeStack.CurrentScope.Origin);
```
vitek-karas, line 2051: This is basically a second "foreach ( ... in type.Methods)" - would look nicer if we merged them into one foreach.

test/Mono.Linker.Tests.Cases/Attributes/TypeWithDynamicInterfaceCastableImplementationAttributeIsKept.cs
jtschuster: Yes, the defaults are to ignore descriptor files and "skip" trim mode for references.

vitek-karas (approved):
The change looks good for servicing.

Sorry I used this as a way to suggest lot of cleanup, which we should not do in servicing. But please consider the comments for future PR into main.

vitek-karas:
This actually fixes a functional issue as well - recursive generics cause stackoverflow without this change:
https://github.com/dotnet/linker/issues/3155

PR #3095, opened 2022-10-31
Get rid of warnings about old version of NuGet.Protocol

agocke:
The analyzer testing infrastructure uses an old version of NuGet.Protocol, which is generating warnings. This upgrades to a newer version.


akoeplinger:
@agocke did this make it to main/7.0? I see the same warnings about NuGet.Protocol there. Besides that, according to nuget.org the 5.8.0 version has a vulnerability as well so we should probably upgrade to 5.9.3: https://www.nuget.org/packages/NuGet.Protocol/5.8.0

PR #3100, opened 2022-11-03
Increase build versions again

agocke:
Looks like the new builds weren't quite new enough. I tracked down the MSBuild reference and it looks like it's a transitive dependency on netstandard2.0 that's the problem. 17.2.2 should remove that dependency and still be supported for all net6.0 platforms.


agocke:
Assuming this works (finally) some of this will need to be brought into main as well.

PR #3103, opened 2022-11-08
Add feature attributes design

sbomer:
This is an attempt to specify the behavior of the various `Requires`* attributes we have today. Unlike https://github.com/dotnet/linker/pull/2177, this is intended to specify the allowed and disallowed semantics for a generalized feature attribute, not the specific formatting of the warnings. I think https://github.com/dotnet/linker/pull/2177 would be helpful in addition to this spec to give more concrete examples, but I want this to be a concise summary of the rules.

See the motivation section for details.

### DIFF
--- a/docs/design/feature-attribute-semantics.md
+++ b/docs/design/feature-attribute-semantics.md
@@ -51,15 +51,18 @@ Access to members with a _feature requirement_ is always allowed from a _feature
 
 ## Feature available scope
 
-Methods and constructors (including static constructors) with a _feature requirement_ are in a _feature available_ scope.
+The following constructs with a _feature requirement_ are also in a _feature available_ scope:
+- Methods
+- Constructors (including static constructors)
 
-Methods, constructors (including static constructors), and fields declared in a class or struct with a _feature requirement_ are also in a _feature available_ scope.
+When a class or struct has a _feature requirement_, the following members are in a _feature available_ scope:
+- Methods
+- Constructors (including static constructors)
+- Fields
+- Properties
+- Events
 
-Properties and events declared in a class or struct with a _feature requirement_ are also in a _feature available_ scope.
-
-Lambdas and local functions with a _feature requirement_ are also in a _feature available_ scope.
-
-Note that the _feature available_ scope for a type does not extend to nested types.
+Note that the _feature available_ scope for a type does not extend to nested types or to members of base classes or interfaces implemented by the type.
 
 ## Feature requirement
 


docs/design/feature-attributes.md
```diff
@@ -0,0 +1,181 @@
...
+
+`RequiresFeature` may be used on methods, constructors, or classes only.
+
+The use of this attribute establishes a [_feature requirement_](#feature-requirement) for the attributed type or member, which restricts access to the attributed type or member (and in some cases to other related IL) in certain ways. It also establishes a [_feature available_](#feature-available-scope) scope (which includes the attributed member but may also include other related IL) wherein access to members with a _feature requirement_ is allowed.
+
+Access to members with a _feature requirement_ is always allowed from a _feature available_ scope, and never produces feature warnings. The restrictions created by _feature requirement_ only limit access from scopes outside of _feature available_, where certain access patterns produce warnings.
+
+## Feature available scope
+
+Methods and constructors (including static constructors) with a _feature requirement_ are in a _feature available_ scope.
```
vitek-karas, line 35: I think we've decided we don't want to support these attributes on static constructors because it's too hard to determine when the static constructor is going to be called. That's why we introduced requirements on type scope, as that covers the static constructor as well (and has a much easier definition of access).

vitek-karas, line 35: I find this wording a bit confusing... but it's probably just me.

In this case the "insides" of the method are in the feature available scope - the method itself ... not really.
For example, method defined on a type which has feature requirement (the type has it) IS in a feature available scope - as a whole.

But maybe it's not worth making the distinction.

docs/design/feature-attributes.md
```diff
@@ -0,0 +1,181 @@
...
+
+Methods, constructors (including static constructors), and fields declared in a class or struct with a _feature requirement_ are also in a _feature available_ scope.
+
+Properties and events declared in a class or struct with a _feature requirement_ are also in a _feature available_ scope.
+
+Lambdas and local functions inside of a method in a _feature available_ scope are also in a _feature available_ scope.
+
+Note that nested types declared in a type that is in a _feature available_ scope are not necessarily in a _feature available_ scope.
```
vitek-karas, line 43: "are not necessarily" - this doesn't sound like a rule :wink:. 
Are they or are they not? (I think we said that they are.... right?)

docs/design/feature-attributes.md
```diff
@@ -0,0 +1,181 @@
...
+
+When `RequiresFeature` is used on a method or constructor (except static constructors), this declares a _feature requirement_ for the method.
+
+Static constructors never have a _feature requirement_. `RequiresFeature` on a static constructor is not supported.
+
+### Classes
+
+When `RequiresFeature` is used on a class, this declares a _feature requirement_ for the class.
+
+When a class has a _feature requirement_, this creates a _feature requirement_ for the following members of the class:
```
vitek-karas, line 57: I think it's worth mentioning that this list IS the definition of "feature requirement on the class", there's no more to it. Specifically - accessing the type itself does not create a feature requirement - that is `typeof(Foo)` does not warn if `Foo` has a `RUC` on it.

vitek-karas: Also maybe of note that it makes no requirements/effect on the base class and interfaces.

docs/design/feature-attributes.md
```diff
@@ -0,0 +1,181 @@
...
+
+For analyzers which don't have visibility into the compiler-generated code for nested functions, nested functions declared in a method which is in a _feature available_ scope have a _feature requirement_.
+
+Note that a lambda or local function inherits _feature requirement_ from the enclosing user method, not from an enclosing lambda or local function if one is present.
+
+## Validation behavior
+
+### RequiresFeatureAttribute
+
+`RequiresFeatureAttribute` on a static constructor is disallowed.
```
vitek-karas, line 100: Above you use just `RequiresFeature` (without the "Attributes").... just seems inconsistent.

docs/design/feature-attributes.md
```diff
@@ -0,0 +1,181 @@
...
+            var some = new SomeFeatureProvider();
+            some.Use();
+        }
+        OtherFeatureProvider other = new();
+        other.Use();
+    }
+}
+```
+
+Perhaps we could introduce the simplified model as an optional strict mode for people who are interested in rewriting their code for maximal size savings.
```
vitek-karas, line 181: I think it depends on how popular requirements on type are going to be. I can't think of a similar optimization for requirements on members which so far are much more common.

I agree that this feels more like an "analyzer only" feature, as opposed to "hard requirement for the system".

docs/design/feature-attributes.md
```diff
@@ -0,0 +1,181 @@
+# Feature attributes
+
+This specification defines the semantics of "feature attributes", in terms of a hypothetical `RequiresFeature` attribute type. The rules described here are designed to be applicable to any attribute that describes feature or capability of the code or the platform.
+
+## Motivation
```
vitek-karas, line 5: I think another motivation should be code sharing - if the various attribute share the same behavior the analyzers can be shared as well, which helps a lot.

vitek-karas (approved):
Thanks a lot for writing this down!

The obvious next step would be to describe how the existing attribute differ from this model and what we can do to fix that.

jtschuster (commented):
Comments are from before your latest commit

docs/design/feature-attribute-semantics.md
```diff
@@ -0,0 +1,200 @@
...
+
+The advantage of the specified model is that it allows some references without warning, giving some extra flexibility and making it easier to migrate existing code. The downside is that it might lead to preserving more code, whereas a simplified model could guarantee that all code related to a disabled feature is removed.
+
+Here is an example of a pattern which does not warn in the current model, but would warn with a simplified model. Assume that the code under `SomeFeatureIsSupported` is removed when the feature is unavailable.
+
+```csharp
+class FeatureConsumer {
+    static void Run() {
+        SomeFeatureProvider? some;
+        if (Features.SomeFeatureIsSupported)
```
jtschuster, line 160: If the API to check for feature/capability availability is uniform, this could be completely removed, right? Then we would be able to use the simplified model and not have the extra warnings? Not that that's how it works right now, but it could be a possibility.

jtschuster (commented):
Great writeup, thank you!

I think the rules could be more precise if they are worded with "_feature X requirement_" and "_feature X available_ scope" to clarify that each feature requirement and scope is independent, but that's not really the most important thing right now.

docs/design/feature-attribute-semantics.md
jtschuster: Oh, I see, I missed that, thank you!

docs/design/feature-attributes.md
vitek-karas: I realized the signature problem as well - it actually extends to method attributes as well... so...
I think it's OK to stick with the current wording - if it's too difficult to understand we could add a longer description about the fact that it applies to signature and body...

vitek-karas (approved):
Nice work!

docs/design/feature-attribute-semantics.md
```diff
@@ -0,0 +1,203 @@
...
+
+## RequiresFeatureAttribute
+
+`RequiresFeatureAttribute` may be used on methods, constructors, or classes only.
+
+The use of this attribute establishes a [_feature requirement_](#feature-requirement) for the attributed type or member, which restricts access to the attributed type or member (and in some cases to other related IL) in certain ways. It also establishes a [_feature available_](#feature-available-scope) scope (which includes the attributed member but may also include other related IL) wherein access to members with a _feature requirement_ is allowed.
+
+Access to members with a _feature requirement_ is always allowed from a _feature available_ scope, and never produces feature warnings. The restrictions created by _feature requirement_ only limit access from scopes outside of _feature available_, where certain access patterns produce warnings.
+
+## Feature available scope
```
agocke, line 52: Content for this section looks good, but it's interesting that this is represented as bullets in the feature requirements section, but prose here.

For someone implementing the spec, you might want a more structured notation, i.e. the following constructs declared with a _feature requirement_ are also in a _feature available scope_: etc.

PR #3108, opened 2022-11-10
Warn on static constructor reflection access

sbomer:
Introduces a warning for reflection access to static constructors.

Fixes https://github.com/dotnet/linker/issues/3104

### DIFF
--- a/test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresOnStaticConstructor.cs
+++ b/test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresOnStaticConstructor.cs
@@ -4,13 +4,13 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.Runtime.CompilerServices;
 using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Text;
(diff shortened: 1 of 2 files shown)


MichalStrehovsky:
We also have an intrinsic around this in `RuntimeHelpers.RunClassConstructor`. I assume it just falls out but calling it out just in case it doesn't.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresOnStaticConstructor.cs
```diff
@@ -9,6 +9,7 @@
 using System.Threading.Tasks;
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 using Mono.Linker.Tests.Cases.Expectations.Helpers;
+using System.Reflection;
```
vitek-karas, line 12: Nit: Sort

vitek-karas (approved):

vitek-karas:
This does nothing about the `RuntimeHelpers.RunClassConstructor` per Michal's comment above.
We should look into that - if it needs an annotation or RUC or something else.

marek-safar:
Do module initializers have a similar problem?

vitek-karas (approved):

PR #3109, opened 2022-11-11
Fixes 3072 handling intrinsics on TypeInfo

vitek-karas:
The problem is that GetIntrinsicIdForMethod operates on MethodDefinition but some parts of HandleCall operate on MethodReference. If the calling code references for example TypeInfo.GetMethods intrinsics detection finds this as being the GetMethods intrinsics but HandleCall doesn't and falls through to the default case which should basically never happen.

The fix is to use MethodDefinition in both places.

Added a test which needed to be done in IL since the last time TypeInfo.GetMethods existed as a method (and not inherited from Type.GetMethods) was in netstandard1.2. So in the IL the call is made to a MethodReference TypeInfo.GetMethods which reproes the problem.

This is a servicing fix for https://github.com/dotnet/linker/issues/3072 into 7.0.


vitek-karas:
Once approved I'll port the tests over to main, where the bug doesn't repro since the affected code has been changed. But we still want the tests.

jeffschwMSFT (approved):
approved.  we will take for consideration in 7.0.x

agocke (approved):

PR #3113, opened 2022-11-14
Add more context for unexpected errors in optimizer step

marek-safar:


src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
```diff
@@ -47,28 +47,32 @@ public void ProcessMethod (MethodDefinition method)
...
-			inliner.RewriteBody ();
+				//
+				// Note: The inliner cannot run before reducer rewrites body as it
+				// would require another recomputing offsets due to instructions replacement
+				// done by inliner
+				//
+				var inliner = new CallInliner (method.Body, this);
+				inliner.RewriteBody ();
+			} catch (Exception e) {
+				throw new InternalErrorException ($"Could not process method '{method.GetDisplayName ()}' body", e);
```
vitek-karas, line 74:
```suggestion
				throw new InternalErrorException ($"Could not process the body of method '{method.GetDisplayName ()}'.", e);
```

vitek-karas: I assume we rely on callstack to tell us that it came from this optimizer (otherwise it might be worth mentioning it in the message).

vitek-karas (approved):

src/linker/Linker.Steps/UnreachableBlocksOptimizer.cs
marek-safar: Yes, we do.

PR #3117, opened 2022-11-15
Tests for TypeInfo intrinsics

vitek-karas:
This is a port of the tests from https://github.com/dotnet/linker/pull/3109 (the fix is not needed since the code now works differently and doesn't have the problem)


test/Mono.Linker.Tests.Cases/DataFlow/Dependencies/TypeInfoCalls.il
```diff
@@ -0,0 +1,112 @@
+// Metadata version: v4.0.30319
+.assembly extern System.Runtime
+{
+  .publickeytoken = (B0 3F 5F 7F 11 D5 0A 3A )                         // .?_....:
+  .ver 7:0:0:0
```
marek-safar, line 5: nit: I think version is not needed

marek-safar (approved):

PR #3124, opened 2022-11-18
Make default rooting behavior consistent

sbomer:
Now specifying an assembly as a root without any other qualifiers will default to rooting everything. Fixes https://github.com/dotnet/linker/issues/1541

### DIFF
--- a/src/ILLink.Tasks/LinkTask.cs
+++ b/src/ILLink.Tasks/LinkTask.cs
@@ -43,11 +43,8 @@ namespace ILLink.Tasks
 		/// <summary>
 		///   The names of the assemblies to root. This should contain
 		///   assembly names without an extension, not file names or
-		///   paths. Exactly which parts of the assemblies get rooted
-		///   is subject to change. Currently these get passed to
-		///   illink with "-a", which roots the entry point for
-		///   executables, and everything for libraries. To control
-		///   the behaviour explicitly, set RootMode metadata.
+		///   paths. The default is to root everything in these assemblies.
+		//    For more fine-grained control, set RootMode metadata.
 		/// </summary>
 		[Required]
 		public ITaskItem[] RootAssemblyNames { get; set; }


src/linker/Linker/Driver.cs
```diff
@@ -1115,11 +1115,9 @@ static string[] ReadLines (string file)
 			return null;
 		}
 
-		AssemblyRootMode? ParseAssemblyRootsMode (string s)
+		AssemblyRootMode? ParseAssemblyRootMode (string s)
 		{
 			switch (s.ToLowerInvariant ()) {
-			case "default":
```
vitek-karas, line 1121: Is there any chance we get this on the command line?
Just thinking about backward compat.

vitek-karas (approved):

src/linker/Linker/Driver.cs
vitek-karas: I think it's OK to break this as well - especially since this is 8.0 only and we will have TFM-specific linker then.

src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
```diff
@@ -279,9 +279,9 @@ Copyright (c) .NET Foundation. All rights reserved.
       <ManagedAssemblyToLink Include="@(_TrimmableManagedAssemblyToLink)" IsTrimmable="true" />
     </ItemGroup>
 
-    <!-- Root the main assembly, whether or not it has IsTrimmable set. -->
+    <!-- Root the main assembly entry point. -->
     <ItemGroup>
-      <TrimmerRootAssembly Include="@(IntermediateAssembly)" />
+      <TrimmerRootAssembly Include="@(IntermediateAssembly)" RootMode="EntryPoint" />
```
jonathanpeppers, line 284: @sbomer for Android apps are actually class libraries, there is no `static void Main()` because Android's entry point is either an `Android.App.Application` or `Android.App.Activity`.

I think this change causes:
```xml
ILLink error IL1034: Root assembly 'MyAndroidApp, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' does not have entry point.
```

Should I do this in the Android workload?
```
<ItemGroup>
  <TrimmerRootAssembly Update="@(TrimmerRootAssembly)" RootMode="Library" />
</ItemGroup>
```

PR #3131, opened 2022-11-23
Root entire assembly for non-exe projects

sbomer:
After fixing the root behavior in https://github.com/dotnet/linker/pull/3124, the linker update https://github.com/dotnet/runtime/pull/78020 is failing to build wasm test projects because they `PublishTrimmed` non-exe test projects which don't have an entry point:
```
ILLink : error IL1034: Root assembly 'System.Runtime.Tests, Version=7.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' does not have entry point.
```

This isn't really a supported scenario so we could work around it in dotnet/runtime, but in this case I think it makes sense to put the logic in the official targets here. This will root the entire assembly for non-exe projects, which is what the "default" root mode did (for non-exes) before https://github.com/dotnet/linker/pull/3124.

### DIFF
--- a/src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
+++ b/src/ILLink.Tasks/build/Microsoft.NET.ILLink.targets
@@ -279,9 +279,11 @@ Copyright (c) .NET Foundation. All rights reserved.
       <ManagedAssemblyToLink Include="@(_TrimmableManagedAssemblyToLink)" IsTrimmable="true" />
     </ItemGroup>
 
-    <!-- Root the main assembly entry point. -->
     <ItemGroup>
-      <TrimmerRootAssembly Include="@(IntermediateAssembly)" RootMode="EntryPoint" />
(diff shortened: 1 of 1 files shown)


MichalStrehovsky:
In general it's preferable to do workarounds for unsupported things local to where we do the unsupported thing (so in the runtime repo in this case).

It is extra code to maintain and extra potential bugs. This diff would introduce a bug for OutputType=WinExe.

vitek-karas:
I agree with Michal - the SDK doesn't support trimming anything but apps right now. Adding support for trimming libraries is much more complicated and thus I would not even try. I would prefer this change to be in the runtime.

Also - if we were to support classlibs in the SDK this way, then the correct solution would be to do "library" rooting for the assembly, not "all". The wasm tests in runtime relied on `all` because they're tests, they don't want to be trimmed, but I would prefer it they made this requirement explicit.

PR #3132, opened 2022-11-24
Use new submodule URL for dotnet/cecil

akoeplinger:
GitHub does a redirect but it's better to be consistent.


marek-safar (approved):

premun:
@akoeplinger I'm quite happy I implemented support for this event in the VMR and I was thinking maybe it was never going to happen. Now let's see how they VMR sync goes

PR #3147, opened 2022-12-01
Warn on reflection access to instance methods on RUC types

sbomer:
Fixes the hole for methods in https://github.com/dotnet/linker/issues/3140. We don't warn for RUC attributes on methods in RUC types, but instead prevent reflection access from creating these attributes by warning on the reflection access to methods (including instance methods).

This leaves the attribute warnings for fields as-is - so instance fields will get attribute warnings unlike instance methods, because we don't warn on reflection access to instance fields in RUC types. This is inconsistent with the analyzer behavior, which doesn't warn for attributes in this case. I may try to address that in a follow-up.

### DIFF
--- a/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/linker/Linker.Steps/MarkStep.cs
@@ -3071,7 +3071,6 @@ namespace Mono.Linker.Steps
 			bool skipWarningsForOverride;
 			bool isReflectionAccessCoveredByRUC;
 			RequiresUnreferencedCodeAttribute? requiresUnreferencedCode;
-			// Attribute properties are marked as a result of marking attribute instances, not because 
 			if (dependencyKind == DependencyKind.AttributeProperty) {
 				// Property assignment in an attribute instance.
 				// This case is more like a direct method call than reflection, and should
--- a/src/linker/Linker/Annotations.cs
+++ b/src/linker/Linker/Annotations.cs
@@ -601,7 +601,7 @@ namespace Mono.Linker
 		/// Determines if method is within a declared RUC scope - this typically means that trim analysis
 		/// warnings should be suppressed in such a method.
 		/// </summary>
-		/// <remarks>Unlike <see cref="DoesMemberRequireUnreferencedCode(IMemberDefinition, out RequiresUnreferencedCodeAttribute?)"/>
+		/// <remarks>Unlike <see cref="DoesMethodRequireUnreferencedCode(IMemberDefinition, out RequiresUnreferencedCodeAttribute?)"/>
 		/// if a declaring type has RUC, all methods in that type are considered "in scope" of that RUC. So this includes also
 		/// instance methods (not just statics and .ctors).</remarks>
 		internal bool IsInRequiresUnreferencedCodeScope (MethodDefinition method, [NotNullWhen (true)] out RequiresUnreferencedCodeAttribute? attribute)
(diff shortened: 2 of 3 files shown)


test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresOnClass.cs
```diff
@@ -515,11 +515,22 @@ static void TestNameOfDoesntWarn ()
...
+			[ExpectedWarning ("IL2026", "DerivedWithRequiresOnType.Method()", ProducedBy = ProducedBy.Analyzer)]
 			[ExpectedWarning ("IL2026", "InterfaceWithoutRequires.Method(Int32)")]
 			[ExpectedWarning ("IL2026", "InterfaceWithoutRequires.Method()")]
 			[ExpectedWarning ("IL2026", "ImplementationWithRequiresOnType.Method()")]
+			// Linker skips warnings for interface overrides, assuming it is covered by RUC on the interface method.
+			[ExpectedWarning ("IL2026", "ImplementationWithRequiresOnType.Method(Int32)", ProducedBy = ProducedBy.Analyzer)]
+			// Linker incorrectly skips warnings for derived method, under the assumption that
+			// it will be covered by the base method. But in this case the base method
+			// is unannotated (and the mismatch produces no warning because the derived
+			// type has RUC).
```
vitek-karas, line 532: Do we need a bug for this?

vitek-karas (approved):

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3065,15 +3065,31 @@ void ProcessAnalysisAnnotationsForMethod (MethodDefinition method, DependencyKin
...
 				return;
 
-			// All override methods should have the same annotations as their base methods
-			// (else we will produce warning IL2046 or IL2092 or some other warning).
-			// When marking override methods via DynamicallyAccessedMembers, we should only issue a warning for the base method.
-			bool skipWarningsForOverride = dependencyKind == DependencyKind.DynamicallyAccessedMember && method.IsVirtual && Annotations.GetBaseMethods (method) != null;
+			bool skipWarningsForOverride;
+			bool isReflectionAccessCoveredByRUC;
+			RequiresUnreferencedCodeAttribute? requiresUnreferencedCode;
+			// Attribute properties are marked as a result of marking attribute instances, not because
```
jtschuster, line 3074: Is this comment incomplete?

src/linker/Linker.Steps/MarkStep.cs
```diff
@@ -3065,15 +3065,31 @@ void ProcessAnalysisAnnotationsForMethod (MethodDefinition method, DependencyKin
...
+				// Property assignment in an attribute instance.
+				// This case is more like a direct method call than reflection, and should
+				// be logically similar to what is done in ReflectionMethodBodyScanner for method calls.
+				skipWarningsForOverride = false;
+				isReflectionAccessCoveredByRUC = Annotations.DoesMethodRequireUnreferencedCode (method, out requiresUnreferencedCode);
+			} else {
+				// All override methods should have the same annotations as their base methods
+				// (else we will produce warning IL2046 or IL2092 or some other warning).
+				// When marking override methods via DynamicallyAccessedMembers, we should only issue a warning for the base method.
+				skipWarningsForOverride = dependencyKind == DependencyKind.DynamicallyAccessedMember && method.IsVirtual && Annotations.GetBaseMethods (method) != null;
```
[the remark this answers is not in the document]
jtschuster: Looks like this is commented on in the tests below. If we change this line to 
```suggestion
				skipWarningsForOverride = dependencyKind == DependencyKind.DynamicallyAccessedMember && method.IsVirtual && Annotations.GetBaseMethods (method)?.Any(m => !m.DeclaringType.IsInterface) == true;
```
Do the tests follow the expected behavior?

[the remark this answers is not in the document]
jtschuster: Tested this and it's a bit more complicated (as it always is) and should be in a separate fix later if it's not expected behavior.

test/Mono.Linker.Tests.Cases/RequiresCapability/RequiresOnClass.cs
```diff
@@ -544,10 +555,12 @@ static void TestDAMAccess ()
 				typeof (DerivedWithRequiresOnTypeOverBaseWithNoRequires).RequiresPublicMethods ();
```
jtschuster, line 555: Do we need to update these comments? Looks like at least BaseWithRequiresOnType is warning now.

src/linker/Linker/Annotations.cs
```diff
@@ -630,23 +631,27 @@ internal bool DoesMemberRequireUnreferencedCode (IMemberDefinition member, [NotN
 			};
 		}
 
-		internal bool ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode (ICustomAttributeProvider? originMember)
+		internal bool ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode (ICustomAttributeProvider? originMember, [NotNullWhen (true)] out RequiresUnreferencedCodeAttribute? attribute)
```
tlakollo, line 634: Now that we return an attribute in this method I think is the same as DoesMemberRequireUnreferencedCode, and given that you changed the calls to DoesMemberRequireUnreferencedCode to ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode I think now nothing calls DoesMemberRequireUnreferencedCode, DoesMethodRequireUnreferencedCode, and DoesFieldRequireUnreferencedCode. I'm fine with 'ShouldSuppressAnalysisWarningsForRequiresUnreferencedCode' method is more readable and correct but probably we should get rid of unused methods

PR #3151, opened 2022-12-06
Fix analyzer test generation for non-public Main

sbomer:
In https://github.com/dotnet/runtime/pull/79011 @tlakollo discovered that the ILLink analyzer testcase generator was generating extra testcases in dotnet/runtime, compared to linker. It turned out that in dotnet/linker, testcases with non-public Main methods were not getting generated, because the generator was getting a reference assembly for Mono.Linker.Tests.Cases.dll (which doesn't contain the non-public entry point), whereas dotnet/runtime has ref assembly generation turned off (so the analyzer was correctly discovering testcases based on the implementation assembly).

This fixes the issue by turning off ref assembly generation for all projects except those specifically designed to be ref assemblies.

The CompilerGeneratedCodeSubstitutions is disabled as in https://github.com/dotnet/runtime/pull/79011.


test/ILLink.RoslynAnalyzer.Tests/generated/ILLink.RoslynAnalyzer.Tests.Generator/ILLink.RoslynAnalyzer.Tests.TestCaseGenerator/Inheritance.Interfaces.StaticInterfaceMethodsTests.g.cs
```diff
@@ -8,7 +8,7 @@ public sealed partial class StaticInterfaceMethodsTests : LinkerTestBase
 	{
 
 		[Fact]
-		public Task BaseProvidesInterfaceMethod ()
+		public Task VarianceBasic ()
```
tlakollo, line 11: Do we know why this happens? I do see that the VarianceBasic had a private Main method and is fine that we now have the Fact being generated. But why the BaseProvidesInterfaceMethod disappeared? And why there are no other tests? inside the Inheritance.Interfaces.StaticInterfaceMethods folder I see 10 files.

tlakollo (approved):

PR #3156, opened 2022-12-09
Add a test which uncovered stack overflow in 7.0

vitek-karas:
Recurisve generics with interface marking annotation used to cause stackoverflow in the linker. The test now passes since the problem was fixed in https://github.com/dotnet/linker/pull/3073

This seems to be the test for https://github.com/dotnet/linker/issues/3155 - it produces the exact same stack frames and stack overflow when ran on 7.0.


vitek-karas:
I verified that with https://github.com/dotnet/linker/pull/3094 the test passes as well.

vitek-karas:
Once merged we should add this test to https://github.com/dotnet/linker/pull/3094 as well.

marek-safar (approved):

jtschuster (approved):
Thank you!

Issue #3157, opened 2022-12-10
ILLink : error IL1012 with macOS, .NET 7.0


nilaoda:
```
  Optimizing assemblies for size. This process might take a while.
ILLink : error IL1012: IL Trimmer has encountered an unexpected error. Please report the issue at https://github.com/dotnet/linker/issues [/Users/runner/work/N_m3u8DL-RE/N_m3u8DL-RE/src/N_m3u8DL-RE/N_m3u8DL-RE.csproj]
  Fatal error in IL Linker
  Unhandled exception. System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'target')
     at Mono.Cecil.Cil.ILProcessor.InsertAfter(Instruction target, Instruction instruction)
     at Mono.Linker.LinkerILProcessor.InsertAfter(Instruction target, Instruction instruction)
     at Mono.Linker.Steps.MarkStep.ProcessPendingTypeChecks()
     at Mono.Linker.Steps.MarkStep.Process()
     at Mono.Linker.Steps.MarkStep.Process(LinkContext context)
     at Mono.Linker.Pipeline.ProcessStep(LinkContext context, IStep step)
     at Mono.Linker.Pipeline.Process(LinkContext context)
     at Mono.Linker.Driver.Run(ILogger customLogger)
     at Mono.Linker.Driver.Main(String[] args)
```
Detail here:  
https://github.com/nilaoda/N_m3u8DL-RE/actions/runs/3663371272/jobs/6193112246

vitek-karas:
This was fixed by https://github.com/dotnet/linker/pull/3088 but that fix is only available in .NET 8.0 SDK.
Unfortunately the fix is rather large so it raises the risk of taking the fix for servicing. I'll look into it.

In theory there might be a workaround as the bug is order dependent - if the trimmer sees the code in different order it might work, but it's really hard to guess which order will work and which won't.

vitek-karas:
https://github.com/dotnet/linker/pull/3161 is the 7.0 port of the fix.

vitek-karas:
Note that this issue is only the trimmer, it doesn't apply to NativeAOT (which also does triming), so if you manage to get your app working in NativeAOT, this issue should not impact that.

PR #3161, opened 2022-12-13
[release/7.0] Fix branch removal in compiler generated code

vitek-karas:
7.0 port of #3088

## Customer impact
In the worst case this case cause the trimmer to crash, preventing the app from being trimmed. In other cases this may lead to incorrect warnings or trimming behavior. The bug is order dependent so it's unpredictable (even though it's deterministic), while workaround may be possible by changing order of things, it would be very fragile.
See reported issue: https://github.com/dotnet/linker/issues/3157

## Risk
While the change affects lot of places in the trimmer, it's mostly mechanical (see below for the technical description of the change). The main idea is to prevent access to Cecil's instructions list directly and always go through a helper which makes sure that all necessary modifications to the instructions of the method were applied first. As such the risk is relatively low, since it doesn't change how most things work, it only changes order - making sure that body modifications happen before the body is inspected for marking.

## Testing
This change added targeted tests which reproed the problem and they cover several variations of the problem. The change has been in main (8.0) for more than a month without any known issues.
I validated that this change fixes the problem reported in #3157 

## Technical details about the change
Changes to processing of compiler generated methods lead to a state where we don't call constant prop and branch removal in all cases before we mark instructions of the method. This can lead to overmarking

This change fixes this by making sure that the branch removal executes on the method in all cases before we mark instructions of the method.

The change guarantees that all accesses to Body are after the constant prop/branch removal happened on the method.

This does have one possibly negative impact: the issue described in https://github.com/dotnet/linker/issues/2937 is now consistent and happens always.

Added tests.

Note that there's still a whole in analysis of compiler generated code around state machines, see https://github.com/dotnet/linker/issues/3087

Basically if there's a local function which is going to be removed due to branch removal and if the body of that method contains code which produces a warning due to generic parameter validation, such warning will always be generated even though it's "dead" code and even if it's suppressed via RUC or similar.

In such case the analysis can't figure out to which method the local function belongs (since the call site has been removed).

### DIFF
--- a/src/linker/Linker.Dataflow/InterproceduralState.cs
+++ b/src/linker/Linker.Dataflow/InterproceduralState.cs
@@ -14,22 +14,24 @@ using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.Single
 
 namespace Mono.Linker.Dataflow
 {
-	// Wrapper that implements IEquatable for MethodBody.
-	readonly record struct MethodBodyValue (MethodBody MethodBody);
-
 	// Tracks the set of methods which get analyzer together during interprocedural analysis,
 	// and the possible states of hoisted locals in state machine methods and lambdas/local functions.
 	struct InterproceduralState : IEquatable<InterproceduralState>
 	{
-		public ValueSet<MethodBodyValue> MethodBodies;
+		public ValueSet<MethodIL> MethodBodies;
 		public HoistedLocalState HoistedLocals;
 		readonly InterproceduralStateLattice lattice;
 
-		public InterproceduralState (ValueSet<MethodBodyValue> methodBodies, HoistedLocalState hoistedLocals, InterproceduralStateLattice lattice)
+		public InterproceduralState (ValueSet<MethodIL> methodBodies, HoistedLocalState hoistedLocals, InterproceduralStateLattice lattice)
 			=> (MethodBodies, HoistedLocals, this.lattice) = (methodBodies, hoistedLocals, lattice);
 
 		public bool Equals (InterproceduralState other)
 			=> MethodBodies.Equals (other.MethodBodies) && HoistedLocals.Equals (other.HoistedLocals);
+
+		public override bool Equals (object? obj)
+			=> obj is InterproceduralState state && Equals (state);
+
+		public override int GetHashCode () => base.GetHashCode ();
 
 		public InterproceduralState Clone ()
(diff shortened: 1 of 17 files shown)


marek-safar:
Needs tactics approval

src/linker/Linker.Dataflow/InterproceduralState.cs
```diff
@@ -14,23 +14,25 @@
...
+		public InterproceduralState (ValueSet<MethodIL> methodBodies, HoistedLocalState hoistedLocals, InterproceduralStateLattice lattice)
 			=> (MethodBodies, HoistedLocals, this.lattice) = (methodBodies, hoistedLocals, lattice);
 
 		public bool Equals (InterproceduralState other)
 			=> MethodBodies.Equals (other.MethodBodies) && HoistedLocals.Equals (other.HoistedLocals);
 
+		public override bool Equals (object? obj)
+			=> obj is InterproceduralState state && Equals (state);
+
+		public override int GetHashCode () => base.GetHashCode ();
```
jtschuster, line 34: I'm not super familiar with the default GetHashCode implementation, but will this cause a different hash code for objects that would be equal by Equals() but are different instances?
And if so, should this be `HashUtils.Combine(MethodBodies.GetHashCode(), HoistedLocals.GetHashCode())`?

vitek-karas: Hmm - that's a good point. Now I don't think this is a real problem because we don't use this struct as dictionary key anywhere, but you're right, the hash code is wrong.

src/linker/Linker/MethodIL.cs
```diff
@@ -0,0 +1,31 @@
...
+
+		public MethodDefinition Method => Body.Method;
+
+#pragma warning disable RS0030 // Wrapper which provides safe access to the property
+		public Collection<Instruction> Instructions => Body.Instructions;
+		public Collection<ExceptionHandler> ExceptionHandlers => Body.ExceptionHandlers;
+		public Collection<VariableDefinition> Variables => Body.Variables;
+#pragma warning restore RS0030
+
+		public static MethodIL Create (MethodBody body) => new MethodIL (body);
```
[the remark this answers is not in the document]
vitek-karas: Actually this method is listed in the banned APIs - so we will warn if you try to call this directly.
I ran into this exact problem, so I "fixed" it via a banned API.

PR #3167, opened 2022-12-21
Fix mapping of redeclared type parameters

sbomer:
We were skipping the type parameter mapping of compiler-generated types whose only generic parameters were those implicitly created for the declaring type's type parameters.

For the testcase in question, the nested state machine inherited a generic parameter from the display class. This was causing unnecessary warnings in a field assignment that assigned `this` (an instance of the display class) to a field on the state machine. In IL, that assignment references a field type like `DisplayClass<T>` where `T` is the generic parameter on the state machine. Here we were properly mapping type parameters of the display class back to the annotated enclosing method's type parameters, so we could tell that the "target" required `PublicMethods`. But the substituted `T` from the state machine was not mapped, causing the mismatch.

Fixes the issue mentioned in https://github.com/dotnet/roslyn/issues/46646#issuecomment-1339419988.


marek-safar:
@sbomer what is the plan with this PR?

sbomer:
Moved to https://github.com/dotnet/runtime/pull/86353

PR #3179, opened 2023-01-09
Fix new warning which shows up with latest roslyn

vitek-karas:
First encountered in https://github.com/dotnet/runtime/pull/80235, so copying the change to dotnet/linker


tlakollo (approved):
LGTM

PR #3185, opened 2023-01-13
Fix race condition in build

sbomer:
This is an attempt to fix a race condition in our build, where Mono.Cecil was being built twice into the same output path. See https://github.com/dotnet/installer/pull/15228#issuecomment-1381987801 for context.

Before this change, I saw this line twice in the output:
```
Mono.Cecil -> C:\Users\svbomer\src\linker\artifacts\bin\Mono.Cecil\Debug\netstandard2.0\Mono.Cecil.dll
```

After this change, I only see it once. I don't understand why `SetTargetFramework` was needed in the first place in https://github.com/dotnet/linker/pull/1891#discussion_r593494437, but it no longer seems necessary. Probably the recent changes to the cecil build (arcade onboarding) caused this to become a problem.

I have been able to build locally >100 times in a row with this change, without hitting the race, so I think this fixes it. Previously I was seeing it fail about one in ten builds locally.


marek-safar:
I don't think this change is fixing Mono.Cecil repeated builds, but it does not seem to hurt either.

marek-safar (approved):

ViktorHofer (approved):
Looks good. SetTargetFramework shouldn't be set when the referenced project doesn't multi-target as otherwise you could run into double evaluation and double builds. Great find!

PR #3186, opened 2023-01-16
Fix a small bug in message origin for warnings from static field access on generic type

vitek-karas:
The warning needs to be generated from the origin where the field reference comes from - so if it's a method body, then from the method body. This is because the instantiation is determined by that place, not the field itself (the field is defined on an open generic definition of the type).

This syncs the relevants tests with a NativeAOT work-in-progress change, but in generately it adds a lot more cases and some better comments.

### DIFF
--- a/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterWarningLocation.cs
+++ b/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterWarningLocation.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq.Expressions;
 using System.Security.Policy;
 using Mono.Linker.Tests.Cases.Expectations.Assertions;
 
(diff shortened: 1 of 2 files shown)


test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterWarningLocation.cs
```diff
@@ -61,6 +67,18 @@ class DerivedWithTwoMatching<
 				: BaseWithTwo<TPublicMethods, TPublicFields>
 			{ }
 
+			[ExpectedWarning ("IL2091")]
+			class DerivedWithOnlyStaticMethodReference<TUnknown> : BaseWithPublicMethods<TUnknown>
+			{
+				// The method body in this case looks like:
+				//     BaseWithPublicMethods<TUnknown>.GetMethods ()
+				// The type instantiation needs to be validated and in this case it produces a warning.
+				// This is no different from the same code being part of a completely unrelate method/class.
```
tlakollo, line 76:
```suggestion
				// This is no different from the same code being part of a completely unrelated method/class.
```

test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterWarningLocation.cs
```diff
@@ -115,6 +135,13 @@ public static void Test ()
 			}
 		}
 
+		// Method parameter instantiation doesn't technically need to be validated on its own
+		// Especially in AOT compiler the type of the method parameter won't event exist in the compilation
+		// unless something passes a real value to it, in which case that value creation will do all the necessary
+		// validations. So the declaration of the parameter type alone doesn't not create an analysis hole
```
tlakollo, line 141:
```suggestion
		// Especially in AOT compiler the type of the method parameter won't even exist in the compilation
		// unless something passes a real value to it, in which case that value creation will do all the necessary
		// validations. So the declaration of the parameter type alone doesn't create an analysis hole
```

tlakollo (approved):
Minor spelling stuff, LGTM

PR #3187, opened 2023-01-20
Match SDK versions of MSBuild framework

sbomer:
The version of `Microsoft.Build.Framework` that we depend on has a transitive dependency on `System.Drawing.Common` that is being flagged due to a security vulnerablility in https://github.com/dotnet/sdk/issues/29927. The dependency is probably unused, but we would like to fix it to prevent further false positive reports.

This is being referenced as a transitive dependency of `Microsoft.Build.Framework`: Microsoft.Build.Framework/17.0.0-preview-21267-01 -> System.Security.Permissions/4.7.0 -> System.Windows.Extensions/4.7.0 -> System.Drawing.Common/4.7.0. 

`dotnet/sdk` uses a lower version (15.4.8) of `Microsoft.Build.Framework`, which doesn't have a dependency on System.Drawing.Common. `dotnet/linker` used to reference the same version but we bumped it in https://github.com/dotnet/linker/pull/1840 and https://github.com/dotnet/linker/pull/2047. @marek-safar do you know why we made the upgrade? My inclination is to change back to 15.4.8, since I don't think we can actually be depending on any API surface newer than what ships with the SDK. If there's a requirement to reference a higher version, please let me know.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -14,10 +14,12 @@
     <MicrosoftNETSdkILPackageVersion>7.0.0-rtm.22507.1</MicrosoftNETSdkILPackageVersion>
     <!-- see https://github.com/dotnet/runtime/issues/1338 -->
     <MicrosoftNETCoreILAsmVersion>$(MicrosoftNETSdkILPackageVersion)</MicrosoftNETCoreILAsmVersion>
-    <!-- SRM version should match the SDK version at https://github.com/dotnet/sdk/blob/master/eng/Versions.props -->
+
+    <!-- These should match the SDK versions at https://github.com/dotnet/sdk/blob/master/eng/Versions.props -->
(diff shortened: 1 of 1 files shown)


marek-safar:
I think it was to sync with dotnet/runtime https://github.com/dotnet/runtime/blob/43b8dcf48e26019ce2aca5ae2f249001693b42c2/eng/Versions.props#L175

If this dependency versioning should probably be moved to arcade though to avoid further problems in source build.

ViktorHofer:
> Got it, thanks! Another approach would be to upgrade to the version used by dotnet/runtime, which was also recently bumped to fix component governance warnings: https://github.com/dotnet/runtime/pull/77678.

Yes that's the preferred path.

ViktorHofer:
> https://github.com/dotnet/sdk/pull/30141

Unsure. I think what they use works for source-build so it's probably fine to keep it pinned to the lower version. cc @mmitche @marcpopMSFT

PR #3192, opened 2023-01-24
Make external cecil projects build in isolation

ViktorHofer:
Updates ProjectReferences that point to the two referenced Mono.Cecil projects in order not flow the globally set RepoRoot and VersionsPropsPath properties. That guarantees that these projects are always built deterministically (VS vs dotnet build vs build.cmd).

For more context, see the offline mail conversation between @vitek-karas, @sbomer and me.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -7,6 +7,12 @@
     <CallTarget Targets="VSTest" Condition="'$(IsTestProject)' == 'true'" />
   </Target>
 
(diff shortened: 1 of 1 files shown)


ViktorHofer:
Converting to draft as builds are failing. Taking a look.

tlakollo:
I think this is not needed anymore given that dotnet/linker no longer produces the illink package and that runtime uses a different way to obtain cecil. In any case, if this is needed probably needs to be done in runtime instead.

PR #3202, opened 2023-02-17
Mark .overrides from marked methods in a copy assembly

jtschuster:
Port dotnet/runtime#82197 to dotnet/linker


sbomer:
I don't think we need to keep porting changes back from runtime (unless they are servicing fixes), now that we have started shipping packages from dotnet/runtime.

marek-safar:
This needs to target right 7.x branch instead

PR #3217, opened 2023-06-02
[release/7.0] Fix endless loop in illink analyzer

vitek-karas:
This is a port of the 8.0 change: https://github.com/dotnet/runtime/pull/86449

I verified that the repro from https://github.com/dotnet/runtime/issues/86379 is fixed with this change.

This is a 7.0 fix of https://github.com/dotnet/runtime/issues/86379

## Impact

If the right pattern occurs in the analyzed code, the analyzer will hang. This leads to hanging the build as well as causing problems with IDEs relying on a language service with the analyzer.

## Workaround

The only workaround is to disable the trim analyzer on the affected project - which means losing all of the diagnostics it provides.

## Risk

The fix is slightly risky in that it changes how some internal state of the analyzer is maintained (makes more things immutable). The fix has been merged into .NET 8 main for about 2 weeks now and we haven't seen any problems with it.

/cc @agocke

### DIFF
--- a/test/Mono.Linker.Tests/Tests/GetDisplayNameTests.cs
+++ b/test/Mono.Linker.Tests/Tests/GetDisplayNameTests.cs
@@ -191,18 +191,18 @@ namespace Mono.Linker.Tests
 		}
 
 		[DisplayName ("Mono.Linker.Tests.GetDisplayNameTests.MethodWithGenericTypeArgument(IList<GetDisplayNameTests.GenericClassOneParameter<Byte*[]>>)")]
-		public static void MethodWithGenericTypeArgument (IList<GenericClassOneParameter<byte*[]>> p)
+		public unsafe static void MethodWithGenericTypeArgument (IList<GenericClassOneParameter<byte*[]>> p)
 		{
 		}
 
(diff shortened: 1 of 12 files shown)


test/Mono.Linker.Tests/Tests/GetDisplayNameTests.cs
```diff
@@ -191,18 +191,18 @@ public class NestedGenericClassMultipleParameters<U, V>
 		}
 
 		[DisplayName ("Mono.Linker.Tests.GetDisplayNameTests.MethodWithGenericTypeArgument(IList<GetDisplayNameTests.GenericClassOneParameter<Byte*[]>>)")]
-		public static void MethodWithGenericTypeArgument (IList<GenericClassOneParameter<byte*[]>> p)
+		public unsafe static void MethodWithGenericTypeArgument (IList<GenericClassOneParameter<byte*[]>> p)
```
sbomer, line 194: Just curious - what's the reason for this change?

sbomer (approved):

test/Mono.Linker.Tests/Tests/GetDisplayNameTests.cs
vitek-karas: Compiler reported a warning on this - I assume it's me using a new version of the compiler.

PR #3232, opened 2023-10-10
[release/7.0] Port fix to avoid hangs on array value state exponential growth

vitek-karas:
This a port of https://github.com/dotnet/runtime/pull/87634. It fixes an effective hang due to state value exponential growth.

Also ports the related tests.


vitek-karas:
This is a proactive port - we don't have a current ask for this fix on 7. But the origin of https://github.com/dotnet/runtime/issues/87596 suggests that this might be encountered on 7 as well.

sbomer (approved):

vitek-karas:
Replaced by a combined PR https://github.com/dotnet/linker/pull/3234.

PR #3234, opened 2023-10-11
[release/7.0] Port of stability fixes for trimmer and anlyzer

vitek-karas:
> [!NOTE]
> This says release/7.0 but is actually a backport to 6.0 because the 7.0 linker is used for <= 7.0.

## Customer Impact

These are all crashes/hangs. Each of these has been customer reported in .NET 8, although we have only received a customer issue for the first one in .NET 7. The theory here is that these were somewhat hard to track down and led to a bad customer experience (build hangs), and if we would backport one, it's probably worth it to backport all. We can separate each of them if necessary. The analyzer crash (third in the list) will almost certainly occur if someone tries to use a new language version while sticking on an old framework. This hasn't been common yet, but we guess that it will become more so over time.

 
This ports these PR (at least the parts relevant to trimmer and analyzer) from 8.0:
* https://github.com/dotnet/runtime/pull/82818 - disallow nested array values, this prevents crashes and hangs caused by deep recursion
* https://github.com/dotnet/runtime/pull/87634 - limit the size of the value state to track during analysis. If the state would get too big, we give up on the analysis and simply ignore that value (treat it as unknown). This prevents all kinds of hangs caused by typically exponential growth of the value state.
* https://github.com/dotnet/runtime/pull/88836 - avoid throwing exceptions in the analyzer when an unrecognized code pattern is found. This avoids crashes of the analyzer (and causing build breaks) if a pattern of code we've never encountered is used (applies to new patterns introduced in new language versions as well).

See the specific commits for a bit more detail on the ports.

## Testing

The change adds tests for the first two changes. Adding tests for the 3rd is difficult (how to introduce unknown code patterns - if don't know what they are).

## Risk

Pretty low. Only to trim scenarios. All of these changes simply detect pathological cases and turn them into much simpler state which is not going to harm the analysis.

### DIFF
--- a/src/ILLink.RoslynAnalyzer/DataFlow/CapturedReferenceValue.cs
+++ b/src/ILLink.RoslynAnalyzer/DataFlow/CapturedReferenceValue.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Diagnostics;
 using ILLink.Shared.DataFlow;
 using Microsoft.CodeAnalysis;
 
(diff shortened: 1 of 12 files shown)


sbomer (approved):
LGTM, thank you!

jeffschwMSFT (approved):
approved. we will take for consideration in 7.0.x

PR #3238, opened 2023-11-27
Remove all files except for README.md

sbomer:
@ViktorHofer suggested doing this to prevent the archived sources from main showing up in code search.


ViktorHofer:
It would probably be best to also remove all commits. See corefx for an example: https://github.com/dotnet/corefx. While the repository is already archived, that's not important here.

We created an "archived" branch that only includes a commit for the README.md. We set that branch as the default branch and deleted the main branch.

sbomer:
I'm personally not too keen on deleting `main`, since I sometimes look here for the history from before we moved over to dotnet/runtime. But making a new empty `archive` branch that's the default seems fine to me. I pushed https://github.com/dotnet/linker/tree/archive. @agocke I don't have permissions - do you want to do that?

agocke:
Done

agocke (approved):

sbomer:
Thanks!

ViktorHofer:
> I'm personally not too keen on deleting main, since I sometimes look here for the history from before we moved over

That's a great use case for a tag. A branch expresses that HEAD isn't fixed and commits are being added. But I don't feel strongly about keeping the main branch around or not.

PR #3245, opened 2024-03-26
Add new official build pipeline

agocke:


eng/official.yml
```diff
@@ -0,0 +1,156 @@
...
+  repositories:
+  - repository: 1ESPipelineTemplates
+    type: git
+    name: 1ESPipelineTemplates/1ESPipelineTemplates
+    ref: refs/tags/release
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+  parameters:
+    pool:
+      name: Azure-Pipelines-1ESPT-ExDShared
```
akoeplinger: this is the wrong pool, we should be using NetCore1ESPool-Svc-Internal here.

given this is only really used for the SDLSources stage you can also use this setting:
```
  parameters:
    sdl:
      sourceAnalysisPool:
        name: NetCore1ESPool-Svc-Internal
        image: 1es-windows-2022
        os: windows
```

eng/official.yml
```diff
@@ -0,0 +1,156 @@
...
+          enablePublishBuildAssets: true
+          enableMicrobuild: true
+          jobs:
+          - job: Windows_NT
+            pool:
+              ${{ if eq(variables.officialBuild, 'false') }}:
+                vmImage: windows-latest
+              ${{ if eq(variables.officialBuild, 'true') }}:
+                name: NetCore1ESPool-Internal
+                demands: ImageOverride -equals 1es-windows-2019
```
akoeplinger: does this even work? it should have `image` and `os` keys

eng/official.yml
```diff
@@ -0,0 +1,156 @@
...
+            testResultsFormat: vstest
+          enablePublishBuildAssets: true
+          enableMicrobuild: true
+          jobs:
+          - job: Windows_NT
+            pool:
+              ${{ if eq(variables.officialBuild, 'false') }}:
+                vmImage: windows-latest
+              ${{ if eq(variables.officialBuild, 'true') }}:
+                name: NetCore1ESPool-Internal
```
akoeplinger: we should be using NetCore1ESPool-Svc-Internal for servicing branches

eng/official.yml
```diff
@@ -0,0 +1,156 @@
...
+      jobs:
+      - template: /eng/common/templates/jobs/jobs.yml@self
+        parameters:
+          enableTelemetry: true
+          helixRepo: dotnet/linker
+          enablePublishUsingPipelines: true
+          enablePublishBuildArtifacts: true
+          ${{ if eq(variables.officialBuild, 'false') }}:
+            enablePublishTestResults: true
+            testResultsFormat: vstest
```
akoeplinger: we can probably clean up the pipeline to not run tests etc

akoeplinger (commented):
is the cecil submodule update intentional?

agocke:
@akoeplinger Trying to fix it. There's a package downgrade due to the submodule referencing net40

eng/official.yml
agocke: True

sbomer (approved):
LGTM if it works

agocke:
CG is complaining, so I think it works, but we will need to update some dependencies.

agocke:
I'm going to merge this and come back for the dependency update in a second PR.

PR #3246, opened 2024-03-26
Update MSBuild reference version

agocke:


sbomer (approved):
What problem is this fixing?

agocke:
Theoretically it updates transitive nuget references. I’m still validating though

agocke:
OK, summary is that the test assets are old. The only other bump is MSBuild, but these are just reference assemblies, so they don't actually carry anything important for us.

PR #3254, opened 2024-08-20
Upgrade Microsoft.IO.Redist

agocke:


sbomer (approved):
LGTM, thanks. In general I wonder if we should be using a MicrosoftBuildFrameworkVersion matching dotnet/sdk, instead of bumping to the latest MSBuild and then having to fix up the transitive dependencies when there are cg alerts, similar to what I suggested in https://github.com/dotnet/linker/pull/3187#issue-1550187825.

agocke:
I think that makes sense for 8.0, but for 6.0 a targeted fix seems appropriate and low risk.

